Issue #36, opened 2014-12-05
Remove unnecessary exports used in only in tests


krizhanovsky:
Remove unnecessary symbol exports. See https://github.com/natsys/tempesta/commit/b0220410ad9585a964ea184b4718a503ec094ce2#commitcomment-8848675 for reference and check all other places of exported symbols.

PR #40, opened 2015-01-12
Remove unnecessary exports used in only in tests

vdmit11:
The commit introduces a new macro called DEBUG_EXPORT_SYMBOL().
It acts like the EXPORT_SYMBOL(), but only in a debug build.
In a release build such symbols are eliminated so the global symbol table
is not polluted. That is done for symbols used only in unit tests.

related to: #36


krizhanovsky:
Looks good. Commit and close #36 issue.

PR #42, opened 2015-01-22
Tempesta FW configuration framework.

vdmit11:
The commit introduces base implementation for the Tempesta FW configuration
framework and its integration with existing modules.

Basic aspects of the implementation are:
 - Configuration is stored as a plaintext file which is loaded and
   parsed right in the kernel space. That allows to pass data structures
   to the kernel.
 - Late binding is utilized, modules register their callbacks to handle
   input data.
 - A little DSL is used: modules provide their configuration specifications
   in a flexible declarative manner.
 - start/stop commands are triggered via sysctl now.
   That allows to all kernel modules be loaded before start working.


tempesta_fw/cfg.c
```diff
@@ -0,0 +1,1563 @@
...
+
+#include "cfg.h"
+
+/* FSM's debug messages are very verbose, so they are turned off by default. */
+#ifdef DEBUG_CFG_FSM
+#define FSM_DBG(...) TFW_DBG(__VA_ARGS__)
+#else
+#define FSM_DBG(...)
+#endif
+
```
vdmit11: AIexander, I am not sure that we need distinct debug levels like TFW_DBG0(), TFW_DBG1(), etc because it would be hard to analyze such log messages because tons of messages from all modules mixed together.
It turns out, at least for me, that it is more handy to turn on and off debug messages for separate modules than switch verbosity levels, so I left these stubs here.

tempesta_fw/addr.c
```diff
@@ -44,48 +45,60 @@ validate_addr(const TfwAddr *addr)
 }
 
 static int
-tfw_inet_pton_ipv4(char **p, struct sockaddr_in *addr)
+tfw_inet_pton_ipv4(const char *pos, struct sockaddr_in *addr)
```
vdmit11: As far as I can remember, the old function was not able things like ":80", so I wrote a new one. That turned out to be faster than enhancing the old implementation.

tempesta_fw/cfg.h
```diff
@@ -0,0 +1,285 @@
...
+	const char *name;	/* [A-Za-z0-9_], starts with a letter. */
+	TfwCfgSpec *specs;	/* An array terminated by a null element. */
+	int  (*init)(void);
+	void (*exit)(void);
+	int  (*setup)(void);
+	int  (*start)(void);
+	void (*stop)(void);
+	void (*cleanup)(void);
+} TfwCfgMod;
+
```
vdmit11: The complicated system of setup/start/stop/cleanup callbacks is actually useful.
For example, look at the tfw_sched_http.c implementation.
Without the setup/cleanup callbacks it would be hard to allocate/free memory for parsed configuration.

tempesta_fw/addr.c
krizhanovsky: Never rewrite code just because you don't understand it and can't properly adjust! I'm always happy to explain unclear pieces of code.

Moreover, the new implementation is poor since it examines the same data many times. Throw it out and fix the previous one.

tempesta_fw/addr.c
```diff
@@ -205,87 +221,24 @@ tfw_inet_pton_ipv6(char **p, struct sockaddr_in6 *addr)
...
-		else if (*p1 == '.') {
-			mode = 4;
-		}
-		else {
-			TFW_ERR("bad string: %s\n", *p);
-			return -EINVAL;
-		}
-	}
+	/* The IPv6 address must be enclosed into square brackets,
+	 * or else we can't distinguish it from the port. */
```
krizhanovsky: Read rfc5956 chapter 6 "Notes on Combining IPv6 Addresses with Port Numbers". The same as above: throw out the new and wrong code and adjust the previous right code. The previous code IIRC does not handle all the cases specified by the RFC, but it did the things better.

tempesta_fw/addr.c
```diff
@@ -205,87 +221,24 @@ tfw_inet_pton_ipv6(char **p, struct sockaddr_in6 *addr)
...
-/* TODO: Remove the export after implementing a configuration framework
- * that will remove the dependency of tfw_sched_http on this function. */
 EXPORT_SYMBOL(tfw_inet_pton);
 
-/**
- * Print UPv4/IPv6 address in network byte order to @buf.
- * @buf must be MAX_ADDR_LEN bytes in size.
- */
-int
-tfw_inet_ntop(const void *addr, char *buf)
```
krizhanovsky: This is useful function at least for logging. Don't throw it out.

tempesta_fw/addr.c
```diff
@@ -565,3 +518,19 @@ tfw_addr_fmt(const TfwAddr *addr, char *out_buf, size_t buf_size)
...
+
+ssize_t
+tfw_addr_sa_len(const TfwAddr *addr)
+{
+        sa_family_t family;
+
+        BUG_ON(!addr);
+
+        family = addr->sa.sa_family;
+        if (unlikely(family_is_not_supported(family))) {
```
krizhanovsky: I just had a look at family_is_not_supported() and validate_addr() once more. At which circumstances do we recieve a packet with unsupported family? I believe ip_rcv()/ipv6_rcv(), the earliest point at which Tempesta works, won't get the packet if it has wrong family. So the check is mostly about programming error. So it must be optimized out w/o DEBUG option.

This validate_addr() must use BUG_ON() rather than WARN_ON().

Regarding this code place: there is no need to inspect the field twice, in amily_is_not_supported() and the below.

tempesta_fw/cfg.h
```diff
@@ -0,0 +1,285 @@
...
+int tfw_cfg_check_single_val(const TfwCfgEntry *e);
+
+int tfw_cfg_parse_children(TfwCfgSpec *self, TfwCfgEntry *parsed_entry);
+
+int tfw_cfg_set_bool(TfwCfgSpec *self, TfwCfgEntry *parsed_entry);
+
+int tfw_cfg_set_int(TfwCfgSpec *spec, TfwCfgEntry *parsed_entry);
+
+typedef struct {
+	int is_multiple_of;
```
krizhanovsky: Usially "is_" prefix means boolean type. However, you use it as integer (e.g. cache.c).

tempesta.sh
```diff
@@ -2,20 +2,23 @@
...
-TFW_CACHE_PATH=$TFW_ROOT/cache
+root=$(dirname "$0")
 
 arg=${1:-}
-ss_path=${SYNC_SOCKET:="./"}
-tdb_path=${TDB:="./"}
+ss_path=${SS_PATH:="$root/sync_socket"}
+tdb_path=${TDB_PATH:="$root/tempesta_db"}
+tfw_path=${TFW_PATH:="$root/tempesta_fw"}
+tfw_cfg_path=${TFW_CFG_PATH:="$root/tempesta_fw.conf"}
```
krizhanovsky: It would be very useful to place default tempesta_fw.conf near to the script. The config should also have verbose comments about all possible configuration optins - the same way as Apache and other good Open Source do it.

tempesta_fw/cfg.c
krizhanovsky: Ok, makes sense. However, usually people
1. prepend mudule specific line to the logging message, like "cfg" for this case, so they can differentiate messages from different modules;
2. use only one environment/build variable with debug level. Will you create 5 different variables and how will you name them if you need 5 different levels of debug messages?

Usually, different debug levels just adds extra spaces/tabs at begin of the message, so we get something like

       simple log message
         debug message
           debug_1 message
             debug_2 message
       other normal message

so the log is easy to examine (and automatically process!) for normal and different level debug messages.

tempesta_fw/cache.c
```diff
@@ -523,14 +530,44 @@ tfw_cache_init(void)
...
+		}
+	},
+	{}
+};
+
+TfwCfgMod tfw_mod_cache = {
+	.name 	 = "cache",
+	.start	 = tfw_cache_start,
+	.stop	 = tfw_cache_stop,
+	.specs	 = tfw_cache_cfg_specs,
```
krizhanovsky: Please initialize structure members in the same order as they are declared - it makes the code easier to read.

tempesta_fw/cfg.h
krizhanovsky: Yeah... But I'm still confused by the fact that any of the existing modules doesn't use setup and start callbacks at the same time. So it seems that they can be aggregates. Also why do we need stop and cleanup at the same time? Why we shouldn't call cleanup routine at the stop?

Meantime, it's obvious that we need to separate init/exit from start/stop. However, why setup/cleanup are separate from start/stop is not clear. Please explain.

tempesta_fw/cfg.c
vdmit11: > 1. prepend mudule specific line to the logging message, like "cfg" for this case, so they can differentiate messages from different modules;

Yea, but then I have to grep logs every time, and do that in a tricky way: I have to remove unwanted messages instead of enabled messages.

Suppose I want to investigate the "cfg" module. Then I need a log containing debug messages only for the "cfg" module, plus regular non-debug messages from all modules. Then I can't simply do `grep cfg` because it wipes out messages from other modules, and I can't simply do `grep -v mod1 | grep -v mod2 | etc` because it filters out non-debug messages. 

I know that could be handled by tricky message formatting and clever scripts, but for me that looks unhandy at the moment.

> 2. use only one environment/build variable with debug level. Will you create 5 different variables and how will you name them if you need 5 different levels of debug messages?

Yes, there will be 5 different entities (not variables but compile-time switches).

But you can't avoid that. If you need to distinguish messages, then you have to name them somehow. Perhaps you choose to prepend some prefixes to messages (like "tfw_cfg: blah" or "tfw_sched_http: blah") and then `grep` them. That actually doesn't make much difference since you have to create 5 prefixes and agree on their naming. 

Compile-time switches look handy for me. Debugging is a part of every-day programming process. I work with the source code all the time, so for me it is worthless to change a single line and run Tempesta FW again, and get a readable text log without additional scripts. I don't want to complicate this process.

Actually this way is only applicable if you can reproduce a bug easily on your machine. Things go complicated when you met a heisenbug or you have to use a remote setup to reproduce the bug. I suggest to postpone the work until such case.

tempesta_fw/cfg.h
vdmit11: Well, the main trick is:
  - `setup` is executed before parsing the configuration and invoking TfwCfgSpec callbacks.
  - `start` is handled after the configuration is fully parsed.

If you leave just one callback, then you loose ability to hook a callback both before and after configuration parsing. And that is the problem, because sometimes you need both points. Some actions (memory allocation) have to be done before starting configuration parsing process, and some actions should be done only after all settings are parsed.

Now, suppose that an error happened in the middle the configuration parsing process. Some memory for parsed configuration is already allocated. How would you free it? You can't call the `stop` callback because nothing is started yet, because `start` has to be invoked only when parsing is done. See the problem? I can't suggest any simple solution other than the `cleanup` callback.

I agree that these callbacks are odd. I'm thinking about separating them in the following way:
  - `init/exit/start/stop` are about the Tempesta FW state, so they should be left
  - `setup/cleanup` are more about configuration parsing, so they should be somehow renamed or moved to a some relevant place

At some point we will need to separate them somehow to implement hot configuration re-loading without stopping Tempesta FW (imagine adding a new back-end server machine to a setup or so). Then we will need to change `setup/cleanup` in a way to call them independently from `start/stop` or so.

As for now, I suggest to leave the callback. Perhaps I should rename them to something like `cfg_setup`/`cfg_cleanup` or even `cfg_alloc`/`cfg_free`?

tempesta_fw/addr.c
vdmit11: Now we have a faster `tfw_addr_fmt()` that does the same thing but without `snprintf()`.
The functions were different, but at some point they started doing the same thing.
In this patch I removed the last reference to the old `tfw_inet_ntop()` so I removed the function.

Also, there is a little naming inconsistency here: `tfw_inet_*()` vs `tfw_addr_*()`.
For me `tfw_addr_*()` looks more suitable for the `addr.c` and `TfwAddr`.
Should I re-name the old `tfw_inet_` functions?

tempesta_fw/addr.c
```diff
@@ -565,3 +518,19 @@ tfw_addr_fmt(const TfwAddr *addr, char *out_buf, size_t buf_size)
 	return (pos - out_buf);
```
krizhanovsky: tfw_addr_fmt(): double check for supported familiy, in validate_addr() and family_is_not_supported(). Please inspect all the calls of the check and remove unnecessary calls.

tempesta_fw/addr.c
krizhanovsky: Ok, good. spnprintf() is really bad

tempesta.sh
vdmit11: Whoops, the default .conf file with comments was exactly there, but it seems I deleted it together with the old file.
Have to write it from the blank.

tempesta_fw/cfg.c
krizhanovsky: Dmitry, all what I wrote is just a common practice for logging. If you need messages of certain level for cfg module, then you just compile the module with CFG_DEBUG=3, where 3 is the desired level of debugging, and grep log for the interesting messages like ```dmesg | grep 'cfg\|othe_mod'```

tempesta_fw/cfg.h
krizhanovsky: I still don't understand the problem. If an error happened in the middle the configuration parsing process while we're scanning list of registered modules, then why not just to call ```stop()``` callbacks for all modules which already started? It's their responsibility to free all allocated memory. If configuration subsystem allocates some memory, then it should be freed at tend of ```tfw_cfg_start_mods()```.

tempesta_fw/cfg.c
```diff
@@ -0,0 +1,1563 @@
...
+	static ctl_table tfw_cfg_sysctl_tbl[] = {
+		{
+			.procname	= "state",
+			.data		= tfw_cfg_sysctl_state_buf,
+			.maxlen		= sizeof(tfw_cfg_sysctl_state_buf) - 1,
+			.mode		= 0644,
+			.proc_handler	= handle_sysctl_state_io,
+		},
+		{}
+	};
```
krizhanovsky: Usually in Linux kernel static tables are placed out of functions.

tempesta_fw/cfg.c
```diff
@@ -0,0 +1,1563 @@
...
+/**
+ * Process command received from sysctl as string (either "start" or "stop").
+ * Do corresponding actions, but only if the state is changed.
+ */
+static int
+handle_state_change(const char *old_state, const char *new_state)
+{
+	bool is_changed = strcasecmp(old_state, new_state);
+	bool is_start = !strcasecmp(new_state, "start");
+	bool is_stop = !strcasecmp(new_state, "stop");
```
krizhanovsky: It seems new_state can't have "start" and "stop" line at the same time. However, you process is_start and is_stop variable as independent in the code. Rather, you should have single variable with ```start``` or ```stop``` value and process it w/o double checking in if statements below.

tempesta_fw/cfg.c
```diff
@@ -0,0 +1,1563 @@
...
+
+	/* The buffer where the whole configuration is stored.
+	 * FIXME: use vmalloc() with the size of the configuration file. */
+	static char cfg_text_buf[65536];
```
krizhanovsky: Yes, really, just call vmalloc() - it's as easy as to write the comment ;)

tempesta_fw/cfg.c
```diff
@@ -0,0 +1,1563 @@
...
+	TFW_DBG("stopping modules...\n");
+
+	MOD_FOR_EACH_REVERSE(mod, mod_list) {
+		MOD_CALL(mod, stop);
+	}
+
+	MOD_FOR_EACH_REVERSE(mod, mod_list) {
+		MOD_CALL(mod, cleanup);
+	}
+
```
krizhanovsky: extra empty line

tempesta_fw/cfg.c
```diff
@@ -0,0 +1,1563 @@
...
+	const char **dest_str = cs->dest;
+	TfwCfgSpecStr *cse = cs->spec_ext;
+	int r, len, len_min, len_max;
+
+	r = tfw_cfg_check_single_val(e);
+	if (r)
+		return r;
+
+	BUG_ON(!dest_str);
+	BUG_ON(!cs);
```
krizhanovsky: What's the sense of the check if you dereference ```cs``` above?

tempesta_fw/cfg.c
```diff
@@ -0,0 +1,1563 @@
...
+	len_min = cse->len_range.min;
+	len_max = cse->len_range.max;
+	if (len_min != len_max && (len < len_min || len > len_max)) {
+		TFW_ERR("the string length (%d) is out of valid range (%d, %d):"
+			" '%s'\n", len, len_min, len_max, in_str);
+		return -EINVAL;
+	}
+
+	memcpy(cse->buf.buf, in_str, len + 1);
+	*dest_str = cse->buf.buf;
```
krizhanovsky: It seems something wrong is going with dest_str. Why do you need the variable and why it's a double pointer? Why not just to call ```cse->dest = cse->buf.buf``` here?

tempesta_fw/cfg.c
```diff
@@ -0,0 +1,1563 @@
...
+}
+
+int
+tfw_cfg_set_int(TfwCfgSpec *cs, TfwCfgEntry *e)
+{
+	int base, r, val;
+	int *dest_int;
+	const char *in_str;
+
+
```
krizhanovsky: extra empty line

tempesta_fw/cfg.h
vdmit11: We can't call `stop()` because nothing is `start()`'ed at this point. The `start()` can only be called only when the whole configuration file is parsed.

Consider this configuration:

    listen :80;
    backend :8080;
    listen [::1]:80;

`listen` belongs to the `sock_frontend` and `backend` belongs to `sock_backend` respectively. Suppose you are parsing this configuration text. How do you know the point, when `sock_frontend` should be started? I can't until I parse the last entry.

So we start all modules at once after the parsing is finished. So in the middle of the parsing process no `start()` callback is called yet.

Usually you think of `stop()` as the reverse of `start()`. There is natural temporal coupling between these callbacks. You don't expect the `stop()` to be called when everything is already stopped.

And if you decide to violate that, then in every `stop()` you have to check whether you are started or not. For me it looks better to keep separate `setup()`/`cleanup()` callbacks there, given that they will be useful in future for hot configuration re-loading while the system is running.

tempesta_fw/sched/tfw_sched_http_rules_example.conf
```diff
@@ -1,22 +0,0 @@
...
-}
-
-hdr_conn = Keep-Alive {
-	127.0.0.5:8080 127.0.0.6:8080
-}
-
-uri ^ / {
-	127.0.0.1:8080
-	127.0.0.2:8080
-	127.0.0.3:8080
```
krizhanovsky: Why the default config file is just removed?

tempesta_fw/sock_frontend.c
```diff
@@ -33,132 +34,152 @@
...
-	up_read(&tfw_cfg.mtx);
-	return r;
+static void
+stop_listen_socks(void)
+{
+	struct socket *sock;
+	int i;
+
+	FOR_EACH_SOCK(sock, i) {
+		kernel_sock_shutdown(sock, SHUT_RDWR);
```
krizhanovsky: Did you check that the function works with synchronous sockets? IIRC it doesn't work due to special socket callbacks. Moreover, SS has designated ss_close().

tempesta_fw/sock_frontend.c
```diff
@@ -33,132 +34,152 @@
...
-	return r;
-err:
-	sock_release(s);
-	return r;
-}
-
-void
-tfw_close_listen_sockets(void)
-{
-	down_read(&tfw_cfg.mtx);
```
krizhanovsky: It seems there is not synchronization with the new configuration subsystem. How two concurrent configuration updates are synchronized?

tempesta_fw/main.c
```diff
@@ -19,103 +20,52 @@
...
-		goto err_connection;
-
+	}
+
+	INIT_TFW_MOD(tfw_mod_cache);
+	INIT_TFW_MOD(tfw_mod_http);
+	INIT_TFW_MOD(tfw_mod_server);
+	INIT_TFW_MOD(tfw_mod_client);
+	INIT_TFW_MOD(tfw_mod_session);
+	INIT_TFW_MOD(tfw_mod_connection);
```
krizhanovsky: It seems connection doesn't have it's own configuration, so it should not be treated as TfwCfgMod. Just call it's tfw_connection_init() function as normal. Please check all other calls. Additional unresonable complexity is always bad.

tempesta_fw/http.c
```diff
@@ -945,22 +961,13 @@ tfw_http_init(void)
...
-{
-	return (tfw_hash_str(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST].field) ^
-	        tfw_hash_str(&req->uri));
-}
-EXPORT_SYMBOL(tfw_http_req_key_calc);
+TfwCfgMod tfw_mod_http = {
+	.name = "http",
+	.init = tfw_http_init,
+	.exit = tfw_http_exit,
+};
```
krizhanovsky: Why http.c is also treated as TfwCfgMod? It's not a good idea to declare TfwCfgMod for each C file.

tempesta_fw/http_match.c
```diff
@@ -64,6 +64,14 @@
...
 #include "http.h"
 
+/* Matching every request against a long table of rules may produce a lot of
+ * debugging messages, so they are turned off by default. */
+#ifdef DEBUG_HTTP_MATCH
+#define MATCH_DBG(...) TFW_DBG(__VA_ARGS__)
+#else
+#define MATCH_DBG(...)
+#endif
+
```
krizhanovsky: Oh, custom debug messages again, ever in main core module. This is messy to have own debug infrastructure for each C file.

tempesta_fw/ptrset.h
```diff
@@ -0,0 +1,296 @@
...
+ * distinct objects like a list of back-end servers for the Tempesta FW.
+ *
+ * Advantages over a linked list are:
+ *  - No need for augmenting the target data structure with sturct list_head.
+ *    This is useful when you want to make a list of objects in an external
+ *    kernel module, but you don't want the main code to know about your module.
+ *  - Lightweight operations.
+ *    Although the asymptotic behavior is bad (O(N) for most operations), the
+ *    code is good at constant factors, so the performance should be slightly
+ *    better for small number of elements.
```
krizhanovsky: It's hard to say whether the pointers are the right solution w/o any application in the code. The last factor says that they should not be at least in the patch set.

The fadventages are doubtful - we usually use intrusive lists to get structures stored in the list in the cache while we're traversing the list, so we get the hot data when we find the required item.

tempesta_fw/ptrset.h
```diff
@@ -0,0 +1,296 @@
+/**
+ *		Tempesta FW
```
krizhanovsky: Why the things is in the patch set? How does it linked with configuration?

tempesta_fw/ptrset.h
```diff
@@ -0,0 +1,296 @@
...
+ *   size_t current_addr_idx;
+ *   TfwAddr *current_addr;
+ *   TfwMyAddrSet addresses;
+ *
+ *   tfw_ptrset_for_each(current_addr, current_addr_idx, addresses) {
+ *           printk("%d\n", current_addr->family);
+ *   }
+ */
+#define tfw_ptrset_for_each(ptr, i, s) \
+	for ((i) = 0; ((i) < (s)->n && (ptr = (s)->ptrs[i])); ++(i))
```
krizhanovsky: Typically foreach is faster for intrusive lists in comparison with such pointer lists.

tempesta_fw/ptrset.h
```diff
@@ -0,0 +1,296 @@
...
+ * Safe for use on multiple CPUs because of the atomic operations.
+ * BUG: atomic_t contains signed int which is overflown.
+ *      Perhaps we should use a regular unsigned int instead of atomic_t here?
+ *      That fixes the bug, and also makes the code faster, but less fair.
+ */
+#define tfw_ptrset_get_rr(s) \
+	__tfw_ptrset_get_rr(__tfw_ptrset_cast(s), __tfw_ptrset_max(s))
+
+static inline void *
+__tfw_ptrset_get_rr(TfwPtrSet *s, int max)
```
krizhanovsky: PtrSet is very specific. The whole it's sens in this function and it's very specific for that request scheduler. So it seems it has no sense to move it to separate library-like file.

tempesta_fw/sched/tfw_sched_http.c
```diff
@@ -1,7 +1,78 @@
 /**
  *		Tempesta FW
  *
+ * Tempesta HTTP scheduler (load-balancing module).
```
krizhanovsky: The changes in the scheduler have nothing in common with configuration framework and must be splitted to separate patch.

tempesta_fw/tempesta.h
```diff
@@ -18,6 +19,7 @@
  * Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  */
 #ifndef __TEMPESTA_H__
+
```
krizhanovsky: Why and what for?

tempesta_fw/sock_frontend.c
```diff
@@ -33,132 +34,152 @@
...
-		}
+	SsProto *proto;
+	struct socket *sock;
+	int i, r;
 
+	FOR_EACH_SOCK(sock, i) {
 		/*
 		 * TODO If multiprotocol support is required, then here we must
 		 * have information for which protocol we're establishing
 		 * the new listener. So TfwAddrCfg must be extended with
```
krizhanovsky: Please update the comment as well - we have no TfwAddrCfg any more.

tempesta_fw/tempesta.h
vdmit11: An accident.

tempesta_fw/cache.c
```diff
@@ -523,14 +530,44 @@ tfw_cache_init(void)
...
 	destroy_workqueue(cache_wq);
 	kmem_cache_destroy(c_cache);
 	kthread_stop(cache_mgr_thr);
-	tdb_close(db);
 }
+
+static TfwCfgSpec tfw_cache_cfg_specs[] = {
+	{
+		"cache", "off",
+		tfw_cfg_set_bool, &cache_cfg.cache
```
krizhanovsky: Please don't place two statements at the same line - it makes the code harder to read

tempesta_fw/cache.c
```diff
@@ -523,14 +530,44 @@ tfw_cache_init(void)
...
+	},
+	{
+		"cache_size", "262144",
+		tfw_cfg_set_int, &cache_cfg.db_size,
```
krizhanovsky: The same - why handler initializer is at the same line as default value?...

PR #45, opened 2015-02-09
 Implement basic back-end server failovering. 

vdmit11:
The change introduces a small set of parallel connections to each
TfwServer. They are switched in a round-robin manner when a message
is scheduled. When a connection is closed, another is chosen and the
closed one is re-connected in background.

Also many connections per server implies that TfwSession doesn't link
TfwClient and TfwServer anymore. Instead, it links client/server sockets.


tempesta_fw/Makefile
```diff
@@ -42,9 +42,8 @@ tempesta_fw-objs = \
 	main.o \
 	pool.o \
 	sched.o \
-	server.o \
+	server_be.o \
```
keshonok: Is there any reason for not calling it something like ```server_backend```?
Also, I am looking for name consistency. We used to have ```sock_backend``` and ```sock_frontend```. Now we have ```server_be`` and ```sock_frontend```. This is quite confusing.

tempesta_fw/cfg.c
```diff
@@ -888,6 +888,73 @@ tfw_cfg_map_enum(const TfwCfgEnum mappings[],
...
+}
+EXPORT_SYMBOL(tfw_cfg_check_multiple_of);
+
+/**
+ * Check that the entry @e has exactly @val_n values.
+ */
+int
+tfw_cfg_check_val_n(const TfwCfgEntry *e, int val_n)
+{
+	if (e->val_n != e->val_n) {
```
keshonok: This is a bug. Must be ```if (e->val_n != val_n)```.

tempesta_fw/cfg.c
```diff
@@ -888,6 +888,73 @@ tfw_cfg_map_enum(const TfwCfgEnum mappings[],
...
+
+/**
+ * Check that integer is in specified range.
+ * Print an error and return non-zero if it is out of range.
+ */
+int
+tfw_cfg_check_range(long value, long min, long max)
+{
+	if (min != max && (value < min || value > max)) {
+		TFW_ERR("the value %ld is out of range: [%ld, %ld]\n",
```
keshonok: I am not sure I like this. This kind of small utility/helper functions should just return the result, and all error processing and reporting should be done by the caller. I can see why you did it this way, precisely to avoid that. Still, my opinion is that this is not the way it should be.

tempesta_fw/cfg.c
```diff
@@ -888,6 +888,73 @@ tfw_cfg_map_enum(const TfwCfgEnum mappings[],
...
+
+/**
+ * Check that integer @value is a multiple of @divisor (print an error
+ * otherwise);
+ */
+int
+tfw_cfg_check_multiple_of(long value, int divisor)
+{
+	if (divisor && (value % divisor)) {
+		TFW_ERR("the value of %ld is not a multiple of %d\n",
```
keshonok: See my comment above, same here.

tempesta_fw/client.c
```diff
@@ -46,8 +46,6 @@ tfw_destroy_client(struct sock *s)
 	conn->hndl = NULL;
 
 	kmem_cache_free(cli_cache, cli);
-
-	conn->sk_destruct(s);
```
keshonok: I don't see any comment in the commit as to why this is incorrect and need to be removed.

tempesta_fw/connection.c
```diff
@@ -109,12 +109,9 @@ tfw_connection_new(struct sock *sk, int type, void *handler,
 	}
 
 	sk->sk_user_data = conn;
-
-	conn->sk_destruct = sk->sk_destruct;
```
keshonok: Why don't we need to call the original socket destruct function?

tempesta_fw/rrptrset.h
```diff
@@ -123,13 +123,31 @@ tfw_ptrset_init(TfwRrPtrSet *s, int max)
 }
 
 static inline int
+tfw_ptret_vacant_n(TfwRrPtrSet *s)
```
keshonok: ```tfw_ptrset_vacant_n```, perhaps?

tempesta_fw/cfg.c
```diff
@@ -888,6 +888,73 @@ tfw_cfg_map_enum(const TfwCfgEnum mappings[],
 EXPORT_SYMBOL(tfw_cfg_map_enum);
 
 /**
+ * Get value of attribute with name @attr_key.
+ * Return @default_val if the attribute is not found in the entry @e.
+ */
+const char *
+tfw_cfg_get_attr(const TfwCfgEntry *e, const char *attr_key,
+		 const char *default_val)
```
keshonok: I see this a lot in the code. From readability point of view, it's easier on an eye when this is written in way where the second line is longer.
```c
tfw_cfg_get_attr(const TfwCfgEntry *e,
                 const char *attr_key, const char *default_val)
```

When some arguments at the end of the list are connected or similar/uniform, then an alternative is something like this:
```c
tfw_cfg_get_attr(const TfwCfgEntry *e, const char *attr_key,
                                       const char *default_val)
```

Both are cleaner and better readable, as well as easier and faster understood.

tempesta_fw/cfg.c
```diff
@@ -888,6 +888,73 @@ tfw_cfg_map_enum(const TfwCfgEnum mappings[],
...
+{
+	const char *key, *val;
+	size_t i;
```
keshonok: This is not as pronounced in the code as the above, but still, it's everywhere as well. Shorter and simpler lines with variables should come first, and longer lines should come last. Unless it's necessary to do otherwise. The reasoning is the same, it's easier to read and comprehend that way. I would put it this way:
```c
size_t i;
const char *key, *val;
```

tempesta_fw/server_be.c
```diff
@@ -0,0 +1,448 @@
...
+		spin_unlock_bh(&besrv->conn_pool_lock);
+
+		while (conn_needed) {
+			sk = tfw_be_srv_connect(besrv);
+			if (!sk) {
+				all_connected = false;
+				break;
+			}
+
+			spin_lock_bh(&besrv->conn_pool_lock);
```
keshonok: I would appreciate a comprehensive comment - why _bh() locks are necessary here.

tempesta_fw/server_be.c
```diff
@@ -0,0 +1,448 @@
...
+ * missing connection.
+ *
+ * If a connection attempt is failed, instead of sleeping forever, it wakes up
+ * periodically and makes another attempt (with unlimited number of retries).
+ */
+static int
+tfw_be_connd_mainloop(void *data)
+{
+	long timeout;
+	int err;
```
keshonok: Both should probably be made local to ```do {} while()``` block.

tempesta_fw/rrptrset.h
```diff
@@ -123,13 +123,31 @@ tfw_ptrset_init(TfwRrPtrSet *s, int max)
...
+
+static inline int
 tfw_ptrset_add(TfwRrPtrSet *s, void *ptr)
 {
 	if (__tfw_ptrset_find_idx(s, ptr) > 0) {
 		TFW_ERR("Can't add ptr %p to set %p - duplicate ptr\n", ptr, s);
 		return -1;
 	}
-	else if (s->n >= s->max) {
+	else if (!tfw_ptret_vacant_n(s)) {
```
keshonok: Please, place ```else if ()``` on the same line as the closing curly brace.

tempesta_fw/session.h
```diff
@@ -34,6 +34,7 @@
  * session.
  */
 typedef struct {
+	struct sock 	*srv_sk;
```
keshonok: Please comment in the code why this should be outside of ```TfwServer{}``` where it belongs on a first thought.

tempesta_fw/Makefile
vdmit11: Yea, perhaps `server_backend` is better.
I just used `be` because I saw it several times across the project as an abbreviation for 'backend'.

I agree that `server_be` and `sock_frontend` are confusing.
Before these changes, there were separate `sock_backend` and `server` units.
`sock_backend` was responsible for connecting to back-end server, and `server` was responsible for `TfwServer` implementation.
And after these changes, I can't separate them in such way anymore because now there is a pool of connections, and basically almost all functions go to `sock_backend.c` and almost nothing is left for `server.c`.
I failed to make these units independed, so I merged them into the single unit.
Also, I can't name the new unit `sock_backend` because it is not only about connecting to back-end servers, but also about providing a specific `TfwServer` implementation. And I can't name it `server.c` because it is not abstract enough for the generic `TfwServer`, it is a kind of specific implementation of it.
I hope you understood my dilemma.

tempesta_fw/cfg.c
vdmit11: Well, I jsut wanted to put all error messages of the configuration parser to the single place, where it is easy to inspect and edit them. 

That is done because these error messages will be seen by end user. I mean, they are not some debugging messages for developers only, I think they will be pretty common messages seen by system administrators who edit configuration files. So likely we would like to work carefully on them in future. Also, by placing error messages here, we have aceess to parser internals and thus print additional verbose info: file/line/position/etc.

In addition, these functions reduce duplicate logic a little bit: this particular function `tfw_cfg_check_range()` is used in `tfw_be_cfg_add_srv()` and in `tfw_cfg_set_int()`. For me it is better to avoid such duplicate error messages because they are subject for change as I described above.

Hope you got my motivation.

tempesta_fw/connection.c
vdmit11: Because re-placing `sk_destruct()` with a custom handler and then calling the original `sk_destruct()` function was a dirty solution. The `sk->sk_destruct` callback is usually occupied with a specific destructor of linux TCP/IP stack, like `inet_release()`. What we did by hooking up the callback is kinda monkey patching, the `sk_destruct` is intended for use with lower level code, and shall not be touched by us.

Now, instead of re-placing the `sk_destruct` and calling it manually later, we have our custom callback in `TfwConnection` which is called from `tfw_connection_close()`. Basically it is cleaner because we don't have to call original `sk_destruct` now, and it doesn't violate encapsulation of lower-level code.

Actually, this little fix is not required for this pull-request. Originally it was, because the callback returned a value that allowed to not deallocate the socket, but then I changed the implementation and the return value was not required anymore, but the callback is survived. I left it because it is a little bit cleaner than calling `sk_destruct` as I described above.

keshonok:
This merge request includes code that had been moved to a separate patch: `switch from sk_destruct to TfwConnection->destruct`

vdmit11:
I think I should make some more changes here: do few fixes, clean up the code and re-organize commits for easier review. I'll hold the pull request for a while.

etc/tempesta_fw.conf
```diff
@@ -42,50 +76,125 @@
 # Default:
 #   listen 80;
 
-# TAG: cache
+
+# TAG: backend
```
krizhanovsky: We have forward proxy in plans, so let's do not stick to poor reverse proxy in our design and code. So let's move from 'backend' to 'server' hereafter.

etc/tfw_sched_http.conf
```diff
@@ -1,157 +0,0 @@
-#
```
krizhanovsky: Did we considered to merge the config with main config? Probably, it's better to keep it separate to separate reverse proxy logic.

tempesta_fw/addr.c
```diff
@@ -515,3 +515,4 @@ tfw_addr_sa_len(const TfwAddr *addr)
         validate_addr(addr);
         return (addr->family == AF_INET6) ? sizeof(addr->v6) : sizeof(addr->v4);
 }
+EXPORT_SYMBOL(tfw_addr_sa_len);
```
krizhanovsky: why not to move to header file instead?

tempesta_fw/cfg.c
krizhanovsky: Agree with Alexey. What 'the value 20 is out of range: [0, 10]' will say in logs? What if we have many items with the value '20' in configuration? Rather we should see something like 'the value of feature_foo (20) is out of range: [0, 10]', so now we know that feature_foo is poorly configured.

tempesta_fw/connection.c
```diff
@@ -119,6 +120,7 @@ tfw_connection_new(struct sock *sk, int type, void *handler,
 
 	return 0;
 }
+EXPORT_SYMBOL(tfw_connection_new);
```
krizhanovsky: It seems lb/ should not be designed as standalone modules, so the function should not be exported. See the below for about tfw_lb_srv_connect().

tempesta_fw/lb/tfw_lb_http.c
```diff
@@ -0,0 +1,1091 @@
...
+	tfw_lb_connd_should_wakeup = true;
+	wake_up(&tfw_lb_connd_wakeup_wq);
+}
+
+/**
+ * Establish a single connection to the back-end server (but don't yet add it
+ * to the server's connection pool).
+ */
+static struct socket *
+tfw_lb_srv_connect(TfwLbSrv *srv)
```
krizhanovsky: The function is mostly copy-paste of tfw_backend_connect() while the logic of the last one is just a special case of tfw_lb_srv_connect(): we can have only one or many connections to one particular server.

tempesta_fw/lb/tfw_lb_dummy.c
```diff
@@ -0,0 +1,204 @@
...
+
+#include "addr.h"
+#include "lb_mod.h"
+#include "log.h"
+#include "connection.h"
+#include "tempesta.h"
+
+MODULE_AUTHOR(TFW_AUTHOR);
+MODULE_DESCRIPTION("Tempesta FW dummy load balancer");
+MODULE_VERSION("0.0.1");
```
krizhanovsky: We start versioning at 0.1.0. See http://semver.org/

tempesta_fw/lb/tfw_lb_dummy.c
```diff
@@ -0,0 +1,204 @@
...
+}
+
+static void
+tfw_lb_dummy_close_cb(struct sock *sk)
+{
+	TFW_ERR_ADDR("backend connection is closed", &tfw_lb_dummy_be_addr);
+}
+
+static int
+tfw_lb_dummy_start(void)
```
krizhanovsky: Copy paste of tfw_lb_srv_connect() and tfw_backend_connect().

tempesta_fw/lb/tfw_lb_dummy.c
```diff
@@ -0,0 +1,204 @@
...
+static int
+tfw_lb_dummy_send_msg(TfwMsg *msg)
+{
+	struct sock *sk;
+
+	BUG_ON(!tfw_lb_dummy_be_sock);
+	sk = tfw_lb_dummy_be_sock->sk;
+
+	if (sk->sk_state != TCP_ESTABLISHED) {
+		TFW_ERR_ADDR("not connected to backend", &tfw_lb_dummy_be_addr);
```
krizhanovsky: Is this really an error and we can't handle it? I guess we should just print a warning here and try to reestablish the connection....

tempesta_fw/connection.c
```diff
@@ -180,9 +182,8 @@ tfw_connection_send_srv(TfwSession *sess, TfwMsg *msg)
 	BUG_ON(srv_conn->sess && srv_conn->sess != sess);
 	srv_conn->sess = sess;
 
-	ss_send(sess->srv->sock, &msg->skb_list);
-
-	return 0;
+	/* XXX: will be refactored. */
```
krizhanovsky: What exactly we need to rework here? Why it can't be done now?

tempesta_fw/connection.c
```diff
@@ -180,9 +182,8 @@ tfw_connection_send_srv(TfwSession *sess, TfwMsg *msg)
...
+	return tfw_lb_send_msg(msg);
```
krizhanovsky: Just noticed that tfw_connection_send_srv() returns int, but nobody (in http.c) is interested in it. It seems we should treat somehow problems on sending a message. Probably, we should initiate connection failovering...

tempesta_fw/lb/tfw_lb_http.c
```diff
@@ -0,0 +1,1091 @@
...
+ * When all connections are established, the thread sleeps until some connection
+ * is lost. It is woken by a callback which is called when connection is closed.
+ * Then the thread iterates over all servers as usual and reestablishes the
+ * closed connection.
+ *
+ * If a connection attempt is failed, instead of sleeping forever, it wakes up
+ * periodically and makes another attempt (with unlimited number of retries).
+ */
+static int
+tfw_lb_connd_mainloop(void *data)
```
krizhanovsky: Threading is common and dummy way, need to think about more clever and performant way to handle the cases. Basically, TCP is event driven - everything done by it is done either by peer segments of local timers. So I reckon we can hook TCP events to handle reconnection in fully synchronous and non-blocking fashion at the same time. This is core idea behind Synchronous Sockets.

tempesta_fw/lb/tfw_lb_http.c
```diff
@@ -0,0 +1,1091 @@
...
+	return best_sk;
+}
+
+static struct sock *
+tfw_lb_sched(TfwLbGroup *group, TfwHttpReq *req)
+{
+	/* Should be done via a function pointer, but KISS for now. */
+	return (group->sched == TFW_LB_SCHED_RR)
+	       ? tfw_lb_sched_rr(group)
+	       : tfw_lb_sched_hash(group, req);
```
krizhanovsky: Good improvement - it's much better to have single powerful scheduler rather than 3 simple schedulers. However, it's still better to call the scheduling functions by callbacks, don't postpone it.

tempesta_fw/lb/tfw_lb_http.c
```diff
@@ -0,0 +1,1091 @@
...
+	return tfw_lb_sched_rr_failover(group);
+}
+
+/**
+ * XXX: the whole tfw_sched_hash is collapsed into this single function.
+ * Rendezvous hashing is included as a failovering strategy.
+ *
+ * XXX: the linear search is slow but it is required for failover switching.
+ * Perhaps we should do binary search first, and then fall back to linear if
+ * the chosen socket is dead.
```
krizhanovsky: Well, basically even with https://github.com/natsys/tempesta/issues/76 in mind we expect a lot of groups, but not so many servers for one group among which we do load balancing and failovering, so linear scan over contigouse cache line aligned array is Ok.

tempesta_fw/lb_mod.c
```diff
@@ -0,0 +1,65 @@
...
+
+int
+tfw_lb_mod_register(const TfwLbMod *mod)
+{
+	BUG_ON(!mod);
+	BUG_ON(!mod->name || !mod->send_msg);
+
+	TFW_DBG("registering load balancer: %s\n", mod->name);
+
+	if (tfw_lb_mod) {
```
krizhanovsky: It's good to protect from concurrent modules registration. The problem also exists in current tfw_sched_register(), but tfw_classifier_register() is properly synchronized.

tempesta_fw/lb/tfw_lb_http.c
```diff
@@ -0,0 +1,1091 @@
...
+
+static TfwLbGroup *
+tfw_lb_select_group(TfwHttpReq *req)
+{
+	TfwLbRule *rule;
+
+	if (!tfw_lb_rules)
+		return tfw_lb_default_group;
+
+	rule = tfw_http_match_req_entry(req, tfw_lb_rules, TfwLbRule, rule);
```
krizhanovsky: Typically, tfw_http_match_req() should be redesigned in context of #76. Snort uses the rules tree instead of list of full rules.

tempesta_fw/lb/tfw_lb_http.c
krizhanovsky: Not for this release. This depends on #83 and should be reworked when ss_connect() is implemented.

vdmit11:
The pull request is obsolete. The code is re-worked and now it is the iss78 branch.

Issue #47, opened 2015-02-10
The functionality to add HTTP headers to passing requests is broken


keshonok:
Currently this is used in ```tfw_http_add_forwarded_for()``` to add an ```X-Forwarded-For:``` header to HTTP requests that Tempesta receives, parses, and then passes to a backend server.

One reason is the use of constructs like ```vaddr = kmap_atomic(skb_frag_page(frag));```, where something like ```frag_addr = skb_frag_address(frag);``` should be used instead. That's not the only reason for broken behaviour though.

krizhanovsky:
```__hdr_add()``` also doesn't satisfy RFC7230 3.2.2 and generates duplicate headers.

keshonok:
Here's an idea for a possible interface that formed in my head while I was working on sticky cookie functionality.

The base for the functionality in this issue is the ability to add/delete/replace HTTP message data that reside in an SKB. Usually HTTP message data is in a `TfwStr{}` structure that may consist of multiple data chunks. SKB data may reside in multiple data chunks as well. That is very similar to `iovec{}` structure that is used to describe data consisting of multiple chunks.

Currently Tempesta can only put linear data to an SKB starting from some offset in that SKB. (see `tfw_http_msg_setup()` and `tfw_http_msg_add_data()`). That forces us to linearize `TfwStr{}` before using this interface. It's very similar in another case of `__hdr_add()` or `__hdr_sub()` that accept linear data only, and can work only with linear areas in SKB (can't cross data chunks).

If, say, describe SKB data chunks in `TfwStr{}` terms, it would be beneficial to have an interface that allowed to copy data from one `TfwStr{}` to another `TfwStr{}`.

For instance, when we need to insert a header to HTTP message that sits in an SKB, we would find or make space in that SKB, and return a `TfwStr{}` that describes that data area. Then just copy the header using the `TfwStr{}` to `TfwStr{}` copy interface. When we need to remove a header, we already have a `TfwStr{}` that holds it. Use it as an argument to the function that does the removal. Replacement of  a header is a combination of the two above, conceptually. Again, an existing `TfwStr{}` is an argument to the function that does it.

In cases like where we need to fill out a whole SKB with an HTTP message data, an iterator would be very useful to keep current position for data copying.

I presume that there might be cases where dealing with linear data is easier. Basically, linear data is just a `TfwStr{}` that consists of a single chunk (`PLAIN`), so it's trivial to provide that interface as well on top of the above interface.

PR #48, opened 2015-02-10
Process SKB data in full when parsing HTTP requests/responses

vdmit11:
1. sock.c:ss_skb_process_skb()
   Process SKB paged fragments in a correct way.
2. connection.c:tfw_connection_put_skb_to_msg()
   Tempesta links SKBs to messages, and thus takes an additional
   ownership of those SKBs. Indicate that to the Linux kernel by
   incrementing the number of SKB users. Later on when the messages
   are destroyed in tfw_http_msg_free(), the SKBs are released
   as well but not destroyed unless there are no more owners.
3. http.c
   Comment out the call to tfw_http_add_forwarded_for() due to
   broken functionality that leads to system crash (issue #47).
4. connection.c:tfw_connection_send_srv()
   Add an extra BUG_ON() to bar the use of a NULL pointer.
5. tempesta_fw/Makefile
   Relax the compilation parameters to allow the compilation pass
   okay with warnings. That became a necessity when a call to
   tfw_http_add_forwarded_for() was commented out.

### DIFF
--- a/tempesta_fw/Makefile
+++ b/tempesta_fw/Makefile
@@ -16,7 +16,7 @@
 # this program; if not, write to the Free Software Foundation, Inc., 59
 # Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 
-EXTRA_CFLAGS += -Werror -I$(src)/../sync_socket -I$(src)/../tempesta_db
+EXTRA_CFLAGS +=  -I$(src)/../sync_socket -I$(src)/../tempesta_db
 
 ifneq ($(NDEBUG), 1)
 EXTRA_CFLAGS += -DDEBUG -O0 -g3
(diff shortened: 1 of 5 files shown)


tempesta_fw/Makefile
```diff
@@ -16,7 +16,7 @@
 # this program; if not, write to the Free Software Foundation, Inc., 59
 # Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 
-EXTRA_CFLAGS += -Werror -I$(src)/../sync_socket -I$(src)/../tempesta_db
+EXTRA_CFLAGS +=  -I$(src)/../sync_socket -I$(src)/../tempesta_db
```
vdmit11: Perhaps it would be better to keep `-Werror` there?
At some point we agreed that all compiler warnings shall be fixed, so this option facilitates that.

vdmit11: Ok, I saw the commit message. Perhaps we should leave if for a while until #47 is fixed.

keshonok: I just want to comment out a single function call. However with ```-Werror``` the code won't compile as I get a compiler warning ```tempesta/tempesta_fw/http.c:690:1: warning: 'tfw_http_add_or_append_forwarded_for' defined but not used [-Wunused-function]```.

keshonok:
This pull request is outdated. Another will be created.

PR #57, opened 2015-02-20
configuration fixes

vdmit11:


krizhanovsky:
The patch set looks good. Please merge to master and close #71.

PR #58, opened 2015-02-20
tfw_sched_hash: implement Highest Random Weight (Rendezvous) hashing

vdmit11:
The HRW hashing method allows to stick HTTP messages to their home back-end
servers, and preserve this mapping even if the number of servers is changed.

In order to achieve that, the tfw_sched_hash is updated. Now it hashes not only
HTTP requests, but also servers (TfwServer objects). For each incoming HTTP
request it searches for a best match among all server hashes.
Each Host/URI hash has only one best matching TfwSrver hash which is chosen,
and thus any request always goes to its home server unless it is offline.

### DIFF
--- a/tempesta_fw/addr.c
+++ b/tempesta_fw/addr.c
@@ -515,3 +515,4 @@ tfw_addr_sa_len(const TfwAddr *addr)
         validate_addr(addr);
         return (addr->family == AF_INET6) ? sizeof(addr->v6) : sizeof(addr->v4);
 }
+EXPORT_SYMBOL(tfw_addr_sa_len);
--- a/tempesta_fw/sched/tfw_sched_hash.c
+++ b/tempesta_fw/sched/tfw_sched_hash.c
@@ -2,13 +2,22 @@
  *		Tempesta FW
  *
  * Hash-based HTTP request scheduler.
+ *
  * The scheduler computes hash of URI and Host header fields of a HTTP request
- * and uses the hash value as an index in the array of servers added with
- * the tfw_sched_add_srv() function. Therefore, requests with the same URI and
- * Host are mapped to the same server (unless the list of servers is changed).
+ * and uses the hash value as a key for searching a TfwServer object.
+ * Hence, HTTP requests with the same URI and Host always go to the same server
+ * unless it is offline.
+ *
+ * Also, the scheduler utilizes the Rendezvous hashing (Highest Random Weight)
+ * method that allows to stick every HTTP message to its server, preserving
+ * this mapping when other servers go down or new servers added.
+ * The scheduler hashes not only HTTP requests, but also servers (TfwServer
+ * objects), and for each incoming HTTP request it searches for a best match
+ * among all server hashes. Each Host/URI hash has only one best matching
+ * TfwSrver hash which is chosen, and thus any request always goes to its home
+ * server unless it is offline.
  *
  * TODO:
- *  - Replace the hash function (currnlty djb2 is used).
  *  - Refactoring: there is simpilar logic in all scheduler modules related
  *    to lists of TfwServer objects. The code should be extracted and re-used.
  *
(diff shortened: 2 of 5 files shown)


krizhanovsky:
Left few comments, looks good.

vdmit11:
Done some fixes. @krizhanovsky, please review it again.

tempesta_fw/addr.c
```diff
@@ -515,3 +515,4 @@ tfw_addr_sa_len(const TfwAddr *addr)
         validate_addr(addr);
         return (addr->family == AF_INET6) ? sizeof(addr->v6) : sizeof(addr->v4);
 }
+EXPORT_SYMBOL(tfw_addr_sa_len);
```
krizhanovsky: Commented many times: do not pollute space of exported symbols names. The function is pretty short and must be moved to .h file instead of being exported.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -29,84 +39,189 @@
...
 
+#include "http_msg.h"
 #include "log.h"
 #include "sched.h"
-#include "http_msg.h"
 
 MODULE_AUTHOR(TFW_AUTHOR);
 MODULE_DESCRIPTION("Tempesta hash-based scheduler");
-MODULE_VERSION("0.0.1");
+MODULE_VERSION("0.0.2");
```
krizhanovsky: I referenced a link to the versioning scheme in one of my previous reviews. Please fix the version.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -29,84 +39,189 @@
...
+ *  - A weak hash function adds unfairness to the load balancing.
+ *    There may be a case when a server pulls all load from all other servers.
+ *    And such condition (if it occurs although it is quite improbable) is
+ *    quite stable: it cannot be fixed by adding/removing servers and restarting
+ *    the Tempesta FW instance.
+ *  - For every HTTP request, we have to scan the list of all servers to find
+ *    a matching one with the highest weight. That adds some overhead.
+ *    Currently the linear search is used. We may switch to binary search, but
+ *    benchmarks are required to prove that it will behave better on the small
+ *    number of servers that we usually have.
```
krizhanovsky: I have the benchmarks: https://github.com/krizhanovsky/NatSys-Lab/blob/master/array_scans.cc

The program is only 136 lines in length, so it requires almost the same time to check the performance case as to write the comment ;) Don't be lazy.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -29,84 +39,189 @@
...
 	int ret = 0;
+	TfwSrvHash *new_list_entry;
+	TfwSrvHashList *new_list, *old_list;
+	size_t new_list_size, old_list_size;
+
+	spin_lock_bh(&srv_hashes_update_lock);
+
+	old_list = tfw_srv_hash_list;
+	old_list_size = TFW_SRV_HASH_LIST_SIZE(old_list->n);
+	new_list_size = TFW_SRV_HASH_LIST_SIZE(old_list->n + 1);
```
krizhanovsky: Usually growing in exponentionaly is more preferred than realloc and copy the array for each server. Just keep in ming #76.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -114,42 +229,73 @@ static int
...
-		servers[servers_n] = NULL;
+	TfwSrvHash *curr;
+	TfwSrvHashList *new_list, *old_list;
+	size_t new_list_size, old_list_size;
+
+	spin_lock_bh(&srv_hashes_update_lock);
+
+	old_list = tfw_srv_hash_list;
+	old_list_size = TFW_SRV_HASH_LIST_SIZE(old_list->n);
+	new_list_size = TFW_SRV_HASH_LIST_SIZE(old_list->n - 1);
```
krizhanovsky: Actually there is no need to shrink the list - ever 64K servers is just about 1MB of memory.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -114,42 +229,73 @@ static int
...
+	if (!new_list) {
+		ret = -ENOMEM;
+		goto out;
+	}
+
+	/* Copy entries to the new list excluding the one being deleted. */
+	for (curr = old_list->srv_hashes; curr->srv; ++curr) {
+		if (curr->srv != srv)
+			new_list->srv_hashes[new_list->n++] = *curr;
 	}
```
krizhanovsky: Why not to use memove()? It's much faster.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -29,84 +39,189 @@
...
+	(sizeof(TfwSrvHashList) + ((n) + _nulls) * sizeof(TfwSrvHash));	\
+})
 
-	return -1;
-}
+/**
+ * A NULL-terminated array of TfwSrvHash that stores all servers added to the
+ * scheduler via tfw_sched_hash_add_srv().
+ *
+ * Concurrent RCU updaters must be synchronized with servers_update_lock.
```
krizhanovsky: Comment seems outdated, there is no symbol with name "servers_update_lock"

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -114,42 +229,73 @@ static int
...
-		.add_srv = tfw_sched_hash_add_srv,
-		.del_srv = tfw_sched_hash_del_srv
-	};
+	int r;
 
 	LOG("init\n");
 
-	return tfw_sched_register(&tfw_sched_hash_mod);
+	/* Allocate a dummy list to avoid NULL checks in other functions. */
+	tfw_srv_hash_list = kzalloc(TFW_SRV_HASH_LIST_SIZE(0), GFP_KERNEL);
```
krizhanovsky: Why to allocate an empty array if you know that you'll insert some entries soon? Why not to allocate some adequate values, like 16?

vdmit11:
The pull request is obsolete. Now these changes are the part of the iss78 branch.

Issue #23, opened 2014-09-26
fix defects introduced by commit 1cd52cf


vdmit11:
1. Violation of encapsulation in `tfw_session_free()`.
The session-related code should not know about HTTP internals.
Need to add a generic destructor method to `TfwMsg`, set it up in `tfw_http_msg_create()` and call the destructor from `tfw_session_free()`.

2. tfw_http_msg_free() should be NULL-safe like kfree()

3. A TfwSession is bound with a client in tfw_connection_recv(), so at any point when tfw_sess_conn() is called the `cli` should not be NULL, so it is better to plce BUG() there instead of returning NULL when `cli` is NULL.

See comments for: 1cd52cf8f0052c00125e9f08d50518e1c49dc5e8

vdmit11:
Actually this issue is still not fully fixed.

This defect is still not resolved:
> Violation of encapsulation in tfw_session_free().

I failed to create any quick elegant fix for it. Everything I tried was ugly and bloated.

The essense of the problem (for me) is:
 1. The `TfwSession` is abstract. It should not contain any HTTP-related fields.
 2. But actually want to put some HTTP-related fields to a session. For example, a HTTP pipeline which is actually stored in the `TfwSession` and we are trying to fix it here.

Also, I think that:
 - The items `1.` and `2.` do contradict with each other.
 - We can't remove the item `1.` because it is a core part of the architecture.
 - We also can't remove the item `2.` becaus it is the essential complexity: pipelining works in this way and we have to implement it somehow.

I think the only way to solve this issue is to introduce a `TfwHttpSession`. The structure should complement the `TfwSession` and contain all the HTTP-session-related data.

Then, I see two possible implementations:

**1. composition:**
```
typedef struct {
     void *user_data;   /* Points to TfwHttpSession. */
} TfwSession;

typedef struct {
    struct list_head pipeline;
} TfwHttpSession;
```
In this case the `TfwHttpSession` is an independed object.

**2. interface/implementation**
```
typedef struct {
    TfwSession sess;
    struct list_head pipeline;
} TfwHttpSession;
```
In this case TfwSession is just an abstract "interface" (like a `file` or `inode`).
Various protocols implement their own sessions: `TfwHttpSession`, `TfwHttp2Session`, etc.

For me, the second approach looks better because:
 - `TfwSession` is abstract. It just declares how a specific object can behave. Hence, it should not have any default implementation. In terms of OOP, it is an interface or an abstract class, but not a specific implementation.
[...]

PR #59, opened 2015-02-20
switch from sk_destruct to TfwConnection->destruct

vdmit11:
Before the commit TfwConnection-related code was overwriting the
sk_destruct field of struct sock with a custom destructor function.
The original sk_destruct function was saved in TfwConnection and
had to be called manually from the custom destructor.

This change simply moves the destructor from sk_destruct to the
TfwConnection->destruct field. Now you don't need to call the
original socket destructor manually, and the field was occupied
anyway, so this commit makes code simpler for free.

### DIFF
--- a/tempesta_fw/client.c
+++ b/tempesta_fw/client.c
@@ -46,8 +46,6 @@ tfw_destroy_client(struct sock *s)
 	conn->hndl = NULL;
 
 	kmem_cache_free(cli_cache, cli);
-
-	conn->sk_destruct(s);
 }
 
 TfwClient *
--- a/tempesta_fw/connection.c
+++ b/tempesta_fw/connection.c
@@ -84,13 +84,11 @@ tfw_connection_free(TfwConnection *c)
  * A downcall for new connection called to set necessary callbacks
  * when a traditional Sockets connect() is calling.
  *
- * @destructor Is a function placed to sk->sk_destruct.
- * The original callback is saved to TfwConnection->sk_destruct and the passed
- * function must call it manually.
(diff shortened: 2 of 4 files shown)


keshonok:
My opinion is that this patch is premature and doesn't solve the core problem - that sockets are not destroyed properly, and an sk_destruct function is not called as it should.

The technique where the original destructor function is replaced with a new one, and the new function needs to call the original when it's run - that is a pretty standard technique called chaining. This technique is used throughout the linux kernel as well.

The original problem still stands. We need to solve the original problem.

keshonok:
After further research I believe that indeed it's better to do it this way. We don't even need to call it a `destructor`. The major point here is that we want to free Tempesta resources as soon as a socket leaves the `ESTABLISHED` state and starts the closing process. We don't really care how long the final TCP handshakes are gonna be for, or how long it would take the kernel to finally close a socket.

Socket's `sk_destruct` is a different matter that needs to be solved separately, and it doesn't affect the release of socket-related resources in Tempesta.

krizhanovsky:
The patch beacks socket-connection linkage. Previously if somebody calls sk_free(), then the connection destructor is called. Now this is not the case.

Linked with #23 somewhat.

vdmit11:
> Previously if somebody calls sk_free(), then the connection destructor is called. Now this is not the case.
Now the destructor is called from ss_do_close() which is called when the socket leaves the `TCP_ESTABLISHED` state, so the Tempesta FW frees its resources early and doesn't wait until the socket is actually deallocated. Isn't it the right way to go?

PR #64, opened 2015-02-27
fix check for kernel thread context in tdb_file_open()

vdmit11:
The check is done as following:
	BUG_ON(current->mm != &init_mm);

That triggers a BUG() every time you start Tempesta FW which cache
enabled. The condition is always false because current->mm is NULL
for kernel threads (see linux/Documentation/vm/active_mm.txt).


krizhanovsky:
Already fixed.

Issue #50, opened 2015-02-16
Slow HTTP prevention


krizhanovsky:
req_con_limit/frang must be extended by detection and prevention Slow HTTP attack.

keshonok:
Implemented in 1ca833c

PR #70, opened 2015-03-07
HTTP DoS protection (static limits in the req_conn_limit module)

vdmit11:


tempesta_fw/classifier/req_conn_limit.c
```diff
@@ -37,6 +37,7 @@
 #include "../classifier.h"
```
krizhanovsky: Please rename the file to frang.c. This module does more than just requests and connections limiting.

tempesta_fw/classifier/req_conn_limit.c
```diff
@@ -37,6 +37,7 @@
...
 #include "../connection.h"
 #include "../gfsm.h"
+#include "../http.h"
 #include "../log.h"
 
 MODULE_AUTHOR(TFW_AUTHOR);
```
krizhanovsky: Don't forget to update the version. It should be 0.2.0 now.

tempesta_fw/classifier/req_conn_limit.c
```diff
@@ -364,45 +608,80 @@ rcl_init(void)
...
+	 * The suggested solution is to extend the GFSM with the support of
+	 * "lightweight hooks" that behave like plain functions rather than FSM.
+	 * That should solve all these problems listed above:
+	 *  - Such hook may be unregistered any time it is not executed.
+	 *  - The hook doesn't need an ID, so no need to maintain a global list
+	 *    of all known hooks in the GFSM code.
+	 *  - No need to create a dummy FSM just to call the hook.
+	 *    This is easy to comprehend, and perhaps faster since the GFSM
+	 *    doesn't need to switch FSMs.
+	 */
```
krizhanovsky: The comment isn't relevant, this is just an improper GFSM usage. GFSM allows you to register a FSM, not just a hook. When you call tfw_gfsm_register_fsm() you register FSM handler which typicaly should run switch for current state of the tracked object. So you shouldn't have been registering the second FSM, but you should register second hook as you do.

tempesta_fw/http.h
```diff
@@ -34,9 +34,10 @@
 #define TFW_HTTP_PF_CRLF		(TFW_HTTP_PF_CR | TFW_HTTP_PF_LF)
 
 typedef enum {
-	TFW_HTTP_METH_GET	= 0,
-	TFW_HTTP_METH_HEAD	= 1,
-	TFW_HTTP_METH_POST	= 2,
+	TFW_HTTP_METH_GET,
+	TFW_HTTP_METH_HEAD,
+	TFW_HTTP_METH_POST,
+	_TFW_HTTP_METH_COUNT,
```
krizhanovsky: _TFW_HTTP_METH_COUNT is always the last one, so don't leave coma after it.

tempesta_fw/gfsm.h
```diff
@@ -70,7 +70,11 @@
  */
 enum {
 	TFW_FSM_HTTP,
-	TFW_FSM_RCL,	/* Request connection limiting classifier */ 
+
+	/* Request connection limiting classifier */
+	TFW_FSM_RCL_REQ,
+	TFW_FSM_RCL_CHUNK,
```
krizhanovsky: The change is wrong, see comments above.

tempesta_fw/classifier/req_conn_limit.c
```diff
@@ -264,86 +368,226 @@ static TfwClassifier rcl_class_ops = {
...
+		}
+		val = val * 10 + *p - '0';
+	}
+
+	*out_val = val;
+	return 0;
+}
+
+/* TODO: refactoring: get rid of these sysctl handlers,
+ * replace them with Tempesta's cfg framework. */
```
krizhanovsky: Why didn't you do it now?

tempesta_fw/classifier/req_conn_limit.c
```diff
@@ -251,11 +257,109 @@ rcl_req_limit(RclAccount *ra, struct sock *sk)
...
+
+	if (!rcl_http_field_len)
+		return TFW_PASS;
+
+	TFW_HTTP_FOR_EACH_HDR_FIELD(field, end, req) {
+		if (tfw_str_len(field) > rcl_http_field_len) {
+			TFW_DBG("rcl: http_field_len limit is reached\n");
+			return TFW_BLOCK;
+		}
+	}
```
krizhanovsky: So Cookie is also limited by this routine while Cookie is very specific header. Usually, we can limit HTTP header by 256 characters or something like that, while common Cookie is much larger. Cookies must be limited separately.

tempesta_fw/classifier/req_conn_limit.c
```diff
@@ -251,11 +257,109 @@ rcl_req_limit(RclAccount *ra, struct sock *sk)
...
+
+	r = rcl_http_uri_len_limit(req);
+	if (r)
+		return r;
+	r = rcl_http_field_len_limit(req);
+	if (r)
+		return r;
+	r = rcl_http_body_len_limit(req);
+	if (r)
+		return r;
```
krizhanovsky: This is not a FSM, rather it's just a sequence of checks. What if we extend the module up to tens of checks? Rather you should track current request state - wheter URL is parsed, whether Cookie is parsed, whether body is parsed and so on. Why should we call all the checks if only part of URI is parsed?

tempesta_fw/classifier/req_conn_limit.c
```diff
@@ -251,11 +257,109 @@ rcl_req_limit(RclAccount *ra, struct sock *sk)
...
+rcl_http_chunk_handler(void *obj, unsigned char *data, size_t len)
+{
+	int r;
+	TfwConnection *c = (TfwConnection *)obj;
+	TfwHttpReq *req = container_of(c->msg, TfwHttpReq, msg);
+
+	r = rcl_http_methods_check(req);
+	if (r)
+		return r;
+	r = rcl_http_len_limit(req);
```
krizhanovsky: Why do you call the same checks set for fully parsed request and only small chunk? This is wrong and inefficient.

tempesta_fw/classifier/req_conn_limit.c
```diff
@@ -251,11 +257,109 @@ rcl_req_limit(RclAccount *ra, struct sock *sk)
 }
 
 static int
+rcl_http_uri_len_limit(const TfwHttpReq *req)
+{
+	/* FIXME: tfw_str_len() iterates over chunks to calculate the length.
+	 * This is too slow. The value must be stored in a TfwStr field. */
```
krizhanovsky: This is because you improperly designed FSM. Actually you did set of callbacks instead of FSM. See https://github.com/natsys/tempesta/pull/70/files#r26332092 .

tempesta_fw/classifier/req_conn_limit.c
```diff
@@ -251,11 +257,109 @@ rcl_req_limit(RclAccount *ra, struct sock *sk)
...
+		}
+	}
+
+	return TFW_PASS;
+}
+
+static int
+rcl_http_body_len_limit(const TfwHttpReq *req)
+{
+	if (rcl_http_body_len && tfw_str_len(&req->body) > rcl_http_body_len) {
```
krizhanovsky: The check should work only for POST requests.

tempesta_fw/classifier/req_conn_limit.c
```diff
@@ -251,11 +257,109 @@ rcl_req_limit(RclAccount *ra, struct sock *sk)
...
+
+static int
+rcl_http_field_len_limit(const TfwHttpReq *req)
+{
+	const TfwStr *field, *end;
+
+	if (!rcl_http_field_len)
+		return TFW_PASS;
+
+	TFW_HTTP_FOR_EACH_HDR_FIELD(field, end, req) {
```
krizhanovsky: Poorly designed. Say we get 10 chunks and each chunk keeps a header. So you check all the headers on each next header and introduce O(n^2) complexity instead of O(n).

tempesta_fw/classifier/req_conn_limit.c
```diff
@@ -251,11 +257,109 @@ rcl_req_limit(RclAccount *ra, struct sock *sk)
...
+		return r;
+	r = rcl_http_len_limit(req);
+	if (r)
+		return r;
+
+	return 0;
+}
+
+static int
+rcl_http_chunk_handler(void *obj, unsigned char *data, size_t len)
```
krizhanovsky: Not all the checks are implemented. Following are required:

     - Content-Type must be present in POST request
      - List of allowed Content-Type values
      - Host header must be present
      - Host header must not be IP address
      - HTTP headers must not be doubled (set flag)
      - client_body and client_header timeouts

And separate check of Cookie size

etc/tfw_frang.conf
```diff
@@ -0,0 +1,45 @@
...
+#	new_connection_rate NUM;
+#	new_connection_burst NUM;
+#	concurrent_connections NUM;
+#	http_uri_len NUM;
+#	http_field_len NUM;
+#	http_body_len NUM;
+#	http_host_is_required true|false;
+#	http_methods [METHOD]...;
+#	http_ct_is_required true|false;
+#	http_ct_vals ["CONTENT_TYPE"]...;
```
krizhanovsky: Please describe each configuration option.

etc/tfw_frang.conf
```diff
@@ -0,0 +1,45 @@
...
+#
+# Syntax:
+#   frang_limits {
+#	request_rate NUM;
+#	request_burst NUM;
+#	new_connection_rate NUM;
+#	new_connection_burst NUM;
+#	concurrent_connections NUM;
+#	http_uri_len NUM;
+#	http_field_len NUM;
```
krizhanovsky: "field" is unclear, should he ```http_header_len``` instead.

tempesta_fw/classifier/frang.c
```diff
@@ -0,0 +1,695 @@
...
+#include "../tempesta_fw.h"
+#include "../classifier.h"
+#include "../connection.h"
+#include "../gfsm.h"
+#include "../http.h"
+#include "../log.h"
+
+MODULE_AUTHOR(TFW_AUTHOR);
+MODULE_DESCRIPTION("Tempesta static limiting classifier");
+MODULE_VERSION("0.1.1");
```
krizhanovsky: Again , "Don't forget to update the version. It should be 0.2.0 now."

etc/tfw_frang.conf
keshonok: Actually, the correct term is `field`, or `header field`. `Header` refer to the whole header that consist of multiple header fields.

tempesta_fw/classifier/frang.c
```diff
@@ -0,0 +1,695 @@
...
+		if (ra->last_ts + GC_TO < jiffies / HZ)
+			hash_del(&ra->hentry);
+	}
+	if (!ra) {
+		spin_unlock(&hb->lock);
+
+		/*
+		 * Add new client account.
+		 * Other CPUs should not add the same account while we
+		 * allocating the account w/o lock.
```
krizhanovsky: Why other CPU can't add the same account?

tempesta_fw/classifier/frang.c
```diff
@@ -0,0 +1,695 @@
...
+
+static int
+frang_http_field_len_limit(const TfwHttpReq *req)
+{
+	const TfwStr *field, *end;
+
+	if (!frang_cfg.http_field_len)
+		return TFW_PASS;
+
+	TFW_HTTP_FOR_EACH_HDR_FIELD(field, end, req) {
```
krizhanovsky: The comment is still not fixed: https://github.com/natsys/tempesta/pull/70#discussion_r26331794

tempesta_fw/classifier/frang.c
```diff
@@ -0,0 +1,695 @@
...
+	}
+
+	return TFW_PASS;
+}
+
+static int
+frang_http_body_len_limit(const TfwHttpReq *req)
+{
+	if (frang_cfg.http_body_len &&
+	    tfw_str_len(&req->body) > frang_cfg.http_body_len) {
```
krizhanovsky: Comment https://github.com/natsys/tempesta/pull/70#discussion_r26333138 is still not fixed

tempesta_fw/classifier/frang.c
```diff
@@ -0,0 +1,695 @@
...
+err_fsm:
+	tfw_classifier_unregister();
+err_class:
+	tfw_cfg_mod_unregister(&frang_cfg_mod);
+err_cfg:
+	kmem_cache_destroy(frang_mem_cache);
+	return r;
+}
+
+module_init(frang_init);
```
krizhanovsky: Comment https://github.com/natsys/tempesta/pull/70#discussion_r26333673 is still not fully implemented.

tempesta_fw/classifier/frang.c
```diff
@@ -0,0 +1,695 @@
...
+
+	if (!frang_cfg.http_ct_is_required || !frang_cfg.http_ct_vals ||
+	    req->method != TFW_HTTP_METH_POST)
+		return TFW_PASS;
+
+	/* Find the Content-Type header.
+	 *
+	 * XXX: Should we make the header "special"?
+	 * It would speed up this function, but bloat the HTTP parser code,
+	 * and pollute the headers table.
```
krizhanovsky: Yes, definitely. Please do this.

tempesta_fw/classifier/frang.c
```diff
@@ -0,0 +1,695 @@
...
+	field = &req->h_tbl->tbl[TFW_HTTP_HDR_HOST].field;
+	if (!field->ptr) {
+		TFW_DBG("frang: the Host header is missing\n");
+		return TFW_BLOCK;
+	}
+
+	/* FIXME: here should be a check that the Host value is not an IP
+	 * address. Need a fast routine that supports compound TfwStr.
+	 * Perhaps should implement a tiny FSM or postpone the task until we
+	 * have a good regex library. */
```
krizhanovsky: Propose a quick solution. Probably we'll be Ok to live with it. I believe nothing special from TfwStr isn't required and you can do this with current implementation.

tempesta_fw/classifier/frang.c
```diff
@@ -0,0 +1,905 @@
...
+	return TFW_PASS;
+}
+
+static int
+frang_http_field_len_special(const TfwHttpReq *req)
+{
+	const TfwStr *field, *end;
+
+	FOR_EACH_HDR_FIELD_SPECIAL(field, end, req) {
+		if (tfw_str_len(field) > frang_cfg.http_field_len) {
```
krizhanovsky: This function and the above check a header length against the same configuration argument. However, the task was check header length for Cookie (not all "special" headers) and all other headers separately since Cookie is usually much bigger than any other headers.

So (1) there should be different configuration options and (2) we're interested in Cookie header and 'all other' (special and raw) headers.

tempesta_fw/classifier/frang.c
```diff
@@ -0,0 +1,905 @@
...
+	 * timeout state in the loop. But when we're past these states, we
+	 * don't want to run through them on each run again, and just loop
+	 * in FieldDup and FieldLen states. I guess that can be done with
+	 * some clever FSM programming, but this is plain simpler.
+	 */
+	if (frang_cfg.clnt_hdr_timeout
+	    && (skb != head_skb) && FSM_HDR_STATE(req->frang_st)) {
+		unsigned long start = req->tm_header;
+		unsigned long delta = frang_cfg.clnt_hdr_timeout;
+		if (time_is_after_jiffies(start + delta))
```
krizhanovsky: We should do TFW_WARN() on reaching limits for this module. Please correct this and check other places.
Currently we just log the messages in dmesg and surely this is bottleneck in DDoS. The more performant logger is in our TODO.

krizhanovsky:
Good to merge except the comments above.
Please fix them and merge the branch to muster.
#50 should be closed as well.

krizhanovsky:
Just replace the linear search by binary one and merge the branch to master.

PR #88, opened 2015-04-01
Tempesta sticky cookie implementation

krizhanovsky:


etc/tempesta_fw.conf
```diff
@@ -18,6 +18,7 @@
 #
 # Default:
 #   backend 127.0.0.1:8080;
+backend 192.168.10.1:8080;
```
krizhanovsky: This is you local settings, don't forget to remove this from final merge.

tempesta_fw/http.c
```diff
@@ -696,6 +697,7 @@ tfw_http_append_forwarded_for(TfwHttpMsg *m)
 static int
 tfw_http_add_or_append_forwarded_for(TfwHttpMsg *m)
 {
+	return 0;
```
krizhanovsky: Don't forget to remove

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+	} else if (ret == 0) {
+		return tfw_http_sticky_notfound(hm);
+	} else if (ret == 1) {
+		return tfw_http_sticky_found(hm, &value);
+	} else {
+		TFW_WARN("Multiple Tempesta sticky cookies found: %d\n", ret);
+		return -1;
+	}
+	/*NOTREACHED*/
+	return -1;
```
krizhanovsky: These 6 lines above should be just

         TFW_WARN("Multiple Tempesta sticky cookies found: %d\n", ret);
         return -1;

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+#include <crypto/sha.h>
+#include "cfg.h"
+#include "client.h"
+#include "http.h"
+#include "http_msg.h"
+
+typedef struct tfw_str {
+	u_char	*ptr;
+	size_t	 len;
+} tfw_str_t;
```
krizhanovsky: There is no need for the data structure, please use TfwStr.

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+
+	for (i = 0; i < hm->h_tbl->size; i++) {
+		hdr = &hm->h_tbl->tbl[i].field;
+		if (tfw_str_eq_cstr(hdr, field, len,
+				    TFW_STR_EQ_PREFIX | TFW_STR_EQ_CASEI)) {
+			break;
+		}
+	}
+	if (i < hm->h_tbl->size) {
+		/* DIRTY HACK TO COMPENSATE FOR PARSER BUG */
```
krizhanovsky: What the bug exactly is? Please create appropriate issue for the bug.

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+		eptr--;
+	s->len = eptr - s->ptr + 1;
+
+	sptr = skip_spaces(s->ptr);
+	s->len -= sptr - s->ptr;
+	s->ptr = sptr;
+}
+
+static TfwStr *
+tfw_http_field(TfwHttpMsg *hm, const char *field, size_t len)
```
krizhanovsky: This is relatively generic HTTP function, so it's better to move it to http.c

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+	TfwStr *hdr;
+
+	if ((hdr = tfw_http_field(hm, field->ptr, field->len)) == NULL) {
+		return 0;
+	}
+	len = tfw_str_len(hdr) + 1;
+	if ((buf = tfw_pool_alloc(hm->pool, len)) == NULL) {
+		return -ENOMEM;
+	}
+	len = tfw_str_to_cstr(hdr, buf, len);
```
krizhanovsky: In general Tempesta aims zero-copy. There should be strong reason to copy data.

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+	return (ptr - buf);
+}
+
+/*
+ * Send an HTTP 302 response to the client. The response redirects
+ * the client to the same URI as the original request, but includes
+ * 'Set-Cookie:' header field that sets Tempesta sticky cookie.
+ */
+static int
+tfw_http_send_302(TfwHttpMsg *hm)
```
krizhanovsky: This should be more generic function which creates particular HTTP error response depending on passed arguments. Also should be moved to http.c

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+ * quick access to current SKB and page fragment. State is passed
+ * and updated on each call to these functions.
+ */
+typedef struct tfw_msg_add_state {
+	struct sk_buff *skb;
+	unsigned int	fragnum;
+} tfw_mastate_t;
+
+static void
+tfw_http_msg_add_data(tfw_mastate_t *state, TfwMsg *msg, char *data, size_t len)
```
krizhanovsky: Too generic HTTP routine, move it to http.c

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+	tfw_http_msg_add_data(&state, msg, buf, len);
+	tfw_http_msg_add_data(&state, msg, S_CRLF, SLEN(S_CRLF));
+
+	tfw_http_msg_add_data(&state, msg, S_F_CONTENT_LENGTH,
+					   SLEN(S_F_CONTENT_LENGTH));
+	tfw_http_msg_add_data(&state, msg, "0", 1);
+	tfw_http_msg_add_data(&state, msg, S_CRLF, SLEN(S_CRLF));
+
+	tfw_http_msg_add_data(&state, msg, S_F_LOCATION, SLEN(S_F_LOCATION));
+	tfw_http_msg_add_data(&state, msg, S_HTTP, SLEN(S_HTTP));
```
krizhanovsky: What's the sense to call the function so many times just with static strings as an argument? Why not to call it only once with concatenated static strings?

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+tfw_http_msg_setup(tfw_mastate_t *state, TfwMsg *msg, size_t len)
+{
+	struct page *page;
+	struct sk_buff *skb;
+	int i_frag, i_skb, nr_skb_frags;
+	int nr_frags = DIV_ROUND_UP(len, PAGE_SIZE);
+	int nr_skbs = DIV_ROUND_UP(nr_frags, MAX_SKB_FRAGS);
+
+	/*
+	 * TODO: Make sure to create SKBs with payload size <= MSS
```
krizhanovsky: Won't GSO handle the case?

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+
+/*
+ * Send an HTTP 302 response to the client. The response redirects
+ * the client to the same URI as the original request, but includes
+ * 'Set-Cookie:' header field that sets Tempesta sticky cookie.
+ */
+static int
+tfw_http_send_302(TfwHttpMsg *hm)
+{
+	u_char buf[256];
```
krizhanovsky: That's not good to allocate such buffers on stack. Also the buffer is used for copying, so it's better to avoid it completely.

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+	s->ptr = sptr;
+}
+
+static TfwStr *
+tfw_http_field(TfwHttpMsg *hm, const char *field, size_t len)
+{
+	int i;
+	TfwStr *hdr;
+
+	for (i = 0; i < hm->h_tbl->size; i++) {
```
krizhanovsky: Cookie should be made special header, so we can avoid the loop.

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+{
+	int ret;
+	u_char *valptr, *endptr;
+	const tfw_str_t s_field = tfw_string("Cookie:");
+	tfw_str_t value = { 0 };
+
+	/*
+	 * Find a 'Cookie:' header field in the request.
+	 * The search for Tempesta sticky cookie within the field.
+	 * Note that there can be only one "Cookie:" header field.
```
krizhanovsky: Why? HTTP request frequently has a serveral Cookie headers.

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+ * to the HTTP response' header block.
+ */
+static int
+tfw_http_sticky_add(TfwHttpMsg *hm, u_char *value, size_t len)
+{
+	int ret;
+	char buf[256], *ptr = buf;
+
+	/* Set up complete 'Set-Cookie:' header field */
+	memcpy(ptr, S_F_SET_COOKIE, SLEN(S_F_SET_COOKIE));
```
krizhanovsky: You don't care about data copies at all. At least for this case you could initialize buf with the static string to avoid at least this memcpy().

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+	ptr += SLEN(S_F_SET_COOKIE);
+	memcpy(ptr, tfw_cfg_sticky.name.ptr, tfw_cfg_sticky.name.len);
```
krizhanovsky: Why not just reserve enough data in the buffer and initialize it previously by S_F_SET_COOKIE?

tempesta_fw/http_sticky.c
```diff
@@ -0,0 +1,740 @@
...
+	return SLEN(S_V_DATE);
+}
+
+static size_t
+tfw_http_prep_sticky(u_char *buf, u_char *value, size_t len)
+{
+	u_char *ptr = buf;
+
+	while (len--) {
+		snprintf(ptr, 100, "%02x", (*value++ & 0x0FF));
```
krizhanovsky: snprintf() is harmful for performance, Dmitry did the faster implementation.

tempesta_fw/http_sticky.c
keshonok: Perhaps. After some digging in the Linux network stack code, I could see that whenever SKBs are filled  in the TCP stack, they are carefully made not to exceed MSS, including all headers. The same issue may arise when adding extra headers to an HTTP request, as that operation may cause the SKB payload to exceed MSS.

I guess this issue needs extra research, and it needs a clear written analysis of whats and whys.

tempesta_fw/http_sticky.c
keshonok: I have found a good answer with references here: http://stackoverflow.com/questions/16305814/are-multiple-cookie-headers-allowed-in-an-http-request Perhaps I should put the RFC and section numbers in the comment.

tempesta_fw/http_sticky.c
keshonok: I surely agree. The major reason for that was that `TfwStr{}` has no notion of an offset. There's no readily available way to get a substring of a '`TfwStr{}` string, because you can't point at the middle of a chunk in the scope of a `TfwStr{}`. Also, there's not current support for getting an offset of a substring match when doing a substring search, which might also be just because of the same issue.

There are two typical examples that I've encountered here. One is that I need to get the value of a header field: everything minus the actual `"<Header Field Name>: "`. The other is to find a substring, then get another substring that follows (as in <cookie name>=<cookie value>. Provided that we get the position of a matched substring (which we can't right now), that position can occur in the middle of a chunk.

Perhaps we could introduce a notion of an offset into `TfwStr{}`. That would take some extra space in the structure, but it would let us do things described above. Alternatively, we can split a chunk on the fly at an offset. That would require us to extend an array of chunks that make up an `TfwStr{}`, copy all chunks after the offset, and insert a new chunk in the middle that has data right up to the offset.

So, would adding an offset to `TfwStr{}` be a viable solution? Tha would require some work implementing and updating the rest of the codebase, but then we would be able to avoid copying nearly at all times, I believe.

tempesta_fw/http_sticky.c
keshonok: I am not sure I understand. Each particular HTTP error response may require different headers, as in this particular case it requires `Location: ` header. There could be a lot of variation between different responses. Could you please elaborate on how you see it done in a generic way?

tempesta_fw/http_sticky.c
keshonok: Yes, that would be beneficial. But I also use this function to search for `User-Agent: ` header field. Would `User-Agent: ` header field be a good candidate for being a special header as well? That would be great.

tempesta_fw/http_sticky.c
krizhanovsky: I have not particular ideas about exact implementation of generic function for forming error responses. However, it's clear that we'll need to generate different error codes, so we need some generic implementation. Sticky cookies is just at first case when we need error messaging....

tempesta_fw/http_sticky.c
krizhanovsky: > Would User-Agent: header field be a good candidate for being a special header as well?

Yes. This is very crucial header for us

tempesta_fw/http_sticky.c
krizhanovsky: > So, would adding an offset to TfwStr{} be a viable solution?

I don't think so. As I wrote don't care about copying for now.

tempesta_fw/http_sticky.c
krizhanovsky: Yes, good point. In fact I did see multiple Cookies during our DPI session. However, IIRC RFC7230-7235 also prohibits duplicate headers. Meantime, if client is RFC compliant and really doesn't duplicate headers, probably proxies passing the request add their own cookies (it seems what we do on our own - just put the new header).

So the right solution to check the newest RFC 723[0-5] for client and proxy headers and do the work according to the standard.

Issue #56, opened 2015-02-19
Kernel panic when freeing a Tempesta message


keshonok:
This appears to be a case of SKB's double-free - freeing an SKB that had been freed already.

```
Loading Tempesta kernel modules...
[    8.219962] [tempesta] Initializing Tempesta FW kernel module...
[    8.220504] [tempesta] init: cfg_if
[    8.220800] [tempesta] init: http
[    8.221066] [tempesta] init: server
[    8.221363] [tempesta] init: client
[    8.221650] [tempesta] init: session
[    8.221944] [tempesta] init: connection
[    8.222268] [tempesta]   register cfg: cache
[    8.222613] [tempesta]   register cfg: sock_backend
[    8.223006] [tempesta]   register cfg: sock_frontend
[    8.238490] [tempesta] Registering new scheduler: dummy
Starting Tempesta...
[    8.240987] [tempesta] got state via sysctl: start
[    8.241387] [tempesta]   reading configuration file...
[    8.241800] [tempesta]   reading file: /home/natsys-lab/tfw-run/tempesta/etc/tempesta_fw.conf
[    8.242864] [tempesta]   file size: 2224 bytes
[    8.243236] [tempesta]   read by offset: 0
[    8.243717] [tempesta]   read by offset: 2224
[    8.244194] [tempesta] starting all modules...
[    8.244559] [tempesta]   parsing configuration and pushing it to modules...
[    8.245134] [tempesta]   spec handle: 'listen'
[    8.245616] [tempesta]   created front-end socket: sk=ffff88003cc2e040
[    8.246197] [tempesta]   use default entry: 'cache off;'
[    8.246758] [tempesta]   spec handle: 'cache'
[    8.247231] [tempesta]   use default entry: 'cache_size 262144;'
[    8.247975] [tempesta]   spec handle: 'cache_size'
[    8.248682] [tempesta]   use default entry: 'cache_dir /opt/tempesta/cache;'
[    8.249596] [tempesta]   spec handle: 'cache_dir'
[    8.250104] [tempesta]   use default entry: 'backend 127.0.0.1:8080;'
[    8.250533] [tempesta]   spec handle: 'backend'
[    8.250833] [tempesta]   starting modules...
[    8.251111] [tempesta]   mod_start(): cache
[    8.251383] [tempesta]   mod_start(): sock_backend
[    8.251695] [tempesta]   Starting thread: tfw_bconnd
[    8.252101] [tempesta]   mod_start(): sock_frontend
[...]

PR #95, opened 2015-04-17
When freeing a message, correctly free SKBs that make the message.

keshonok:
This is a solution to #56. The key is knowing whether an SKB is in
Tempesta's full possession or not. It is from the moment it's been
taken out of a socket's receive queue, and until it's sent out, i.e.
put in a socket's send queue. If a message is freed during that time
frame, Tempesta is responsible of freeing SKBs that make the message.
Otherwise, the Linux networking stack will take care of that when it
finishes sending those SKBs out.

### DIFF
--- a/tempesta_fw/connection.c
+++ b/tempesta_fw/connection.c
@@ -130,22 +130,6 @@ tfw_connection_close(struct sock *sk)
 	return 0;
 }
 
-void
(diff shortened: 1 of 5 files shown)


keshonok:
This pull request is premature.

PR #104, opened 2015-05-18
Provide a descriptive warning message for each limit exceedance

keshonok:


tempesta_fw/http.h
```diff
@@ -165,9 +165,9 @@ typedef struct {
 	TfwStr			uri_path;
 	unsigned int		frang_st;
 	unsigned int		hdr_rawid;
+	unsigned int		body_len;
```
krizhanovsky: Why? How about downloading/uploading HDMI video file or something like that?

krizhanovsky:
There is only one comment. Other changes are good to be merged to master.

Issue #119, opened 2015-06-19
Crash on No server group scheduler


krizhanovsky:
Prerequisites:

  CentOS 7, Apache httpd or Nginx as backend server on the same node with Tempesta.

Scenario:

1. start apache httpd or nginx on 8080 port;

2. start Tempesta with default configuration;

3. send a HTTP request to Tempesta

The system crashes on the request processing:

    [tempesta]   parser: Req_Hdr(16:0): c=0xd(.), r=-2
    [tempesta]   parser: Req_HdrDone(93:0): c=0xa(.), r=-2
    [tempesta]   parse msg body: flags=0x2 content_length=0
    [tempesta]   request parsed: len=288 parsed=288 msg_len=288 res=0
    [tempesta]   GFSM return code 0
    [tempesta] ERROR: No server group scheduler
    [tempesta] ERROR: Can't get an appropriate server connection for a session
    [sync_sockets] Warning: can't process app data on socket ffff880045399180
    [tempesta]   close client socket: sk=ffff880045399180, conn=ffff88004651b5a0, cli=ffff880046439670
    [sync_sockets]   Close socket ffff880045399180 (account=1)
    [sync_sockets]   ss_do_close: sk ffff880045399180, sk->sk_socket           (null), state (Established)
    ------------[ cut here ]------------
    WARNING: at net/core/stream.c:201 sk_stream_kill_queues+0x12c/0x130()
    Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tfw_sched_dummy(O) tempesta_fw(O) tempesta_db(O) nf_conntrack_ipv4 nf_defrag_ipv4 xt_conntrack nf_conntrack ipt_REJECT xt_CHECKSUM iptable_mangle bridge stp llc ip6table_filter ip6_tables iptable_filter ip_tables ebtable_nat ebtables xfs exportfs ppdev dm_mirror parport_pc parport dm_region_hash serio_raw microcode i2c_piix4 i2c_core pcspkr dm_log dm_mod uinput btrfs xor zlib_deflate raid6_pq libcrc32c ata_generic pata_acpi e1000 ata_piix floppy ipv6 autofs4
    CPU: 1 PID: 335 Comm: systemd-journal Tainted: G           O 3.10.10+ #3
    Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.8.1-20150318_183358- 04/01/2014
[...]

Issue #111, opened 2015-06-15
Serve requests with responses from cache when there's no backend server


keshonok:
Currently, a request is blocked if there's no backend server to forward it to.

However Tempesta as a Web Server is considered a default behaviour now. If there's no backend server, requests should be served from cache. Requests should be blocked only when there's no cached response.

krizhanovsky:
Linked with #119. We should return 404 response if there is no cache entry as well as a backend where we can fetch the entry from.

keshonok:
This issue is implemented as described. Separate issues are opened for deeper problems that became evident after analysis of this issue. See the analysis in #119, and see new issues that refer to #119.

keshonok:
Merged to master: 92bf359

PR #121, opened 2015-06-22
Modify request processing logic (#119, #111)

keshonok:
Default operating mode of Tempesta is a Web Server now. If there's
no backend server available, requests should be served from cache.

With this change a response are now looked up in cache. If not found,
an attempt is made to forward the request to a backend server. When
that fails (no backend server available), an HTTP 404 response is sent
to the client. As most of this logic should run after a response is
looked up in cache, it's been moved to tfw_http_req_cache_cb() from
tfw_http_req_process().

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -156,6 +156,7 @@ tfw_http_msg_setup(TfwHttpMsg *hm, size_t len)
 #define S_HTTP			"http://"
 
 #define S_302			"HTTP/1.1 302 Found"
+#define S_404			"HTTP/1.1 404 Not Found"
 #define S_502			"HTTP/1.1 502 Bad Gateway"
 
 #define S_F_HOST		"Host: "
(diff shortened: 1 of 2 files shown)


tempesta_fw/http.c
```diff
@@ -1070,23 +1126,48 @@ tfw_http_adjust_resp(TfwHttpResp *resp)
...
 			return;
-		/* Send the request to appropriate server. */
+		}
+		if (tfw_http_adjust_req(req))
+			goto send_404;		/* Should it be an HTTP 502? */
+
+		/* Add request to the connection. */
+		list_add_tail(&req->msg.msg_list, &conn->msg_queue);
+
+		/* Send request to the server. */
```
krizhanovsky: A request is processed in some NUMA node while the corresponding cached response can live in different NUMA node. So to minimize memory transfers we delegate request processing to appropriate NUMA node by deferred work in tfw_cache_req_process(). This asynchronous processing has nothing with _new_ response processing if cache can't fulfill current request. If we don't find a cached entity, then we should decide which NUMA node should create and deliver a response and schedule a work for that node. Meantime, we should do as less as possible work with the request fields on the remove node.

So moving the logic from tfw_http_req_process() to tfw_http_req_cache_cb() is incorrect.

tempesta_fw/http.c
```diff
@@ -1070,23 +1126,48 @@ tfw_http_adjust_resp(TfwHttpResp *resp)
...
+		tfw_connection_send(req->conn, (TfwMsg *)resp);
 	} else {
-		if (tfw_http_adjust_req(req))
+		/* Dispatch request to an appropriate server. */
+		TfwConnection *conn = tfw_sched_get_srv_conn((TfwMsg *)req);
+		if (!conn)
+			goto send_404;
+		r = tfw_http_sticky_req_process((TfwHttpMsg *)req);
+		if (r < 0) {
+			goto send_404;		/* Should it be an HTTP 502? */
```
krizhanovsky: I'd say 500 error, since there were no upstream server response so far.

tempesta_fw/http.c
```diff
@@ -1070,23 +1126,48 @@ tfw_http_adjust_resp(TfwHttpResp *resp)
...
+		} else if (r > 0) {
+			/* Response sent, nothing to do */
 			return;
-		/* Send the request to appropriate server. */
+		}
+		if (tfw_http_adjust_req(req))
+			goto send_404;		/* Should it be an HTTP 502? */
```
krizhanovsky: I'd say 500 error, since there were no upstream server response so far.

tempesta_fw/http.c
```diff
@@ -1070,23 +1126,48 @@ tfw_http_adjust_resp(TfwHttpResp *resp)
...
 		 * TODO should we adjust it somehow?
 		 */
-		tfw_connection_send(conn, (TfwMsg *)resp);
+		tfw_connection_send(req->conn, (TfwMsg *)resp);
 	} else {
-		if (tfw_http_adjust_req(req))
+		/* Dispatch request to an appropriate server. */
+		TfwConnection *conn = tfw_sched_get_srv_conn((TfwMsg *)req);
+		if (!conn)
+			goto send_404;
```
krizhanovsky: This is not good and expected behavior (no cache and no upstream server), so please leave warning log message.

krizhanovsky:
Good to merge

Issue #113, opened 2015-06-15
Regulate reconnect timeouts


keshonok:
When an established connection is lost, or a connect attempt has failed, a reconnect is initiated and then repeated until a connection is established. Each reconnect attempt is made after a timeout that currently constant.

The timeout duration should be more flexible and possibly depend on circumstances. In a simpler case it should increase with each failed reconnect attempt, and it should be reset after a reconnect is successful. Reconnect attempts should be stopped altogether when timeout gets too big. That would mean, for instance, moving a backend server to offline state for server connections.

keshonok:
Exponential backoff algorithm appears to be a good and popular choice for this kind of reconnect delays.

keshonok:
Implemented in 00d792d

PR #128, opened 2015-06-24
Implement exponential backoff algorithm for reconnect timeouts (#113)

keshonok:
The reconnect timeout is now changed from MIN to MAX value with each
consecutive attempt according to a variant of exponential backoff
algorithm. The timeout is reset when a reconnect results in success.


krizhanovsky:
good to merge

Issue #127, opened 2015-06-24
Kernel crash at tempesta_fw/server.c:39 when Tempesta is stopped


keshonok:
Different circumstances. It could be an attempts to stop a running Tempesta with a stop command. Or it could be Tempesta stopping when it encountered an unrecoverable error while loading (on start command). The common part of the trace is below.

```
[    8.603159] [tempesta]   stopping already stared modules
[    8.603637] [tempesta]   mod_stop(): sock_srv
[    8.604061] [tempesta]   Free msg:           (null)
[    8.604510] [tempesta]   Free msg:           (null)
[    8.604936] [tempesta]   Free msg:           (null)
[    8.605373] [tempesta]   Free msg:           (null)
[    8.605808] [tempesta]   mod_stop(): http_sticky
[    8.606221] [tempesta]   mod_stop(): cache
[    8.824699] [tdb] Close table cache
[    8.825051] [tempesta]   mod_stop(): filter
[    8.841880] [tdb] Close table filter
[    8.842253] [tempesta]   spec cleanup: 'listen'
[    8.842673] [tempesta]   spec cleanup: 'server'
[    8.843115] ------------[ cut here ]------------
[    8.843537] kernel BUG at /home/natsys-lab/tfw-run/tempesta/tempesta_fw/server.c:39!
[    8.844018] invalid opcode: 0000 [#1] SMP 
[    8.844018] Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tfw_sched_dummy(O) tempesta_fw(O) tempesta_db(O)
[    8.844018] CPU: 0 PID: 2074 Comm: sysctl Tainted: G        W  O 3.10.10-tempesta #2
[    8.844018] task: ffff8800389b6100 ti: ffff880038c22000 task.ti: ffff880038c22000
[    8.844018] RIP: 0010:[<ffffffffa004cd22>]  [<ffffffffa004cd22>] tfw_destroy_server+0x25/0x3f [tempesta_fw]
[    8.844018] RSP: 0018:ffff880038c23c90  EFLAGS: 00010202
[    8.844018] RAX: 0000000000000001 RBX: ffff880038c23df0 RCX: a000000000000000
[    8.844018] RDX: ffffffffa005bff0 RSI: 1268000000000000 RDI: ffffffffa005bf88
[    8.844018] RBP: ffff880038c23c98 R08: ffffffff81acb070 R09: 0000000000000000
[...]

PR #129, opened 2015-06-24
Use correct list to get servers to destroy on Tempesta stop (#127)

keshonok:


krizhanovsky:
Good catch. Please merge to master

Issue #80, opened 2015-03-18
Synchronous Socket: unnecessary skb postponing


krizhanovsky:
Based on Alexey's notes.

ss_tcp_process_proto_skb() calls postpone_skb() callback if connection_recv() returns SS_POSTPONE. postpone_skb() points to tfw_connection_postpone_skb() which just adds the skb to conn->msg->skb_list. Meantime, ss_tcp_process_skb() firstly calls put_skb_to_msg() callback before executing ss_tcp_process_proto_skb() and the callback points to tfw_connection_put_skb_to_msg() which also adds the skb to the same list. So postponing the skb has no sense.

keshonok:
That is correct summary of my notes. Closely related is issue #31.

An incoming SKB is processed by `ss_tcp_process_skb()`. First thing it does is call `put_skb_to_msg()` callback that points to `tfw_connection_put_skb_to_msg()`. That function creates a new connection message if needed, and then adds the SKB to that message's list `conn->msg->skb_list` by executing `ss_skb_queue_tail(&conn->msg->skb_list, skb);`

Then `ss_tcp_process_proto_skb()` is called on every piece of data in the SKB. In turn, it calls the `connection_recv()` callback. On `TFW_POSTPONE` result from `connection_recv()` it calls `postpone_skb()` callback that points `tfw_connection_postpone_skb()`. The only purpose of `tfw_connection_postpone_skb()` is executing `ss_skb_queue_tail(&conn->msg->skb_list, skb);` - again, on the same SKB, the same message, and the same connection.

Granted, that doesn't do any harm as the SKB is not actually added to the same list twice due to safely measures.

All of that means the following:

* The whole `if (r == SS_POSTPONE)` case can safely be removed from `ss_tcp_process_proto_skb()`. That will eliminate unnecesary multiple calls to `tfw_connection_postpone_skb()`.
* `.postpone` hook in `ssocket_hooks{}` is not needed as well.
* The whole `tfw_connection_postpone_skb()` function can safely be removed as unused and unnecessary.

PR #132, opened 2015-06-26
Eliminate unnecessary SKB postponing (#80)

keshonok:
Due to specifics of SKB processing in Tempesta postponing has become
a dummy operation that does nothing and is completely unnecessary.

### DIFF
--- a/tempesta_fw/sock.c
+++ b/tempesta_fw/sock.c
@@ -198,13 +198,9 @@ ss_tcp_process_proto_skb(struct sock *sk, unsigned char *data, size_t len,
 
 	read_lock(&sk->sk_callback_lock);
 	r = SS_CALL(connection_recv, sk, data, len);
-	if (r == SS_POSTPONE) {
-		SS_CALL(postpone_skb, sk->sk_user_data, skb);
-		r = SS_OK;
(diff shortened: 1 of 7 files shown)


tempesta_fw/sock.c
```diff
@@ -198,13 +198,9 @@ ss_tcp_process_proto_skb(struct sock *sk, unsigned char *data, size_t len,
...
 	read_lock(&sk->sk_callback_lock);
 	r = SS_CALL(connection_recv, sk, data, len);
-	if (r == SS_POSTPONE) {
-		SS_CALL(postpone_skb, sk->sk_user_data, skb);
-		r = SS_OK;
-	}
 	read_unlock(&sk->sk_callback_lock);
 
-	return r;
+	return (r == SS_POSTPONE) ? SS_OK : r;
```
krizhanovsky: Dirty fix.

If we don't want to process TFW_POSTPONE return code, the all called functions (tfw_http_req_process() and tfw_http_resp_process()) should not return the value. However, the functions do handles POSTPONE logic, so they should leave the place where the logic is handled while the lower, network, layer should be unaware about POSTPONE.

Issue #46, opened 2015-02-09
sk->sk_destruct() is not being called when connection is closed


vdmit11:
Scenario:

1. start nginx
2. start Tempesta FW
3. `wget http://localhost` - should pass and give you `index.html`
3. stop nginx
4. `wget http://locslhost` - crashes

That happens because when the connection to nginx is closed, no `sk->sk_destruct()` is called.

We put `tfw_destroy_server()` to `sk->sk_destruct()`, so when the connection is closed, the server is not deleted from the load-balancing scheduler as it should be, and you get a NULL pointer dereference when you try to acess its connection.

Generally, the problem is wider than the fact that `tfw_destroy_server()` is not invoked.
Original socket destructor (usually `inet_release()`) is not invoked as well, so we likely have some memory leak here.

A notable thing that when you call `sock_release()`, the `sk->sk_destruct()` is invoked as it should be.
We need to understand why `sk->sk_destruct()` is not invoked when a TCP connection is closed, but is invoked when you do `sock_release()`.

keshonok:
With the way Synchronous Sockets are implemented now, we never reach `TCP_CLOSE` state. That is the root cause of this and some other problems.

Right now we do a forceful "non-sleeping" socket close at `TCP_CLOSE_WAIT` state by calling `ss_do_close()`. However that doesn't close the socket properly. This procedure breaks TCP protocol handshake on socket closing. It also breaks proper moves from one TCP state to another, and thus doesn't execute proper actions required at each state. That includes the final actions where the `sk_destruct()` function is called. By the end the socket must be clear of any write data, and that's the condition under which the `sk_destruct()` is called.

By the way, calling `ss_do_close()` at `TCP_CLOSE_WAIT` state loses the received FIN packet as it's not processed (it's discarded) but "assumed". If we don't call `ss_do_close()` at `TCP_CLOSE_WAIT` state, then we do catch the FIN packet in `ss_tcp_data_ready()` -> `ss_tcp_process_data()`. The latter then calls `ss_do_close()` as soon as it recognizes a FIN packet, and then it's all the same.

Related issues are: #52, and possibly #62, #63.

krizhanovsky:
An attempt to fix the issue was done in https://github.com/natsys/tempesta/pull/59

krizhanovsky:
It seems we ever have the problem in simple scenario on normal connection closing after successful data transfer in proxy mode:

        .....................
[...]

PR #133, opened 2015-06-29
Release a hold on a socket when closing it. (#46)

keshonok:
In some cases sock_put() was not called. Because of that, some
sockets were not closed properly: sk->sk_destruct() was not called.

### DIFF
--- a/tempesta_fw/sock.c
+++ b/tempesta_fw/sock.c
@@ -276,6 +276,8 @@ ss_tcp_process_skb(struct sk_buff *skb, struct sock *sk, unsigned int off,
 }
 
 /**
+ * This is main body of the socket close function in Sync Sockets.
+ *
  * inet_release() can sleep (as well as tcp_close()), so we make our own
  * non-sleepable socket closing.
  *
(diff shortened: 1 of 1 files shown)


tempesta_fw/sock.c
```diff
@@ -440,6 +440,13 @@ ss_do_close(struct sock *sk)
 	}
 }
 
+static void
+ss_do_close(struct sock *sk)
+{
+	__ss_do_close(sk);
+	sock_put(sk);
+}
+
```
krizhanovsky: Why can't we just call ss_close() which does sock_put()? Why do we need 3 functions for socket closing?

keshonok: `ss_close()` is called where is can be used in process context, so in includes additional locking. `ss_do_close()` is called in SoftIRQ context, and it does not need any extra locking around it, the socket is most probably locked already anyway. `__ss_do_close()` is the main body of the close function, and it is therefore called from both `ss_close()` and `ss_do_close()`.

`sock_put()` should be called outside of locks, that's why it's not in `__ss_do_close()` (or `ss_do_close()` before this patch).

krizhanovsky: Ok, makes sense. Just add appropriate comment to the functions.

krizhanovsky:
Tested. Good to merge after applying mentioned comments.

PR #141, opened 2015-07-09
Implement new SKB data processing design

keshonok:
A couple of major points:
1. An SKB is handed over to Tempesta from Sync Sockets, and Tempesta becomes
   responsible for it. That includes releasing an SKB when necessary. Because
   of that, processing of data in an SKB is moved to Tempesta. Sync Sockets
   expects now that each SKB is processed in full, and just needs to know if
   processing results in success or failure. Due to connection callback design
   TfwMsg{}.skb_offset is introduced to keep the current offset of data in the
   current SKB.
2. Data is parsed in separate chunks that an SKB holds. In case of pipelined
   HTTP requests processing may stop in the middle of data chunk. However the
   number of parsed bytes is not returned to the caller due to internal design.
   To deal with that, conn->msg is used in combination with TFW_PASS result code.
3. Now SKBs or requests/responses are never used after they were released.

### DIFF
--- a/tempesta_fw/connection.c
+++ b/tempesta_fw/connection.c
@@ -130,17 +130,27 @@ tfw_connection_send(TfwConnection *conn, TfwMsg *msg)
  * the use of sk->sk_user_data.
  */
 int
-tfw_connection_recv(struct sock *sk, unsigned char *data, size_t len)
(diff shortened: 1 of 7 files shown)


keshonok:
This design doesn't fit the architecture well. A different design is being implemented.

PR #142, opened 2015-07-09
Configuration: configure protocol for listening sockets

stefanstrogin:
From the configuration example
listen 192.168.0.1:80 proto=http;
Now attribute "proto" is handled with possible values "http" and "https".

Signed-off-by: Stefan Strogin <stefan.strogin@gmail.com>


krizhanovsky:
Good to merge after fixing Alexey's note.

stefanstrogin:
Fixed.

Issue #146, opened 2015-07-11
frang: bad request body length calculation


krizhanovsky:
Currently Frang tries to calculate current request body length based on current chunk of processed skb. That's wrong. ```TFW_HTTP_PARSE_BODY``` must be responsible for the calculation while appropriate state of Frang FSM should only analyze already fully preprares TfwHttpReq with correct ```body_len``` value.

keshonok:
It appears that the current length of an HTTP message's body is already calculated by the parser as `TfwMsg{}.body.len`. `Frang` needs to use it instead of calculating the current body length on its own. That makes code related to `TfwHttpMsg{}.body_len`, as well as the `body_len` member of the structure, redundant in `Frang` and `http.h`.

keshonok:
fixed with df3ff1d, abfbf9f.

PR #156, opened 2015-07-20
Use actual current HTTP message body length in Frang (#146)

keshonok:
Current HTTP message body length is accumulated by the parser, so there's
no need to do the same in Frang. Just use the value from the parser. Also,
Frang runs on an SKB now, not on a data chunk from an SKB. Current SKB is
passed as an argument. That makes the code a tad bit simpler.

### DIFF
--- a/tempesta_fw/http.h
+++ b/tempesta_fw/http.h
@@ -186,7 +186,6 @@ typedef struct {
 	unsigned int		hdr_rawid;
 	unsigned long		tm_header;
 	unsigned long		tm_bchunk;
-	unsigned long		body_len;
 	unsigned long		hash;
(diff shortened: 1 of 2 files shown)


tempesta_fw/http.h
```diff
@@ -186,7 +186,6 @@ typedef struct {
 	unsigned int		hdr_rawid;
 	unsigned long		tm_header;
 	unsigned long		tm_bchunk;
-	unsigned long		body_len;
```
krizhanovsky: Please remove corresponding comment line above as well.

krizhanovsky:
Good to merge with fix of the minor comment.

Issue #159, opened 2015-07-21
Incorrect processing of pipelined messages


keshonok:
Currently, when a pipelined request is encountered, the relevant SKB is cloned and attached to the new, pipelined message. As requests are sent out to a server, that leads to all sorts of issues. For instance, when an SKB contains exactly two requests, these requests are duplicated in the new SKB, and server receives four requests instead of two. It's even more troubling when pipelined requests are split between SKBs, and server receives half-formed requests.

Due to logic of inner working, this new message and SKB cloning operation needs to be done right after the current message is parsed and before it's processed further. That's because the current message may be deleted as result of further processing, and then the SKB containing the pipelined message will be lost.

On another note, currently schedulers don't know anything about pipelined messages and treat each message equally. However, it may be required to send a set of pipelined messages to the same server.

The main goals are:
1. No HTTP messages should be sent to a server in duplicate or malformed.
2. It must be possible to send pipelined messages to a single server or to different servers, based on settings and/or HTTP message contents.

While these goals sound simple, they may require certain architecture changes.

The relevant, or rather paired issue here is #160. Issues that are raised here are very similar to issues described there.

krizhanovsky:
skb data is typically a chain(s) of data blocks, so here we should do following:

1. do something like skb_clone(), which won't share the original skb data;

2. pass the buffers containing all messages after the first one (lets consider N pipelined messages) to the new clone;

3. split the buffer which contains end of the first message and header of the second to two buffers and assign the second buffer to the second clone;

4. adjust original skb data pointers, sizes etc;

5. do all the current processing over curent skb and continue parsing the new skb.

Probably it's not a big deal to do the steps above. Meantime, if we done with parsing of the first message, then the tail of the second could be on the way and we do not know when we receive the rest of second message, so we can't ask scheduler about whether it's going to send them to the same server or not. It's better to do all job right now to avoid congestions.

keshonok:
Fixed with 13360a9.

PR #167, opened 2015-07-30
Split SKBs with pipelined requests into separate SKBs for each request.

keshonok:
This commit addresses issues described in #159. SKBs that contain multiple
requests are split into separate SKBs with each SKB containing exactly one
request. In essence, request pipelining is eliminated, and all pipelined
requests are turned into a series of separate requests. The rationale for
that can be found in comments to #159.


krizhanovsky:
Good to merge.

PR #168, opened 2015-07-30
Rework ss_tcp_process_data() for proper FIN handling.

keshonok:
Data SKBs may have FIN flag set as well. Process that case.

### DIFF
--- a/tempesta_fw/sock.c
+++ b/tempesta_fw/sock.c
@@ -426,11 +426,10 @@ static void
 ss_tcp_process_data(struct sock *sk)
 {
 	bool tcp_fin, droplink = true;
-	int r, count, processed = 0;
+	int processed = 0;
 	unsigned int off;
 	struct sk_buff *skb, *tmp;
 	struct tcp_sock *tp = tcp_sk(sk);
-	void *conn;
 
 	skb_queue_walk_safe(&sk->sk_receive_queue, skb, tmp) {
 		if (unlikely(before(tp->copied_seq, TCP_SKB_CB(skb)->seq))) {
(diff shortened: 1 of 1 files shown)


tempesta_fw/sock.c
```diff
@@ -441,106 +442,111 @@ ss_tcp_process_data(struct sock *sk)
...
 		__skb_unlink(skb, &sk->sk_receive_queue);
 
+		tcp_fin = tcp_hdr(skb)->fin;
 		off = tp->copied_seq - TCP_SKB_CB(skb)->seq;
 		if (tcp_hdr(skb)->syn)
 			off--;
-		if (off < skb->len) {
-			int r, count = skb->len - off;
-			void *conn;
+		if (off < skb->len)
```
krizhanovsky: The case should be marked as likely()

tempesta_fw/sock.c
```diff
@@ -441,106 +442,111 @@ ss_tcp_process_data(struct sock *sk)
...
 		}
+		tp->copied_seq += count;
+		processed += count;
+
+		if (tcp_fin)
+			goto data_fin;
+		continue;
+lone_fin:
+		__kfree_skb(skb);
+data_fin:
```
krizhanovsky: The function control flow becomes to complex - too many gotos, labels and booleans. Sometimes it's better to use if statements and do little copy-paste (e.g. there is not a big deal to copy-paste ```SS_DBG(".."); ++tp->copied_seq;``` for FIN processing.

krizhanovsky:
good to merge

PR #172, opened 2015-08-04
Make a writable copy of request data in HTTP parser test

i-rinat:
Test for HTTP parser includes case normalization, which could result in writing to read-only segment.

The patch makes a copy of strings to ensure HTTP parser always provided with writable memory area.


krizhanovsky:
Good to merge.

Issue #165, opened 2015-07-28
Possible double call of __ss_do_close()


keshonok:
When calling `ss_close()` in Tempesta, eventually `_ss_do_close()` is called. In the process it comes to `TCP_FIN_WAIT1` socket state that is handled by `ss_tcp_state_change()` socket callback. In the callback `ss_droplink()` is called for `TCP_FIN_WAIT1` state, where `__ss_do_close()` is called as well. This leads to believe that a double call of `__ss_do_close()` may occur which is undesirable and wrong.

In practice, however, that doesn't seem to happen. Find out why.

keshonok:
`TCP_FIN_WAIT1` state is reached only in the process of a socket close. In that process `tcp_close_state()` is called, which moves a socket from `TCP_ESTABLISHED` state to `TCP_FIN_WAIT1` state, along with sending FIN to the other end of connection. `ss_state_change()` socket callback (`sk->sk_data_ready()`) is never called in this process, so that's why there's no double call of `__ss_do_close()` in Sync Sockets.

The case for `TCP_FIN_WAIT1` state should be removed from `ss_state_change()` as unnecessary.

`sk->sk_state_change()` **is** called for `TCP_FIN_WAIT2` state though - the next state after `TCP_FIN_WAIT1`.

krizhanovsky:
It seems we just should leave TCP_CLOSE_WAIT in the condition to be able to disassociate all resources from the socket on passive close, while we should not hook socket state change on active close. This is a bug anyway and should be fixed.

keshonok:
Fixed in 56f0852

PR #174, opened 2015-08-05
Remove the case TCP_FIN_WAIT1 state from ss_state_change() (#165)

keshonok:
Socket's state is moved from TCP_ESTABLISHED to TCP_FIN_WAIT_1 when a socket
is actively closed. sk->sk_state_change() is never called when this happens.
So, TCP_FIN_WAIT1 state is completely unnecessary in ss_state_change().


krizhanovsky:
Good to merge

PR #175, opened 2015-08-07
Generate sticky cookie when User-Agent header field is missing.

keshonok:
User-Agent header field is not mandatory and may be missing from an HTTP request.
Tempesta cookie is still generated for the request, just User-Agent header value
is not meshed in.


krizhanovsky:
Good to merge

PR #177, opened 2015-08-08
fill all output string fields in tfw_http_field_value()

i-rinat:
`tfw_http_field_value()` creates a new dynamically allocated string. So one can expect it to fill all TfwStr fields, including `.flags`.

### DIFF
--- a/tempesta_fw/http_sticky.c
+++ b/tempesta_fw/http_sticky.c
@@ -140,6 +140,7 @@ tfw_http_field_value(TfwHttpMsg *hm, const TfwStr *field_name, TfwStr *value)
 	ptr = strim(buf + field_name->len);
 	value->ptr = ptr;
 	value->len = len - (ptr - buf);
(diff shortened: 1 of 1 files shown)


krizhanovsky:
If ```value``` argument is fully written in ```tfw_http_field_value()```, then there is no reason to write zeros at the automatic variable initialization passed to the function. So I'd suggest to remove zeroing of the variable.

keshonok:
The caller expects that the contents of `value` is changed only where there's a successful result. In a recent case a missing `User-Agent` header field was considered a meaningful result as well (just not a complete success), where `value.len` remained a zero as the header field was not found.

So, if initialization of automatic variable `value` is removed, then all fields of the `value` variable will need to be explicitly set not just on successful result, but on failure as well.

krizhanovsky:
Oh, then it seems the cleanup is invalid and we should just leave original version of the code.

PR #179, opened 2015-08-10
Add round-robin scheduler tests

avbelov23:


tempesta_fw/t/unit/sched_helper.c
```diff
@@ -0,0 +1,70 @@
+#include "sched_helper.h"
```
krizhanovsky: Please add headers to all new files.

krizhanovsky:
Good tests.
Good to merge with the small cleanup.
Please do similar tests for hash and http schedulers.

PR #180, opened 2015-08-11
add trailing CRLF in Set-Cookie header

i-rinat:
Set-Cookie header miss trailing CRLF which could make HTTP response message invalid


krizhanovsky:
Good to merge

PR #186, opened 2015-08-17
code cleanups

i-rinat:


tempesta_fw/classifier/frang.c
```diff
@@ -401,8 +401,15 @@ frang_http_ct_check(const TfwHttpReq *req)
...
-			 curr->str);
+		char *buf;
+		size_t len = field->len + 1;
+
+		buf = tfw_pool_alloc(req->pool, len);
+		if (buf == NULL)
+			return TFW_BLOCK;
+		tfw_str_to_cstr(field, buf, len);
+
+		TFW_WARN("frang: restricted Content-Type value: %s\n", buf);
```
krizhanovsky: ```tfw_str_to_cstr()``` is deprecated due to increased memory footprint: TfwStr is designed to work with chunked strings and the routine breaks the concept. Since Tempesta is security application, we can expect very large, specially crafted, header values. Thus, ```tfw_pool_alloc()``` must be able to return very large memory areas, surely with some cost on performance. Does it have sense just to print warning? In this case it's better just to print first chunk of the string. In most normal cases headers are single chunked.

Meantime, it seems there is a bug in original TFW_WARN(): TfwStr does not guarantee zero-terminated strings, so the printing must be done with limited length.

i-rinat:
Removed calling of `tfw_str_to_cstr()`.

tempesta_fw/classifier/frang.c
```diff
@@ -401,8 +401,11 @@ frang_http_ct_check(const TfwHttpReq *req)
 		}
 	}
 	if (!curr->str) {
-		TFW_WARN("frang: restricted Content-Type value: %s\n",
-			 curr->str);
+		TfwStr *s;
+		/* take first chunk */
+		s = TFW_STR_PLAIN(field) ? field : TFW_STR_CHUNK(field, 0);
```
krizhanovsky: ```TFW_STR_CHUNK(s)``` returns just ```s``` if it's plain, so ```? : ``` operator is unnecessary here.

krizhanovsky:
Good to merge with minor fix.

PR #190, opened 2015-08-19
Add configuration mod find function

avbelov23:


tempesta_fw/cfg.c
```diff
@@ -1734,9 +1735,13 @@ tfw_cfg_mod_register(TfwCfgMod *mod)
 		return -EPERM;
 	}
 
+	write_lock_bh(&cfg_mods_lock);
```
krizhanovsky: Do we call the functions concurrently with softirqs? Otherwise there is no reason to use _bh version of the lock.

krizhanovsky:
Looks good

PR #197, opened 2015-08-21
Add kallsyms helper and sample

avbelov23:


tempesta_fw/t/unit/kallsyms_helper.c
```diff
@@ -0,0 +1,50 @@
+/**
+ *              Tempesta FW
+ *
+ * Copyright (C) 2012-2014 NatSys Lab. (info@natsys-lab.com).
```
krizhanovsky: This file is new, so only line at the below for 2015 is needed.

krizhanovsky:
Good to merge

PR #201, opened 2015-08-22
http_parser: change level only for duplicate strings

i-rinat:
`hdr_chunk_fixup()` should go down a level only if string is a duplicate.


krizhanovsky:
good to merge

Issue #204, opened 2015-08-23
Unit tests building problem


krizhanovsky:
There is linkage problem on building the tests:

  CC [M]  /root/tempesta/tempesta_fw/t/unit/main.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/test.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/helpers.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/test_addr.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/test_cfg.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/test_hash.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/test_http_match.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/test_tfw_str.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/test_http_parser.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/sched_helper.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/test_sched_rr.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/test_sched_hash.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/test_http_sticky.o
  CC [M]  /root/tempesta/tempesta_fw/t/unit/kallsyms_helper.o
  LD [M]  /root/tempesta/tempesta_fw/t/unit/tfw_test.o
WARNING: /root/tempesta/tempesta_fw/t/unit/tfw_test.o(.text+0x443d9): Section mismatch in reference from the function test_suite__http_sticky() to the function .init.text:tfw_http_sticky_init()
The function test_suite__http_sticky() references
the function __init tfw_http_sticky_init().
This is often because test_suite__http_sticky lacks a __init 
annotation or the annotation of tfw_http_sticky_init is wrong.

  Building modules, stage 2.
  MODPOST 9 modules
WARNING: /root/tempesta/tempesta_fw/t/unit/tfw_test.o(.text+0x443d9): Section mismatch in reference from the function test_suite__http_sticky() to the function .init.text:tfw_http_sticky_init()
The function test_suite__http_sticky() references
the function __init tfw_http_sticky_init().
This is often because test_suite__http_sticky lacks a __init 
annotation or the annotation of tfw_http_sticky_init is wrong.

krizhanovsky:
[...]

PR #209, opened 2015-08-24
#204. Fix building problem

avbelov23:


tempesta_fw/t/unit/test_http_sticky.c
```diff
@@ -34,6 +34,8 @@
 #define __init
 #endif
 
+#include "http_msg.c"
+
 #include "http_sticky.c"
 
 #include "http.c"
```
krizhanovsky: Do we still need to include http.c?

avbelov23: Yes.
WARNING: "tfw_http_send_502" undefined!
WARNING: "tfw_http_prep_302" undefined!
WARNING: "tfw_http_prep_hexstring" undefined!

krizhanovsky:
Yeah, I see. Good to merge.

PR #213, opened 2015-08-24
frang: check length of currently parsed header too

i-rinat:
Peeking into currently parsed header line and check its length
to block malicious request as fast as possible.


krizhanovsky:
Will be part of ak-158

Issue #210, opened 2015-08-24
404 Response is not received by a client.


keshonok:
This occurs with current master at dd9ec9f.
Cache is off in the configuration file, no back end server is configured. Client runs on host machine, Tempesta runs in VM. Simple request like `wget -d -S <tempesta_IP_address>`.

It appears that the response is sent out, but it's not received by a client.

```
lta@bastinda:~/tfw-run/wget$ wget -d -S 192.168.10.230
Setting --server-response (serverresponse) to 1
DEBUG output created by Wget 1.15 on linux-gnu.

URI encoding = ‘UTF-8’
--2015-08-24 13:38:22--  http://192.168.10.230/
Connecting to 192.168.10.230:80... connected.
Created socket 3.
Releasing 0x0000000000967b30 (new refcount 0).
Deleting unused 0x0000000000967b30.

---request begin---
GET / HTTP/1.1
User-Agent: Wget/1.15 (linux-gnu)
Accept: */*
Host: 192.168.10.230
Connection: Keep-Alive

---request end---
HTTP request sent, awaiting response...
```

```
Starting nginx: nginx.
Loading Tempesta kernel modules...
[   16.180559] [tdb] Start Tempesta DB
[   16.332776] [tempesta] Initializing Tempesta FW kernel module...
[   16.333215] [tempesta] init: cfg_if
[   16.333462] [tempesta] init: tls
[   16.333684] [tempesta] init: http
[   16.333912] [tempesta] init: http_sticky
[   16.335853] [tempesta] init: server
[   16.336143] [tempesta] init: client
[   16.336400] [tempesta] init: sock_srv
[   16.336662] [tempesta] init: sock_clnt
[   16.336931] [tempesta]   register cfg: filter
[   16.337252] [tempesta]   register cfg: cache
[   16.337548] [tempesta]   register cfg: http_sticky
[   16.337877] [tempesta]   register cfg: sock_srv
[   16.338193] [tempesta]   register cfg: sock_clnt
[   16.355056] [tempesta] Registering new scheduler: dummy
[   16.367538] [tempesta]   tfw_sched_hash: init
[   16.367878] [tempesta] Registering new scheduler: hash
[   16.381922] [tempesta]   tfw_sched_http: init
[   16.382260] [tempesta]   register cfg: tfw_sched_http
[   16.382613] [tempesta] Registering new scheduler: http
[   16.396381] [tempesta]   tfw_sched_rr: init
[...]

PR #218, opened 2015-08-25
Fix tfw_http_msg_write() (#210)

keshonok:


tempesta_fw/http_msg.c
```diff
@@ -559,40 +559,58 @@ tfw_http_msg_iter_init(TfwHttpMsg *hm, TfwMsgIter *it)
...
 			frag = ss_skb_frag_next(&hm->msg.skb_list,
 						&it->skb, &it->frag);
+			it->frag_off = 0;
+			/*
+			 * If all data from the chunk has been copied,
+			 * then switch to next chunk. Otherwise, stay
+			 * in the current chunk.
+			 */
+			/* Note that n_copy equals f_room here. */
+			if (c_size == n_copy) {
```
krizhanovsky: It'd be easier to understand if you replace the condition by ```c_size < f_room```

tempesta_fw/ss_skb.h
```diff
@@ -167,6 +167,23 @@ ss_skb_adjust_data_len(struct sk_buff *skb, int delta)
...
+ * In that case, return PAGE_SIZE for the fragment size.
+ * See implementation of ss_skb_alloc().
+ */
+static inline unsigned int
+ss_skb_frag_size(skb_frag_t *frag, unsigned int frag_size)
+{
+        return frag_size ? skb_frag_size(frag) : PAGE_SIZE;
+}
+
+
```
krizhanovsky: extra blank line

tempesta_fw/http_msg.c
```diff
@@ -549,50 +549,109 @@ tfw_http_msg_hdr_add(TfwHttpMsg *hm, TfwStr *hdr)
...
+ * to support consecutive writes. This function is intended to work
+ * together with tfw_http_msg_write() that uses the @it iterator.
+ */
+TfwHttpMsg *
+tfw_http_msg_create(TfwMsgIter *it, int type, size_t data_len)
+{
+	TfwHttpMsg *hm = tfw_http_msg_alloc(type);
+
+	if ((hm == NULL) || (data_len == 0))
+		return NULL;
```
krizhanovsky: It seems we should return either @hm if data_len == 0 (@hm is still valid new message, but with zero body) or free the message before return. I'd prefer the first one.

tempesta_fw/http_msg.c
```diff
@@ -549,50 +549,109 @@ tfw_http_msg_hdr_add(TfwHttpMsg *hm, TfwStr *hdr)
...
+/*
+ * Fill up an HTTP message @hm with data from string @data. An iterator
+ * @it is used to support multiple calls to this functions after set up.
+ * This function can only be called after a call to tfw_http_msg_setup().
+ * It works only with empty SKB space prepared by tfw_http_msg_setup().
+ * It should not be used under any other circumstances.
+ */
 int
-tfw_http_msg_write(TfwHttpMsg *hm, TfwMsgIter *it, const TfwStr *data)
+tfw_http_msg_write(TfwMsgIter *it, TfwHttpMsg *hm, const TfwStr *data)
```
krizhanovsky: Not a big deal, but usually return values are the last one.

tempesta_fw/http_msg.h
```diff
@@ -24,8 +24,9 @@
 #include "http.h"
 
 typedef struct {
-	int		frag;
-	int		frag_off;
+	unsigned int	frag;
+	unsigned int	frag_off;
+	unsigned int	frag_size;
```
krizhanovsky: It seem you forgot to remove frag_size.

tempesta_fw/http_msg.c
keshonok: On return from this function we expect that it has successfully created both the message and all SKBs with data. So it has to be all or nothing here.

Issue #212, opened 2015-08-24
Failures at HTTP parser unit test


krizhanovsky:
After #47 fixes we get HTTP parser errors in unit test:

        tfw_test: split: head_len=43 [GET /a/b/c/dir/?foo=1&bar=2#abcd HTTP/1.1
        ], tail_len=2 [
        ]
        [tempesta]   parse 43 client data bytes (GET /a/b/c/dir/?foo=1&bar=2#abcd HTTP/1.1
        ) on req=ffff88002c5f2040
        [tempesta]   enter FSM at state 0
        [tempesta]   parser: Req_0(0:0): c=0x47(G), r=-2
        [tempesta]   parser: Req_Method(1:0): c=0x47(G), r=-2
        [tempesta]   parser: Req_Uri(11:0): c=0x2f(/), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x61(a), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x2f(/), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x62(b), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x2f(/), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x63(c), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x2f(/), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x64(d), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x69(i), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x72(r), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x2f(/), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x3f(?), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x66(f), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x6f(o), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x6f(o), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x3d(=), r=-2
        [tempesta]   parser: Req_UriAbsPath(22:0): c=0x31(1), r=-2
[...]

PR #225, opened 2015-08-27
#212. Fix unit tests

avbelov23:


tempesta_fw/http_msg.c
```diff
@@ -46,7 +46,7 @@ tfw_http_msg_hdr_val(TfwStr *hdr, int id, TfwStr *val)
 
 	BUG_ON(TFW_STR_PLAIN(hdr));
 	BUG_ON(TFW_STR_DUP(hdr));
-	BUG_ON(nlen >= hdr->len);
+	BUG_ON(nlen > hdr->len);
```
krizhanovsky: When headers w/o spaces and any values are valid? If a header has length exactly as from hdr_lens, it means that its total length is just a header name plus ":". Moreover, this is used only for special header and it seems this is invalid case.

tempesta_fw/http_msg.c
```diff
@@ -54,10 +54,9 @@ tfw_http_msg_hdr_val(TfwStr *hdr, int id, TfwStr *val)
 
 	TFW_STR_FOR_EACH_CHUNK(c, hdr, {
 		BUG_ON(!c->len);
-		BUG_ON(c->len >= val->len);
```
krizhanovsky: Did you see crash on the assertion? Typically it should always be true. Just ">=" should be replaced by ">".

avbelov23: Yes. Example: TfwString(chank1: "Connection: ", chank2: "close").

Issue #205, opened 2015-08-23
Eliminate sleeping locks in Tempesta sockets API code


keshonok:
Tempesta implements its own version of sockets API functions. Some of them still use `lock_sock()` and `release_sock()` that cannot be used in SoftIRQ context as they may sleep/block. An obvious example is `ss_connect()`.

All other Tempesta sockets API functions need to be verified as well. There's a chance that some of them use sleeping/blocking locks somewhere inside the kernel. In that case those parts of the kernel code need to be brought in into `sock.c` and put inside non-blocking locks.

keshonok:
Also, there's a bug in `ss_connect()` here: https://github.com/natsys/tempesta/blob/master/tempesta_fw/sock.c#L1031 The socket is not unlocked on return.

krizhanovsky:
Yeah, just faced the bug:

        [sync_sockets]   ss_send: sk ffff880044ad4180, sk->sk_socket           (null), state (Established)
        [sync_sockets]   ss_send:208 entail skb=ffff88007a96d080 data_len=31 len=397
        [sync_sockets]   ss_send:226 sk=ffff880044ad4180 is_queue_empty=0 tcp_send_head(sk)=ffff88007a96d080 sk->sk_state=1
        [sync_sockets]   ss_tcp_state_change: sk ffff880044ad4180, sk->sk_socket           (null), state (Close Wait)
        [sync_sockets]   Peer connection closing
        [tempesta]   connection lost: 127.0.0.1:8080
        [tempesta]   Free msg:           (null)
        [tempesta]   Free msg: ffff880046033040
        [tempesta]   free skb ffff88007b17a200: truesize=2269 sk=          (null), destructor=          (null) users=1 type=Conn_Clnt
        [sync_sockets]   Close socket ffff880044ad4180 (account=1)
        [sync_sockets]   __ss_do_close: sk ffff880044ad4180, sk->sk_socket           (null), state (Close Wait)
        [sync_sockets]   free rcv skb ffff880044703108
        [sync_sockets]   ss_tcp_state_change: sk ffff880046033780, sk->sk_socket           (null), state (Established)
        [tempesta]   connected: 127.0.0.1:8080
        [sync_sockets]   ss_tcp_state_change: sk ffff880044ad4880, sk->sk_socket           (null), state (Close Wait)
        [sync_sockets]   Peer connection closing
        [tempesta]   connection lost: 127.0.0.1:8080
        [tempesta]   Free msg:           (null)
[...]

PR #227, opened 2015-08-31
#205. Eliminate sleeping locks in Tempesta sockets API code

avbelov23:


tempesta_fw/sock.c
```diff
@@ -1036,45 +1039,283 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
...
+	int err;
+
+	old_state = sk->sk_state;
+	if (!((1 << old_state) & (TCPF_CLOSE | TCPF_LISTEN)))
+		goto out;
+
+	/* Really, if the socket is already in listen state
+	 * we can only allow the backlog to be adjusted.
+	 */
+	if (old_state != TCP_LISTEN) {
```
keshonok: Do we not want to support TCP Fast Open?
If we really don't, then we should at least have a comment about that here.

tempesta_fw/sock.c
```diff
@@ -1036,45 +1039,283 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
...
-		.type = sk->sk_type,
-		.state = SS_UNCONNECTED
-	};
+	int err;
+
 	BUG_ON(sk->sk_type != SOCK_STREAM);
-	return inet_listen(&sock, backlog);
+
+	bh_lock_sock(sk);
+	err = ss_inet_listen(sk, backlog);
```
keshonok: `ss_inet_listen()` is never called in any context other than from `ss_listen()`. Therefore there's no reason to have it as a separate function

tempesta_fw/sock.c
```diff
@@ -1036,45 +1039,283 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
...
-	release_sock(sk);
+	bh_unlock_sock(sk);
 	return err;
 }
 EXPORT_SYMBOL(ss_connect);
 
 /*
- * The original functions are inet_bind() and inet6_bind().
- * These two can be made a bit shorter should that become necessary.
+ * Bind for IPv4 protocol. The original functions are inet_bind().
```
keshonok: The original function~~s~~ ~~are~~ is inet_bind().

tempesta_fw/sock.c
```diff
@@ -1036,45 +1039,283 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
...
+ *
+ * NOTE: This code assumes that both IPv4 and IPv6 are compiled in as
+ * part of the Linux kernel, and not as separate loadable kernel modules.
```
keshonok: I don't think this comment is relevant here (as well as below). It's relevant for `ss_inet_create()` as the original implementation tries to load the module if the code is not in the kernel yet, while the Sync Sockets implementation skips that.

tempesta_fw/sock.c
```diff
@@ -1036,45 +1039,283 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
...
+	inet->inet_daddr = 0;
+	inet->inet_dport = 0;
+	sk_dst_reset(sk);
+	err = 0;
+out:
+	return err;
+}
+
+/*
+ * Bind for IPv6 protocol. The original functions inet6_bind().
```
keshonok: The original function~~S~~ **is** inet6_bind().

tempesta_fw/sock.c
```diff
@@ -1036,45 +1039,283 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
...
+ *
+ * NOTE: This code assumes that both IPv4 and IPv6 are compiled in as
+ * part of the Linux kernel, and not as separate loadable kernel modules.
```
keshonok: Please the comment above on the same issue.

tempesta_fw/sock.c
```diff
@@ -1036,45 +1039,283 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
...
+	 *  is temporarily down)
+	 */
+	err = -EADDRNOTAVAIL;
+	if (!sysctl_ip_nonlocal_bind &&
+	    !(inet->freebind || inet->transparent) &&
+	    addr->sin_addr.s_addr != htonl(INADDR_ANY) &&
+	    chk_addr_ret != RTN_LOCAL &&
+	    chk_addr_ret != RTN_MULTICAST &&
+	    chk_addr_ret != RTN_BROADCAST)
+		goto out;
```
keshonok: As a matter of fact, we do set `inet_sk(sk)->freebind = 1;` in `sock_clnt.c`, so the condition above is never true. Do we need to keep this code block?

tempesta_fw/sock.c
```diff
@@ -1036,45 +1039,283 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
...
+	    !ns_capable(net->user_ns, CAP_NET_BIND_SERVICE))
+		goto out;
+
+	/*      We keep a pair of addresses. rcv_saddr is the one
+	 *      used by hash lookups, and saddr is used for transmit.
+	 *
+	 *      In the BSD API these are the same except where it
+	 *      would be illegal to use them (multicast/broadcast) in
+	 *      which case the sending device address is used.
+	 */
```
keshonok: I believe that there's no need to copy this kind of elaborate comments from original function here. We have a reference to the original function, and that should be sufficient. There are big blocks of comments in other places as well.

tempesta_fw/sock.c
```diff
@@ -1036,45 +1039,283 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
...
+	    addr->sin_addr.s_addr != htonl(INADDR_ANY) &&
+	    chk_addr_ret != RTN_LOCAL &&
+	    chk_addr_ret != RTN_MULTICAST &&
+	    chk_addr_ret != RTN_BROADCAST)
+		goto out;
+
+	snum = ntohs(addr->sin_port);
+	err = -EACCES;
+	if (snum && snum < PROT_SOCK &&
+	    !ns_capable(net->user_ns, CAP_NET_BIND_SERVICE))
```
keshonok: This check surely doesn't hurt, but do we need it?

tempesta_fw/sock.c
```diff
@@ -1036,45 +1039,283 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
...
+	 *      which case the sending device address is used.
+	 */
+
+	/* Check these errors (active socket, double bind). */
+	err = -EINVAL;
+	if (sk->sk_state != TCP_CLOSE || inet->inet_num)
+		goto out;
+
+	inet->inet_rcv_saddr = inet->inet_saddr = addr->sin_addr.s_addr;
+	if (chk_addr_ret == RTN_MULTICAST || chk_addr_ret == RTN_BROADCAST)
```
keshonok: Do we need this check?

tempesta_fw/sock.c
```diff
@@ -1036,45 +1039,283 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
...
+	if (snum)
+		sk->sk_userlocks |= SOCK_BINDPORT_LOCK;
+	inet->inet_sport = htons(inet->inet_num);
+	inet->inet_dport = 0;
+	inet->inet_daddr = 0;
+unlock:
+	rcu_read_unlock();
+out:
+	return err;
+}
```
keshonok: Basically, most questions here are similar to questions to `ss_inet_bind()` above.

tempesta_fw/sock.c
```diff
@@ -1036,45 +1039,283 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
...
+
+	if (addr_len < SIN6_LEN_RFC2133)
+		return -EINVAL;
+
+	addr_type = ipv6_addr_type(&addr->sin6_addr);
+	if ((addr_type & IPV6_ADDR_MULTICAST))
+		return -EINVAL;
+
+	snum = ntohs(addr->sin6_port);
+	if (snum && snum < PROT_SOCK && !ns_capable(net->user_ns, CAP_NET_BIND_SERVICE))
```
krizhanovsky: Too long line, please keep 80 characters as maximum line length.

Issue #184, opened 2015-08-12
Integrity of conn->msg_queue may be compromised


keshonok:
For caching purposes, HTTP requests that are sent out to a back end server are kept in a list until a matching response comes. The list `conn->msg_queue` is maintained for each connection to back end servers.

These requests may come on different CPU, yet they may be scheduled to go out on the same connection. That means these requests may be inserted into the list in parallel. As responses are received, matching requests are removed from the list, also in parallel with possible inserts.

Right now access to the list is not serialized, and so the integrity of the list can be compromised. All access to the list must be serialized.

keshonok:
Fixed via 7047279.

Issue #118, opened 2015-06-18
ss_send() may attempt to entail data SKBs to inactive or closed socket


keshonok:
When a request comes, Tempesta chooses a backend server to forward that request to.  By the time Tempesta actually starts putting request's SKBs to the socket connected to the backend server (destination socket), that socket may have been closed or be in the process of closing. `ss_send()` would then attempt to entail SKBs to a socket that is closed or is being closed.

There's no easy solution.
* Certain measures must be taken to not let the kernel completely delete the destination socket structure.
* The destination socket must be checked to ensure that it still belongs to Tempesta. For that it must be locked (`sk->sk_callback_lock`) around access to `sk->sk_user_data` (see issue #107). However, as `ss_send()` runs under the same lock on the source socket, we have the same case of classic deadlock as in issue #61.

Also, #114 is related to this issue.

krizhanovsky:
Backend socket transmission is done after the request scheduling, i.e. the scheduler 'decides' the socket as alive and backend as normally operational, while it's not actually. This is unexpected condition, so Tempesta must return 500 error to client.

krizhanovsky:
This is quite similar to #114. So consider this tasks as a task for race condition when the server is chosen when it was alive, but it dies which a request is scheduled to the connection of the server. This case is rare, so moved to next release.

keshonok:
Fixed with a set of patches in #237.

Issue #114, opened 2015-06-15
State of a backend server is not taken into account when scheduling requests


keshonok:
Start backend servers. Start Tempesta with configuration that points to those backend servers. Do an HTTP request with `wget` and see that it's successful. Bring backend servers down. See that Tempesta performs reconnect attempts at some small intervals, and fails as the servers are down. Do another HTTP request with `wget` and see Tempesta (and kernel) crash.

The problem here is that a backend server is still chosen by Tempesta for forwarding the request to, even though it is offline. Tempesta doesn't keep or makes use of current state of a backend server.

The crash may happen in different places depending on current Tempesta state, but more often than not it's something like this:

```
[   54.914695] [tempesta]   request parsed: len=50 parsed=50 msg_len=50 res=0
[   54.914695] [tempesta]   GFSM return code 0
[   54.914695] [tempesta]   Matching request: ffff880038bf4040, list: ffff880038616040
[   54.914695] [tempesta]   tfw_sched_http: use server group: 'default'
[   54.914695] BUG: unable to handle kernel NULL pointer dereference at 0000000000000012
[   54.914695] IP: [<ffffffffa004e03f>] ss_send+0x4f/0x2e6 [tempesta_fw]
[   54.914695] PGD 38a3c067 PUD 38a64067 PMD 0 
[   54.914695] Oops: 0000 [#1] SMP 
[   54.914695] Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tfw_sched_dummy(O) tempesta_fw(O) tempesta_db(O)
[   54.914695] CPU: 0 PID: 0 Comm: swapper/0 Tainted: G        W  O 3.10.10-tempesta #2
[   54.914695] task: ffffffff81565440 ti: ffffffff81554000 task.ti: ffffffff81554000
[   54.914695] RIP: 0010:[<ffffffffa004e03f>]  [<ffffffffa004e03f>] ss_send+0x4f/0x2e6 [tempesta_fw]
[   54.914695] RSP: 0018:ffff88003fc038a0  EFLAGS: 00010202
[   54.914695] RAX: 0000000000000000 RBX: ffff880038644cc0 RCX: ffff880038bf4040
[   54.914695] RDX: ffff880038bf42c0 RSI: ffff880038bf42c0 RDI: 0000000000000000
[   54.914695] RBP: ffff88003fc038f8 R08: ffff8800385ceb30 R09: 0000000000000000
[...]

Issue #228, opened 2015-09-02
Soft lockup on Tempesta start


krizhanovsky:
Seems related to https://github.com/natsys/tempesta/commit/8a130862e2bdd0b81bf814198e65bb85ec20c3f3

On tempesta start:

         # ./tempesta.sh -f --start 
         Loading Tempesta kernel modules...
         Load Frang
         Starting Tempesta...

         Message from syslogd@localhost at Sep  2 17:29:12 ...
          kernel:BUG: soft lockup - CPU#0 stuck for 22s! [sysctl:3726]

Dmesg shows ```ss_connect()``` in call trace:

        [tdb] Start Tempesta DB
        [tempesta] Initializing Tempesta FW kernel module...
        [tempesta] init: cfg_if
        [tempesta] init: tls
        [tempesta] init: http
        [tempesta] init: http_sticky
        [tempesta] init: server
        [tempesta] init: client
        [tempesta] init: sock_srv
        [tempesta] init: sock_clnt
        [tempesta]   register cfg: filter
        [tempesta]   register cfg: cache
        [tempesta]   register cfg: http_sticky
        [tempesta]   register cfg: sock_srv
        [tempesta]   register cfg: sock_clnt
        [tempesta] Registering new scheduler: dummy
        [tempesta]   tfw_sched_hash: init
        [tempesta] Registering new scheduler: hash
        [tempesta]   tfw_sched_http: init
        [tempesta]   register cfg: tfw_sched_http
        [tempesta] Registering new scheduler: http
        [tempesta]   tfw_sched_rr: init
        [tempesta] Registering new scheduler: round-robin
        [tempesta]   register cfg: frang
        [tempesta] Warning: frang module can't be unloaded, so all allocated resources won't freed
        [tempesta] got state via sysctl: start
        [tempesta]   reading configuration file...
        [tempesta]   reading file: /root/tempesta/etc/tempesta_fw.conf
        [tempesta]   file size: 466 bytes
        [tempesta]   read by offset: 0
        [tempesta]   read by offset: 466
        [tempesta] starting all modules...
        [tempesta]   parsing configuration and pushing it to modules...
        [tempesta]   spec handle: 'server'
        [tempesta]   new server group: 'default'
        [tempesta]   spec handle: 'cache'
        [tempesta]   spec handle: 'frang_limits'
        [tempesta]   spec handle: 'request_rate'
        [tempesta]   spec handle: 'request_burst'
        [tempesta]   spec handle: 'connection_rate'
[...]

PR #237, opened 2015-09-07
Solution to issues #114, #118, #184.

keshonok:


tempesta_fw/connection.c
```diff
@@ -121,7 +132,17 @@ tfw_connection_destruct(TfwConnection *conn)
...
+	struct sock *sk;
+
+	spin_lock(&conn->splock);
+	sk = conn->sk;
+	if (sk)
+		ss_sock_hold(sk);
+	spin_unlock(&conn->splock);
+	if (sk) {
+		ss_send(sk, &msg->skb_list);
+		sock_put(sk);
```
keshonok: Use `ss_sock_put()`

tempesta_fw/connection.h
```diff
@@ -105,9 +143,13 @@ tfw_connection_validate_cleanup(TfwConnection *conn)
 	BUG_ON(!conn);
 	BUG_ON(!list_empty(&conn->list));
 	BUG_ON(!list_empty(&conn->msg_queue));
+	BUG_ON(atomic_read(&conn->refcnt) & ~1);
 	BUG_ON(conn->msg);
+
+	spin_lock(&conn->splock);
 	BUG_ON(conn->peer);
 	BUG_ON(conn->sk);
+	spin_unlock(&conn->splock);
```
keshonok: Perhaps locks are not needed here. What do you think?

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -183,13 +183,23 @@ tfw_sched_hash_update_data(TfwSrvGroup *sg)
...
+			 * It should be assumed that scheduler's data is
+			 * only semi-accurate at any point of time.
+			 */
+			spin_lock(&conn->splock);
 			if (!conn->sk) {
+				spin_unlock(&conn->splock);
 				++conn_idx;
 				continue;
 			}
+			spin_unlock(&conn->splock);
```
keshonok: Perhaps locks are unnecessary here. A read of `conn->sk` should appear atomic provided that `sk` member is properly aligned by compiler.

tempesta_fw/sched/tfw_sched_rr.c
```diff
@@ -114,9 +114,20 @@ tfw_sched_rr_update_data(TfwSrvGroup *sg)
...
+			 * to use it. That has to be dealt with elsewhere.
+			 * It should be assumed that scheduler's data is
+			 * only semi-accurate at any point of time.
+			 */
+			spin_lock(&conn->splock);
+			if (!conn->sk) {
+				spin_unlock(&conn->splock);
 				continue;
+			}
+			spin_unlock(&conn->splock);
```
keshonok: Perhaps locks are unnecessary here. A read of `conn->sk` should appear atomic provided that `sk` member is properly aligned by compiler.

tempesta_fw/connection.c
```diff
@@ -50,32 +54,39 @@ tfw_connection_init(TfwConnection *conn)
...
+ * This function does an opposite to what tfw_connection_link_sk() does.
+ * It must be called under write lock on sk->sk_callback_lock to modify
+ * sk->sk_user_data, or the socket must not be bound or connected.
+ * Socket close may occur in SoftIRQ or in user context at STOP time,
+ * and in both cases it's protected with a lock on sk->sk_callback_lock.
+ * conn->sk is modified under a separate lock.
  */
 void
-tfw_connection_unlink_sk(TfwConnection *conn)
+tfw_connection_unlink_sk(TfwConnection *conn, struct sock *sk)
```
krizhanovsky: The sk argument is used for BUG_ON statement only, so I think compiler complains to unused argument if you compile Tempesta w/o DEBUG option. Probably it's better to check sk in the caller function.

tempesta_fw/sock_srv.c
```diff
@@ -119,13 +119,13 @@ tfw_sock_srv_connect_try(TfwSrvConnection *srv_conn)
 	}
 
 	sock_set_flag(sk, SOCK_DBG);
-	tfw_connection_link_sk(&srv_conn->conn, sk);
+	tfw_connection_link_from_sk(conn, sk);
```
krizhanovsky: Nice. This is what I mentioned in 2nd paragraph of comment https://github.com/natsys/tempesta/issues/228#issuecomment-137195351 for issue #228.

tempesta_fw/sock_srv.c
```diff
@@ -119,13 +119,13 @@ tfw_sock_srv_connect_try(TfwSrvConnection *srv_conn)
...
 	ss_set_callbacks(sk);
 
 	r = ss_connect(sk, &addr->sa, tfw_addr_sa_len(addr), 0);
 	if (r) {
 		TFW_ERR("can't initiate a connect to server: error %d\n", r);
-		tfw_connection_unlink_sk(&srv_conn->conn);
+		tfw_connection_unlink_sk(conn, sk);
```
krizhanovsky: You can add assertion ```conn->sk == sk``` just before call of the function or use define wrapper like tfw_connection_unlink_sk_check().

tempesta_fw/connection.h
```diff
@@ -96,6 +102,38 @@ typedef struct {
...
+	atomic_inc(&conn->refcnt);
+}
+
+static inline bool
+tfw_connection_put(TfwConnection *conn)
+{
+	if (unlikely(!conn))
+		return false;
+	if (likely(atomic_read(&conn->refcnt) == 1))
+		smp_rmb();
```
krizhanovsky: Do we need the barrier on x86-64 which we only support?
I created #241 for the compile-time assertion.

tempesta_fw/http.c
```diff
@@ -290,11 +294,42 @@ tfw_http_conn_msg_alloc(TfwConnection *conn)
...
+void
+tfw_http_conn_msg_free(TfwHttpMsg *hm)
+{
+	TfwConnection *msg_conn = NULL;
+
+	if (hm) {
+		msg_conn = hm->conn;
+		tfw_http_msg_free(hm);
+	}
+	if (msg_conn && tfw_connection_put(msg_conn)) {
```
krizhanovsky: ```tfw_connection_put()``` already checks the argument for NULL, so no need to check msg_conn here.

tempesta_fw/http.c
```diff
@@ -290,11 +294,42 @@ tfw_http_conn_msg_alloc(TfwConnection *conn)
...
+		TFW_CONN_TYPE(msg_conn) & Conn_Clnt
+			? tfw_cli_conn_free(msg_conn)
+			: tfw_srv_conn_free(msg_conn);
+	}
```
krizhanovsky: The function could be bit simplier:

        if (unlikely(!hm))
                return;
        
        if (tfw_connection_put(hm->conn)) {
                TFW_CONN_TYPE(msg_conn) & Conn_Clnt
                        ? tfw_cli_conn_free(hm->conn)
                        : tfw_srv_conn_free(hm->conn);
        }
        
        tfw_http_msg_free(hm);

tempesta_fw/sock_srv.c
```diff
@@ -100,17 +100,17 @@ typedef struct {
...
 
 /**
- * Initiate a new connect attempt without blocking
- * until the connection is established.
+ * Initiate a non-blocking connect attempt.
+ * Returns immediately without waiting until a connection is established.
  */
 static int
-tfw_sock_srv_connect_try(TfwSrvConnection *srv_conn)
+tfw_sock_srv_connect_try(TfwConnection *conn)
```
krizhanovsky: Changes in whole patch https://github.com/natsys/tempesta/commit/c5a198fc8d3795e03d1c43311ba20c6ac8e81c2c are doubtful. It's somewhat like broken incapsulation. sock_srv.c is a set of functions working with TfwSrvConnection, a derived class from TfwConnection. So the static functions of sock_srv.c a likely working with various members of TfwSrvConnection (probably some of them in future rather than now), so it's better to keep them like static methods of a class (in C++ terms, TfwSrvConnection in our case) accepting the class as the first argument. However, external interfaces of the class (e.g. tfw_srv_conn_free()) are really could be called with TfwConnection. 

Thus generic rule for such cases: external interfaces can work with generic base class, while internal (static) functions should work with derived class.

tempesta_fw/sock_clnt.c
```diff
@@ -123,15 +121,21 @@ tfw_sock_clnt_new(struct sock *sk)
...
 		sk, conn, cli);
 	return 0;
 
 err_conn_init:
-	tfw_connection_unlink_peer(conn);
-	tfw_connection_unlink_sk(conn);
 	tfw_connection_destruct(conn);
-	tfw_cli_conn_free(conn);
+	if (tfw_connection_put(conn))
+		tfw_cli_conn_free(conn);
```
krizhanovsky: Double connection freeing. The second freeing is at tfw_connection_destruct() -> conn_destruct hook  which is tfw_http_conn_destruct() -> tfw_http_conn_msg_free() -> tfw_connection_put() & tfw_cli_conn_free()

tempesta_fw/sock_clnt.c
```diff
@@ -123,15 +121,21 @@ tfw_sock_clnt_new(struct sock *sk)
 		goto err_conn_init;
 	}
 
+	spin_lock(&conn->splock);
+	tfw_connection_link_sk(conn, sk);
+	tfw_connection_link_peer(conn, (TfwPeer *)cli);
```
krizhanovsky: TfwPeer lives longer than TfwConnection by its definition, so access to it should not be managed by shorter living objects like TfwConnection. Moreover, the unlink operation is synchronized by TfwPeer->conn_lock.

It's unclear why we need to synchronize access to conn->peer. If it must be synchronized, then probably we should do this in different way...

tempesta_fw/connection.c
krizhanovsky: Looks bit unusual bedause double sk checking on most probable workflow, I'd rewrite it like

        sk = conn->sk;
        if (unlikely(!sk)) {
                spin_unlock(&conn->splock);                                    
                return;                                                        
        }                                                                      
        
        ss_sock_hold(sk);
        spin_unlock(&conn->splock);                                            
        
        ss_send(sk, &msg->skb_list);                                           
        
        ss_sock_put(sk);

tempesta_fw/sock_clnt.c
```diff
@@ -157,10 +161,14 @@ tfw_sock_clnt_drop(struct sock *sk)
 	 */
 	r = tfw_classify_conn_close(sk);
 
+	spin_lock(&conn->splock);
+	tfw_connection_unlink_sk(conn, sk);
 	tfw_connection_unlink_peer(conn);
```
krizhanovsky: The function should not zero conn->peer - it has no sense.

tempesta_fw/sock_srv.c
```diff
@@ -437,13 +448,13 @@ tfw_sock_srv_add_conns(TfwServer *srv, int conns_n)
...
+tfw_sock_srv_del_conns(TfwServer *srv)
 {
-	TfwSrvConnection *srv_conn, *tmp;
+	TfwConnection *conn, *tmp;
 
-	list_for_each_entry_safe(srv_conn, tmp, &srv->conn_list, conn.list) {
-		tfw_connection_unlink_peer(&srv_conn->conn);
-		tfw_srv_conn_free(srv_conn);
+	list_for_each_entry_safe(conn, tmp, &srv->conn_list, list) {
+		tfw_connection_unlink_peer(conn);
```
krizhanovsky: The function should not zero conn->peer - it has no sense.

tempesta_fw/connection.h
krizhanovsky: Yeah, agree. Reading a pointer is atomic operation and this is just a validation.

tempesta_fw/sched/tfw_sched_hash.c
krizhanovsky: Yes, exactly

tempesta_fw/sched/tfw_sched_rr.c
krizhanovsky: Yes, as the above

tempesta_fw/http_sticky.c
```diff
@@ -249,13 +259,25 @@ tfw_http_sticky_add(TfwHttpMsg *hm, u_char *value, size_t len)
...
+	 * to protect that with a lock. If the client connection
+	 * is already gone, then there's nothing we can do here.
+	 */
+	spin_lock(&hmreq->conn->splock);
+	client = (TfwClient *)hmreq->conn->peer;
+	if (client)
+		tfw_http_prep_hexstring(buf, client->cookie.hmac, len);
+	spin_unlock(&hmreq->conn->splock);
+	if (client == NULL)
+		return -ENOENT;
```
krizhanovsky: Which data exactly and where can be concurrently accessed that we must protect it with the lock?
There is also client value double check.

tempesta_fw/sock.c
```diff
@@ -145,6 +145,8 @@ ss_send(struct sock *sk, const SsSkbList *skb_list)
 	/* Synchronize concurrent socket writting in different softirqs. */
 	bh_lock_sock_nested(sk);
 
+	if (sk->sk_state != TCP_ESTABLISHED)
+		goto out;
```
krizhanovsky: We definitely can send data in CLOSE_WAIT state. Probably we also can send in SYN_SENT and/or SYN_RCVD - need to check.

tempesta_fw/connection.h
```diff
@@ -54,20 +54,26 @@ enum {
  * Session/Presentation layer (in OSI terms) handling.
```
krizhanovsky: Please add comment that TfwConnection can live longer than underlying TCP connection, for example in connection failovering case. This is important design point which means that TfwConnection could be alive and valid, while sk member in it fully unusable.

Meantime TfwPeer is the property of TfwConnection and it's always the same for the same connection, i.e. TfwConnection has no sense w/o live TfwPeer and can't change its peer.

tempesta_fw/connection.h
```diff
@@ -54,20 +54,26 @@ enum {
...
 	SsProto			proto;
 	struct list_head	list;
 	struct list_head	msg_queue;
+	spinlock_t		msg_qlock;
+	atomic_t		refcnt;
 
 	TfwMsg			*msg;
 	TfwPeer 		*peer;
 	struct sock		*sk;
+	spinlock_t		splock;
```
krizhanovsky: It'd be better to place the member near from msg_qlock` to avoid unnecessary memory holes due to alignment.

tempesta_fw/http_sticky.c
```diff
@@ -232,8 +240,10 @@ tfw_http_sticky_set(TfwHttpMsg *hm)
 	+ STICKY_NAME_MAXLEN + 1 + STICKY_KEY_MAXLEN * 2 + 2
 
 static int
-tfw_http_sticky_add(TfwHttpMsg *hm, u_char *value, size_t len)
+tfw_http_sticky_add(TfwHttpMsg *hmresp, TfwHttpMsg *hmreq)
 {
+	TfwClient *client;
+	unsigned int len = sizeof(client->cookie.hmac);
 	char buf[len * 2];
 	TfwStr set_cookie = {
```
krizhanovsky: Bad stack memory alignment: 8, 4, 40 and 24 bytes are allocated. It's better to place them like 4, 8, 24, 40 to minimized 8-byte alignment

tempesta_fw/sock_srv.c
```diff
@@ -186,20 +187,24 @@ static int
...
 	int r;
-	TfwSrvConnection *srv_conn = sk->sk_user_data;
-	TfwServer *srv = (TfwServer *)srv_conn->conn.peer;
+	TfwConnection *conn = sk->sk_user_data;
+	TfwServer *srv = (TfwServer *)conn->peer;
+
+	/* Link Tempesta with the socket. */
+	spin_lock(&conn->splock);
+	tfw_connection_link_to_sk(conn, sk);
+	spin_unlock(&conn->splock);
```
krizhanovsky: Again, tfw_connection_link_to_sk() is just a 8-byte assignment. What do you synchronize?

tempesta_fw/sock_srv.c
```diff
@@ -213,28 +218,29 @@ static int
...
 
 	TFW_DBG_ADDR("connection lost", &srv->addr);
 
-	/* Revert tfw_sock_srv_connect_complete(). */
-	tfw_sg_update(srv->sg);
-	tfw_connection_destruct(&srv_conn->conn);
+	/* Withdraw from socket activity. */
+	spin_lock(&conn->splock);
+	tfw_connection_unlink_sk(conn, sk);
+	spin_unlock(&conn->splock);
```
krizhanovsky: Also it seems there is nothing to synchronize. There are other similar locking places at the below, need to carefully review the locks. If the locks are needed, then probably it's better to move them to tfw_connection_unlink_sk() and similar functions.

tempesta_fw/sock_srv.c
```diff
@@ -283,35 +291,44 @@ static const SsHooks tfw_sock_srv_ss_hooks = {
...
+	 * taken under conn->splock as we would get lock inversion. It is
+	 * taken in Sync Sockets, and conn->splock is taken under that lock.
+	 * So we need to adhere to correct order of locking.
 	 */
+	spin_lock(&conn->splock);
+	sk = conn->sk;
+	if (sk)
+		ss_sock_hold(sk);
+	spin_unlock(&conn->splock);
 	if (sk) {
```
krizhanovsky: The same as with tfw_connection_send - double variable check on main path.

tempesta_fw/sock_srv.c
```diff
@@ -283,35 +291,44 @@ static const SsHooks tfw_sock_srv_ss_hooks = {
...
+		/* Withdraw from socket activity and close the socket. */
 		ss_callback_write_lock(sk);
```
krizhanovsky: By the way, why do you acquire sk_callback_lock for writing here?

tempesta_fw/http.c
```diff
@@ -446,25 +488,27 @@ tfw_http_req_cache_cb(TfwHttpReq *req, TfwHttpResp *resp, void *data)
...
-			tfw_http_msg_free((TfwHttpMsg *)req);
+			tfw_http_conn_msg_free((TfwHttpMsg *)req);
 			return;
 		}
 		if (tfw_http_adjust_req(req))
 			goto send_500;
 
-		/* Add request to the connection. */
+		/* Dispatch request to an appropriate server. */
+		conn = tfw_sched_get_srv_conn((TfwMsg *)req);
```
krizhanovsky: This is TfwConnection user - why don't you increment reference counter of the connection? Typically, we should do this just when we get the connection from scheduler data structure.

tempesta_fw/http.c
```diff
@@ -309,11 +344,13 @@ tfw_http_conn_destruct(TfwConnection *conn)
 {
 	TfwMsg *msg, *tmp;
 
-	tfw_http_msg_free((TfwHttpMsg *)conn->msg);
-
+	spin_lock(&conn->msg_qlock);
 	list_for_each_entry_safe(msg, tmp, &conn->msg_queue, msg_list)
```
krizhanovsky: It's good to have assertion here ```BUG_ON(conn == msg->conn);```

tempesta_fw/http.c
```diff
@@ -446,25 +488,27 @@ tfw_http_req_cache_cb(TfwHttpReq *req, TfwHttpResp *resp, void *data)
...
 		if (tfw_http_adjust_req(req))
 			goto send_500;
 
-		/* Add request to the connection. */
+		/* Dispatch request to an appropriate server. */
+		conn = tfw_sched_get_srv_conn((TfwMsg *)req);
+		if (conn == NULL) {
+			TFW_ERR("Unable to find a backend server\n");
+			goto send_404;
+		}
```
krizhanovsky: Schedulers should make a decision on plain unmodified request, so the scheduling should be done before any request mangling.

Issue #203, opened 2015-08-23
LWS* between header and value for raw headers


avbelov23:
LWS* was not handled between header and value for raw headers in match_hdr_raw().
Currently "X-Hdr:foo" is not equal to "X-Hdr: foo.

avbelov23:
Now TEST(http_match, headers_eq) fall from tempesta_fw/t/unit/test_http_match.c

PR #238, opened 2015-09-07
#203. LWS* between name and value and after value for raw headers

avbelov23:
In accordance with 3.2 RFC 7230 header value can contain groups of whitespace characters. We consider them the equivalent of a single whitespace character, if there is whitespace character / characters in the rule in the same position. For unification, also work with the keys, although whitespace characters are not permitted in the key.

### DIFF
--- a/tempesta_fw/http_match.c
+++ b/tempesta_fw/http_match.c
@@ -62,6 +62,7 @@
  */
 
 #include <linux/cache.h>
+#include <linux/ctype.h>
 #include "http_match.h"
 #include "http_msg.h"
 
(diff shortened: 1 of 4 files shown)


tempesta_fw/http_match.c
```diff
@@ -201,27 +202,99 @@ match_hdr(const TfwHttpReq *req, const TfwHttpMatchRule *rule)
...
+			while (p != pend && c != cend) {
+				if (tolower(*p) != tolower(*c)) {
+					/* If the position of the strings have a
+					 * different number of space characters,
+					 * or one of the strings have space
+					 * characters after ':', missing their
+					 */
+					if (unlikely((
+						(isspace(*p) &&
+							(prev == *p ||
```
krizhanovsky: It's not a normal case if some of the strings start with space, but it's good to initialized ```prev``` with some value or do implicit assertion that the strings don't start w/ space. I'm wondering why compiler doesn't notify about possible uninitialized ```prev``` using.

tempesta_fw/http_match.c
```diff
@@ -201,27 +202,99 @@ match_hdr(const TfwHttpReq *req, const TfwHttpMatchRule *rule)
...
+								prev == ':'))
+						||
+						(isspace(*c) &&
+							(prev == *c ||
+								prev == ':'))
```
krizhanovsky: ```prev``` is checked against ':' twice. You can easily rewrite the expression with only one check. Meantime there is nothing bad to do couple of simple checks with the same body instead of this one large condition - compiler also generates just set of conditional jumps.

tempesta_fw/http_match.c
```diff
@@ -201,27 +202,99 @@ match_hdr(const TfwHttpReq *req, const TfwHttpMatchRule *rule)
...
+						))) {
+						goto state_sp;
```
krizhanovsky: Regardless whether ```p``` or ```c``` is space you're trying to move both of them - just unnecessary conditional branches in _MOVE_TO_COND.

tempesta_fw/http_match.c
```diff
@@ -201,27 +202,99 @@ match_hdr(const TfwHttpReq *req, const TfwHttpMatchRule *rule)
...
+	}
 
-		/* TODO: handle LWS* between header and value for raw headers.
-		 * (currently "X-Hdr:foo" is not equal to "X-Hdr: foo").
-		 */
-		if (tfw_str_eq_cstr(hdr, rule->arg.str, rule->arg.len, flags))
-			return true;
+state_common:
+			while (p != pend && c != cend) {
+				if (tolower(*p) != tolower(*c)) {
```
krizhanovsky: Why do you downcase ```p```? You can do this at configuration stage and avoid the operation in runtime.

tempesta_fw/http_match.c
```diff
@@ -201,27 +202,99 @@ match_hdr(const TfwHttpReq *req, const TfwHttpMatchRule *rule)
...
+						))) {
+						goto state_sp;
+					}
+
+					return false;
+				}
+
+				p++;
+				c++;
+				prev = *(p - 1);
```
krizhanovsky: Why not to move the statement before ```p++``` or just use classic ```prev = *p++``` to avoid the decrement?

tempesta_fw/http_match.c
```diff
@@ -201,27 +202,99 @@ match_hdr(const TfwHttpReq *req, const TfwHttpMatchRule *rule)
...
+					 * or one of the strings have space
+					 * characters after ':', missing their
+					 */
+					if (unlikely((
+						(isspace(*p) &&
+							(prev == *p ||
+								prev == ':'))
+						||
+						(isspace(*c) &&
+							(prev == *c ||
```
krizhanovsky: It seems the lines "aaa: \tbbb" and "aaa: bbb" treated as not equal (```*c == '\t' while ```prev == ' '```), however tab and space are LWS in HTTP.

tempesta_fw/http_match.c
```diff
@@ -201,27 +202,99 @@ match_hdr(const TfwHttpReq *req, const TfwHttpMatchRule *rule)
...
+			}
+			_TRY_NEXT_CHUNK(goto state_common, );
+
+			if (c == cend && p == pend) {
+				return true;
+			}
+
+			if (p == pend && flags & TFW_STR_EQ_PREFIX) {
+				return true;
+			}
```
krizhanovsky: I'd place the code before ```_TRY_NEXT_CHUNK(goto state_common, );``` because now you're doing double check of ```p != pend``` if rule is finished.

tempesta_fw/http_match.c
```diff
@@ -201,27 +202,99 @@ match_hdr(const TfwHttpReq *req, const TfwHttpMatchRule *rule)
...
+
+state_sp:
+			_MOVE_TO_COND(p, pend, !isspace(*p));
+			_MOVE_TO_COND(c, cend, !isspace(*c));
+			_TRY_NEXT_CHUNK(goto state_sp, return p == pend);
+			goto state_common;
```
krizhanovsky: Case 1: we're from while() by goto statement. Since while() check passed, we know that neither c nor p reach end of their strings, so why do we try the next chunk? At least one unnecessary condition, but also unclear code.

Case 2: we're here due to end of pattern - what's the sense to try the next chunk?

Code between lines 282-295 do 3 checks for ```c == end``` - this has no sense.

I'd say you're trying to generalize different cases and pay by unnecessary checks and conditional jumps (which are expensive) for the generalization.

tempesta_fw/http_match.c
```diff
@@ -201,27 +202,99 @@ match_hdr(const TfwHttpReq *req, const TfwHttpMatchRule *rule)
...
+					 */
+					if (unlikely((
+						(isspace(*p) &&
+							(prev == *p ||
+								prev == ':'))
+						||
+						(isspace(*c) &&
+							(prev == *c ||
+								prev == ':'))
+						))) {
```
krizhanovsky: If you do such complex conditions, please properly align the statements by spaces to make them more readable:

                                        if (unlikely(((isspace(*p)
                                                       && (prev == *p
                                                           || prev == ':'))
                                                     || (isspace(*c)
                                                         && (prev == *c
                                                             || prev == ':')))))
                                        {
                                                goto state_sp;
                                        }

tempesta_fw/sched/tfw_sched_http.c
```diff
@@ -302,6 +304,13 @@ tfw_sched_http_cfg_handle_match(TfwCfgSpec *cs, TfwCfgEntry *e)
 	memcpy(rule->rule.arg.str, in_arg, arg_len);
 	rule->rule.arg.type = tfw_sched_http_cfg_arg_tbl[field];
 
+	if (rule->rule.field == TFW_HTTP_MATCH_F_HDR_RAW) {
+		p = rule->rule.arg.str;
+		while ((*p = tolower(*p))) {
+			p++;
+		}
+	}
```
krizhanovsky: This is http_match specific code and you already has 2 instances of the code, so it's good to move as function to http_match.[ch]

Issue #243, opened 2015-09-08
Failures at HTTP parser unit test


avbelov23:
[  887.903086] tfw_test: finish - failed 2055 assertions

PR #244, opened 2015-09-08
#243. Failures at HTTP parser unit test

avbelov23:
Incorrectly calculated the length of the value of special headers in tfw_http_msg_hdr_val and because of this, comparison tfw-string and null-terminated strings work incorrectly. It does not take into account possible whitespace characters in the first chunk after a header key.

### DIFF
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -28,7 +28,7 @@
 #include "ss_skb.h"
 
 /**
- * Fills @val with second part of specula HTTP header containing the header
+ * Fills @val with second part of special HTTP header containing the header
  * value.
  */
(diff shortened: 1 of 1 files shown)


tempesta_fw/http_msg.c
```diff
@@ -51,13 +51,13 @@ tfw_http_msg_hdr_val(TfwStr *hdr, int id, TfwStr *val)
 	BUG_ON(id >= TFW_HTTP_HDR_RAW);
```
krizhanovsky: Please, fix also misprint in header comment: "specula" -> "special".

krizhanovsky:
Good to merge with additional comment fix.

PR #252, opened 2015-09-10
tfw_fuzzer: fuzzing requests and responses in the HTTP parser unit test: #150

stefanstrogin:


tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -85,45 +165,64 @@ reset_req(size_t data_len)
  *  >  0 - EOF: all possible fragments are parsed, terminate the loop.
  */
 int
-do_split_and_parse(unsigned char *req_str)
+do_split_and_parse(unsigned char *str, int type)
```
krizhanovsky: The function should be declared as static

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -85,45 +165,64 @@ reset_req(size_t data_len)
...
+		reset_req(strlen(str));
+	else if (type == FUZZ_RESP)
+		reset_resp(strlen(str));
+	else
+		BUG();
 
 	/* First iteration. */
-	if (!req_len) {
-		req_len = strlen(req_str);
+	if (!len) {
```
krizhanovsky: ```len``` isn't initialized at this point. The same story with ```chunks``` below

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -47,11 +49,88 @@ reset_req(size_t data_len)
...
+}
+
+int
+split_and_parse_n(unsigned char *str, int type, size_t len, int chunks)
+{
+	int chlen = len / chunks, rem = len % chunks, pos = 0, step, r;
+	char tmp1, tmp2;
+
+	while (pos < len)
+	{
```
krizhanovsky: This doesn't satisfy our coding style: https://github.com/krizhanovsky/linux-3.10.10-sync_sockets/blob/master/Documentation/CodingStyle

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -131,39 +230,58 @@ do_split_and_parse(unsigned char *req_str)
...
+		r = tfw_http_parse_resp(resp, head_buf, head_len);
 	if (r != TFW_POSTPONE)
 		return r;
 
 	/* Parse the tail. */
-	return tfw_http_parse_req(req, tail_buf, tail_len);
+	if (type == FUZZ_REQ)
+		return tfw_http_parse_req(req, tail_buf, tail_len);
+	else
+		return tfw_http_parse_resp(resp, tail_buf, tail_len);
```
krizhanovsky: Whole the function and split_and_parse_n() are messy and must be reworked. Is there a real reason why to copy the piece of code

                if (type == FUZZ_REQ)                                          
                        r = tfw_http_parse_req(req, buf, size);          
                else                                                           
                        r = tfw_http_parse_resp(resp, buf, size); 

4 times? While 2 chunks and more are handled differently? Why at the end of do_split_and_parse() we call the parser exactly twice? All the thing make a code reader wondering and must be reworked.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -47,11 +49,88 @@ reset_req(size_t data_len)
...
+		}
+
+		if (pos) {
+			tmp1 = *(str + pos - 1);
+			*(str + pos - 1) = 0;
+		}
+		if (pos + step < len) {
+			tmp2 = *(str + pos + step);
+			*(str + pos + step) = 0;
+		}
```
krizhanovsky: This is very unusual thing, but it has sense. Please always comment such code.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -85,45 +165,64 @@ reset_req(size_t data_len)
...
 	++head_len;
 	--tail_len;
 
-	/* Done all iterations?. */
-	if (head_len == req_len) {
-		req_len = head_len = tail_len = 0;
-		return 1;
+	if (head_len == len) {
+		++chunks;
+		return 0;
```
avbelov23: Here we go to the testing of the request after the return success (0) without parsing.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -47,11 +49,88 @@ reset_req(size_t data_len)
...
+	free_resp();
+	alloc_resp(data_len);
+}
+
+static void
+free_msgs(void)
+{
+	free_req();
+	free_resp();
+}
```
krizhanovsky: Some of the functions above are used only once in the code and actually just a 2-line wrappers - why do you use them? Too many and too similar (e.g. alloc_resp() and test_resp_alloc()) makes problems to understand the code - a code reader must keep them all in their head.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -85,45 +165,64 @@ reset_req(size_t data_len)
  *  >  0 - EOF: all possible fragments are parsed, terminate the loop.
  */
 int
-do_split_and_parse(unsigned char *req_str)
+do_split_and_parse(unsigned char *str, int type)
 {
 	static char head_buf[PAGE_SIZE];
 	static char tail_buf[PAGE_SIZE];
-	static size_t req_len, head_len, tail_len;
+	static size_t len, head_len, tail_len;
```
krizhanovsky: Well, we define the variables static to make a loop. This is not the better and easier to understand solution. The function should be reworked as set of loops and functions

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -365,13 +483,53 @@ TEST(http_parser, parses_connection_value)
...
 }
 
+TEST(http_parser, fuzzer)
+{
+	char *str;
+	int i = 0, ret;
+	str = vmalloc(2 * 1024 * 1024);
+
+	for (i = 0; i < 2; i++) {
+		TEST_LOG("fuzz request i: %d\n", i);
```
krizhanovsky: What ```i``` is? Need a better naming, comments or better pint message.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -365,13 +483,53 @@ TEST(http_parser, parses_connection_value)
...
+		ret = fuzz_gen(str, 5, FUZZ_REQ);
+		switch (ret) {
+		case FUZZ_VALID:
+			FOR_REQ(str);
+			break;
+		case FUZZ_INVALID:
+			EXPECT_BLOCK_REQ(str);
+			break;
+		case FUZZ_END:
+			goto end;
```
krizhanovsky: -EINVAL returned from fuzz_gen() is not handled in this switch and swith at the below.

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,563 @@
...
+	v |= add_transfer_encoding(str);
+	strcat(str, "\r\n");
+	v |= add_duplicates(str, TRANSFER_ENCODING, &add_transfer_encoding);
+
+	strcat(str, "\r\n");
+
+	v |= add_body(str, type);
+
+	for (i = 0; i < move; i++) {
+		ret = gen_vector_move((N_FIELDS - 1 + i * ((i % 2) ? 3 : 2)) % N_FIELDS);
```
krizhanovsky: Too long line, please fit 80 characters

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,563 @@
...
+#define A_URI "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_.~!*'();:@&=+$,/?%#[]"
+#define A_URI_INVAL "`\\\03\023\07\r"
+#define A_USER_AGENT A_URI
+#define A_USER_AGENT_INVAL A_URI_INVAL
+#define A_HOST "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-[]:"
+#define A_HOST_INVAL "\\`~!*'();@&=+$,/?%#\03\023\07\r"
+#define A_X_FORWARDED_FOR A_HOST
+#define A_X_FORWARDED_FOR_INVAL A_HOST_INVAL
+
+static const char *a_body = "ABCxyz013\r\n\t-_.~!*'();:@&=+$,/?%#[]\n\r\023\07\013\014\x89\x90\xa0\xc0";
```
krizhanovsky: Too long line.

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,563 @@
...
+}
+
+static int add_method(char *s1)
+{
+	struct fuzz_msg r = methods[gen_vector[METHOD].i];
+
+	if (r.s) {
+		strcat(s1, r.s);
+		return r.inval;
+	} else {
```
krizhanovsky: Usually we don't make ELSE branch if IF branch ends with return.

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,563 @@
...
+		if (gen_vector[CONTENT_LENGTH].i % 2) {
+			add_string(s1, gen_vector[METHOD].i,
```
krizhanovsky: gen_vector[METHOD].i = 0, so do we generate empty method here and return FUZZ_VALID?

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,563 @@
...
+		else if (i % 3)
+			addch(s1, '\r');
+		else
+			addch(s1, ' ');
+	}
+
+	return s1;
+}
+
+static int add_method(char *s1)
```
krizhanovsky: Please read our coding style: function return type and function namre are always on the different lines.

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,563 @@
...
+		if (gen_vector[CONTENT_LENGTH].i % 2) {
+			add_string(s1, gen_vector[URI_FILE].i,
+				   gen_vector[URI_FILE].a_val);
+			return FUZZ_VALID;
+		} else
+			add_string(s1, gen_vector[URI_FILE].i,
+				   gen_vector[URI_FILE].a_inval);
+	}
+
+	return FUZZ_INVALID;
```
krizhanovsky: There are a lot of generating functions with very similar bodies, so it seems they can be reworked to single function which will also accept header name (or "" if it's not a header generator) and field constant (e.g. URI_FILE).

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,563 @@
...
+	strcat(str, "\r\n");
+	v |= add_duplicates(str, CONTENT_TYPE, &add_content_type);
+
+	v |= add_content_length(str);
+	strcat(str, "\r\n");
+	v |= add_duplicates(str, CONTENT_LENGTH, &add_content_length);
+
+	v |= add_transfer_encoding(str);
+	strcat(str, "\r\n");
+	v |= add_duplicates(str, TRANSFER_ENCODING, &add_transfer_encoding);
```
krizhanovsky: Not enough headers. Here is an example of standard browser request:

         GET / HTTP/1.1
         Host: 172.16.0.5
         User-Agent: Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:38.0) Gecko/20100101 Firefox/38.0
         Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
         Accept-Language: en-US,en;q=0.5
         Accept-Encoding: gzip, deflate
         Connection: keep-alive

So we should at least test these Accept-* headers. Also Cookie and Set-Cookie are missed.

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,563 @@
...
+	if (type == FUZZ_REQ)
+		return FUZZ_VALID;
+
+	rc = resp_code[gen_vector[RESP_CODE].i].s;
+	if (!strstr(rc, "100") || !strstr(rc, "101") || !strstr(rc, "102") ||
+			!strstr(rc, "204") || !strstr(rc, "304"))
+		return FUZZ_INVALID;
+
+	return FUZZ_VALID;
+}
```
krizhanovsky: Chunked body should be tested, see https://tools.ietf.org/html/rfc7230#section-4.1

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,563 @@
...
+static struct fuzz_msg conn_val[] = {{"keep-alive", 0}, {"close", 0},
+	{"upgrade", 0}, {NULL, }};
+static struct fuzz_msg ua_val[] = {{"Wget/1.13.4 (linux-gnu)", 0},
+	{"Mozilla/5.0", 0}, {NULL, }};
+static struct fuzz_msg hosts[] = {{"localhost", 0}, {"127.0.0.1", 0},
+	{"example.com", 0}, {"xn--80aacbuczbw9a6a.xn--p1ai", 0}, {NULL, }};
+static struct fuzz_msg content_type[] = {{"text/html;charset=utf-8", 0},
+	{"image/jpeg", 0}, {"text/plain", 0}, {NULL, }};
+static struct fuzz_msg content_length[] = {{"0", 0}, {"10", 1}, {"-42", 1},
+	{"146", 1}, {"0100", 1}, {"100500", 1}, {NULL, }};
```
krizhanovsky: It's good to test invalid string values like ```efg``` and very large numbers (large than 64 bit) like "289938893745782652477004635711254524". The same for chunks length in chunked transfer encoding.

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,563 @@
...
+		v |= add_method(str);
+		add_spaces(str);
+
+		add_uri_path_start(str);
+		if (!gen_vector[URI_PATH_DEPTH].i)
+			addch(str, '/');
+		for (i = 0; i < gen_vector[URI_PATH_DEPTH].i; ++i) {
+			addch(str, '/');
+			v |= add_uri_file(str);
+		}
```
krizhanovsky: URI can be extremely large, like tens of kilobytes. We should test such cases.

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,563 @@
...
+		return r.inval;
+	} else {
+		if (gen_vector[CONTENT_LENGTH].i % 2) {
+			add_string(s1, gen_vector[USER_AGENT].i,
+				   gen_vector[USER_AGENT].a_val);
+			return FUZZ_VALID;
+		} else
+			add_string(s1, gen_vector[USER_AGENT].i,
+				   gen_vector[USER_AGENT].a_inval);
+	}
```
krizhanovsky: Some time ago we saw very large User-Agent value, it was Internet Explorer with sequence of tons of updates. It hase several kilobytes in size. Generally speaking, it's good to generate such large values for all headers at least once.

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,563 @@
...
+	v |= add_body(str, type);
+
+	for (i = 0; i < move; i++) {
+		ret = gen_vector_move((N_FIELDS - 1 + i * ((i % 2) ? 3 : 2)) % N_FIELDS);
+		if (ret == FUZZ_END)
+			break;
+	}
+	if (v & FUZZ_INVALID)
+		return FUZZ_INVALID;
+	return ret;
```
krizhanovsky: #150 ends by the list of papers with many HTTP attacks and no one of them is implemented in the fuzzer. I think most of them can be encoded as special values for header values (e.g. request smuggling should be just a one-two special values for any of the headers). Please review the papers for examples of HTTP attacks and accurately implement them in the fuzzer.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -47,11 +49,88 @@ reset_req(size_t data_len)
...
+		step = chlen;
+
+		if (rem) {
+			step += rem;
+			rem = 0;
+		}
+
+		if (pos) {
+			tmp1 = *(str + pos - 1);
+			*(str + pos - 1) = 0;
```
avbelov23: I do not see the need for this code.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -323,6 +328,73 @@ TEST(http_parser, fills_hdr_tbl)
...
+		"Dummy5: 5\r\n"
+		"Dummy6: 6\r\n"
+		"Content-Length: 0\r\n"
+		"Content-Type: text/html; charset=iso-8859-1\r\n"
+		"Dummy7: 7\r\n"
+		"Dummy8: 8\r\n"
+		"Dummy9: 9\r\n"
+		"Cache-Control: max-age=0, private, min-fresh=42\r\n"
+		"Expires: Tue, 31 Jan 2012 15:02:53 GMT\r\n"
+		"Keep-Alive: timeout=600, max=65526\r\n"
```
krizhanovsky: Please also add Transfer-Encoding

tempesta_fw/t/unit/tfw_fuzzer.c
```diff
@@ -0,0 +1,541 @@
...
+	int i, tmp, v = FUZZ_VALID;
+	static int curr_duplicates = 0;
+
+	if (curr_duplicates++ % DUPLICATES_PERIOD)
+		return FUZZ_VALID;
+
+	for (i = 0; i < curr_duplicates % MAX_DUPLICATES; ++i) {
+		if (gen_vector[t].dissipation) {
+			tmp = gen_vector[t].i;
+			gen_vector[t].i = (gen_vector[t].i + i) % gen_vector[t].size;
```
krizhanovsky: Too long line

Issue #256, opened 2015-09-11
The use of ss_close() in frang.c leads to kernel panic or a possible deadlock.


keshonok:
There's a code path in `frang.c` that makes use of `ss_close()`: https://github.com/natsys/tempesta/blob/master/tempesta_fw/classifier/frang.c#L303 If it is triggered, then `ss_close()` is called. Calling it in these circumstances causes kernel panic on a `BUG_ON(sk->sk_user_data == NULL)`. 
```
[   16.602319] [tempesta]   frang_req_limit: close connection
[   16.602319] ------------[ cut here ]------------
[   16.602319] kernel BUG at /home/natsys-lab/tfw-run/tempesta/tempesta_fw/sock.c:427!
[   16.602319] invalid opcode: 0000 [#1] SMP 
[   16.602319] Modules linked in: tfw_frang(O) tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tfw_sched_dummy(O) tempesta_fw(O) tempesta_db(O)
[   16.602319] CPU: 0 PID: 0 Comm: swapper/0 Tainted: G           O 3.10.10-tempesta+ #31
[   16.602319] task: ffffffff81565440 ti: ffffffff81554000 task.ti: ffffffff81554000
[   16.602319] RIP: 0010:[<ffffffffa00363d8>]  [<ffffffffa00363d8>] ss_close+0x25/0x6e [tempesta_fw]
[   16.602319] RSP: 0018:ffff88003fc03818  EFLAGS: 00010202
[   16.602319] RAX: 0000000000000001 RBX: ffff8800385d4100 RCX: 0000000000000000
[   16.602319] RDX: 0000000000000002 RSI: ffffffff81565a20 RDI: ffff8800385d4100
[   16.602319] RBP: ffff88003fc03820 R08: 0000000000000000 R09: 0000000000000000
[   16.602319] R10: 0000000000000001 R11: ffff88003fc034ae R12: ffff8800388ffb80
[   16.602319] R13: ffff8800385d48e0 R14: 0000000000000000 R15: ffff8800385d48e0
[   16.602319] FS:  0000000000000000(0000) GS:ffff88003fc00000(0000) knlGS:0000000000000000
[   16.602319] CS:  0010 DS: 0000 ES: 0000 CR0: 000000008005003b
[   16.602319] CR2: 00000000004112a0 CR3: 0000000038b30000 CR4: 00000000000006b0
[...]

PR #259, opened 2015-09-12
#256. Remove the posible deadlock

avbelov23:
The deadlock may be when frag wish block a request, closing the socket, again acquire hb->lock


krizhanovsky:
Good to merge

Issue #265, opened 2015-09-13
Poor DoS handling


krizhanovsky:
1. run Tempesta with enabled Frang
2. use some simple DoS tool like tcping or any other tool which established many connections to hit conn_max
3. try to access protected upstream through Tempesta from other IP address.

There are at least two problems:

1. Frang doesn't log blocked IP address, however it's expected. Actually there should be logging function which does common thing w/o current copy & paste and with logging src IP;
2. you can't reach the protected site - it seems you'll also be blocked.

krizhanovsky:
Performance optimizations for rate limiting and ports checking: https://github.com/natsys/tempesta/commit/7905f1016dc7162818c93eed9b2d6d807187c169

PR #267, opened 2015-09-15
Fix the counting of the number of concurrent connections. (#265)

keshonok:
The number of concurrent connections is maintained by classifier that
is called when a connection is established or closed. The number is
inremented when new connection is established. If the limit has been
exceeded on the number, the request is blocked by classifier. It's
expected that the connection is closed when that happens. Classifier
is called again on connection close event, and that's when the number
of concurrent connections is decremented. However, the connection for
the client is not initialized at the time classifier is called. That
lead to the connection drop hook not being called from Sync Sockets,
and therefore classifier was not called on connection close event.
This fix makes sure that classifier is called with connection drop
event when new connection is blocked by classifier.


krizhanovsky:
Good to merge

Issue #257, opened 2015-09-11
test_http_parser: wrong parsing of a request splitted into many chunks 


stefanstrogin:
1) Compile Tempesta with the patch set from https://github.com/natsys/tempesta/tree/ss-fuzzer
2) Run all http_parser unit tests.
It fails on the test parses_req_uri on parsing "GET /a/b/c/dir/?foo=1&bar=2#abcd HTTP/1.1\r\n\r\n".
May be related with #207?
```
[tempesta]   parse 15 client data bytes (GET /a/b/c/dir/) on req=ffff88002ff85040
[tempesta]   enter FSM at state 0
[tempesta]   parser: Req_0(0:0): c=0x47(G), r=-2
[tempesta]   parser: Req_Method(1:0): c=0x47(G), r=-2
[tempesta]   parser: Req_Uri(11:0): c=0x2f(/), r=-2
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x61(a), r=-2
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x2f(/), r=-2
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x62(b), r=-2
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x2f(/), r=-2
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x63(c), r=-2
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x2f(/), r=-2
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x64(d), r=-2
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x69(i), r=-2
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x72(r), r=-2
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x2f(/), r=-2
[tempesta]   store header chunk len=15 data=ffffffffa023ece0 hdr=<0x0,0,          (null)>
[tempesta]   parse 15 client data bytes (?foo=1&bar=2#ab) on req=ffff88002ff85040
[tempesta]   enter FSM at state 22
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x3f(?), r=-2
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x66(f), r=-2
[tempesta]   parser: Req_UriAbsPath(22:0): c=0x6f(o), r=-2
[...]

PR #268, opened 2015-09-15
#257. Fix chunk begins of split fields

avbelov23:


tempesta_fw/http_parser.c
```diff
@@ -101,9 +101,16 @@ st: __attribute__((unused)) 						\
 	TFW_DBG("parser: " #st "(%d:%d): c=%#x(%c), r=%d\n",		\
 		st, parser->_i_st, c, isprint(c) ? c : '.', r);
 
-#define __FSM_EXIT(field)						\
+#define __FSM_EXIT(p)							\
```
krizhanovsky: ```p``` is used as local variable in all FSM functions, so the change isn't good.

tempesta_fw/http_parser.c
```diff
@@ -101,9 +101,16 @@ st: __attribute__((unused)) 						\
...
 do {									\
-	if (field) { /* staticaly resolved */				\
+	if (p) { /* staticaly resolved */				\
+		TfwStr *field = (p);					\
```
krizhanovsky: Why do we need the temporary variable ```field```?

avbelov23: Sametimes as field we pass NULL (void*) and we can not do field->ptr.

avbelov23: Sometimes as ''field we pass NULL(void) and we can't do ''field->ptr

krizhanovsky:
As discussed ____FSM_MOVE_LAMBDA() calls tfw_http_msg_hdr_chunk_fixup() for uninitialized parser->hdr even for URI and Host parsing, which is surely wrong. The reason for the bug is that tfw_http_msg_set_data() (usually called from tfw_http_msg_hdr_chunk_fixup()) is not called for parser->uri_path. So the merge request is wrong and higher level logic must be fixed.

Issue #260, opened 2015-09-12
http_parser: crash when parsing splitted incorrectly duplicated headers


stefanstrogin:
1) Compile Tempesta with the patch set from https://github.com/natsys/tempesta/tree/ss-fuzzer
2) Comment all tests except `fuzzer` in TEST_SUITE(http_parser).
```
 TEST_SUITE(http_parser)
 {
        TEST_TEARDOWN(free_msgs);
 
+#if 0
        TEST_RUN(http_parser, parses_req_method);
        TEST_RUN(http_parser, parses_req_uri);
        TEST_RUN(http_parser, fills_hdr_tbl);
        TEST_RUN(http_parser, blocks_suspicious_x_forwarded_for_hdrs);
        TEST_RUN(http_parser, parses_connection_value);
+#endif
        TEST_RUN(http_parser, fuzzer);
 }
```
3)
```
--- a/tempesta_fw/t/unit/tfw_fuzzer.c
+++ b/tempesta_fw/t/unit/tfw_fuzzer.c
@@ -87,7 +87,7 @@ static struct {
        /* TRANSFER_ENCODING_NUM */
        {0, 5, "", NULL},
        /* DUPLICATES */
-       {0, 3, "", NULL},
+       {1, 3, "", NULL},
        /* BODY_SIZE */
        {0, 10, "", NULL}
 };
```
4) ./run_all_tests.sh
It should crash:
```
[tempesta]   open header at char [H], skb=ffff88002f6a3800
[tempesta]   parser: RGen_LWS(10000:37): c=0x20( ), r=-2
[tempesta]   parser: RGen_LWS(10000:37): c=0x31(1), r=-2
[tempesta]   enter FSM at state 37
[tempesta]   parser: Req_HdrHostV(37:0): c=0x31(1), r=-2
[tempesta]   Free msg: ffff88002ec7e040
[tempesta]   free skb ffff88002f6a3800: truesize=1344 sk=          (null), destructor=          (null) users=1 type=Unknown
tfw_test: split: head_len=46 [GET http://file.html HTTP/1.1
Connection: clo], tail_len=598 [se
Connection: upgrade
User-agent: Mozilla/5.0
User-agent: Wget/1.13.4 (linux-gnu)
Host: localhost
Host: 127.0.0.1
X-Forwarded-For: localhost
X-Forwarded-For: 127.0.0.1
Content-Type: text/html;charset=utf-8
Content-Type: image/jpeg
Content-Length: 0
[...]

PR #269, opened 2015-09-16
#260. Fix crash. Second argument of tfw_stricmpspn is TfwStr

avbelov23:


krizhanovsky:
Good catch good to merge ;)

PR #270, opened 2015-09-17
Fix the way reconnect works on failover.

keshonok:
A bug was introduced in one of previous commits. On connect error or
a disconnect the failover procedure takes place, and then the connection
is released. In the process of connection release a new connect attempt
is initiated immediately. When that fails as well, it leads to failover
procedure again, and the circle continues. As a result, new connect
attempts are repeated at the same (very short) intervals, which is not
the original intent. This fix restores the original intent. New connect
attempts are always started in deferred context. The first attempt is
started after a very short interval, and consecutive connect attempts
are progressively delayed.

### DIFF
--- a/tempesta_fw/sock_srv.c
+++ b/tempesta_fw/sock_srv.c
@@ -69,7 +69,7 @@
  */
 
 /** The wakeup interval between failed connection attempts. */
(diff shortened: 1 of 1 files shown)


krizhanovsky:
Good to merge

PR #272, opened 2015-09-17
Instead of BUG_ON(), let schedulers return NULL for connections

keshonok:


tempesta_fw/sched/tfw_sched_rr.c
```diff
@@ -69,11 +69,17 @@ tfw_sched_rr_get_srv_conn(TfwMsg *msg, TfwSrvGroup *sg)
 	conn_list = &srv_list->conn_lists[idx];
 
 	/* 2. Select a connection in the same round-robin manner. */
-	BUG_ON(!conn_list || !conn_list->conn_n);
+	if (unlikely(!conn_list || !conn_list->conn_n))
```
krizhanovsky: conn_list is never NULL, since we get an item of the array.

krizhanovsky:
Good to merge with removed conn_list check for NULL

Issue #275, opened 2015-09-18
Skip header fieild name


avbelov23:
``(echo -en "GET / HTTP/1.1\r\nCache-"; sleep 1; echo -en "Control: max-"; sleep 1; echo -en "age=0\r\n\r\n") | nc 127.0.0.1 80``

        [  406.420841] [tempesta]   parser: Req_Hdr(32:0): c=0x43(C), r=-2
        [  406.420841] [tempesta]   open header at char [C], skb=ffff88003970de88
        [  406.420842] [tempesta]   parser: Req_HdrC(38:0): c=0x61(a), r=-2
        [  406.420843] [tempesta]   parser: Req_HdrCa(39:0): c=0x63(c), r=-2
        [  406.420844] [tempesta]   parser: Req_HdrCac(40:0): c=0x68(h), r=-2
        [  406.420845] [tempesta]   parser: Req_HdrCach(41:0): c=0x65(e), r=-2
        [  406.420846] [tempesta]   parser: Req_HdrCache(42:0): c=0x2d(-), r=-2
        [  406.420862] [tempesta]   store header chunk len=22 data=ffff88003b9eb8d0 hdr=<0x0,0,ffff88003b9eb8e0>
        ...
        [  407.419146] [tempesta]   parse 13 client data bytes (Control: max-) on req=ffff88003aa90040
        [  407.419147] [tempesta]   enter FSM at state 43
        [  407.419149] [tempesta]   parser: Req_HdrCache_(43:0): c=0x43(C), r=-2
        [  407.419150] [tempesta]   parser: Req_HdrCache_C(44:0): c=0x6f(o), r=-2
        [  407.419159] [tempesta]   parser: Req_HdrCache_Co(45:0): c=0x6e(n), r=-2
        [  407.419160] [tempesta]   parser: Req_HdrCache_Con(46:0): c=0x74(t), r=-2
        [  407.419161] [tempesta]   parser: Req_HdrCache_Cont(47:0): c=0x72(r), r=-2
        [  407.419162] [tempesta]   parser: Req_HdrCache_Contr(48:0): c=0x6f(o), r=-2
        [  407.419163] [tempesta]   parser: Req_HdrCache_Contro(49:0): c=0x6c(l), r=-2
        [  407.419164] [tempesta]   parser: Req_HdrCache_Control(50:0): c=0x3a(:), r=-2
        [  407.419165] [tempesta]   parser: RGen_LWS(10000:51): c=0x20( ), r=-2
[...]

PR #276, opened 2015-09-19
#275. Fix loss a header name

avbelov23:


tempesta_fw/http_parser.c
```diff
@@ -1133,7 +1134,7 @@ __req_parse_cache_control(TfwHttpReq *req, unsigned char *data, size_t len)
 		if (c == '=')
 			__FSM_I_MOVE(Req_I_CC_Ext);
 		if (IN_ALPHABET(c, hdr_a))
-			__FSM_I_MOVE(Req_I_CC);
+			__FSM_I_MOVE_n(Req_I_CC, 0);
```
krizhanovsky: Since you don't move ```p```, the call is effectively equal to ```goto Req_I_CC```, so use __FSM_I_JMP()

Issue #188, opened 2015-08-18
TfwPool extension


krizhanovsky:
Currently TfwPool allocations are limited by single page in total, so chunks accounting for large HTTP messages is impossible. The pool must be extended to grow automatically as current page is exhausted.

krizhanovsky:
Dummy pool implementation w/o ```free``` interface leads to CPU data cache starvation: any new allocation works with new memory location, so just accessed cache lines must be evicted and the new memory block must be loaded to CPU cache.

Thus very simple freeing interface, which probably allows memory leakage (no worries - the whole pool is dropped at the end of HTTP message processing), can gain more performance.

PR #278, opened 2015-09-21
#188. TfwPool extension

krizhanovsky:


tempesta_fw/pool.h
```diff
@@ -28,9 +28,14 @@
 #define TFW_POOL_ZERO	0x1
 
 typedef struct {
-	unsigned char *base; /* base address of the pool */
-	unsigned int order; /* total size of the pool */
+	struct list_head list;
```
krizhanovsky: Do we really need double linked list? Probably we just should keep head and tail of the list... Double linked lists are too complex for fast synchronization (e.g. lock-free techniques).

tempesta_fw/pool.c
```diff
@@ -35,15 +36,22 @@
 TfwPool *
 __tfw_pool_new(size_t n)
 {
-	unsigned int order = (n + sizeof(TfwPool)) >> PAGE_SHIFT;
+	unsigned int order = (n + sizeof(TfwPool) + sizeof(TfwPoolChunk)) >> PAGE_SHIFT;
```
krizhanovsky: Why do we need to keep TfwPool in each pool chunk? If we really need to do this, then ____cacheline_aligned probably has no sense.

tempesta_fw/pool.c
```diff
@@ -99,6 +119,10 @@ EXPORT_SYMBOL(tfw_pool_realloc);
 void
 tfw_pool_free(TfwPool *p)
 {
-	free_pages((unsigned long)p, p->order);
+	TfwPoolChunk *chunk;
+
+	list_for_each_entry(chunk, &p->chunks, list) {
```
krizhanovsky: Use safe version of the function since you're deleting the list items

tempesta_fw/pool.c
```diff
@@ -70,21 +89,22 @@ EXPORT_SYMBOL(tfw_pool_alloc);
 void *
 tfw_pool_realloc(TfwPool *p, void *ptr, size_t old_n, size_t new_n)
 {
-	unsigned char *p_tmp = ptr;
+	unsigned char *tmp_ptr = ptr;
 	void *a;
+	TfwPoolChunk *chunk;
+
```
krizhanovsky: It's good to call here ```try_free()``` function (https://github.com/natsys/tempesta/issues/188#issuecomment-138092421): if ```ptr``` is the last allocated item, then we can reuse it's memory. So it's good to implement ```try_free``` at the same time.

tempesta_fw/pool.c
```diff
@@ -53,15 +61,26 @@ void *
 tfw_pool_alloc(TfwPool *p, size_t n)
 {
 	void *a;
+	TfwPoolChunk *chunk;
+
```
krizhanovsky: The pool is growing by pages, so we need BUG_ON or if statement if a user requests more than ```PAGE_SIZE - sizeof(blah-blah-blah)``` bytes.

tempesta_fw/pool.c
```diff
@@ -35,33 +36,65 @@
 TfwPool *
 __tfw_pool_new(size_t n)
 {
-	unsigned int order = (n + sizeof(TfwPool)) >> PAGE_SHIFT;
+	unsigned int order = get_order(n + sizeof(TfwPool) + sizeof(TfwPoolChunk));
```
krizhanovsky: too long line

tempesta_fw/pool.h
```diff
@@ -53,5 +61,6 @@ TfwPool *__tfw_pool_new(size_t n);
 void *tfw_pool_alloc(TfwPool *p, size_t n);
 void *tfw_pool_realloc(TfwPool *p, void *ptr, size_t old_n, size_t new_n);
 void tfw_pool_free(TfwPool *p);
+void tfw_pool_try_free(TfwPool *p, void *ptr, size_t n);
```
krizhanovsky: Would be good to rename ```tfw_pool_free``` to ```tfw_pool_destroy``` otherwise we get two very different functions with very similar names.

tempesta_fw/pool.h
```diff
@@ -27,11 +28,18 @@
...
-typedef struct {
-	unsigned char *base; /* base address of the pool */
-	unsigned int order; /* total size of the pool */
+typedef struct TfwPoolChunk TfwPoolChunk;
+
+struct TfwPoolChunk {
+	TfwPoolChunk *next;
+	unsigned char *base; /* base address of the pool chunk */
+	unsigned int order; /* total size of the pool chunk */
 	unsigned int off; /* current offset, INVAIRANT: off < size */
```
krizhanovsky: The comment is outdated: we have no field ```size``` in the data structure

tempesta_fw/pool.c
```diff
@@ -35,33 +36,65 @@
...
+	TfwPoolChunk *chunk;
+
+	chunk = p->head;
+	while (chunk != NULL) {
+		if (chunk->base <= tmp_ptr &&
+		    tmp_ptr < chunk->base + TFW_POOL_CHUNK_SIZE(chunk)) {
+			break;
+		}
+		chunk = chunk->next;
+	}
```
krizhanovsky: You don't need linear scan. The chunk can be retrieved simply by ```(unsigned long)ptr & PAGE_MASK)``` since TfwPoolChunk is always at head of the page where ptr is.

tempesta_fw/pool.c
```diff
@@ -99,6 +126,26 @@ EXPORT_SYMBOL(tfw_pool_realloc);
...
+	while (chunk != NULL) {
+		next = chunk->next;
+		free_pages((unsigned long)chunk, chunk->order);
+		chunk = next;
+	}
 }
 EXPORT_SYMBOL(tfw_pool_free);
+
+void
+tfw_pool_try_free(TfwPool *p, void *ptr, size_t n)
```
krizhanovsky: Please place the function just after tfw_pool_realloc() - it's good to keep logically near functions together

tempesta_fw/pool.c
```diff
@@ -99,6 +126,26 @@ EXPORT_SYMBOL(tfw_pool_realloc);
...
+{
+	TfwPoolChunk *chunk;
+
+	chunk = __tfw_find_chunk(p, ptr);
+	if (ptr + n == chunk->base + chunk->off) {
+		chunk->off -= n;
+		memset(ptr, 0, n);
```
krizhanovsky: Well, memset() and TFW_POOL_ZERO was a poor thing, and now it goes worse - ```tfw_pool_try_free()``` has sens only if it's extremely lightweight and fast, otherwise our pool goes to be common memory allocator. So we should not do the memory zeroing if we're not sure do we need this or not.

Please review _all_ memory allocation in the project. For example, tfw_http_msg_alloc() requests zeroed memory, by also calls memset() for just allocated memory.

It's good to create the pool w/o any flags, but extend tfw_pool_alloc() and tfw_pool_realloc() with the flag, so if the called needs zeroed memory, then it should request it only for the required memory chunk.

tempesta_fw/classifier/frang.c
```diff
@@ -479,8 +479,10 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
 	if (!tfw_addr_pton(buf, &addr)) {
 		frang_msg("Host header field contains IP address",
 			  &ra->addr, "\n");
+		tfw_pool_try_free(req->pool, buf, len);
 		return TFW_BLOCK;
 	}
+	tfw_pool_try_free(req->pool, buf, len);
 	return TFW_PASS;
 }
```
krizhanovsky: Frang is not the only place where try_free has sense. It seems tfw_http_sticky_get() also does memory allocation which can be freed when the function finishes. @keshonok, please comment.

tempesta_fw/pool.c
avbelov23: No. __get_free_pages may return several pages.

Issue #282, opened 2015-09-24
Data SKB with FIN flag is not processed


keshonok:
When data SKB with FIN flag comes, `sk_state_change()` is called first, before `sk_data_ready()` is called. Because of that, the socket is closed prematurely by Sync Sockets, and the data SKB with FIN flag is discarded. In the end, an HTTP message never completes and is discarded as well.

This is linked with commit 83af006

keshonok:
Fixed in a8f1af6.

keshonok:
The fix appears to be wrong as it leads to a possible deadlock. See the traces below.
```
[tempesta] Warning: frang: connections max num. exceeded for ::ffff:c0a8:a01: 201 (lim=200)
[tempesta] Warning: frang: connections max num. exceeded for ::ffff:c0a8:a01: 201 (lim=200)

=============================================
[ INFO: possible recursive locking detected ]
3.10.10+ #6 Tainted: G           O
---------------------------------------------
swapper/0/0 is trying to acquire lock:
 (slock-AF_INET/1){+.-...}, at: [<ffffffffa03e828d>] ss_send+0x5d/0x390 [tempesta_fw]

but task is already holding lock:
 (slock-AF_INET/1){+.-...}, at: [<ffffffff814a0ebd>] tcp_v4_rcv+0x80d/0xb50

other info that might help us debug this:
 Possible unsafe locking scenario:

       CPU0
       ----
  lock(slock-AF_INET/1);
  lock(slock-AF_INET/1);

 *** DEADLOCK ***

 May be due to missing lock nesting notation

4 locks held by swapper/0/0:
 #0:  (rcu_read_lock){.+.+..}, at: [<ffffffff8143e460>] __netif_receive_skb_core+0x110/0x900
 #1:  (rcu_read_lock){.+.+..}, at: [<ffffffff8147895d>] ip_local_deliver_finish+0x3d/0x340
 #2:  (slock-AF_INET/1){+.-...}, at: [<ffffffff814a0ebd>] tcp_v4_rcv+0x80d/0xb50
 #3:  (clock-AF_INET){++--..}, at: [<ffffffffa03e78e1>] ss_tcp_process_data+0x1b1/0x320 [tempesta_fw]

stack backtrace:
CPU: 0 PID: 0 Comm: swapper/0 Tainted: G           O 3.10.10+ #6
Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS Bochs 01/01/2011
[...]

PR #285, opened 2015-09-24
Process data SKB in ss_tcp_state_change() when FIN is received. (#282)

keshonok:
When data SKB is received with FIN flag, sk_state_change() handler
is called first by the kernel. The usual reaction to that is to drop
the connection, close the socket, and start the failover procedure.
However, data SKB has not been processed in that case which lead to
loss of an HTTP message that could not complete. This patch makes
sure the outstanding SKBs are processed before the connection is
dropped and failover procedure started.

### DIFF
--- a/tempesta_fw/sock.c
+++ b/tempesta_fw/sock.c
@@ -718,7 +718,6 @@ ss_tcp_data_ready(struct sock *sk, int bytes)
 			 * SKBs linked with the currently processed message.
 			 */
 			ss_droplink(sk);
-		}
 	}
 	else {
 		/*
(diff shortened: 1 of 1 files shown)


tempesta_fw/sock.c
```diff
@@ -469,7 +469,7 @@ ss_droplink(struct sock *sk)
  * TODO:
  * -- process URG
  */
-static void
+static int
```
krizhanovsky: ```droplink``` is bool, so it's good to return bool as well.

krizhanovsky:
Good to merge w/ small adjustment

PR #286, opened 2015-09-26
Use correct free() function when freeing a message.

keshonok:
In tfw_http_msg_create_sibling() a new message is created with
tfw_http_conn_msg_alloc(). The message is linked with an existing
connection instance, which increments the refcnt of the instance.
That means that freeing of the message must be done using correct
function that decrements the refcnt of the connection instance.


krizhanovsky:
Good catch, Alexey. Good to merge

PR #293, opened 2015-10-05
copy request body to modifiable memory in test_sched_http.c

i-rinat:
`tfw_http_parse_req()` can modify data in-place, therefore buffer needs to be writable.


krizhanovsky:
Good to merge

PR #296, opened 2015-10-06
Fix the insertion of an HTTP header field in ss_skb.c

keshonok:


krizhanovsky:
Good catch and good fix to be merged

PR #302, opened 2015-10-07
store chunk size in TfwPoolChunk at creation time

i-rinat:
`c->order` was kept uninitialized, and only filled at next segment allocation time.


krizhanovsky:
Good to merge

PR #310, opened 2015-10-12
Change sched.c to use RCU locks, and add thorough comments.

keshonok:


tempesta_fw/sched.c
```diff
@@ -25,87 +25,114 @@
...
  */
 TfwScheduler *
 tfw_sched_lookup(const char *name)
 {
 	TfwScheduler *sched;
 
-	read_lock_bh(&sched_lock);
-
-	list_for_each_entry(sched, &sched_list, list) {
+	rcu_read_lock();
```
krizhanovsky: Why not ```rcu_read_lock_bh()```? The function can be called in process context (configuration) and softirq.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -171,7 +171,7 @@ __calc_conn_hash(TfwServer *srv, size_t conn_idx)
 }
 
 static void
-tfw_sched_hash_update_data(TfwSrvGroup *sg)
+tfw_sched_hash_update_data(TfwSrvGroup *sg, TfwSrvConnection *srv_conn)
```
krizhanovsky: ```srv_conn``` is unused here

tempesta_fw/sched/tfw_sched_rr.c
```diff
@@ -33,20 +34,103 @@ MODULE_LICENSE("GPL");
 #define ERR(...) TFW_ERR(BANNER __VA_ARGS__)
```
krizhanovsky: You almost fully rewrote the scheduler code, please update minor version on such large changes. If the code gets large new functionality or redesign, then minor version is updated. If it gets significant bugfixes, then patch level version is updated.

tempesta_fw/sched/tfw_sched_rr.c
```diff
@@ -33,20 +34,103 @@ MODULE_LICENSE("GPL");
...
+ * @n_s_ent and @n_c_ent only increase and never decrease.
+ *
+ * @n_srv may increase or decrease. It represents the current size
+ * of @s_pentry array, which is the number of schedulable servers.
+ */
+typedef struct TfwRrSchedData {
+	struct rcu_head		rcu;
+	spinlock_t		slock;
+	atomic64_t		rr_cnt;
+	atomic_t		n_srv;
```
krizhanovsky: Why do we need it atomic if it's read under RCU and updated under slock?

tempesta_fw/sched/tfw_sched_rr.c
```diff
@@ -54,92 +138,353 @@ typedef struct {
...
+
+	sched_data = rcu_dereference(sg->sched_data);
+	BUG_ON(sched_data == NULL);
+
+	if (tfw_connection_live(&s_conn->conn))
+		tfw_sched_rr_go_live(sched_data, s_conn);
+	else
+		tfw_sched_rr_go_dead(sched_data, s_conn);
+
+	rcu_read_unlock();
```
krizhanovsky: Wow... The things going there look so complex.... I believe the redesign is required, basicaly RR is the most simple thing and should not be so complex.

PR #311, opened 2015-10-13
Let ss_tcp_data_ready() do FIN processing. (#282)

keshonok:


krizhanovsky:
Good to merge

keshonok:
Unfortunately, this still won't fix the problem. See explanation here: https://github.com/natsys/tempesta/issues/282#issuecomment-148369580.

Issue #299, opened 2015-10-07
Fixing TFW_LOG,ERR,WARN


vladtcvs:
TFW_ERR,LOG,WARN uses printk directly in DEBUG=1,2,3 case, and this is bad, and net_err_ratelimited without DEBUG. This leads to losing important messages, but we need to prevent message flood. We need to find some way to fix this.

krizhanovsky:
Basically you should not see problems with net_x_ratelimited() functions. If you do see, then the systems floods the log and the exact flooding messages must be fixed. If you need a lot of information to be printed, then use DEBUG=1 (user level debugging).

PR #313, opened 2015-10-14
Ls 299 02

sergsever:
Добавил в printk стандартные констаны KERN_ERR и т.д.


krizhanovsky:
The pull requests is obviously broken since it contains many other commits non relating to the #299.

PR #318, opened 2015-10-19
Fixed #199

### LINKED ISSUE #199: Sticky Cookies: poor User-Agent and Cookie header processing
Currently Cookie and User-Agent aren't special header, so the module uses a loop in ```tfw_http_field_raw()``` to find the headers. The headers must be made special to remove the loop and improve the module performance. It worth to mention that the headers are crucial for users identification and very useful in any HTTP processing.

Also ```tfw_http_field_value()``` uses deprecated ```tfw_str_to_cstr()``` and basically repeats functionality of new ```tfw_http_msg_hdr_val()```.

```tfw_str_to_cstr()``` should be finally removed when this issue and #192 are fixed.

### PR DESCRIPTION
**Pull request by almk277**
#199

### CODE REVIEW

File: tempesta_fw/http.h [hunk 1]
```diff
@@ -158,6 +158,8 @@ typedef enum {
 
 	TFW_HTTP_HDR_CONNECTION = TFW_HTTP_HDR_NONSINGULAR,
 	TFW_HTTP_HDR_X_FORWARDED_FOR,
+	TFW_HTTP_HDR_USER_AGENT,
+	TFW_HTTP_HDR_COOKIE,
```
keshonok: These two should be singular.

tempesta_fw/http_parser.c
```diff
@@ -70,7 +83,8 @@ int __fsm_const_state;							\
 int __fsm_n __attribute__((unused));					\
 size_t __fsm_sz __attribute__((unused));				\
 unsigned char *__fsm_ch __attribute__((unused));			\
-TfwStr *__fsm_str __attribute__((unused));
+TfwStr *__fsm_str __attribute__((unused));				\
```
keshonok: This variable is unused in the code now. Perhaps it's no harm to keep it anyway, but maybe it's better to remove it altogether.

tempesta_fw/http_sticky.c
```diff
@@ -316,9 +307,11 @@ tfw_http_sticky_found(TfwHttpMsg *hm, TfwStr *value)
 	/*
 	 * Do nothing for now. The request is passed to a backend server.
 	 */
-	/* XXX This assumes that 'value' is a linear TfwStr{}. */
-	TFW_DBG("Sticky cookie found: \"%.*s\"\n",
-		(int)value->len, (char *)value->ptr);
+	if (likely(TFW_STR_PLAIN(value)))
```
keshonok: The downside of this is that when debugging is compiled out that will leave an empty `if()` block. Perhaps it's better to print full or partial value depending on whether the string is compound or not, and do all of that within `TFW_DBG()` statement.

tempesta_fw/http_sticky.c
```diff
@@ -78,67 +79,70 @@ tfw_http_sticky_send_302(TfwHttpMsg *hm)
...
+	} state = StateName;
+
+	BUG_ON(!TFW_STR_PLAIN(str));
+
+	TFW_STR_FOR_EACH_CHUNK(chunk, cookie, {
+		switch (state) {
+
+		case StateName:
+			if ((chunk->flags & TFW_STR_NAME)
+			    && tfw_str_eq_cstr(chunk, str->ptr, str->len,
```
keshonok: Let's say that data come in one byte at a time, so a cookie name will be scattered across multiple chunks. In that case this will not work as intended.

tempesta_fw/http_sticky.c
almk277: Things gets complicated. Maybe, it's better to have general TFW_STR_DBG macro.
Also, an empty `if()` just optimizes out, it should not harm.

tempesta_fw/http.h
```diff
@@ -152,6 +152,8 @@ typedef enum {
 	TFW_HTTP_HDR_HOST,
 	TFW_HTTP_HDR_CONTENT_LENGTH,
 	TFW_HTTP_HDR_CONTENT_TYPE,
+	TFW_HTTP_HDR_USER_AGENT,
+	TFW_HTTP_HDR_COOKIE,
```
krizhanovsky: Please add to the comment above that Cookie is singular header due to RFC 6265 5.4. Also please find and document the requirement for User-Agent.

tempesta_fw/http_parser.c
```diff
@@ -70,7 +83,7 @@ int __fsm_const_state;							\
 int __fsm_n __attribute__((unused));					\
 size_t __fsm_sz __attribute__((unused));				\
 unsigned char *__fsm_ch __attribute__((unused));			\
-TfwStr *__fsm_str __attribute__((unused));
+unsigned int __fsm_chunkflags __attribute__((unused)) = 0;
```
krizhanovsky: The variable is used in interior FSM, so additional interior-specific macro should be added to define variables for interior FSM and do not increase stack size for main FSM.

tempesta_fw/http_sticky.c
```diff
@@ -78,67 +78,87 @@ tfw_http_sticky_send_302(TfwHttpMsg *hm)
...
+
+	BUG_ON(!TFW_STR_PLAIN(&tfw_cfg_sticky.name));
+
+	TFW_STR_INIT(&tmp);
+	tmp.flags = __TFW_STR_COMPOUND;
+
+	TFW_STR_FOR_EACH_CHUNK(chunk, cookie, {
+		--n;
+		switch (state) {
+
```
krizhanovsky: We don't place empty line between ```switch``` and ```case```

tempesta_fw/http_sticky.c
```diff
@@ -78,67 +78,87 @@ tfw_http_sticky_send_302(TfwHttpMsg *hm)
...
+			if (chunk->flags & TFW_STR_NAME) {
+				/*
+				 * Create temporary compound string, starting
+				 * with this chunk.
+				 * We do not use it's overall length now,
+				 * so do not set it.
+				 */
+				tmp.ptr = (void*)chunk;
+				__TFW_STR_CHUNKN_SET(&tmp, n);
+				if (tfw_str_eq_cstr(&tmp, cstr, clen,
```
krizhanovsky: There is no sense to call relatively expensive ```tfw_str_eq_cstr()```: ```cstr``` is plain C-string and ```chunk``` is also a plain string (we don't support chunked chunks).

tempesta_fw/http_sticky.c
```diff
@@ -78,67 +78,87 @@ tfw_http_sticky_send_302(TfwHttpMsg *hm)
...
+		StateVal,
+		StateValChunks,
+	} state = StateName;
+
+	BUG_ON(!TFW_STR_PLAIN(&tfw_cfg_sticky.name));
+
+	TFW_STR_INIT(&tmp);
+	tmp.flags = __TFW_STR_COMPOUND;
+
+	TFW_STR_FOR_EACH_CHUNK(chunk, cookie, {
```
krizhanovsky: Cookies can be large and this piece of code could be really hot (this is the intention of #199).

Never use the dummy FSM style in performance critical code, please read http://natsys-lab.blogspot.ru/2014/11/the-fast-finite-state-machine-for-http.html for more details.

Moreover, TfwStr containing Cookie is never plain so you don't need generic TfwStr loop provided by TFW_STR_FOR_EACH_CHUNK() and can use very simple common for() loop.

tempesta_fw/http_parser.c
```diff
@@ -383,7 +424,7 @@ enum {
  * (e.g. Content-Length which is doubled by TfwHttpMsg.conent_length)
  * to mangle row skb data.
  */
-#define TFW_HTTP_PARSE_SPECHDR_VAL(st_curr, st_i, hm, func, id)		\
+#define TFW_HTTP_PARSE_SPECHDR_VAL_GEN(st_curr, st_i, hm, func, id, saveval) \
```
krizhanovsky: "GEN" is usually used for "generate", I'd prefer "__TFW_HTTP_PARSE_SPECHDR_VAL" for the new macro

tempesta_fw/http_parser.c
```diff
@@ -458,7 +503,7 @@ __FSM_STATE(st_curr) {							\
 
 /*
  * Parse raw (common) HTTP headers.
- * Note that some of these (like Cookie or User-Agent) can be extremely large.
+ * Note that some of these can be extremely large.
```
krizhanovsky: Why do you remove the comment?

tempesta_fw/http_parser.c
```diff
@@ -1237,6 +1306,95 @@ __req_parse_x_forwarded_for(TfwHttpMsg *msg, unsigned char *data, size_t len)
...
+	unsigned char *p = data;
+	unsigned char *orig_data = data;
+	unsigned char c = *p;
+	__FSM_DECLARE_VARS();
+
+	/*
+	 * Here we build header value string manually
+	 * to split it in chunks: chunk bounds are
+	 * at least at name start, value start and value end.
+	 * This simplifies cookie search, http_sticky uses it.
```
krizhanovsky: Please add a comment with particular RFC chapter which describes the parsing rules.

tempesta_fw/http_parser.c
```diff
@@ -1490,6 +1648,17 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 				__FSM_MOVE_n(RGen_LWS, 16);
 			}
 			__FSM_MOVE(Req_HdrX);
+		case 'u':
+			if (likely(p + 10 <= data + len
+				&& *(p + 1) == 's'
+				&& *(p + 1) == 'e'
+				&& C8_INT_LCM(p + 3, 'r', '-', 'a', 'g',
```
krizhanovsky: You need explicitly check '-' because ```C8_INT_LCM()``` can convert '\r' to '-'

tempesta_fw/http_sticky.c
almk277: But name could be spread over multiple chunks, so we can not just compare a chunk - we should look at chunk sequence at `tmp`.

tempesta_fw/http_parser.c
almk277: The comment is about raw headers, but Cookie and User-Agent are special now.

tempesta_fw/http_parser.c
```diff
@@ -1237,6 +1306,95 @@ __req_parse_x_forwarded_for(TfwHttpMsg *msg, unsigned char *data, size_t len)
...
+		__FSM_I_MOVE_flags(Req_I_CookieVal, TFW_STR_VALUE);
+	}
+
+	__FSM_STATE(Req_I_CookieVal) {
+		if (unlikely(c == ';'))
+			/* do not save ';' yet */
+			__FSM_I_MOVE_chunk_n(Req_I_CookieSP, -1);
+		if (unlikely(c == '\r' || c == ' ')) {
+			/* do not save LWS */
+			__FSM_I_CHUNK_FIXUP(-1);
```
krizhanovsky: Please rework the macro, so it'll accept 0 instead of -1 (see the chat discussion).

tempesta_fw/http_parser.c
```diff
@@ -126,14 +136,58 @@ do {									\
...
  * FSMs, and so _I_ in the name means "interior" FSM.
  */
+
 #define __FSM_I_EXIT()			goto done
-#define __FSM_I_MOVE_n(to, n)						\
+#define __FSM_I_CHUNK_FIXUP(n)						\
+do {									\
+	/* Save symbols until current, plus n symbols more */	\
+	int slen = p - data + n;					\
+	__chunk_finish(msg, data, slen, &__fsm_chunkflags);		\
```
krizhanovsky: ```__fsm_chunkflags``` is defined in only one place, so the macros using the variable are also used in only one place, so it's better to remove them and do the code w/o macros.

krizhanovsky:
Good to merge with the small cleanups on Cookie parsing specific macros.

Issue #326, opened 2015-10-22
Not filled req->crlf


vladtcvs:
At the begining of `frang_http_req_process()` add:
```
TFW_DBG("%s: =========> crlf->ptr = %p, crlf->len = %d, crlf->flags = %x\n",
        __func__, req->crlf.ptr, req->crlf.len, req->crlf.flags);
```

On the host that runs Tempesta do:
```
telnet 127.0.0.1 80<enter>
GET http://localhost/ HTTP/1.1<enter>
Host: localhost<enter>
<enter>
```

The result is:
```
[tempesta]   frang_http_req_process: =========> crlf->ptr =           (null), crlf->len = 0, crlf->flags = 0
[tempesta]   frang_http_req_process: =========> crlf->ptr =           (null), crlf->len = 0, crlf->flags = 0
[tempesta]   frang_http_req_process: =========> crlf->ptr = ffff8801172d48d0, crlf->len = 0, crlf->flags = 0
```

So, crlf->ptr gets assigned to, but length remains zero.
Below is the full trace of the request/response.

```
[sync_sockets]   ss_tcp_data_ready: sk ffff880116d7f4c0, sk->sk_socket           (null), state (Established)
[tempesta]   Link new msg ffff880105b9f020 with connection ffff880105bf3a70
[tempesta]   Add skb ffff880116faae88 to message ffff880105b9f020
[tempesta]   GFSM exec fsm 0, state 0x2000
[tempesta]     Received 32 client data bytes on conn=ffff880105bf3a70 msg=ffff880105b9f020
[tempesta]   parse 32 client data bytes (GET http://localhost/ HTTP/1.1
) on req=ffff880105b9f020
[tempesta]     Request parsed: len=32 parsed=32 msg_len=32 res=-1
[tempesta]   GFSM move from 0x2000 to 0x1
[tempesta]   GFSM switch from fsm 0 at state 1 to fsm 2 at state 0x400
[tempesta]   GFSM exec fsm 2, state 0x2400
[tempesta]   frang_http_req_process: =========> crlf->ptr =           (null), crlf->len = 0, crlf->flags = 0
[sync_sockets]   ss_tcp_data_ready: sk ffff880116d7f4c0, sk->sk_socket           (null), state (Established)
[tempesta]   Add skb ffff880116faac48 to message ffff880105b9f020
[tempesta]   GFSM exec fsm 0, state 0x2001
[...]

Issue #151, opened 2015-07-12
Frang: RFC 7230 5.4 Host validation


krizhanovsky:
RFC 7230 5.4 says:

        A client MUST send a Host header field in all HTTP/1.1 request
        messages.  If the target URI includes an authority component, then a
        client MUST send a field-value for Host that is identical to that
        authority component, excluding any userinfo subcomponent and its "@"
        delimiter (Section 2.7.1).  If the authority component is missing or
        undefined for the target URI, then a client MUST send a Host header
        field with an empty field-value.

So Frang (```frang_http_host_check()```) must compare values of Host header with req->host and properly handle empty Host header values.

krizhanovsky:
Also HTTP parser must support [Punycode](https://en.wikipedia.org/wiki/Punycode)

vladtcvs:
punycode support added

vladtcvs:
depends on #326

vladtcvs:
fix merged

PR #327, opened 2015-10-22
fill req->crlf.len: #326

vladtcvs:
filling crlf.len - bug #326

### DIFF
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -58,6 +58,16 @@ __field_finish(TfwHttpMsg *hm, TfwStr *field,
 	field->flags |= TFW_STR_COMPLETE;
 }
 
+/* Same as __field_finish, but p points on the last character of the string */
+static inline void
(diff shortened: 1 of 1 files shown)


tempesta_fw/http_parser.c
```diff
@@ -1592,8 +1602,10 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 
 	/* Request headers are fully read. */
 	__FSM_STATE(Req_HdrDone) {
-		if (c == '\n')
+		if (c == '\n') {
+			__field_finish_curpos(msg, &req->crlf, data, p);
```
krizhanovsky: There is no sense to introduce the new function of 2 lines and called only once.
Please see comment in #326 - probably we should set crlf.len by 2, however setting TFW_STR_COMPLETE is appreciated as well...

vladtcvs:
patch is included into patch for #151 - Pull request #329

PR #329, opened 2015-10-26
Item #151, #326 fixing

vladtcvs:
check if Host: == host in URI

Handle empty Host: header

remove unused vars

fix variables names

use hardware crc32 only if supported

### DIFF
--- a/tempesta_fw/classifier/frang.c
+++ b/tempesta_fw/classifier/frang.c
@@ -464,28 +464,79 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
 {
 	TfwAddr addr;
 	TfwStr field;
+	int ret = TFW_BLOCK;
+	char *hdrhost = NULL;
+
+	/* 1) we check, if Host: header is present.
+	 * 2) we check, that Host: header is not a IP address.
+	 * 3) if URI has form "http://host:port/path", we check if host in URI
+	 * 	is equal to Host: header. As a particular case, if host
+	 * 	in uri is empty, Host: header also should be empty
+	 * 4) if URI has form "/path", we check, that host is not empty
+	 *
+	 * (RFC 7230 5.4):
+	 * 	A client MUST send a Host header field in all HTTP/1.1 request
+	 * 	messages.  If the target URI includes an authority component,
+	 * 	then a client MUST send a field-value for Host that is
+	 * 	identical to that authority component, excluding any userinfo
+	 * 	subcomponent and its "@" delimiter (Section 2.7.1).
+	 * 	If the authority component is missing or undefined for the
+	 * 	target URI, then a client MUST send a Host header field
+	 * 	with an empty field-value. */
+
+	TFW_DBG2("Start http host check\n");
+
+	BUG_ON(req == NULL);
+	BUG_ON(req->h_tbl == NULL);
 
 	if (TFW_STR_EMPTY(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST])) {
+		TFW_DBG3("Host header is missed\n");
 		frang_msg("Host header field", &ra->addr, " is missed\n");
 		return TFW_BLOCK;
 	}
-
 	tfw_http_msg_hdr_val(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST],
 			     TFW_HTTP_HDR_HOST, &field);
 
-	/*
-	 * FIXME: here should be a check that the Host value is not an IP
-	 * address. Need a fast routine that supports compound TfwStr.
-	 * Perhaps should implement a tiny FSM or postpone the task until we
-	 * have a good regex library.
-	 * For now just linearize the Host header field TfwStr{} string.
-	 */
-	if (!tfw_addr_pton(&field, &addr)) {
-		frang_msg("Host header field contains IP address",
-			  &ra->addr, "\n");
-		return TFW_BLOCK;
+	if (!TFW_STR_EMPTY(&field) && field.len > 0) {
+		if (!tfw_addr_pton(&field, &addr)) {
+			frang_msg("Host header field contains IP address",
+				&ra->addr, "\n");
+			goto finish;
+		}
 	}
-	return TFW_PASS;
+
+	if (req->flags & TFW_HTTP_URI_FULL) {
+		if (!TFW_STR_EMPTY(&field)) {
+			if ((hdrhost = tfw_pool_alloc(req->pool,
+				field.len + 1)) == NULL)
+			{
+				TFW_ERR("Can not allocate memory\n");
+				goto finish;
+			}
+			tfw_str_to_cstr(&field, hdrhost, field.len + 1);
+		}
+
+		/* Host: header must be equal to host in URI */
+		if (!tfw_str_eq_cstr(&req->host, hdrhost,
+					 field.len, TFW_STR_EQ_CASEI))
+		{
+			frang_msg("Host header is not equal to host in URL",
+				  &ra->addr, "\n");
+			goto finish;
+		}
+	} else {
+		TFW_DBG3("No host in URI\n");
+		if (TFW_STR_EMPTY(&field) || field.len == 0) {
+			frang_msg("Host header is empty", &ra->addr, "\n");
+			goto finish;
+		}
+	}
+	ret = TFW_PASS;
+finish:
+	if (hdrhost)
+		tfw_pool_free(req->pool, hdrhost, field.len + 1);
+
+	return ret;
 }
 
 /*
@@ -499,7 +550,7 @@ enum {
 };
 
 enum {
-	Frang_Req_0,
+	Frang_Req_0 = 0,
 
 	Frang_Req_Hdr_Start,
 	Frang_Req_Hdr_Method,
@@ -532,16 +583,52 @@ int __fsm_const_state = Frang_Req_0; /* make compiler happy */
 #define __FRANG_FSM_START(st)						\
 switch(st)
 
+#if defined(DEBUG) && (DEBUG >= 3)
+const char *__state_name_array[] = {
+	"Frang_Req_0",
+
+	"Frang_Req_Hdr_Start",
+	"Frang_Req_Hdr_Method",
+	"Frang_Req_Hdr_UriLen",
+	"Frang_Req_Hdr_FieldDup",
+	"Frang_Req_Hdr_FieldLen",
+	"Frang_Req_Hdr_FieldLenFinal",
+	"Frang_Req_Hdr_Crlf",
+	"Frang_Req_Hdr_Host",
+	"Frang_Req_Hdr_ContentType",
+
+	"Frang_Req_Hdr_NoState",
+
+	"Frang_Req_Body_Start",
+	"Frang_Req_Body_Timeout",
+	"Frang_Req_Body_ChunkCnt",
+	"Frang_Req_Body_Len",
+
+	"Frang_Req_Body_NoState",
+
+	"Frang_Req_Done"
+};
+
+#define __state_name(state) ((state >= 0 && state <= Frang_Req_Done) ?	\
+				__state_name_array[state] :		\
+				"Wrong state")
+#endif /* defined(DEBUG) && (DEBUG >= 3) */
+
+/* NOTE: we use the fact, that if DEBUG < 3, TFW_DBG3() is empty, so
+ * we can use it with undefined arguments, such as
+ * __state_name(__fsm_const_state), which is defined only when DEBUG >= 3
+ */
 #define __FRANG_FSM_FINISH()						\
 done:									\
-	TFW_DBG3("Finish FRANG FSM at state %d\n", __fsm_const_state);	\
+	TFW_DBG3("Finish FRANG FSM at state %d = %s\n",			\
+		__fsm_const_state, __state_name(__fsm_const_state));	\
 	TFW_DBG3("Frang return %s\n", r == TFW_PASS ? "PASS" : "BLOCK");\
 	req->frang_st = __fsm_const_state;
 
 #define __FRANG_FSM_STATE(st)						\
 case st:								\
 st: __attribute__((unused))						\
-	TFW_DBG3("enter FRANG FSM at state %d\n", st);			\
+	TFW_DBG3("enter FRANG FSM at state %d = %s\n", st, __state_name(st));\
 	__fsm_const_state = st; /* optimized out to constant */
 
 #define __FRANG_FSM_EXIT()	goto done;
@@ -689,7 +776,7 @@ frang_http_req_process(FrangAcc *ra, TfwConnection *conn, struct sk_buff *skb,
 	 * If not, continue checks on header fields.
 	 */
 	__FRANG_FSM_STATE(Frang_Req_Hdr_Crlf) {
-		if (!TFW_STR_EMPTY(&req->crlf))
+		if (req->crlf.flags & TFW_STR_COMPLETE)
 			__FRANG_FSM_JUMP(Frang_Req_Hdr_FieldLenFinal);
 		__FRANG_FSM_JUMP_EXIT(Frang_Req_Hdr_FieldDup);
 	}
--- a/tempesta_fw/http.h
+++ b/tempesta_fw/http.h
@@ -77,13 +77,6 @@ enum {
 	TFW_HTTP_FSM_DONE	= TFW_GFSM_HTTP_STATE(TFW_GFSM_STATE_LAST)
 };
 
-/**
- * All helping information for current HTTP parsing state of a message.
- */
-#define TFW_HTTP_PF_CR			0x01
-#define TFW_HTTP_PF_LF			0x02
-#define TFW_HTTP_PF_CRLF		(TFW_HTTP_PF_CR | TFW_HTTP_PF_LF)
-
 typedef enum {
 	TFW_HTTP_METH_NONE,
 	TFW_HTTP_METH_GET,
@@ -119,22 +112,24 @@ typedef struct {
  * (e.g. process LWS using @state while current state is saved in @_i_st
  * or using @_i_st parse value of a header described.
  *
- * @state	- current parser state;
- * @_i_st	- helping (interior) state;
  * @to_go	- remaining number of bytes to process in the data chunk;
  *		  (limited by single packet size and never exceeds 64KB)
+ * @state	- current parser state;
+ * @_i_st	- helping (interior) state;
  * @to_read	- remaining number of bytes to read;
+ * @_hdr_tag	- describes, which header should be closed in case of
+ *		  the empty header (see RGEN_LWS_empty)
  * @_tmp_acc	- integer accumulator for parsing chunked integers;
  * @_tmp_chunk	- currently parsed (sub)string, possibly chunked;
  * @hdr		- currently parsed header.
  */
 typedef struct tfw_http_parser {
-	unsigned char	flags;
 	unsigned short	to_go;
 	int		state;
 	int		_i_st;
 	int		to_read;
 	unsigned long	_tmp_acc;
+	unsigned int	_hdr_tag;
 	TfwStr		_tmp_chunk;
 	TfwStr		hdr;
 } TfwHttpParser;
@@ -186,6 +181,8 @@ typedef struct {
 /* Request flags */
 #define TFW_HTTP_STICKY_SET		0x0100	/* Need 'Set-Cookie` */
 #define TFW_HTTP_FIELD_DUPENTRY		0x0200	/* Duplicate field */
+/* URI has form http://authority/path, not just /path */
+#define TFW_HTTP_URI_FULL		0x0400
 
 /**
  * Common HTTP message members.
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -32,7 +32,7 @@
  * value.
  */
 void
-tfw_http_msg_hdr_val(TfwStr *hdr, int id, TfwStr *val)
+tfw_http_msg_hdr_val(TfwStr *hdr, unsigned id, TfwStr *val)
 {
 	static const size_t hdr_lens[] = {
 		[TFW_HTTP_HDR_HOST]	= sizeof("Host:") - 1,
@@ -45,17 +45,23 @@ tfw_http_msg_hdr_val(TfwStr *hdr, int id, TfwStr *val)
 	TfwStr *c, *end;
 	int nlen = hdr_lens[id];
 
-	if (unlikely(TFW_STR_EMPTY(hdr))) {
-		TFW_STR_INIT(val);
-		return;
-	}
-	BUG_ON(TFW_STR_PLAIN(hdr));
 	BUG_ON(TFW_STR_DUP(hdr));
-	BUG_ON(nlen >= hdr->len);
 	BUG_ON(id >= TFW_HTTP_HDR_RAW);
+
+	/* Only Host: header is allowed to be empty
+	 * If header string is plain, it is always empty header.
+	 * Not empty headers are compount strings. */
+	BUG_ON(id == TFW_HTTP_HDR_HOST ? (nlen > hdr->len) :
+		(nlen >= hdr->len || TFW_STR_PLAIN(hdr)));
 
 	*val = *hdr;
 
+	/* Field value, if it exist, lies in the separate chunk.
+	 * So we skip several first chunks, containing field name,
+	 * to get the field value. If we have field with empty value,
+	 * we get an empty string with val->len = 0 and val->ptr from the
+	 * last name's chunk, but it is unimportant.
+	 */
 	TFW_STR_FOR_EACH_CHUNK(c, hdr, end) {
 		BUG_ON(!c->len);
 
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -45,6 +45,7 @@
 enum {
 	RGen_LWS = 10000,
 	RGen_LF,
+	RGen_LWS_empty,
 };
 
 /**
@@ -104,7 +105,7 @@ do {									\
 	p += n;								\
 	if (unlikely(p >= data + len || !*p)) {				\
 		r = TFW_POSTPONE; /* postpone to more data available */	\
-		__fsm_const_state = to; /* start from state @to nest time */\
+		__fsm_const_state = to; /* start from state @to next time */\
 		/* Close currently parsed field chunk. */		\
 		tfw_http_msg_field_chunk_fixup(msg, field, data, len);	\
 		__FSM_EXIT()						\
@@ -606,35 +607,53 @@ __FSM_STATE(prefix ## _Done) {						\
 	return TFW_BLOCK;						\
 }
 
-/*
+#define RGEN_LWS_common_cases(st)					\
+	case ' ':							\
+	case '\t':							\
+		__FSM_MOVE(st);						\
+	default:							\
+		parser->state = parser->_i_st;				\
+		parser->_i_st = 0;					\
+		BUG_ON(unlikely(p >= data + len || !*p));		\
+		goto fsm_reenter;
+
+/* In request we should pass empty headers:
+ * RFC 7230 5.4:
+ * ....
+ * ....
+ * If the authority component is missing or
+ * undefined for the target URI, then a client MUST send a Host header
+ * field with an empty field-value.
+ *
+ * NOTE: using of RGEN_LWS_empty should be matched with
+ * the BUG_ON() statements in tfw_http_msg_hdr_val function
+ *
  * Read LWS at arbitrary position and move to stashed state.
  * This is bit complicated (however you can think about this as
  * a plain pushdown automaton), but reduces FSM code size.
  */
+#define RGEN_LWS_empty()						\
+__FSM_STATE(RGen_LWS_empty) {						\
+	switch (c) {							\
+	case '\r':							\
+	case '\n':							\
+		tfw_http_msg_hdr_chunk_fixup(msg, data, p - data);      \
+		if (tfw_http_msg_hdr_close(msg, parser->_hdr_tag))      \
+			return TFW_BLOCK;				\
+		if (c == '\r')						\
+			__FSM_MOVE(RGen_LF);				\
+		__FSM_JMP(RGen_LF);					\
+	RGEN_LWS_common_cases(RGen_LWS_empty)				\
+	}								\
+}
+
 #define RGEN_LWS()							\
 __FSM_STATE(RGen_LWS) {							\
 	switch (c) {							\
-	case '\r':							\
-		if (likely(!(parser->flags & TFW_HTTP_PF_CRLF))) {	\
-			parser->flags |= TFW_HTTP_PF_CR;		\
-			__FSM_MOVE(RGen_LWS);				\
-		}							\
-		return TFW_BLOCK;					\
 	case '\n':							\
-		if (likely(!(parser->flags & TFW_HTTP_PF_LF))) {	\
-			parser->flags |= TFW_HTTP_PF_LF;		\
-			__FSM_MOVE(RGen_LWS);				\
-		}							\
+	case '\r':							\
 		return TFW_BLOCK;					\
-	case ' ':							\
-	case '\t':							\
-		__FSM_MOVE(RGen_LWS);					\
-	default:							\
-		parser->flags &= ~TFW_HTTP_PF_CRLF;			\
-		parser->state = parser->_i_st;				\
-		parser->_i_st = 0;					\
-		BUG_ON(unlikely(p >= data + len || !*p));		\
-		goto fsm_reenter;					\
+	RGEN_LWS_common_cases(RGen_LWS)					\
 	}								\
 }
 
@@ -1343,10 +1362,16 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 	 * req->userinfo, reset req->host and fill it.
 	 */
 	__FSM_STATE(Req_UriAuthorityStart) {
+		req->flags |= TFW_HTTP_URI_FULL;
 		if (likely(isalnum(c) || c == '.' || c == '-')) {
 			*p = LC(*p);
 			tfw_http_msg_set_data(msg, &req->host, p);
 			__FSM_MOVE_f(Req_UriAuthority, &req->host);
+		} else if (likely(c == '/')) {
+			TFW_DBG3("Handling http:///path\n");
+			tfw_http_msg_set_data(msg, &req->host, p);
+			__field_finish(msg, &req->host, data, p);
+			__FSM_MOVE_f(Req_UriAbsPath, &req->uri_path);
 		}
 		return TFW_BLOCK;
 	}
@@ -1480,8 +1505,9 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 			if (!(req->body.flags & TFW_STR_COMPLETE)) {
 				tfw_http_msg_set_data(msg, &req->crlf, p);
 				__FSM_MOVE(Req_HdrDone);
-			} else
+			} else {
 				__FSM_MOVE(Req_Done);
+			}
 		}
 		if (unlikely(c == '\n')) {
 			if (!(req->body.flags & TFW_STR_COMPLETE)) {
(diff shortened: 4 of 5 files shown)


tempesta_fw/http.h
```diff
@@ -232,6 +232,7 @@ typedef struct {
 	TFW_HTTP_MSG_COMMON;
 	TfwStr			host;
 	TfwStr			uri_path;
+	int			uri_full;
```
keshonok: I think it's better to introduce a flag, something like `TFW_HTTP_HAS_URI` or similar, instead of extending the request structure.

tempesta_fw/http_parser.c
```diff
@@ -1302,6 +1270,7 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 
 	__FSM_STATE(Req_Uri) {
 		if (likely(c == '/')) {
+			req->uri_full = 0;
```
keshonok: I believe this is unnecessary.

tempesta_fw/classifier/frang.c
```diff
@@ -531,6 +581,53 @@ enum {
 	Frang_Req_Done
 };
 
+const char *
+state_name(unsigned state)
```
keshonok: * It's used only in debug printouts, so it's better to compile it conditionally as well, at the same debug level it's used at.
* Perhaps it's better to distinguish it from regular functions somehow, like `__state_name()` or similar.
* Perhaps it's better to use a static array of strings instead of a run-time function.

tempesta_fw/classifier/frang.c
```diff
@@ -531,6 +581,53 @@ enum {
...
+	case Frang_Req_Body_NoState:
+		return "Frang_Req_Body_NoState";
+
+	case Frang_Req_Done:
+		return "Frang_Req_Done";
+	}
+	return "Wrong state";
+}
+
+
```
keshonok: Extra blank line.

tempesta_fw/http_parser.c
```diff
@@ -1436,10 +1408,12 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 	__FSM_STATE(Req_Hdr) {
 		if (unlikely(c == '\r')) {
 			if (!(req->body.flags & TFW_STR_COMPLETE)) {
+				TFW_LOG("CRLF start\n");
```
keshonok: Do we need to log this every time an HTTP message is processed?

tempesta_fw/http_parser.c
```diff
@@ -1592,8 +1601,11 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 
 	/* Request headers are fully read. */
 	__FSM_STATE(Req_HdrDone) {
-		if (c == '\n')
+		if (c == '\n') {
+			TFW_LOG("CRLF end\n");
```
keshonok: Do we need to log this every time an HTTP message is processed?

tempesta_fw/http_parser.c
```diff
@@ -1325,6 +1296,7 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 	 * to lower case.
 	 */
 	__FSM_STATE(Req_UriHostStart) {
+		req->uri_full = 1;
```
keshonok: See a comment above regarding the use of a flag instead.

tempesta_fw/http_parser.c
```diff
@@ -1495,7 +1470,41 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 		}
 	}
 
-	RGEN_LWS();
+	__FSM_STATE(RGen_LWS) {
```
keshonok: What was the reason that required the removal of `RGen_LWS()` macro, and moving its whole contents to the body of this function, as well as to the body of `tfw_http_parse_resp()`?

tempesta_fw/http_parser.c
```diff
@@ -1592,8 +1601,11 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 
 	/* Request headers are fully read. */
 	__FSM_STATE(Req_HdrDone) {
-		if (c == '\n')
+		if (c == '\n') {
+			TFW_LOG("CRLF end\n");
+			__field_finish(msg, &req->crlf, data, p + 1);
```
keshonok: The fact that we don't set the length of `req->crlf` string until it's complete is somewhat cumbersome to remember and rely on. For that reason I believe that using `TFW_STR_EMPTY()` on `req->crlf` is cumbersome as well. Now that we set `req->crlf` correctly, I believe the correct way of checking for it in `Frang` is checking for `TFW_STR_COMPLETE` flag. What do you think?

tempesta_fw/classifier/frang.c
```diff
@@ -542,14 +639,15 @@ switch(st)
...
 	req->frang_st = __fsm_const_state;
 
-#define __FRANG_FSM_STATE(st)						\
-case st:								\
-st: __attribute__((unused))						\
-	TFW_DBG3("enter FRANG FSM at state %d\n", st);			\
+#define __FRANG_FSM_STATE(st)							\
+case st:									\
+st: __attribute__((unused))							\
+	TFW_DBG3("enter FRANG FSM at state %d = %s\n", st, state_name(st));	\
```
keshonok: Line lengths exceed 80 characters.

tempesta_fw/classifier/frang.c
```diff
@@ -482,18 +484,66 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+			ret = TFW_BLOCK;
+			goto finish;
+		}
+	} else {
+		TFW_DBG("No host in url\n");
+	}
+
+finish:
+	if (hdrhost)
+		tfw_pool_free(req->pool, hdrhost, hdrhostlen + 1/* + hdrhostulen_buf*/);
```
keshonok: Line length exceeds 80 characters.

tempesta_fw/classifier/frang.c
```diff
@@ -482,18 +484,66 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+		frang_msg("Host field not equal to host in URI",
+			  &ra->addr, "\n");
+		ret = TFW_BLOCK;
+		goto finish;
+	}
+
+	if (urlhostlen) {
+		if (hdrhostlen) {
+			if (hdrhostlen != urlhostlen || strcmp(hdrhost, urlhost) != 0) {
+				frang_msg("Host header is not equal to host in URL",
```
keshonok: Line length exceeds 80 characters for the above two lines.

tempesta_fw/classifier/frang.c
```diff
@@ -482,18 +484,66 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
-	if ((buf = tfw_pool_alloc(req->pool, len)) == NULL)
-		return TFW_BLOCK;
-	tfw_str_to_cstr(&field, buf, len);
-	if (!tfw_addr_pton(buf, &addr)) {
+
+	urlhostlen = req->host.len;
+	hdrhostlen = field.len;
+
+	if (urlhostlen) {
+		if ((urlhost = tfw_pool_alloc(req->pool, urlhostlen + 1)) == NULL) {
```
keshonok: Line length exceeds 80 characters.

keshonok: Do we need to linearize `req->host`? Why can't we use `tfw_str_eq_cstr()` for comparison with `field` (`hdrhost`)?

tempesta_fw/classifier/frang.c
```diff
@@ -482,18 +484,66 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+		if (hdrhostlen) {
+			if (hdrhostlen != urlhostlen || strcmp(hdrhost, urlhost) != 0) {
+				frang_msg("Host header is not equal to host in URL",
+					&ra->addr, "\n");
+				ret = TFW_BLOCK;
+				goto finish;
+			}
+		} else {
+			frang_msg("Host header is empty",
+			  &ra->addr, "\n");
```
keshonok: These two lines should be one line.

tempesta_fw/classifier/frang.c
```diff
@@ -482,18 +484,66 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
 			  &ra->addr, "\n");
-		tfw_pool_free(req->pool, buf, len);
-		return TFW_BLOCK;
+		ret = TFW_BLOCK;
+		goto finish;
 	}
-	tfw_pool_free(req->pool, buf, len);
-	return TFW_PASS;
+
+	if (req->uri_full && urlhostlen == 0 && hdrhostlen != 0) {
```
keshonok: I am not convinced that this is a correct interpretation of RFC.

tempesta_fw/classifier/frang.c
```diff
@@ -482,18 +484,66 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+			}
+		} else {
+			frang_msg("Host header is empty",
+			  &ra->addr, "\n");
+			ret = TFW_BLOCK;
+			goto finish;
+		}
+	} else {
+		TFW_DBG("No host in url\n");
+	}
```
keshonok: * Initialize `int ret = TFW_BLOCK` instead of `TFW_PASS` at function start.
* Remove all `ret = TFW_BLOCK` above, leaving just `goto finish;`
* Put `ret = TFW_PASS` here.

tempesta_fw/classifier/frang.c
```diff
@@ -462,38 +462,84 @@ frang_http_ct_check(const TfwHttpReq *req, FrangAcc *ra)
...
 frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
 {
+	int ret = TFW_BLOCK;
 	TfwAddr addr;
+	char *hdrhost = NULL;
+	size_t hdrhostlen = 0;
 	TfwStr field;
-	char *buf;
-	int len;
+
```
keshonok: It's better to reorder the variables so that they take less space on stack.

tempesta_fw/classifier/frang.c
```diff
@@ -462,38 +462,84 @@ frang_http_ct_check(const TfwHttpReq *req, FrangAcc *ra)
...
 		frang_msg("Host header field", &ra->addr, " is missed\n");
 		return TFW_BLOCK;
 	}
-
+	TFW_DBG3("Host: header present. Length = %i, flags = 0x%04X\n",
+			 req->h_tbl->tbl[TFW_HTTP_HDR_HOST].len,
+			 (int)(req->h_tbl->tbl[TFW_HTTP_HDR_HOST].flags));
 	tfw_http_msg_hdr_val(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST],
-			     TFW_HTTP_HDR_HOST, &field);
+		     TFW_HTTP_HDR_HOST, &field);
```
keshonok: Improper alignment.

tempesta_fw/classifier/frang.c
```diff
@@ -462,38 +462,84 @@ frang_http_ct_check(const TfwHttpReq *req, FrangAcc *ra)
...
+		goto finish;
+	}
+
+	tfw_str_to_cstr(&field, hdrhost, hdrhostlen + 1);
+	if (!tfw_addr_pton(hdrhost, &addr)) {
 		frang_msg("Host header field contains IP address",
-			  &ra->addr, "\n");
-		tfw_pool_free(req->pool, buf, len);
-		return TFW_BLOCK;
+			&ra->addr, "\n");
```
keshonok: Improper alignment.

tempesta_fw/classifier/frang.c
```diff
@@ -462,38 +462,84 @@ frang_http_ct_check(const TfwHttpReq *req, FrangAcc *ra)
...
+		goto finish;
 	}
-	tfw_pool_free(req->pool, buf, len);
-	return TFW_PASS;
+
+	if (req->flags & TFW_HTTP_URI_FULL) {
+		/* Host: header must be equal to host in URI */
+		if (!tfw_str_eq_cstr(&req->host, hdrhost,
+				hdrhostlen, TFW_STR_EQ_CASEI))
```
keshonok: Improper alignment.

tempesta_fw/classifier/frang.c
```diff
@@ -462,38 +462,84 @@ frang_http_ct_check(const TfwHttpReq *req, FrangAcc *ra)
...
+		{
+			frang_msg("Host header is not equal to host in URL",
+					&ra->addr, "\n");
```
keshonok: Improper alignment.

tempesta_fw/classifier/frang.c
```diff
@@ -462,38 +462,84 @@ frang_http_ct_check(const TfwHttpReq *req, FrangAcc *ra)
...
+	* Perhaps should implement a tiny FSM or postpone the task until we
+	* have a good regex library.
+	* For now just linearize the Host header field TfwStr{} string.
+	*/
+	hdrhostlen = field.len;
+	if ((hdrhost = tfw_pool_alloc(req->pool, hdrhostlen + 1)) == NULL) {
+		TFW_ERR("Can not allocate memory\n");
+		goto finish;
+	}
+
```
keshonok: Logically, the code above and and the code below belong to the same logical block, so I'd remove the empty line here .

tempesta_fw/classifier/frang.c
```diff
@@ -531,6 +577,35 @@ enum {
 	Frang_Req_Done
 };
 
+const char *__state_name_array[] = {
```
keshonok: The array is only used in debug statements. Please make it so it's compiled conditionally.

tempesta_fw/classifier/frang.c
```diff
@@ -531,6 +577,35 @@ enum {
...
+	"Frang_Req_Body_Timeout",
+	"Frang_Req_Body_ChunkCnt",
+	"Frang_Req_Body_Len",
+
+	"Frang_Req_Body_NoState",
+
+	"Frang_Req_Done"
+};
+
+#define __state_name(state) ((state >= 0 && state <= Frang_Req_Done) ?	\
```
keshonok: Please make it compiled conditionally.

tempesta_fw/http.h
```diff
@@ -83,6 +83,7 @@ enum {
 #define TFW_HTTP_PF_CR			0x01
 #define TFW_HTTP_PF_LF			0x02
 #define TFW_HTTP_PF_CRLF		(TFW_HTTP_PF_CR | TFW_HTTP_PF_LF)
+#define TFW_HTTP_URI_FULL		0x04
```
keshonok: Wrong place. The block of definitions is for `parser->flags`, while you're using it in `req->flags`

tempesta_fw/http.h
```diff
@@ -135,6 +136,7 @@ typedef struct tfw_http_parser {
 	int		_i_st;
 	int		to_read;
 	unsigned int	_tmp_acc;
+	int		_header_tag;
```
keshonok: Unnecessary. This should be controlled via parser flags.

tempesta_fw/http_msg.c
```diff
@@ -45,16 +45,23 @@ tfw_http_msg_hdr_val(TfwStr *hdr, int id, TfwStr *val)
 	TfwStr *c;
 	int nlen = hdr_lens[id];
 
-	BUG_ON(TFW_STR_PLAIN(hdr));
 	BUG_ON(TFW_STR_DUP(hdr));
-	BUG_ON(nlen >= hdr->len);
+	BUG_ON(nlen > hdr->len);
```
keshonok: What has changed that `>=` is no longer valid?

tempesta_fw/http_msg.c
```diff
@@ -45,16 +45,23 @@ tfw_http_msg_hdr_val(TfwStr *hdr, int id, TfwStr *val)
 	TfwStr *c;
 	int nlen = hdr_lens[id];
 
-	BUG_ON(TFW_STR_PLAIN(hdr));
```
keshonok: I'd create a specific `BUG_ON()` rather than removing it completely. For instance, in this case the only exception is the `Host:` header. Why not modify the `BUG_ON()` to crash on anything else if a string is plain?

tempesta_fw/http_msg.c
```diff
@@ -45,16 +45,23 @@ tfw_http_msg_hdr_val(TfwStr *hdr, int id, TfwStr *val)
...
 	BUG_ON(TFW_STR_DUP(hdr));
-	BUG_ON(nlen >= hdr->len);
+	BUG_ON(nlen > hdr->len);
 	BUG_ON(id >= TFW_HTTP_HDR_RAW);
 
 	*val = *hdr;
 
+	/* Field value, if it exist, lies in the separate chunk.
+	 * So we pass several first chunks, containing field name,
```
keshonok: we skip

tempesta_fw/http_msg.c
```diff
@@ -45,16 +45,23 @@ tfw_http_msg_hdr_val(TfwStr *hdr, int id, TfwStr *val)
...
+	 * to get the field value. If we have field with empty value,
+	 * we get an empty string with length = 0 and pnt from the
+	 * last name chunk, but it is unimportant.
```
keshonok: I am not sure I understand the comment. Also, what is `pnt`?

tempesta_fw/http_parser.c
```diff
@@ -1248,6 +1217,7 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 	unsigned char c = *p;
 	__FSM_DECLARE_VARS();
 
+	parser->_header_tag = -1;
```
keshonok: Unnecessary when flags are used.

tempesta_fw/http_parser.c
```diff
@@ -1495,7 +1477,69 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
...
+	 * ....
+	 * If the authority component is missing or
+	 * undefined for the target URI, then a client MUST send a Host header
+	 * field with an empty field-value.
+	 */
+	__FSM_STATE(RGen_LWS) {
+		switch (c) {
+		case '\r':
+			TFW_DBG3("\\r symbol\n");
+			if (parser->_header_tag >= 0) {
```
keshonok: There's no need to check for this case every time we process a header field. We know exactly when we need this particular case. For that reason even the flag mentioned above (or a new member of the parser structure, for that matter) is not needed.

It makes sense to leave the original `RGen_LWS` where it was, and replace it with a set of macros like the original `RGen_LWS()` and a new `RGen_LWS_empty()` that would block or pass a header with an empty value string. Most of the code in these macros is exactly the same, and one block of code will need to be parametrized.

tempesta_fw/http_parser.c
```diff
@@ -1495,7 +1477,69 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
...
+				 * possibility of empty value. For such headers,
+				 * we set parser->_header_tag. If this field
+				 * is set, and header is empty, we call 
+				 * tfw_http_msg_hdr_close for it. So, this
+				 * field has 2 purposes: as flag, that empty
+				 * header is possible and shows, with which
+				 * header we work at the moment.
+				 */
+				if (likely(!(parser->flags & TFW_HTTP_PF_CRLF))) {
+					TFW_DBG3("CR and LF flag not set. Setting CR.\n");
```
keshonok: I believe that all debug printout in this state should not be here in the final version. This includes all `TFW_DBG3()` calls in this state.

tempesta_fw/http_parser.c
```diff
@@ -1592,8 +1636,11 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 
 	/* Request headers are fully read. */
 	__FSM_STATE(Req_HdrDone) {
-		if (c == '\n')
+		if (c == '\n') {
+			TFW_DBG3("CRLF end\n");
```
keshonok: Excessive debug printout. Please remove when you're finished with debugging.

tempesta_fw/http_parser.c
```diff
@@ -103,8 +103,9 @@ done:									\
 do {									\
 	p += n;								\
 	if (unlikely(p >= data + len || !*p)) {				\
+		TFW_DBG3("go to postpone\n");				\
```
keshonok: Excessive debug printout. Please remove when you're finished with debugging.

tempesta_fw/http_parser.c
```diff
@@ -1436,10 +1414,12 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 	__FSM_STATE(Req_Hdr) {
 		if (unlikely(c == '\r')) {
 			if (!(req->body.flags & TFW_STR_COMPLETE)) {
+				TFW_DBG3("CRLF start\n");
```
keshonok: Excessive debug printout. Please remove when you're finished with debugging.

tempesta_fw/http_parser.c
```diff
@@ -1495,7 +1477,69 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
...
+				} else {
+					TFW_DBG3("CR or LF already set\n");
+				}
+				tfw_http_msg_hdr_chunk_fixup(msg, data, p - data);
+				if (tfw_http_msg_hdr_close(msg, parser->_header_tag))
+					return TFW_BLOCK;
+				parser->_header_tag = -1;
+				__FSM_MOVE(RGen_LF);
+			} else {
+				if (likely(!(parser->flags & TFW_HTTP_PF_LF))) {
```
keshonok: This appears to be a bug. This block of code should be the same as in the original macro/state.

tempesta_fw/http_msg.c
vladtcvs: case nlen == hdr->len is valid, it is a case of empty value

keshonok: Would you please add a comment saying exactly that?

tempesta_fw/classifier/frang.c
```diff
@@ -531,6 +574,40 @@ enum {
...
+	"Frang_Req_Body_Timeout",
+	"Frang_Req_Body_ChunkCnt",
+	"Frang_Req_Body_Len",
+
+	"Frang_Req_Body_NoState",
+
+	"Frang_Req_Done"
+};
+
+
```
keshonok: Excess empty line.

tempesta_fw/classifier/frang.c
```diff
@@ -531,6 +574,40 @@ enum {
...
+#define __state_name(state) ((state >= 0 && state <= Frang_Req_Done) ?	\
+				__state_name_array[state] :		\
+				"Wrong state")
+#else
+#define __state_name(state) ("")
```
keshonok: Do we really need this without debugging?

tempesta_fw/http_msg.c
```diff
@@ -45,13 +45,28 @@ tfw_http_msg_hdr_val(TfwStr *hdr, int id, TfwStr *val)
 	TfwStr *c;
 	int nlen = hdr_lens[id];
 
-	BUG_ON(TFW_STR_PLAIN(hdr));
 	BUG_ON(TFW_STR_DUP(hdr));
-	BUG_ON(nlen >= hdr->len);
 	BUG_ON(id >= TFW_HTTP_HDR_RAW);
 
+	/* Only Host: header is allowed to be empty */
+	if (id == TFW_HTTP_HDR_HOST) {
```
keshonok: There are cases where `BUG_ON()` may be compiled out. That would leave an empty block here. Please put everything in `BUG_ON()` statements.

tempesta_fw/http_parser.c
```diff
@@ -1237,6 +1207,63 @@ __req_parse_x_forwarded_for(TfwHttpMsg *msg, unsigned char *data, size_t len)
...
+/* In request we should pass empty headers:
+ * RFC 7230 5.4:
+ * ....
+ * ....
+ * If the authority component is missing or
+ * undefined for the target URI, then a client MUST send a Host header
+ * field with an empty field-value.
+ *
+ * ATTENTION: using of RGEN_LWS_empty should be corresponding with
+ * tfw_http_msg_hdr_val function
```
keshonok: Please use `NOTE:`. Also, the test is not clear. Did you mean something like the following?
```The use of RGEN_LWS_empty() should be matched with the BUG_ON() statements in tfw_http_msg_hdr_val() function```.

File: tempesta_fw/http_parser.c [hunk 42]
```diff
@@ -1237,6 +1207,63 @@ __req_parse_x_forwarded_for(TfwHttpMsg *msg, unsigned char *data, size_t len)
...
+ */
+#define RGEN_LWS_common_cases(move_state)			\
```
keshonok: Please use the name conventions that are already used in the macros. In this case, I believe, it should be `(st)`.

tempesta_fw/http_parser.c
```diff
@@ -1237,6 +1207,63 @@ __req_parse_x_forwarded_for(TfwHttpMsg *msg, unsigned char *data, size_t len)
...
+		if (likely(!(parser->flags & TFW_HTTP_PF_CRLF))) {	\
+			parser->flags |= TFW_HTTP_PF_CR;		\
+			__FSM_MOVE(RGen_LWS);				\
+		}							\
+		return TFW_BLOCK;					\
+	RGEN_LWS_common_cases(RGen_LWS)					\
+	}								\
+}
+
+
```
keshonok: Excess empty line.

tempesta_fw/http_parser.c
keshonok: Please move the macro back to the part of the source file where macros are grouped together.

tempesta_fw/classifier/frang.c
vladtcvs: yes, it is one block. empty line for more comfortable reading. But it's a matter of taste

tempesta_fw/classifier/frang.c
```diff
@@ -540,18 +583,63 @@ int __fsm_const_state = Frang_Req_0; /* make compiler happy */
 #define __FRANG_FSM_START(st)						\
 switch(st)
 
+#if DEBUG == 3
+
```
keshonok: Excess empty line.

tempesta_fw/classifier/frang.c
```diff
@@ -540,18 +583,63 @@ int __fsm_const_state = Frang_Req_0; /* make compiler happy */
 #define __FRANG_FSM_START(st)						\
 switch(st)
 
+#if DEBUG == 3
```
keshonok: Please use brackets. See `log.h` for examples.

tempesta_fw/classifier/frang.c
```diff
@@ -540,18 +583,63 @@ int __fsm_const_state = Frang_Req_0; /* make compiler happy */
...
 	TFW_DBG3("Frang return %s\n", r == TFW_PASS ? "PASS" : "BLOCK");\
 	req->frang_st = __fsm_const_state;
 
 #define __FRANG_FSM_STATE(st)						\
 case st:								\
 st: __attribute__((unused))						\
-	TFW_DBG3("enter FRANG FSM at state %d\n", st);			\
+	TFW_DBG3("enter FRANG FSM at state %d = %s\n", st, __state_name(st));\
+	__fsm_const_state = st; /* optimized out to constant */
+
```
keshonok: Excess empty line.

tempesta_fw/classifier/frang.c
```diff
@@ -540,18 +583,63 @@ int __fsm_const_state = Frang_Req_0; /* make compiler happy */
...
+#else
+
```
keshonok: Excess empty line.

tempesta_fw/classifier/frang.c
```diff
@@ -540,18 +583,63 @@ int __fsm_const_state = Frang_Req_0; /* make compiler happy */
...
 	req->frang_st = __fsm_const_state;
 
 #define __FRANG_FSM_STATE(st)						\
 case st:								\
 st: __attribute__((unused))						\
-	TFW_DBG3("enter FRANG FSM at state %d\n", st);			\
+	TFW_DBG3("enter FRANG FSM at state %d = %s\n", st, __state_name(st));\
+	__fsm_const_state = st; /* optimized out to constant */
+
+#else
```
keshonok: You can use a comment to visually distinguish this statement from the rest of the code. Something like that:
```C
#else /* ( DEBUG == 3) */
```

tempesta_fw/classifier/frang.c
```diff
@@ -540,18 +583,63 @@ int __fsm_const_state = Frang_Req_0; /* make compiler happy */
...
+#define __FRANG_FSM_FINISH()						\
+done:									\
+	req->frang_st = __fsm_const_state;
+
+#define __FRANG_FSM_STATE(st)						\
+case st:								\
+st: __attribute__((unused))						\
 	__fsm_const_state = st; /* optimized out to constant */
 
+#endif
```
keshonok: Use a comment to visually distinguish this statement from the rest of the code. Something like that:
```C
#endif /* ( DEBUG == 3) */
```

tempesta_fw/http_msg.c
```diff
@@ -45,13 +45,23 @@ tfw_http_msg_hdr_val(TfwStr *hdr, int id, TfwStr *val)
...
+	 * Not empty headers are compount strings. */
+	BUG_ON(id == TFW_HTTP_HDR_HOST ? (nlen > hdr->len) :
+		(nlen >= hdr->len || TFW_STR_PLAIN(hdr)));
+
 	*val = *hdr;
 
+	/* Field value, if it exist, lies in the separate chunk.
+	 * So we skip several first chunks, containing field name,
+	 * to get the field value. If we have field with empty value,
+	 * we get an empty string with val->len = 0 and val->pnt from the
```
keshonok: val->ptr

tempesta_fw/http_parser.c
```diff
@@ -604,11 +605,49 @@ __FSM_STATE(prefix ## _Done) {						\
 	return TFW_BLOCK;						\
 }
 
-/*
- * Read LWS at arbitrary position and move to stashed state.
- * This is bit complicated (however you can think about this as
- * a plain pushdown automaton), but reduces FSM code size.
```
krizhanovsky: Don't remove the comment about the automaton construction.

tempesta_fw/http_parser.c
```diff
@@ -604,11 +605,49 @@ __FSM_STATE(prefix ## _Done) {						\
...
+/* In request we should pass empty headers:
+ * RFC 7230 5.4:
+ * ....
+ * ....
+ * If the authority component is missing or
+ * undefined for the target URI, then a client MUST send a Host header
+ * field with an empty field-value.
+ *
+ * NOTE: using of RGEN_LWS_empty should be matched with
+ * the BUG_ON() statements in tfw_http_msg_hdr_val function
```
krizhanovsky: Please place the comment about RGEN_LWS_empty just before the macro and finish the sentence with dot.

tempesta_fw/http_parser.c
```diff
@@ -604,11 +605,49 @@ __FSM_STATE(prefix ## _Done) {						\
...
  */
+#define RGEN_LWS_common_cases(st)				\
```
krizhanovsky: The macro is just same piece of code for the macros below, so I'd prefer name it ```__RGEN_LWS```

tempesta_fw/http_parser.c
```diff
@@ -604,11 +605,49 @@ __FSM_STATE(prefix ## _Done) {						\
...
+	parser->_i_st = 0;					\
+	BUG_ON(unlikely(p >= data + len || !*p));		\
+	goto fsm_reenter;
+
+#define RGEN_LWS_empty()						\
+__FSM_STATE(RGen_LWS_empty) {						\
+	switch (c) {							\
+	case '\r':							\
+		if (unlikely(parser->flags & TFW_HTTP_PF_LF))		\
+			return TFW_BLOCK;				\
```
krizhanovsky: It seems the check is also good for RGen_LWS, aren't it?

tempesta_fw/http_parser.c
```diff
@@ -604,11 +605,49 @@ __FSM_STATE(prefix ## _Done) {						\
...
+		parser->flags |= TFW_HTTP_PF_CR;			\
+		tfw_http_msg_hdr_chunk_fixup(msg, data, p - data);	\
+		if (tfw_http_msg_hdr_close(msg, parser->_hdr_tag))	\
+			return TFW_BLOCK;				\
+		__FSM_MOVE(RGen_LF);					\
+	RGEN_LWS_common_cases(RGen_LWS_empty)				\
+	}								\
+}
+
```
krizhanovsky: The change should be bit more complex.

The initial ```RGEN_LWS()``` was written by RFC 2616 which allows line folding for headers. I.e. following request is correct in terms of RFC 2616:

        GET / HTTP/1.1\r\n
        Host:    \r\n
                foo.com\r\n
        Connection: close\r\n
        \r\n

Note that spaces precedes "foo.com" and this is required by RFC 2616. However, the logic is broken in initial ```RGEN_LWS()``` and "Connection: close" will be treated as Host header value in following request:

        GET / HTTP/1.1\r\n
        Host:\r\n
        Connection: close\r\n
        \r\n

Moreover, the feature is prohibited in RFC 7230:

        This specification deprecates such line folding except within the message/http
        media type (Section 8.3.1).

We don't use the parser to process body of ```message/http``` content type. RFC 7230 defines ```field-value``` as ```*( field-content / obs-fold )``` which could be zero length according to ABNF, RFC 5234, chapter 3.6. So we should process zero-length HTTP headers normally.

tempesta_fw/classifier/frang.c
```diff
@@ -464,28 +464,80 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
-	 * Perhaps should implement a tiny FSM or postpone the task until we
-	 * have a good regex library.
-	 * For now just linearize the Host header field TfwStr{} string.
-	 */
-	if (!tfw_addr_pton(&field, &addr)) {
-		frang_msg("Host header field contains IP address",
-			  &ra->addr, "\n");
-		return TFW_BLOCK;
+	if (!TFW_STR_EMPTY(&field))
+		hdrhostlen = field.len;
```
krizhanovsky: ```hdrhostlen``` is not needed, you can use field.len hereafter.

tempesta_fw/classifier/frang.c
```diff
@@ -464,28 +464,80 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+
+	if (hdrhostlen > 0) {
+		if (!tfw_addr_pton(&field, &addr)) {
+			frang_msg("Host header field contains IP address",
+				&ra->addr, "\n");
+			goto finish;
+		}
+
+		if ((hdrhost = tfw_pool_alloc(req->pool, hdrhostlen + 1)) == NULL) {
```
krizhanovsky: The line is longer than 80 characters

tempesta_fw/classifier/frang.c
```diff
@@ -464,28 +464,80 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+			TFW_ERR("Can not allocate memory\n");
+			goto finish;
+		}
+
+		tfw_str_to_cstr(&field, hdrhost, hdrhostlen + 1);
 	}
-	return TFW_PASS;
+
+	if (req->flags & TFW_HTTP_URI_FULL) {
```
krizhanovsky: If the condition is false, then you don't need to allocate ```hdrhost``` and linearize ```field```

tempesta_fw/http.h
```diff
@@ -119,22 +112,25 @@ typedef struct {
  * (e.g. process LWS using @state while current state is saved in @_i_st
  * or using @_i_st parse value of a header described.
  *
- * @state	- current parser state;
- * @_i_st	- helping (interior) state;
+ * @flags	- flags. Now used flags: TFW_HTTP_PF_CR, TFW_HTTP_PF_LF
```
krizhanovsky: You're removing flags field, so the comment also should be removed.

tempesta_fw/http.h
```diff
@@ -186,6 +182,8 @@ typedef struct {
 /* Request flags */
 #define TFW_HTTP_STICKY_SET		0x0100	/* Need 'Set-Cookie` */
 #define TFW_HTTP_FIELD_DUPENTRY		0x0200	/* Duplicate field */
+/* URI has form http://authority/path, not just /path */
+#define TFW_HTTP_URI_FULL		0x0400
```
krizhanovsky: In case of ```http://authority/path``` ```req->host``` should not be empty. So why do you need the flag?

krizhanovsky:
Good to merge

PR #331, opened 2015-10-27
Fix #192: zero-copy tfw_addr_pton

### LINKED ISSUE #192: tfw_addr_pton() and tfw_str_to_cstr() rework
Currently Host header must be converted to C-string before passing to ```tfw_addr_pton()```, so there is unnecessary copying and additional preasure on TfwPool. Meantime ```tfw_str_to_cstr()``` was considered as deprecated and should not be used. If C-string must be passed to ```tfw_addr_pton()```, then automatic plain TfwStr could be defined.

Example of the issue is ```frang_http_host_check()```.

almk277:
#192
Iterating over C-string is replaced with iterating over TfwStr.

### DIFF
--- a/tempesta_fw/addr.c
+++ b/tempesta_fw/addr.c
@@ -40,30 +40,35 @@ validate_addr(const TfwAddr *addr)
 }
 
 static int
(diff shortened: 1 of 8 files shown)


krizhanovsky:
Good to merge

Issue #230, opened 2015-09-03
HTTP bomber


krizhanovsky:
High performance tool based on synchronous sockets and fuzzer (#150) must be made to do multi-gigabit stress testing for Web servers.

The tool must have following options:

* number of concurrent connections;

* number of messages sent through one connection (i.e. how frequenty new connection will be established)

krizhanovsky:
Please implement connections termination by RST as well as using FIN to cause stress on SS subsystem.

krizhanovsky:
Please also test the tool against any regular Web server like Apache HTTPD or Nginx.

PR #335, opened 2015-10-29
#230. HTTP bomber

avbelov23:


tempesta_fw/t/unit/tfw_bomber.c
```diff
@@ -0,0 +1,496 @@
...
+static int tfw_connects = 16;
+static int tfw_messages = 4;
+static char *tfw_server = "127.0.0.1:80";
+
+module_param(tfw_threads, int, S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP);
+module_param(tfw_connects, int, S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP);
+module_param(tfw_messages, int, S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP);
+module_param(tfw_server, charp, 0);
+
+MODULE_PARM_DESC(tfw_server, "Server host address and optional port nunber");
```
krizhanovsky: Please add description for other parameters.

tempesta_fw/t/unit/tfw_bomber.c
```diff
@@ -0,0 +1,496 @@
...
+static void
+tfw_bomber_send_msgs(void)
+{
+	int i, k, m;
+
+	for (i = 0; i < tfw_threads; i++) {
+		for (k = 0; k < tfw_connects; k++) {
+			for (m = 0; m < tfw_messages; m++) {
+				if (tfw_bomber_desc[i][k].sk) {
+					msg_send(&tfw_bomber_desc[i][k]);
```
krizhanovsky: Why do we do this in single thread? Rather we should send messages from _all_ the threads concurrently.

tempesta_fw/t/unit/tfw_bomber.c
```diff
@@ -0,0 +1,496 @@
...
+	TfwHttpMsg *req;
+	TfwMsgIter it;
+
+	sk = desc->sk;
+	BUG_ON(!sk);
+
+	atomic_inc(&tfw_bomber_request_nsend);
+
+	len = 1 * 1024 * 1024;
+	str = vmalloc(len);
```
krizhanovsky: This is the killer for performance. Just create an array of ```void *``` with size of ```tfw_threads```, and preallocate memory for each member of the array. After than each thread can use the memory regions addressing them by their identifiers.

tempesta_fw/t/unit/tfw_bomber.c
```diff
@@ -0,0 +1,496 @@
...
+		task = kthread_create(tfw_bomber_thread_connect, (void *)(long)i,
+				      "tfw_bomber_thread_connect_%02d", i);
+		if (IS_ERR_OR_NULL(task)) {
+			ret = PTR_ERR(task);
+			SS_ERR("Unable to create a thread: %s%02d (%d)\n",
+			       "tfw_bomber_thread_connect", i, ret);
+			break;
+		}
+		tfw_bomber_connect_task[i] = task;
+	}
```
krizhanovsky: The design is simply wrong: ```tfw_threads``` threads just do reconnects in ```tfw_bomber_thread_connect``` while the only one thread ```tfw_bomber_thread_finish``` sends messages. Each working thread must establish ```tfw_connects``` and send ```tfw_messages``` through each of them. When ```tfw_messages``` are sent through a connection, the connection must be reestablished by the working thread. Thus we need one more argument - ```tfw_total_messages``` which defines number of iterations in a thead as ```iter = tfw_total_messages / tfw_threads / tfw_connects / tfw_messages```.

tempesta_fw/t/unit/bomber.c
```diff
@@ -0,0 +1,485 @@
...
+
+	fuzz_set_only_valid_gen(true);
+
+	if (tfw_addr_pton(&TFW_STR_FROM(server), &tfw_bmb_server_address)) {
+		TFW_ERR("Unable to parse server's address: %s", server);
+		return -EINVAL;
+	}
+	TFW_DBG("Started bomber module, server's address is %s\n", server);
+
+	tfw_bmb_desc = kmalloc(nthreads * sizeof(tfw_bmb_desc_t *), GFP_KERNEL);
```
krizhanovsky: You do several allocations of ```nthreads``` of different structures. The code could be simpler if you join the structures and do only one allocation for one array of the merged structures.

tempesta_fw/t/unit/bomber.c
```diff
@@ -0,0 +1,485 @@
...
+		struct task_struct *task;
+
+		task = kthread_create(tfw_bmb_worker, (void *)(long)i, "worker");
+		if (IS_ERR_OR_NULL(task)) {
+			ret = PTR_ERR(task);
+			TFW_ERR("Unable to create thread: (%d)\n", ret);
+			goto err_create_tasks;
+		}
+		tfw_bmb_tasks[i] = task;
+	}
```
krizhanovsky: Again, what's the sense of 3 different loops with the same counter?

krizhanovsky:
Looks good for me

tempesta_fw/t/unit/tfw_fuzzer.h
```diff
@@ -32,11 +34,18 @@ typedef enum {
 	N_FIELDS,
 } field_t;
 
-int fuzz_gen(char *str, char *end, field_t start, int move, int type);
+typedef struct {
+	int i[N_FIELDS];
+	bool is_only_valid;
+	bool is_chancked_body;
```
keshonok: chunked?

tempesta_fw/t/unit/bomber.c
```diff
@@ -0,0 +1,521 @@
...
+stop_threads:
+	tfw_bmb_stop_threads();
+	tfw_bmb_free();
+
+	return r;
+}
+
+static void
+tfw_bmb_exit(void)
+{
```
keshonok: Should it be empty?

avbelov23: Yes. There is no sense here to free resources. It can be in the init.

tempesta_fw/t/unit/tfw_fuzzer.h
avbelov23: Yes. Chunked. Thanks!

Issue #317, opened 2015-10-19
HTTP parser: incorrect Expires header parsing


avbelov23:
If a response is split in the Expires field-value, then __resp_parse_expires() is unable to continue parsing properly.

Response:

        parse 161 server data bytes (HTTP/1.1 200 OK
        Accept-Ranges:none
        Set-Cookie:name=value
        ETag:"56d-9989200-1132c580"
        Server:Apache/2.2.17 (Win32) PHP/5.3.5
        Expires:Tue, 31 Jan 2012 15:02:5)
        
        parse 157 server data bytes (3 GMT
        Connection:close
        Content-Type:image/jpeg
        Content-Type:image/jpeg
        Content-Length:0
        Transfer-Encoding:identity
        Cache-Control:max-stale=0

        0

        )

Log:

        [14541.678816] [sync_sockets]   parse 157 server data bytes (3 GMT
        Connection:close
        Content-Type:image/jpeg
        Content-Type:image/jpeg
        Content-Length:0
        Transfer-Encoding:identity
        Cache-Control:max-stale=0
        
        0
        
        ) on resp=ffff88005c77b040
        [14541.694513] [sync_sockets]       enter FSM at state 62
        [14541.697435] [sync_sockets]       parser: Resp_HdrExpiresV(62:11): c=0x33(3), r=-2
        [14541.702058] [tempesta]       store field chunk len=0 data=ffffc90002dd30a1 field=        <0x200,9,ffff88005c77b2c8>
        [14541.708310] [sync_sockets]       enter FSM at state 11
        [14541.711312] [sync_sockets]       parser: Resp_I_ExpSec(11:11): c=0x33(3), r=-2
        [14541.715678] [sync_sockets]       parser: Resp_I_EoL(16:16): c=0xa(.), r=-2
        [14541.719835] [sync_sockets]       parse raw header __resp_parse_expires: ret=-2 data_len=160

PR #342, opened 2015-11-07
#317: incorrect parsing of "Expires" seconds

almk277:
#317

### DIFF
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -1830,11 +1830,15 @@ enum {
 	/* Expires header */
 	Resp_I_Expires,
 	Resp_I_ExpDate,
+	Resp_I_ExpMonthSP,
 	Resp_I_ExpMonth,
 	Resp_I_ExpYearSP,
 	Resp_I_ExpYear,
+	Resp_I_ExpHourSP,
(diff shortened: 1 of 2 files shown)


tempesta_fw/http_parser.c
```diff
@@ -2112,14 +2112,16 @@ __resp_parse_expires(TfwHttpResp *resp, unsigned char *data, size_t len)
...
 	__FSM_STATE(Resp_I_EoL) {
-		if (c == '\r')
-			return p - data;
-		return CSTR_NEQ;
+		/* Skip rest of line: ' GMT'. */
+		__fsm_sz = len - (size_t)(p - data);
+		__fsm_ch = memchr(p, '\r', __fsm_sz);
+		if (__fsm_ch)
+			return __fsm_ch - data;
+		__FSM_I_MOVE_n(Resp_I_EoL, __fsm_sz);
```
krizhanovsky: The fix seems incomplete.

What about other numeric values? Did you check that test_http_parser.c doesn't fail on split Expires header any more?

Since it seems we cannot control length of numeric values, then please add checks for ```parser->_tmp_acc``` values.

krizhanovsky:
Good to merge

Issue #348, opened 2015-11-10
Close client connection on client's request


keshonok:
Some benchmarking tools expect that the other side of connection is an HTTP server, and expects certain standard behaviour from the server. In particular, they expect that the server will close a connection  when `Connection: close` header field is send in an HTTP request.

Tempesta, on the other hand, attempts to keep connections open as long as possible, and never closes connections with clients at this time. That confuses benchmarking tools, which stops them from doing their job.

Tempesta should close client connections when clients request that.

keshonok:
The issue is actually multifold.
- If a client sends the `Connection: close` header in an HTTP request, then the connection with the client should be closed after a response to that request is sent to the client.
- The above is still true when a response from a back-end server cannot be forwarded to the client for any reason, such as an error while processing or adjusting the response.
- The above is also true for 302, 404, 500, or 502 error responses that are generated by Tempesta itself when a request cannot be forwarded to a back-end server.
- In addition, the response must contain the same `Connection: close` header in an HTTP response when that header is present in an HTTP request.

keshonok:
Implemented via 15daa97.

PR #352, opened 2015-11-13
Add "Connection:" header to an HTTP response when necessary. (#348)

keshonok:


tempesta_fw/http.c
```diff
@@ -160,35 +123,57 @@ tfw_http_prep_hexstring(char *buf, u_char *value, size_t len)
...
 TfwHttpMsg *
-tfw_http_prep_302(TfwHttpMsg *hm, TfwStr *cookie)
+tfw_http_prep_302(TfwHttpMsg *hmreq, TfwStr *cookie)
 {
+#define S_302_PART_01	S_302 S_CRLF S_F_DATE
+#define S_302_PART_02	S_CRLF S_F_CONTENT_LENGTH "0" S_CRLF S_F_LOCATION
+#define S_302_PART_03	S_CRLF S_F_SET_COOKIE
+#define S_302_FIXLEN	SLEN(S_302_PART_01 S_V_DATE S_302_PART_02 S_302_PART_03)
+#define S_302_KEEP	S_CRLF S_H_CONN_KA
+#define S_302_CLOSE	S_CRLF S_H_CONN_CLOSE
```
krizhanovsky: Please don't place so many defines inside the function, it's good to place them bit ahead of the function.

tempesta_fw/http.c
```diff
@@ -197,30 +182,49 @@ tfw_http_prep_302(TfwHttpMsg *hm, TfwStr *cookie)
...
+	TfwStr *crlf = __TFW_STR_CH(msg, TFW_STR_CHUNKN(msg) - 1);
 	TfwHttpMsg *resp;
 	TfwMsgIter it;
 
+	if (conn_flag) {
+		unsigned long crlf_len = crlf->len;
+		if (conn_flag == TFW_HTTP_CONN_CLOSE) {
+			crlf->ptr = S_H_CONN_CLOSE;
+			crlf->len = SLEN(S_H_CONN_CLOSE);
+		} else if (conn_flag == TFW_HTTP_CONN_KA) {
```
krizhanovsky: Currently we support only few values for Connection header, while it's much more complex. So other flags will be introduced and it's better to check the flags by ```&```.

tempesta_fw/http.c
```diff
@@ -160,35 +123,57 @@ tfw_http_prep_hexstring(char *buf, u_char *value, size_t len)
...
 			{ .ptr = S_302_PART_02, .len = SLEN(S_302_PART_02) }
 		},
-		.len = SLEN(S_302_PART_01 S_302_PART_02 S_V_DATE),
+		.len = SLEN(S_302_PART_01 S_V_DATE S_302_PART_02),
 		.flags = 3 << TFW_STR_CN_SHIFT
 	};
+	TfwStr crlfcrlf = { .ptr = S_CRLFCRLF, .len = SLEN(S_CRLFCRLF) };
+	TfwStr crlf_keep = { .ptr = S_302_KEEP, .len = SLEN(S_302_KEEP) };
+	TfwStr crlf_close = { .ptr = S_302_CLOSE, .len = SLEN(S_302_CLOSE) };
+	TfwStr host, *crlf = &crlfcrlf;
```
krizhanovsky: Well, the variables are constant, so it seems there is no sense to allocate them on the stack on each function call and we can make them static.

krizhanovsky:
Good to merge with few cleanups

Issue #354, opened 2015-11-13
HTTP parser: Array index out of bounds


avbelov23:
`(echo -en "GET / HTTP/1.1\r\nContent-Length:0"; sleep 1; echo -en "\r\nHost:test\r\n\r\n") | nc 127.0.0.1 80` is blocked sometimes
The bug is not always played back as there is the reading of the array bounds - https://github.com/natsys/tempesta/blob/master/tempesta_fw/http_parser.c#L296

PR #356, opened 2015-11-16
#354. HTTP parser: Array index out of bounds

avbelov23:


tempesta_fw/http_parser.c
```diff
@@ -303,6 +301,9 @@ parse_int_a(unsigned char *data, size_t len, const unsigned long *delimiter_a,
 		*acc = *acc * 10 + *p - '0';
 	}
 
+	if (unlikely(p - data == len))
```
krizhanovsky: Double check is not acceptable in this performance critical function. Please rewrite the loop as while() with all necessary checks inside.

krizhanovsky:
Good to merge

PR #361, opened 2015-11-20
Close client connections when requested by clients. (#348)

keshonok:
When "Connection: close" header is present in an HTTP request from a client,
close the connection after the response is sent to that client.

### DIFF
--- a/tempesta_fw/connection.h
+++ b/tempesta_fw/connection.h
@@ -209,23 +209,6 @@ tfw_connection_live(TfwConnection *conn)
 	return conn->sk && ss_sock_live(conn->sk);
 }
 
-/*
- * Drop a connection. This function may be called only when there's
- * at least one extra reference to the connection (@conn->refcnt),
- * so if the reference is dropped here it's never the last one.
- * See comments to ss_do_droplink() and ss_droplink().
- *
- * Note: Don't put a BUG_ON() on @conn->refcnt in this function to
- * enforce the condition mentioned above. That may lead to a race
- * condition with ss_tcp_process_data(). If this function is called
- * under wrong conditions, the check may also be invalid.
- */
-static inline void
-tfw_connection_drop(TfwConnection *conn)
-{
-	ss_droplink(conn->sk);
-}
-
 /**
  * Check that TfwConnection resources are cleaned up properly.
  */
--- a/tempesta_fw/connection.h
+++ b/tempesta_fw/connection.h
@@ -209,6 +209,17 @@ tfw_connection_live(TfwConnection *conn)
 	return conn->sk && ss_sock_live(conn->sk);
 }
 
+/*
+ * Drop a connection. This function may be called only when there's
+ * at least one extra reference to the connection (@conn->refcnt),
+ * so if the reference is dropped here it's never the last one.
+ * See comments to ss_do_droplink() and ss_droplink().
+ *
+ * Note: Don't put a BUG_ON() on @conn->refcnt in this function to
+ * enforce the condition mentioned above. That may lead to a race
+ * condition with ss_tcp_process_data(). If this function is called
+ * under wrong conditions, the check may also be invalid.
+ */
 static inline void
 tfw_connection_drop(TfwConnection *conn)
 {
--- a/tempesta_fw/connection.h
+++ b/tempesta_fw/connection.h
@@ -209,6 +209,12 @@ tfw_connection_live(TfwConnection *conn)
 	return conn->sk && ss_sock_live(conn->sk);
 }
 
+static inline void
+tfw_connection_drop(TfwConnection *conn)
+{
+	ss_droplink(conn->sk);
+}
+
 /**
  * Check that TfwConnection resources are cleaned up properly.
  */
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -611,7 +611,7 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
 	unsigned int skb_len = skb->len;
 
 	BUG_ON(!conn->msg);
-	BUG_ON(off >= skb_len);
+	BUG_ON(data_off >= skb_len);
 
 	TFW_DBG2("Received %u client data bytes on conn=%p msg=%p\n",
 		 skb_len - off, conn, conn->msg);
@@ -701,6 +701,16 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
 			}
 		}
 		/*
+		 * Complete HTTP message has been collected and successfully
+		 * processed. Mark the message as complete in @conn, because
+		 * further handling of @conn depends on that. Future SKBs
+		 * will be put in a new message.
+		 * Otherwise, the function returns from inside the loop.
+		 * @conn->msg holds the reference to the message, which can
+		 * be used to release it.
+		 */
+		conn->msg = NULL;
+		/*
 		 * The request should either be stored or released.
 		 * Otherwise we lose the reference to it and get a leak.
 		 * As it may be released, save the needed flag for later use.
@@ -732,17 +742,6 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
 			conn->msg = (TfwMsg *)hmsib;
 		}
 	}
-	/*
-	 * All data in the SKB has been successfully
-	 * processed. Complete HTTP messages have been
-	 * collected, and stored or released. Future SKBs
-	 * should be put in a new message.
-	 *
-	 * Otherwise, the function just returns from inside
-	 * the loop. @conn->msg contains the reference to
-	 * a message, which can be used to release it.
-	 */
-	conn->msg = NULL;
 
 	return r;
 }
@@ -796,7 +795,6 @@ tfw_http_popreq(TfwConnection *conn)
 	spin_lock(&conn->msg_qlock);
 	if (unlikely(list_empty(&conn->msg_queue))) {
 		spin_unlock(&conn->msg_qlock);
-		TFW_WARN("Empty request queue\n");
 		return NULL;
 	}
 	msg = list_first_entry(&conn->msg_queue, TfwMsg, msg_list);
@@ -836,7 +834,7 @@ tfw_http_resp_process(TfwConnection *conn, struct sk_buff *skb,
 	unsigned int skb_len = skb->len;
 
 	BUG_ON(!conn->msg);
-	BUG_ON(off >= skb_len);
+	BUG_ON(data_off >= skb_len);
 
 	TFW_DBG2("received %u server data bytes on conn=%p msg=%p\n",
 		skb->len - off, conn, conn->msg);
@@ -951,10 +949,22 @@ next_resp:
 		 * for data exchange until that gets impossible.
 		 */
 		if ((req = tfw_http_popreq(conn)) != NULL) {
+			/*
+			 * Complete HTTP message has been collected and
+			 * successfully processed. Mark the message as
+			 * complete in @conn, because further handling
+			 * of @conn depends on that. Future SKBs will
+			 * be put in a new message.
+			 * Otherwise, the function returns from inside
+			 * the loop. @conn->msg holds the reference to
+			 * the message, which can be used to release it.
+			 */
+			conn->msg = NULL;
 			tfw_cache_resp_process((TfwHttpResp *)hmresp,
 					       req, tfw_http_resp_cache_cb);
 		} else {
-			/* conn->msg will get NULLed in the process. */
+			/* @conn->msg will get NULLed in the process. */
+			TFW_WARN("Paired request missing\n");
 			tfw_http_conn_msg_free(hmresp);
 		}
 
@@ -968,18 +978,7 @@ next_resp:
 			conn->msg = (TfwMsg *)hmsib;
 		}
 	}
-	/*
-	 * All data in the SKB has been processed, and the processing
-	 * is successful. A complete HTTP message has been collected,
-	 * sent to a client, and then released.
-	 *
-	 * Otherwise, the function just returns from inside the loop.
-	 * conn->msg contains the reference to a message, which can
-	 * be used to release it.
-	 *
-	 * @conn->msg is NULLed when the message is freed/released.
-	 * Future SKBs will be put in a new message.
-	 */
+
 	return r;
 }
 
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -366,27 +366,37 @@ tfw_http_conn_msg_free(TfwHttpMsg *hm)
 	tfw_http_msg_free(hm);
 }
 
-static void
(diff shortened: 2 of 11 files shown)


tempesta_fw/http.c
```diff
@@ -689,6 +701,14 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
 		tfw_cache_req_process((TfwHttpReq *)hmreq,
 				      tfw_http_req_cache_cb);
 
+		/*
+		 * At this stage the connection may have been dropped
+		 * and the socket closed. Further processing of data
+		 * in the SKB makes little sense.
+		 */
+		if (unlikely(!ss_sock_active(req_sk)))
+			return TFW_PASS;
```
krizhanovsky: Pessimism is bad. It's better to optimistically go further and correctly handle connection drop: anyway connection can be closed in any given time.

tempesta_fw/http.c
```diff
@@ -516,18 +516,20 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+		 * is either passed through from the back-end server, or
+		 * it is generated from the cache, so unrefer all its data.
 		 */
-		if (tfw_http_adjust_resp(resp, req))
-			return;
-		tfw_connection_send(req->conn, (TfwMsg *)resp, true);
+		if (tfw_http_adjust_resp(resp, req) == 0)
+			tfw_connection_send(req->conn, (TfwMsg *)resp, true);
+		if (req_flags & TFW_HTTP_CONN_CLOSE)
+			ss_droplink(req_sk);
```
krizhanovsky: Hm, shouldn't we do ```tfw_http_conn_msg_free((TfwHttpMsg *)req)``` here? It seems the job is done and we don't see where caller frees ```req```....

tempesta_fw/http.c
```diff
@@ -569,15 +576,19 @@ tfw_http_req_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
...
 	tfw_http_send_404((TfwHttpMsg *)req);
 	tfw_http_conn_msg_free((TfwHttpMsg *)req);
+	if (req_flags & TFW_HTTP_CONN_CLOSE)
+		ss_droplink(req_sk);
 	return;
 send_500:
 	tfw_http_send_500((TfwHttpMsg *)req);
 	tfw_http_conn_msg_free((TfwHttpMsg *)req);
+	if (req_flags & TFW_HTTP_CONN_CLOSE)
+		ss_droplink(req_sk);
```
krizhanovsky: Why not to call

	if (req_flags & TFW_HTTP_CONN_CLOSE)
		ss_droplink(req_sk);

from ```tfw_http_conn_msg_free()``` or some other wrapping function? If we drop the connection on SS layer, then we also should drop TfwConnection and associated messages.

tempesta_fw/sock.c
```diff
@@ -572,7 +580,19 @@ ss_tcp_process_data(struct sock *sk)
...
+			 * processing the receive queue and return.
+			 *
+			 * NOTE: This function is also called in TCP_CLOSE_WAIT
+			 * state from ss_tcp_state_change(), so that has to be
+			 * a valid socket state as well.
+			 */
+			if (unlikely(!ss_sock_active(sk))) {
+				SS_DBG("Link is dropped: sk = %p\n", sk);
+				return false;
+			}
```
krizhanovsky: The same as above - why do we need pessimistic workflow?

tempesta_fw/sock.c
```diff
@@ -539,8 +549,6 @@ ss_tcp_process_data(struct sock *sk)
 			 * the current SKB. Hand the SKB over to the
 			 * upper layer for processing.
 			 */
-			read_lock(&sk->sk_callback_lock);
```
krizhanovsky: Why don't we need the lock any more?

tempesta_fw/http.c
krizhanovsky: Why ```ss_close()``` can not be used here?

tempesta_fw/http.c
```diff
@@ -435,20 +455,45 @@ tfw_http_msg_create_sibling(TfwHttpMsg *hm, struct sk_buff **skb,
...
-		return TFW_HTTP_MSG_HDR_XFRM(hm, "Connection", "close",
-					     TFW_HTTP_HDR_CONNECTION, 0);
+		r = TFW_HTTP_MSG_HDR_XFRM(hm, "Connection", "close",
+					  TFW_HTTP_HDR_CONNECTION, 0);
+		if (r)
+			TFW_ERR("%s '%s' header: msg %p\n",
+				"Error setting", "Connection: close", hm);
+		else
+			TFW_DBG2("%s '%s' header: msg %p\n",
+				 "Set", "Connection: close", hm);
```
krizhanovsky: There is no sense to pass constant strings via ```%s``` argument.

tempesta_fw/http.c
```diff
@@ -463,11 +508,11 @@ tfw_http_add_x_forwarded_for(TfwHttpMsg *hm)
...
 	if (r)
-		TFW_ERR("can't add X-Forwarded-For header for %.*s to msg %p",
-			(int)(p - buf), buf, hm);
+		TFW_ERR("%s 'X-Forwarded-For: %.*s': msg %p\n",
+			"Error adding", (int)(p - buf), buf, hm);
 	else
-		TFW_DBG2("added X-Forwarded-For header for %*s\n",
-			 (int)(p - buf), buf);
+		TFW_DBG2("%s 'X-Forwarded-For: %.*s': msg %p\n",
+			 "Added", (int)(p - buf), buf, hm);
```
krizhanovsky: Again: why constant stings are moved to parameters?

tempesta_fw/http.c
```diff
@@ -517,17 +562,19 @@ static void
 tfw_http_req_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
 {
 	int r;
-	TfwConnection *conn;
+	int req_conn_close = !!(req->flags & TFW_HTTP_CONN_CLOSE);
```
krizhanovsky: ```req_conn_close``` is integer, so there is no need type conversion to boolean

tempesta_fw/http.c
```diff
@@ -382,6 +382,24 @@ tfw_http_conn_destruct(TfwConnection *conn)
...
+		 * requests in the queue that are kept until a paired
+		 * response comes. That will never happen now. Send
+		 * a client an error response. If the connection with
+		 * a client must be closed after a response is sent
+		 * to that client, then close the connection now.
+		 *
+		 * Hold @msg->conn reference through @msg until
+		 * the connection is dropped.
+		 *
+		 * Note: It's essential that there's no incoming
```
krizhanovsky: It'd be good to change the comment such that it's OK to close connection after receiving "Connection: close" regardless client activity.

tempesta_fw/connection.h
```diff
@@ -220,6 +220,23 @@ tfw_connection_live(TfwConnection *conn)
...
+ *
+ * Note: Don't put a BUG_ON() on @conn->refcnt in this function to
+ * enforce the condition mentioned above. That may lead to a race
+ * condition with ss_tcp_process_data(). If this function is called
+ * under wrong conditions, the check may also be invalid.
+ */
+static inline void
+tfw_connection_drop(TfwConnection *conn)
+{
+	ss_droplink(conn->sk);
```
krizhanovsky: The design problem still exists: you're exporting the new function ```ss_droplink()``` just to call ```connection_drop``` callback which is essentially placed at the same logical level as ```tfw_connection_drop()```.

tempesta_fw/http.c
```diff
@@ -563,24 +615,28 @@ tfw_http_req_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
...
 	tfw_http_send_404((TfwHttpMsg *)req);
+	if (req_conn_close)
+		tfw_connection_drop(cli_conn);
 	tfw_http_conn_msg_free((TfwHttpMsg *)req);
 	return;
 send_500:
 	tfw_http_send_500((TfwHttpMsg *)req);
+	if (req_conn_close)
+		tfw_connection_drop(cli_conn);
 	tfw_http_conn_msg_free((TfwHttpMsg *)req);
```
krizhanovsky: ```tfw_http_conn_msg_free()``` is used in http.c only, so it should be made static

tempesta_fw/http.c
```diff
@@ -563,24 +615,28 @@ tfw_http_req_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
...
-	tfw_connection_send(conn, (TfwMsg *)req, false);
-
+	tfw_connection_send(srv_conn, (TfwMsg *)req, false);
 	goto conn_put;
+
 send_404:
 	tfw_http_send_404((TfwHttpMsg *)req);
+	if (req_conn_close)
+		tfw_connection_drop(cli_conn);
 	tfw_http_conn_msg_free((TfwHttpMsg *)req);
```
krizhanovsky: The 3 lines above are ugly (see https://github.com/natsys/tempesta/pull/361/files#r47531757). You introduce ```tfw_connection_drop()``` just to call ```tfw_sock_clnt_drop()``` (for client, the server logic is the same). Meantime, ```tfw_http_conn_msg_free()``` calls ```tfw_cli_conn_release()``` after reference counter decrement. So we're in the same code place (sock_clnt.c) but, through 2 functions and using callbacks through SS. The design is too complicated and potentially buggy.

tempesta_fw/http.c
```diff
@@ -683,12 +740,37 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
...
+		 * @conn->msg holds the reference to the message, which can
+		 * be used to release it.
+		 */
+		conn->msg = NULL;
+		/*
 		 * The request should either be stored or released.
 		 * Otherwise we lose the reference to it and get a leak.
+		 * As it may be released, save the needed flag for later use.
 		 */
+		req_conn_close = !!(hmreq->flags & TFW_HTTP_CONN_CLOSE);
```
krizhanovsky: Why do we need ```req_conn_close``` if it's used only once?

tempesta_fw/http.c
```diff
@@ -683,12 +740,37 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
...
 		/*
+		 * Complete HTTP message has been collected and successfully
+		 * processed. Mark the message as complete in @conn, because
+		 * further handling of @conn depends on that. Future SKBs
+		 * will be put in a new message.
+		 * Otherwise, the function returns from inside the loop.
+		 * @conn->msg holds the reference to the message, which can
+		 * be used to release it.
+		 */
+		conn->msg = NULL;
```
krizhanovsky: Why do we need the change? Initially it was done to properly allocate a new message in ```tfw_connection_recv()```. While ```tfw_http_req_process()``` is called only from ```tfw_connection_recv()```, when the message is already allocated. Since you're assigns ```hmsib``` to ```conn->msg``` few lines below, it seems you need the change for ```tfw_cache_req_process((TfwHttpReq *)hmreq, tfw_http_req_cache_cb)```. And this isn't clear for me: both the function don't do anything with ```conn->msg``` and it should be properly freed by ```tfw_http_conn_msg_unlink()``` on connection closing...

tempesta_fw/http.c
```diff
@@ -683,12 +740,37 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
...
+		 * must be dropped after a response is sent to that client,
+		 * if the client sends "Connection: close" header field in
+		 * the request. Subsequent requests from the client coming
+		 * over the same connection are ignored.
+		 *
+		 * Note: This connection's @conn must not be dereferenced
+		 * from this point on.
+		 */
+		if (req_conn_close)
+			return TFW_STOP;
```
krizhanovsky: Sibling message ```hmsib``` is allocated, but not freed. You should not create it if ```req_conn_close == true```.

tempesta_fw/http.c
```diff
@@ -913,10 +988,22 @@ tfw_http_resp_process(TfwConnection *conn, struct sk_buff *skb,
...
+			 * Complete HTTP message has been collected and
+			 * successfully processed. Mark the message as
+			 * complete in @conn, because further handling
+			 * of @conn depends on that. Future SKBs will
+			 * be put in a new message.
+			 * Otherwise, the function returns from inside
+			 * the loop. @conn->msg holds the reference to
+			 * the message, which can be used to release it.
+			 */
+			conn->msg = NULL;
```
krizhanovsky: The same as for request processing: why should we clear ```conn->msg```?

tempesta_fw/sync_socket.h
```diff
@@ -75,7 +75,13 @@ static inline bool ss_sock_live(struct sock *sk)
 	return (sk->sk_state == TCP_ESTABLISHED);
 }
 
-static inline bool ss_can_send(struct sock *sk)
+/*
+ * Socket is in a usable state that allows processing
+ * and sending of HTTP messages. This function must
+ * be used consistently across all involved functions.
+ *
```
keshonok: Extra line, should be removed.

tempesta_fw/http.c
```diff
@@ -517,17 +562,19 @@ static void
...
+		 * is either passed through from the back-end server, or
+		 * it is generated from the cache, so unrefer all its data.
 		 */
-		if (tfw_http_adjust_resp(resp, req))
-			return;
-		tfw_connection_send(req->conn, (TfwMsg *)resp, true);
+		if (tfw_http_adjust_resp(resp, req) == 0)
+			tfw_connection_send(cli_conn, (TfwMsg *)resp, true);
+		if (req_conn_close)
+			tfw_connection_drop(cli_conn);
```
keshonok: Looks like there was a bug here after recent changes in cache processing. `req` and `resp` should be deleted here.

krizhanovsky: Yeah, +1

tempesta_fw/sock.c
```diff
@@ -457,6 +457,28 @@ ss_close(struct sock *sk)
...
+ *   called by the kernel as the socket will be closed.
+ * - If ss_do_droplink() is called first, then ss_droplink() sees that
+ *   the socket is not live and does not execute the body again.
+ *
+ * All of this ensures that there's always an extra reference to the
+ * socket from Tempesta, until Tempesta is done with the connection.
+ *
+ * XXX: The reasoning above may get invalid if Synchronous Sockets
+ * is changed in regards to the design of socket closing procedure.
+ */
```
krizhanovsky: This still doesn't explain the necessity of ```ss_droplink()``` while we have ```ss_close()```. The functions still do very similar things and having both of them looks like broken design.

tempesta_fw/sock.c
```diff
@@ -532,48 +579,61 @@ ss_tcp_process_data(struct sock *sk)
 			off--;
 		if (likely(off < skb->len)) {
 			int r, count = skb->len - off;
-			void *conn;
+			void *udata = sk->sk_user_data;
```
krizhanovsky: ```conn``` is better name than unclear ```udata```

Issue #81, opened 2015-03-20
TLS


krizhanovsky:
Must be implemented TLS support on top of latest stable release of [libressl](http://www.libressl.org/) ported to Linux kernel.

stefanstrogin:
Here are results of benchmarks of mbed tls, wolfssl and libressl.
1. ssl_read() execution time and speed of data decoding.
https://docs.google.com/spreadsheets/d/1Gg04OLxRSgYL0Y2vwqf4CHpNc8Q_lkKf_P7OnRWPYfI/edit
(unfortunately in case of libressl ssl_read decodes faster than we can read data from the socket).
2. ssl connection time:
https://docs.google.com/spreadsheets/d/1SvSjL2OiqbTEZeqTSzbcXylFC8dHwl0lO6bE95VSVjw/edit

Modificated libressl's s_client is used to connect to stud and send 16 MB of random (from /dev/urandom) data at a time (in the 1st test). Stud doesn't resend the received data neither processes it anyhow just reading it to a buffer.

krizhanovsky:
Since the difference is too large between the results, it seems the benchmarks require more observations. Probably there are communication issues (actually there are and we saw some of them) and generation of crypto-context on negotiation phase.

Also there are some [licensing issues](https://people.gnome.org/~markmc/openssl-and-the-gpl.html) and generally speaking OpenSSL (and LibreSSL corespondingly) is GPL incompatible.

stefanstrogin:
The same graphs with logarithmic scale.
1. ![ssl_read-log](https://cloud.githubusercontent.com/assets/10670399/9150600/77681718-3dea-11e5-85da-5d26b2492ea3.png)
2. ![ssl_connect-log](https://cloud.githubusercontent.com/assets/10670399/9150601/7982af2c-3dea-11e5-8870-c875811dbe94.png)

krizhanovsky:
A simple benchmark, probably based on [Stud](https://github.com/bumptech/stud), on top of tempesta/ktest framework must be implemented to compare performance of the resulting implementation against LibreSSL, GnuTLS and wolfSSL.

krizhanovsky:
[SSL session caching](https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslsessioncache) should be implemented using TDB to replicate sessions to other nodes for full connections failovering.

krizhanovsky:
[...]

PR #369, opened 2015-12-10
#81. Compilation test server as kernel module

avbelov23:


tls/programs/ssl/ssl_server.c
```diff
@@ -22,28 +23,23 @@
...
     !defined(MBEDTLS_ENTROPY_C) || !defined(MBEDTLS_SSL_TLS_C) || \
     !defined(MBEDTLS_SSL_SRV_C) || !defined(MBEDTLS_NET_C) ||     \
     !defined(MBEDTLS_RSA_C) || !defined(MBEDTLS_CTR_DRBG_C) ||    \
-    !defined(MBEDTLS_X509_CRT_PARSE_C) || !defined(MBEDTLS_FS_IO) || \
-    !defined(MBEDTLS_PEM_PARSE_C)
+    !defined(MBEDTLS_X509_CRT_PARSE_C) || !defined(MBEDTLS_PEM_PARSE_C)
 int main( void )
 {
-    mbedtls_printf("MBEDTLS_BIGNUM_C and/or MBEDTLS_CERTS_C and/or MBEDTLS_ENTROPY_C "
+    printk("MBEDTLS_BIGNUM_C and/or MBEDTLS_CERTS_C and/or MBEDTLS_ENTROPY_C "
```
krizhanovsky: Why don't you use printing level or better ```pr_info()``` function?

tls/programs/ssl/ssl_server.c
```diff
@@ -387,14 +356,25 @@ int main( void )
...
-          && MBEDTLS_FS_IO && MBEDTLS_PEM_PARSE_C */
+          && MBEDTLS_PEM_PARSE_C */
+
+static int __init
+tfw_ssl_srv_init(void)
+{
+    main();
+
+    return 0;
+}
```
krizhanovsky: This is ugly. It'd be better to just rename ```main()``` to ```tfw_ssl_srv_init()```

tls/ttls/net.c
```diff
@@ -31,142 +32,25 @@
...
-    return( 0 );
-}
+#include <linux/kernel.h>
+#include <linux/module.h>
+#include <net/sock.h>
 
 /*
  * Initialize a context
  */
 void mbedtls_net_init( mbedtls_net_context *ctx )
```
krizhanovsky: Please adjust the coding style when you touch a code. In this case the function header should look like

      void
      mbedtls_net_init(mbedtls_net_context *ctx)

and tab should be used instead of 4 spaces.

tls/ttls/net.c
```diff
@@ -486,42 +180,7 @@ int mbedtls_net_recv( void *ctx, unsigned char *buf, size_t len )
...
-        if( errno == EINTR )
-            return( MBEDTLS_ERR_SSL_WANT_READ );
-#endif
-
-        return( MBEDTLS_ERR_NET_RECV_FAILED );
-    }
-
-    /* This call will not block */
-    return( mbedtls_net_recv( ctx, buf, len ) );
+    return( 0 );
```
krizhanovsky: return doesn't need parentheses

tls/ttls/net.c
```diff
@@ -31,142 +32,25 @@
...
-        }
-
-        close( ctx->fd );
-        ret = MBEDTLS_ERR_NET_CONNECT_FAILED;
-    }
-
-    freeaddrinfo( addr_list );
-
-    return( ret );
+    return( 0 );
```
krizhanovsky: The function is totally empty, so why do we need to leave it? And why don't we need connection logic now?

Return doesn't need parentheses.

tls/ttls/net.c
```diff
@@ -174,112 +58,36 @@ int mbedtls_net_connect( mbedtls_net_context *ctx, const char *host, const char
  */
 int mbedtls_net_bind( mbedtls_net_context *ctx, const char *bind_ip, const char *port, int proto )
 {
```
krizhanovsky: Please add comment to port the logic to SS

tls/ttls/net.c
```diff
@@ -289,139 +97,36 @@ int mbedtls_net_accept( mbedtls_net_context *bind_ctx,
                         void *client_ip, size_t buf_size, size_t *ip_len )
```
krizhanovsky: Please add comment to port the logic to SS

tls/ttls/net.c
```diff
@@ -289,139 +97,36 @@ int mbedtls_net_accept( mbedtls_net_context *bind_ctx,
...
-
-    if( ret < 0 )
-    {
-        if( net_would_block( bind_ctx ) != 0 )
-            return( MBEDTLS_ERR_SSL_WANT_READ );
+    struct socket *srv_socket = bind_ctx->socket;
+    struct socket *cl_socket;
 
-        return( MBEDTLS_ERR_NET_ACCEPT_FAILED );
+    ret = sock_create(PF_INET, SOCK_STREAM, IPPROTO_TCP, &cl_socket);
```
krizhanovsky: Why do you create a new socket? It seems you should use already bound and listening socket.

tls/ttls/net.c
```diff
@@ -289,139 +97,36 @@ int mbedtls_net_accept( mbedtls_net_context *bind_ctx,
...
+    client_ctx->socket = cl_socket;
 
     return( 0 );
 }
+EXPORT_SYMBOL(mbedtls_net_accept);
 
 /*
  * Set the socket blocking or non-blocking
  */
 int mbedtls_net_set_block( mbedtls_net_context *ctx )
```
krizhanovsky: Well, we should not have blocking IO, so please remove such functions or define them as __deprecated

tls/ttls/net.c
```diff
@@ -289,139 +97,36 @@ int mbedtls_net_accept( mbedtls_net_context *bind_ctx,
...
-    !defined(EFI32)
-    u_long n = 0;
-    return( ioctlsocket( ctx->fd, FIONBIO, &n ) );
-#else
-    return( fcntl( ctx->fd, F_SETFL, fcntl( ctx->fd, F_GETFL ) & ~O_NONBLOCK ) );
-#endif
+    return( 0 );
 }
 
 int mbedtls_net_set_nonblock( mbedtls_net_context *ctx )
```
krizhanovsky: Any IO should be nonblocking, so this logic should be integrated to socket initialization routines. Meantime the function should be removed or declared as deprecated

tls/ttls/net.c
```diff
@@ -429,19 +134,7 @@ int mbedtls_net_set_nonblock( mbedtls_net_context *ctx )
  */
 void mbedtls_net_usleep( unsigned long usec )
```
krizhanovsky: When do we need sleep logic? Probably this also should be deprecated function

tls/ttls/net.c
```diff
@@ -449,36 +142,37 @@ void mbedtls_net_usleep( unsigned long usec )
...
-    ret = (int) read( fd, buf, len );
+    struct socket *sock = ((mbedtls_net_context *)ctx)->socket;
+    struct msghdr msg;
+    struct iovec iov;
+    mm_segment_t oldfs;
+    int size;
+
+    if(sock == NULL) {
+        return -1;
+    }
```
krizhanovsky: From CodingStyle:

Do not unnecessarily use braces where a single statement will do.              
        
        if (condition)
                action();

tls/ttls/net.c
```diff
@@ -529,49 +188,45 @@ int mbedtls_net_recv_timeout( void *ctx, unsigned char *buf, size_t len,
...
-            return( MBEDTLS_ERR_SSL_WANT_WRITE );
+    struct socket *sock = ((mbedtls_net_context *)ctx)->socket;
+    struct msghdr msg;
+    struct iovec iov;
+    int size;
+    mm_segment_t oldfs;
+
+    if(sock == NULL) {
+        return -1;
+    }
```
krizhanovsky: CodingStyle:

Do not unnecessarily use braces where a single statement will do.              
        
        if (condition)
                action();

tls/ttls/platform.c
```diff
@@ -37,7 +38,9 @@
 #if !defined(MBEDTLS_PLATFORM_STD_CALLOC)
 static void *platform_calloc_uninit( size_t n, size_t size )
 {
-    return vmalloc(n * size);
+    void *p = vmalloc(n * size);
+    memset(p, 0, n * size);
+    return p;
```
krizhanovsky: There is ```vzalloc()``` call. Can we use ```kmalloc()``` instead? And probably we should use ATOMIC call. Anyway, verbose comment is required here. Probably we'll have many issues with the allocation when use it in Tempesta.

tls/ttls/net.c
```diff
@@ -31,547 +32,227 @@
...
+		switch(ret) {
+#if defined EAGAIN
+		case -EAGAIN:
+#endif
+#if defined EWOULDBLOCK && EWOULDBLOCK != EAGAIN
+		case -EWOULDBLOCK:
 #endif
+			return MBEDTLS_ERR_SSL_WANT_READ;
+		}
+printk("%i\n", ret);
```
krizhanovsky: Looks like debugging garbage

tls/ttls/net.c
```diff
@@ -31,547 +32,227 @@
...
-            return( MBEDTLS_ERR_SSL_WANT_READ );
+	ret = srv_socket->ops->accept(srv_socket, cl_socket, 0);
+	if (ret < 0) {
+		switch(ret) {
+#if defined EAGAIN
+		case -EAGAIN:
+#endif
+#if defined EWOULDBLOCK && EWOULDBLOCK != EAGAIN
+		case -EWOULDBLOCK:
 #endif
```
krizhanovsky: Don't you know what is defined? Usually such ifdefs are used in crossplathform code, not in our case.

tls/ttls/net.c
```diff
@@ -31,547 +32,227 @@
...
+	set_fs(oldfs);
+
+	if(ret < 0) {
+		switch(ret) {
+#if defined EAGAIN
+		case -EAGAIN:
+#endif
+#if defined EWOULDBLOCK && EWOULDBLOCK != EAGAIN
+		case -EWOULDBLOCK:
+#endif
```
krizhanovsky: Don't you know what is defined? Usually such ifdefs are used in crossplathform code, not in our case.

tls/ttls/net.c
```diff
@@ -31,547 +32,227 @@
...
```
krizhanovsky: Don't you know what is defined? Usually such ifdefs are used in crossplathform code, not in our case.

tls/ttls/net.h
```diff
@@ -134,34 +137,6 @@ int mbedtls_net_accept( mbedtls_net_context *bind_ctx,
                         void *client_ip, size_t buf_size, size_t *ip_len );
 
 /**
- * \brief          Set the socket blocking
- *
- * \param ctx      Socket to set
- *
- * \return         0 if successful, or a non-zero error code
- */
-int mbedtls_net_set_block( mbedtls_net_context *ctx );
```
krizhanovsky: Well, when you remove some symbol, please check whether there is some of it users. Now I see

         $ find . -name \*.[ch] |xargs grep --color mbedtls_net_set_block
         ./ttls/compat-1.3.h:#define net_set_block mbedtls_net_set_block
         ./programs/ssl/ssl_server2.c:        ret = mbedtls_net_set_block( &client_fd );
         ./programs/ssl/ssl_client2.c:        ret = mbedtls_net_set_block( &server_fd );
         ./programs/ssl/ssl_client2.c:            ret = mbedtls_net_set_block( &server_fd );

```/programs/ssl/*``` are likely to be adjusted in near future, so leave them. While ```./ttls/compat-1.3.h``` seems like a header for compatibility for programs written for old versions of mbedtls, so please remove it.

krizhanovsky:
Good to merge after the small cleanups.

tls/ttls/net.c
avbelov23: Thank you!

Issue #365, opened 2015-12-05
TLS: too large stack frame


krizhanovsky:
At some circumstances TLS requires too large stack:

         [root@vps222136 tempesta]# make
         make -C tempesta_db
         make[1]: Entering directory `/root/tempesta/tempesta_db'
         make -C libtdb
         make[2]: Entering directory `/root/tempesta/tempesta_db/libtdb'
         make[2]: Nothing to be done for `all'.
         make[2]: Leaving directory `/root/tempesta/tempesta_db/libtdb'
         make -C tdbq
         make[2]: Entering directory `/root/tempesta/tempesta_db/tdbq'
         make[2]: Nothing to be done for `all'.
         make[2]: Leaving directory `/root/tempesta/tempesta_db/tdbq'
         make[1]: Leaving directory `/root/tempesta/tempesta_db'
         make -C /lib/modules/4.1.12+/build M=/root/tempesta modules
         make[1]: Entering directory `/root/linux-4.1-tfw'
           CC [M]  /root/tempesta/tls/ttls/aes.o
         /root/tempesta/tls/ttls/aes.c: In function 鈥?[01maes_gen_tables鈥?
         /root/tempesta/tls/ttls/aes.c:465:1: error: the frame size of 2064 bytes is larger than 2048 bytes [-Werror=frame-larger-than=]
          }
          ^
         cc1: all warnings being treated as errors
         make[4]: *** [/root/tempesta/tls/ttls/aes.o] Error 1
         make[3]: *** [/root/tempesta/tls/ttls] Error 2
         make[2]: *** [/root/tempesta/tls] Error 2
         make[1]: *** [_module_/root/tempesta] Error 2
         make[1]: Leaving directory `/root/linux-4.1-tfw'
         make: *** [build] Error 2

This bug is crucial since we can't build the project. Please fix it ASAP.

PR #370, opened 2015-12-10
#365. TLS: too large stack frame

avbelov23:
There is large stack frame in AES tables generation process. Disable generation.


tls/ttls/config.h
```diff
@@ -300,7 +300,7 @@
  *
  * Uncomment this macro to store the AES tables in ROM.
  */
-//#define MBEDTLS_AES_ROM_TABLES
+#define MBEDTLS_AES_ROM_TABLES
```
krizhanovsky: We don't need the config option at all, so please leave static tables only

krizhanovsky:
good to merge

Issue #371, opened 2015-12-13
Fuzzer generates invalid HTTP requests and reports them as valid


krizhanovsky:
Start Apache HTTPD on 8080 port and run bomber on the same host:

        ./tfw_bomber.sh -a 127.0.0.1:8080 -c 1 -i 1 -m 1 -t 1 -v --start

dmesg will show request

        GET http://localhost/file.html HTTP/1.0^M
        Host:localhost^M
        Accept:text/plain^M
        Accept-Language:ru^M
        Accept-Encoding:chunked^M
        Cookie:name=value^M
        X-Forwarded-For:localhost^M
        User-agent:Wget/1.13.4 (linux-gnu)^M
        Connection:keep-alive^M
        Content-Type:text/html;charset=utf-8^M
        Content-Length:10000^M
        Transfer-Encoding:chunked^M
        Transfer-Encoding:chunked^M
        Transfer-Encoding:identity^M
        Cache-Control:no-cache^M
        ^M
        2710^M
        $$$;;;;}}}};;;;$$$$RRRRZZZ?@@''@@??^^

        ^^??@@''@@??Vtvtztvt7tvie)Lx7qt?#,zv66uuyyugZ^CaZ_W{>GffknuKIOQSQOYaYOQSQOIKIPRTRPZbZlnlrlnlj1317131jlnlrlnljVX));;))}}))9.tN[}!'rky;^"{*,=,*>`9}<fS2v[bdqlsIMMQQMMYYMMQQMMIIsswwss44sCCCCCCCSSSSSSSSCCCCCCCC===tvxvt353tvaar!l3mxpmyz322uuuuGUZ^CSVh uvCbfojq')';')'}')';')'$qsqwqsq4qs?:@:?:+; ;^; ;+:?:@:?:wy0y))))>>>>))97@o:>#+[/,;^`<*,*?[;`9;,A+LQ:fdmho$$$$$$$$$$$$$$$JJJJJJJJJJJ?@@??++  ^^  ++??@@??NNRRVtvtz131ztva2[Dhzit?9,z3yyuuCCGQR^CSR_O!)GffkvmKIOIKIWIKIOIKIHJLJPJLJXJLJlnljtvtz131ztvtjlnlj

The response is treated by Apache HTTPD as invalid:

        [tfw_bomber] Server response:
        ------------------------------
        HTTP/1.1 400 Bad Request^M
        Date: Sun, 13 Dec 2015 18:23:29 GMT^M
[...]

PR #372, opened 2015-12-14
#371.  Fuzzer generates invalid HTTP requests and reports them as valid

avbelov23:


tempesta_fw/t/tfw_fuzzer.c
```diff
@@ -164,7 +164,7 @@ static struct {
 	{sizeof(content_len) / sizeof(fuzz_msg), 2, "0123456789", A_URI, 1, 1,
 		MAX_CONTENT_LENGTH_LEN},
 	/* TRANSFER_ENCODING */
-	{sizeof(transfer_encoding) / sizeof(fuzz_msg), 0, NULL, NULL, 0, 1},
+	{sizeof(transfer_encoding) / sizeof(fuzz_msg), 0, NULL, NULL, 1, 1},
```
krizhanovsky: It seems RFC allows many Transfer-Encoding headers, so the fix seems wrong.

krizhanovsky:
good to merge

Issue #353, opened 2015-11-13
Memory corruption when adding/changing Connection: header in an HTTP response.


keshonok:
The conditions are as follows:

- Current master branch at fad205a
- Back-end HTTP server is running on the same host with Tempesta.
- The kernel is 3.10.10, and has SLUB debugging turned on.
- Simple configuration file:
```
cache 0;
server 127.0.0.1:8080;
```
- Run the following command:
```
(echo -en "GET / HTTP/1.1\r\nHost: 127.0.0.1\r\nConnection: close\r\n\r\n"; sleep 1;) | nc 127.0.0.1 80
```

The expectation is that the HTTP response contains `Connection: close` header field.

However, the response appears to corrupted. Please see the trace of two consecutive runs of the above command. Pay attention to the `Connection:` header field. On first call it's partly overwritten with the next header field. On the second response `kkkkkkk` means that the memory area has been freed.

```
[root@localhost tempesta]# (echo -en "GET / HTTP/1.1\r\nHost: 127.0.0.1\r\nConnection: close\r\n\r\n"; sleep 1;) | nc 127.0.0.1 80
HTTP/1.1 403 Forbidden
Date: Fri, 13 Nov 2015 11:30:20 GMT
Server: Apache/2.4.6 (CentOS) OpenSSL/1.0.1e-fips mod_fcgid/2.3.9
Last-Modified: Tue, 17 Jun 2014 16:00:47 GMT
ETag: "1310-4fc0a3f32a9c0"
Accept-Ranges: bytes
Content-Length: 4880
Keep-Alive: timeout=5, max=100
ConContent-Type: text/html; charset=UTF-8

<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd"><html><head>
<meta http-equiv="content-type" content="text/html; charset=UTF-8">
                <title>Apache HTTP Server Test Page powered by CentOS</title>
                <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">

    <!-- Bootstrap -->
    <link href="css/bootstrap.min.css" rel="stylesheet">
    <link rel="stylesheet" href="css/open-sans.css" type="text/css" />

<style type="text/css"><!--              

body {
  font-family: "Open Sans", Helvetica, sans-serif;
  font-weight: 100;
  color: #ccc;
  background: rgba(10, 24, 55, 1);
  font-size: 16px;
}

[...]

PR #379, opened 2015-12-23
Convert linear data to page fragments (#353)

keshonok:


keshonok:
A better solution is implemented in de9f645

Issue #377, opened 2015-12-21
Remove "Keep-Alive:" header from responses that carry "Connection: close" header


keshonok:
Tempesta sits between a client and a back-end server. It attempts to keep connections to a back-end server open for prolonged time. If a client sends `Connection: close` header in a request, Tempesta replaces it with `Connection: keep-alive` header as it passes the request to the server. When response comes, `Connection: keep-alive` header from the server is changed back to `Connection: close`. However, there's also `Keep-Alive:` header in server's response that contradicts with `Connection: close` header. For instance, it could be something like this: `Keep-Alive: timeout=5, max=100`.

`Keep-Alive:` header should be removed from responses in these cases.

Related issue: #348.

krizhanovsky:
There is TODO comment in the code https://github.com/natsys/tempesta/blob/master/tempesta_fw/http.c#L552

keshonok:
Implemented in 23f7c46.

PR #381, opened 2015-12-25
Remove "Keep-Alive:" from responses with "Connection: close". (#377)

keshonok:


tempesta_fw/http.c
```diff
@@ -498,6 +498,45 @@ tfw_http_set_hdr_connection(TfwHttpMsg *hm, int conn_flg)
...
+	case TFW_HTTP_CONN_KA:
+		/*
+		 * If present, "Keep-Alive" header informs the other side
+		 * of the timeout policy for a connection. Otherwise, it's
+		 * presumed that default policy is in action.
+		 *
+		 * TODO: Add/Replace "Keep-Alive" header when Tempesta
+		 * implements connection timeout policies and the policy
+		 * for the connection differs from default policy.
+		 */
```
krizhanovsky: Well, for instance Nginx's ```keepalive_timeout``` option is very useful for DDoS protection, so it's good to introduce the option now for Tempesta. Also don't forget to updatae README.

tempesta_fw/http.c
```diff
@@ -498,6 +498,45 @@ tfw_http_set_hdr_connection(TfwHttpMsg *hm, int conn_flg)
...
+		return 0;
+	default:
+		/*
+		 * "Keep-Alive" header mandates that "Connection: keep-alive"
+		 * header in present in HTTP message. In HTTP/1.1 connections
+		 * are keep-alive by default. If we want to add "Keep-Alive"
+		 * header then "Connection: keep-alive" header must be added
+		 * as well. TFW_HTTP_CONN_KA flag will force the addition of
+		 * uConnection: keep-alive" header to HTTP message.
```
krizhanovsky: mispring: ```u``` instead of ```"``` at begin of the line

tempesta_fw/http_msg.c
```diff
@@ -527,7 +527,6 @@ tfw_http_msg_hdr_xfrm(TfwHttpMsg *hm, char *name, size_t n_len,
 			if (tfw_http_msg_grow_hdr_tbl(hm))
 				return -ENOMEM;
 		orig_hdr = &ht->tbl[hid];
-		BUG_ON(!TFW_STR_EMPTY(orig_hdr));
```
krizhanovsky: Why? Typically we can see empty TfwStr as a header for special headers directly indexed while raw header must not be empty: if we added the header, then it shouldn't be empty. Meantime, if we deleted the header previously, then why do we try to update/delete/add it again?

tempesta_fw/str.c
```diff
@@ -238,9 +244,11 @@ __cstricmpspn(const char *s1, const char *s2, int n, int stop)
...
 		c2 = tolower(*s2++);
 		if (c1 != c2)
-			return c1 < c2 ? -1 : 1;
-		if (!c1 || c1 == stop)
-			break;
+			return -1;
+		if (!c1)
+			return 0;
+		if (c1 == stop)
+			return 1;
```
krizhanovsky: Why? Actually ```stop``` plays the same role as '\0', so why  the called need to know whether ```stop``` faced or not?

krizhanovsky:
good to merge

Issue #387, opened 2015-12-28
Implement keepalive_timeout option in Tempesta


keshonok:
This option should be similar to one in Nginx. Please see the Nginx option description here: http://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_timeout.

Also, README should be updated with the description of this option.

Related issue: #100.

PR #389, opened 2015-12-30
#387. Implement keepalive_timeout option in Tempesta

avbelov23:


etc/tempesta_fw.conf
```diff
@@ -85,12 +85,24 @@
...
+# Default:
+#   keepalive_timeout 75;
+
 # TAG: listen
 # 
 # Tempesta FW listening address.
 #
 # Syntax:
-#   listen PORT | IPADDR[:PORT]
+#   listen PORT | IPADDR[:PORT] [keepalive_timeout=TIMEOUT]
```
krizhanovsky: There is no need to maintain ```keepalive_timeout``` as an argument of ```listen``` option

etc/tempesta_fw.conf
```diff
@@ -85,12 +85,24 @@
...
+#
+# Syntax:
+#   keepalive_timeout TIMEOUT
+#
+# TIMEOUT is a timeout in secounds during which a keep-alive client connection
+# will stay open on the server side for each 'listen' entry. When not specified
+# explicitly or set the zero value, keep-alive client connections disables.
+#
+# Default:
+#   keepalive_timeout 75;
```
krizhanovsky: Please also update README.md

tempesta_fw/sock_clnt.c
```diff
@@ -45,28 +45,80 @@
...
+ * @list	- An entry in the tfw_listen_socks list.
+ * @addr	- The IP address specified in the configuration.
+ * @ka_timeout	- The keep-alive timeout.
+ */
+typedef struct {
+	SsProto			proto;
+	struct sock		*sk;
+	struct list_head	list;
+	TfwAddr			addr;
+	int			ka_timeout;
```
krizhanovsky: There is no need ```ka_timeout``` on listening socket.

tempesta_fw/sock_clnt.c
```diff
@@ -45,28 +45,62 @@
  * ------------------------------------------------------------------------
  */
 
+/**
+ * TfwConnection extension for client sockets.
+ *
+ * @conn		- The base structure. Must be the first member.
+ * @keepalive_timer	- The keep-alive timer for the connection.
+ * @timeout		- The keep-alive timeout for the connection.
```
krizhanovsky: There is no field ```timeout```

tempesta_fw/sock_clnt.c
```diff
@@ -45,28 +45,62 @@
...
+			tfw_cli_conn_release(&cli_conn->conn);
+	}
+}
 
-static TfwConnection *
+static TfwCliConnection *
 tfw_cli_conn_alloc(void)
 {
-	TfwConnection *conn;
+	TfwCliConnection *cli_conn;
```
krizhanovsky: Don't rename variables in this function and at the below.

krizhanovsky:
Good to merge with small cleanups

README.md
```diff
@@ -96,6 +96,18 @@ listen 127.0.0.1:8001;
 listen [::1]:8001;
 ```
 
+### Keep-alive timeout
+
+Tempesta may use a single TCP connection to send and receive multiple HTTP requests/responses. The syntax is as follows:
+```
+keepalive_timeout TIMEOUT
+```
+`TIMEOUT` is a timeout in secounds during which a keep-alive client connection will stay open on the server side for each 'listen' entry. The zero value disables keep-alive client connections. Default value is 75.
```
keshonok: seconds

etc/tempesta_fw.conf
```diff
@@ -85,6 +85,18 @@
 #   There is a special group called "default". All "server" entries defined
 #   outside an "srv_group" added to the default group implicitly.
 
+# TAG: keepalive_timeout
+#
+# Syntax:
+#   keepalive_timeout TIMEOUT
+#
+# TIMEOUT is a timeout in secounds during which a keep-alive client connection
```
keshonok: seconds

etc/tempesta_fw.conf
```diff
@@ -85,6 +85,18 @@
...
+# will stay open on the server side for each 'listen' entry. The zero value
+# disables keep-alive client connections.
+#
```
keshonok: Here and above:

It's unclear what is `the server side`. Perhaps, this should be changed to `Tempesta`. Also, my understanding is that the option is not for `listen` entries, but rather for client connections accepted on listening ports.

tempesta_fw/sock_clnt.c
```diff
@@ -45,28 +45,62 @@
...
 static struct kmem_cache *tfw_cli_conn_cache;
+static int tfw_cli_cfg_ka_timeout = -1;
+
+static void
+tfw_sock_cli_keepalive_timer_cb(unsigned long data)
+{
+	TfwCliConnection *cli_conn = (TfwCliConnection *)data;
+	TFW_DBG("Client timeout end\n");
+	if (ss_close(cli_conn->conn.sk) == SS_OK) {
+		tfw_connection_unlink_from_sk(cli_conn->conn.sk);
```
keshonok: Copy&Paste could be dangerous. I think this code should be moved to a separate function called from here and from`tfw_sock_clnt_drop()`. Also, comments to this code should be moved to the new function from `tfw_sock_clnt_drop()`.

tempesta_fw/sock_clnt.c
```diff
@@ -451,6 +529,13 @@ TfwCfgMod tfw_sock_clnt_cfg_mod  = {
 			.allow_repeat = true,
 			.cleanup = tfw_sock_clnt_cfg_cleanup_listen
 		},
+		{
+			"keepalive_timeout",
+			"75",
+			tfw_sock_clnt_cfg_handle_keepalive,
+			.allow_repeat = true,
```
keshonok: Right now the keep-alive timeout is the one and only for all listening ports. So this option should not be allowed to berepeated in the configuration file.

tempesta_fw/sock_clnt.c
```diff
@@ -45,28 +45,62 @@
...
+ * TfwConnection extension for client sockets.
+ *
+ * @conn		- The base structure. Must be the first member.
+ * @keepalive_timer	- The keep-alive timer for the connection.
+ * @timeout		- The keep-alive timeout for the connection.
+ *
+ */
+typedef struct {
+	TfwConnection		conn;
+	struct timer_list	keepalive_timer;
```
keshonok: It would change this to `ka_timer` as the name appears to be very long. It's purely cosmetic though, so please feel free to keep your own name.

README.md
```diff
@@ -96,6 +96,18 @@ listen 127.0.0.1:8001;
 listen [::1]:8001;
 ```
 
+### Keep-alive timeout
+
+Tempesta may use a single TCP connection to send and receive multiple HTTP requests/responses. The syntax is as follows:
+```
+keepalive_timeout TIMEOUT
+```
+`TIMEOUT` is a timeout in secounds during which a keep-alive client connection will stay open on Tempesta for each 'listen' entry. The zero value disables keep-alive client connections. Default value is 75.
```
keshonok: "seconds", "in Tempesta".
"for each 'listen' entry" should be removed.

etc/tempesta_fw.conf
```diff
@@ -85,6 +85,18 @@
 #   There is a special group called "default". All "server" entries defined
 #   outside an "srv_group" added to the default group implicitly.
 
+# TAG: keepalive_timeout
+#
+# Syntax:
+#   keepalive_timeout TIMEOUT
+#
+# TIMEOUT is a timeout in secounds during which a keep-alive client connection
+# will stay open on Tempesta for each 'listen' entry. The zero value
```
keshonok: "for each 'listen' entry" should be removed.

tempesta_fw/http.c
```diff
@@ -388,15 +389,17 @@ tfw_http_conn_msg_free(TfwHttpMsg *hm)
...
+tfw_http_conn_cli_dropfree(TfwHttpMsg *hm)
 {
-	BUG_ON(!(TFW_CONN_TYPE(hmreq->conn) & Conn_Clnt));
+	BUG_ON(!(TFW_CONN_TYPE(hm->conn) & Conn_Clnt));
 
-	if (hmreq->flags & TFW_HTTP_CONN_CLOSE) {
-		if (ss_close(hmreq->conn->sk) == SS_OK)
-			tfw_sock_clnt_drop(hmreq->conn->sk);
+	if ((hm->flags & TFW_HTTP_CONN_CLOSE) ||
+	    (hm->version < TFW_HTTP_VER_11 && !(hm->flags & TFW_HTTP_CONN_KA)))
```
keshonok: It appears that this logic is already present in the correct place in a more recent version of `http.c`. Please see here: https://github.com/natsys/tempesta/blob/master/tempesta_fw/http.c#L779

tempesta_fw/sock_clnt.c
```diff
@@ -45,28 +46,78 @@
  * ------------------------------------------------------------------------
  */
 
+/**
+ * TfwConnection extension for client sockets.
+ *
+ * @conn		- The base structure. Must be the first member.
+ * @keepalive_timer	- The keep-alive timer for the connection.
```
keshonok: ka_timer

krizhanovsky:
good to merge

PR #395, opened 2016-01-15
fix #344 in tests kallsyms were replaced by including c-files

### LINKED ISSUE #344: Tests can't find necessary functions
When running, test suite searches for some function addresses with `get_sym_ptr()`. But if a function is static, it could be inlined, so test can't find it and crashes.

Steps to reproduce:
`make DEBUG=1 test`
leads to kernel crash.

One possibility is to include source files directly.

sergsever:


krizhanovsky:
good to merge

PR #397, opened 2016-01-21
Make TX work on the same CPU.

keshonok:


tempesta_fw/work_queue.h
```diff
@@ -56,6 +58,8 @@ tfw_wq_push(TfwRBQueue *wq, void *ptr, int cpu, struct irq_work *work)
 
 	if (raw_smp_processor_id() != cpu)
 		irq_work_queue_on(work, cpu);
+	else
+		raise_softirq(NET_TX_SOFTIRQ);
```
krizhanovsky: If we're in softirq, then why should we raise it again?

krizhanovsky:
good to merge

PR #405, opened 2016-01-26
Check for CPU locality in the right place in ss_do_send().

keshonok:


krizhanovsky:
good to merge

krizhanovsky:
Good to merge

Issue #385, opened 2015-12-28
Don't add requests to disconnected server's @conn


keshonok:
When connection to a back end server is dropped, server's `conn` structure is cleared in `tfw_http_conn_destruct()`. The list of requests to the back end is cleared as well. In the meanwhile, server's `conn` structure stays put as long as there are references to it. When all references are gone server's `conn` structure is completely destroyed and freed.

At the same time requests continue coming in. Server socket closing may occur in one thread at the same time as another thread attempts to send a request to the server. The socket closing thread clears `conn->msg_queue` and expect that it's empty after that. The other thread adds another request to that same queue.

We should not add requests to a inactive server connection.

Below is the trace of this bug occurence. The bug occurs here: https://github.com/natsys/tempesta/blob/41f188f74648d04e8ce9332c11a57e00b2288444/tempesta_fw/connection.c#L75

```
[50116.780780] ------------[ cut here ]------------
[50116.780782] kernel BUG at /home/natsys-lab/tfw-run/tempesta/tempesta_fw/connection.c:75!
[50116.780783] invalid opcode: 0000 [#1] SMP 
[50116.780785] Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) dccp_diag dccp tcp_diag inet_diag 9p fscache microcode ppdev pcspkr 9pnet_virtio 9pnet parport_pc parport virtio_balloon serio_raw i2c_piix4 i2c_core acpi_cpufreq uinput xfs libcrc32c ata_generic pata_acpi virtio_net virtio_pci virtio_ring virtio ata_piix floppy dm_mirror dm_region_hash dm_log dm_mod ipv6 autofs4 [last unloaded: tempesta_db]
[50116.780804] CPU: 0 PID: 6394 Comm: nginx Tainted: G           O    4.1.12-ab+ #6
[50116.780805] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS Bochs 01/01/2011
[50116.780806] task: ffff88000222c1d0 ti: ffff88008d610000 task.ti: ffff88008d610000
[50116.780806] RIP: 0010:[<ffffffffa038d17c>]  [<ffffffffa038d17c>] tfw_connection_destruct+0x3c/0x40 [tempesta_fw]
[50116.780812] RSP: 0018:ffff88013fc03a28  EFLAGS: 00010212
[...]

PR #406, opened 2016-01-28
Move tfw_connection_destruct() to correct place. (#385)

keshonok:


tempesta_fw/http.c
```diff
@@ -659,8 +661,12 @@ tfw_http_req_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
 	spin_unlock(&srv_conn->msg_qlock);
 
 	/* Send request to the server. */
-	if (tfw_connection_send(srv_conn, (TfwMsg *)req, false))
+	if (tfw_connection_send(srv_conn, (TfwMsg *)req, false)) {
+		spin_lock(&srv_conn->msg_qlock);
+		list_del(&req->msg.msg_list);
+		spin_unlock(&srv_conn->msg_qlock);
```
krizhanovsky: Why do we need to delete the message from the list? Server connections are expected to be a contention point to service many concurrent clients, so unnecessary locking is unwished...

krizhanovsky:
good to merge

PR #410, opened 2016-01-31
Perform socket closing procedure only once for each socket

keshonok:


tempesta_fw/sock.c
```diff
@@ -1018,9 +1021,14 @@ ss_tx_action(void)
...
 			bh_unlock_sock(sk);
-			sock_put(sk); /* paired with ss_send() */
 			break;
 		case SS_CLOSE:
+			if (!ss_sock_live(sk)) {
+				SS_DBG("%s: Socket inactive: sk %p\n",
+					__func__, sk);
+				bh_unlock_sock(sk);
+				break;
+			}
```
krizhanovsky: It'd be better to move the check to ```ss_do_close()``` near to check ```sk``` for NULL.

krizhanovsky:
Good to merge with small cleanup

PR #412, opened 2016-02-02
Fix the connection destruct procedure

keshonok:


tempesta_fw/http.c
```diff
@@ -608,15 +611,22 @@ tfw_http_req_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
...
-		 * it is generated from the cache, so unrefer all its data.
+		 * The request is served from cache.
+		 * Send the response as is and unrefer its data.
 		 */
-		if (tfw_http_adjust_resp(resp, req) == 0)
-			tfw_cli_conn_send(req->conn, (TfwMsg *)resp, true);
+		if (tfw_http_adjust_resp(resp, req))
+			goto resp_err;
+		if (tfw_cli_conn_send(req->conn, (TfwMsg *)resp, true))
+			goto resp_err;
```
krizhanovsky: It seems there is no sense to try to send client an error message if we couldn't send them normal message. When error message can be sent if normal message can't?

tempesta_fw/procfs.h
```diff
@@ -54,15 +56,26 @@ typedef struct {
...
+ * @entries	- The number of cache entries.
  */
 typedef struct {
-	u64	cache_hit;
-	u64	cache_miss;
+	u64	hits;
+	u64	misses;
+	u64	stores;
+	u64	evictions;
+	u64	entries;
```
krizhanovsky: There is no understanding how to mange the new counters, so there is no sense to add them now.

krizhanovsky:
Good to merge

Issue #414, opened 2016-02-04
Incorrect handling of responses to HEAD requests


keshonok:
Response to a HEAD request has the header part that is identical to that of a response to an identical GET request. The difference with a response to a GET request is that response to a HEAD request doesn't have an actual body even though the header part may indicate that it does. Please see [RFC2616 Section 9.4](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4).

Currently Tempesta doesn't know the difference and expects a body in a response to a HEAD request.

The RFC also mentions the consequences to caching that need to be taken care of.

keshonok:
Fixed with commit c5a44bd.

PR #415, opened 2016-02-07
Handle responses to HEAD method requests. (#414)

keshonok:
According to RFC2616 Section 9.4 the HEAD method is identical to GET
except that the server MUST NOT return a message-body in the response.
However, the parser expects the presense of the body when it sees that
the response header has "Content-Length" header field with specified
body length. This change makes the parser know that the response body
is not present even though the response header indicates that it does.

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -333,8 +333,8 @@ tfw_http_conn_msg_alloc(TfwConnection *conn)
 		TfwHttpReq *req;
 
 		spin_lock(&conn->msg_qlock);
-		req = (TfwHttpReq *)list_first_entry(&conn->msg_queue,
-						     TfwMsg, msg_list);
+		req = (TfwHttpReq *)list_first_entry_or_null(&conn->msg_queue,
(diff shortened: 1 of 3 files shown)


tempesta_fw/http.c
```diff
@@ -327,10 +327,19 @@ tfw_http_conn_msg_alloc(TfwConnection *conn)
...
                 TFW_INC_STAT_BH(clnt.rx_messages);
-        else
+        } else {
+		TfwHttpReq *req;
+
+		spin_lock(&conn->msg_qlock);
+		req = (TfwHttpReq *)list_first_entry(&conn->msg_queue,
+						     TfwMsg, msg_list);
+		spin_unlock(&conn->msg_qlock);
+		if (req && (req->method == TFW_HTTP_METH_HEAD))
```
krizhanovsky: ```list_first_entry()``` returns the list head instead of NULL if the list is empty. So ```req``` check is invalid.

krizhanovsky:
Good to merge after the trivial fix

Issue #419, opened 2016-02-09
[HTTP] Don't pipeline non-idempotent method requests


krizhanovsky:
[RFC 7230 6.3.2](https://tools.ietf.org/html/rfc7230#section-6.3.2) prohibits pipelining of non-idempotent method requests. So if we receive such client request, then:
1. we still should read following messages from client connection, process them and enqueue them to outgoing server connections, but don't call `ss_send()` for them immediately;
2. if we receive idempotent request from a client, we should check whether there are non-idempotent requests in the server connection waiting to be sent. Probably we should manage this using some connection flag. In this sense non-idempotent request is synchronous "barrier" for requests transmission;
3. if we receive a response from a server for non-idempotent request, then typically we have some queued idempotent requests, so after the response processing we must call `ss_send()` for waiting idempotent requests.

See also [Making HTTP Pipelining Usable on the Open Web](http://tools.ietf.org/html/draft-nottingham-http-pipeline-01).

krizhanovsky:
Consider this [recommendations](https://tools.ietf.org/html/draft-nottingham-http-pipeline-01) and the author's [blog post](https://www.mnot.net/blog/2011/07/11/what_proxies_must_do) about proper implementation of HTTP pipelining.

krizhanovsky:
Actually the definition of idempotent request is application specific(see [my presentation](http://www.slideshare.net/profyclub_ru/web-tempesta-technologies?ref=http://junior.highload.ru/2016/abstracts/2064), slides 6-8), e.g. `GET /update?a=2` could be non-idempotent while POST request with Web-search parameters can be idempotent. So a configuration options like existing `cache_methods` and `caching policy` must be introduced to specify idempotent requests. Some DDoS mitigation techniques switches on caching of responses for non-idempotent requests (e.g. we can send the same responses to non-idempotent POSTs), but we still not pipeline them to avoid consistency issues on application side.

krizhanovsky:
Please don't forget to create a task for @sergsever to create appropriate functional test. Verbose test logic description (what it should send and how verify the results) is appreciated.

krizhanovsky:
[...]

PR #423, opened 2016-03-09
Send responses in correct order. Handle non-idempotent requests. (#419)

keshonok:


tempesta_fw/connection.h
```diff
@@ -87,6 +87,7 @@ typedef struct {
 	struct list_head	msg_queue;
 	spinlock_t		msg_qlock;
 	atomic_t		refcnt;
+	unsigned long		flags;
```
krizhanovsky: Please don't forget to update comments describing data structures.

tempesta_fw/http.c
```diff
@@ -59,6 +59,9 @@ int ghprio; /* GFSM hook priority. */
 #define S_H_CONN_KA		S_F_CONNECTION S_V_CONN_KA S_CRLFCRLF
 #define S_H_CONN_CLOSE		S_F_CONNECTION S_V_CONN_CLOSE S_CRLFCRLF
 
+static void tfw_http_resp_forward(TfwHttpReq *req,
+				  TfwHttpResp *resp, bool unref_data);
```
krizhanovsky: Please avoid forward function declarations. If you think that there are program modules in the compilation unit, so you arrange functions according to the modules, then it seems the compilation unit should be separated to the program modules.

tempesta_fw/http.c
```diff
@@ -414,30 +417,45 @@ tfw_http_conn_init(TfwConnection *conn)
...
+
+/*
+ * Connection with a peer is released.
+ * Destroy connection's remaining resources.
+ */
+static void
+tfw_http_conn_release(TfwConnection *conn)
+{
+	if (TFW_CONN_TYPE(conn) & Conn_Srv)
+		tfw_http_conn_srv_release(conn);
```
krizhanovsky: Well, if we don't need to do anything for client connections, then lets rework `tfw_connection_release()` as well as all the callback calls - there is no sense to call callback just to check that we don't need to do anything.

tempesta_fw/http.c
```diff
@@ -414,30 +417,45 @@ tfw_http_conn_init(TfwConnection *conn)
...
+	struct list_head zap_queue;
+
+	TFW_DBG3("%s: conn = %p\n", __func__, srv_conn);
+	BUG_ON(!(TFW_CONN_TYPE(srv_conn) & Conn_Srv));
+
+	INIT_LIST_HEAD(&zap_queue);
+
+	spin_lock(&srv_conn->msg_qlock);
+	list_splice_tail_init(&srv_conn->msg_queue, &zap_queue);
+	spin_unlock(&srv_conn->msg_qlock);
```
krizhanovsky: Who can use `srv_conn->msg_queue`? I.e. why do we need the lock?

tempesta_fw/http.c
```diff
@@ -615,35 +634,212 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
 	return tfw_http_set_hdr_connection(m, conn_flg);
 }
 
+static inline bool
+tfw_http_req_is_idempotent(TfwHttpReq *req)
+{
+	return (req->method != TFW_HTTP_METH_POST);
```
krizhanovsky: According to the new design, the function should return flag for !NON_IDEMPOTENT flag.

tempesta_fw/http.c
```diff
@@ -615,35 +634,212 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+ * Forward request @req to server connection @srv_conn.
+ * Can be called in both SoftIRQ and kernel thread contexts.
+ */
+static void
+tfw_http_conn_req_forward(TfwConnection *srv_conn, TfwHttpReq *req)
+{
+	TFW_DBG3("%s: srv_conn = %p, req = %p\n", __func__, srv_conn, req);
+	BUG_ON(!(TFW_CONN_TYPE(srv_conn) & Conn_Srv));
+
+	spin_lock_bh(&srv_conn->msg_qlock);
```
krizhanovsky: Ok, the code is outdated: now everything is running in softirq.

tempesta_fw/http.c
```diff
@@ -615,35 +634,212 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+	if (!tfw_http_req_is_idempotent(req))
+		req->flags |= TFW_HTTP_SYNC_REQ;
```
krizhanovsky: In the design draft `TFW_HTTP_SYNC_REQ` seems the same as `NON_IDEMPOTENT`. I think the last name is better since it follows RFC name.

tempesta_fw/http.c
```diff
@@ -615,35 +634,212 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+	list_add_tail(&req->msg.fwd_list, &srv_conn->msg_queue);
+	if (srv_conn->flags & TFW_CONN_FWD_HOLD) {
+		spin_unlock_bh(&srv_conn->msg_qlock);
+		return;
```
krizhanovsky: Hm, just return? So `tfw_http_req_cache_cb()` now ends with assumption that the request was sent to server, however it actually wasn't. Who and when sends the request now?

tempesta_fw/http.c
```diff
@@ -615,35 +634,212 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+
 /*
- * Depending on results of processing of a request, either send the request
- * to an appropriate server, or return the cached response. If none of that
- * can be done for any reason, return HTTP 404 or 500 error to the client.
+ * Forward request @req to server connection @srv_conn.
+ * Can be called in both SoftIRQ and kernel thread contexts.
+ */
+static void
+tfw_http_conn_req_forward(TfwConnection *srv_conn, TfwHttpReq *req)
```
krizhanovsky: I'd prefer shorter suffix `fwd` for the group of the function - it's common for forwarding logic and much shorter.

tempesta_fw/sock_clnt.c
```diff
@@ -164,12 +164,14 @@ tfw_sock_clnt_new(struct sock *sk)
...
 
 	/* Link Tempesta with the socket and the peer. */
-	tfw_connection_revive(conn);
 	tfw_connection_link_to_sk(conn, sk);
 	tfw_connection_link_from_sk(conn, sk);
 	tfw_connection_link_peer(conn, (TfwPeer *)cli);
 	ss_set_callbacks(sk);
 
+	/* The connection is available to others now. */
+	tfw_connection_revive(conn);
```
krizhanovsky: Why did you move the refcounter initialization? When you "publish" connection to the world anything can happen with it, so we can go to improper manipulations with uninitialized refcounter.

tempesta_fw/http.c
```diff
@@ -615,35 +634,212 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+	list_for_each_entry_safe(req, tmp, &out_queue, msg.seq_list) {
+		list_del_init(&req->msg.seq_list);
+		resp = (TfwHttpResp *)req->resp;
+		/*
+		 * Verify that the client connection is not closing.
+		 * If it is, then discard @req and @resp, and return.
+		 * Remaining requests from the client will be taken
+		 * care of when the connection is released.
+		 */
+		if (!tfw_connection_nfo(cli_conn))
```
krizhanovsky: "nfo" means "Not FailOvering". Since client connections can't be failovered, but now we call `tfw_connection_put_to_death()` for them, it's good to rename the function.

tempesta_fw/http.c
```diff
@@ -615,35 +634,212 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+			goto loop_discard;
+		/*
+		 * Close the client connection in case of an error.
+		 * Otherwise, the correct order of responses may be broken.
+		 */
+		unref_data = (resp->flags & TFW_HTTP_UNREF_DATA);
+		if (tfw_cli_conn_send(cli_conn, (TfwMsg *)resp, unref_data))
+			req->flags |= TFW_HTTP_CONN_CLOSE;
```
krizhanovsky: Currently we try to send 500 error code to the client if we can't send response. (BTW it seems senseless to try to send error message if we couldn't send normal message). Why do you set connection close flag for the request now? Why don't you delete the request and close the connection as currently?

tempesta_fw/http.c
```diff
@@ -615,35 +634,212 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+
+	TFW_DBG3("%s: conn = %p\n", __func__, srv_conn);
+	BUG_ON(!(srv_conn->flags & TFW_CONN_FWD_HOLD));
+
+	INIT_LIST_HEAD(&zap_queue);
+	INIT_LIST_HEAD(&err_queue);
+
+	spin_lock_bh(&srv_conn->msg_qlock);
+	end = container_of(fwd_queue, TfwHttpReq, msg.fwd_list);
+	list_for_each_entry_safe(req, tmp, fwd_queue, msg.fwd_list) {
```
krizhanovsky: `srv_conn->msg_queue` contains messages sent to the server and waiting for response. It seems we should process new `backlog` server queue, which contains backlogged messages while there is nonidempotent message is in `msg_queue`, here.

tempesta_fw/http.c
```diff
@@ -615,35 +634,212 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+			list_del_init(&req->msg.seq_list);
+			spin_unlock_bh(&req->conn->msg_qlock);
+		}
+		tfw_http_conn_cli_dropfree((TfwHttpMsg *)req);
+	}
+
+	/* Requests that couldn't be forwarded due to an error. */
+	list_for_each_entry_safe(req, tmp, &err_queue, msg.fwd_list) {
+		list_del_init(&req->msg.fwd_list);
+		tfw_http_send_500((TfwHttpMsg *)req);
```
krizhanovsky: This is strange practice to try to send error message to client if we couldn't send normal message to them. Let's fix all the places and just drop connections.

Issue #424, opened 2016-03-10
Some @conn are not released in Tempesta if "ab" benchmark is interrupted with Ctrl-C


keshonok:
This issue is more obvious when a large number of connection is specified. The test below was interrupted with Ctrl-C after 12000 requests. The result is that 408 connections remain "active" in Tempesta. That means, internally, that these connections had not been dropped by Tempesta, and the connection structures had not been released.

```
$ ab -s 180 -k -n 20000 -c 2000 http://192.168.10.230:80/
This is ApacheBench, Version 2.3 <$Revision: 1528965 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 192.168.10.230 (be patient)
Completed 2000 requests
Completed 4000 requests
Completed 6000 requests
Completed 8000 requests
Completed 10000 requests
Completed 12000 requests
^C

Server Software:        Apache/2.4.6
Server Hostname:        192.168.10.230
Server Port:            80

Document Path:          /
Document Length:        63 bytes

Concurrency Level:      2000
Time taken for tests:   0.326 seconds
Complete requests:      12804
Failed requests:        11385
   (Connect: 0, Receive: 0, Length: 11385, Exceptions: 0)
Non-2xx responses:      11385
Keep-Alive requests:    12804
Total transferred:      1712361 bytes
HTML transferred:       89397 bytes
Requests per second:    39318.05 [#/sec] (mean)
Time per request:       50.867 [ms] (mean)
Time per request:       0.025 [ms] (mean, across all concurrent requests)
Transfer rate:          5135.01 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    1   4.7      0      22
Processing:     0   23  53.8      8     259
Waiting:        0   23  53.8      8     259
Total:          0   24  57.6      8     275

Percentage of the requests served within a certain time (ms)
  50%      8
  66%     11
  75%     14
  80%     16
  90%     30
  95%    240
  98%    267
  99%    270
 100%    275 (longest request)

# cat /proc/tempesta/perfstat 
Cache hits                              : 0
Cache misses                            : 0
Client messages received                : 13862
Client messages forwarded               : 13755
Client messages served from cache       : 0
Client messages parsing errors          : 0
Client messages filtered out            : 0
Client messages other errors            : 12255
Client connection attempts              : 1058
[...]

PR #425, opened 2016-03-10
Add client connection hook to handle connection errors. (#424)

keshonok:


tempesta_fw/sock_clnt.c
```diff
@@ -212,9 +212,30 @@ tfw_sock_clnt_drop(struct sock *sk)
...
+/*
+ * The hook is executed when a client connection is terminated due to
+ * an error of any kind.
+ */
+static int
+tfw_sock_clnt_error(struct sock *sk)
+{
+	return tfw_sock_clnt_do_drop(sk, "connection error");
+}
+
```
krizhanovsky: Please don't introduce dummy wrappers which do exactly the same things except bit different debug output. Normal closing and error closing are treated the same way. That doesn't seem right, but this is how the things work now. We'll need to rethink this...

tempesta_fw/sock_clnt.c
```diff
@@ -212,9 +212,30 @@ tfw_sock_clnt_drop(struct sock *sk)
...
 static const SsHooks tfw_sock_clnt_ss_hooks = {
 	.connection_new		= tfw_sock_clnt_new,
 	.connection_drop	= tfw_sock_clnt_drop,
+	.connection_error	= tfw_sock_clnt_error,
```
krizhanovsky: Ingress RST are processed in ```tcp_reset()``` which calls ```sk->sk_error_report()```. Isn't it better to hook the socket callback from SS? I'm confused by the fact that TCP_CLOSE state processed in our ```ss_tcp_state_change()``` can be achieved by many (I think) ways, not always when an error occured. So it seems it'd be more accurate to call ```SsHooks->connection_error``` from ```sk->sk_error_report()``` only.

PR #438, opened 2016-03-22
fix #65

### LINKED ISSUE #65: cfg: wrong position is printed in log messages in case of parser error
Scenario:
1. Take the default configuration file that consists of comments only.
2. Type few random characters (e.g. `sdafasdg`) in the middle of the file. 
3. Start Tempesta FW.

Something like this appears in the log:

```
[   10.728074] [tempesta] ERROR: syntax error
[   10.728561] [tempesta] ERROR: configuration parsing error:
[   10.728561] 
[   10.728561] #   cache_size 1048576; # 1 GiB
[   10.728561] #
[   10.728561] # Default:
[   10.728561] #   cache_size 262144;  # 256 MiB
[   10.728561] 
[   10.728561] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

It shows the wrong position in the configuration file.
That happens because the parser tries to read more tokens and eventually reaches EOF without meeting a `;` character which is a syntax error.
The position where the parser stops is the end of the buffer, so it shows some contents around this position which is not what you expect to see.

Generally this is not about EOF, the tokenizer always eats comments implicitly, so you always see a wrong position in the log message if there are comments after it.
The default configuration file contains a lot of comments, so this is an issue.

TODO:
1. Try to display an actual position where an error happens. Perhaps need to show position of the previous token, not the current one. Or a position where the current entry started, but where the parser terminated.
2. Add more info: show file/line and expected/got clauses.

sergsever:


tempesta_fw/cfg.c
```diff
@@ -1368,13 +1368,9 @@ mod_stop(TfwCfgMod *mod)
...
 {
-	const char *start = max((ps->pos - 80), ps->in);
-	int len = ps->pos - start;
+	long unsigned int len = ps->pos - ps->in;
 
-	TFW_ERR("configuration parsing error:\n"
-		"%.*s\n"
-		"^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n",
-		len, start);
+	TFW_ERR("configuration parsing error:%lu;%s\n", len, ps->e.name);
```
krizhanovsky: ```len``` is the total configuration file length - it doesn't help to locate the syntax error at all. Can we print line number instead?

Please add space after colon.

Next bad thing is that dmesg is full of warning messages:

        [  628.235453] [tempesta] ERROR: don't know how to handle: 'adfdffa'
        [  628.236993] ------------[ cut here ]------------
        [  628.238220] WARNING: CPU: 0 PID: 4495 at /root/tempesta/tempesta_fw/cfg.c:1425 tfw_cfg_parse_mods_cfg+0x10 8/0x1d0 [tempesta_fw]()
        ......
        [  628.306569] [tempesta] ERROR: configuration parsing error:43;adfdffa
        [  628.308057] ------------[ cut here ]------------
        [  628.309169] WARNING: CPU: 0 PID: 4495 at /root/tempesta/tempesta_fw/cfg.c:1373 tfw_cfg_parse_mods_cfg+0x14 2/0x1d0 [tempesta_fw]()
        ......
        [  628.364936] [tempesta] ERROR: can't parse configuration data
        [  628.366461] ------------[ cut here ]------------
        [  628.367817] WARNING: CPU: 0 PID: 4495 at /root/tempesta/tempesta_fw/cfg.c:1465 handle_sysctl_state_io+0x3d9/0x4a0 [tempesta_fw]()
        .........
        [  628.420853] [tempesta]     spec cleanup: 'server'
        [  628.422035] [tempesta] ERROR: failed to start modules
        [  628.423145] ------------[ cut here ]------------
        [  628.424612] WARNING: CPU: 0 PID: 4495 at /root/tempesta/tempesta_fw/cfg.c:1635 handle_sysctl_state_io+0x383/0x4a0 [tempesta_fw]()

Basically we have verbose enough error reporting and there is no need to the warnings with calltraces - they just pollute the log and make it hard to read.

krizhanovsky:
Good to merge

Issue #437, opened 2016-03-20
WARNING: CPU: 4 PID: 0 at arch/x86/kernel/pci-nommu.c:34 nommu_map_page+0x75/0x80()


keshonok:
```
[18478.740830] ------------[ cut here ]------------
[18478.746109] WARNING: CPU: 4 PID: 0 at arch/x86/kernel/pci-nommu.c:34 nommu_map_page+0x75/0x80()
[18478.755472] Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) nfsd auth_rpcgss oid_registry nfs_acl nfs lockd grace fscache sunrpc ast ttm intel_powerclamp drm_kms_helper coretemp crct10dif_pclmul crc32_pclmul snd_pcm snd_timer drm aesni_intel snd aes_x86_64 lrw gf128mul psmouse joydev glue_helper iTCO_wdt ablk_helper iTCO_vendor_support evdev i2c_algo_bit i2c_i801 i2c_core lpc_ich cryptd syscopyarea sysfillrect sysimgblt tpm_tis soundcore mfd_core serio_raw pcspkr i7core_edac tpm button shpchp acpi_cpufreq processor edac_core thermal_sys ipmi_watchdog ipmi_si ipmi_poweroff ipmi_devintf ipmi_msghandler autofs4 ext4 crc16 mbcache jbd2 sg sr_mod sd_mod cdrom hid_generic usbhid hid ahci ehci_pci libahci ixgbe uhci_hcd mdio libata ehci_hcd e1000e crc32c_intel
[18478.832470]  dca usbcore ptp scsi_mod usb_common pps_core [last unloaded: tempesta_db]
[18478.839978] CPU: 4 PID: 0 Comm: swapper/4 Tainted: G        W  O    4.1.12+ #1
[18478.848005] Hardware name: Quanta                 S89/QR1-890/QSSC-890  /S89/QR1-890/QSSC-890  , BIOS S89_3A10 10/21/2010
[18478.859787]  ffffffff819cfbbd ffff88007a6038d8 ffffffff8161dea5 0000000000040400
[18478.868106]  0000000000000000 ffff88007a603918 ffffffff810670aa 0000000000000000
[18478.876420]  00058000000eed79 ffff880058a09098 000000000000012d ffff880057fed2d0
[18478.884741] Call Trace:
[...]

Issue #436, opened 2016-03-20
BUG in __split_pgfrag: unable to handle kernel NULL pointer dereference


keshonok:
This has happened in one of the runs of the `ab` benchmark. Difficult to reproduce without a special test request that consists of specially crafted SKBs.
```
[52016.303192] BUG: unable to handle kernel NULL pointer dereference at 0000000000000094
[52016.311901] IP: [<ffffffffa06bccd9>] __split_pgfrag+0x5d9/0x930 [tempesta_fw]
[52016.319916] PGD 0
[52016.322761] Oops: 0000 [#1] SMP
[52016.326743] Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) nfsd auth_rpcgss oid_registry nfs_acl nfs lockd grace fscache sunrpc ast ttm drm_kms_helper intel_powerclamp drm snd_pcm coretemp crct10dif_pclmul crc32_pclmul joydev snd_timer snd aesni_intel aes_x86_64 lrw gf128mul glue_helper ablk_helper i2c_algo_bit psmouse iTCO_wdt iTCO_vendor_support evdev soundcore syscopyarea sysfillrect cryptd lpc_ich sysimgblt pcspkr mfd_core serio_raw i2c_i801 tpm_tis tpm i2c_core button acpi_cpufreq processor i7core_edac thermal_sys edac_core shpchp ipmi_watchdog ipmi_si ipmi_poweroff ipmi_devintf ipmi_msghandler autofs4 ext4 crc16 mbcache jbd2 sg sr_mod cdrom sd_mod hid_generic usbhid hid ehci_pci uhci_hcd ahci libahci ehci_hcd ixgbe e1000e mdio usbcore libata
[52016.404040]  crc32c_intel dca ptp scsi_mod usb_common pps_core [last unloaded: tempesta_db]
[52016.411965] CPU: 1 PID: 13 Comm: ksoftirqd/1 Tainted: G        W  O    4.1.12+ #1
[52016.420184] Hardware name: Quanta                 S89/QR1-890/QSSC-890  /S89/QR1-890/QSSC-890  , BIOS S89_3A10 10/21/2010
[52016.431907] task: ffff8800592aa8d0 ti: ffff8800592b0000 task.ti: ffff8800592b0000
[...]

Issue #435, opened 2016-03-19
kernel BUG at net/core/skbuff.c:2437


keshonok:
This happens on the latest Tempesta kernel with the SKB page allocator, and Tempesta version that has `unref_data` always set to `false`, even in the case where usually set to `true`. The test was over a 10G link.

The trace shows that the 

```
[52518.672355] kernel BUG at net/core/skbuff.c:2437!
[52518.677478] invalid opcode: 0000 [#1] SMP 
[52518.682041] Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) nls_utf8 fuse btrfs xor raid6_pq ntfs vfat msdos fat dm_mod nfsd auth_rpcgss oid_registry nfs_acl nfs lockd grace fscache sunrpc snd_pcm ast ttm intel_powerclamp coretemp snd_timer crct10dif_pclmul snd crc32_pclmul drm_kms_helper aesni_intel drm aes_x86_64 lrw gf128mul i2c_algo_bit glue_helper iTCO_wdt joydev psmouse soundcore ablk_helper iTCO_vendor_support cryptd i2c_i801 lpc_ich evdev i2c_core syscopyarea mfd_core serio_raw sysfillrect tpm_tis sysimgblt tpm pcspkr button acpi_cpufreq processor i7core_edac shpchp edac_core thermal_sys ipmi_watchdog ipmi_si ipmi_poweroff ipmi_devintf ipmi_msghandler autofs4 ext4 crc16 mbcache jbd2 sg sr_mod sd_mod cdrom hid_generic usbhid hid ehci_pci uhci_hcd ahci libahci ehci_hcd e1000e ixgbe libata crc32c_intel mdio dca usbcore ptp scsi_mod usb_common pps_core [last unloaded: tempesta_db]
[52518.771363] CPU: 1 PID: 0 Comm: swapper/1 Tainted: G        W  O    4.1.12+ #1
[52518.779233] Hardware name: Quanta                 S89/QR1-890/QSSC-890  /S89/QR1-890/QSSC-890  , BIOS S89_3A10 10/21/2010
[52518.790870] task: ffff8800590fa290 ti: ffff8800590fc000 task.ti: ffff8800590fc000
[52518.799054] RIP: 0010:[<ffffffff814dc650>]  [<ffffffff814dc650>] skb_copy_and_csum_bits+0x2e0/0x2f0
[...]

Issue #433, opened 2016-03-17
kernel BUG at net/core/skbuff.c:1801


keshonok:
This issue is nearly identical to #430 as it happens around a very similar piece of code, namely after an unsuccessful call to `skb_copy_bits()`. Again, that leads me to believe that there's an issue with SKB's internal integrity.

The difference from #430 is that this time a 1G network adapter was used controlled by the `e1000` driver. So at least this issue is not just about the `ixgbe` driver.

```
[47434.863552] kernel BUG at net/core/skbuff.c:1801!
[47434.868908] invalid opcode: 0000 [#1] SMP 
[47434.873699] Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) nfsd auth_rpcgss oid_registry nfs_acl nfs lockd grace fscache sunrpc intel_powerclamp coretemp kvm crct10dif_pclmul crc32_pclmul ast ttm aesni_intel aes_x86_64 drm_kms_helper snd_pcm drm joydev snd_timer snd lrw gf128mul glue_helper iTCO_wdt soundcore ablk_helper iTCO_vendor_support i2c_algo_bit cryptd psmouse lpc_ich evdev serio_raw i2c_i801 tpm_tis mfd_core i2c_core pcspkr tpm button acpi_cpufreq processor shpchp i7core_edac edac_core thermal_sys ipmi_watchdog ipmi_si ipmi_poweroff ipmi_devintf ipmi_msghandler autofs4 ext4 crc16 mbcache jbd2 sg sr_mod cdrom sd_mod hid_generic usbhid hid ehci_pci ahci uhci_hcd libahci ixgbe e1000e ehci_hcd libata dca crc32c_intel ptp scsi_mod usbcore usb_common pps_core mdio [last unloaded: tempesta_db]
[47434.955823] CPU: 6 PID: 0 Comm: swapper/6 Tainted: G           O    4.1.12+ #4
[47434.963864] Hardware name: Quanta                 S89/QR1-890/QSSC-890  /S89/QR1-890/QSSC-890  , BIOS S89_3A10 10/21/2010
[47434.975648] task: ffff88005be1a3d0 ti: ffff88005be20000 task.ti: ffff88005be20000
[47434.983983] RIP: 0010:[<ffffffff8151fdd2>]  [<ffffffff8151fdd2>] __pskb_pull_tail+0x382/0x390
[...]

PR #440, opened 2016-03-25
Process GRO-ed SKBs in a correct way.

keshonok:
This fixes kernel crashes in multiple places as noted in issues
#430, #431, #432, #433, #435, #437.

The core issue was the ss_skb_split() is not supposed to work with
GRO-ed SKBs that has data in skb_shinfo(skb)->frag_list. Splitting
of those SKBs resulted in the appearance of all kinds of invalid
SKBs that caused total havoc in the kernel. It also triggered #436
that will be fixed separately.

### DIFF
--- a/tempesta_fw/sock.c
+++ b/tempesta_fw/sock.c
@@ -447,7 +447,7 @@ ss_tcp_process_data(struct sock *sk)
 {
 	bool tcp_fin, droplink = true;
 	int processed = 0;
-	unsigned int off;
+	unsigned int off, skb_len, skb_seq;
 	struct sk_buff *skb, *tmp;
 	struct sk_buff *frag_list = NULL;
 	struct tcp_sock *tp = tcp_sk(sk);
@@ -484,11 +484,22 @@ ss_tcp_process_data(struct sock *sk)
 			goto out;
 		}
 
+		/* Save the original len and seq for reporting. */
+		skb_len = skb->len;
+		skb_seq = TCP_SKB_CB(skb)->seq;
+
+		off = tp->copied_seq - skb_seq;
+		if (TCP_SKB_CB(skb)->tcp_flags & TCPHDR_SYN)
+			off--;
+
+		/* SKB may be freed in processing. Save the flag. */
+		tcp_fin = TCP_SKB_CB(skb)->tcp_flags & TCPHDR_FIN;
+
 		/*
 		 * When GRO is used, multiple SKBs may be joined in one
 		 * big SKB. These SKBs are attached through frag_list.
 		 * Interpret the big SKB as a set of separate smaller
-		 * SKBs for processing. Move the FIN flag if necessary.
+		 * SKBs for processing.
 		 */
 		frag_list = skb_shinfo(skb)->frag_list;
 		if (frag_list) {
@@ -499,13 +510,6 @@ ss_tcp_process_data(struct sock *sk)
 			}
 			skb_shinfo(skb)->frag_list = NULL;
 		}
-
-		off = tp->copied_seq - TCP_SKB_CB(skb)->seq;
-		if (TCP_SKB_CB(skb)->tcp_flags & TCPHDR_SYN)
-			off--;
-
-		/* SKB may be freed in processing. Save the flag. */
-		tcp_fin = TCP_SKB_CB(skb)->tcp_flags & TCPHDR_FIN;
 repeat:
 		if (likely(off < skb->len)) {
 			int r, count = skb->len - off;
@@ -548,7 +552,7 @@ repeat:
 				off = 0;
 				skb = frag_list;
 				frag_list = frag_list->next;
-				skb->next = skb->prev = NULL;
+				skb->next = NULL;
 				goto repeat;
 			}
 			if (tcp_fin) {
@@ -556,22 +560,24 @@ repeat:
 				++tp->copied_seq;
 				goto out;
 			}
-			continue;
-		}
-		if (tcp_fin) {
+		} else if (frag_list) {
+			off -= skb->len;
+			__kfree_skb(skb);
+			skb = frag_list;
+			frag_list = frag_list->next;
+			skb->next = NULL;
+			goto repeat;
+		} else if (tcp_fin) {
 			__kfree_skb(skb);
 			SS_DBG("Link FIN: sk %p\n", sk);
 			++tp->copied_seq;
 			goto out;
+		} else {
+			SS_WARN("recvmsg bug: overlapping TCP segment at %X"
+				" seq %X rcvnxt %X len %x\n",
+				tp->copied_seq, skb_seq, tp->rcv_nxt, skb_len);
+			__kfree_skb(skb);
 		}
-		SS_WARN("recvmsg bug: overlapping TCP segment at %X"
-			" seq %X rcvnxt %X len %x\n",
-			tp->copied_seq, TCP_SKB_CB(skb)->seq,
-			tp->rcv_nxt, skb->len);
-		/* Have the remainder of frag_list freed as well. */
-		skb_shinfo(skb)->frag_list = frag_list;
-		frag_list = NULL;
-		__kfree_skb(skb);
 	}
 	droplink = false;
 out:
--- a/tempesta_fw/sock.c
+++ b/tempesta_fw/sock.c
@@ -492,15 +492,10 @@ ss_tcp_process_data(struct sock *sk)
 		 */
 		frag_list = skb_shinfo(skb)->frag_list;
 		if (frag_list) {
-			struct sk_buff *frag_skb, *last_skb = frag_list;
+			struct sk_buff *frag_skb;
 			skb_walk_frags(skb, frag_skb) {
 				frag_skb->nohdr = 0;
 				ss_skb_adjust_data_len(skb, -frag_skb->len);
-				last_skb = frag_skb;
-			}
-			if (TCP_SKB_CB(skb)->tcp_flags & TCPHDR_FIN) {
-				TCP_SKB_CB(skb)->tcp_flags &= ~TCPHDR_FIN;
-				TCP_SKB_CB(last_skb)->tcp_flags |= TCPHDR_FIN;
 			}
 			skb_shinfo(skb)->frag_list = NULL;
 		}
@@ -508,15 +503,14 @@ ss_tcp_process_data(struct sock *sk)
 		off = tp->copied_seq - TCP_SKB_CB(skb)->seq;
 		if (TCP_SKB_CB(skb)->tcp_flags & TCPHDR_SYN)
 			off--;
-repeat:
+
 		/* SKB may be freed in processing. Save the flag. */
 		tcp_fin = TCP_SKB_CB(skb)->tcp_flags & TCPHDR_FIN;
-
+repeat:
 		if (likely(off < skb->len)) {
 			int r, count = skb->len - off;
 			void *conn = rcu_dereference_sk_user_data(sk);
 
-			BUG_ON(tcp_fin && frag_list);
 			/*
 			 * If @sk_user_data is unset, then this connection
 			 * had been dropped in a parallel thread. Dropping
@@ -535,13 +529,21 @@ repeat:
 			 * is returned back to the Linux kernel.
 			 */
 			if (r < 0) {
-				SS_DBG("Bad socket %p data processing, %d\n",
-				       sk, r);
+				SS_DBG("Processing error: sk %p r %d\n", sk, r);
 				goto out; /* connection dropped */
 			}
 			tp->copied_seq += count;
 			processed += count;
 
+			if (r == SS_STOP) {
+				if (tcp_fin) {
+					SS_DBG("Data FIN: sk %p\n", sk);
+					++tp->copied_seq;
+					goto out;
+				}
+				SS_DBG("Stop processing: sk %p\n", sk);
+				break;
+			}
 			if (frag_list) {
 				off = 0;
 				skb = frag_list;
@@ -550,19 +552,15 @@ repeat:
 				goto repeat;
 			}
 			if (tcp_fin) {
-				SS_DBG("Data FIN received: sk %p\n", sk);
+				SS_DBG("Data FIN: sk %p\n", sk);
 				++tp->copied_seq;
 				goto out;
-			}
-			if (r == SS_STOP) {
-				SS_DBG("Stop processing data: sk %p\n", sk);
-				break;
 			}
 			continue;
 		}
 		if (tcp_fin) {
 			__kfree_skb(skb);
-			SS_DBG("Link FIN received: sk %p\n", sk);
+			SS_DBG("Link FIN: sk %p\n", sk);
 			++tp->copied_seq;
 			goto out;
 		}
@@ -570,7 +568,7 @@ repeat:
 			" seq %X rcvnxt %X len %x\n",
 			tp->copied_seq, TCP_SKB_CB(skb)->seq,
 			tp->rcv_nxt, skb->len);
-		/* Get the remains of frag_list freed as well. */
+		/* Have the remainder of frag_list freed as well. */
 		skb_shinfo(skb)->frag_list = frag_list;
 		frag_list = NULL;
 		__kfree_skb(skb);
--- a/tempesta_fw/sock.c
+++ b/tempesta_fw/sock.c
@@ -149,6 +149,9 @@ ss_do_send(struct sock *sk, SsSkbList *skb_list)
 		skb->ip_summed = CHECKSUM_PARTIAL;
 		skb_shinfo(skb)->gso_segs = 0;
 
+		/* @skb should be rerouted on forwarding. */
+		skb_dst_drop(skb);
+
 		/*
 		 * TODO Mark all data with PUSH to force receiver to consume
 		 * the data. Currently we do this for debugging purposes.
@@ -480,10 +483,6 @@ ss_tcp_process_data(struct sock *sk)
 			SS_WARN("Error uncloning ingress skb: sk %p\n", sk);
 			goto out;
 		}
-
-		/* @skb should be rerouted on forwarding. */
-		skb_dst_drop(skb);
-		skb->dev = NULL;
 
 		/*
 		 * When GRO is used, multiple SKBs may be joined in one
--- a/tempesta_fw/sock.c
+++ b/tempesta_fw/sock.c
@@ -432,6 +432,95 @@ ss_close_sync(struct sock *sk)
 }
 EXPORT_SYMBOL(ss_close_sync);
 
+/*
+ * Process a single SKB.
+ */
+static int
+ss_tcp_process_skb(struct sock *sk, struct sk_buff *skb, int *processed)
+{
+	bool tcp_fin;
+	int r = 0, offset, count;
+	void *conn;
+	struct sk_buff *frag_list;
+	struct tcp_sock *tp = tcp_sk(sk);
+
+	/* Calculate the offset into the SKB. */
+	offset = tp->copied_seq - TCP_SKB_CB(skb)->seq;
+	if (TCP_SKB_CB(skb)->tcp_flags & TCPHDR_SYN)
+		offset--;
+
+	/* SKB may be freed in processing. Save the flag. */
+	tcp_fin = TCP_SKB_CB(skb)->tcp_flags & TCPHDR_FIN;
+
+	/*
+	 * When GRO is used, multiple SKBs may be merged into
+	 * one big SKB. These SKBs are linked in via frag_list.
+	 * Interpret the big SKB as a set of separate smaller
+	 * SKBs for processing. Make the top SKB first in the
+	 * frag_list.
+	 */
+	frag_list = skb_shinfo(skb)->frag_list;
+	if (frag_list) {
+		struct sk_buff *f_skb;
+		skb_walk_frags(skb, f_skb) {
+			f_skb->nohdr = 0;
+			ss_skb_adjust_data_len(skb, -f_skb->len);
+		}
+		skb_shinfo(skb)->frag_list = NULL;
+	}
+	skb->next = frag_list;
+	frag_list = skb;
+
+	while (frag_list) {
+		skb = frag_list;
+		frag_list = frag_list->next;
+
+		if (unlikely(offset >= skb->len)) {
+			offset -= skb->len;
+			__kfree_skb(skb);
+			continue;
+		}
+		skb->next = NULL;
+
+		count = skb->len - offset;
+		tp->copied_seq += count;
+		*processed += count;
+
+		conn = rcu_dereference_sk_user_data(sk);
+		/*
+		 * If @sk_user_data is unset, then this connection
+		 * had been dropped in a parallel thread. Dropping
+		 * a connection is serialized with the socket lock.
+		 * The receive queue must be empty in that case,
+		 * and the execution path should never reach here.
+		 */
+		BUG_ON(conn == NULL);
+
+		r = SS_CALL(connection_recv, conn, skb, offset);
+
+		if (r < 0) {
+			SS_DBG("Processing error: sk %p r %d\n", sk, r);
+			goto out; /* connection dropped */
+		} else if (r == SS_STOP) {
+			SS_DBG("Stop processing: sk %p\n", sk);
+			break;
+		}
+	}
+	if (tcp_fin) {
+		SS_DBG("Data FIN: sk %p\n", sk);
+		++tp->copied_seq;
+		r = SS_DROP;
+	}
+out:
+	while (frag_list) {
+		skb = frag_list;
+		frag_list = frag_list->next;
+		__kfree_skb(skb);
+	}
+
+	return r;
+}
+
 /**
  * Receive data on TCP socket. Very similar to standard tcp_recvmsg().
  *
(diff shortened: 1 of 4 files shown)


tempesta_fw/sock.c
```diff
@@ -210,13 +210,13 @@ ss_send(struct sock *sk, SsSkbList *skb_list, bool pass_skb)
 		ss_skb_queue_head_init(&sw.skb_list);
 		for (skb = ss_skb_peek(skb_list); skb; skb = ss_skb_next(skb)) {
 			/* tcp_transmit_skb() will clone the skb. */
-			skb_copy = pskb_copy_for_clone(skb, GFP_ATOMIC);
-			if (!skb_copy) {
+			twin_skb = pskb_copy_for_clone(skb, GFP_ATOMIC);
```
keshonok: I renamed it as I experimented with different SKB copy functions, one of which is `skb_copy()`. That caused the name conflict. I suggest that we keep this change.

tempesta_fw/sock.c
```diff
@@ -482,15 +485,39 @@ ss_tcp_process_data(struct sock *sk)
 		skb_dst_drop(skb);
 		skb->dev = NULL;
```
keshonok: It appears that `skb->dev = NULL` is always done by the Linux stack in `tcp_v4_do_rcv()`: http://lxr.free-electrons.com/source/net/ipv4/tcp_ipv4.c?v=4.1#L1626

tempesta_fw/sock.c
```diff
@@ -482,15 +485,39 @@ ss_tcp_process_data(struct sock *sk)
 		skb_dst_drop(skb);
```
keshonok: Shouldn't this be done in the egress path, i.e. in `ss_do_send()`? Until that time it doesn't matter at all as the SKB is taken out of the Linux TCP/IP stack and orphaned.

krizhanovsky: Maybe so, I don't mind to move it to ```ss_do_send()```. The intention to add the call here was to "forget" routing information for ingress skb and process it further the same way as locally generated skbs.

tempesta_fw/sock.c
krizhanovsky: Yeah, it seems it's safe to remove the line.

tempesta_fw/sock.c
```diff
@@ -482,15 +485,39 @@ ss_tcp_process_data(struct sock *sk)
...
+			if (TCP_SKB_CB(skb)->tcp_flags & TCPHDR_FIN) {
+				TCP_SKB_CB(skb)->tcp_flags &= ~TCPHDR_FIN;
+				TCP_SKB_CB(last_skb)->tcp_flags |= TCPHDR_FIN;
+			}
+			skb_shinfo(skb)->frag_list = NULL;
+		}
+
 		off = tp->copied_seq - TCP_SKB_CB(skb)->seq;
-		if (tcp_hdr(skb)->syn)
+		if (TCP_SKB_CB(skb)->tcp_flags & TCPHDR_SYN)
```
krizhanovsky: Why the change is required?

tempesta_fw/sock.c
```diff
@@ -482,15 +485,39 @@ ss_tcp_process_data(struct sock *sk)
...
 			off--;
+repeat:
+		/* SKB may be freed in processing. Save the flag. */
+		tcp_fin = TCP_SKB_CB(skb)->tcp_flags & TCPHDR_FIN;
```
krizhanovsky: Why don't you use ```tcp_hdr(skb)->fin``` ?

tempesta_fw/sock.c
```diff
@@ -482,15 +485,39 @@ ss_tcp_process_data(struct sock *sk)
...
+		 * SKBs for processing. Move the FIN flag if necessary.
+		 */
+		frag_list = skb_shinfo(skb)->frag_list;
+		if (frag_list) {
+			struct sk_buff *frag_skb, *last_skb = frag_list;
+			skb_walk_frags(skb, frag_skb) {
+				frag_skb->nohdr = 0;
+				ss_skb_adjust_data_len(skb, -frag_skb->len);
+				last_skb = frag_skb;
+			}
```
krizhanovsky: Then ```ss_skb_process()``` shall not walk through the fragments. And this is good to do in this patch to clearly express the logic movement.

tempesta_fw/sock.c
```diff
@@ -542,13 +542,6 @@ ss_tcp_process_data(struct sock *sk)
...
 			processed += count;
 
-			if (frag_list) {
-				off = 0;
-				skb = frag_list;
-				frag_list = frag_list->next;
-				skb->next = skb->prev = NULL;
-				goto repeat;
-			}
 			if (tcp_fin) {
```
krizhanovsky: So if FIN is actually set in the last fragment we don't process all the fragments but first.

tempesta_fw/ss_skb.c
```diff
@@ -164,58 +162,65 @@ __lookup_pgfrag_room(struct sk_buff *skb, int len)
...
  * Somewhat like skb_shift().
+ * Make room for @n fragments starting with slot @from.
  *
- * Beware: @from can be equal to MAX_SKB_FRAGS if we need to insert a new
- * fragment after the last one.
+ * Beware: @from can be equal to MAX_SKB_FRAGS when a new fragment
+ * is inserted after the last one.
  */
 static int
 __extend_pgfrags(struct sk_buff *skb, struct sk_buff *pskb, int from, int n)
```
krizhanovsky: pskb is unused argument

tempesta_fw/ss_skb.c
```diff
@@ -527,70 +575,102 @@ static int
...
 
+/*
+ * Move an SKB from frag_list to the proper place in the list of SKB
+ * for an HTTP message.
+ *
+ * During the manipulations on SKB fragments a new SKB may be created
+ * to hold new or relocated fragments. The new SKB is stored in the
+ * original SKB's frag_list. After the manipulations on fragments are
+ * completed, the new SKB is inserted after the original SKB in the
+ * list of SKBs for an HTTP message.
```
krizhanovsky: The comment is unclear. What is "proper place"? Currently there are strict order for skb processing, so fragments were also at proper place in sense of processing order. It's clear from the code that skb is moved from fragment to Tempesta's list of skbs. However, it's absolutely unclear why fragments don't work for us anymore and we need to linearize the skb list.

tempesta_fw/ss_skb.c
```diff
@@ -237,6 +246,7 @@ __new_pgfrag(struct sk_buff *skb, struct sk_buff *pskb, int size, int i,
 			return -ENOMEM;
 	}
 
+	/* Make room for @shift fragments starting with slot @i. */
 	if (__extend_pgfrags(skb, pskb, i, shift)) {
```
krizhanovsky: pskb is actually unused, but still passed as an argument.

tempesta_fw/ss_skb.c
```diff
@@ -299,36 +316,43 @@ __split_linear_data(struct sk_buff *skb, struct sk_buff *pskb,
...
 			return NULL;
 	} else {
-		/*
-		 * Probably we should delete the whole second part
-		 * of the linear data or move it to page fragment.
-		 */
-		tail_len += len;
-		BUG_ON(tail_len < 0);
-		if (tail_len && __extend_pgfrags(skb, pskb, 0, 1))
+		if (__extend_pgfrags(skb, pskb, 0, 1))
```
krizhanovsky: Why do you extend fragments array if we don't need to insert a new fragment? If ```tail_len == 0```, i.e. we delete the whole tail, then this is just straightforward head length adjustment.

tempesta_fw/ss_skb.c
```diff
@@ -345,96 +369,106 @@ __split_linear_data(struct sk_buff *skb, struct sk_buff *pskb,
...
- * @return pointer to the new data room or NULL on failure.
+ * Note: @off is always within the borders of fragment @i. It can
+ * point at the start of a fragment, but it can never point at the
+ * location right after the end of a fragment. In other words, @off
+ * can be zero, but it can not be equal to the size of fragment @i.
+ *
+ * @return pointer to the room for new data, or NULL on failure.
  */
 static void *
 __split_pgfrag_add(struct sk_buff *skb, struct sk_buff *pskb, int i, int off,
```
krizhanovsky: Unused ```pskb``` again

tempesta_fw/ss_skb.c
```diff
@@ -345,96 +369,106 @@ __split_linear_data(struct sk_buff *skb, struct sk_buff *pskb,
...
 {
-	int split, dst_i = i + !!off;
-	skb_frag_t *frag_dst, *frag = &skb_shinfo(skb)->frags[i];
+	int tail_len;
 	struct sk_buff *skb_dst;
+	skb_frag_t *frag_dst, *frag = &skb_shinfo(skb)->frags[i];
 
+	SS_DBG("%s: skb [%p] i [%d] off [%d] len [%d] fragsize [%d]\n",
+		__func__, skb, i, off, len, skb_frag_size(frag));
 	BUG_ON((off < 0) || (off >= skb_frag_size(frag)));
```
krizhanovsky: Please add ```BUG_ON(off + len < skb_frag_size(frag))``` to be sure that we never call the function with ```off``` pointing to the end of fragment.

tempesta_fw/ss_skb.c
```diff
@@ -345,96 +369,106 @@ __split_linear_data(struct sk_buff *skb, struct sk_buff *pskb,
...
 
 	/*
-	 * Try to append data to a page fragment. If 'off' is zero,
-	 * then try to append data to a preceding page fragment
+	 * Try to append data to the page fragment. If @off is zero,
+	 * then try to append data to the preceding page fragment
 	 * if there's any. Go for other solutions if there's no room.
 	 */
-	if (!split && (off || i)) {
+	if (off || i) {
```
krizhanovsky: The condition is wrong: we're entering to the branch ever if we must insert data into middle of the fragment. Meantime, if shall be ```if (!off && i)```, i.e. try to append the data to previous fragment (i != 0) if we need to add data just before the fragment (!off).

Since now we don't handle case adding data just after i'th fragment, the internals of the branch must also be amended - ```frag_dst``` is always ```frag```, so we actually just should check current ```frag``` and can do this in the condition above.

tempesta_fw/ss_skb.c
```diff
@@ -345,96 +369,106 @@ __split_linear_data(struct sk_buff *skb, struct sk_buff *pskb,
...
 }
 
 /**
- * Delete @len (the value is postive now) bytes from @frag.
- * @return pointer just after deleted fragment.
+ * Delete @len (the value is positive now) bytes from @frag.
+ * @return a pointer at the location right after the deleted fragment.
  */
 static void *
 __split_pgfrag_del(struct sk_buff *skb, struct sk_buff *pskb, int i, int off,
```
krizhanovsky: unused ```pskb```

tempesta_fw/ss_skb.c
```diff
@@ -461,36 +500,49 @@ __split_pgfrag_del(struct sk_buff *skb, struct sk_buff *pskb, int i, int off,
...
 		ss_skb_adjust_data_len(skb, -len);
 		return skb_frag_address(frag);
 	}
+	/* Quick path: delete the tail part of a fragment. */
+	if (off + len == skb_frag_size(frag)) {
+		skb_frag_size_sub(frag, len);
+		ss_skb_adjust_data_len(skb, -len);
+		++i;
+		goto lookup_next_ptr;
+	}
```
krizhanovsky: Good catch!

tempesta_fw/ss_skb.c
```diff
@@ -498,16 +550,12 @@ __split_pgfrag_del(struct sk_buff *skb, struct sk_buff *pskb, int i, int off,
...
-	/* Try to find next data chunk after deleted fragment. */
+	/* Get the next fragment after the deleted fragment. */
 	if (i < si->nr_frags)
 		return skb_frag_address(&si->frags[i]);
-	skb_dst = skb_shinfo(pskb ? : skb)->frag_list;
-	if (skb_headlen(skb_dst))
-		return skb_dst->data;
-	if (skb_shinfo(skb_dst)->nr_frags)
-		return skb_frag_address(&skb_shinfo(skb_dst)->frags[0]);
-	/* ...not found, return invalid address try next skb. */
```
krizhanovsky: Hmmm, it seems you abandoned skb fragments support just to remove the piece of code.... I don't see problems with the code while really ugly ```return (void *)1``` is still there.

tempesta_fw/ss_skb.c
```diff
@@ -527,70 +575,102 @@ static int
...
+			it->ptr = __split_pgfrag(skb, pskb, i, offset, len);
 			goto done;
 		}
 	}
 
-	skb_walk_frags(skb, frag_i) {
-		int r = __skb_fragment(frag_i, skb, pspt, len, it);
-		if (r != -ENOENT)
-			return r;
-	}
```
krizhanovsky: Why do you delete the code? ```ss_skb_defraglist()``` actually doesn't move the whole fragments list to ```TFW_SKB_CB()``` list. If you'd move the list, then you need to convert skb fragment list to Tempesta's list, which use different skb fields to organize the lists. This is double list traversals (firstly to convert the list and secondly to process the list) and there is no sense for the overhead, whereas the lists are typically short.

tempesta_fw/ss_skb.c
```diff
@@ -601,19 +681,18 @@ __skb_fragment(struct sk_buff *skb, struct sk_buff *pskb, char *pspt,
...
-ss_skb_get_room(struct sk_buff *skb, char *pspt, unsigned int len, TfwStr *it)
+ss_skb_get_room(SsSkbList *head, struct sk_buff *skb,
+		char *pspt, unsigned int len, TfwStr *it)
 {
-	if (unlikely(len > PAGE_SIZE)) {
-		SS_WARN("Trying to get too large skb room, size=%u\n", len);
-		return -E2BIG;
-	}
+	int r = __skb_fragment(skb, NULL, pspt, len, it);
+	ss_skb_defraglist(head, skb);
```
krizhanovsky: It's unclear why do you do this, but anyway there is no sense to put the new fragment for ```frag_list``` and move to the skb list. It's rather better to pass ```head``` to ```__skb_fragment()```.

tempesta_fw/ss_skb.c
```diff
@@ -498,16 +550,12 @@ __split_pgfrag_del(struct sk_buff *skb, struct sk_buff *pskb, int i, int off,
...
+	/* Get the next fragment after the deleted fragment. */
 	if (i < si->nr_frags)
 		return skb_frag_address(&si->frags[i]);
-	skb_dst = skb_shinfo(pskb ? : skb)->frag_list;
-	if (skb_headlen(skb_dst))
-		return skb_dst->data;
-	if (skb_shinfo(skb_dst)->nr_frags)
-		return skb_frag_address(&skb_shinfo(skb_dst)->frags[0]);
-	/* ...not found, return invalid address try next skb. */
+	/* Return invalid address, try next skb. */
```
krizhanovsky: We spent a lot of time to understand were do we need the return value, so it'd be good to reference ```ss_skb_cutoff_data()``` in the comment.

tempesta_fw/ss_skb.c
```diff
@@ -527,70 +575,102 @@ static int
...
+	skb_frag_list_init(skb);
+}
+
+static inline void *
+ss_skb_data_address(struct sk_buff *skb)
+{
+	if (skb_headlen(skb))
+		return skb->data;
+	if (skb_shinfo(skb)->nr_frags)
+		return skb_frag_address(&skb_shinfo(skb)->frags[0]);
```
krizhanovsky: I'm not sure is it possible that skb has only ```frag_list```, but the function should either lookup the data in ```skb->frag_list``` or make an assertion that this is impossible case.

tempesta_fw/sock.c
```diff
@@ -509,38 +533,59 @@ ss_tcp_process_data(struct sock *sk)
...
+				}
+				SS_DBG("Stop processing: sk %p\n", sk);
+				break;
+			}
+			if (frag_list) {
+				off = 0;
+				skb = frag_list;
+				frag_list = frag_list->next;
+				skb->next = NULL;
+				goto repeat;
```
krizhanovsky: This is internal loop on skb fragments and you have 3 places where you do iterations over the list. The function also becomes to long, so it must be reworked: current loop over ```sk->sk_receive_queue``` stays in the function while the loop over ```skb->frag_list``` must be moved to other function and replaced by normal loop construction instead of goto. Probably the body of the internal loop is good to move to 3rd function.

tempesta_fw/ss_skb.c
keshonok: Here is the case where we know for sure that we need to insert a new fragment. There's a case where a whole tail is removed from the linear part. That case is processed before this one, not shown in this patch.

Also, `tail_len` can never be zero. If it was that would mean that we've got `pspt` pointing at the end of data in the linear area, because `tail_len` is calculated as `tail_len = (char *)skb_tail_pointer(skb) - pspt;`

krizhanovsky: Yeah, you're right

tempesta_fw/ss_skb.c
keshonok: In the manipulations on the SKB it may be necessary to create a new SKB to hold extra fragments. The new SKB is created in `__extend_pgfrags()`. The new SKB has to be inserted into the SKB list for an HTTP message in correct place, after the current SKB that is being processed. That operation requires the list head. So the list head will have to be passed through the whole call chain in `ss_skb.c` all the way down to `__extend_pgfrags()`. 

Is that what we want to do?

Instead, I temporarily store it in SKB's `frag_list` as now we don't have data there. When the control is returned back to one of the top functions in the call chain that has the pointer to the list head, the SKB is moved to the SKB list for the HTTP message.

tempesta_fw/ss_skb.c
keshonok: In this sequence of patches we don't have data in `frag_list` anymore, guuaranteed. I will put a `BUG_ON()` to assert that.

tempesta_fw/ss_skb.c
keshonok: In this sequence of patches we don't have data in `frag_list` of an SKB anymore. `frag_list` is used as the palce to temporarily store the new SKB that is created in the process of manipulating the SKB's fragments and data in those fragments. The "proper" place is after the SKB that just had been processed. That is explained in the lengthy comment above.

Please, see also the comment below: https://github.com/natsys/tempesta/pull/440/files#r58870927

tempesta_fw/ss_skb.c
keshonok: In this series of patches we don't have data in `frag_list` anymore.
`ss_skb_defraglist()` moves just the new SKB that was created while doing an add or delete operation on data in the SKB. There can be only one such SKB. There's no double list traversal.
Please see also the comments below on the same topic.

tempesta_fw/ss_skb.c
keshonok: We get here when we don't have knowledge of where the next data is. Thus this `return (void *)1`. How do you propose we can eliminate that?

krizhanovsky: I don't propose to remove the code. Rather I argue to not to remove ```frag_list`` since it has no sense and do better commenting for the code.

tempesta_fw/ss_skb.c
krizhanovsky: Ok. But don't forget assertions that there is no frag_list

tempesta_fw/ss_skb.c
krizhanovsky: That's not the right technique to pass function return value through list of fragments. The function should be removed and the new fragment must be handled properly.

tempesta_fw/ss_skb.c
krizhanovsky: Yeah, BUG_ON is very wished

tempesta_fw/ss_skb.c
krizhanovsky: I don't see a problem to pass list head by a function argument. Also you can use some return value.

tempesta_fw/ss_skb.c
```diff
@@ -186,8 +186,9 @@ __extend_pgfrags(struct sk_buff *skb, struct sk_buff *pskb, int from, int n)
 		 * Allocate a new SKB to hold @n_excess fragments.
 		 * Put it on frag_list of the current SKB.
 		 */
-		if ((nskb = alloc_skb(0, GFP_ATOMIC)) == NULL)
+		if ((nskb = alloc_skb(MAX_TCP_HEADER, GFP_ATOMIC)) == NULL)
```
krizhanovsky: Why do we need TCP header in the fragment?

tempesta_fw/ss_skb.c
```diff
@@ -164,129 +162,155 @@ __lookup_pgfrag_room(struct sk_buff *skb, int len)
...
-			psi->frag_list = skb_frag;
-		}
+		/*
+		 * Allocate a new SKB to hold @n_excess fragments.
+		 * Put it on frag_list of the current SKB.
+		 */
+		nskb = alloc_skb(MAX_TCP_HEADER, GFP_ATOMIC);
+		if (nskb == NULL)
+			return -ENOMEM;
+		skb_reserve(nskb, MAX_TCP_HEADER);
```
krizhanovsky: Why do we need to be able to place TCP header in the fragment?

tempesta_fw/ss_skb.c
```diff
@@ -207,9 +208,10 @@ __extend_pgfrags(struct sk_buff *skb, struct sk_buff *pskb, int from, int n)
 		}
 		skb_shinfo(nskb)->nr_frags += n_excess;
 	}
+
 	/* Make room for @n fragments in the SKB. */
 	n_shift = si->nr_frags - from - n_excess;
-	if (n_shift)
+	if (n_shift > 0)
```
krizhanovsky: Probably it has sense to add BUG_ON(n_shift < 0)

tempesta_fw/ss_skb.c
```diff
@@ -164,129 +162,155 @@ __lookup_pgfrag_room(struct sk_buff *skb, int len)
...
+		skb_shinfo(nskb)->nr_frags += n_excess;
+		it->skb = nskb;
 	}
 
-	memmove(&si->frags[from + n], &si->frags[from],
-		(si->nr_frags - from - n_frag) * sizeof(skb_frag_t));
-	si->nr_frags += n - n_frag;
+	/* Make room for @n fragments in the SKB. */
+	n_shift = si->nr_frags - from - n_excess;
+	if (n_shift > 0)
```
krizhanovsky: Probably it has sense to add BUG_ON(n_shift < 0)

tempesta_fw/ss_skb.c
```diff
@@ -461,136 +512,185 @@ __split_pgfrag_del(struct sk_buff *skb, struct sk_buff *pskb, int i, int off,
...
+	/*
+	 * Use @it to hold the return values from __split_pgfrag()
+	 * and __split_linear_data(). @it->ptr, @it->skb, and
+	 * @it->flags may be set to actual values. If a new SKB is
+	 * allocated, then it is stored in @it->skb. @it->ptr holds
+	 * the pointer either to data after the deleted data, or to
+	 * the area for new data. @it->flags indicates that @it->ptr
+	 * points to data in @it->skb.
+	 */
+	memset(it, 0, sizeof(TfwStr));
```
krizhanovsky: Why do you need this? If something is wrong, then we return error code. If not, then ```it```'s members aren't used.

keshonok: If the result of modification is positive (there was no error), that doesn't mean we have the relevant values in `it->skb` and `it->ptr`. `it->skb` is not set if the modification didn't require a new SKB. `it->ptr` is not set if it should be set to an area in the next SKB that is not known at that time. So, in that case, we need to see that whether the values have changed or not. For that reason we can't have garbage in `it` and need to make sure it's cleared before use.

The block of code below at the label `done` deals with interpreting the  values of `it->skb`, `it->ptr`, and it->flags`.

tempesta_fw/ss_skb.c
keshonok: It turns out that when these SKBs get into TCP/IP stack for sending, the stack expects that SKBs have sufficient room to do an `skb_pull()`. If there's no headroom, it crashes in skb_pull().

The total absence of headroom is okay for SKB fragments (from `frag_list`), but it is not in this case.

tempesta_fw/sock.c
```diff
@@ -432,6 +432,102 @@ ss_close_sync(struct sock *sk)
...
+			ss_skb_adjust_data_len(skb, -f_skb->len);
+		}
+		skb_shinfo(skb)->frag_list = NULL;
+	}
+	skb->next = frag_list;
+	frag_list = skb;
+
+	while (frag_list) {
+		skb = frag_list;
+		frag_list = frag_list->next;
```
krizhanovsky: ```skb``` is never NULL, you assign ```skb``` to ```frag_list```, which is just a loop iterator, check it for NULL and reassign ```frag_list``` to ```skb``` - looks tricky and ugly. Why not to use plain for() loop?

krizhanovsky:
Good to merge without https://github.com/natsys/tempesta/pull/440/commits/607ddc750907d753c325379010cb1679b0d66873

Issue #418, opened 2016-02-09
Kernel panic when Tempesta starts with an error in the config file


sergsever:
There's an error in `tempesta_fw.conf` file - the `listen` option is listed twice. When Tempesta starts we get kernel panic:

```
kernel BUG at /usr/src/projects/tempesta/tempesta_fw/sched/tfw_sched_rr.c:113!
invalid opcode: 0000 [#1] SMP 
Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) ppdev serio_raw pcspkr i2c_piix4 microcode parport_pc parport i2c_core acpi_cpufreq ipv6 autofs4 ata_generic pata_acpi crc32c_intel e1000 floppy ata_piix
CPU: 0 PID: 716 Comm: apache2 Tainted: G           O    4.1.12 #2
Hardware name: VMware, Inc. VMware Virtual Platform/440BX Desktop Reference Platform, BIOS 6.00 07/31/2013
task: ffff88003b09d570 ti: ffff88003b0a0000 task.ti: ffff88003b0a0000
RIP: 0010:[<ffffffffa01fc14c>]  [<ffffffffa01fc14c>] tfw_sched_rr_update_data+0xbc/0xc0 [tfw_sched_rr]
RSP: 0018:ffff88003fc03a28  EFLAGS: 00010246
RAX: ffffffffa01fc090 RBX: ffff88003b1bcb98 RCX: 0000000000000000
RDX: ffff88003fc11038 RSI: ffff88003fc0d758 RDI: ffff88003b11ea40
RBP: ffff88003fc03a28 R08: 0000000000000000 R09: 6e6f632020205d61
R10: 0000000000002698 R11: 3231203a74736f6c R12: ffff880039473800
R13: ffffffffa01ca4fc R14: ffff880037c35f50 R15: ffff88003b1b40f8
FS:  00007f00bf7f6700(0000) GS:ffff88003fc00000(0000) knlGS:0000000000000000
CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
CR2: 00007f00ccb849f0 CR3: 000000003a6d7000 CR4: 00000000000006f0
Stack:
 ffff88003fc03a38 ffffffffa01b781d ffff88003fc03aa8 ffffffffa01bacee
 ffff880039473898 2e302e3732310246 0000000000312e30 0000000000000000
[...]

PR #442, opened 2016-03-27
sock_clnt: check for duplicate listener addresses

milabs:
This refers #418 and prevents of using multiple equals `listen` directives.


krizhanovsky:
good to merge

PR #443, opened 2016-03-27
Make sure the pool pointer is valid.

keshonok:
There's an execution path where the pool pointer can be NULL. That
happens when an HTTP message structure is allocated statically or
on stack (one of TfwHttpMsg{}, TfwHttpReq{}, or TfwHttpResp{}).


krizhanovsky:
Good to merge

Issue #444, opened 2016-03-29
Block HTTP messages using LF only as line terminators instead of full CRLF


krizhanovsky:
RFC 7230 allows, but doesn't require to support, messages with lines terminated with LF only, without CR. While Tempesta FW processes lines termination with LF only as well as with full CRLF, it requires more complex (and slower) parsing and complicates HTTP messages transformation. Currently, there are bugs in cutoff logic: https://github.com/natsys/tempesta/blob/master/tempesta_fw/http_msg.c#L447 and https://github.com/natsys/tempesta/blob/master/tempesta_fw/http_msg.c#L486 - header trailers are considered 2 bytes in size.

So we should support CRLF only terminations.

krizhanovsky:
Commit https://github.com/natsys/tempesta/pull/340/commits/3ffc05323776f92d16dfa6b87c81c403f485cb15 relates to the issue.

milabs:
RFC 2616 says that:

~~~
19.3 Tolerant Applications

   Clients SHOULD be tolerant in parsing the Status-Line and servers
   tolerant when parsing the Request-Line. In particular, they SHOULD
   accept any amount of SP or HT characters between fields, even though
   only a single SP is required.

   The line terminator for message-header fields is the sequence CRLF.
   However, we recommend that applications, when parsing such headers,
   recognize a single LF as a line terminator and ignore the leading CR.
~~~

As for now, RFC 2616 obsoleted by RFC 7230. But there still many implementations which doesn't follow the strict CRLF rules. So, is it useful for the Tempesta not to support them?

krizhanovsky:
Frankly speaking I didn't see any real client using LF only as line terminator.

milabs:
There is a question about validness of the `\r...\r\r\r\n` sequence. Google and Yandex has no united meaning of how to deal with it:

1. Google
~~~
$ echo -en "GET / HTTP/1.1\r\nHost: www.google.com\r\n\r\n" | socat - TCP:www.google.com:80
HTTP/1.1 302 Found
...

$ echo -en "GET / HTTP/1.1\r\r\r\r\r\nHost: www.google.com\r\n\r\n" | socat - TCP:www.google.com:80
HTTP/1.1 302 Found
...
~~~

2. Yandex
~~~
$ echo -en "GET / HTTP/1.1\r\nHost: www.yandex.ru\r\n\r\n" | socat - TCP:www.yandex.ru:80
HTTP/1.1 302 Found

[...]

PR #451, opened 2016-04-14
Allow only LF and CRLF in while parsing HTTP

milabs:

### DIFF
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -43,8 +43,12 @@
 /* Common states. */
 enum {
 	RGen_LWS = 10000,
-	RGen_LF,
 	RGen_LWS_empty,
+	RGen_EoL, RGen__EoL,
+	RGen_Hdr,
+	RGen_Body,
+	RGen_BodyChunkExt,
+	RGen_BodyReadChunk,
 };
 
 /**
@@ -57,6 +61,31 @@ __field_finish(TfwHttpMsg *hm, TfwStr *field,
 	tfw_http_msg_field_chunk_fixup(hm, field, begin, end - begin);
 	field->flags |= TFW_STR_COMPLETE;
 }
+
+/**
+ * Check whether a character is CR or LF.
+ */
+#define IS_CR_OR_LF(c) (c == '\r' || c == '\n')
+
+/**
+ * Scans the initial @n bytes of the memory area pointed to by @s for the first
+ * occurance of EOL character.
+ */
+static inline unsigned char *
+memchreol(const unsigned char *s, size_t n)
+{
+	while (n) {
+		if (IS_CR_OR_LF(*s))
+			return (unsigned char *)s;
+		s++, n--;
+	}
+	return NULL;
+}
+
+/**
+ * Check whether a character is a whitespace (RWS/OWS/BWS according to RFC7230).
+ */
+#define IS_WS(c) (c == ' ' || c == '\t')
 
 /**
  * GCC 4.8 (CentOS 7) does a poor work on memory reusage of automatic local
@@ -440,10 +469,9 @@ __FSM_STATE(st_curr) {							\
 		/* The header value is fully parsed, move forward. */	\
 		if (saveval)						\
 			tfw_http_msg_hdr_chunk_fixup(msg, p, __fsm_n);	\
-		if (tfw_http_msg_hdr_close(msg, id))			\
-			return TFW_BLOCK;				\
-		parser->_i_st = I_0;					\
-		__FSM_MOVE_n(RGen_LF, __fsm_n + 1); /* skip \r */	\
+		parser->_i_st = RGen_EoL;				\
+		parser->_hdr_tag = id;					\
+		__FSM_MOVE_n(RGen_LWS_empty, __fsm_n); /* skip OWS */	\
 	}								\
 }
 
@@ -478,10 +506,9 @@ __FSM_STATE(st_curr) {							\
 		BUG_ON(__fsm_n < 0);					\
 		/* The header value is fully parsed, move forward. */	\
 		tfw_http_msg_hdr_chunk_fixup(msg, p, __fsm_n);		\
-		if (tfw_http_msg_hdr_close(msg, TFW_HTTP_HDR_RAW))	\
-			return TFW_BLOCK;				\
-		parser->_i_st = I_0;					\
-		__FSM_MOVE_n(RGen_LF, __fsm_n + 1); /* skip \r */	\
+		parser->_i_st = RGen_EoL;				\
+		parser->_hdr_tag = TFW_HTTP_HDR_RAW;			\
+		__FSM_MOVE_n(RGen_LWS_empty, __fsm_n); /* skip OWS */	\
 	}								\
 }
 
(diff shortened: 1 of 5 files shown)


tempesta_fw/http_parser.c
```diff
@@ -2717,10 +2667,16 @@ tfw_http_parse_resp(void *resp_data, unsigned char *data, size_t len)
 
 	__FSM_START(parser->state) {
 
+	/* ----------------    EOL    ---------------- */
+
+	#define TFW_HTTP_RESP_EOL
+	#include "http_parser_eol.h"
+	#undef TFW_HTTP_RESP_EOL
+
```
keshonok: Do we really need this in a separate file? I would say it should be in this file.

tempesta_fw/str.h
```diff
@@ -92,13 +92,17 @@
...
+ *                it should be 0 if the string has no EOL at all, 1 for LF and
+ *                2 for CRLF);
  * @flags	- 3 most significant bytes for number of chunks of compound
  * 		  string and the least significant byte for flags;
  */
 typedef struct {
 	void		*ptr;
 	struct sk_buff	*skb;
 	unsigned long	len;
+	unsigned char   eol_len;
```
keshonok: I believe something like `uint8_t` would be more appropriate.

Also, I personally prefer compact and concise names that are in line with the names of other members of the structure.  For instance, like `eolsz` or `leol`. This is totally a matter of taste, however.

tempesta_fw/http_parser_eol.h
```diff
@@ -0,0 +1,127 @@
...
+		 *   CR LF CR    -> empty-line (incomplete)
+		 *   CR LF LF    -> empty-line
+		 *   CR LF CR LF -> empty-line
+		 */
+		switch (parser->_tmp.eol) {
+		case 0xa:
+		case 0xda:
+			/* Skip headers that were not opened */
+			if (msg->parser.hdr.ptr) {
+				tfw_str_set_eol_len(&parser->hdr, 1 + !!(parser->_tmp.eol == 0xda));
```
keshonok: Perhaps, `!!(parser->_tmp.eol & 0xf0)`? Not sure which is better.

tempesta_fw/http_parser_eol.h
```diff
@@ -0,0 +1,127 @@
...
+				if (tfw_http_msg_hdr_close(msg, parser->_hdr_tag))
+					return TFW_BLOCK;
+			}
```
keshonok: An explicit `/* Fall through. */` comment is usually a good thing to have.

tempesta_fw/http_parser_eol.h
```diff
@@ -0,0 +1,127 @@
...
+		/*
+		 * Set empty-line mark only if LFxx or CRLFxx was
+		 * catched and crlf wasn't completed yet by
+		 * @__field_finish function.
+		 */
+		if ((parser->_tmp.eol & 0xf0) == 0xa0) {
+			if (!(msg->crlf.flags & TFW_STR_COMPLETE)) {
+				tfw_http_msg_set_data(msg, &msg->crlf, p);
+			}
+		}
```
keshonok: Generally, according to the code style, curly braces are not needed here.

tempesta_fw/http_parser_eol.h
```diff
@@ -0,0 +1,127 @@
...
+		switch (parser->_tmp.eol) {
+		case 0xa: case 0xda:
+			parser->_tmp.eol = 0;
+			if (!(msg->crlf.flags & TFW_STR_COMPLETE)) {
+				__FSM_JMP(RGen_Hdr); /* Header field (1) */
+			} else if (!(msg->body.flags & TFW_STR_COMPLETE)) {
+				__FSM_JMP(RGen_Body); /* Chunk of data (2) */
+			} else {
+				__FSM_JMP(RGen_Hdr); /* Header field (3) */
+			}
```
keshonok: Generally, according to the code style, curly braces are not needed here.

tempesta_fw/http_parser.c
milabs: Defining a macro reduces readability as it needs special formatting with back-slashes... So, in that case I decide not to use macro but move whole the EOL-logic into the separate file. I think that all the http-parser code needs to be moved into the separate directory, also.

tempesta_fw/http_parser_eol.h
milabs: The same.

tempesta_fw/http_parser_eol.h
```diff
@@ -0,0 +1,127 @@
...
+		}
+		return TFW_BLOCK;
+	}
+	__FSM_MOVE(RGen__EoL);
```
keshonok: I am a bit confused by having this statement here, and having `parser->_tmp.eol = 0;` in the code above. It appears that in most cases it will have to be executed twice.

tempesta_fw/str.h
milabs: Personally, I like explicit type names like `uint8_t`. But as I see it, Tempesta doesn't use such a types. So, what's the benefit of using `uint8_t` here instead of `unsigned char`?

tempesta_fw/http_parser_eol.h
milabs: `parse->_tmp.eol` is shared with `parser->_tmp.acc`. The later is used for example for parsing integers. So, we need explicitly ensure that after exit from the EoL handler such a value have been cleared to not reflect integer accumulation.

tempesta_fw/str.h
keshonok: Well, in my opinion, the type defines the use intention. The intention for this member is to represent a size, not a character.

milabs: Seems that not `long` not `char` does not assumes any intention of use. The point was that if whole the project doesn't use any 'uintXX_t` types there is no good reason to use `uint8_t` or similar in that case. Also, `char` is a plain-C type witch carries 8-bit signed/unsigned value. That's the only what we needed here.

tempesta_fw/str.h
```diff
@@ -193,6 +197,46 @@ tfw_str_updlen(TfwStr *s, const char *curr_p)
...
+	return s->len + s->eol_len;
+}
+
+/**
+ * Converts containig EOL string to string with EOL by cutting of @s
+ * length and setting @eol_len to @len. That function may be usefull
+ * with hand-crafted strings like `Cookie: SID=deadbeef\r\n`.
+ */
+static inline void
+tfw_str_fixup_eol(TfwStr *s, unsigned int len)
```
keshonok: As it stands now, the term `fixup` is used to mean a proper completion of a header field: either closing it when it is complete, or adding a new chunk of data to it. Perhaps, something like `tfw_str_strip_eol()` would be a better name? Or `tfw_str_trim_eol()`, or `tfw_str_conv_eol()`?

Also, this function looks dangerous. I think that a `BUG_ON(s->eol_len)` is justified. The real-world use case is somewhat ambiguous. If the purpose is to convert a `TfwStr{}` with `EOL` to one without `EOL`, then `len` appears to be superficial here, and it's dangerous to use it blindly. Perhaps we should really check that there are `EOL` characters at the end of the string, strip those characters, and set `s->eol_len` accordingly to correct size.

Also, technically speaking, the length of the chunks of `TfwStr{}` should be adjusted as well. That makes the matter more complicated. If `EOL` consists of two characters it may happen that they are in different chunks. If this function is intended only for plain strings, then that should be enforced via a `BUG_ON()`.

tempesta_fw/str.h
krizhanovsky: Personally I prefer classic C types, but I don't mind types like ```uint32_t``` - that's up to programmer what to use.

PR #452, opened 2016-04-14
frang: check URI length even if string is not completed

### LINKED ISSUE #445: Check URI length in Frang after each chunk of data is processed
It turns out that the check for URI length is postponed until a complete URI string is accumulated from an HTTP message. A malicious user may send a request with an extraordinary long URI (read: limitless length), and that may present a vector of DoS attack.

The check should be performed as the URI string is accumulated, i.e. after each chunk of data is processed by the parser.

~~~
	/* Ensure that length of URI is within limits. */
	__FRANG_FSM_STATE(Frang_Req_Hdr_UriLen) {
		if (frang_cfg.http_uri_len) {
			if (!(req->uri_path.flags & TFW_STR_COMPLETE)) {
				__FRANG_FSM_EXIT();
			}
			r = frang_http_uri_len(req, ra);
		}
		__FRANG_FSM_MOVE(Frang_Req_Hdr_FieldDup);
	}
~~~

https://github.com/natsys/tempesta/blob/master/tempesta_fw/classifier/frang.c#L744

milabs:
Check URI length in Frang after each chunk of data is processed.

This fixes #445.

### DIFF
--- a/tempesta_fw/classifier/frang.c
+++ b/tempesta_fw/classifier/frang.c
@@ -741,10 +741,9 @@ frang_http_req_process(FrangAcc *ra, TfwConnection *conn, struct sk_buff *skb,
 	/* Ensure that length of URI is within limits. */
 	__FRANG_FSM_STATE(Frang_Req_Hdr_UriLen) {
 		if (frang_cfg.http_uri_len) {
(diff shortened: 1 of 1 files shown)


keshonok:
Looks good to me.

krizhanovsky:
good to merge

Issue #182, opened 2015-08-12
http_parser: optimize parsing using vector instructions (AVX)


stefanstrogin:
Now HTTP parser cycles to process every single character. Improve it using AVX to process strings.

krizhanovsky:
The most crucial piece is x_HdrOther processing which should potentially scan long strings and verify allowed alphabet (currently we do simple memchr()).

krizhanovsky:
Depends on #69

milabs:
Common HTTP-headers processing fixed by 04ca164fc5fbb55b468f1e425b5e47dacbfcff08

krizhanovsky:
According to the [benchmark](http://natsys-lab.blogspot.ru/2016/09/how-to-analyze-performance-of-your.html) small HTTP flood-like request doesn't stress the parser at all, rather Linux I/O is the bottleneck. Meantime, large strings in request significantly hurt performance in HTTP parser.

A request with significantly large strings (e.g. see [example of very long User-Agent strings](http://stackoverflow.com/questions/654921/how-big-can-a-user-agent-string-geat)) like

        $ ./wrk -t 8 -c 64 -d 60s --header 'Connection: keep-alive' --header
	'Upgrade-Insecure-Requests: 1' --header 'User-Agent: Mozilla/4.0
	(compatible; MSIE 8.0; Windows NT 6.0; Trident/4.0; (R1 1.6); SLCC1;
	.NET CLR 2.0.50727; InfoPath.2; OfficeLiveConnector.1.3;
	OfficeLivePatch.0.0; .NET CLR 3.5.30729; .NET CLR 3.0.30618;
	66760635803; runtime 11.00294; 876906799603; 97880703; 669602703;
	9778063903; 877905603; 89670803; 96690803; 8878091903; 7879040603;
	999608065603; 799808803; 6666059903; 669602102803; 888809342903;
	696901603; 788907703; 887806555703; 97690214703; 66760903; 968909903;
	796802422703; 8868026703; 889803611803; 898706903; 977806408603;
	976900799903; 9897086903; 88780803; 798802301603; 9966008603;
	66760703; 97890452603; 9789064803; 96990759803; 99960107703;
	8868087903; 889801155603; 78890703; 8898070603; 89970603; 89970539603;
	89970488703; 8789007603; 87890903; 877904603; 9887077703; 798804903;
	97890264603; 967901703; 87890703; 97690420803; 79980706603;
[...]

PR #453, opened 2016-04-14
Allow only LF and CRLF in while parsing HTTP (v2)

milabs:

This is a second version of the patch set which takes into account comments passed on v1.

### DIFF
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -444,7 +444,7 @@ __hdr_del(TfwHttpMsg *hm, int hid)
 
 	/* Delete the underlying data. */
 	TFW_STR_FOR_EACH_DUP(dup, hdr, end) {
-		if (ss_skb_cutoff_data(dup, 0, 2))
+		if (ss_skb_cutoff_data(dup, 0, tfw_str_eolen(dup)))
 			return TFW_BLOCK;
 	};
 
@@ -483,7 +483,7 @@ __hdr_sub(TfwHttpMsg *hm, char *name, size_t n_len, char *val, size_t v_len,
 
 	if (!TFW_STR_DUP(orig_hdr) && hdr.len <= orig_hdr->len) {
 		/* Rewrite the header in-place. */
-		if (ss_skb_cutoff_data(orig_hdr, hdr.len, 2))
+		if (ss_skb_cutoff_data(orig_hdr, hdr.len, tfw_str_eolen(orig_hdr)))
 			return TFW_BLOCK;
 		if (tfw_strcpy(orig_hdr, &hdr))
 			return TFW_BLOCK;
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -43,8 +43,12 @@
 /* Common states. */
 enum {
 	RGen_LWS = 10000,
-	RGen_LF,
 	RGen_LWS_empty,
+	RGen_EoL, RGen__EoL,
+	RGen_Hdr,
+	RGen_Body,
+	RGen_BodyChunkExt,
+	RGen_BodyReadChunk,
 };
 
 /**
@@ -57,6 +61,31 @@ __field_finish(TfwHttpMsg *hm, TfwStr *field,
 	tfw_http_msg_field_chunk_fixup(hm, field, begin, end - begin);
 	field->flags |= TFW_STR_COMPLETE;
 }
+
+/**
+ * Check whether a character is CR or LF.
+ */
+#define IS_CR_OR_LF(c) (c == '\r' || c == '\n')
+
+/**
+ * Scans the initial @n bytes of the memory area pointed to by @s for the first
+ * occurance of EOL character.
+ */
+static inline unsigned char *
+memchreol(const unsigned char *s, size_t n)
+{
+	while (n) {
+		if (IS_CR_OR_LF(*s))
+			return (unsigned char *)s;
+		s++, n--;
+	}
+	return NULL;
+}
+
+/**
+ * Check whether a character is a whitespace (RWS/OWS/BWS according to RFC7230).
+ */
+#define IS_WS(c) (c == ' ' || c == '\t')
 
 /**
  * GCC 4.8 (CentOS 7) does a poor work on memory reusage of automatic local
@@ -440,10 +469,9 @@ __FSM_STATE(st_curr) {							\
 		/* The header value is fully parsed, move forward. */	\
 		if (saveval)						\
 			tfw_http_msg_hdr_chunk_fixup(msg, p, __fsm_n);	\
-		if (tfw_http_msg_hdr_close(msg, id))			\
-			return TFW_BLOCK;				\
-		parser->_i_st = I_0;					\
-		__FSM_MOVE_n(RGen_LF, __fsm_n + 1); /* skip \r */	\
+		parser->_i_st = RGen_EoL;				\
+		parser->_hdr_tag = id;					\
+		__FSM_MOVE_n(RGen_LWS_empty, __fsm_n); /* skip OWS */	\
 	}								\
 }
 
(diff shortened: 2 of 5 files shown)


keshonok:
Looks good to me.

tempesta_fw/http_parser.c
```diff
@@ -1455,10 +1453,16 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 
 	__FSM_START(parser->state) {
 
+	/* ----------------    EOL   ---------------- */
+
+	#define TFW_HTTP_REQ_EOL
+	#include "http_parser_eol.h"
+	#undef TFW_HTTP_REQ_EOL
```
krizhanovsky: The parser code is organized to minimize size of jumps as much as possible, e.g. you probably noted that "improbable" states are placed at the end of parsing functions. So it's better to place EOL processing code somewhere near to finishing header lines or starting a new header.

Yes, the parser code is large and clumsy, but moving only one state to other file doesn't make the code better. I though about reworking the logic to 3 files: common definitions and functions and request and response processing. I think it'll be good to do at the time of reworking current code to use SSE.

tempesta_fw/http_parser_eol.h
```diff
@@ -0,0 +1,139 @@
...
+		 *   CR LF CR    -> empty-line (incomplete)
+		 *   CR LF LF    -> empty-line
+		 *   CR LF CR LF -> empty-line
+		 */
+		switch (parser->_tmp.eol) {
+		case 0xa:
+		case 0xda:
+			/* Skip headers that were not opened */
+			if (msg->parser.hdr.ptr) {
+				tfw_str_set_eolen(&parser->hdr, 1 + !!(parser->_tmp.eol == 0xda));
```
krizhanovsky: Too long line, please fit 80 characters lines.

tempesta_fw/http_parser_eol.h
```diff
@@ -0,0 +1,139 @@
...
+		case 0xada:
+		case 0xdad:
+		case 0xdaa:
+		case 0xdada:
+			goto GoodLookingEOL;
+		}
+
+		return TFW_BLOCK;
+
+	GoodLookingEOL:
```
krizhanovsky: Usually we place labels w/o idents. Also CamelCase is used for custom types just to avoid '_t' suffixes.

tempesta_fw/http_parser_eol.h
```diff
@@ -0,0 +1,139 @@
...
+
+		/*
+		 * Set empty-line mark only if LFxx or CRLFxx was
+		 * catched and crlf wasn't completed yet by
+		 * @__field_finish function.
+		 */
+		if ((parser->_tmp.eol & 0xf0) == 0xa0) {
+			if (!(msg->crlf.flags & TFW_STR_COMPLETE)) {
+				tfw_http_msg_set_data(msg, &msg->crlf, p);
+			}
```
krizhanovsky: We don't use curly braces for one line IF bodies.

tempesta_fw/http_parser_eol.h
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		/*
+		 * Check for the empty-line (EOL + EOL) mixture here as
+		 * it can be handled immediately.
+		 */
+		switch (parser->_tmp.eol) {
+		case 0xaa: case 0xada: case 0xdaa: case 0xdada:
```
krizhanovsky: Please don't aggregate case statements into one line

tempesta_fw/http_parser_eol.h
```diff
@@ -0,0 +1,139 @@
...
+		/*
+		 * Can't make desicion right now as the EOL-sequence
+		 * is incomplete. So, let's try to do it on the next
+		 * pass.
+		 */
+
+		__FSM_MOVE(RGen__EoL);
+	}
+
+	TFW_DBG3("parser: eol %08lx +%02x(%c)\n", parser->_tmp.eol, c, isprint(c) ? c : '.');
```
krizhanovsky: Too long line

tempesta_fw/http_parser_eol.h
```diff
@@ -0,0 +1,139 @@
...
+ * this program; if not, write to the Free Software Foundation, Inc., 59
+ * Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ */
+
+__FSM_STATE(RGen_EoL) {
+	parser->_tmp.eol = 0;
+	/* Pass through */
+}
+__FSM_STATE(RGen__EoL) {
+	if (likely(IS_CR_OR_LF(c))) {
```
krizhanovsky: The macro is defined in http_parser.c and also used here and there, so the file is not a separate logic, but just a piece of large logic from http_parser.c. So I agree with @keshonok to move the file content to http_parser.c

tempesta_fw/http_parser.c
```diff
@@ -59,6 +63,31 @@ __field_finish(TfwHttpMsg *hm, TfwStr *field,
...
+static inline unsigned char *
+memchreol(const unsigned char *s, size_t n)
+{
+	while (n) {
+		if (IS_CR_OR_LF(*s))
+			return (unsigned char *)s;
+		s++, n--;
+	}
+	return NULL;
+}
```
krizhanovsky: You replaces vectorized ```memchr()``` by this singe-byte-step loop with 2 conditions and that's not good. Moreover, this is essentially bit specialized ```strcspn()```, which can be used instead. Please use [PCMPSTR](http://halobates.de/pcmpstr-js/pcmp.html).

There are also consider relating issue https://github.com/natsys/tempesta/issues/182, which probably partially fixed by the patch.

tempesta_fw/http_msg.c
```diff
@@ -483,7 +483,7 @@ __hdr_sub(TfwHttpMsg *hm, char *name, size_t n_len, char *val, size_t v_len,
 
 	if (!TFW_STR_DUP(orig_hdr) && hdr.len <= orig_hdr->len) {
```
krizhanovsky: The note https://github.com/natsys/tempesta/issues/444#issuecomment-205768369 is still not fixed

tempesta_fw/http_parser.c
```diff
@@ -563,89 +582,68 @@ do {									\
 
 #define TFW_HTTP_PARSE_BODY(prefix)					\
```
krizhanovsky: ```prefix``` is not used any more. It seems now the macro does the same for request and response parsing, so can we replace it by a function? Such that compiler can optimize the code by inlining or not the code.

tempesta_fw/http_parser.c
milabs: As for using `strcspn()` then according to [lxr](http://lxr.free-electrons.com/source/lib/string.c?v=4.1#L449) I don't really see that it has better implementation than `memchreol()`. Should I play with SSE 4.2 in that patch or can we separate that task from fixing CR/LF issues? Personally, I'll prefer not to mix different tasks in on patch set.

krizhanovsky: If ```strcspn()``` is just like ```memchreol()```, then why do we need to introduce our own function? One can find SSE implementation of ```strcspn()``` and can replace the poor kernel implementation by a better implementation of the standard functions. Meantime, it isn't obvious that ```memchreol()``` has many good SSE implementations.

I agree that it's better to do SSE optimizations in different patch. Just please make a TODO comment in the code or update #182 issue, such that we won't forget about the necessary optimizations.

tempesta_fw/http_parser.c
milabs: No sure that we can do it in a such manner as `RGen__EoL` state now needs to jump to `RGen_Body` state which will impossible to address if it will resides in a separate function. Or may be I don't understand of what you meant?

krizhanovsky: Ok, I see, you're right.

Issue #446, opened 2016-04-07
CRLF at the end of a header field can't be removed correctly in an SKB


keshonok:
All HTTP message headers are stored in a `TfwStr{}` structure in Tempesta. Tempesta doesn't store CRLF at the end of each HTTP header field. That data is considered superficial as it can be regenerated if necessary, so there's no real need to spend resources on storing it.

However, the standard allows for a single LF without CR as the end of a header field. That means the size of the ending sequence of a header field can vary between 1 or 2 bytes. We can't just remove 2 bytes at the end of a header fields like it's done now in Tempesta.

The other issue lies in the specific way that Tempesta uses to remove data from an SKB. The data is described by `TfwStr{}` structure, and it has one very specific property: each chunk of `TfwStr{}` doesn't cross the borders of a contiguous chunk of data in an SKB. A contiguous chunk of data in an SKB can be the linear part of an SKB, or a paged fragment.

However, as mentioned above, CRLF is not stored in `TfwStr{}`, so it's not guaranteed to have the same property as a chunk of `TfwStr{}`. That's exactly what happens: an attempt to remove 2 bytes of CRLF fails as CR and LF are located in different fragments of an SKB, or even in different SKBs. A different procedure needs to be developed for deletion of data by size, not by `TfwStr{}`.

This issue depends on #444.

keshonok:
Fixed via 24d4deb.

PR #456, opened 2016-04-18
Correcty remove the EOL bytes of a header field. (#446)

keshonok:
Now the EOL bytes are removed by length and therefore can cross
boundaries of fragments that make up an SKB.

### DIFF
--- a/tempesta_fw/ss_skb.c
+++ b/tempesta_fw/ss_skb.c
@@ -464,7 +464,7 @@ __split_pgfrag_add(struct sk_buff *skb, int i, int off, int len, TfwStr *it)
 		ss_skb_adjust_data_len(skb_dst, tail_len);
 	}
 
-	/* Get the SKB and the address of new data. */
+	/* Get the SKB and the address for new data. */
 	it->flags = !(i < MAX_SKB_FRAGS - 1);
 	frag_dst = it->flags ? &skb_shinfo(it->skb)->frags[0] : frag + 1;
 	it->ptr = skb_frag_address(frag_dst);
@@ -639,6 +639,9 @@ __skb_fragment(struct sk_buff *skb, struct sk_buff *pskb,
 	offset = pspt - (char *)skb->data;
 
 	if ((offset >= 0) && (offset < d_size)) {
+		int t_size = d_size - offset;
+		/* This affects only negative @len. */
+		len = (-len <= t_size) ? len : -t_size;
 		ret = __split_linear_data(skb, pspt, len, it);
 		goto done;
 	}
@@ -650,6 +653,9 @@ __skb_fragment(struct sk_buff *skb, struct sk_buff *pskb,
 		offset = pspt - (char *)skb_frag_address(frag);
 
 		if ((offset >= 0) && (offset < d_size)) {
+			int t_size = d_size - offset;
+			/* This affects only negative @len. */
+			len = (-len <= t_size) ? len : -t_size;
 			ret = __split_pgfrag(skb, i, offset, len, it);
 			goto done;
 		}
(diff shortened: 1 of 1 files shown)


tempesta_fw/ss_skb.c
```diff
@@ -711,7 +717,8 @@ __skb_fragment(struct sk_buff *skb, struct sk_buff *pskb,
 	it->flags = 0;
 	BUG_ON(it->skb == NULL);
 
-	return 0;
+	/* Return the length of processed data. */
+	return (len > 0 ? len : -len);
```
krizhanovsky: Now it seems we can process less than ```len``` bytes, shouldn't we return ```t_size``` instead?

tempesta_fw/ss_skb.c
```diff
@@ -734,7 +744,8 @@ ss_skb_get_room(struct sk_buff *skb, char *pspt, unsigned int len, TfwStr *it)
 int
 ss_skb_cutoff_data(const TfwStr *hdr, int skip, int tail)
 {
-	int r;
+	int r, f_len;
+	void *f_ptr;
```
krizhanovsky: There is no sense to introduce the variable. As many variables you have to keep in mind, as harder the code is understand to.

tempesta_fw/ss_skb.c
```diff
@@ -744,21 +755,33 @@ ss_skb_cutoff_data(const TfwStr *hdr, int skip, int tail)
...
+		r = __skb_fragment(c->skb, NULL, f_ptr, f_len, &it);
+		if (r < 0)
 			return r;
+		BUG_ON(r != -f_len);
 		skip = 0;
 	}
 
 	/* Cut off the tail. */
 	if (tail) {
+		struct sk_buff *f_skb;
```
krizhanovsky: Why do we need the variables?

tempesta_fw/ss_skb.c
```diff
@@ -744,21 +755,33 @@ ss_skb_cutoff_data(const TfwStr *hdr, int skip, int tail)
...
 		BUG_ON(it.ptr == NULL);
 		BUG_ON(it.skb == NULL);
-		r = __skb_fragment(it.skb, NULL, it.ptr, -tail, &it_zero);
-		if (r)
-			SS_WARN("Cannot delete hdr tail\n");
-		return r;
+		while (tail) {
+			f_ptr = it.ptr;
+			f_skb = it.skb;
+			it = it_zero;
```
krizhanovsky: just do ```memset()```, there is no need to use ```it_zero```

tempesta_fw/ss_skb.c
keshonok: I didn't want to recalculate `f_len` for the `BUG_ON()` below. Also, we definitely need them in the case of a tail removal (see below). So this way it's uniform and easy to understand.

tempesta_fw/ss_skb.c
keshonok: `it` is used to return the values that we may need if `EOL` is split between fragments or SKBs. For that `it` needs to be zeroed (see below). For that reason we can't use `it->ptr` and `it->skb` as functions arguments.

tempesta_fw/ss_skb.c
keshonok: Yes, but `it` values are checked on return from `__split_linear_data()` or `__split_pgfrag()`, and that depends on tha fact that `it` values **may** have changed if they had been set. To see if they have changed, we need to have the initial values for `it` member zeroed.

I will use `memset()` instead.

krizhanovsky: Ok, agreed

tempesta_fw/ss_skb.c
krizhanovsky: I only see ```f_len``` usage in ```__skb_fragment()``` and ```BUG_ON()```. We don't care about BUG statements, because if you need more performance, then you can disable the checks. So we have only one usage of the variable, and it's better to avoid the variable.

krizhanovsky:
good to merge

PR #457, opened 2016-04-18
cleanup cache entry size calculation

### LINKED ISSUE #450: HTTP response contains garbage at the end
Here are two similar requests to NGINX.

The first one is the direct request and the second one is through the Tempesta.

~~~
$ printf 'GET / HTTP/1.1\r\nHost: 192.168.88.245:80\r\nConnection: close\r\n\r\n' | socat -T1 -,ignoreeof TCP:192.168.88.245:80 | hexdump -C
00000000  48 54 54 50 2f 31 2e 31  20 32 30 30 20 4f 4b 0d  |HTTP/1.1 200 OK.|
00000010  0a 53 65 72 76 65 72 3a  20 6e 67 69 6e 78 2f 31  |.Server: nginx/1|
00000020  2e 38 2e 31 0d 0a 44 61  74 65 3a 20 53 75 6e 2c  |.8.1..Date: Sun,|
...
00000320  75 20 66 6f 72 20 75 73  69 6e 67 20 6e 67 69 6e  |u for using ngin|
00000330  78 2e 3c 2f 65 6d 3e 3c  2f 70 3e 0a 3c 2f 62 6f  |x.</em></p>.</bo|
00000340  64 79 3e 0a 3c 2f 68 74  6d 6c 3e 0a              |dy>.</html>.|
0000034c
~~~

~~~
$ printf 'GET / HTTP/1.1\r\nHost: 192.168.88.245:8080\r\nConnection: close\r\n\r\n' | socat -T1 -,ignoreeof TCP:192.168.88.245:8080 | hexdump -C
00000000  48 54 54 50 2f 31 2e 31  20 32 30 30 20 4f 4b 0d  |HTTP/1.1 200 OK.|
00000010  0a 43 6f 6e 74 65 6e 74  2d 4c 65 6e 67 74 68 3a  |.Content-Length:|
00000020  20 36 31 32 0d 0a 43 6f  6e 74 65 6e 74 2d 54 79  | 612..Content-Ty|
...
00000320  65 6d 3e 54 68 61 6e 6b  20 79 6f 75 20 66 6f 72  |em>Thank you for|
00000330  20 75 73 69 6e 67 20 6e  67 69 6e 78 2e 3c 2f 65  | using nginx.</e|
00000340  6d 3e 3c 2f 70 3e 0a 3c  2f 62 6f 64 79 3e 0a 3c  |m></p>.</body>.<|
00000350  2f 68 74 6d 6c 3e 0a 00  00 00 00 00 00 00 00 00  |/html>..........|
00000360  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00000390  00 00 00 00 00 00                                 |......|
00000396
~~~

As we can see, Tempesta response has the garbage at the end. The problem can be reproduced at least with 18afd4a5d74d32d74ca77870cb38872d4c271960

milabs:
Please, pull

### DIFF
--- a/tempesta_fw/cache.c
+++ b/tempesta_fw/cache.c
@@ -652,13 +652,18 @@ tfw_cache_build_resp_hdr(TDB *db, TfwHttpResp *resp, TfwStr *hdr,
 	for (d = 0; d < dn; ++d) {
 		s = (TfwCStr *)*p;
 		BUG_ON(s->flags);
+		TFW_STR_INIT(&dups[d]);
 		dups[d].len = s->len;
 		dups[d].flags = s->flags;
 		*p += TFW_CSTR_HDRLEN;
 		if ((r = tfw_cache_write_field(db, trec, resp, it, p, s->len,
-					       hdr)))
+					       &dups[d])))
 			return r;
 	}
+
+	hdr->ptr = dups;
+	__TFW_STR_CHUNKN_SET(hdr, dn);
+	hdr->flags |= TFW_STR_DUPLICATE;
 
 	return 0;
 }
@@ -777,6 +782,7 @@ tfw_cache_build_resp(TfwCacheEntry *ce)
 		goto err;
 
 	for (h = 0; h < ce->hdr_num; ++h) {
+		TFW_STR_INIT(resp->h_tbl->tbl + h);
 		if (tfw_cache_build_resp_hdr(db, resp, resp->h_tbl->tbl + h,
 					     &trec, &it, &p))
 			goto err;
--- a/tempesta_fw/cache.c
+++ b/tempesta_fw/cache.c
@@ -64,6 +64,9 @@ typedef struct {
 	long		body;
 } TfwCacheEntry;
 
+#define CE_BODY_SIZE							\
+	(sizeof(TfwCacheEntry) - offsetof(TfwCacheEntry, ce_body))
+
 /**
  * String header for cache entries used for TfwStr serialization.
  *
@@ -167,7 +170,7 @@ tfw_cache_entry_key_eq(TDB *db, TfwHttpReq *req, TfwCacheEntry *ce)
 	    != ce->key_len)
 		return false;
 
-	t_off = sizeof(*ce) - offsetof(TfwCacheEntry, ce_body);
+	t_off = CE_BODY_SIZE;
 	TFW_CACHE_REQ_KEYITER(c, req, u_end, h_start, h_end) {
 		if (!trec)
 			return false;
@@ -308,6 +311,9 @@ tfw_cache_strcpy_eol(char **p, TdbVRec **trec,
 
 	BUG_ON(TFW_STR_DUP(src));
 
+	if (unlikely(!src->len))
+		return 0;
+
 	TFW_STR_FOR_EACH_CHUNK(c, src, end) {
 		if ((n = tfw_cache_strcpy(p, trec, c, *tot_len)) < 0) {
 			TFW_ERR("Cache: cannot copy chunk of string\n");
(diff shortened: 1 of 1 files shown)


tempesta_fw/cache.c
```diff
@@ -457,7 +457,7 @@ __cache_add_node(TDB *db, TfwHttpResp *resp, TfwHttpReq *req,
 	size_t data_len = sizeof(*ce) + req->uri_path.len
 			  + req->h_tbl->tbl[TFW_HTTP_HDR_HOST].len
 			  + sizeof(TfwCStr) * resp->h_tbl->off
-			  + resp->msg.len;
+			  + resp->body.len;
```
krizhanovsky: Hm, why? ```data_len``` is the length of data stored in cache as TfwCacheEntry. Plus to ```ce```, descriptor of the stored cache entry, we store URI and Host header - this is cache key which is used to match stored cache entry against request. Next we store binary representation of all response headers to be able to quickly read them and generate a response ready for possible headers adjustments. ```TfwCStr``` is used to describe each header string. So so far we calculated only length of auxilary data. So we add length of whole response message including status line, all the header and message body - all the things will be stored.

milabs:
@krizhanovsky @keshonok please, review

tempesta_fw/cache.c
```diff
@@ -634,6 +634,8 @@ tfw_cache_build_resp_hdr(TDB *db, TfwHttpResp *resp, TfwStr *hdr,
 	TfwStr *dups;
 	TfwCStr *s = (TfwCStr *)*p;
 
+	TFW_STR_INIT(hdr);
```
keshonok: In my opinion, If `hdr` is intended to be in initial, zeroed state, then that should be be done in the calling function, and not here. The initial state of `hdr` should be enforced then via a proper `BUG_ON()`.

tempesta_fw/cache.c
```diff
@@ -308,6 +308,9 @@ tfw_cache_strcpy_eol(char **p, TdbVRec **trec,
 
 	BUG_ON(TFW_STR_DUP(src));
 
+	if (!src->len)
+		goto out;
```
keshonok: Why not just `return 0;` here?

milabs: Single point of return.

tempesta_fw/cache.c
```diff
@@ -460,15 +458,49 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwHttpReq *req,
 	return 0;
 }
 
+static size_t
+__cache_entry_size(TfwHttpResp *resp, TfwHttpReq *req)
+{
+	long n;
+	size_t size = sizeof(TfwCacheEntry) - offsetof(TfwCacheEntry, ce_body);
```
keshonok: I would say that this (and the above occurrence of the same statement) should be replaced with a macro or an inline function (I prefer the latter, but it's just me).

tempesta_fw/cache.c
keshonok: Can't that that it makes sense here. The thread of execution is finished here, so it's best to return right here instead of mixing it with the other thread of execution. The two threads of execution don't have any common code.

tempesta_fw/cache.c
```diff
@@ -308,6 +311,9 @@ tfw_cache_strcpy_eol(char **p, TdbVRec **trec,
 
 	BUG_ON(TFW_STR_DUP(src));
 
+	if (!src->len)
```
krizhanovsky: It's good to use ```unlikely()`` for the check

krizhanovsky:
good to merge

PR #458, opened 2016-04-19
Use pskb throughout ss_skb.c to set it->skb and it->ptr in place

keshonok:


tempesta_fw/ss_skb.c
```diff
@@ -313,8 +368,17 @@ __split_linear_data(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
 	 */
 	if (unlikely((len < 0) && (tail_len == -len))) {
 		ss_skb_put(skb, len);
-		if (skb_is_nonlinear(skb))
+		if (skb_is_nonlinear(skb)) {
 			it->ptr = skb_frag_address(&skb_shinfo(skb)->frags[0]);
+			it->skb = pskb ? pskb : skb;
+		} else {
+			struct sk_buff *dskb = __next_skb_frag(skb, pskb);
```
krizhanovsky: You've just checked ```pskb``` in ```__next_skb_frag()``` and checks it again here. That's clumsy. Moreover exactly the same piece of code is copy-pasted in ```__split_pgfrag_del()```. Thus the piece of code must be inline function not using ```__next_skb_frag()```, but rather accurately implementing the logic with minimum branches. Also, ```a ? a : b``` usually is expressed as ```a ? : b```.

tempesta_fw/ss_skb.c
```diff
@@ -262,7 +305,7 @@ __new_pgfrag(struct sk_buff *skb, int size, int i, int shift, TfwStr *it)
 	 */
 	if (i == MAX_SKB_FRAGS) {
 		i = 0;
-		skb = it->skb;
+		skb = __next_skb_frag(skb, pskb);
```
krizhanovsky: So instead of one write in ```__extend_pgfrags()``` and one read here you do 3 writes and 1 branch in ```__insert_skb_frag()``` and 1 read with a branch here. I don't think the change makes the code better.

tempesta_fw/ss_skb.c
```diff
@@ -262,7 +314,7 @@ __new_pgfrag(struct sk_buff *skb, int size, int i, int shift, TfwStr *it)
...
```
krizhanovsky: So instead of one write in __extend_pgfrags() and one read here you do 3 writes and 1 branch in ```__insert_skb_frag()``` called in ```__extend_pgfrags()``` and 1 read with a branch here. I don't think the change makes the code better.

tempesta_fw/ss_skb.c
```diff
@@ -313,8 +368,17 @@ __split_linear_data(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
...
 	if (unlikely((len < 0) && (tail_len == -len))) {
 		ss_skb_put(skb, len);
-		if (skb_is_nonlinear(skb))
+		if (skb_is_nonlinear(skb)) {
 			it->ptr = skb_frag_address(&skb_shinfo(skb)->frags[0]);
+			it->skb = pskb ? pskb : skb;
+		} else {
+			struct sk_buff *dskb = __next_skb_frag(skb, pskb);
+			if (dskb)
+				it->skb = pskb ? pskb : skb;
```
krizhanovsky: The change is very strange. ```pskb``` meand "parent skb", while ```it->skb``` is actullay ```skb``` owning data where ```it->ptr``` points to. Meantime you always write top level (parent) ```skb``` to ```it->skb```.

tempesta_fw/ss_skb.c
```diff
@@ -602,9 +665,10 @@ static int
 __skb_fragment(struct sk_buff *skb, struct sk_buff *pskb,
 	       char *pspt, int len, TfwStr *it)
```
krizhanovsky: ```__skb_fragment()``` is always called with ```pskb == NULL```, the only one case where it's called with non-NULL ```pskb``` is recursive call. Isn't it better to introduce ```skb_fragment()``` which doesn't have ```pskb``` argument at all and just calls ```__skb_fragment()``` with NULL as ```pskb```?

tempesta_fw/ss_skb.c
```diff
@@ -160,6 +160,49 @@ __lookup_pgfrag_room(struct sk_buff *skb, int len)
...
 }
 
+/*
+ * Insert the new SKB at the correct location in the list of fragments
+ * of the parent SKB. If @skb is one of the fragments, then insert the
+ * new SKB after @skb in the frag_list. Otherwise, insert the new SKB
+ * as the first fragment.
+ */
+static void
+__insert_skb_frag(struct sk_buff *skb,
```
krizhanovsky: Why do you break line after the first argument?

tempesta_fw/ss_skb.c
```diff
@@ -226,16 +226,30 @@ __extend_pgfrags(struct sk_buff *skb, struct sk_buff *pskb,
...
-		if (nskb == NULL)
-			return -ENOMEM;
 		/*
 		 * The number of page fragments that don't fit in the SKB
 		 * after the room is prepared for @n page fragments.
 		 */
 		n_excess = si->nr_frags + n - MAX_SKB_FRAGS;
 
+		/*
+		 * Use the next SKB fragment if there room there for
```
krizhanovsky: "if there room there for" -> "if there is room for"?

tempesta_fw/ss_skb.c
```diff
@@ -183,17 +225,31 @@ __extend_pgfrags(struct sk_buff *skb, int from, int n, TfwStr *it)
...
-			return -ENOMEM;
-
 		/*
 		 * The number of page fragments that don't fit in the SKB
 		 * after the room is prepared for @n page fragments.
 		 */
 		n_excess = si->nr_frags + n - MAX_SKB_FRAGS;
 
+		/*
+		 * Use the next SKB fragment if there room there for
```
krizhanovsky: "if there room there for" -> "if there is room for"?

tempesta_fw/ss_skb.c
```diff
@@ -183,17 +225,31 @@ __extend_pgfrags(struct sk_buff *skb, int from, int n, TfwStr *it)
...
+		 * @n_excess page fragments. Otherwise, allocate a new
+		 * SKB fragment to hold @n_excess page fragments.
+		 */
+		nskb = __next_skb_frag(skb, pskb);
+		if (nskb && !skb_headlen(nskb)
+		    && (skb_shinfo(nskb)->nr_frags <= MAX_SKB_FRAGS - n_excess))
+		{
+			int r = __extend_pgfrags(nskb, NULL, 0, n_excess);
```
krizhanovsky: Yeah, nice optimization!

tempesta_fw/ss_skb.c
keshonok: I believe that is the correct behaviour. Suppose we store an skb fragment in `it->ptr`, and then do another modification that involves that skb. The modification may require that a new skb is created, as in `__extend_pgfrags()`. Then the new skb would be attached to the current skb as it's skb fragment which is wrong because the current skb in this case is an skb fragment itself. Skb fragments don't have fragments in the Linux kernel. Parent skb is the real skb, the others are skb fragments.

krizhanovsky: Intention about ```it->skb``` is simple: when we process ```it->ptr``` we could wish to change/examine corresponding ```skb```, so ```it->skb``` points to appropriate ```skb```. Sure parent skb also contain the data pointed by ```it->ptr```, but we'll lose CPU cycles to find the ```skb``` fragment. So ```it->ptr``` should be either in linear or paged data, not in fragments.

tempesta_fw/ss_skb.c
```diff
@@ -700,8 +701,7 @@ __skb_fragment(struct sk_buff *skb, struct sk_buff *pskb,
 
 	if ((offset >= 0) && (offset < d_size)) {
 		unsigned int t_size = d_size - offset;
```
milabs: Shouldn't we use here `int` type and not `unsigned int` one? Even the result is supposed to be positive, conversion from `t_size` to `-t_size` below looks suspicious.

tempesta_fw/ss_skb.c
```diff
@@ -533,80 +615,74 @@ __split_pgfrag_del(struct sk_buff *skb, int i, int off, int len, TfwStr *it)
...
 /**
- * Fragment @skb to add some room if @len > 0 or delete data otherwise.
+ * Add room for data to @skb if @len > 0 or delete data otherwise.
+ * Most of the time that is done by fragmenting the @skb.
+ * @bskb is the base skb if @skb is an skb fragment.
+ * Otherwise it equals @skb.
  */
 static int
-__skb_fragment(struct sk_buff *skb, struct sk_buff *pskb,
+__skb_fragment(struct sk_buff *skb, struct sk_buff *bskb,
```
krizhanovsky: Well comment https://github.com/natsys/tempesta/pull/458#discussion_r61856771 is still not fixed. You've just renamed `pskb` to `bskb`, but the only one place where `bskb` differs from `skb` is internal `__skb_fragment()` call. So the function should be split: `skb_fragment()` should accept `skb` only while `__skb_fragment()`, called from `skb_fragment()` should accept `bskb` as well as `skb`.

tempesta_fw/ss_skb.c
```diff
@@ -160,6 +160,61 @@ __lookup_pgfrag_room(struct sk_buff *skb, int len)
...
 
+/*
+ * Insert the new SKB at the correct location in the list of fragments
+ * of the base SKB. If @skb is one of the fragments, then insert the
+ * new SKB after @skb in the frag_list. Otherwise, insert the new SKB
+ * as the first fragment in the list.
+ */
+static void
+__insert_skb_frag(struct sk_buff *skb,
+		  struct sk_buff *bskb, struct sk_buff *nskb)
```
krizhanovsky: why `bskb` isn't defined at previous line, why do you break line in this way?

tempesta_fw/ss_skb.c
```diff
@@ -220,10 +289,11 @@ __extend_pgfrags(struct sk_buff *skb, int from, int n, TfwStr *it)
 
 /*
  * Make room for @shift fragments starting with slot @i. Then make
- * a new fragment in slot @i that can hold @size bytes, and it set up.
+ * a new fragment in slot @i that can hold @size bytes, and set it up.
  */
 static int
-__new_pgfrag(struct sk_buff *skb, int size, int i, int shift, TfwStr *it)
+__new_pgfrag(struct sk_buff *skb,
+	     struct sk_buff *bskb, int size, int i, int shift)
```
krizhanovsky: Again why do you break lines in this way? Looks very strange.

tempesta_fw/ss_skb.c
```diff
@@ -337,37 +397,37 @@ __split_linear_data(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
...
+	 */
+	if (len < 0) {
+		skb->tail += len;
+		skb->len += len;
+		ss_bskb_adjust_len(skb, bskb, len);
+		tail_len += len;
+		tail_off -= len;
 	}
+	skb->tail -= tail_len;
+	skb->data_len += tail_len;
```
krizhanovsky: why `skb->truesize` isn't updated anymore?

tempesta_fw/ss_skb.c
```diff
@@ -386,21 +446,32 @@ __split_linear_data(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
...
  * @return 0 on success, -errno on failure.
- * @return SKB in @it->skb if new SKB is allocated.
  * @return pointer to the room for new data in @it->ptr.
- * @return @it->flags is set if @it->ptr points to data in it->skb.
+ * @return pointer to SKB with data at @it->ptr in @it->skb.
  */
 static int
-__split_pgfrag_add(struct sk_buff *skb, int i, int off, int len, TfwStr *it)
+__split_pgfrag_add(struct sk_buff *skb, struct sk_buff *bskb,
+		   int i, int off, int len, TfwStr *it)
```
krizhanovsky: strange line break

tempesta_fw/ss_skb.c
```diff
@@ -297,25 +367,20 @@ __split_linear_data(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
...
-	 * Quick and unlikely path: just advance the skb tail pointer.
-	 * Note that this only works when we make room. When we remove,
-	 * pspt points at the start of the data chunk to remove. In that
-	 * case, tail_len can never be zero.
-	 */
-	if (unlikely(!tail_len && len <= ss_skb_tailroom(skb))) {
-		BUG_ON(len < 0);
-		it->ptr = ss_skb_put(skb, len);
-		return 0;
-	}
```
krizhanovsky: Hm, it seems you moved the fast path to `__split_pgfrag_add()` and this doesn't look correct. See comment in `__split_pgfrag_add()`.

tempesta_fw/ss_skb.c
```diff
@@ -386,21 +446,32 @@ __split_linear_data(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
...
+	 * Fast path: just advance the skb tail pointer.
+	 * If @off is zero and @i is zero (first paged fragment),
+	 * then see if there's sufficient room in the linear part.
+	 */
+	if (unlikely(!off && !i && (len <= ss_skb_tailroom(skb)))) {
+		ss_bskb_adjust_len(skb, bskb, len);
+		it->ptr = ss_skb_put(skb, len);
+		it->skb = bskb;
+		return 0;
+	}
```
krizhanovsky: If I understand correctly you moved the quick path here from `__split_linear_data()` just because of `offset < d_size` in `__skb_fragment()`. IIRC we discussed the condition and agreed that we should use `<=`.

tempesta_fw/ss_skb.c
```diff
@@ -533,80 +615,74 @@ __split_pgfrag_del(struct sk_buff *skb, int i, int off, int len, TfwStr *it)
...
-	if (i < si->nr_frags)
+	/* Get the address of data after the deleted data. */
+	if (i < si->nr_frags) {
 		it->ptr = skb_frag_address(&si->frags[i]);
+		it->skb = bskb;
+	} else {
+		struct sk_buff *dskb;
+		it->skb = __next_skb(skb, &bskb, &dskb);
+		it->ptr = __skb_data_address(dskb, bskb);
+	}
```
krizhanovsky: copy&paste from `__split_linear_data()`. This should be done as a function

        static inline void
        __lookup_next_ptr(struct sk_buff *skb, struct sk_buff *bskb, TfwStr *it, int i)
        {
                if (i < si->nr_frags) {                                                
                        it->ptr = skb_frag_address(&si->frags[i]);                     
                        it->skb = bskb;                                                
                } else {                                                               
                        struct sk_buff *dskb;                                          
                        it->skb = __next_skb(skb, &bskb, &dskb);                       
                        it->ptr = __skb_data_address(dskb, bskb);                      
                }               
        }

and should be called with `0` and `i` as last argument correspondingly.

tempesta_fw/ss_skb.c
```diff
@@ -656,13 +729,18 @@ __skb_fragment(struct sk_buff *skb, struct sk_buff *pskb,
...
 			len = max(len, -t_size);
-			ret = __split_pgfrag(skb, i, offset, len, it);
+			ret = __split_pgfrag(skb, bskb, i, offset, len, it);
 			goto done;
 		}
 	}
 
+	/* skb fragments are not expected in skb fragments. */
+	if (skb != bskb)
+		return -ENOENT;
```
krizhanovsky: Shouldn't we use `WARN_ON()` instead? This isn't an input error, but actually our assumption about how underlying code works. And this assumption can be correct or not. So we should see errors or warnings rather than dropping client connections.

krizhanovsky:
Looks good for me

PR #459, opened 2016-04-19
Allow only LF and CRLF in while parsing HTTP (v3)

milabs:

Review history:
- https://github.com/natsys/tempesta/pull/451 (v1)
- https://github.com/natsys/tempesta/pull/453 (v2)

Changes since v2:
- fulfilling the CodingStyle (braces and fill-column limit)
- don't use separate header file for EOL
- cache's code reflects that s_line not contains EOL anymore
- __hdr_del / __hdr_sub are updated to use EOL
- common (raw) headers correctness checking (colon + alphabet)

### DIFF
--- a/tempesta_fw/cache.c
+++ b/tempesta_fw/cache.c
@@ -47,7 +47,7 @@
  * @hdr_num	- numbder of headers;
  * @hdr_len	- length of whole headers data;
  * @key		- the cache enty key (URI + Host header);
- * @status	- pointer to status line;
+ * @status	- pointer to status line  (with trailing CRLFs);
  * @hdrs	- pointer to list of HTTP headers (with trailing CRLFs);
  * @body	- pointer to response body (with a prepending CRLF);
  */
@@ -293,6 +293,42 @@ tfw_cache_strcpy(char **p, TdbVRec **trec, TfwStr *src, size_t tot_len)
 }
 
 /**
+ * Copies plain or compound (chunked) TfwStr @src to TdbRec @trec may be
+ * appending EOL marker at the end.
+ *
+ * @src is copied (possibly with EOL appended)
+ * @return number of copied bytes on success and negative value otherwise.
+ */
+static long
+tfw_cache_copy_str(char **p, TdbVRec **trec,
+		   TfwStr *src, size_t *tot_len, bool eol)
+{
+	long n, copied = 0;
+	TfwStr *c, *end;
+
+	BUG_ON(TFW_STR_DUP(src));
+
+	TFW_STR_FOR_EACH_CHUNK(c, src, end) {
+		if ((n = tfw_cache_strcpy(p, trec, c, *tot_len)) < 0) {
+			TFW_ERR("Cache: cannot copy chunk of string\n");
+			return -ENOMEM;
+		}
+		*tot_len -= n;
+		copied += n;
+	}
+
+	if (eol) {
+		if ((n = tfw_cache_strcpy(p, trec, &g_crlf, *tot_len)) < 0)
+			return -ENOMEM;
+		BUG_ON(n != SLEN(S_CRLF));
+		*tot_len -= n;
+		copied += n;
+	}
+
+	return copied;
+}
+
+/**
  * Deep HTTP header copy to TdbRec.
  * @src is copied in depth first fashion to speed up upcoming scans.
  * @return number of copied bytes on success and negative value otherwise.
@@ -301,7 +337,7 @@ static long
 tfw_cache_copy_hdr(char **p, TdbVRec **trec, TfwStr *src, size_t *tot_len)
 {
 	long n = sizeof(TfwCStr), copied;
-	TfwStr *dup, *dup_end, *chunk, *chunk_end;
+	TfwStr *dup, *dup_end;
 
 	if (unlikely(src->len >= TFW_CSTR_MAXLEN)) {
 		TFW_WARN("Cache: trying to store too big string %lx\n",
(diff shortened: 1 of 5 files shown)


tempesta_fw/http_msg.c
```diff
@@ -475,21 +475,65 @@ __hdr_sub(TfwHttpMsg *hm, char *name, size_t n_len, char *val, size_t v_len,
...
+		goto slow_path;
+
+	BUG_ON(!tfw_str_eolen(orig_hdr));
+
+	/*
+	 * We are trying to keep line endings as they were received, so
+	 * adjustment of @hdr's last chunk is needed in case of single LF.
+	 */
+	if (tfw_str_eolen(orig_hdr) == 1)
+		++hlc->ptr, --hlc->len, --hdr.len;
```
keshonok: I wouldn't use the comma operator. I would just make these separate statements.

tempesta_fw/http_msg.c
```diff
@@ -475,21 +475,65 @@ __hdr_sub(TfwHttpMsg *hm, char *name, size_t n_len, char *val, size_t v_len,
...
+	 *
+	 *        hdr     | Foo: ......Ba | rEOL   5) slow_path
+	 */
+	if (hdr.len > orig_hdr->len) {
+		/*
+		 * Indeed, 4) is the luckiest case and we can reuse @orig_hdr's
+		 * EOL space by cutting off the @hdr's EOL chunk at all.
+		 */
+		if ((hdr.len - hlc->len) == orig_hdr->len) {
+			hdr.len -= hlc->len, TFW_STR_CHUNKN_SUB(&hdr, 1);
```
keshonok: Same as above, I wouldn't use the comma operator. I would just make these separate statements.

tempesta_fw/http_msg.c
```diff
@@ -475,21 +475,65 @@ __hdr_sub(TfwHttpMsg *hm, char *name, size_t n_len, char *val, size_t v_len,
...
+	 *
+	 *        hdr     | Foo: BarEOL.. |        1) cutoff, fast_path
+	 *        hdr     | Foo: ..BarEOL |        2) cutoff, fast_path
+	 *
+	 *        hdr     | Foo: ...BarEO | L      3) slow_path (CRLF only)
+	 *        hdr     | Foo: .....Bar | EOL    4) reuse EOL, fast_path
+	 *
+	 *        hdr     | Foo: ......Ba | rEOL   5) slow_path
+	 */
+	if (hdr.len > orig_hdr->len) {
```
keshonok: There's a lots of case listed here. In my opinion this only clouds the issue.

The pre-condition here is that EOL bytes for both the original string and the substitute string are normalized and made the same. So the real condition for the fast path is that the length of the substitute string without the EOL bytes must be no larger than the length of the original string which is already counted without the EOL bytes. That is the only case when we can use the fast path. Moreover, if the length of the substitute string without the EOL bytes is exactly the same as the length of the original string (already without the EOL bytes), then we don't need to cut off any bytes. We can just copy the substitute string over the original string leaving the EOL bytes intact.

In all other cases we must go for the slow path.

Here is a sample code that expesses this logic in a clear and obvious way. It also has less branches (goto). What do you guys think?


```C
/* The same for both strings. */
eolen = tfw_str_eolen(orig_hdr);

/*
 * EOL bytes are not a part of a header field string in Tempesta.
 * Therefore only @orig_hdr->len bytes at most can be copied over.
 * The EOL has been normalized to be the same for both strings.
 * If the substitute string without the EOL fits into that space,
 * then the fast path can be used. Otherwise, go by the slow path.
 */
hdrlen_wo_eol = hdr.len - eolen;
if (hdrlen_wo_eol <= orig_hdr->len) {
        if (hdrlen_wo_eol == orig_hdr->len) {
                hdr.len -= eolen;
                TFW_STR_CHUNKN_SUB(&hdr, 1);
                goto copy_over;
        }
        if (ss_skb_cutoff_data(orig_hdr, hdr.len, eolen))
                return TFW_BLOCK;
copy_over:
        return tfw_strcpy(orig_hdr, &hdr) ? TFW_BLOCK : 0;
}
/* Generic and slower path. */
```

tempesta_fw/http_msg.c
```diff
@@ -475,21 +475,65 @@ __hdr_sub(TfwHttpMsg *hm, char *name, size_t n_len, char *val, size_t v_len,
 			{ .ptr = name,	.len = n_len },
 			{ .ptr = ": ",	.len = 2 },
 			{ .ptr = val,	.len = v_len },
-			{ .ptr = "\r\n", .len = 2 }
+			{ .ptr = "\r\n",.len = 2 } /* may be adjusted */
 		},
 		.len = n_len + v_len + 4,
 		.flags = 4 << TFW_STR_CN_SHIFT
 	};
+	TfwStr *hlc = TFW_STR_LAST(&hdr);
```
keshonok: Perhaps it's better to use just `&hdr[3]` or something like that as that is calculated at compile time versus the run time.

tempesta_fw/http_msg.c
milabs: You've missed 3-rd case (see my description). Let's assume that `orig_hdr = { 10 } + CRLF` and `hdr = { 9 + CRLF }`. You can't use `ss_skb_cutoff_data` and must fall to the slow path.

keshonok: I am not sure I follow. So, we have:
- `orig_hdr = { 10 } + CRLF` -- that makes 10 bytes that we can copy.
- `hdr = { 9 + CRLF }` -- That makes 9 bytes (without `CRLF`) that we need to copy.
So it appears that we can copy these 9 bytes (without CRLF), and cut off 1 byte between the end of the substitute string and the original CRLF. Am I missing something here?

It also appears that we don't need to remove the tail at all in the fast path. Also, that means that we need to deal with the substitute string without EOL in the fast path as EOL is the same for both strings. The code sample above is not correct in that regard.

tempesta_fw/cache.c
```diff
@@ -293,6 +293,42 @@ tfw_cache_strcpy(char **p, TdbVRec **trec, TfwStr *src, size_t tot_len)
...
 
 /**
+ * Copies plain or compound (chunked) TfwStr @src to TdbRec @trec may be
+ * appending EOL marker at the end.
+ *
+ * @src is copied (possibly with EOL appended)
+ * @return number of copied bytes on success and negative value otherwise.
+ */
+static long
+tfw_cache_copy_str(char **p, TdbVRec **trec,
```
krizhanovsky: The name is just too close to ```tfw_cache_strcpy()``` and I expect that there will be a lof of confusions about differences between the functions. I'd propose something like ```tfw_cache_streolcpy``` or ```tfw_cache_strcpy_eol```.

PR #460, opened 2016-04-19
Allow only LF and CRLF in while parsing HTTP (v4)

milabs:

Review history:
- https://github.com/natsys/tempesta/pull/451 (v1)
- https://github.com/natsys/tempesta/pull/453 (v2)
- https://github.com/natsys/tempesta/pull/459 (v3)

Changes since v3:
- __hdr_sub reimplemented
- tfw_cache_copy_str -> tfw_cache_strcpy_eol

Changes since v2:
- fulfilling the CodingStyle (braces and fill-column limit)
- don't use separate header file for EOL
- cache's code reflects that s_line not contains EOL anymore
- __hdr_del / __hdr_sub are updated to use EOL
- common (raw) headers correctness checking (colon + alphabet)

### DIFF
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -444,7 +444,7 @@ __hdr_del(TfwHttpMsg *hm, int hid)
 
 	/* Delete the underlying data. */
 	TFW_STR_FOR_EACH_DUP(dup, hdr, end) {
-		if (ss_skb_cutoff_data(dup, 0, 2))
+		if (ss_skb_cutoff_data(dup, 0, tfw_str_eolen(dup)))
 			return TFW_BLOCK;
 	};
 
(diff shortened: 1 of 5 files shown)


tempesta_fw/http_msg.c
```diff
@@ -475,21 +475,56 @@ __hdr_sub(TfwHttpMsg *hm, char *name, size_t n_len, char *val, size_t v_len,
...
+
+	/*
+	 * We are trying to keep line endings as they were received, so
+	 * adjustment of @hdr's last chunk is needed in case of single LF.
+	 */
+	if (tfw_str_eolen(orig_hdr) == 1) {
+		--hdr.len;
+		++hlc->ptr;
+		--hlc->len;
+	}
```
krizhanovsky: It seems ```hlc``` is used for length only, ```hlc->ptr``` isn't used. So can we just use int variable instead?

keshonok:
Looks good to me.

Issue #375, opened 2015-12-19
TfwBomber functional test


krizhanovsky:
A functional test using Tempesta bomber's fuzzer must be developed to test regressions and basic Tempesta functionality. This is mandatory tool to prevent dummy bugs like #374 due to inaccurate commits. Valid and invalid queries and all cache modes must be testes.

Depends on #428.

krizhanovsky:
Following automated tests must be implemented (the list should be updated):

* small bomber fuzzying with valid and invalid queries;

* IPv6 compatibility;

krizhanovsky:
We're not going to spend more resource for TfwBomber, https://github.com/tempesta-tech/tempesta/issues/879#issuecomment-1485862857

PR #462, opened 2016-04-22
Sl 396

### LINKED ISSUE #396: Need to fix some errors in old functional(python) tests.
In tfw.py - start/stop parameters.
in fragmented_requests.py - formating.

sergsever:
Fix #396 and #375.


tempesta_fw/t/functional/tests/helpers/tfw.py
```diff
@@ -1,40 +1,50 @@
...
 
 import os
 import subprocess
 import sys
 
-from . import teardown
+import teardown
 
 __author__ = 'NatSys Lab'
 __copyright__ = 'Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).'
```
avbelov23: Tempesta Technologies Inc.

tempesta_fw/t/functional/tests/fragmented_requests.py
```diff
@@ -0,0 +1,106 @@
...
+
+# Start Tempesta FW and a back-end server with a default configuration.
+def run():
+	c = conf.Config('etc/tempesta_fw.conf')
+	c.add_option('cache', '0')
+	c.add_option('listen', '8081')
+	c.add_option('server', '127.0.0.1:80')
+
+	vs_get = b"GET / HTTP/1.0\r\nhost: loc\r\n\r\n"
+	s_get = b"GET http:localhost:80/index.html HTTP/1.0\r\n\
```
milabs: `http:localhost:80/index.html` is it valid? Also, it seems that `s_get` is never used.

tempesta_fw/t/functional/tests/fragmented_requests.py
```diff
@@ -0,0 +1,106 @@
...
+Connection: Keep-Alive\r\n\
+host: localhost\r\n\r\n"
+	tfw.start()
+	print("tfw start\n")
+	s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+	s.connect(("127.0.0.1",8081))
+	s.sendall(vs_get)
+	data = s.recv(1024)
```
milabs: Should the `data` be checked for some expected result?

tempesta_fw/t/functional/tests/fragmented_requests.py
```diff
@@ -0,0 +1,106 @@
...
+
+__author__ = 'Temesta Technologies Inc.'
+__copyright__ = 'Copyright (C) 2016 Tempesta Technologies Inc. (info@natsys-lab.com).'
+__license__ = 'GPL2'
+
+import conf
+import tfw
+import socket
+
+req_get = b'\
```
milabs: unused?

tempesta_fw/t/functional/tests/fragmented_requests.py
```diff
@@ -0,0 +1,106 @@
...
+def fragmentize_str(s, frag_size):
+	"""
+	Split a string into a list of equal N-sized fragmen.
+	>>> fragmentize_str("foo12bar34baz", 3)
+	['foo', '12b', 'ar3', '4ba', 'z']
+	"""
+	return [s[i:i+frag_size]  for i in range(0, len(s), frag_size)]
+
+
+def backend_callback(method, path, headers, body):
```
milabs: Unused?

tempesta_fw/t/functional/tests/helpers/be.py
```diff
@@ -0,0 +1,102 @@
...
+
+__author__ = 'NatSys Lab'
+__copyright__ = 'Copyright (C) 2016 NatSys Lab. (info@natsys-lab.com).'
+__license__ = 'GPL2'
+
+def start(*args, **kwargs):
+	"""A shortcut for BackendHTTPServer() constructor."""
+	return BackendHTTPServer(*args, **kwargs)
+
+def _dummy_callback(method, uri, headers, body):
```
milabs: Unused?

tempesta_fw/t/functional/tests/helpers/be.py
```diff
@@ -0,0 +1,102 @@
...
+	# returned by the callback.
+	# Read body  and push it to the callback
+
+	# The callback must return a tuple of exactly 3 elements:
+	# (http_code, headers_dict, body_str)
+
+	# Send response fields provided by the callback.
+
+	# At this point Tempesta FW parser blocks HTTP/1.0 requests
+protocol_version = 'HTTP/1.1'
```
milabs: Bad indentation!

tempesta_fw/t/functional/tests/helpers/teardown.py
```diff
@@ -12,8 +12,8 @@
 import sys
 from threading import *
 
-__author__ = 'NatSys Lab'
-__copyright__ = 'Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).'
+__author__ = 'Tempesta Technologies Inc.'
+__copyright__ = 'Copyright (C) 2014-2016 Tempesta Technologies Inc.. (info@natsys-lab.com).'
```
milabs: Two dots at the end.

tempesta_fw/t/functional/tests/test_cache.py
```diff
@@ -0,0 +1,41 @@
...
+		return 'Test cache'
+
+	def run(self):
+		tfw.start()
+
+		for x in range(0, 2):
+			s = socket(AF_INET, SOCK_STREAM)
+			s.connect(('127.0.0.1', 8081))
+			s.sendall(self.vs_get)
+			data = s.recv(1024)
```
milabs: Data is not checked for the result.

sergsever: If the problem existed, after second _send_ would be a *crash*, so if second *recv()* is Ok, then test is Ok.

tempesta_fw/t/functional/tests/test_cache.py
```diff
@@ -0,0 +1,48 @@
...
+		self.res = False
+		self.vs_get = b"GET http://localhost:80/ HTTP/1.1\r\n\
+Host: localhost\r\nConnection: Keep-alive\r\nSet-Cookie: session=42\r\n\r\n"
+# to reproduce the bug, a backend (apache) has to return in response double
+# headers.
+		self.apache_cfg = conf.Config('/etc/apache2/apache2.conf',
+					      new=False)
+		self.apache_cfg.add_string('Header add Set-Cookie: \"s=42\"')
+		self.apache_cfg.add_string('Header add Set-Cookie: \"s=42\"')
+		self.cfg = conf.Config("etc/tempesta_fw.conf")
```
krizhanovsky: Ok, I see why do you pass configuration path as an argument to `conf.Config()`. However, there is only one Tempesta config as only one Apache config, it's better to define two appropriate subclasses inherited from `conf` which implement Tempesta and Apached config logic. The code with the sublcasses wild be more readable.

tempesta_fw/t/functional/tests/test_cache.py
```diff
@@ -0,0 +1,48 @@
...
+	def __init__(self):
+		self.res = False
+		self.vs_get = b"GET http://localhost:80/ HTTP/1.1\r\n\
+Host: localhost\r\nConnection: Keep-alive\r\nSet-Cookie: session=42\r\n\r\n"
+# to reproduce the bug, a backend (apache) has to return in response double
+# headers.
+		self.apache_cfg = conf.Config('/etc/apache2/apache2.conf',
+					      new=False)
+		self.apache_cfg.add_string('Header add Set-Cookie: \"s=42\"')
+		self.apache_cfg.add_string('Header add Set-Cookie: \"s=42\"')
```
krizhanovsky: Why do you add the configuration option twice? Bug?

tempesta_fw/t/functional/tests/test_cache.py
```diff
@@ -0,0 +1,48 @@
...
+			s = socket(AF_INET, SOCK_STREAM)
+			s.connect(('127.0.0.1', 8081))
+			s.sendall(self.vs_get)
+			data = s.recv(1024)
+			if len(data) > 0:
+				self.res = True
+				s.close()	
+			else:
+				self.res = False
+				s.close()
```
krizhanovsky: Since this is cache test, please verify /proc/tempesta/perfstat for cache hits and cache misses and ensure that the there is exactly one new cache miss or there are two cache hits. Note that the entry can be already cached due to previous tests, so it's better to use special URI to ensure that the resource is not cached by previous tests.

tempesta_fw/t/functional/tests/test_cache.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
```
krizhanovsky: Please add haeader comment to ALL THE TESTS describing what and how the particular test verifies.

tempesta_fw/t/functional/tests/test_frang.py
```diff
@@ -0,0 +1,516 @@
...
+	def get_name(self):
+		return 'test Frang'
+
+	def run(self):
+		tests = [self.body_chunks(), self.header_chunks(),\
+ self.conn_burst(), self.request_burst(),\
+ self.body_timeout(), self.field_len(), self.body_len(),\
+ self.host_required(), self.ct_required(), self.req_method(),\
+ self.ct_vals(), self.uri_len(), self.request_rate(), self.conn_rate(),\
+ self.conn_max()]
```
krizhanovsky: Poor identation here and at in the many places above.

tempesta_fw/t/functional/tests/test_frang.py
```diff
@@ -0,0 +1,516 @@
...
+		self.cfg.add_option('http_ct_vals', '[\"text/html\"]')
+		self.cfg.add_option('http_body_chunk_cnt', '1')
+		self.cfg.add_end_of_section()
+		tfw.start_with_frang()
+		print("tfw start\n")
+		self.s = socket(AF_INET, SOCK_STREAM)
+		self.s.connect(('127.0.0.1', 8081))
+		self.s.send(part1)
+		self.s.send(part2)
+		self.s.send(part3)
```
krizhanovsky: It seems the data can be buffered by I/O library or OS, so the test isn't reliable, To avoid OS buffering TCP_NODELAY socket option should be used. Also please ensure that Python I/O layer doesn't do any buffering.

tempesta_fw/t/functional/tests/test_frang.py
```diff
@@ -0,0 +1,516 @@
...
+		data = self.s.recv(1024)
+		if len(data) == 0:
```
krizhanovsky: Does it mean that the connection was terminated? Can we satisfy the check if Tempestat is missbehaving and returns empty reposnse? I propose to grep dmesg (using some generic function of course, such that it can be used from many places) to check for frang limiting warning.

tempesta_fw/t/functional/tests/test_frang.py
```diff
@@ -0,0 +1,516 @@
...
+		self.s.send(self.vs_get)
+		data = self.s.recv(1024)
+		if len(data) == 0:
+			self.res = True
+			tcount += 1
+
+		print("res:", self.res)
+		self.s.close()
+		time.sleep(5)
+		tfw.stop()
```
krizhanovsky: No empty line before the method.

Also please write good comments for each method what the method verifies and how the verification works.

tempesta_fw/t/functional/tests/test_frang.py
```diff
@@ -0,0 +1,516 @@
...
+		self.cfg = conf.Config('etc/tempesta_fw.conf')
+		self.cfg.add_option('cache', '0')
+		self.cfg.add_option('listen', '8081')
+		self.cfg.add_option('server', '127.0.0.1:80')
+
+	def uri_len(self):
+		global tcount
+
+		self.res = False
+		print("uri\n")
```
krizhanovsky: Looks like forgotten debug prints. There are many such prints. Please clean them out.

tempesta_fw/t/functional/tests/test_frang.py
```diff
@@ -0,0 +1,516 @@
...
+		tfw.start_with_frang()
+		self.s.connect(("127.0.0.1",8081))
+		print("tfw start\n")
+		try:
+			self.s.send(self.vs_get)
+			data = self.s.recv(1024)
+			if len(data) == 0:
+				self.res = True
+				tcount += 1
+		except OSError as e:
```
krizhanovsky: Why the exception is not handled in many other similar places? It seems reading and verification of response should be moved to some generic function, such that the logic becomes more robust and consistent.

tempesta_fw/t/functional/tests/test_frang.py
```diff
@@ -0,0 +1,516 @@
...
+		self.cfg.add_section('frang_limits')
+		self.cfg.add_option('request_rate', '5')
+
+		self.cfg.add_end_of_section()
+		tfw.start_with_frang()
+		print("tfw start\n")
+		self.s.connect(("127.0.0.1",8081))
+
+		self.vs_get = b"GET / HTTP/1.0\r\nhost: loc\r\n" +\
+b"Connection: Keep-Alive\r\n\r\n"
```
krizhanovsky: Identation problem again. please fix all the issues in the file!

tempesta_fw/t/functional/tests/test_frang.py
```diff
@@ -0,0 +1,516 @@
...
+
+		try:
+			socks = []
+			conncount = 0
+			port = 8095
+			for x in range(0,7):
+				s = socket(AF_INET, SOCK_STREAM)
+				s.settimeout(2)
+				s.setsockopt(SOL_SOCKET, SO_REUSEADDR, 1)
+				s.bind(('127.0.0.6', port))
```
krizhanovsky: The logic seems broken. There is no sense to bind all the connection to the same port. I reckon you simply get exceptions here since 7 connection can not live with the same addresses and ports concurrently.

tempesta_fw/t/functional/tests/test_frang.py
```diff
@@ -0,0 +1,516 @@
...
+b"Connection: Keep-Alive\r\n\r\n"
+
+		try:
+			conncount = 0
+			port = 8095
+			for x in range(0,7):
+				self.s = socket(AF_INET, SOCK_STREAM)
+				self.s.settimeout(2)
+				self.s.setsockopt(SOL_SOCKET, SO_REUSEADDR, 1)
+				self.s.bind(('127.0.0.5', port))
```
krizhanovsky: The same as above

tempesta_fw/t/functional/tests/test_frang.py
```diff
@@ -0,0 +1,516 @@
...
+		self.__init__()
+		self.cfg.add_section('frang_limits')
+		self.cfg.add_option('client_header_timeout', '1')
+		self.cfg.add_end_of_section()
+		tfw.start_with_frang()
+		print("tfw start\n")
+		self.s = socket(AF_INET, SOCK_STREAM)
+		self.s.connect(('127.0.0.1', 8081))
+		self.s.send(part1)
+		time.sleep(1)
```
krizhanovsky: Probably it's better to sleep bit more than 1 second to exceed the timeout with certainty.

tempesta_fw/t/functional/tests/test_frang.py
```diff
@@ -0,0 +1,516 @@
...
+		self.__init__()
+		self.cfg.add_section('frang_limits')
+		self.cfg.add_option('client_body_timeout', '1')
+		self.cfg.add_end_of_section()
+		tfw.start_with_frang()
+		print("tfw start\n")
+		self.s = socket(AF_INET, SOCK_STREAM)
+		self.s.connect(('127.0.0.1', 8081))
+		self.s.send(part1)
+		time.sleep(1)
```
krizhanovsky: The same as above

tempesta_fw/t/functional/tests/test_parser.py
```diff
@@ -0,0 +1,111 @@
...
+				data = s.recv(1024)
+			except socket.error as e:
+				self.res = "exception:{}".format(e)
+				print("exception:{}".format(e))
+
+			s.close()
+			i += 1
+			if len(data) > 0:
+				parser = tfwparser.TFWParser()
+				if len(parser.get_body(data)) == 0:
```
krizhanovsky: Body is not necessary THML, so it's unclear why do you measure data length between `<html>` and `</html>` tags. Seems like some misconception... I believe you can just check full body length, w/o HTML parsing.

tempesta_fw/t/functional/tests/test_parser.py
```diff
@@ -0,0 +1,111 @@
...
+						self.res += "bodies "
+						self.res += "not equals\n"
+					else:
+						self.res += "bodies equals\n"
+				status = parser.get_status(data)
+				self.res += "status:{}\n".format(status)
+			
+		tfw.stop()	 
+		be.stop(be_pid)
+		print(self.res)
```
krizhanovsky: Again, no empty line between different methods

tempesta_fw/t/functional/tests/test_parser.py
```diff
@@ -0,0 +1,111 @@
...
+		self.resp = b'HTTP/1.0' + b' 200 - OK\n'
+		date = datetime.datetime.utcnow().strftime("%a, %d %b %Y" +\
+							   " %H:%M:%S GMT")
+		self.resp += b"Date: " + date + b"\r\n"
+		self.resp += b"Server: be python\r\n" 
+		self.resp += b'\r\n<html><body>content</body></html>\r\n\r\n'
+		self.run_test(1)
+	def run_test(self, num):
+
+		vs_get = b"GET / HTTP/1.0\nHost: loc\n" +\
```
krizhanovsky: Why do you use HTTP/1.0?

tempesta_fw/t/functional/tests/test_parser.py
```diff
@@ -0,0 +1,111 @@
...
+		self.resp = b'HTTP/1.0' + b' 200 - OK\r\n'
+		date = datetime.datetime.utcnow().strftime("%a, %d %b %Y" +\
+							   " %H:%M:%S GMT")
+		self.resp += b"Date: " + date + b"\r\n" 
+		self.resp += b"Server: be python\r\n\r\n"
+		self.resp += b'\r\n<html>content</html>\r\n\r\n'
+		self.run_test(2)
+		 
+# A response without the Content-Length header an without a body. 
+# For now (11.11.2016) tempesta's return status - 404.
```
krizhanovsky: Ok, this is test for #619. But the comment is very strange and it seems the subject for investigation. It seems you have some misconfiguration - Tempesta must just return response without body in this test. Please learn the reason for 404 response and fix the tests.

tempesta_fw/t/functional/tests/test_cache.py
sergsever: A problem was with double headers.

tempesta_fw/t/functional/tests/test_frang.py
sergsever: A port is not the same  - it is incremented in a loop, so for the Tempesta it looks like new connections from different clients.

tempesta_fw/t/functional/tests/test_frang.py
sergsever: It has proved to be enough.

PR #463, opened 2016-04-24
Fix erroneous accumulation of processed data in HTTP parser

### LINKED ISSUE #461: Incorrect/abnormal header field string in Tempesta
This occurs in current master at c434e5c. The test is as follows:
- The host machine is Ubuntu 14.04 LTS.
- The VM is QEMU+KVM running Centos 7, Tempesta kernel, and Tempesta.
- The host is connected to the VM via a bridge. The network driver in the VM is `virtio-net`.
- The back end server is Apache or Nginx that run on the host.
- The benchmark is `ab`, and it also runs on the host.

The test is to get a large file (95k) with `ab` benchmark without the `-k` parameter:
```
ab -r -n 1000000 -c 1000 http://192.168.10.230:80/tempesta.html
```

To catch this abnormal header field, the following patch is applied to master. It causes the kernel crash when an abnormal header string is encountered.
```diff
diff --git a/tempesta_fw/http_msg.c b/tempesta_fw/http_msg.c
index cc1f528..fd6bd42 100644
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -573,6 +573,12 @@ tfw_http_msg_hdr_xfrm(TfwHttpMsg *hm, char *name, size_t n_len,
                orig_hdr = &ht->tbl[hid];
        }
 
+if (orig_hdr->len > 32) {
+        printk(KERN_ERR "%s: ============================\n", __func__);
+        tfw_str_dprint(orig_hdr, __func__);
+        printk(KERN_ERR "%s: ============================\n", __func__);
+        BUG();
+}
        if (unlikely(append && hid < TFW_HTTP_HDR_NONSINGULAR)) {
                TFW_WARN("Appending to nonsingular header %d\n", hid);
                return -ENOENT;
diff --git a/tempesta_fw/ss_skb.c b/tempesta_fw/ss_skb.c
index 7bb827b..e41aa9f 100644
--- a/tempesta_fw/ss_skb.c
+++ b/tempesta_fw/ss_skb.c
@@ -111,6 +111,7 @@ __check_frag_room(struct sk_buff *skb, skb_frag_t *frag, int len)
        struct page *pg = skb_frag_page(frag);
        skb_frag_t *frag2, *ret = frag;
 
+return NULL;
        if ((refcnt = page_count(pg)) == 1)
                return frag; /* no other users */
 
diff --git a/tempesta_fw/str.c b/tempesta_fw/str.c
index a0f80d0..7395307 100644
--- a/tempesta_fw/str.c
+++ b/tempesta_fw/str.c
@@ -410,20 +410,20 @@ tfw_str_to_cstr(const TfwStr *str, char *out_buf, int buf_size)
 }
 EXPORT_SYMBOL(tfw_str_to_cstr);
 
-#ifdef DEBUG
+// #ifdef DEBUG
 void
 tfw_str_dprint(TfwStr *str, const char *msg)
 {
        TfwStr *dup, *dup_end, *c, *chunk_end;
 
-       TFW_DBG("%s: addr=%p skb=%p len=%lu flags=%x:\n", msg,
+       printk(KERN_ERR "%s: addr=%p skb=%p len=%lu flags=%x:\n", msg,
                str, str->skb, str->len, str->flags);
        TFW_STR_FOR_EACH_DUP(dup, str, dup_end) {
-               TFW_DBG("  duplicate %p, len=%lu, flags=%x:\n",
+               printk(KERN_ERR "  duplicate %p, len=%lu, flags=%x:\n",
                        dup, dup->len, dup->flags);
                TFW_STR_FOR_EACH_CHUNK(c, dup, chunk_end)
-                       TFW_DBG("   len=%lu, ptr=%p '%.*s'\n", c->len,
+                       printk(KERN_ERR "   len=%lu, ptr=%p '%.*s'\n", c->len,
                                c->ptr, (int)c->len, (char *)c->ptr);
        }
 }
-#endif
+// #endif
diff --git a/tempesta_fw/str.h b/tempesta_fw/str.h
index d747ba0..133679e 100644
--- a/tempesta_fw/str.h
+++ b/tempesta_fw/str.h
@@ -246,10 +246,10 @@ bool tfw_str_eq_cstr(const TfwStr *str, const char *cstr, int cstr_len,
 
 size_t tfw_str_to_cstr(const TfwStr *str, char *out_buf, int buf_size);
 
-#ifdef DEBUG
+// #ifdef DEBUG
 void tfw_str_dprint(TfwStr *str, const char *msg);
-#else
-#define tfw_str_dprint(str, msg)
-#endif
+// #else
+// #define tfw_str_dprint(str, msg)
+// #endif
 
 #endif /* __TFW_STR_H__ */
```

Here is the trace from the crash:
```
[ 1561.470147] ------------[ cut here ]------------
[ 1561.470147] kernel BUG at /root/tempesta/tempesta_fw/http_msg.c:580!
[ 1561.470147] invalid opcode: 0000 [#1] SMP
[ 1561.470147] Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) ppdev microcode i2c_piix4 9pnet_virtio 9pnet virtio_balloon pcspkr parport_pc i2c_core parport acpi_cpufreq ip_tables ata_generic pata_acpi virtio_net serio_raw virtio_pci virtio_ring virtio ata_piix floppy ipv6 autofs4 [last unloaded: tempesta_db]
[ 1561.470147] CPU: 1 PID: 0 Comm: swapper/1 Tainted: G           O    4.1.12-ab+ #12
[ 1561.470147] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS Bochs 01/01/2011
[ 1561.470147] task: ffff88011afe4290 ti: ffff88011afe8000 task.ti: ffff88011afe8000
[ 1561.470147] RIP: 0010:[<ffffffffa021f289>]  [<ffffffffa021f289>] tfw_http_msg_hdr_xfrm+0x4a9/0x4c0 [tempesta_fw]
[ 1561.470147] RSP: 0018:ffff88013fc837a8  EFLAGS: 00010202
[ 1561.470147] RAX: 0000000000000033 RBX: ffff88009aadc1d8 RCX: 0000000000000000
[ 1561.470147] RDX: 0000000000000001 RSI: ffff88013fc8d7f8 RDI: 0000000000000300
[ 1561.470147] RBP: ffff88013fc83918 R08: 0000000000000016 R09: 0000000000000003
[ 1561.470147] R10: 0000000000000001 R11: 0000000000000000 R12: ffffffffa0237a94
[ 1561.470147] R13: 0000000000000006 R14: ffff88009aadc020 R15: ffffffffa0237aab
[ 1561.470147] FS:  0000000000000000(0000) GS:ffff88013fc80000(0000) knlGS:0000000000000000
[ 1561.470147] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[ 1561.470147] CR2: 00007f38ede45000 CR3: 000000011949e000 CR4: 00000000000006e0
[ 1561.470147] Stack:
[ 1561.470147]  ffff88013fc83848 0000000000000000 0000000300000000 0000000000000016
[ 1561.470147]  ffff88013fc83868 0000000000000000 0000000000000020 0000040000000000
[ 1561.470147]  ffff88011afe4b60 ffff880119fd6e58 0000000000000000 00000000000003d4
[ 1561.470147] Call Trace:
[ 1561.470147]  <IRQ>
[ 1561.470147]  [<ffffffff810b5bd9>] ? __lock_acquire+0xc09/0x1cb0
[ 1561.470147]  [<ffffffffa021c78a>] tfw_http_adjust_resp+0x7a/0xf0 [tempesta_fw]
[ 1561.470147]  [<ffffffffa021cd31>] tfw_http_resp_cache_cb+0x31/0xc0 [tempesta_fw]
[ 1561.470147]  [<ffffffffa0218832>] tfw_cache_process+0xe2/0x140 [tempesta_fw]
[ 1561.470147]  [<ffffffffa021c55d>] ? tfw_http_popreq+0x3d/0x60 [tempesta_fw]
[ 1561.470147]  [<ffffffffa021cfa3>] tfw_http_msg_process+0x1e3/0x600 [tempesta_fw]
[ 1561.470147]  [<ffffffffa021be12>] __gfsm_fsm_exec+0x62/0xa0 [tempesta_fw]
[ 1561.470147]  [<ffffffffa021c054>] tfw_gfsm_dispatch+0x14/0x20 [tempesta_fw]
[ 1561.470147]  [<ffffffffa021b4f8>] tfw_connection_recv+0x98/0xb0 [tempesta_fw]
[ 1561.470147]  [<ffffffffa022b2ab>] ss_tcp_process_data+0x1ab/0x430 [tempesta_fw]
[ 1561.470147]  [<ffffffffa022ba08>] ss_tcp_data_ready+0x68/0xb0 [tempesta_fw]
[ 1561.470147]  [<ffffffff8158e5ca>] tcp_rcv_established+0x3ea/0x740
[ 1561.470147]  [<ffffffff815997d3>] tcp_v4_do_rcv+0x103/0x430
[ 1561.470147]  [<ffffffff8159b1d5>] tcp_v4_rcv+0xb05/0xc20
[ 1561.470147]  [<ffffffff810b2f34>] ? __lock_is_held+0x54/0x80
[ 1561.470147]  [<ffffffff815716ff>] ip_local_deliver_finish+0x13f/0x3f0
[ 1561.470147]  [<ffffffff8157160c>] ? ip_local_deliver_finish+0x4c/0x3f0
[ 1561.470147]  [<ffffffff81571c2a>] ip_local_deliver+0x9a/0xb0
[ 1561.470147]  [<ffffffff81571168>] ip_rcv_finish+0x168/0x5c0
[ 1561.470147]  [<ffffffff81571f2f>] ip_rcv+0x2ef/0x410
[ 1561.470147]  [<ffffffff81571000>] ? inet_del_offload+0x40/0x40
[ 1561.470147]  [<ffffffff81527178>] __netif_receive_skb_core+0x328/0xcd0
[ 1561.470147]  [<ffffffff81527b3d>] __netif_receive_skb+0x1d/0x60
[ 1561.470147]  [<ffffffff81527bfe>] process_backlog+0x7e/0x240
[ 1561.470147]  [<ffffffff81527c63>] ? process_backlog+0xe3/0x240
[ 1561.470147]  [<ffffffff815297c5>] ? net_rx_action+0x95/0x4e0
[ 1561.470147]  [<ffffffff81529887>] net_rx_action+0x157/0x4e0
[ 1561.470147]  [<ffffffff810628b6>] __do_softirq+0xe6/0x5d0
[ 1561.470147]  [<ffffffff81063065>] irq_exit+0xa5/0xb0
[ 1561.470147]  [<ffffffff81038895>] smp_call_function_single_interrupt+0x35/0x40
[ 1561.470147]  [<ffffffff81622e20>] call_function_single_interrupt+0x70/0x80
[ 1561.470147]  <EOI>
[ 1561.470147]  [<ffffffff810835d5>] ? atomic_notifier_call_chain+0x5/0x110
[ 1561.470147]  [<ffffffff8100dd35>] ? default_idle+0x25/0x200
[ 1561.470147]  [<ffffffff8100dd33>] ? default_idle+0x23/0x200
[ 1561.470147]  [<ffffffff8100e88f>] arch_cpu_idle+0xf/0x20
[ 1561.470147]  [<ffffffff810aaaeb>] cpu_startup_entry+0x25b/0x4f0
[ 1561.470147]  [<ffffffff810ec12c>] ? clockevents_register_device+0xec/0x1c0
[ 1561.470147]  [<ffffffff81039443>] start_secondary+0x143/0x170
[ 1561.470147] Code: e8 f0 54 3f e1 48 83 7b 10 20 4c 8b 85 a8 fe ff ff 44 8b 8d a4 fe ff ff 44 8b 9d a0 fe ff ff 8b 8d 9c fe ff ff 0f 86 44 fc ff ff <0f> 0b 45 31 e4 e9 72 fc ff ff 66 66 66 66 2e 0f 1f 84 00 00 00
[ 1561.470147] RIP  [<ffffffffa021f289>] tfw_http_msg_hdr_xfrm+0x4a9/0x4c0 [tempesta_fw]
[ 1561.470147]  RSP <ffff88013fc837a8>
```

In my case, if the back end server is Apache, I get the following abnormal string:
```
tfw_http_msg_hdr_xfrm: addr=ffff880036b882f8 skb=ffff8800baa86000 len=271 flags=302:
[ 5432.164057]   duplicate ffff880036b882f8, len=271, flags=302:
[ 5432.164057]    len=12, ptr=ffff8800ba8fd840 'Keep-Alive: '
[ 5432.164057]    len=9, ptr=ffff8800ba8fd84c 'timeout=5'
[ 5432.164057]    len=250, ptr=ffff8800ba8fd75c 'HTTP/1.1 200 OK^M
Date: Thu, 21 Apr 2016 17:27:51 GMT^M
Server: Apache/2.4.7 (Ubuntu)^M
Last-Modified: Mon, 04 Apr 2016 13:55:03 GMT^M
ETag: "17527-52fa9134df040"^M
Accept-Ranges: bytes^M
Content-Length: 95527^M
Vary: Accept-Encoding^M
Keep-Alive: timeout=5^M'
```
If the back end server is Nginx, I get the following abnormal string:
```
[ 1561.461037] tfw_http_msg_hdr_xfrm: addr=ffff88009aadc1d8 skb=ffff88011a566000 len=74 flags=302:
[ 1561.461037]   duplicate ffff88009aadc1d8, len=74, flags=302:
[ 1561.469170]    len=8, ptr=ffff88009c7e9249 'Server: '
[ 1561.470147]    len=20, ptr=ffff88009c7e9251 'nginx/1.4.6 (Ubuntu)'
[ 1561.470147]    len=46, ptr=ffff88009c7e9238 'HTTP/1.1 200 OK^M
Server: nginx/1.4.6 (Ubuntu)^M'
```

`wget` directly to these back ends produce the following response headers:
```
  HTTP/1.1 200 OK
  Date: Thu, 21 Apr 2016 21:23:21 GMT
  Server: Apache/2.4.7 (Ubuntu)
  Last-Modified: Mon, 04 Apr 2016 13:55:03 GMT
  ETag: "17527-52fa9134df040"
  Accept-Ranges: bytes
  Content-Length: 95527
  Vary: Accept-Encoding
  Keep-Alive: timeout=5
  Connection: Keep-Alive
  Content-Type: text/html
```
```
  HTTP/1.1 200 OK
  Server: nginx/1.4.6 (Ubuntu)
  Date: Thu, 21 Apr 2016 21:23:53 GMT
  Content-Type: text/html
  Content-Length: 95527
  Last-Modified: Tue, 12 Apr 2016 21:46:37 GMT
  Connection: keep-alive
  ETag: "570d6cbd-17527"
  Accept-Ranges: bytes
```

milabs:
This patch-set fixes #461 issue by synchronizing fix-ups and data movement what was the reason of erroneous accumulation of processed data.


tempesta_fw/http_parser.c
```diff
@@ -103,17 +92,105 @@ memchreol(const unsigned char *s, size_t n)
...
+	TfwHttpMsg	* msg = (TfwHttpMsg *)(ptr);			\
+	TfwHttpParser	* parser = &msg->parser;			\
+	unsigned char	* p = data;					\
+	unsigned char	c = *p;						\
+	int		__fsm_const_state;				\
+	int		__maybe_unused __fsm_n;				\
+	size_t		__maybe_unused __fsm_sz;			\
+	unsigned char	__maybe_unused * __fsm_ch;			\
+	TfwStr		__maybe_unused * chunk = &parser->_tmp_chunk;	\
+	;
```
keshonok: Even though a compiler may do this job, it's always a good idea to pack stack items as dense as possible, without holes when possible. It might be a good idea to move `unsigned char c` definition to the end of the block.

Also, this is not really the C language style (or the kernel source style) to leave a space between asterisk and the variable in the definition.

tempesta_fw/http_parser.c
```diff
@@ -103,6 +103,19 @@ memchreol(const unsigned char *s, size_t n)
 #define IS_WS(c) (c == ' ' || c == '\t')
 
 /**
+ * When parsing it's necessary to not forget to update @data and @len
+ * properly. So, presence of __data_{} macroses will be helpful to reduce amount
```
keshonok: `presence of __data_{} macroses will be helpful to reduce amount` -> `__data_{} macros help to reduce the amount`.
Usually I try my best not to fix comments. Sorry about that.

tempesta_fw/http_parser.c
```diff
@@ -1842,21 +1871,21 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
...
 
-		tfw_http_msg_hdr_open(msg, p);
+		__msg_hdr_open(p);
 
 		switch (LC(c)) {
 		case 'c':
 			__FSM_MOVE(Req_HdrC);
 		case 'h':
-			if (likely(p + 5 <= data + len
+			if (likely(__data_available(p, 5)
```
keshonok: It might be a good idea to use real strings in cases like this. That would make the code more readable. Say, in this particular case we could do this:
```C
#define SLEN(s)    (sizeof(s) - 1)
if (likely(__data_available(p, SLEN("Host:"))
             && C4_INT_LCM(p + 1, 'o', 's', 't', ':'))) {
        parser->_i_st = Req_HdrHostV;
        parser->_hdr_tag = TFW_HTTP_HDR_HOST;
        __FSM_MOVE_n(RGen_LWS_empty, SLEN("Host:"));
}
```
We use `SLEN()` macro in other parts of Tempesta code in similar fashion, even though, I believe, it's always local to a module.

tempesta_fw/http_parser.c
```diff
@@ -103,17 +92,105 @@ memchreol(const unsigned char *s, size_t n)
 #define IS_WS(c) (c == ' ' || c == '\t')
 
 /**
+ * When parsing it's necessary to not forget to update @data and @len
+ * properly. So, presence of __data_{} macroses will be helpful to reduce amount
+ * of direct @data/@len manipulations.
+ */
+#define __data_offset(pos)						\
+	(size_t)((pos) - data)
+#define __data_margin(pos)						\
```
keshonok: I believe this is supposed to mean the number of bytes that is not processed yet. `margin` is a very confusing term for that purpose. I would suggest to use the term `remainder`, like `__data_remainder()`

tempesta_fw/http_parser.c
```diff
@@ -103,17 +92,105 @@ memchreol(const unsigned char *s, size_t n)
...
+	(len - __data_offset(pos))
+#define __data_move_to(pos)						\
+	len -= __data_offset(pos), data = pos;
```
keshonok: Uh, I guess I'll let it slide this time. :-)

milabs:
Sorry, but the patch-set is currently unstable. I'm working on it's completion.

tempesta_fw/http_parser.c
krizhanovsky: +1

tempesta_fw/http_parser.c
```diff
@@ -103,17 +92,104 @@ memchreol(const unsigned char *s, size_t n)
...
+ * When parsing it's necessary to not forget to update @data and @len
+ * properly. So, __data_{} macros help to reduce the amount of direct
+ * @data/@len manipulations.
+ */
+#define __data_offset(pos)						\
+	(size_t)((pos) - data)
+#define __data_remain(pos)						\
+	(len - __data_offset(pos))
+#define __data_rebase(pos)						\
+	len -= __data_offset(pos), data = pos;
```
krizhanovsky: Would be good to make it as ```do { .. } while (0)``` statement.

tempesta_fw/http_parser.c
krizhanovsky: Yeah, +1, that would make the code better looking.

tempesta_fw/http_parser.c
```diff
@@ -140,22 +216,25 @@ do {									\
...
+#define __FSM_MOVE_nff(to, n, field, fixup)				\
 do {									\
 	p += n;								\
-	if (unlikely(p >= data + len || !*p)) {				\
+	if (unlikely(__data_offset(p) >= len)) {			\
 		r = TFW_POSTPONE; /* postpone to more data available */	\
 		__fsm_const_state = to; /* start from state @to next time */\
 		/* Close currently parsed field chunk. */		\
-		tfw_http_msg_field_chunk_fixup(msg, field, data, len);	\
+		if (fixup) __msg_field_fixup(field, data + len);	\
```
krizhanovsky: I think coding style requires to place IF statement and its body on different lines.

tempesta_fw/http_parser.c
```diff
@@ -103,17 +92,104 @@ memchreol(const unsigned char *s, size_t n)
...
+					     __data_offset(pos));	\
+		__data_rebase(pos);					\
+	}								\
+} while (0)
+#define __msg_hdr_finish(pos)						\
+({									\
+	int eolen = 1 + !!((parser->_tmp.eol & 0xff) == 0xda);		\
+	tfw_str_set_eolen(&parser->hdr, eolen);				\
+	tfw_http_msg_hdr_close(msg, msg->parser._hdr_tag);		\
+})
```
krizhanovsky: The macros are relatively complex, so please use empty lines between them (they are almost functions).

tempesta_fw/http_parser.c
```diff
@@ -103,17 +92,104 @@ memchreol(const unsigned char *s, size_t n)
...
+		tfw_http_msg_field_chunk_fixup(msg, field, data,	\
+					       __data_offset(pos));	\
+		__data_rebase(pos);					\
+	}								\
+} while (0)
+#define __msg_field_finish(field, pos)					\
+do {									\
+	__msg_field_fixup(field, pos);					\
+	(field)->flags |= TFW_STR_COMPLETE;				\
+} while (0)
```
krizhanovsky: The macros are relatively complex, so please use empty lines between them (they are almost functions).

tempesta_fw/http_parser.c
milabs: There was at least 2 points in @keshonok's comment. What exactly I should to correct?

tempesta_fw/http_parser.c
milabs: Nope. There would to much copy-paste here in that case as token length used no only in `__data_available` . I think we shouldn't mix all the possible enhancements in one patch-set.

tempesta_fw/http_parser.c
krizhanovsky: Basically both the comments are good, so it would be goot to fix both of them

tempesta_fw/http_parser.c
krizhanovsky: Yeah, we can do this in other patch. But anyway the note is good

tempesta_fw/http_parser.c
milabs: Not sure that the comment about the compiler's behavior is actual for nowadays. Let's give the compiler to do it's job well especially while speaking for local on-stack variables layout. There is no information on the gcc (x86-64) manual page about the question. Moreover, suppose that `unsigned char c` is moved at the end of the declaration. But nothing prevents us to make another variable declaration after the `__FSM_DECLARE_VARS` (see the code below, sometimes local variables get's declared just after the macro.

krizhanovsky: Actually compilers don't do the job well, e.g. https://github.com/natsys/tempesta/blob/master/tempesta_fw/http_parser.c#L106 , so we have used the rule as a coding style.

However, you're right - I just made a simple test:

        $ cat stack.c 
        extern int call(long a, long b, long c);
        
        int func (int argc, char *argv[])
        {
        	volatile char b = *argv[0];
        	volatile int a = argc;
        	volatile short e = *argv[1];
        	volatile long d = *(long *)argv[1];
        	return call(a + b, a + d, b - e);
        }
        $ gcc --version|head -1
        gcc (GCC) 5.1.1 20150618 (Red Hat 5.1.1-4)
        $ gcc -O2 -c -S -o stack.s stack.c
        $ grep -A 20 '\<func:' stack.s
        func:
        .LFB0:
        	.cfi_startproc
        	subq	$24, %rsp
        	.cfi_def_cfa_offset 32
        	movq	(%rsi), %rax
        	movzbl	(%rax), %eax
        	movb	%al, 1(%rsp)
        	movq	8(%rsi), %rax
        	movl	%edi, 4(%rsp)
        	movsbw	(%rax), %dx
        	movq	(%rax), %rax
        	movw	%dx, 2(%rsp)
        	movq	%rax, 8(%rsp)
        	movsbl	1(%rsp), %edx
        	movswl	2(%rsp), %r8d
        	movslq	4(%rsp), %rsi
        	movq	8(%rsp), %rcx
        	movsbl	1(%rsp), %edi
        	movl	4(%rsp), %eax
        	subl	%r8d, %edx

So compiler uses only 16 bytes for the data and properly pack them on stack.

tempesta_fw/http_parser.c
```diff
@@ -446,20 +522,30 @@ enum {
...
-	if (msg->parser.hdr.ptr) {
-		/* LF and CRLF are the only valid EOL values */
-		int eolen = 1 + !!(parser->_tmp.eol == 0xda);
-		tfw_str_set_eolen(&parser->hdr, eolen);
-		return tfw_http_msg_hdr_close(msg, parser->_hdr_tag);
+
+	if (parser->hdr.ptr)
+		return __msg_hdr_finish(p);
+
+	if (TFW_CONN_TYPE(msg->conn) & Conn_Srv) {
```
krizhanovsky: Why this is done for responses only?

tempesta_fw/http_parser.c
```diff
@@ -103,17 +92,104 @@ memchreol(const unsigned char *s, size_t n)
...
+do {									\
+	tfw_http_msg_set_data(msg, field, pos);				\
+	__data_rebase(pos);						\
+} while (0)
+#define __msg_field_fixup(field, pos)					\
+do {									\
+	if (__data_offset(pos)) {					\
+		tfw_http_msg_field_chunk_fixup(msg, field, data,	\
+					       __data_offset(pos));	\
+		__data_rebase(pos);					\
```
krizhanovsky: It seems the macro is about the new idea to move ```data```, which is unclear. Lets discuss it in the chat...

tempesta_fw/http_parser.c
```diff
@@ -140,22 +216,25 @@ do {									\
...
-		tfw_http_msg_field_chunk_fixup(msg, field, data, len);	\
+		if (fixup) __msg_field_fixup(field, data + len);	\
 		__FSM_EXIT()						\
 	}								\
 	c = *p;								\
 	goto to;							\
 } while (0)
 
+#define __FSM_MOVE_nf(to, n, field)					\
+	__FSM_MOVE_nff(to, n, field, 1)
```
krizhanovsky: In other places ```__FSM_MOVE_nff()``` is called with the same 2nd, 3rd, 4th arguments, so maybe it's better to introduce macro like ```__FSM_MOVE_nofixup(to)    __FSM_MOVE_nff(to, 0, NULL, 1)```?

tempesta_fw/http_parser.c
milabs: Only responses needs to handle `s_line`. Requests doesn't store the first HTTP-request line separately.

PR #470, opened 2016-05-04
Offset must be a signed type in __skb_fragment()

keshonok:
Also, make sure that we don't let through requests for deletion that
skip over the whole length if the string. That also ensures that the
CRLF tail is not deleted alone without the deletion of a part of the
tail part of the string.

Also, simplify several statements by using standard abs() and max()
functions.


milabs:
Looks good to me.

krizhanovsky:
good to merge

PR #473, opened 2016-05-07
Rework data manipulations in HTTP-parser

milabs:
Add some macros. Fix problem with incorrect `msg->crlf` handling in case of sparse `CRLF` (the problem existed earlier and was found while working on https://github.com/natsys/tempesta/issues/461#issuecomment-213615573)


krizhanovsky:
Good to merge

Issue #207, opened 2015-08-23
http_parser: correct handling of parsing headers splitted into 3 and more chunks


stefanstrogin:
Now it's presumed that input header string may be splitted into 2 chunks. But it may be split into any amount of chunks. We should correctly handle it.

stefanstrogin:
It should be tested with #150 
(`6. Current version of do_split_and_parse() splits request only to 2 pieces, but it has sense to split it into more pieces, typically also using sliding number of pieces, to stress compound strings`).

krizhanovsky:
Actually, we normally handle long and/or unknown header values, i.e. the headers could contain any number of chunks. However, the parser tries to eat small string (like "keep-alive" or "close" for Connection header) quickly and presumes that the could not contain more than 2 chunks. See comment in http_parser.c:

         * The parser enforces few sane limitations:
         *
         *      - short fields (like numeric Content-Length) could be carried by not
         *        more than 2 data chunks - the bigger number of chunks means some
         *        dirty games like Slow HTTP attack.

milabs:
Fixed by https://github.com/natsys/tempesta/pull/476

PR #476, opened 2016-05-08
Fixes for the #474 and #207 issues

### LINKED ISSUE #474: Incorrect parsing of Transfer-Encoding when operating in byte-by-byte mode
There is a problem with parsing `Transfer-Encoding: chunked` when the request is sending byte by byte. Here is the sample script that generates the request:

```perl
#!/usr/bin/perl -w

use strict;
use Time::HiRes;

$| = 1;

my $r = "\r\r\r\r\n\n\n\nGET / HTTP/1.1\nHost: localhost\r\nTransfer-Encoding: chunked\r\n\n8\r\n12345678\n0;ext=value\r\nSome-Header: some-value\r\nSome-Header: some-value\n\n";
for my $c (unpack("(a)*", $r)) {
    print $c;
    print STDERR $c;
    Time::HiRes::sleep 0.1;
}
```

To reproduce the problem one can use 5d19cca and run:

```shell
$ perl request.pl | socat -,ignoreeof TCP:192.168.88.220:3128




GET / HTTP/1.1
Host: localhost
Transfer-Encoding: chun2016/05/07 16:59:07 socat[2259] E write(5, 0x1d07650, 1): Broken pipe
```

So, Tempesta breaks the connection in the middle of `Transfer-Encoding` header parsing.

milabs:
Fixes #474, #207

### DIFF
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -830,8 +830,6 @@ __FSM_STATE(RGen_Body) {						\
 			return TFW_BLOCK;				\
 		default:						\
 			BUG_ON(__fsm_n < 0);				\
-			if (unlikely(__fsm_n == 0))			\
-				return TFW_BLOCK;			\
 			parser->to_read = parser->_tmp.acc;		\
 			if (!parser->to_read)				\
 				msg->body.flags |= TFW_STR_COMPLETE;	\
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -1277,6 +1277,9 @@ enum {
 	Req_I_H_Port,
 	/* Cache-Control header */
 	Req_I_CC,
+	Req_I_CC_m,
+	Req_I_CC_n,
+	Req_I_CC_o,
 	Req_I_CC_MaxAgeV,
 	Req_I_CC_MinFreshV,
 	Req_I_CC_Ext,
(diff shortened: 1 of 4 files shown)


tempesta_fw/str.c
```diff
@@ -364,6 +364,51 @@ tfw_str_eq_cstr(const TfwStr *str, const char *cstr, int cstr_len,
...
+		tmp.ptr += sizeof(TfwStr);
+
+		TFW_STR_CHUNKN_SUB(&tmp, 1);
+	}
+
+	TFW_WARN("Desired position is outside the string");
+out:
+	return r;
+}
+EXPORT_SYMBOL(tfw_str_eq_cstr_pos);
```
krizhanovsky: Please add appropriate tests to t/unit/test_tfw_str.c and make sure that all corner cases are handled properly

tempesta_fw/http_parser.c
```diff
@@ -479,19 +476,21 @@ enum {
 
 /* Parsing helpers. */
 #define TRY_STR_LAMBDA(str, lambda, state)				\
-	r = __chunk_strncasecmp(chunk, p, __data_remain(p),		\
-				str, sizeof(str) - 1);			\
-	switch (r) {							\
+	if (!chunk->ptr)						\
+		chunk->ptr = p;						\
```
krizhanovsky: Should ```chunk->len``` be initialized as well?

tempesta_fw/http_parser.c
```diff
@@ -335,44 +335,41 @@ static const unsigned long hdr_a[] ____cacheline_aligned = {
 #define CSTR_NEQ		TFW_BLOCK	/* -2 */
 #define CSTR_BADLEN		-3
 /**
- * Compare two chunks of data with const pattern @str.
- * If two chunks are less than @tot_len in length and if @chunk is empty,
- * then store @p in @chunk and postpone the comparison util next piece of data.
+ * Compare a mixed pair of strings with the string @src of length @tot_len where
```
keshonok: Perhaps -- ... with the string @str?

tempesta_fw/http_parser.c
```diff
@@ -335,44 +335,41 @@ static const unsigned long hdr_a[] ____cacheline_aligned = {
...
 
-	if (unlikely(tot_len > cn + len)) {
-		if (cn)
-			return CSTR_BADLEN;
-		chunk->ptr = p;
-		chunk->len = len;
-		return CSTR_POSTPONE;
-	}
+	if (offset > str_len ||
+	    (tolower(*p) != tolower(*(str + offset))))
```
krizhanovsky: unlikely() ?

tempesta_fw/http_parser.c
```diff
@@ -659,8 +657,10 @@ good_looking_eol:							\
 __FSM_STATE(st_curr) {							\
 	BUG_ON(__data_offset(p) > len);					\
 	__fsm_sz = __data_remain(p);					\
-	if (parser->_i_st == I_0)					\
+	if (parser->_i_st == I_0) {					\
+		TRY_STR_INIT();						\
```
krizhanovsky: It seems you initialize ```parser->chunk`` before and after ```TRY_STR()```/```TRY_STR_LAMBDA()``` - can we do this before the matching only?

Issue #411, opened 2016-01-31
Memory leak in current Tempesta kernel


keshonok:
Current Tempesta kernel is at tempesta-tech/linux-4.1-tfw@d3eb3e3. Tempesta runs in VM. The back end server is Apache running in the same VM. The only difference with Apache's default config is that Apache runs with `MaxKeepAliveRequests 0` option. Tempesta is tested with `ab` benchmark tool as follows:
```
ab -k -n 150000 -c 200 http://192.168.10.230:80/
```
The kernel loses a significant amount of available memory after each run of the benchmark. That can be seen by executing in the VM the command `$ cat /proc/meminfo` after each run of the benchmark. After several runs (about 7-10 runs or so in a VM with 4Gb RAM) the kernel exhaust all available memory. The OOM task killer starts killing tasks, but at that time it can't really help.

It appears the the problem is in the commit tempesta-tech/linux-4.1-tfw@6b6e899 named `Page allocator for socket buffers`. It has been verified that Tempesta kernel at tempesta-tech/linux-4.1-tfw@c3dabd9 which is right before that commit doesn't have a noticeable memory leak. For verification please run the kernel at tempesta-tech/linux-4.1-tfw@c3dabd9 commit, and use [ab-hdr-corrupt](https://github.com/natsys/tempesta/commits/ab-hdr-corrupt) branch rebased on top of current Tempesta master at 7c0dbc2.

keshonok:
Not sure this is related to the memory leak issue, but I have also encountered this bug once that could be due to memory corruption. This happens when the VM's RAM is being pressured by running the above mentioned benchmark at the time the memory almost run out.

```
BUG: Bad page state in process swapper/0  pfn:0d9a7
page:ffffea00003669c0 count:1 mapcount:0 mapping:          (null) index:0x0
flags: 0x800000000000000()
page dumped because: nonzero _count
Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) 9p ppdev microcode i2c_piix4 parport_pc pcspkr 9pnet_virtio 9pnet serio_raw virtio_balloon i2c_core parport acpi_cpufreq uinput xfs libcrc32c ata_generic pata_acpi virtio_net virtio_pci virtio_ring virtio ata_piix floppy dm_mirror dm_region_hash dm_log dm_mod ipv6 autofs4
[...]

PR #480, opened 2016-05-16
Fix two memory leaks

milabs:
This fixes two leaks, found while digging #411

### DIFF
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -796,25 +796,14 @@ tfw_http_conn_msg_unlink(TfwHttpMsg *m)
 void
 tfw_http_msg_free(TfwHttpMsg *m)
 {
-	struct sk_buff *skb;
-
 	TFW_DBG3("Free msg=%p\n", m);
 	if (!m)
 		return;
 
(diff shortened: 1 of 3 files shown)


tempesta_fw/http_msg.c
```diff
@@ -796,25 +796,14 @@ tfw_http_conn_msg_unlink(TfwHttpMsg *m)
...
-			 " destructor=%p users=%d type=%s\n",
-			 skb, skb->truesize, skb->sk, skb->data,
-			 skb->destructor, atomic_read(&skb->users),
-			 m->conn && TFW_CONN_TYPE(m->conn) & Conn_Clnt
-			 ? "Conn_Clnt"
-			 : m->conn && TFW_CONN_TYPE(m->conn) & Conn_Srv
-			   ? "Conn_Srv" : "Unknown");
-		kfree_skb(skb);
-	}
+	/* TODO: use ss_skb_list_print() to show SKBs queue */
```
krizhanovsky: Isn't it better just to move the slightly modified piece of code to the new function ```ss_skb_list_print()``` rather than write the TODO?

Issue #434, opened 2016-03-18
/root/tempesta/tempesta_fw/sock_srv.c:286 suspicious rcu_dereference_check() usage!


keshonok:
I have seen this just once so far.

```
[ 1371.313316] ===============================
[ 1371.317998] [ INFO: suspicious RCU usage. ]
[ 1371.322689] 4.1.12+ #12 Tainted: G        W  O   
[ 1371.327904] -------------------------------
[ 1371.332590] /root/tempesta/tempesta_fw/sock_srv.c:286 suspicious rcu_dereference_check() usage!
[ 1371.341809] 
[ 1371.341809] other info that might help us debug this:
[ 1371.341809] 
[ 1371.351412] 
[ 1371.351412] rcu_scheduler_active = 1, debug_locks = 0
[ 1371.359018] 2 locks held by swapper/0/0:
[ 1371.363494]  #0:  (((&icsk->icsk_retransmit_timer))){+.-...}, at: [<ffffffff8110f8e5>] call_timer_fn+0x5/0x4b0
[ 1371.374179]  #1:  (slock-AF_INET){+.-...}, at: [<ffffffff8159fbb1>] tcp_write_timer+0x21/0x90
[ 1371.383400] 
[ 1371.383400] stack backtrace:
[ 1371.388953] CPU: 0 PID: 0 Comm: swapper/0 Tainted: G        W  O    4.1.12+ #12
[ 1371.396882] Hardware name: Quanta                 S89/QR1-890/QSSC-890  /S89/QR1-890/QSSC-890  , BIOS S89_3A10 10/21/2010
[ 1371.408507]  0000000000000001 ffff88007c803c88 ffffffff8165fc2f 0000000000000011
[ 1371.416661]  ffffffff81c145c0 ffff88007c803cb8 ffffffff810e4637 ffff88002c558bb8
[ 1371.424807]  ffff88003bea8040 ffff88003bea8268 0000000000000006 ffff88007c803d18
[ 1371.432948] Call Trace:
[ 1371.436053]  <IRQ>  [<ffffffff8165fc2f>] dump_stack+0x4c/0x65
[ 1371.442489]  [<ffffffff810e4637>] lockdep_rcu_suspicious+0xe7/0x120
[ 1371.449438]  [<ffffffffa0683efc>] tfw_sock_srv_do_failover.isra.10+0x10c/0x130 [tempesta_fw]
[...]

PR #483, opened 2016-05-17
Remove RCU-like access to sk->sk_user_data (#434)

keshonok:
Replace it with plain and simple access. Now that Tempesta code
runs completely under socket lock, the use of RCU-like access
is no longer needed.


krizhanovsky:
Looks good. Could you please update ./t/sync_sockets/kernel/sync_*.c and ./t/bomber.c as well?

PR #486, opened 2016-05-18
Fix crash while working with the cache 

milabs:
Crash happens when duplicate string being read from the cache.


krizhanovsky:
good to merge

Issue #491, opened 2016-05-20
HTTP: trailing headers


krizhanovsky:
We should verify trailing headers according to `Trailer` header, [RFC 7230 4.4](https://tools.ietf.org/html/rfc7230#section-4.4). Moreover, since Tempesta moves trailing headers (is it consistent with RFC?) to the message header, it must remove `Trailer` header, i.e. do not cache it at all. So `Trailer` is hop-by-hop header in our logic.

ikoveshnikov:
Also note, that not all headers are allowed to appear in trailer part. https://tools.ietf.org/html/rfc7230#section-4.1.2

Issue #489, opened 2016-05-19
Incomplete parsing of chunked-body


milabs:
The problem is that current parsing of chunked body omits some body parts is case of `Transfer-Encoding: chunked`. The reason of the problem is incorrect handling of body in https://github.com/natsys/tempesta/blob/master/tempesta_fw/http_parser.c#L779.

krizhanovsky:
Chunked transfer encoding is wide spread, so the issue is crucial.

PR #492, opened 2016-05-20
Fix chunked body parsing issue

milabs:


tempesta_fw/http.h
```diff
@@ -139,7 +139,7 @@ typedef struct {
 	int		state;
 	int		_i_st;
 	int		to_read;
-	union {
+	struct {
 		unsigned long acc;
 		unsigned long eol;
 	} _tmp;
```
krizhanovsky: Why do we need `_tmp`, why not to embed `acc` and `eol` directly to `TfwHttpParser`?

tempesta_fw/cache.c
```diff
@@ -436,6 +436,8 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwHttpReq *req,
 		n = field - resp->h_tbl->tbl;
 		/* Skip hop-by-hop headers. */
 		h = (n < TFW_HTTP_HDR_RAW && hbh_hdrs[n]) ? &empty : field;
+		/* Skip trailer fields from being saved (FIXME: #491). */
+		h = tfw_str_eolen(h) ? h : &empty;
```
krizhanovsky: #491 doesn't require not to store trailer header in cache, but rather just `Trailer` header which describes the trailer header. Removing arbitrary server headers could break Web application.

tempesta_fw/http_msg.c
```diff
@@ -321,6 +321,14 @@ tfw_http_msg_hdr_close(TfwHttpMsg *hm, int id)
 done:
 	*h = hm->parser.hdr;
 
+	/*
+	 * We must distinguish headers opened after the body from generic
+	 * headers, so mark them by setting @eolen value to 0. It helps us to
+	 * prevent saving them to the cache (FIXME: #491).
+	 */
+	if (hm->body.flags & TFW_STR_COMPLETE)
+		tfw_str_set_eolen(h, 0);
```
krizhanovsky: Standard doesn't say anything about treating trailer headers in different manner than usual header, so why should we do this?

tempesta_fw/http_parser.c
```diff
@@ -866,6 +850,10 @@ __FSM_STATE(RGen_BodyChunkExt) {					\
 	return TFW_BLOCK;						\
 }
 
+#define TFW_HTTP_UPDATE_BODY_IF_OPENED()				\
+	if (msg->body.ptr)						\
+		tfw_http_msg_add_data_ptr(msg, &msg->body, data, len);
```
krizhanovsky: I think the macro is redundant. At any FSM state of HTTP message chunk processing we have `data` pointing to begin of the chunk. So we can set up the body chunk at any time. The macro is especially bad for HTTP requests which usually don't have body at all, but we still do the branching (BTW it would be a good place to use `unlikely`)

tempesta_fw/http_parser.c
```diff
@@ -598,14 +599,12 @@ good_looking_eol:							\
 		case 0xdada:						\
 			if (!(msg->crlf.flags & TFW_STR_COMPLETE)) {	\
 				__msg_crlf_finish(p);			\
-				parser->_tmp.eol = 0;			\
-				TFW_HTTP_INIT_BODY_PARSING(msg, RGen_Body); \
+				__FSM_JMP(RGen_Body);			\
```
krizhanovsky: I didn't get why don't we move to `RGen_BodyChunk` here?

tempesta_fw/http_parser.c
```diff
@@ -644,7 +641,7 @@ good_looking_eol:							\
 	 * The next chunk of chunked-body payload (RFC 7320, 4.1.2)	\
 	 * needs to be handled.						\
 	 */								\
-	__FSM_JMP(RGen_Body);						\
+	__FSM_JMP(RGen_BodyChunk);					\
```
krizhanovsky: I didn't get the logic. It seems at this point you don't know transfer encoding, we only do the check at `RGen_Body` state. At the state we check TFW_HTTP_CHUNKED and jump to `RGen_BodyChunk`... Why do you directly jump to `RGen_BodyChunk`?

tempesta_fw/http_parser.c
```diff
@@ -774,59 +771,47 @@ __FSM_STATE(RGen_HdrOtherV) {						\
...
-#define TFW_HTTP_PARSE_BODY()						\
-/* Read request|response body. */					\
-__FSM_STATE(RGen_Body) {						\
+}									\
+__FSM_STATE(RGen_BodyChunk) {						\
 	TFW_DBG3("read body: to_read=%d\n", parser->to_read);		\
+	__fsm_sz = __data_remain(p);					\
+	/* Open body for updading only if it were not opened */		\
+	if (!msg->body.ptr)						\
+		tfw_http_msg_add_data_ptr(msg, &msg->body, p, __fsm_sz);\
```
krizhanovsky: This is essentially `TFW_HTTP_UPDATE_BODY_IF_OPENED()`. Why don't you use the macro here?

milabs: `TFW_HTTP_UPDATE_BODY_IF_OPENED` check for `msg->body.ptr` not the `!msg->body->ptr`. This case is related to body opening.

tempesta_fw/http.h
milabs: That's all by historical reasons. We had `_tmp_acc` some time ago, then we had `_tmp.{acc,eol}` as a union. Now we have `_tmp.acc` and `_tmp.eol`. Seems that next we'll have is `acc` and `eol`, but not is that patch.

tempesta_fw/cache.c
milabs: Trailing headers stays in the body. This change prevents them from being restored from the cache.

krizhanovsky: We need to move the headers to normal headers place. But the fix is not about #489, so it seems we just don't need to do anything in the patch. I think trail headers are relatively rare and we should properly fix them in $491.

tempesta_fw/http.h
krizhanovsky: Why not to do the dummy substitution now if you touch the piece of code anyway?

tempesta_fw/http_parser.c
krizhanovsky: Ok, I see

tempesta_fw/http_parser.c
```diff
@@ -598,14 +599,12 @@ good_looking_eol:							\
 		case 0xdada:						\
```
krizhanovsky: In line 545: there is no need to do `& 0xf` since you accept 0xd and 0xa only in the IF statement.

tempesta_fw/http_parser.c
```diff
@@ -448,7 +449,7 @@ parse_int_hex(unsigned char *data, size_t len, unsigned long *acc)
 	unsigned char *p;
 
 	for (p = data; p - data < len; ++p) {
-		if (unlikely(isspace(*p) || (*p == ';')))
+		if (unlikely(IS_CR_OR_LF(*p) || (*p == ';')))
 			return p - data;
```
krizhanovsky: It seems the parser logic is broken. Assume we receive

         GET / HTTP/1.1\r\n
         Transfer-Encoding: chunked\r\n
         \r\n
         \r\n

`parse_int_hex()` returns 0 from the `if` statement when its called at `RGen_BodyChunk` state. `parser->_tmp.acc` stays zero, so we add TFW_STR_COMPLETE to `body` string. `RGen_BodyChunkExt` finds the last CRLF and moves to `RGen_EoL`. `RGen__EoL` finds that `crlf` and `body` are complete and returns TFW_PASS from the FSM. However, there is no mandatory `last-chunk`.

tempesta_fw/http_parser.c
krizhanovsky: Ok, my previous comment for the place seems invalid. However, it'd be good to rename `RGen_BodyChunk` since now it looks like processing chunked body, but in fact the state also handles chunked messages as well ans `Content-Length` defined. Or you can just write a comment just before the state clarifying the state logic.

tempesta_fw/http_parser.c
krizhanovsky: I reviewed the macro more carefully and the comment is right. HTTP parser is a pushdown automaton and we really can check `msg->body.ptr` (as a stack value), but there is no sense to do this for _all_ states. Rather we should do this _not_earlier_ than reach body processing states. So the macro breaks automata concepts.

tempesta_fw/http_parser.c
```diff
@@ -785,84 +672,111 @@ do {									\
...
-			if (!parser->to_read)				\
-				msg->body.flags |= TFW_STR_COMPLETE;	\
-			parser->_acc = 0;				\
-			__FSM_B_MOVE_n(RGen_BodyChunkExt, __fsm_n);	\
-		}							\
-	}								\
+	/* prevent @parse_int_hex false positives */			\
+	if (!parser->to_read && !isxdigit(c))				\
+		return TFW_BLOCK;					\
+	if (!parser->to_read)						\
```
keshonok: There's no need to check the condition twice.

milabs: Nope. This is required as `parser->to_read` may contains non-zero `Content-Length` value.

keshonok: What I meant was this:
```C
if (!parser->to_read) {
    if (!isxdigit(c))
        return TFW_BLOCK;
    __FSM_JMP(RGen_BodyChunkLen);
}
```

tempesta_fw/http_msg.c
```diff
@@ -332,13 +332,21 @@ tfw_http_msg_hdr_close(TfwHttpMsg *hm, int id)
 	return TFW_PASS;
 }
 
+/**
+ * Fixup the new data chunk to the @str.
+ *
+ * Like @tfw_http_msg_field_chunk_fixup, but besides the last one, it returns
+ * the operation result which can be checked on the caller side. One day, there
+ * must be only one of them.
```
krizhanovsky: Would be good to mark this by "TODO" -  so the comment will be move explicit and someone will merge the functions.

tempesta_fw/cache.c
```diff
@@ -526,8 +527,10 @@ __cache_entry_size(TfwHttpResp *resp, TfwHttpReq *req)
 	/* Add status line length + CRLF */
 	size += resp->s_line.len + SLEN(S_CRLF);
 
-	/* Add body size */
+	/* Add body size accounting CRLF after the last chunk */
 	size += resp->body.len;
+	if (resp->flags & TFW_HTTP_CHUNKED)
+		size += SLEN(S_CRLF);
```
krizhanovsky: Will the lengths and offset calculations break if we get '\n' only as EoL for chunks?

tempesta_fw/http_parser.c
```diff
@@ -498,153 +489,51 @@ enum {
...
-				__msg_crlf_open(p);			\
+	/* The header may be unopened in case of parsing s_line */	\
+	if (!parser->hdr.ptr)						\
+		__FSM_MOVE_nofixup(RGen_Hdr);				\
+	tfw_str_set_eolen(&parser->hdr, 1 + !!(parser->_eol == 0xda));	\
+	/* Zero length means that we've got an empty-line */		\
+	if (unlikely(!parser->hdr.len)) {				\
+		if (!(msg->crlf.flags & TFW_STR_COMPLETE)) {		\
+			msg->crlf = parser->hdr;			\
+			msg->crlf.flags |= TFW_STR_COMPLETE;		\
```
krizhanovsky: It seems you don't set `skb`, `len` and `eolen` for `msg.crlf` - why? Probably for now there is no place where we need `crlf->skb`, but if we'd to use it for some data mangling we'll hit memory corruption, i.e. in general it's good to follow TfwStr semantics and keep it consistent. Next, what if we get '\n' only instead of `\r\n' as message CRLF? See https://github.com/tempesta-tech/tempesta/pull/492/files#r64988311 as well.

tempesta_fw/http_parser.c
```diff
@@ -786,84 +674,112 @@ do {									\
...
-#define __FSM_B_MOVE(to)	__FSM_B_MOVE_n(to, 1)
+#define __FSM_B_MOVE(to)						\
+	__FSM_B_MOVE_n(to, 1)
 
-#define TFW_HTTP_INIT_BODY_PARSING(msg, to_state)			\
+#define TFW_HTTP_INIT_BODY_PARSING(msg)					\
 do {									\
 	TFW_DBG3("parse msg body: flags=%#x content_length=%lu\n",	\
 		 msg->flags, msg->content_length);			\
 	/* RFC 2616 4.4: firstly check chunked transfer encoding. */	\
```
krizhanovsky: Ok, for now, but don't forget in future to update RFC 2616 references to RFC 723[0-5].

tempesta_fw/http_parser.c
milabs: `eolen` was set on hdr at line 523 and in case of empty-line detected (CRLF) whole the header will be copied to `crlf`. So, the `crlf->skb`, `crlf->len` (0) and `crlf->ptr` will be copied from the hdr.

tempesta_fw/http_parser.c
```diff
@@ -786,84 +674,112 @@ do {									\
...
-		__fsm_n = parse_int_hex(p, __fsm_sz, &parser->_tmp.acc);\
-		TFW_DBG3("len=%zu ret=%d to_read=%lu\n",		\
-			 __fsm_sz, __fsm_n, parser->_tmp.acc);		\
-		switch (__fsm_n) {					\
-		case CSTR_POSTPONE:					\
-			/* Not all data has been parsed. */		\
-			__FSM_B_MOVE_n(RGen_Body, __fsm_sz);		\
-		case CSTR_BADLEN: /* bad header length */		\
-		case CSTR_NEQ: /* bad header value */			\
+		/* prevent @parse_int_hex false positives */		\
```
krizhanovsky: Not a big deal, but it's better to follow normal English grammar in comments: start sentence from Capital letter and finish it with dot. However, short trailing comments which are "part of the sentence", like `case CSTR_NEQ: /* bad header value */\` above can bypass the rules. Just a small note to make code nice.

tempesta_fw/http_parser.c
krizhanovsky: Oh, I see, thanks.

krizhanovsky:
good to merge

Issue #494, opened 2016-05-22
Cache: cache only specified request methods


krizhanovsky:
New configuration option `cache_meth` must be added. The option must accept space separated list of cacheable methods, e.g.

          cache_meth GET POST;

All methods, not listed in the configuration option, must not be cached. GET only must be cached by default. Tempesta must write through all non-cached methods, i.e. just proxy them w/o caching.

Please update README.md as well.

krizhanovsky:
Cache key for POST requests also must contain request body, see [Caching POST Responses with Nginx](https://medium.com/programming-articles/caching-post-responses-with-nginx-1c0c064bb6b0) for details.

krizhanovsky:
POST requests as well as other [non-safe](https://tools.ietf.org/html/rfc7231#section-4.2.1)/or unknown  methods must be unchacheable. Cacheable POST is a topic for #506.

keshonok:
Implemented in b7845d2.

PR #504, opened 2016-05-24
Add cache_method directive: cache the specified request methods only. (#494)

keshonok:


tempesta_fw/cache.c
```diff
@@ -158,6 +159,33 @@ static TfwStr g_crlf = { .ptr = S_CRLF, .len = SLEN(S_CRLF) };
 	}								\
 	for ( ; c != h_end; ++c, c = (c == u_end) ? h_start : c)
 
+static unsigned int tfw_cache_method_noncacheable = (1 << TFW_HTTP_METH_POST);
```
krizhanovsky: Please don't use so long names, `tfw_cache_meth_nc` would be enough for the variable and the functions at the below.

krizhanovsky:
Good to merge after several cleanups

PR #505, opened 2016-05-24
Set TCP layer's tcp_gso_segs, instead of skb's gso_segs.

keshonok:
skb's gso_segs, gso_size and gso_type are set by the TCP layer at
a later stage.


krizhanovsky:
good to merge

PR #512, opened 2016-05-25
Fix CID 111488

milabs:


keshonok:
Looks good to me.

tempesta_fw/classifier/frang.c
```diff
@@ -742,8 +742,9 @@ frang_http_req_process(FrangAcc *ra, TfwConnection *conn, struct sk_buff *skb,
 	__FRANG_FSM_STATE(Frang_Req_Hdr_UriLen) {
 		if (frang_cfg.http_uri_len) {
 			r = frang_http_uri_len(req, ra);
-			if (!(req->uri_path.flags & TFW_STR_COMPLETE))
+			if (!(req->uri_path.flags & TFW_STR_COMPLETE)) {
 				__FRANG_FSM_JUMP_EXIT(Frang_Req_Hdr_UriLen);
+			}
```
krizhanovsky: Wow, the macro is poorly written. It's better to use `do { } while (0)` construction in the macro to prevent further issues like this. Also it's good to do the same for `__FRANG_FSM_MOVE()`

milabs: Ok, I'll take it.

krizhanovsky:
good to merge

Issue #538, opened 2016-05-28
kernel BUG at /root/tempesta/tempesta_fw/cfg.c:628!


keshonok:
To reproduce, just omit the semicolon that closes a directive located in a configuration section between curly brackets. For example:
```
srv_group dummy {
    server 127.0.0.1
}
```

```
[ 5055.180793] kernel BUG at /root/tempesta/tempesta_fw/cfg.c:628!
[ 5055.181100] invalid opcode: 0000 [#1] SMP
[ 5055.181100] Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) i2c_piix4 i2c_core ppdev virtio_balloon 9pnet_virtio 9pnet microcode pcspkr acpi_cpufreq parport_pc parport ip_tables ata_generic pata_acpi virtio_net serio_raw virtio_pci virtio_ring virtio ata_piix floppy ipv6 autofs4 [last unloaded: tempesta_db]
[ 5055.181100] CPU: 0 PID: 8017 Comm: sysctl Tainted: G        W  O    4.1.12-ab+ #28
[ 5055.181100] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS Bochs 01/01/2011
[ 5055.181100] task: ffff8800ba4d0d10 ti: ffff8800baa9c000 task.ti: ffff8800baa9c000
[ 5055.181100] RIP: 0010:[<ffffffffa02c8eb4>]  [<ffffffffa02c8eb4>] parse_cfg_entry+0x3f4/0x400 [tempesta_fw]
[ 5055.181100] RSP: 0018:ffff8800baa9fa48  EFLAGS: 00010202
[ 5055.181100] RAX: 0000000000000002 RBX: ffffffffa02c8bd5 RCX: 0000000000000000
[ 5055.181100] RDX: 0000000000000031 RSI: 0000000000000001 RDI: 0000000000000000
[ 5055.181100] RBP: ffff8800baa9fa88 R08: 0000000000000000 R09: ffff880036b8a460
[ 5055.181100] R10: 0000000000000001 R11: 0000000000000000 R12: 0000000000000006
[ 5055.181100] R13: ffff8800baa9fb68 R14: ffff8800baa9fb18 R15: ffff8800baa9fb18
[ 5055.181100] FS:  00007f075cd72740(0000) GS:ffff88013fc00000(0000) knlGS:0000000000000000
[...]

PR #539, opened 2016-05-28
Expect an error when processing a directive within a cfg section. (#538)

keshonok:


tempesta_fw/cfg.c
```diff
@@ -623,6 +623,9 @@ parse_cfg_entry(TfwCfgParserState *ps)
 		PFSM_COND_MOVE(ps->t == TOKEN_LITERAL, PS_MAYBE_EQSIGN);
 		FSM_COND_JMP(ps->t == TOKEN_SEMICOLON, PS_SEMICOLON);
 		FSM_COND_JMP(ps->t == TOKEN_LBRACE, PS_LBRACE);
+
+		ps->err = -EINVAL;
+		FSM_JMP(PS_EXIT);
```
krizhanovsky: It seems there are similar places w/o proper handling of unexpected input. I.e. PS_MAYBE_EQSIGN - do we have the same issue in something like

        sticky name =

or

        srv_group sched= {
            .....

?

Issue #495, opened 2016-05-22
Cache: selective resource caching


krizhanovsky:
We must be able to cache or bypass the cache selectively depending on resource URI. Following configuration must be used to fulfill a request from the cache (also please refer #471 since the extension is required for Web-server mode):

        location prefix "/static/" {
                cache_fulfill suffix ".jpg" ".JPG" ".png" ".PNG";
        }

It's also required to be able to bypass cache for specific URLs:

        location suffix ".php" {
                cache_bypass prefix "/nocache/";
        }

Simple array-based matched should be implemented. Array could be limited by 32 or 64 items.

`location` isn't mandatory grouping. Rather, it's just a base for grouing HTTP management directives.  In this case `location` just help to match prefix and suffix of URI. So `cache_bypass` and `cache_fulfill` can be defined at top level of configuration file.

krizhanovsky:
Lets consider this example which should give an idea how the operation should work:

        1       cache 1;
        2 
        3       cache_fulfill suffix ".css";
        4       cache_bypass prefix "/scripts/";
        5
        6       location prefix "/static/" {
        7               cache_fulfill suffix ".png";
        8               cache_fulfill suffix ".jpg";
        9               cache_bypass prefix "/static/html";
        10      }

1st line defines overall cache mode: switched on/of. If `cache 0`, then just ignore any `cache_fulfill`/`cache_bypass` directives, Tempesta works as pure HTTP proxy w/o any caching.

Lines 3 and 4 defines default policy. If they don't match, then do not do any caching. This means that cache isn't checked at all for the request and corresponding response also isn't stored in the cache.

Few examples:

1. `GET /`: firstly check location - it doesn't match, check lines 3 & 4 - no match, so there is no caching for the request.

2. Add `cache_fulfill * *` at line 5, so this wildcard matches everything, so 1st example will be fulfilled from the cache. In this case we also can safely remove 3rd line since it's overwritten by the new line. Note that sequence of the directives is important, such that if we place the new directive at line 2, then 3rd and 4th lines will have no effect.

3. `GET /static/img/1.png`: `location` matches, 7th line matches, so request is fulfilled from the cache or forwarded to upstream if there is not cached entry and received response is cached;

4. `GET /scripts/do.pl`: `location` doesn't match, check 3rd line - no match, 4th line patches and we bypass the cache;

[...]

Issue #471, opened 2016-05-05
Web-server mode


krizhanovsky:
Web-serve mode must be implemented. The mode can be configured in following way

        srv_group natsys { server 198.199.95.78; }
        srv_group heavy { server 198.199.95.78; } # heavy static & dynamic
        
        vhost tempesta-tech {
        	# URI begining with "?" (dynamic content) pass to heavy server group.
        	location "?" {
        		proxy_pass heavy;
        	}
        	# All others are processed locally.
        	index	index.html;
        	root_db	/opt/tempesta/db/www_tt.tdb
        }
        
        # Virtual hosts
        sched_http_rules {
        	match	natsys		host	suffix	"natsys-lab.com";
        	match	tempesta-tech	host	suffix	"tempesta-tech.com";
        }

So following tasks must be done:

1. http scheduler: suffix matching;

2. virtual host (```vhost```) logic

3. ```location``` logic;

4. handling ```index``` page (Do we actually need to configure index?);

5. ```root_db``` TDB logic (v0.2) without dynamic eviction. Depends on TDB v0.2 (#66, #279 and others) for Web-content management.

krizhanovsky:
Please note https://github.com/tempesta-tech/tempesta/issues/731#issuecomment-377257713 as the syntax is slightly changed.

PR #540, opened 2016-05-28
Implement "suffix" operation for string matching. (#495, #471)

keshonok:


tempesta_fw/str.c
```diff
@@ -410,8 +410,48 @@ tfw_str_eq_cstr_pos(const TfwStr *str, const char *pos, const char *cstr,
...
+		    int cstr_len, tfw_str_eq_flags_t flags)
+{
+	bool ret = false;
+	TfwStr t, tmp = *str;
+	TfwStr *c, *end;
+
+	BUG_ON(TFW_STR_DUP(str));
+	BUG_ON(!cstr || !cstr_len);
+
+	if (offset < 0)
```
krizhanovsky: please use unlikely()

tempesta_fw/str.c
```diff
@@ -410,8 +410,48 @@ tfw_str_eq_cstr_pos(const TfwStr *str, const char *pos, const char *cstr,
...
+		return false;
+	if (offset == 0)
+		return tfw_str_eq_cstr(str, cstr, cstr_len, flags);
+	if (offset + cstr_len > str->len)
```
krizhanovsky: please use unlikely()

tempesta_fw/str.c
```diff
@@ -410,8 +410,48 @@ tfw_str_eq_cstr_pos(const TfwStr *str, const char *pos, const char *cstr,
 }
 EXPORT_SYMBOL(tfw_str_eq_cstr_pos);
 
+bool
+tfw_str_eq_cstr_off(const TfwStr *str, ssize_t offset, const char *cstr,
```
krizhanovsky: Please add header comment for the function and appropriate set of tests in t/unit/test_tfw_str.c

tempesta_fw/http_match.c
```diff
@@ -96,8 +96,8 @@
  *   "Host  :  natsys-lab.com"
  */
 static bool
-hdr_val_eq(const TfwHttpReq *req, tfw_http_hdr_t id, const char *val,
-	   int val_len, tfw_str_eq_flags_t f)
+hdr_val_eq(const TfwHttpReq *req, tfw_http_hdr_t id, tfw_http_match_op_t op,
```
krizhanovsky: Please add appropriate set of tests in t/unit/test_http_match.c

tempesta_fw/str.c
keshonok: In fact, this is not an unlikely event considering how this function is used in comparisons. The length of the pattern is subtracted from the length of the URI path, and it may well be that the URI is shorter in an HTTP message.

krizhanovsky: ok

krizhanovsky:
good to merge

PR #542, opened 2016-05-29
sock: take header reference correctly (fixes #411)

### LINKED ISSUE #411: Memory leak in current Tempesta kernel

milabs:


krizhanovsky:
Looks good for me

PR #543, opened 2016-05-30
cache - location, cache-bypass, cache-fulfill

keshonok:


tempesta_fw/Makefile
```diff
@@ -50,6 +50,7 @@ tempesta_fw-objs = \
 	sock_srv.o \
 	ss_skb.o \
 	stress.o \
+	tfw_tfwcfg.o \
```
krizhanovsky: It's better to name the new module `vhost`, `location` is a part of virtual host logic.

tempesta_fw/tfw_tfwcfg.h
```diff
@@ -0,0 +1,59 @@
...
+} TfwCfgCacheMatch;
+
+typedef struct {
+	tfw_match_t		op;
+	const char		*arg;
+	unsigned int		len;
+	TfwCfgCacheMatch	**cam;
+	unsigned int		cam_sz;
+	unsigned int		cam_max;
+} TfwCfgLocation;
```
krizhanovsky: Lets remove `Cfg` from `TfwCfgCacheMatch` and `TfwCfgLocation` - there is not much about configuration, but mostly about caching per location.

tempesta_fw/cache.c
```diff
@@ -190,6 +191,58 @@ tfw_cache_msg_cacheable(TfwHttpReq *req)
...
 
+/*
+ * Match the given request against strings specified in "location",
+ * "cache_bypass", and "cache_fulfil" directives and according to
+ * match operators specified in the directives.
+ *
+ * Return the constant that identifies the action on a request.
+ * Currently the actions are "bypass", "fulfill", or "default".
+ */
+tfw_stmt_t
```
krizhanovsky: It seems the function should be static.

tempesta_fw/cache.c
```diff
@@ -190,6 +191,58 @@ tfw_cache_msg_cacheable(TfwHttpReq *req)
...
+{
+	TfwCfgLocation *loc;
+	TfwCfgCacheMatch *cam;
+
+	loc = tfw_location_match(&req->uri_path);
+	cam = tfw_camatch_match(loc, &req->uri_path);
+	if (cam)
+		return cam->stmt;
+	return TFW_D_CACHE_DEFAULT;
+}
```
krizhanovsky: Well, I think the things should be done in different way. Now we're doing foundation for #471. If you see at the configuration example, then you notice that there is some _context_, which defines how to process HTTP messages (cache or not, service from local database and so on - just keep in mind `location` directive of Nginx). This means that at least `TfwHttpReq` (but probably `TfwHttpResp` as well) should have a pointer to the context.

For now _context_ means `TfwLocation`, but #471 will introduce _vhost_, the real _context_. Probably you'll find useful to define `TfwVhost` now and the structure will contain array of `TfwLocation`s.

Thus it seems that http.c should call vhost.c (i.e. the function should be moved either to http.c or vhost.c) to determine appropriate vhost for a request. Next, cache.c will also call vhost.[ch] (or just read `TfwVhost` members directly) to get caching policy.

tempesta_fw/tfw_tfwcfg.c
```diff
@@ -0,0 +1,676 @@
...
+tfw_tfwcfg_init(void)
+{
+	int size;
+
+	BUG_ON(tfwcfg_camatch);
+	BUG_ON(tfwcfg_location);
+
+	/* Array of location directives. */
+	size = sizeof(TfwCfgLocation) * TFW_LOCATION_ARRAY_SZ;
+	tfwcfg_location = kzalloc(size, GFP_KERNEL);
```
krizhanovsky: Why to allocate the array dynamically if we can allocate it statically at compile time? Usually compiler do better optimization work if static data is used, so always use static data if possible.

tempesta_fw/tfw_tfwcfg.c
```diff
@@ -0,0 +1,676 @@
...
+		TFW_ERR("Unable to allocate memory"
+			" for location directives: %d bytes\n", size);
+		return -ENOMEM;
+	}
+	tfwcfg_location_sz = 0;
+	tfwcfg_location_max = TFW_LOCATION_ARRAY_SZ;
+
+	/* Array of cache action directives. */
+	size = sizeof(TfwCfgCacheMatch) * TFW_CAMATCH_ARRAY_SZ;
+	tfwcfg_camatch = kzalloc(size, GFP_KERNEL);
```
krizhanovsky: The same - array is static.

tempesta_fw/cache.c
```diff
@@ -190,6 +191,58 @@ tfw_cache_msg_cacheable(TfwHttpReq *req)
...
+ * Currently the actions are "bypass", "fulfill", or "default".
+ */
+tfw_stmt_t
+tfw_cache_action(TfwHttpReq *req)
+{
+	TfwCfgLocation *loc;
+	TfwCfgCacheMatch *cam;
+
+	loc = tfw_location_match(&req->uri_path);
+	cam = tfw_camatch_match(loc, &req->uri_path);
```
krizhanovsky: You check `loc == NULL` in `tfw_camatch_match()` - that's safe. However, `loc == NULL` is expectable, so each time when `tfw_location_match()` returns NULL you do far call just to check the value and return.

tempesta_fw/cache.c
```diff
@@ -190,6 +191,58 @@ tfw_cache_msg_cacheable(TfwHttpReq *req)
...
+	if (cam)
+		return cam->stmt;
+	return TFW_D_CACHE_DEFAULT;
```
krizhanovsky: There is no sense to treat default action as 3rd case - just return `TFW_D_CACHE_BYPASS` by default.

tempesta_fw/cache.c
```diff
@@ -190,6 +191,58 @@ tfw_cache_msg_cacheable(TfwHttpReq *req)
...
+	/*
+	 * Process requests according to "location", "cache_bypass",
+	 * and "cache_fulfill" directives.
+	 */
+	if (!resp) {
+		tfw_stmt_t stmt = tfw_cache_action(req);
+		if ((stmt == TFW_D_CACHE_DEFAULT)
+		    || (stmt == TFW_D_CACHE_BYPASS))
+		{
+			req->cache_ctl.flags |= TFW_HTTP_CC_NO_STORE;
```
krizhanovsky: No! `no-store` is set only when `Connection-Control: no-store` is set, it has nothing to do with our manipulations.

tempesta_fw/cache.c
```diff
@@ -190,6 +191,58 @@ tfw_cache_msg_cacheable(TfwHttpReq *req)
...
+ * the resulting decision.
+ */
+static bool
+tfw_cache_employ(TfwHttpReq *req, TfwHttpResp *resp)
+{
+	/*
+	 * Process requests according to "location", "cache_bypass",
+	 * and "cache_fulfill" directives.
+	 */
+	if (!resp) {
```
krizhanovsky: Why to check `resp` for NULL? If you received a response - check it for cacheability, if you process request - check the request. There is no sense in extra branching.

tempesta_fw/cache.c
```diff
@@ -573,6 +626,8 @@ tfw_cache_add(TfwHttpResp *resp, TfwHttpReq *req)
 
 	if (!cache_cfg.cache || !tfw_cache_msg_cacheable(req))
 		return true;
+	if (!tfw_cache_employ(req, resp))
```
krizhanovsky: `resp` isn't NULL here and you don't need `req` here. You only decide whether `resp` is cacheable here.

tempesta_fw/cache.c
```diff
@@ -611,10 +666,10 @@ tfw_cache_process(TfwHttpReq *req, TfwHttpResp *resp,
 	TfwWorkTasklet *ct;
 	TfwCWork cw;
 
-	if (!cache_cfg.cache || !tfw_cache_msg_cacheable(req)) {
-		action(req, resp);
-		return 0;
-	}
+	if (!cache_cfg.cache || !tfw_cache_msg_cacheable(req))
+		goto dont_cache;
+	if (!tfw_cache_employ(req, resp))
```
krizhanovsky: Yes, this is the only place where we should check `req` and `resp` for NULL because we don't know what we're processing, request or response.

tempesta_fw/tfw_tfwcfg.c
```diff
@@ -0,0 +1,676 @@
...
+ */
+static const char *
+tfw_enum_string(const char * const * enum_string, int enum_max, int enum_val)
+{
+	if ((enum_val < 0) || (enum_val >= enum_max))
+		return enum_string[enum_max];
+	if (enum_string[enum_val])
+		return enum_string[enum_val];
+	return enum_string[enum_max];
+}
```
krizhanovsky: The function is used in only one place, so it's not needed. See comment for `tfw_stmt_string()`.

tempesta_fw/tfw_tfwcfg.c
```diff
@@ -0,0 +1,676 @@
...
+static const char * const __read_mostly __tfwcfg_enum_stmt[] = {
+	[0 ... _TFW_D_COUNT]	= NULL,
+	[TFW_D_CACHE_BYPASS]	= "cache_bypass",
+	[TFW_D_CACHE_FULFILL]	= "cache_fulfill",
+	[_TFW_D_COUNT]		= "UNKNOWN",
+};
+const char *
+tfw_stmt_string(tfw_stmt_t stmt)
+{
+	return tfw_enum_string(__tfwcfg_enum_stmt, _TFW_D_COUNT, stmt);
```
krizhanovsky: Why not just 

     switch (stmt) {
     case TFW_D_CACHE_BYPASS:
            return "cache_bypass";
     case TFW_D_CACHE_FULFILL:
            return "cache_fulfill";
     default:
            return "UNKNOWN";
      }

The code is clear and fast.

tempesta_fw/tfw_tfwcfg.c
```diff
@@ -0,0 +1,676 @@
...
+		return enum_string[enum_val];
+	return enum_string[enum_max];
+}
+
+/* Mappings and functions for configuration directives (statements). */
+static const TfwCfgEnum const __read_mostly __tfwcfg_stmt_enum[] = {
+	{ "cache_bypass",	TFW_D_CACHE_BYPASS },
+	{ "cache_fulfill",	TFW_D_CACHE_FULFILL },
+	{}
+};
```
krizhanovsky: Nobody uses it. Garbage?

tempesta_fw/tfw_tfwcfg.h
```diff
@@ -0,0 +1,59 @@
...
+ * ithe name of the variable short and concise.
+ */
+typedef enum {
+	/* Cache directives. */
+	TFW_D_CACHE_BYPASS,
+	TFW_D_CACHE_FULFILL,
+	TFW_D_CACHE_DEFAULT,
+
+	_TFW_D_COUNT
+} tfw_stmt_t;
```
krizhanovsky: Cache bypass or fulfill is just a binary operation. It's unclear for me why do we need to handle 4-value enum instead of 2-value enum or just boolean type...

tempesta_fw/tfw_tfwcfg.h
```diff
@@ -0,0 +1,59 @@
...
+	tfw_match_t		op;
+	const char		*arg;
+	unsigned int		len;
+	TfwCfgCacheMatch	**cam;
+	unsigned int		cam_sz;
+	unsigned int		cam_max;
+} TfwCfgLocation;
+
+
+const char * tfw_stmt_string(tfw_stmt_t stmt);
```
krizhanovsky: The function isn't used anywhere except tfw_tfwcfg.c

tempesta_fw/cache.c
keshonok: Generally speaking it has not. However, the action is the same: Do not store the response in cache. By RFC, if a request has this option, then the response may not be stored in cache.

tempesta_fw/cache.c
keshonok: Ultimately, it does need the request when making the decision on whether to cache the response or not. Request's Cache-Control arguments that appear pretty soon need to be checked.

tempesta_fw/cache.c
keshonok: Ultimately, the decision for a response needs data from the request. Please see below.

tempesta_fw/tfw_tfwcfg.c
keshonok: I used it a lot while debugging. This function belongs to `TFW_DBG()` statements. I'll put in a few of them in different places.

tempesta_fw/tfw_tfwcfg.c
keshonok: Used in the above function, for debugging.

tempesta_fw/tfw_tfwcfg.c
keshonok: This is a generic function that accepts the array like the one above, and emits the text representation. Highly useful in debugging.

tempesta_fw/tfw_tfwcfg.h
keshonok: Yes, this is clearer now. Just bypass by default.

tempesta_fw/cache.c
krizhanovsky: Sure. The flag is set by RFC, while the functionality isn't about RFC. So we should not mangle the flag in this patch.

tempesta_fw/cache.c
krizhanovsky: Ok, agree. However, the main point was that you do extra check for NULL, but you have the knowledge about ingress arguments in most of the cases.

tempesta_fw/tfw_tfwcfg.c
krizhanovsky: For now the functions are used as they used now. They complex and slow. There is no sense to introduce additional complexity and slowness w/o necessarity.

tempesta_fw/vhost.c
```diff
@@ -0,0 +1,608 @@
...
+};
+
+/*
+ * Find a matching cache policy directive. Strings are compared
+ * according to the match operator in the directive. A pointer
+ * to the matching TfwCaPolicy structure is returned if the
+ * match is found. Null is returned if there's no match.
+ */
+static bool
+__tfw_capolicy_match_fn(TfwCaPolicy *capo, TfwStr *arg)
```
krizhanovsky: The function is used only once in `tfw_capolicy_match()`, so there is no sense in the separate function as well as in temporal variables like `match_fn` and `capo`.

tempesta_fw/vhost.h
```diff
@@ -0,0 +1,82 @@
...
+ * @cmd	- One of defined in tfw_capo_t.
+ * @op	- Match operator: eq, prefix, suffix, etc.
+ * @arg	- String for the match operator.
+ * @len	- Length of the sting in @arg.
+ */
+typedef struct {
+	short		cmd;
+	short		op;
+	const char	*arg;
+	unsigned int	len;
```
krizhanovsky: Please change `arg` and `len` places to avoid memory vasting on alignment

tempesta_fw/vhost.h
```diff
@@ -0,0 +1,82 @@
...
+ * @len		- Length of the sting in @arg.
+ * @capo	- Array of pointers to Cache Policy definitions.
+ * @capo_sz	- Size of @capo array.
+ */
+typedef struct {
+	short		op;
+	const char	*arg;
+	unsigned int	len;
+	TfwCaPolicy	**capo;
+	unsigned int	capo_sz;
```
krizhanovsky: `capo_sz` should be before `capo` to memory vasting on alignment

tempesta_fw/vhost.h
```diff
@@ -0,0 +1,82 @@
...
+ * @loc_sz	- Size of @loc array.
+ * @loc_dflt	- Group of default policies.
+ * @loc_dflt_sz	- Size of @loc_dflt.
+ */
+typedef struct {
+	TfwLocation	*loc;
+	unsigned int	loc_sz;
+	TfwLocation	*loc_dflt;
+	unsigned int	loc_dflt_sz;
+} TfwVhost;
```
krizhanovsky: The same: arange arguments to avoid memory vasting on alignment

tempesta_fw/cache.c
```diff
@@ -190,6 +191,73 @@ tfw_cache_msg_cacheable(TfwHttpReq *req)
...
+	 */
+	loc = vhost->loc_dflt;
+	if (loc && loc->capo_sz) {
+		if ((capo = tfw_capolicy_match(loc, arg)))
+			return capo->cmd;
+	} else {
+		TfwVhost *vhost_dflt = tfw_vhost_get_default();
+		if (vhost == vhost_dflt)
+			return TFW_D_CACHE_BYPASS;
+
```
krizhanovsky: Please comment here or at `tfw_vhost_dflt` initialization that `loc_dflt` is always global default caching policy. This place made me think about couple of minutes what's going on. This is tricky place which we have discussed a lot.

tempesta_fw/cache.c
```diff
@@ -190,6 +191,73 @@ tfw_cache_msg_cacheable(TfwHttpReq *req)
...
+		return false;
+	}
+	/* cache_fulfill - work as usual in cache mode. */
+	BUG_ON(cmd != TFW_D_CACHE_FULFILL);
+
+	return true;
+}
+
+static bool
+tfw_cache_employ_resp(TfwHttpReq *req, TfwHttpResp *resp)
```
krizhanovsky: Argument `resp` is unused

krizhanovsky:
Well done!

PR #545, opened 2016-06-01
 	Handle a response that is terminated by closing the connection.

keshonok:


tempesta_fw/http_parser.c
```diff
@@ -685,29 +686,86 @@ do {									\
...
+	if (msg->content_length) {					\
+		parser->to_read = msg->content_length;			\
+		__FSM_MOVE_nofixup(RGen_BodyStart);			\
+	}								\
+	/* Process the body until the connection is closed. */		\
+	tfw_http_msg_set_data(msg, &msg->body, p);			\
+	__FSM_MOVE_nofixup(Resp_BodyUnlim);				\
+}
+
+#define TFW_HTTP_PARSE_BODY_UNLIM()					\
```
milabs: Why not to include this to the `TFW_HTTP_INIT_RESP_BODY_PARSING`?

tempesta_fw/http.c
```diff
@@ -1055,6 +1146,7 @@ tfw_http_resp_process(TfwConnection *conn, struct sk_buff *skb,
 			TFW_INC_STAT_BH(serv.msgs_parserr);
 			return TFW_BLOCK;
 		case TFW_POSTPONE:
+			((TfwHttpResp *)hmresp)->skb_offset = off;
```
milabs: Looks like ugly workaround. Why not to add `off` and `skb` to the `TfwGState` struct and avoid passing them via the arguments?

tempesta_fw/http_parser.c
```diff
@@ -685,29 +686,86 @@ do {									\
...
+	 * TODO: msg->h_tbl[TFW_HTTP_HDR_TRANSFER_ENCODING]		\
+	 * is present and there's NO "chunked" coding.			\
+	 * Process the body until the connection is closed.		\
+	 */								\
+	if (msg->content_length) {					\
+		parser->to_read = msg->content_length;			\
+		__FSM_MOVE_nofixup(RGen_BodyStart);			\
+	}								\
+	/* Process the body until the connection is closed. */		\
+	tfw_http_msg_set_data(msg, &msg->body, p);			\
```
milabs: As `p` points to the last EOL character, not the first body one, we can't do `tfw_http_msg_set_data` here. Need to `__FSM_MOVE_nofixup` first.

tempesta_fw/http_parser.c
keshonok: Frankly, I didn't want to mix the init body macro with the body processing macro. I agree that these two go side by side, yet I would still like them to be seen as separate. In the code I see a separate `TFW_HTTP_INIT_RESP_BODY_PARSING();`, and then a separate `TFW_HTTP_PARSE_BODY_UNLIM()`, and I know that one is to initialize the parsing of the body, while the other is to actually process the body.

milabs: Is there any cases where those macros will be used separately?

tempesta_fw/http.c
keshonok: Yes, it **is** a workaround. The offset is needed only in the case when the body of a response is terminated by closing of the connection. There's no other case so far where the offset is need after the skb it belongs to has been processed.

The last skb is always the last one in the list for a message, so it's easily reachable anyway, so there's no need to keep a duplicate pointer to it elsewhere. I think. As for the offset, that might be a good idea.

tempesta_fw/http_parser.c
keshonok: Duly noted. Thanks!

tempesta_fw/http_parser.c
keshonok: No, there aren't. In my opinion, this is clarity versus obscurity. Right now it's clearly visible that this is the initialization, and that is the processing. Hiding that doesn't make the code better or more effective, and it doesn't make it clearer. On the contrary, the real case of the processing will be obscurely hidden somewhere in the deep of the source file.

tempesta_fw/http.c
```diff
@@ -974,20 +987,99 @@ tfw_http_popreq(TfwConnection *conn)
...
+		TFW_INC_STAT_BH(serv.msgs_otherr);
+		/* Proceed with processing of the next response. */
+	}
+
+	return 0;
+}
+
+/*
+ *
+ */
```
krizhanovsky: Very meaningful function description ;)

tempesta_fw/http.c
```diff
@@ -446,9 +446,16 @@ tfw_http_conn_release(TfwConnection *conn)
  * Release resources that are not needed anymore, and keep other
  * resources that are needed while there are users of the connection.
  */
+static void tfw_http_resp_terminate(TfwHttpMsg *hm);
```
krizhanovsky: It seems you declared the function here just to avoid `tfw_http_conn_drop()` movement. However, it seems it can be placed below `tfw_http_resp_terminate()`. Please use the declarations only when you absolutely need cyclic dependency (this is the first thing which comes to mind when I see predeclarations).

tempesta_fw/http.c
```diff
@@ -446,9 +446,16 @@ tfw_http_conn_release(TfwConnection *conn)
...
+
 static void
 tfw_http_conn_drop(TfwConnection *conn)
 {
+	if (conn->msg && (TFW_CONN_TYPE(conn) & Conn_Srv)) {
```
krizhanovsky: Hm, what about POST requests? Can they be terminated by connection close as well?

tempesta_fw/http.c
```diff
@@ -974,20 +987,99 @@ tfw_http_popreq(TfwConnection *conn)
...
+	/* Proceed with the error processing */
+error:
+	/*
+	 * Send an error response to the client, otherwise the pairing
+	 * of requests and responses will be broken. If a paired request
+	 * is not found, then something is terribly wrong.
+	 */
+	hmreq = (TfwHttpMsg *)tfw_http_popreq(hmresp);
+	if (unlikely(hmreq == NULL)) {
+		r = TFW_STOP;
```
krizhanovsky: `r` isn't used in return statement

tempesta_fw/http.c
```diff
@@ -955,15 +962,21 @@ tfw_http_resp_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
...
+tfw_http_popreq(TfwHttpMsg *hmresp)
 {
 	TfwMsg *msg;
+	TfwConnection *conn = hmresp->conn;
 
 	spin_lock(&conn->msg_qlock);
 	if (unlikely(list_empty(&conn->msg_queue))) {
 		spin_unlock(&conn->msg_qlock);
+		/* @conn->msg will get NULLed in the process. */
+		TFW_WARN("Paired request missing\n");
```
krizhanovsky: Please add warning about probable HTTP Response Splitting attack.

tempesta_fw/http.c
keshonok: No, I don't think so. According to RFC 7230 3.3.3, only responses may be terminated by closing the connection.

tempesta_fw/http.c
keshonok: Arrrgh... Thanks!

tempesta_fw/http.c
krizhanovsky: Yeah, that's messy hack and according to our discussion we don't need it. Rather we can pass full skb length as `off`

krizhanovsky:
good to merge

PR #546, opened 2016-06-01
Fix bugs in implementation of __hdr_append

milabs:


tempesta_fw/ss_skb.c
```diff
@@ -683,8 +691,10 @@ __skb_fragment(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
 	d_size = skb_headlen(skb);
 	offset = pspt - (char *)skb->data;
 
-	if ((offset >= 0) && (offset < d_size)) {
+	if ((offset >= 0) && (offset <= d_size)) {
 		int t_size = d_size - offset;
+		if (!t_size)
+			goto append;
```
keshonok: Pephaps, `unlikely()` would be helpful here as this should be a rare situation.

tempesta_fw/ss_skb.c
```diff
@@ -712,23 +722,25 @@ __skb_fragment(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
 		d_size = skb_frag_size(frag);
 		offset = pspt - (char *)skb_frag_address(frag);
 
-		if ((offset >= 0) && (offset < d_size)) {
+		if ((offset >= 0) && (offset <= d_size)) {
 			int t_size = d_size - offset;
+			if (!t_size)
```
keshonok: Pephaps, `unlikely()` would be helpful here as this should be a rare situation.

tempesta_fw/ss_skb.c
```diff
@@ -741,17 +753,29 @@ __skb_fragment(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
 		return ret;
 	if ((it->ptr == NULL) || (it->skb == NULL))
 		return -EFAULT;
-	it->len = len;
 
 	/* Return the length of processed data. */
-	return abs(len);
+	return (it->len = abs(len));
```
keshonok: I wouldn't use an assignment in a `return()`. The syntax is correct, of course, but I find this practice doubtful.

tempesta_fw/ss_skb.c
```diff
@@ -691,8 +691,10 @@ __skb_fragment(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
 	d_size = skb_headlen(skb);
 	offset = pspt - (char *)skb->data;
 
-	if ((offset >= 0) && (offset < d_size)) {
+	if ((offset >= 0) && (offset <= d_size)) {
 		int t_size = d_size - offset;
+		if (unlikely(!t_size))
```
krizhanovsky: Technically this is the same as point to first byte of first fragment, so why don't you consider the fast path at line 709?

tempesta_fw/ss_skb.c
```diff
@@ -697,12 +715,13 @@ __skb_fragment(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
 	 * advance the skb tail pointer.
 	 */
 	if (len > 0) {
+		if (unlikely((offset == d_size)))
+			pspt = (char *)skb_frag_address(&si->frags[0]);
 		offset = pspt - (char *)skb_frag_address(&si->frags[0]);
```
krizhanovsky: Basically it's just equal to

         offset = (unlikely((offset == d_size))) ? 0 : pspt - (char *)skb_frag_address(&si->frags[0]);

So there is no need to calculate `pspt` just to get zero at next line.

PR #548, opened 2016-06-01
Fix CID 78557

milabs:


tempesta_fw/http_msg.c
```diff
@@ -841,6 +841,12 @@ tfw_http_msg_alloc(int type)
 	ss_skb_queue_head_init(&hm->msg.skb_list);
 
 	hm->h_tbl = (TfwHttpHdrTbl *)tfw_pool_alloc(hm->pool, TFW_HHTBL_SZ(1));
+	if (!hm->h_tbl) {
```
keshonok: Perhaps, s simpler and faster `tfw_pool_destroy(m->pool);` would suffice here?
Also, perhaps the statement above `ss_skb_queue_head_init(&hm->msg.skb_list);` should better be moved lower and reside after the memory allocation.

We don't have a message yet that is known to the outside of this little function. But calling `tfw_http_msg_free(hm);` would tell that we're freeing a known message (that really doesn't exist yet in all meaningful terms). Extra generic actions would be executed on the message. And a debug message would be printed about freeing a message that didn't appear anywhere yet.

milabs: Ok, but what's about `hm` itself?

milabs: Ok, I got it.

tempesta_fw/http_msg.c
```diff
@@ -838,13 +838,18 @@ tfw_http_msg_alloc(int type)
 	if (!hm)
 		return NULL;
 
-	ss_skb_queue_head_init(&hm->msg.skb_list);
-
 	hm->h_tbl = (TfwHttpHdrTbl *)tfw_pool_alloc(hm->pool, TFW_HHTBL_SZ(1));
+	if (!hm->h_tbl) {
```
krizhanovsky: please use `unlikely()`

tempesta_fw/http_msg.c
```diff
@@ -838,13 +838,18 @@ tfw_http_msg_alloc(int type)
...
+		TFW_WARN("Insufficient memory to create message\n");
+		tfw_pool_destroy(m->pool);
```
krizhanovsky: must be `hm`, not `m`.

tempesta_fw/http_msg.c
```diff
@@ -838,13 +838,18 @@ tfw_http_msg_alloc(int type)
 	if (!hm)
```
krizhanovsky: Please put `unlikely()` here as well.

krizhanovsky:
good to merge with the fixe and cleanups

Issue #526, opened 2016-05-26
[RFC 7234] Process Pragma header


krizhanovsky:
[RFC 7234 5.4](https://tools.ietf.org/html/rfc7234#section-5.4) requires `Pragma` header processing. This is just a subset of `Cache-Control` processing, so just a simple copy & paste.

keshonok:
Implemented in 0b80761.

PR #549, opened 2016-06-02
Process the Pragma header field in requests. (#526)

keshonok:


milabs:
Looks good.

tempesta_fw/http.h
```diff
@@ -265,6 +265,7 @@ typedef struct {
 	unsigned long		tm_header;
 	unsigned long		tm_bchunk;
 	unsigned long		hash;
+	__u8			pragma_no_cache:1;
```
krizhanovsky: I just noticed that we process only 3 request types, so 1 byte is enough to keep `method`, so `pragma_no_cache` can be placed near to `method` and we can put both of them into 2 bytes instead of 16 bytes as now (enum is 4 bytes).

Also why not to introduce new TFW_HTTP_CC_PRAGMA_NO_CACHE flag and use `TfwCacheControl`?

PR #550, opened 2016-06-02
SSE parser

dmleontiev9000:
current version of parser is located in files http_parser.h, http_parser.c, http_parser_impl.c

http_parser_impl.c contains body of parser function, which can be tuned by external definitions.
right now FAST_FORWARD option is supported, which enabled parser to quickly "scroll" over long headers.

sse_parser.c/h should be considered broken. it will be removed as soon as all algorithms will be ported to new version


tempesta_fw/t/unit/sse_parser/sse_parser.c
```diff
@@ -0,0 +1,1552 @@
...
+                0x00, 0x00, 0x20, 0x20, 0x00, 0x00, 0x20, 0x20);
+    c->dq2 = _mm_setr_epi8(
+                0x2F, 0x39, 0x60, 0x66, 0x2F, 0x39, 0x60, 0x66,
+                0x2F, 0x39, 0x60, 0x66, 0x2F, 0x39, 0x60, 0x66);
+    c->dq3 = _mm_set1_epi16(0x00FF);
+    c->dq4 = _mm_setr_epi8(
+                0x30, 0xFF, 0x57, 0xFF, 0x30, 0xFF, 0x57, 0xFF,
+                0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x07, 0x00);
+    c->dq5 = _mm_setr_epi16(
+                16, 16, 1, 1, 0, 0, 0, 0);
```
krizhanovsky: What the constants mean and what do they have to do with quotes?

tempesta_fw/t/unit/sse_parser/sse_parser.c
```diff
@@ -0,0 +1,1552 @@
...
+                          '/', '/',  0 ,  0 ,
+                           0 ,  0 ,  0 ,  0 );
+
+    static const char * http_m[] = {
+        "GET", "PUT", "POST", "COPY", "MOVE", "LOCK", "HEAD", "PATCH",
+        "TRACK", "DELETE", "UNLOCK", "MKCOL", "OPTIONS", "PROPFIND", "PROPPATCH",
+        NULL
+    };
+    XASSERT(sizeof(c->http_method) >= tokenSetLength(http_m));
+    initTokenSet(http_m, &c->http_method, sizeof(c->http_method));
```
krizhanovsky: NULL return code isn't checked.

tempesta_fw/t/unit/sse_parser/sse_parser.c
```diff
@@ -0,0 +1,1552 @@
...
+    XASSERT(sizeof(c->http_schema) >= tokenSetLength(http_sch));
+    initTokenSet(http_sch, &c->http_schema, sizeof(c->http_schema));
+
+    static const char * http_v[] = {
+        "\n", "\r\n", "HTTP/1.0\n", "HTTP/1.0\r\n",
+        "HTTP/1.1\n", "HTTP/1.1\r\n", NULL
+    };
+    XASSERT(sizeof(c->http_version) >= tokenSetLength(http_v));
+    initTokenSet(http_v, &c->http_version, sizeof(c->http_version));
+
```
krizhanovsky: Well, `http_m`, `http_sch` and `http_v` contains very short strings,
so there is no sense to use vector processing. Your becnhmarks
implicitly show this: GOTO outperforms SSE-ff on short strings.
Also keep in ming that `GET / HTTP/1.0\n\n` is a valid request which
we can face on HTTP Flood and we need maximum performance in this case
as well as in normal case with long URIs and Cookies.
Therefore, resulting parser must combine both the approaches and use
vectorized processing for long strings, i.e. URIs and headers. Probably,
you find a heuristic whether to use vectorized parser processing.
I'd imagine it could be length of rest of ingress `data`.
Typically HTTP messages don't contain many spaces sequently, so there is
also no need to use SSE for the processing. However, it'd be good TODO
for Frang to catch such divitating messages... For now we can cath such
requests by limiting size of HTTP fields.

tempesta_fw/t/unit/sse_parser/sse_parser.c
```diff
@@ -0,0 +1,1552 @@
...
+                mask2);
+    __m128i mask4 = _mm_and_si128(
+                mask3,
+                mask1);
+    __m128i vec1 = _mm_cmpgt_epi8(
+                mask4,
+                _mm_setzero_si128());
+    __m128i vec2 = _mm_cmplt_epi8(
+                mask4,
+                _mm_setzero_si128());
```
krizhanovsky: unnecessary line breaks between function arguments are hard to read and doesn't comply our Coding Style.

tempesta_fw/t/unit/sse_parser/sse_parser.c
```diff
@@ -0,0 +1,1552 @@
...
+        {0x80, 0xc0, 0xe0, 0xf0}
+    };
+    //check if buf is aligned
+    if (((long)buf) & 0xF) return NULL;
+
+    TokenSet * ts = (TokenSet*)buf;
+    //count number of words required for map
+    int i, j, k, p, nwords;
+    for(i = 0, k = 0; tokens[i]; ++i) {
+        for(j = 0; tokens[i][j]; ++j);
```
krizhanovsky: This is just sizeof(tokens[i])-1, which can be calcultated statically.

tempesta_fw/t/unit/sse_parser/sse_parser.c
```diff
@@ -0,0 +1,1552 @@
...
+        k += (j+3)>>2; 
+    }
+    //round number of words 
+    nwords = (k+3)>>2;
```
krizhanovsky: `nwords` calculation is almost fully copy&pasted from `tokenSetLength()`

tempesta_fw/t/unit/sse_parser/sse_parser.c
```diff
@@ -0,0 +1,1552 @@
...
+            bytes[k+48] = 0;
+
+            if ((k & 0xF) == 0xF) {
+                k = k + 49;
+            } else {
+                k = k + 1;
+            }
+        }
+
+        //WEIRD CODE!!!!
```
krizhanovsky: senseless bad comment. Basically you should have been comment the loop calculations and `cmap` values.

tempesta_fw/t/unit/sse_parser/sse_parser.c
dmleontiev9000: These constants are for SSE decoder of %xx sequences(%20 for example)

dq0 constant expands 1st and 2nd byte in 4 positions each.

we need 4 copies to perform 4 comparisons c>='0', c<='9', c>='a', c<='f'

dq1 is to lowercase positions which may be  [A-F]

dq2 is actual comparison mask ('0'-1, '9'+1, 'a'-1, 'f'+1)

if comparison was ok, we will get a 00FF word (dq3 is to ensure we found it)

dq4 is used to convert value obtained by dq0-dq1 and masking with result of
dq2-dq3 to decimal values

dq5 is used to combine upper and lower hex symbols into a hex code

2016-06-14 14:17 GMT+04:00 Alexander Krizhanovsky <notifications@github.com>
:

dmleontiev9000: sse_parser should be considered out-dated right now. It's Constants
structure is derived from the structure used in 1st parser version I
developed in April. There may be unused constants, because several
algorithms related to uri normalization are not implemented yet(these
algorithms have to be rewritten)

I've realized the complexity of it's integration into tempesta is too high:
too much checks are needed to keep fixups in proper positions
too much state is to be saved on each iteration
too much overhead
too much SSE code in FSM

http_parser, on other hand, uses few variables to maintain it's state
between iterations

vec = currently observed bytes(x16), including bytes from previous packes
if any
bytes_cached = number of valid bytes in vec, starting from byte 0
bytes_shifted = number of consumed bytes in vec, starting from byte 0

Example:

Let we have 16 bytes in vec: "GET http://host/"
On start of 1st iteration we have bytes_cached = 16, bytes_shifted = 0;
Req_Method iteration eats 3 bytes+space
On start of 2nd iteration we have bytes_cached = 16, bytes_shifted = 4;

Read logic shifts vec 4 bytes right, and inserts 4 new bytes at the left of
the vec.

Space skipping and character set matching algorithms are described in
description.pdf

Number of matched characters is used to determine the number of bytes FSM
can consume at once,
and to get the first byte which does not match character set.:

input="Content-lenght:" --> nbytes = strlen("Content-lenght"); lastchar =
input[nbytes];

2016-06-14 14:24 GMT+04:00 Дмитрий Леонтьев <dm.leontiev7@gmail.com>:

dmleontiev9000: >Well, http_m, http_sch and http_v contains very short strings,

so there is no sense to use vector processing.

In http parser(http_parser.c/http_parser_impl.c), where we expect only
http(https may be added) we use simplier algorithm which
compares as much as possible and returns a bitmask. We expect the data to
come in long
chunks, so we will get "http://" at once. Length of "http://" is 7 bytes so
the comparison mask
will likely have lower 7 bits set to ones.

>keep in ming that GET / HTTP/1.0\n\n is a valid request which
we can face on HTTP Flood

I tried an optimization for that case, but it showed no performance
increase or impact.

>Typically HTTP messages don't contain many spaces sequently
in http_parser, spaces are skipped in tight loop and don't hit performance.

In case of initTokenSet, we use XASSERT macro to ensure we have allocated
enough data.
There is no need to check for NULL. initTokenSet should be run once to
initialize constants.

2016-06-14 14:46 GMT+04:00 Дмитрий Леонтьев <dm.leontiev7@gmail.com>:

krizhanovsky:
Senseless patch

Issue #503, opened 2016-05-24
Tempesta's modified SKB data may get corrupted in skb_linearize()


keshonok:
When modifying SKB's data, Tempesta may map a piece of skb's linear data to a paged fragment. That leads to the following complications that may result in data corruption.
- Once that kind of data mapping is done, any subsequent skb modification cannot use the fast path optimization where `skb->tail` is moved forward towards `skb->end`.
- In certain cases in the egress path the skb may get linearized (`skb_linearize()`). Underneath of that `__pskb_pull_tail()` is used. The function may use the area between `skb->tail` and `skb->end` when copying data. If there's skb's data there that had been mapped to a paged fragment, then the data is corrupted in the process.

This lends itself to a flag that is a property of an skb itself as it need to travel across the Linux kernel along with the skb it belongs to. Given this flag, `skb_linearize()` can be modified to take care of this issue when Tempesta kernel is running.

Also, this flag needs to be checked when the above mentioned fast path modification is possible.

At first glance, there's a straightforward way of checking whether a piece of linear data had been mapped to a paged fragment:
```C
static inline bool
__skb_head_is_mapped(struct sk_buff *skb)
{
        struct page *pg_head;

        if (!skb_is_nonlinear(skb))
                return;

        pg_head = virt_to_head_page(skb->head);
        for (i = 0; i < skb_shinfo(skb)->nr_frags; ++i) {
                skb_frag_t *f = &skb_shinfo(skb)->frags[i];
                if (skb_frag_page(f) == pg_head)
                        return true;
        }

        return false;
}
```

However this is a heavyweight check that may spend lots of cycles. Also, one can imagine a case (though highly improbable), where the mapped paged fragment is shifted/moved to the next skb. In that case, this check would yield incorrect result.

A simple check for a flag would be significantly faster and always correct.

I think that a good place for the flag is `skb_shinfo(skb)->tx_flags`. There's one bit available there that can be used for that purpose. The flag can be called `SKBTX_MAPPED_HEAD` which is in line with the other names there. Then, a simple check for that flag would yield the desired result, and it can be used in both `ss_skb.c` in the fast path case, and in `__pskb_pull_tail()`.

milabs:
Unfortunately, we can't use `skb_shinfo(skb)->tx_flags` as no free bits there. Taking a look at `skb->peeked` instead.

milabs:
Fixed via:
23ef7571af604ddd0637f0a030eb5ff3a61a1fba

[...]

PR #553, opened 2016-06-03
ss-skb: prevent expanding tail of splitted SKB

milabs:
When modifying SKB's data, Tempesta may map a piece of skb's linear
data to a paged fragment. That leads to the following complications
that may result in data corruption:

 - Once that kind of data mapping is done, any subsequent skb
   modification cannot use the fast path optimization where skb->tail
   is moved forward towards skb->end.

This fixes the problem by adding tail locking ability when skb->peeked
is used as a locked/unlocked flag.

Related to #503.

### DIFF
--- a/tempesta_fw/ss_skb.c
+++ b/tempesta_fw/ss_skb.c
@@ -433,6 +433,9 @@ __split_linear_data(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
 	__skb_fill_page_desc(skb, alloc, page, tail_off, tail_len);
 	get_page(page);
 
+	/* Prevent @skb->tail from moving forward */
+	skb->tail_lock = 1;
+
 	/*
 	 * Get the SKB and the address for data. It's either
 	 * the area for new data, or data after the deleted data.
@@ -659,7 +662,7 @@ __split_pgfrag(struct sk_buff *skb, int i, int off, int len, TfwStr *it)
 static inline int
 __split_try_tailroom(struct sk_buff *skb, int len, TfwStr *it)
 {
-	if (len > skb_tailroom(skb))
+	if (len > skb_tailroom_locked(skb))
 		return -ENOSPC;
 	it->ptr = ss_skb_put(skb, len);
 	it->skb = skb;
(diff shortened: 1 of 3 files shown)


tempesta_fw/ss_skb.c
```diff
@@ -420,6 +420,9 @@ __split_linear_data(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
 	__skb_fill_page_desc(skb, alloc, page, tail_off, tail_len);
 	get_page(page);
 
+	/* Prevent futher @skb->tail expanding */
+	__ss_skb_tail_lock(skb);
```
krizhanovsky: Shouldn't we rather just put a check here `WARN_ON(__ss_skb_tail_is_locked(skb))`?

tempesta_fw/ss_skb.c
```diff
@@ -698,7 +701,8 @@ __skb_fragment(struct sk_buff *skb, char *pspt, int len, TfwStr *it)
 	 */
 	if (len > 0) {
 		offset = pspt - (char *)skb_frag_address(&si->frags[0]);
-		if (unlikely(!offset && (len <= ss_skb_tailroom(skb)))) {
+		if (unlikely(!offset && (len <= ss_skb_tailroom(skb)))
+		    && !__ss_skb_tail_is_locked(skb)) {
```
krizhanovsky: IIRC there is kernel code which manipulates with the tail pointer, so shouldn't we rather patch `ss_skb_put()` and `skb_put()`?

tempesta_fw/ss_skb.h
```diff
@@ -208,4 +208,34 @@ int ss_skb_cutoff_data(SsSkbList *skb_list,
...
 		   ss_skb_actor_t actor, void *objdata);
 
+/**
+ * While splitting SKB with @skb_fragment we need to lock skb->tail from
+ * being expanded in case of linear part fragmentation tooks place. We use
+ * skb->ipvs_property as a marker as that field is used only for IPVS and
+ * there is no collision with the kernel.
+ *
+ * Note, that in case of using IPVS with the Tempesta this hack MUST be
+ * verified.
```
krizhanovsky: We try to be a good kernel citizenship, while IPVS is very useful thing which is good to integrate with. Probably it's better just to introduce one more flag somewhere near from `skb_page` - we can loose up to 8 bytes, but it's better than break system interoperability.

PR #554, opened 2016-06-04
sock: drop zero-sized SKBs while in ss_do_send

### LINKED ISSUE #551: Remove empty skbs before sending a message to a recipient
It may occur after message header modifications, that the list of skbs for the message contains skbs that have zero length of data. One extreme example when that may occur is this. Imagine that data comes slowly, byte by byte. In that case each separate byte may come to Tempesta in a separate skb. If a header is deleted in such skb, we get a bunch of empty skbs.

These empty skbs need to be removed from the list before sending a message out. Perhaps the best place for that is `ss_do_send()`.

milabs:
Fixes #551.


krizhanovsky:
good for me

PR #556, opened 2016-06-10
Storing Responses to Authenticated Requests (#3)

avbelov23:


tempesta_fw/cache.c
```diff
@@ -252,10 +252,17 @@ tfw_cache_employ_req(TfwHttpReq *req)
...
 static bool
-tfw_cache_employ_resp(TfwHttpReq *req)
+tfw_cache_employ_resp(TfwHttpResp *resp, TfwHttpReq *req)
 {
 	if (req->cache_ctl.flags & TFW_HTTP_CC_CFG_CACHE_BYPASS)
 		return false;
+	if (!TFW_STR_EMPTY(&req->h_tbl->tbl[TFW_HTTP_HDR_AUTHORIZATION]) &&
+	    !(resp->cache_ctl.flags & TFW_HTTP_CC_PUBLIC) &&
+	    !(resp->cache_ctl.flags & TFW_HTTP_CC_S_MAXAGE) &&
+	    !(resp->cache_ctl.flags & TFW_HTTP_CC_MUST_REV))
```
krizhanovsky: Cache control directives are out of scope of the issue and aren't correct: we should not use cached object only is age of the cache entry reaches specified max age while presence of max age doesn't mean that the response can't be cached. In particular max age equal 0 means this, so you should have been check the value.

tempesta_fw/http.h
```diff
@@ -177,6 +178,7 @@ typedef enum {
 
 	TFW_HTTP_HDR_CONNECTION = TFW_HTTP_HDR_NONSINGULAR,
 	TFW_HTTP_HDR_X_FORWARDED_FOR,
+	TFW_HTTP_HDR_AUTHORIZATION,
```
krizhanovsky: There is no sense to make the header special - we don't process it's content at all.

tempesta_fw/cache.c
avbelov23: RFC 2616, chapter 14.8 (see issue #3 ):
"If the response includes "s-maxage=0", the proxy MUST always revalidate it before re-using it."
I believe that this shows that the cache is possible, but must be revalidation

keshonok:
Major part of this pull request has been merged into #559. Thanks!

PR #557, opened 2016-06-11
rework __lookup_pgfrag_room

milabs:


tempesta_fw/ss_skb.c
```diff
@@ -92,72 +93,93 @@ ss_skb_alloc_pages(size_t len)
...
- * @return pointer to the last fragment in the page.
+ * Find a page that has room for @len bytes.
+ * Return true if the room is found, or false otherwise.
+ * If the room is found, fill @f_out that describes the location of the room.
+ *
+ * The assumption is that there can be sufficient room available at the
+ * start or at the end of a page that holds this SKB's paged fragments.
+ *
+ * For a group of paged fragments located in the same page the available
+ * headroom and tailroom are calculated. If @len is less than both of
```
krizhanovsky: "...If `@len` is **more** than both of..."?

milabs: yep

tempesta_fw/ss_skb.c
```diff
@@ -92,72 +93,93 @@ ss_skb_alloc_pages(size_t len)
...
-			return frag;
+		for (k = i; k < skb_shinfo(skb)->nr_frags; k++) {
+			if (map[k])
+				continue;
+
+			f_this = &skb_shinfo(skb)->frags[k];
+			if (compound_head(skb_frag_page(f_this)) != p_base)
+				continue;
+
+			if ((map[k] = 1, map[i]) > 1)
```
krizhanovsky: `map[k]` and `map[i]` are different values, so assignment to first one and getting value of the second one isn't a good idea. Please don't love `,` operator so much.

milabs: It's shortcut for:
```
map[k] = 1;
if (map[i]) > 1)
    continue;
```

tempesta_fw/ss_skb.c
```diff
@@ -92,72 +93,93 @@ ss_skb_alloc_pages(size_t len)
...
+				continue;
+
+			h_room = min(h_room, f_this->page_offset);
+			t_room = min(t_room, p_size - ss_skb_frag_len(f_this));
+			map[i] = map[i] + !!(len > h_room && len > t_room);
```
krizhanovsky: `map[i] += !!(len > h_room && len > t_room);`?

tempesta_fw/ss_skb.c
```diff
@@ -92,72 +93,93 @@ ss_skb_alloc_pages(size_t len)
...
+			if (map[k])
+				continue;
+
+			f_this = &skb_shinfo(skb)->frags[k];
+			if (compound_head(skb_frag_page(f_this)) != p_base)
+				continue;
+
+			if ((map[k] = 1, map[i]) > 1)
+				continue;
+
```
krizhanovsky: The first iteration of the loop always goes here and it doesn't need to pass the 3 conditions, so it's better just to initialize `map[i]` before the loop.

tempesta_fw/ss_skb.c
krizhanovsky: I see, but the operations have little to do with each other.

krizhanovsky:
Good to merge with small cleanups and improvements

Issue #501, opened 2016-05-23
[Cache] purging


krizhanovsky:
Manual cache purging must be implemented. An example of the purging is

        $ curl -X PURGE <URL>

which sends

        PURGE / HTTP/1.1\r\n
        User-Agent: curl/7.40.0\r\n
        Host: <host>\r\n
        Accept: */*\r\n
        \r\n

Following new configuration options must be introduced:

* `cache_purge` - cache purging mode, binary variable for first implementation, but can be extended in future (e.g. setting eviction strategy and speed for purged entries). `"invalidate"` value (default and the first implemented) means that purged entries are just invalidated and still can be returned to a client under certain conditions (#522, note that for 0.5 we never return stale responses, so this is just fake purging). `"purge"` means real purging backed by TDB deletion operator (this mode depends on #516).

* `cache_purge_acl` - control who can purge the cache by list of IP addresses or netmasks.

See [Nginx](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_purge), [Varnish](https://www.varnish-cache.org/docs/3.0/tutorial/purging.html) and [Squid](http://wiki.squid-cache.org/SquidFaq/OperatingSquid#How_can_I_purge_an_object_from_my_cache.3F) implementations as reference examples.

Wildcard purging is not necessary at this time.

**Use case.** If some content is changed at upstream server, then PURGE request followed by GET request will update appropriate entry in the cache.

keshonok:
Implemented in #558.

PR #558, opened 2016-06-15
Implement manual cache purging using HTTP PURGE method. (#501)

keshonok:


krizhanovsky:
This is new functionality, so please update cache chapter in README.md

tempesta_fw/cache.c
```diff
@@ -950,8 +972,52 @@ cache_req_process_node(TfwHttpReq *req, unsigned long key,
...
+ */
+static int
+tfw_cache_purge_method(TfwHttpReq *req, unsigned long key)
+{
+	TfwVhost *vhost = tfw_vhost_get_default();
+
+	if (!vhost->cache_purge)
+		return tfw_http_send_403((TfwHttpMsg *)req);
+
+	if (vhost->cache_purge_acl) {
```
krizhanovsky: Let's deny PURGE request by default: if admin forgot to set configure it he figure it out quickly otherwise default accepting PURGE request can open security flaws.

tempesta_fw/cache.c
```diff
@@ -950,8 +972,52 @@ cache_req_process_node(TfwHttpReq *req, unsigned long key,
...
+	TdbIter iter;
+	TDB *db = node_db();
+	TfwCacheEntry *ce = NULL;
+
+	if (!(ce = tfw_cache_dbce_get(db, &iter, req, key))) {
+		tfw_cache_dbce_put(ce);
+		return -ENOENT;
+	}
+
+	/* ce->lifetime = 0; */
```
krizhanovsky: Why the function doesn't do nothing? Now we can and we should expire the entry.

keshonok: It will after integration with the other branch.

tempesta_fw/vhost.c
```diff
@@ -521,6 +555,176 @@ tfw_cleanup_locache(TfwCfgSpec *cs)
...
+			return -EINVAL;
+		prefix += 128 - 32;
+		addr->family = AF_INET6;
+	} else if (prefix > 128) {
+		return -EINVAL;
+	}
+	addr->in6_prefix = prefix;
+
+	return 0;
+}
```
krizhanovsky: The two functions above accept `const char` and `TfwAddr`, so it seems it's better to move them to addr.[ch] and rename them such that they parse netmasks becomes clear.

tempesta_fw/vhost.h
```diff
@@ -60,6 +61,12 @@ typedef struct {
 	TfwCaPolicy	**capo;
 } TfwLocation;
 
+/* Cache purge configuration modes. */
+enum {
+	TFW_D_CACHE_PURGE_INVALIDATE,
+	TFW_D_CACHE_PURGE_PURGE,
```
krizhanovsky: Probably `TFW_D_CACHE_PURGE_DELETE` is more clear than `TFW_D_CACHE_PURGE_PURGE`.

tempesta_fw/vhost.c
```diff
@@ -521,6 +555,176 @@ tfw_cleanup_locache(TfwCfgSpec *cs)
...
+	}
+	if (!ce->val_n) {
+		/* Default value for the cache_purge directive. */
+		vhost->cache_purge_mode = TFW_D_CACHE_PURGE_INVALIDATE;
+		goto done;
+	}
+	TFW_CFG_ENTRY_FOR_EACH_VAL(ce, i, val) {
+		if (!strcasecmp(val, "invalidate")) {
+			vhost->cache_purge_mode = TFW_D_CACHE_PURGE_INVALIDATE;
+		} else if (!strcasecmp(val, "purge")) {
```
krizhanovsky: I think `cache_purge = purge` isn't clear enough and it'd be better to rename the option to `delete`, so one reading `cache_purge = delete` get more clear understanding what's going on.

PR #559, opened 2016-06-15
Implement various cache-related features

keshonok:


tempesta_fw/http_parser.c
```diff
@@ -2602,168 +2723,252 @@ __resp_parse_expires(TfwHttpResp *resp, unsigned char *data, size_t len)
...
+__resp_parse_age(TfwHttpResp *resp, unsigned char *data, size_t len)
+{
+	int r = CSTR_NEQ;
+	__FSM_DECLARE_VARS(resp);
+
+	__FSM_START(parser->_i_st) {
+
+	__FSM_STATE(Resp_I_Age) {
+		/* Eat OWS before the node ID. */
+		if (unlikely(IS_WS(c)))
```
milabs: No need to check for `IS_WS(c)` here as `RGen_LWS` eats all the spaces before the token.

tempesta_fw/cache.c
```diff
@@ -59,6 +64,14 @@ typedef struct {
...
 	unsigned int	hdr_num;
 	unsigned int	hdr_len;
+	unsigned int	method: 4;
+	unsigned int	flags: 28;
+	time_t		age;
+	time_t		date;
+	time_t		expires;
+	time_t		req_time;
+	time_t		resp_time;
+	time_t		lifetime;
```
krizhanovsky: Please update header comment as well.

tempesta_fw/cache.c
```diff
@@ -248,17 +261,120 @@ tfw_cache_employ_req(TfwHttpReq *req)
...
+	{
+		time_t lt_max_age, lt_max_stale, lt_min_fresh;
+
+		lt_max_age = lt_max_stale = lt_min_fresh = UINT_MAX;
+		if (req->cache_ctl.flags & TFW_HTTP_CC_MAX_AGE)
+			lt_max_age = req->cache_ctl.max_age;
+		if (req->cache_ctl.flags & TFW_HTTP_CC_MAX_STALE)
+			lt_max_stale += req->cache_ctl.max_stale;
+		if (req->cache_ctl.flags & TFW_HTTP_CC_MIN_FRESH)
+			lt_min_fresh -= req->cache_ctl.min_fresh;
```
keshonok: Of course this should be `=`, and not `+=` or `-=`.

tempesta_fw/cache.c
```diff
@@ -248,17 +261,120 @@ tfw_cache_employ_req(TfwHttpReq *req)
...
+		ce_lifetime = min(min(lt_max_age, lt_max_stale), lt_min_fresh);
+	}
+
```
keshonok: `ce_lifetime = min(ce_lifetime, ce->lifetime);` ???

And `ce_lifetime` shuld be initialized as:
```C
time_t ce_lifetime = UINT_MAX;
```

tempesta_fw/cache.c
```diff
@@ -41,6 +41,11 @@
 #warning "Please set CONFIG_NODES_SHIFT to less than 16"
 #endif
 
+/* Flags stored in a Cache Entry. */
+#define TFW_CE_MUST_REVAL	0x0001		/* MUST revalidate if stale. */
+#define TFW_CE_NO_CACHE		0x0002		/* ALWAYS revalidate. */
```
krizhanovsky: Why do we need the flag? Responses which always must be revalidated shouldn't be stored in cache: if we receive such response then we just bypass the cache.

tempesta_fw/cache.c
```diff
@@ -41,6 +41,11 @@
...
+#define TFW_CE_INVALID		0x0100
```
krizhanovsky: The flag probably has sense, but is unused for now.

tempesta_fw/cache.c
```diff
@@ -248,17 +261,120 @@ tfw_cache_employ_req(TfwHttpReq *req)
...
 static bool
-tfw_cache_employ_resp(TfwHttpReq *req)
+tfw_cache_employ_resp(TfwHttpReq *req, TfwHttpResp *resp)
 {
+	static const unsigned int __read_mostly cacheit =
+		TFW_HTTP_CC_HDR_EXPIRES | TFW_HTTP_CC_MAX_AGE
+		| TFW_HTTP_CC_S_MAXAGE | TFW_HTTP_CC_PUBLIC;
+	static const unsigned int __read_mostly authcan =
+		TFW_HTTP_CC_S_MAXAGE | TFW_HTTP_CC_PUBLIC
+		| TFW_HTTP_CC_MUST_REVAL | TFW_HTTP_CC_PROXY_REVAL;
```
krizhanovsky: Usually defines are used in kernel for such masks - technically should be the same, but just more convenient form.

tempesta_fw/cache.c
```diff
@@ -248,17 +261,120 @@ tfw_cache_employ_req(TfwHttpReq *req)
...
+
 	if (req->cache_ctl.flags & TFW_HTTP_CC_CFG_CACHE_BYPASS)
 		return false;
+
+	if ((req->cache_ctl.flags|resp->cache_ctl.flags) & TFW_HTTP_CC_NO_STORE)
```
krizhanovsky: We use spaces around operators.

tempesta_fw/cache.c
```diff
@@ -248,17 +261,120 @@ tfw_cache_employ_req(TfwHttpReq *req)
...
+	if (!(resp->cache_ctl.flags & TFW_HTTP_CC_IS_PRESENT)
+	    && (resp->cache_ctl.flags & TFW_HTTP_CC_PRAGMA_NO_CACHE))
+		return false;
+	if (resp->cache_ctl.flags & TFW_HTTP_CC_PRIVATE)
+		return false;
+	if ((req->cache_ctl.flags & TFW_HTTP_CC_HDR_AUTHORIZATION)
+	    && !(req->cache_ctl.flags & authcan))
+		return false;
+	if (!(resp->cache_ctl.flags & cacheit) && !tfw_cache_status_bydef(resp))
+		return false;
```
krizhanovsky: There is no sense to check the `req->cache_ctl.flags` and `resp->cache_ctl.flags` so many times against different flags: rather you can just check each of them once for all (or many at least) of the flags - you can group the flags like you did it for `cacheit` and `authcan`.

tempesta_fw/cache.c
```diff
@@ -248,17 +261,120 @@ tfw_cache_employ_req(TfwHttpReq *req)
...
+tfw_cache_calc_lifetime(TfwHttpResp *resp)
+{
+	unsigned int lifetime;
+
+	if (resp->cache_ctl.flags & TFW_HTTP_CC_S_MAXAGE)
+		lifetime = resp->cache_ctl.s_maxage;
+	else if (resp->cache_ctl.flags & TFW_HTTP_CC_MAX_AGE)
+		lifetime = resp->cache_ctl.max_age;
+	else if (resp->cache_ctl.flags & TFW_HTTP_CC_HDR_EXPIRES)
+		lifetime = resp->cache_ctl.expires;
```
krizhanovsky: `Expires` set exact point in time when the entry must be invalidated while `maxage`s specify delta seconds, so `lifetime` must be calculated differently for the cases.

tempesta_fw/cache.c
```diff
@@ -248,17 +261,120 @@ tfw_cache_employ_req(TfwHttpReq *req)
...
+static time_t
+tfw_cache_entry_age(TfwCacheEntry *ce)
+{
+	time_t apparent_age = max_t(time_t, 0, ce->resp_time - ce->date);
+	time_t corrected_age = ce->age + ce->resp_time - ce->req_time;
+	time_t initial_age = max(apparent_age, corrected_age);
+	return (initial_age + tfw_current_timestamp() - ce->resp_time);
+}
+
+static bool
```
krizhanovsky: This is an algorithm which is good to verify by RFC, so please point out appropriate RFC(s) chapters which defines the algorithm in header comment for the function.

tempesta_fw/cache.c
```diff
@@ -616,6 +669,9 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwHttpReq *req,
 	}
 	BUG_ON(tot_len != 0);
 
+	if (resp->cache_ctl.flags
+	    & (TFW_HTTP_CC_MUST_REVAL | TFW_HTTP_CC_MUST_REVAL))
```
krizhanovsky: OR on the constant TFW_HTTP_CC_MUST_REVAL with itself?

tempesta_fw/cache.c
```diff
@@ -1020,7 +1076,10 @@ cache_req_process_node(TfwHttpReq *req, unsigned long key,
 
 	resp = tfw_cache_build_resp(ce);
 out:
-	action(req, resp);
+	if (!resp && (req->cache_ctl.flags & TFW_HTTP_CC_OIFCACHED))
+		tfw_http_send_502((TfwHttpMsg *)req); /* XXX: Change to 504. */
```
krizhanovsky: Why not to implement 504 response now?

tempesta_fw/cache.c
keshonok: Ah, should be `TFW_HTTP_CC_PROXY_REVAL`. Thanks!

tempesta_fw/cache.c
```diff
@@ -251,17 +254,73 @@ tfw_cache_employ_req(TfwHttpReq *req)
 	/* cache_fulfill - work as usual in cache mode. */
 	BUG_ON(cmd != TFW_D_CACHE_FULFILL);
 
+	/* Assume there's no Date: header in the request. */
+	req->cache_ctl.timestamp = tfw_current_timestamp();
```
krizhanovsky: What if THERE IS Date header in `req`?

tempesta_fw/cache.c
```diff
@@ -248,17 +261,120 @@ tfw_cache_employ_req(TfwHttpReq *req)
 	/* cache_fulfill - work as usual in cache mode. */
 	BUG_ON(cmd != TFW_D_CACHE_FULFILL);
 
+	if (req->cache_ctl.flags & TFW_HTTP_CC_NO_CACHE)
+		return false;
+
+	/* Assume there's no Date: header in the request. */
+	req->cache_ctl.timestamp = tfw_current_timestamp();
```
krizhanovsky: What if THERE IS Date header in req?

tempesta_fw/cache.c
```diff
@@ -248,17 +261,120 @@ tfw_cache_employ_req(TfwHttpReq *req)
...
+	if (resp->cache_ctl.flags & TFW_HTTP_CC_PRIVATE)
+		return false;
+	if ((req->cache_ctl.flags & TFW_HTTP_CC_HDR_AUTHORIZATION)
+	    && !(req->cache_ctl.flags & authcan))
+		return false;
+	if (!(resp->cache_ctl.flags & cacheit) && !tfw_cache_status_bydef(resp))
+		return false;
+
+	if (!resp->cache_ctl.timestamp)
+		resp->cache_ctl.timestamp = tfw_current_timestamp();
```
krizhanovsky: Response also can have Date header and it seems the response Date is more important for us than request's one since it's used for freshness calculation. So we shouldn't set the field unconditionally to current timestamp. Or at least mark it as TODO/FIXME.

tempesta_fw/cache.c
```diff
@@ -248,17 +261,120 @@ tfw_cache_employ_req(TfwHttpReq *req)
...
+		lifetime = resp->cache_ctl.max_age;
+	else if (resp->cache_ctl.flags & TFW_HTTP_CC_HDR_EXPIRES)
+		lifetime = resp->cache_ctl.expires;
+	else
+		lifetime = 0;	/* TODO: Heuristic lifetime. */
+
+	return lifetime;
+}
+
+static time_t
```
krizhanovsky: Please point out in the function header comment the RFC chapter which is implemented by the function.

tempesta_fw/http.h
```diff
@@ -95,26 +95,39 @@ enum {
 	TFW_HTTP_VER_20,
 };
 
+/* CC directives common to requests and responses. */
 #define TFW_HTTP_CC_NO_CACHE		0x00000001
 #define TFW_HTTP_CC_NO_STORE		0x00000002
-#define TFW_HTTP_CC_NO_TRANS		0x00000004
-#define TFW_HTTP_CC_NO_OIC		0x00000008
+#define TFW_HTTP_CC_NO_TRANSFORM	0x00000004
```
krizhanovsky: There is no sense to rename short and clear enough constants to longer names.

tempesta_fw/http_parser.c
```diff
@@ -1351,6 +1353,7 @@ __req_parse_cache_control(TfwHttpReq *req, unsigned char *data, size_t len)
 	__FSM_STATE(Req_I_CC_m) {
 		TRY_STR("max-age=", Req_I_CC_MaxAgeV);
 		TRY_STR("min-fresh=", Req_I_CC_MinFreshV);
+		TRY_STR("max_stale", Req_I_CC_MaxStale);
```
krizhanovsky: It seems the field is non-standard-defined, at least comment when we can see this is required.

keshonok: It is defined in the RFC: https://tools.ietf.org/html/rfc7234#section-5.2.1.2

krizhanovsky: RFC 7234 5.2.1.2 doesn't say anything about "max**_**stale" (`_` instead of `-`).

tempesta_fw/cache.c
keshonok: If a response has the `no-cache` CC directive, then according to the RFC 7234 5.2.2.2 that means that the response may be stored in cache. However, it must not be returned to a client without a revalidation. That is especially beneficial for voluminous responses, as a revalidation sequence it much smaller and therefore faster. Are we not going to implement this feature and just bypass the cache?

PR #566, opened 2016-06-19
Get rid of skb_unclone in ss_tcp_process_data

milabs:


tempesta_fw/ss_skb.c
```diff
@@ -976,3 +976,105 @@ ss_skb_split(struct sk_buff *skb, int len)
...
+	/*
+	 * Note, that skb_gro_receive() drops reference to the
+	 * SKB's header via the __skb_header_release(). So, to
+	 * not break the things we must take reference back.
+	 */
+	ss_skb_queue_tail(skb_list, skb);
+	skb_walk_frags(skb, f_skb) {
+		if (f_skb->nohdr) {
+			f_skb->nohdr = 0;
+			atomic_sub(1 << SKB_DATAREF_SHIFT, &skb_shinfo(f_skb)->dataref);
```
krizhanovsky: Too long line

tempesta_fw/ss_skb.c
```diff
@@ -976,3 +976,105 @@ ss_skb_split(struct sk_buff *skb, int len)
...
+
+	if (ss_skb_queue_coalesce_tail(skb_list, skb))
+		goto cleanup;
+
+	skb_walk_frags(skb, f_skb) {
+		if (ss_skb_queue_coalesce_tail(skb_list, f_skb))
+			goto cleanup;
+	}
+
+	consume_skb(skb);
```
krizhanovsky: If `skb_list` is empty, then `__coalesce_frag()` allocates new `skb` and here you free it. It'd be better to optimize the case and consume skb only when we don't need to allocate a new one.

tempesta_fw/ss_skb.c
```diff
@@ -976,3 +976,105 @@ ss_skb_split(struct sk_buff *skb, int len)
...
+		head_frag.size = headlen;
+		head_frag.page.p = virt_to_head_page(skb->head);
+		head_frag.page_offset = skb->data -
+			(unsigned char *)page_address(head_frag.page.p);
+		if (__coalesce_frag(skb_list, &head_frag))
+			return -ENOMEM;
+	}
+
+	for (i = 0; i < skb_shinfo(skb)->nr_frags; i++) {
+		if (__coalesce_frag(skb_list, &skb_shinfo(skb)->frags[i]))
```
krizhanovsky: I think this makes things unnecessary complex. Technically `ss_skb_unroll()` does only following things:

1. if `skb` has head, then shift frags to the left for one position and map `skb->head` page to `skb->frags[0]`;

2. if `skb` has flag_list, then move the skbs to `skb_list`.

So in some cases you can just do plain `memmove()` on `skb->frags` instead of calling `__coalesce_frag()` in the loop. Moreover, if `skb->frags` is full, then you still can do `memmove()` for `nr_frags - 1` items and try to place the last frag in first skb fragment from `frag_list` or, at the end, allocate a new skb.

I think the code can be shorter and more efficient.

krizhanovsky:
Ok, just write TODO comment near from `consume_skb()` to optimize skb reallocation and merge the fix to master.

PR #568, opened 2016-06-22
wq: add sync flag to __tfw_wq_push

milabs:


tempesta_fw/sock.c
```diff
@@ -72,7 +72,7 @@ ss_ipi(struct irq_work *work)
 }
 
 static int
-ss_wq_push(SsWork *sw)
+ss_wq_push(SsWork *sw, bool sync)
```
krizhanovsky: Maybe it's C++'ism, but `bool` isn't `int`: please use `0` for int arguments or `true`/`false` for boolean, but don't intermix them.

tempesta_fw/work_queue.c
```diff
@@ -139,6 +139,10 @@ __tfw_wq_push(TfwRBQueue *q, void *ptr)
 			if (head >= atomic64_read(&q->last_tail) + QSZ) {
 				/* The queue is full, don't wait consumers. */
 				atomic64_set(&pos->head, LLONG_MAX);
+				if (sync) {
+					cpu_relax();
+					goto retry;
+				}
```
krizhanovsky: No, actually the branch is for function returning. We don't need to do second try. If `sync == true`, then just call `__update_guards()`. I think we don't need to call `cpu_relax()` since there are relatively many code to execute.

So the adjustment could be just:

        if (!sync && head >= atomic64_read(&q->last_tail) + QSZ) {

krizhanovsky:
good to merge

PR #570, opened 2016-06-23
Store response's flags and HTTP version in cache entry.

keshonok:


tempesta_fw/cache.c
```diff
@@ -80,6 +80,8 @@ typedef struct {
 	long		status;
 	long		hdrs;
 	long		body;
+	unsigned char	version;
+	unsigned int	hmflags;
```
krizhanovsky: Please comment the fields

krizhanovsky:
Good to merge with small cleanup

Issue #552, opened 2016-06-03
Use __hdr_append() in __hdr_sub().


keshonok:
`__hdr_sub()` uses the slow path when a substitute header doesn't fit in place of the original header. The slow path is a simple `__hdr_del()` followed by `__hdr_add()`. The header is deleted in the middle of the header block of a message, and then the new header is added at the end of the header block.

In extreme case we create a bunch of empty skbs in the process (see #551), and then create another skb for the substitute header.

The slow path is doing a lot of work, more that necessary.  We already have `__hdr_append()` that can be used for the substitution. Instead of complete deletion followed by a complete insertion, it's possible to overwrite part of the original header with data from the substitute header, and add a bit more room for the rest of the substitute header exactly where it is needed.

There's one little nuance in this. When the original header is a duplicate header, then only the first header is substituted, and all duplicates are deleted. Also, one could imagine an optimization that recognizes when a duplicate header immediately follows the original, and then uses the space of both headers for the substitution.

milabs:
> Also, one could imagine an optimization that recognizes when a duplicate header immediately follows the original, and then uses the space of both headers for the substitution.

The implementation I've tried looks over-complicated because of header's `EOL` is not a part of the `TfwStr`. This leads to impossibility of reusing contiguous DUP-string without complication the logic. Also, in current implementation every DUP-string has `0` as it's length. So, total duplicate length computation is required to check if there is enough room to place source header over.

PR #572, opened 2016-06-24
http-msg: replace __hdr_append with __hdr_expand

milabs:
Instead of complete deletion followed by a complete insertion, it's
possible to overwrite part of the original header with data from the
substitute header, and add a bit more room for the rest of the
substitute header exactly where it is needed.

Related to #552.

### DIFF
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -415,33 +415,35 @@ __hdr_add(TfwHttpMsg *hm, const TfwStr *hdr, int hid)
 }
 
 /**
- * Insert ', @hdr' at the end of @orig_hdr
+ * Expand @orig_hdr by appending or replacing with the @hdr.
  * (CRLF is not accounted in TfwStr representation of HTTP headers).
  *
- * Append to first duplicate header, do not produce more duplicates.
+ * Expand the first duplicate header, do not produce more duplicates.
  */
 static int
-__hdr_append(TfwHttpMsg *hm, TfwStr *orig_hdr, const TfwStr *hdr)
+__hdr_expand(TfwHttpMsg *hm, TfwStr *orig_hdr, const TfwStr *hdr, bool append)
 {
 	int r;
 	TfwStr *h, it = {};
 
 	if (TFW_STR_DUP(orig_hdr))
 		orig_hdr = __TFW_STR_CH(orig_hdr, 0);
-	h = TFW_STR_LAST(orig_hdr);
+	BUG_ON(!append && (hdr->len < orig_hdr->len));
 
-	r = ss_skb_get_room(&hm->msg.skb_list, h->skb,
-			    (char *)h->ptr + h->len, hdr->len, &it);
+	h = TFW_STR_LAST(orig_hdr);
+	r = ss_skb_get_room(&hm->msg.skb_list,
+			    h->skb, (char *)h->ptr + h->len,
+			    append ? hdr->len : hdr->len - orig_hdr->len, &it);
 	if (r)
 		return r;
 
-	if (tfw_strcpy(&it, hdr))
-		return TFW_BLOCK;
-	if (tfw_strcat(hm->pool, orig_hdr, &it))
+	if (tfw_strcat(hm->pool, orig_hdr, &it)) {
 		TFW_WARN("Cannot concatenate hdr %.*s with %.*s\n",
 			 PR_TFW_STR(orig_hdr), PR_TFW_STR(hdr));
+		return TFW_BLOCK;
+	}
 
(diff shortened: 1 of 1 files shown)


krizhanovsky:
looks good for me. @keshonok , please review as well

tempesta_fw/http_msg.c
```diff
@@ -476,6 +476,50 @@ __hdr_del(TfwHttpMsg *hm, int hid)
 }
 
 /**
+ * Optimization that recognizes when duplicate headers immediately follows the
+ * original, and then uses the space of such a headers for the substitution.
+ * Also, it is required that all the headers must reside in one SKB.
```
krizhanovsky: Could you please note that the most expected case for the optimization is Set-Cookie.

krizhanovsky:
Looks good for me

tempesta_fw/http_msg.c
```diff
@@ -474,6 +476,52 @@ __hdr_del(TfwHttpMsg *hm, int hid)
...
+{
+	unsigned long len = 0;
+	const TfwStr *s1 = TFW_STR_CHUNK(orig_hdr, 0);
+	const TfwStr *s2 = TFW_STR_LAST(orig_hdr);
+	const TfwStr *c1 = TFW_STR_CHUNK(s1, 0);
+	const TfwStr *c2 = TFW_STR_LAST(s2);
+	TfwStr tail = {};
+
+	if (unlikely(c1->skb != c2->skb))
+		goto slow_path;
```
krizhanovsky: I think this is fine. Yes, probably we can adjust HTTP parser to track continuous duplicates. The adjustment isn't so trivial since some of N duplicates can be continuous and some of them aren't, so just flags aren't enough. HTTP parser is a hot spot, so generally it's unwished to complicate it to get more performance in other, not so hot, place.  Meantime, we can assume that HTTP headers, including probably duplicate Set-Cookies, are usually fit 64KB, one hardware GROed skb, so it's expectable that the condition is usually met.

tempesta_fw/http_msg.c
```diff
@@ -498,15 +546,18 @@ __hdr_sub(TfwHttpMsg *hm, char *name, size_t n_len, char *val, size_t v_len,
 		.flags = 3 << TFW_STR_CN_SHIFT
 	};
 
+	if (unlikely(TFW_STR_DUP(orig_hdr)))
+		return __hdr_sub_dup(hm, orig_hdr, &hdr, hid);
```
krizhanovsky: Again I agree with @keshonok: if the first header of duplicate string is long enough to place the new header, then we can rewrite the header at the first duplicate string regardless other duplicated of original header (whether they're at the same skb or not). The optimization should be small and straightforward. Probably, it has sense to handle the case in `__hdr_sub_dup()`.

tempesta_fw/str.c
```diff
@@ -119,6 +119,23 @@ tfw_str_add_duplicate(TfwPool *pool, TfwStr *str)
 	return dup_str;
 }
 
+/**
+ * Compute the whole string's length, including all the EOL's.
+ */
+unsigned long
+tfw_strlen(const TfwStr *str)
```
krizhanovsky: Yeah, @keshonok made reasonable note about name of the function. Basically, TfwStr has `len` member to get the string length immediately instead of computation via `strlen()` or similar functions, so I thing `strlen`-containg functions will be misleading for TfwStr. I'd name the function like `tfw_str_dup_len()`.

Meantime, I'm also always confused by `tfw_str_total_len()` - knowing TfwStr internals I expect that the function returns length of all the chunks of duplicates. So I'd rename the function to `tfw_str_eol_len()` since `total` doesn't say anything what `total` is: EOL, chunks or duplicates.

tempesta_fw/http_msg.c
```diff
@@ -498,32 +502,36 @@ __hdr_sub(TfwHttpMsg *hm, char *name, size_t n_len, char *val, size_t v_len,
...
-	 * EOL bytes are not a part of a header field string in Tempesta.
-	 * Therefore only @orig_hdr->len bytes at most can be copied over.
-	 * If the substitute string without the EOL fits into that space,
-	 * then the fast path can be used. Otherwise, go by the slow path.
-	 */
-	if (!TFW_STR_DUP(orig_hdr) && (hdr.len <= orig_hdr->len)) {
-		BUG_ON(!tfw_str_eolen(orig_hdr));
-
+	TFW_STR_FOR_EACH_DUP(dst, orig_hdr, end) {
+		if (likely(dst->len < hdr.len))
```
krizhanovsky: Why the condition is likely?

PR #574, opened 2016-06-29
SS: Proper TCP options

### LINKED ISSUE #475: SS: Proper TCP options
TCP options must be handled properly on sending path:

1. immediately send FIN with last portion of data (see https://github.com/natsys/tempesta/issues/376#issuecomment-190323770);

2. use MSG_MORE to optimize packet size at IP layer (https://github.com/natsys/tempesta/blob/master/tempesta_fw/sock.c#L189);

3. don't add PUSH to all segments (https://github.com/natsys/tempesta/blob/master/tempesta_fw/sock.c#L157).

milabs:
This fixes #475

### DIFF
--- a/tempesta_fw/cache.c
+++ b/tempesta_fw/cache.c
@@ -800,19 +800,16 @@ __cache_add_node(TDB *db, TfwHttpResp *resp, TfwHttpReq *req,
 
 }
 
-/**
- * @return false (i.e. don't reuse socket buffers)
- * if @resp is needed for caching.
- */
-bool
-tfw_cache_add(TfwHttpResp *resp, TfwHttpReq *req)
+static void
+tfw_cache_add(TfwHttpResp *resp, TfwHttpReq *req, tfw_http_cache_cb_t action)
 {
 	unsigned long key;
+	bool keep_skb = false;
 
 	if (!cache_cfg.cache || !tfw_cache_msg_cacheable(req))
-		return true;
+		goto out;
 	if (!tfw_cache_employ_resp(req, resp))
-		return true;
+		goto out;
 
 	key = tfw_http_req_key_calc(req);
 
@@ -829,10 +826,14 @@ tfw_cache_add(TfwHttpResp *resp, TfwHttpReq *req)
 	}
 
 	/*
-	 * Cache population is synchronous now, change the return value
-	 * depending on the TODO above.
+	 * Cache population is synchronous now. Don't forget to set
+	 * @keep_skb properly in case of asynchronous operation is being
+	 * performed.
 	 */
-	return true;
+
+out:
+	((TfwMsg *)resp)->ss_flags |= keep_skb ? SS_F_KEEP_SKB : 0;
+	action(req, resp);
 }
 
 static void
@@ -1228,7 +1229,7 @@ tfw_wq_tasklet(unsigned long data)
 
 	while (!tfw_wq_pop(&ct->wq, &cw)) {
 		if (cw.resp) {
-			cw.action(cw.req, cw.resp);
+			tfw_cache_add(cw.resp, cw.req, cw.action);
 		} else if (cw.req->method == TFW_HTTP_METH_PURGE) {
 			tfw_cache_purge_method(cw.req, cw.key);
 		} else {
--- a/tempesta_fw/connection.c
+++ b/tempesta_fw/connection.c
@@ -90,13 +90,11 @@ tfw_connection_release(TfwConnection *conn)
  * for the life of @conn instance. The socket itself may have been
  * closed, but not deleted. ss_send() makes sure that data is sent
  * only on an active socket.
- *
- * @unref_data is true if we won't use @msg any more.
  */
 int
-tfw_connection_send(TfwConnection *conn, TfwMsg *msg, bool unref_data)
+tfw_connection_send(TfwConnection *conn, TfwMsg *msg)
 {
-	return ss_send(conn->sk, &msg->skb_list, unref_data);
+	return ss_send(conn->sk, &msg->skb_list, msg->ss_flags);
 }
 
 int
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -238,7 +238,9 @@ tfw_http_send_resp(TfwHttpMsg *hmreq, TfwStr *msg, const TfwStr *date)
 	tfw_http_prep_date(date->ptr);
 	tfw_http_msg_write(&it, &resp, msg);
 
-	return tfw_cli_conn_send(hmreq->conn, (TfwMsg *)&resp, true);
+	__adjust_resp_flags((TfwHttpResp *)&resp, (TfwHttpReq *)hmreq);
+
+	return tfw_cli_conn_send(hmreq->conn, (TfwMsg *)&resp);
 }
 
 #define S_200_PART_01	S_200 S_CRLF S_F_DATE
@@ -450,37 +452,6 @@ tfw_http_conn_msg_free(TfwHttpMsg *hm)
 	tfw_http_msg_free(hm);
 }
 
-/*
- * Drop the client connection and destroy the request.
- *
- * A connection is dropped in two stages. The socket is closed first.
- * Then the connection linked with the socket is withdrawed from all
- * socket activity, and then deactivated.
- *
- * The socket may be closed either from Tempesta or in Sync Sockets
- * as the reaction to incoming FIN. Both may occur at the same time.
- * Socket lock permits one party only to proceed with the closing.
- * See the comment to ss_close().
- *
- * The connection may be dropped only by the party that closed the
- * socket. Also, it's essential that the connection is dropped only
- * when there's at least one extra reference to it (@conn->refcnt).
- * Then if the reference is dropped here it's never the last one.
- * That won't allow Sync Sockets to destroy the socket and the
- * connection completely until Tempesta is done with the connection.
- * The request holds that extra reference, so it must be freed only
- * after the connection is dropped.
- */
-static inline void
-tfw_http_conn_cli_dropfree(TfwHttpMsg *hmreq)
-{
-	BUG_ON(!(TFW_CONN_TYPE(hmreq->conn) & Conn_Clnt));
-
-	if (hmreq->flags & TFW_HTTP_CONN_CLOSE)
-		ss_close_sync(hmreq->conn->sk, true);
-	tfw_http_conn_msg_free(hmreq);
-}
-
 /**
  * TODO Initialize allocated Client structure by HTTP specific callbacks
  * and FSM.
@@ -512,7 +483,7 @@ tfw_http_conn_release(TfwConnection *conn)
 			&& (((TfwHttpMsg *)msg)->conn == conn));
 		list_del(&msg->msg_list);
 		tfw_http_send_404((TfwHttpMsg *)msg);
-		tfw_http_conn_cli_dropfree((TfwHttpMsg *)msg);
+		tfw_http_conn_msg_free((TfwHttpMsg *)msg);
 		TFW_INC_STAT_BH(clnt.msgs_otherr);
 	}
 	INIT_LIST_HEAD(&conn->msg_queue);
@@ -727,6 +698,8 @@ tfw_http_adjust_req(TfwHttpReq *req)
 	int r;
 	TfwHttpMsg *hm = (TfwHttpMsg *)req;
 
+	__adjust_req_flags(req);
+
 	r = tfw_http_add_x_forwarded_for(hm);
 	if (r)
 		return r;
@@ -746,6 +719,8 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
 {
 	int r, conn_flg = req->flags & __TFW_HTTP_CONN_MASK;
 	TfwHttpMsg *hm = (TfwHttpMsg *)resp;
+
+	__adjust_resp_flags(resp, req);
 
 	r = tfw_http_sticky_resp_process(hm, (TfwHttpMsg *)req);
 	if (r < 0)
@@ -791,12 +766,12 @@ tfw_http_req_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
 		 */
 		if (tfw_http_adjust_resp(resp, req))
 			goto resp_err;
-		if (tfw_cli_conn_send(req->conn, (TfwMsg *)resp, true))
+		if (tfw_cli_conn_send(req->conn, (TfwMsg *)resp))
 			goto resp_err;
 		TFW_INC_STAT_BH(clnt.msgs_fromcache);
 resp_out:
 		tfw_http_conn_msg_free((TfwHttpMsg *)resp);
-		tfw_http_conn_cli_dropfree((TfwHttpMsg *)req);
+		tfw_http_conn_msg_free((TfwHttpMsg *)req);
 		return;
 resp_err:
 		tfw_http_send_500((TfwHttpMsg *)req);
@@ -846,7 +821,7 @@ resp_err:
 	spin_unlock(&srv_conn->msg_qlock);
 
 	/* Send request to the server. */
-	if (tfw_connection_send(srv_conn, (TfwMsg *)req, false)) {
+	if (tfw_connection_send(srv_conn, (TfwMsg *)req)) {
 		spin_lock(&srv_conn->msg_qlock);
 		list_del(&req->msg.msg_list);
 		spin_unlock(&srv_conn->msg_qlock);
@@ -857,12 +832,12 @@ resp_err:
 
 send_404:
 	tfw_http_send_404((TfwHttpMsg *)req);
-	tfw_http_conn_cli_dropfree((TfwHttpMsg *)req);
+	tfw_http_conn_msg_free((TfwHttpMsg *)req);
 	TFW_INC_STAT_BH(clnt.msgs_otherr);
 	return;
 send_500:
 	tfw_http_send_500((TfwHttpMsg *)req);
-	tfw_http_conn_cli_dropfree((TfwHttpMsg *)req);
+	tfw_http_conn_msg_free((TfwHttpMsg *)req);
 	TFW_INC_STAT_BH(clnt.msgs_otherr);
 conn_put:
 	if (tfw_connection_put(srv_conn))
@@ -1052,7 +1027,7 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
 				      tfw_http_req_cache_cb))
 		{
 			tfw_http_send_500(hmreq);
-			tfw_http_conn_cli_dropfree(hmreq);
+			tfw_http_conn_msg_free(hmreq);
 			TFW_INC_STAT_BH(clnt.msgs_otherr);
 			return TFW_PASS;
 		}
@@ -1096,9 +1071,6 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
 static void
 tfw_http_resp_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
 {
-	/* Cache original response before any mangling. */
-	bool unref_data = tfw_cache_add(resp, req);
-
 	/*
 	 * Typically we're at a node far from the node where @resp was
 	 * received, so we do an inter-node transfer. However, this is
@@ -1109,14 +1081,14 @@ tfw_http_resp_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
 	if (tfw_http_adjust_resp(resp, req))
 		goto err;
 
-	if (tfw_cli_conn_send(req->conn, (TfwMsg *)resp, unref_data))
+	if (tfw_cli_conn_send(req->conn, (TfwMsg *)resp))
 		goto err;
 
 	TFW_INC_STAT_BH(serv.msgs_forwarded);
 out:
 	/* Now we don't need the request and the response anymore. */
 	tfw_http_conn_msg_free((TfwHttpMsg *)resp);
-	tfw_http_conn_cli_dropfree((TfwHttpMsg *)req);
+	tfw_http_conn_msg_free((TfwHttpMsg *)req);
 	return;
 err:
 	tfw_http_send_500((TfwHttpMsg *)req);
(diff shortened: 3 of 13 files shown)


tempesta_fw/connection.c
```diff
@@ -90,13 +91,23 @@ tfw_connection_release(TfwConnection *conn)
...
+	const TfwHttpMsg *hm = (TfwHttpMsg *)msg;
+
+	if (TFW_CONN_TYPE(conn) & Conn_Clnt) {
+		if (tfw_http_msg_conn(hm) == TFW_HTTP_CONN_CLOSE)
+			flags |= SS_SEND_F_CONN_CLOSE;
+		if (!(hm->flags & TFW_HTTP_DO_NOT_PASS))
+			flags |= SS_SEND_F_PASS_SKB;
+	} else {
+		/* Add server connection flags */
+	}
```
keshonok: Perhaps, it's better just to leave a comment here and remove an empty branch.
`Add server connection flags in an "else{}" branch if necessary.`

tempesta_fw/http.c
```diff
@@ -1009,7 +1011,7 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
 		 * So, save the needed request flag for later use as it
 		 * may not be accessible later through @req->flags.
 		 */
-		req_conn_close = (hmreq->flags & TFW_HTTP_CONN_CLOSE);
+		req_conn_close = (tfw_http_msg_conn(hmreq) == TFW_HTTP_CONN_CLOSE);
```
keshonok: The line is too long.

tempesta_fw/http.h
```diff
@@ -232,6 +232,9 @@ typedef struct {
 #define TFW_HTTP_VOID_BODY		0x010000	/* Resp to HEAD req */
 #define TFW_HTTP_HAS_HDR_DATE		0x020000	/* Has Date: header */
 
+/* Used to indicate that @SS_SEND_F_PASS_SKB must not be used */
```
keshonok: An empty line before this one is confusing, as empty lines are used to separate blocks of definitions. Here is the block of definitions for responses. I would remove the empty line for that reason.

tempesta_fw/sock.c
```diff
@@ -194,6 +194,9 @@ ss_do_send(struct sock *sk, SsSkbList *skb_list)
 	       smp_processor_id(), __func__,
 	       sk, tcp_send_head(sk), sk->sk_state);
 
+	if ((flags & SS_SEND_F_CONN_CLOSE) && tcp_close_state(sk))
+		ss_tcp_shutdown(sk);
```
keshonok: Perhaps I am missing something here, but I believe that the goal is not just set the FIN flag on the last skb of the outgoing message in the connection that is being closed, but also to **close** the connection right now, not on the next timer tick (SoftIRQ run). Yes, the connection is moved to a closing state. However, Tempesta's close procedure is somewhat different from the "lengthy" kernel's close procedure. Perhaps a discussion is needed here.

tempesta_fw/sock.c
```diff
@@ -113,14 +114,25 @@ ss_skb_entail(struct sock *sk, struct sk_buff *skb)
 	tcb->seq     = tcb->end_seq = tp->write_seq;
 	tcb->tcp_flags = TCPHDR_ACK;
 	tcb->sacked  = 0;
```
keshonok: Do we need tabs here before the `=` character? An attempt to beautify this code has failed anyway. :) Just add `=` after a space, and that's it.

tempesta_fw/http.c
```diff
@@ -759,6 +759,8 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
 	if (r < 0)
 		return r;
 
+	tfw_http_msg_set_conn(hm, conn_flg);
```
keshonok: Nice touch, considering that flags **are** checked now in a response on the way out.

tempesta_fw/cache.c
```diff
@@ -1212,7 +1209,7 @@ tfw_wq_tasklet(unsigned long data)
 
 	while (!tfw_wq_pop(&ct->wq, &cw)) {
 		if (cw.resp) {
-			cw.action(cw.req, cw.resp);
+			tfw_cache_add(cw.resp, cw.req, cw.action);
```
keshonok: Thanks for implementing this. `tfw_cache_add()` is called from the right place now.

tempesta_fw/work_queue.h
```diff
@@ -30,7 +30,7 @@ typedef struct {
 } __WqItem;
 
 #define WQ_ITEM_SZ		sizeof(__WqItem)
-#define TFW_WQ_CHECKSZ(t)	BUILD_BUG_ON(sizeof(t) != WQ_ITEM_SZ)
+#define TFW_WQ_CHECKSZ(t)	BUILD_BUG_ON(sizeof(t) > WQ_ITEM_SZ)
```
krizhanovsky: It seems the change is wrong: different item sizes can lead to memory corruptions.

tempesta_fw/http_parser.c
```diff
@@ -905,14 +905,14 @@ __parse_connection(TfwHttpMsg *hm, unsigned char *data, size_t len)
 
 	__FSM_STATE(I_Conn) {
 		TRY_STR_LAMBDA("close", {
-			if (msg->flags & TFW_HTTP_CONN_KA)
+			if (tfw_http_msg_conn(msg))
 				return CSTR_NEQ;
-			msg->flags |= TFW_HTTP_CONN_CLOSE;
+			tfw_http_msg_set_conn(msg, TFW_HTTP_CONN_CLOSE);
```
krizhanovsky: I don't like the patch. `tfw_http_msg_conn(msg)` doesn't equal to `msg->flags & TFW_HTTP_CONN_KA` and `tfw_http_msg_set_conn(msg, TFW_HTTP_CONN_CLOSE)` does more operations than `msg->flags |= TFW_HTTP_CONN_CLOSE`. Probably in this particular case the statements are equal  and performance penalty is invisible, however the change just makes the code bit nicer (meantime `msg->flags & TFW_HTTP_CONN_KA` says more for me than `fw_http_msg_conn(msg)`) and this `bit nicer` doesn't outweight the badnesses.

tempesta_fw/sock.c
```diff
@@ -549,13 +546,16 @@ ss_tcp_process_data(struct sock *sk)
...
-
 		/* Shared SKBs shouldn't be seen here. */
 		if (skb_shared(skb))
 			BUG();
 
+		__skb_unlink(skb, &sk->sk_receive_queue);
+		skb_orphan(skb);
+
+		/* @skb should be rerouted on forwarding. */
+		skb_dst_drop(skb);
```
krizhanovsky: This is overkill as well. For example, it we consider web-server workload when no ingress skb is forwarded at all. Other example is DDoS attack when most of ingress packets are dropped as well.

I think introducing flag like `TFW_FORWARDED` to skb or passing it through function calls will do good job... However, if skb is generated then `skb_dst_drop()` is just one conditional check and its overhead is the same as for the flag checking.

tempesta_fw/connection.c
```diff
@@ -90,13 +91,22 @@ tfw_connection_release(TfwConnection *conn)
...
+tfw_connection_send(TfwConnection *conn, TfwMsg *msg)
 {
-	return ss_send(conn->sk, &msg->skb_list, unref_data);
+	int flags = 0;
+	const TfwHttpMsg *hm = (TfwHttpMsg *)msg;
+
+	if (TFW_CONN_TYPE(conn) & Conn_Clnt) {
+		if (tfw_http_msg_conn(hm) == TFW_HTTP_CONN_CLOSE)
+			flags |= SS_SEND_F_CONN_CLOSE;
+		if (!(hm->flags & TFW_HTTP_DO_NOT_PASS))
```
krizhanovsky: Nobody sets the flag, so every place which previously had `tfw_cli_conn_send(.., true)` is broken now.

tempesta_fw/connection.c
```diff
@@ -90,13 +91,22 @@ tfw_connection_release(TfwConnection *conn)
...
-tfw_connection_send(TfwConnection *conn, TfwMsg *msg, bool unref_data)
+tfw_connection_send(TfwConnection *conn, TfwMsg *msg)
 {
-	return ss_send(conn->sk, &msg->skb_list, unref_data);
+	int flags = 0;
+	const TfwHttpMsg *hm = (TfwHttpMsg *)msg;
+
+	if (TFW_CONN_TYPE(conn) & Conn_Clnt) {
+		if (tfw_http_msg_conn(hm) == TFW_HTTP_CONN_CLOSE)
+			flags |= SS_SEND_F_CONN_CLOSE;
```
krizhanovsky: Why for client connections only? If we send a message with `Connection: close` to server connection, then we can send it with FIN such that sever will ACKnowledge the FIN in its first data segment and connection will be closed quicker.

tempesta_fw/sock.c
```diff
@@ -245,7 +247,7 @@ ss_send(struct sock *sk, SsSkbList *skb_list, bool pass_skb)
 	 * Schedule the socket for TX softirq processing.
 	 * Only part of @skb_list could be passed to send queue.
 	 */
-	if (ss_wq_push(&sw, false)) {
+	if (ss_wq_push(&sw, (flags & SS_SEND_F_SYNC))) {
```
krizhanovsky: Hm, why did you introduce the flag? Nobody uses it for now...

tempesta_fw/t/unit/test_http_sticky.c
```diff
@@ -149,8 +148,7 @@ tfw_connection_send(TfwConnection *conn, TfwMsg *msg,
 
 	mock.tfw_connection_send_was_called += 1;
 
-	skb = ss_skb_peek(&msg->skb_list);
-	while (skb) {
+	while ((skb = ss_skb_peek(&msg->skb_list))) {
```
krizhanovsky: Endless loop? `ss_skb_peek()` always return first skb in the list, so following `ss_skb_next()` seems should be replaced by `ss_skb_unlink()`.

tempesta_fw/sock.c
```diff
@@ -114,14 +114,25 @@ ss_skb_entail(struct sock *sk, struct sk_buff *skb)
 	tcb->seq     = tcb->end_seq = tp->write_seq;
 	tcb->tcp_flags = TCPHDR_ACK;
 	tcb->sacked  = 0;
-	skb_header_release(skb);
+	__skb_header_release(skb);
```
krizhanovsky: Comment for `__skb_header_release()` says "assuming skb is private to caller", so what if we forward a skb received from local web-server process (the process also takes reference to the skb)? Or do you handle the case by uncloning?..

krizhanovsky:
"use MSG_MORE to optimize packet size at IP layer" in the original task #475 was mostly about https://github.com/tempesta-tech/linux-4.1-tfw/blob/master/net/ipv4/tcp.c#L688: don't send partial TCP segements, so it's good to implement the option.

tempesta_fw/sock.c
krizhanovsky: Actually `ss_tcp_shutdown()` is already partially [implements `ss_do_close()`](https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/sock.c#L339).

It seems we should at least add `tcp_close_state()` call to `ss_tcp_shutdown()` since we have to move the socket to FIN-WAIT-1 state when we send FIN, but it seems `tcp_push()` doesn't analyze the flag for transmitted skbs. I think TCP stack won't close the connection properly when it receives ACK for the FIN. Probably we also have to do `sk->sk_shutdown = SHUTDOWN_MASK`.

Since we don't want and can't to send more data on the socket, we can do `sk_mem_reclaim(sk)` as well.

krizhanovsky:
Please, review possible bug in `ss_do_close()`: the function returns ([here](https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/sock.c#L375) and [here](https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/sock.c#L384)) after socket reference counter [incrementing](https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/sock.c#L363). So it seems socket descriptor leakage is possible.

IIRC we enter `ss_do_close()` with zero refcounter of the socket, so we don't need to do `sock_put()` twice (in `inet_csk_destroy_sock()` and at below of the function) as standard `tcp_close()` does it. However, if we return from the function we never reach zero refcounter on following `ss_do_close()` calls.

tempesta_fw/connection.c
milabs: Actually, the only place where `tfw_cli_conn_send` called with indirect argument is:
https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/http.c#L1112

But, in that case `unref_data` is the result of `bool unref_data = tfw_cache_add(resp, req);`. The latter is always `true` as `tfw_cache_add` do not returns `false` at all.

krizhanovsky: Yeah, I should have been write "every place which previously had tfwi_connection_send(.., false) is broken now". `tfw_cache_add()` is assumed to be changed such that it will return different values. Also there is also https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/http.c#L849 which requires `false` as the last argument.

tempesta_fw/sock.c
milabs: `ss_skb_unroll` makes `skb` private for us. If it were referenced by someone except us, then private copy of SKB chain will be allocated. So, in my opinion, it's no reason to use `skb_header_release` version here anymore.

tempesta_fw/connection.c
milabs: Nope, as `tfw_connection_send(srv_conn, (TfwMsg *)req, false)` is used with the server's connection. In the latter case  it falls to default behavior with `flags = 0`.

tempesta_fw/connection.c
milabs: Does that behavior supposed to be ordinal for the Tempesta to allow someone, except the Tempesta, to close server's connection?

milabs:
@krizhanovsky Well, `ss_do_close` always followed by the `sock_put`, so I do not see the problem with descriptor leakage you've mentioned. Am I missed something?

tempesta_fw/connection.c
```diff
@@ -90,13 +91,22 @@ tfw_connection_release(TfwConnection *conn)
...
-	return ss_send(conn->sk, &msg->skb_list, unref_data);
+	int flags = 0;
+	const TfwHttpMsg *hm = (TfwHttpMsg *)msg;
+
+	if (TFW_CONN_TYPE(conn) & Conn_Clnt) {
+		if (hm->flags & TFW_HTTP_CONN_CLOSE)
+			flags |= SS_SEND_F_CONN_CLOSE;
+		if (!(hm->flags & TFW_HTTP_DO_NOT_PASS))
+			flags |= SS_SEND_F_PASS_SKB;
+	}
```
krizhanovsky: Conceptual architecture is broken. Connection layer is generic - it lays under HTTP, HTTPS and probably other protocols, so connection.[ch] shouldn't include HTTP stuff, while http.[ch] include connection.h. So firstly the change introduces circular dependency. Secondly it breaks program modules dependency.

I'd say this is HTTP responsibility to set appropriate connection flags. Moreover, `tfw_connection_send()` is called for server transmissions as well as for client, so there is no sense to introduce the IF construction when the caller knows exactly who it sends data to.

tempesta_fw/http.c
```diff
@@ -759,6 +759,9 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
 	if (r < 0)
 		return r;
 
+	/* Adjust @resp flags to use them in @tfw_connection_send */
+	hm->flags = (hm->flags & ~__TFW_HTTP_CONN_MASK) | conn_flg;
```
krizhanovsky: Yeah, it seems the right place to set the flags from comment https://github.com/tempesta-tech/tempesta/pull/574/files#r69903646 .

tempesta_fw/connection.c
```diff
@@ -90,13 +91,22 @@ tfw_connection_release(TfwConnection *conn)
...
+	/* Add server connection flags in an "else{}" branch if necessary. */
```
krizhanovsky: I think this is obvious, isn't it?

tempesta_fw/connection.c
krizhanovsky: Ok, agree.

tempesta_fw/connection.c
krizhanovsky: Yeah, agree. At least we don't suppose the behavior for now. However, I still think this is not the right place to set the flags, see https://github.com/tempesta-tech/tempesta/pull/574/files#r69903646

tempesta_fw/sock.c
```diff
@@ -194,6 +195,9 @@ ss_do_send(struct sock *sk, SsSkbList *skb_list)
 	       smp_processor_id(), __func__,
 	       sk, tcp_send_head(sk), sk->sk_state);
 
+	if (flags & SS_SEND_F_CONN_CLOSE)
+		return;
```
krizhanovsky: What's going on here? It seems if we're sending data with connection close flag, then the function just frees the skb list (ever w/o freeing skbs!) and just exits.

Also please remove comment line with MSG_MORE at 180th line. I agree that we don't need MSG_MORE flag since we do just one TCP call for whole message.

tempesta_fw/sock.c
```diff
@@ -1065,6 +1068,41 @@ ss_getpeername(struct sock *sk, struct sockaddr *uaddr, int *uaddr_len)
...
+	if (!ss_sock_live(sk)) {
+		SS_DBG("[%d]: %s: Socket inactive: sk %p\n",
+		       smp_processor_id(), __func__, sk);
+		bh_unlock_sock(sk);
+		return;
+	}
+	ss_do_close(sk);
+	bh_unlock_sock(sk);
+	SS_CALL(connection_drop, sk);
+	sock_put(sk); /* paired with ss_do_close() */
```
krizhanovsky: These 4 lines are just copy & paster among this function and `__ss_do_send()`. Also why don't you check `ss_sock_live()` in `__ss_do_send()`? I think it's better to make `__ss_do_close()` as just these 4 lines with the `ss_sock_live()` check (if it's actually needed) and call it directly from `ss_tx_action()` under the socket lock.

krizhanovsky:
@milabs , "Well, ss_do_close always followed by the sock_put, so I do not see the problem with descriptor leakage you've mentioned. Am I missed something?"

Oh, I see. Thanks.

tempesta_fw/sock.c
```diff
@@ -405,6 +402,15 @@ ss_do_close(struct sock *sk)
 	}
 }
 
+#define __ss_sk_close(sk, lock, unlock, drop)				\
```
krizhanovsky: It's hard to read code like `__ss_sk_close(sk, 1, 1, 1)` - what the numbers are? It's hard to remember. Let's introduce defines like `__ss_sk_close_locked()` or `__ss_sk_close_drop_locked()`.

tempesta_fw/http.h
```diff
@@ -376,4 +376,17 @@ int tfw_http_send_504(TfwHttpMsg *hm);
...
+	((TfwMsg *)req)->ss_flags |= SS_F_KEEP_SKB;
+}
+
+static inline void
+__adjust_resp_flags(TfwHttpResp *resp, const TfwHttpReq *req)
+{
+	if (req->flags & TFW_HTTP_CONN_CLOSE)
+		((TfwMsg *)resp)->ss_flags |= SS_F_CONN_CLOSE;
+}
+
```
krizhanovsky: The functions are used in http.c only, so it seems their place is in http.c

tempesta_fw/sock.c
```diff
@@ -196,6 +188,12 @@ ss_do_send(struct sock *sk, SsSkbList *skb_list)
 	       smp_processor_id(), __func__,
 	       sk, tcp_send_head(sk), sk->sk_state);
 
+	/* If connection close flag is specified, then @ss_do_close is used to
+	 * set FIN on final SKB and push all pending frames to the stack. */
```
krizhanovsky: Please use 

        /*
         * blah blah blah
         */

- style for multi line comments

tempesta_fw/sock.c
```diff
@@ -1087,6 +1084,41 @@ ss_getpeername(struct sock *sk, struct sockaddr *uaddr, int *uaddr_len)
 }
 EXPORT_SYMBOL(ss_getpeername);
 
+static inline void
+__ss_do_send(SsWork *sw)
+{
+	struct sock *sk = sw->sk;
+
+	bh_lock_sock(sk);
+	ss_do_send(sk, &sw->skb_list, sw->flags);
```
krizhanovsky: Usually `xxx` function calls `__xxx` or `do_xxx`, so it has sense to rename the functions.

tempesta_fw/sock.c
```diff
@@ -419,30 +416,30 @@ ss_droplink(struct sock *sk)
...
 	    || smp_processor_id() != sk->sk_incoming_cpu)
 	{
 		SsWork sw = {
 			.sk	= sk,
+			.flags  = flags,
 			.action	= SS_CLOSE,
 		};
 
-		return ss_wq_push(&sw, sync);
+		return ss_wq_push(&sw, false);
```
krizhanovsky: If we pass here e.g. by `!in_softirq()`, then `flags & SS_F_SYNC` can be true. I.e. we need to put the work to the queue, but do this synchronously.

tempesta_fw/sock.c
```diff
@@ -1087,6 +1084,41 @@ ss_getpeername(struct sock *sk, struct sockaddr *uaddr, int *uaddr_len)
...
+		SS_DBG("[%d]: %s: Socket inactive: sk %p\n",
+		       smp_processor_id(), __func__, sk);
+		bh_unlock_sock(sk);
+		return;
+	}
+	ss_do_close(sk);
+	bh_unlock_sock(sk);
+	SS_CALL(connection_drop, sk);
+	sock_put(sk); /* paired with ss_do_close() */
+}
```
krizhanovsky: The functions are called only once from `ss_tx_action()`. Do we really need them?

tempesta_fw/http.h
milabs: `__adjust_resp_flags used` twice, so it's good to have it as inline. `__adjust_req_flags` is used only once, but it has the same purpose as previous. I've change their names to `__init_resp_ss_flags` and `__init_req_ss_flags`.

tempesta_fw/sock.c
milabs: My miss. Thanks.

krizhanovsky:
good to merge

Issue #576, opened 2016-06-30
Soft lockup issues on 10G link with persistent client connections


keshonok:
The environment is as follows:
- latest Tempesta kernel on the current date;
- latest Tempesta build on the current date;
- ixgbe driver 4.4.6 built from sources and loaded before the test is run;
- 10G link between the Tempesta server and the server that runs the ab benchmark;
- Tempesta server has 2Gb RAM and 8 CPU cores;
- ixgbe driver creates RxTx queues CPU affinity by default.

This is the default CPU affinity:
```
for irq in `grep eth3 /proc/interrupts | cut -d: -f1`; do cat /proc/irq/$irq/smp_affinity; done
000001
000002
000004
000008
000010
000020
000040
000080
0000ff
```

On various test runs Tempesta kernel starts spewing soft lockup messages. On the test runs it always occured on CPU1. That, however, doesn't necessarily mean that CPU1 is somehow at fault. It can be any other CPU.

The issue occurs when `ab` is run with `-k` option that causes `ab` to keep the connection open for as long as possible. It's not necessary to have a large number of concurrent connections. the issue happened, for instance, with 128 or 512 concurrent connections.

Below is a collection of various occurrences of this issue in form of kernel backtraces.

```
[  777.205064] NMI watchdog: BUG: soft lockup - CPU#1 stuck for 22s! [ksoftirqd/1:13]
[...]

PR #577, opened 2016-07-06
Schedule cache work to all available CPUs. (#576)

keshonok:


milabs:
:+1:

tempesta_fw/cache.c
```diff
@@ -487,7 +488,8 @@ node_db(void)
 static int
 tfw_cache_sched_cpu(TfwHttpReq *req)
 {
-	return c_nodes[req->node].cpu;
+	struct ca_node_t *node = &ca_nodes[req->node];
+	return node->cpu[node->cpu_idx++ % node->nr_cpus];
```
krizhanovsky: `node->cpu_idx` can be accessed from many CPU cores concurrently, so atomic increment must be used.

milabs: Does it really make sense in comparative with `lock` prefix overhead?

tempesta_fw/cache.c
```diff
@@ -151,10 +151,12 @@ static const int hbh_hdrs[] = {
...
-static struct {
-	int	cpu;
-	TDB	*db;
-} c_nodes[MAX_NUMNODES] __read_mostly;
+static struct ca_node_t {
+	int		cpu[NR_CPUS];
+	unsigned int	cpu_idx;
+	unsigned int	nr_cpus;
+	TDB		*db;
+} ca_nodes[MAX_NUMNODES];
```
krizhanovsky: Why it's renamed and why `__read_mostly` isn't needed?

tempesta_fw/cache.c
```diff
@@ -487,7 +488,8 @@ node_db(void)
 static int
```
krizhanovsky: Please remove TODO comment line above

tempesta_fw/cache.c
keshonok: `ca` is much more readable and understandable than `c` as an abbreviation for `cache`.
`__read_mostly` is not suitable anymore, in my opinion, as `cpu_idx` is continuously incremented now.

tempesta_fw/cache.c
keshonok: I was thinking along the same lines, that strict atomicity is not necessary here.
Do you think it should be a strict atomic increment?

tempesta_fw/cache.c
krizhanovsky: Ok, agree about `__read_mostly`, but disagree about more readable `ca`.

One more point: we use typedefs instead of `struct` type definitions. So there should be `typedef struct { .. } CaNode` instead of `struct ca_node_t`.

tempesta_fw/cache.c
krizhanovsky: If many cores update the same cache line, then there is still huge amount of memory transfers due to frequent invalidation of the cache line. If performance is critical, then the counter should be per_cpu. However, I suppose this isn't a hot spot.

krizhanovsky:
good to merge

Issue #578, opened 2016-07-08
Potential deadlock on net->ipv4.ip_local_ports.lock


keshonok:
This started to show up occasionally in Tempesta performance tests. Latest Tempesta source and kernel. Tempesta is configured to work from cache (in web server mode), so the exchange with Nginx back end server is mininal. Note that Nginx works on the same server. Nginx, however, still resets connections with Tempesta on a periodic basis as configured. From the trace it appears that this deadlock issue happens when Nginx drops the connection from Tempesta, and Tempesta attempts to reconnect.

```
[  150.941291] [tdb] Start Tempesta DB
[  151.040298] [tempesta] Initializing Tempesta FW kernel module...
[  151.080654] [tempesta] Registering new scheduler: hash
[  151.097436] [tempesta] Registering new scheduler: http
[  151.109219] [tempesta] Registering new scheduler: round-robin
[  151.263155] [tempesta] got state via sysctl: start
[  151.268024] [tempesta] starting all modules...
[  151.297167] [tdb] Opened table /opt/tempesta/db/filter.tdb: size=16777216 rec_size=20 base=ffff880078c00000
[  151.661350] [tdb] Opened table /opt/tempesta/db/cache.tdb: size=268435456 rec_size=0 base=ffff880068c00000
[  151.672084] [tempesta] Open listen socket on: 192.168.100.150
[  151.678455] [tempesta] modules are started
[  211.744055] 
[  211.745555] =================================
[  211.749925] [ INFO: inconsistent lock state ]
[  211.754299] 4.1.12+ #2 Tainted: G           O   
[  211.759666] ---------------------------------
[  211.764753] inconsistent {SOFTIRQ-ON-W} -> {IN-SOFTIRQ-R} usage.
[  211.771507] swapper/0/0 [HC0[0]:SC1[1]:HE0:SE0] takes:
[  211.777383]  (&(&net->ipv4.ip_local_ports.lock)->seqcount){+.+-..}, at: [<ffffffff815051a1>] __inet_hash_connect+0x61/0x3a0
[  211.789370] {SOFTIRQ-ON-W} state was registered at:
[  211.795028]   [<ffffffff810c4aa1>] __lock_acquire+0xa71/0x2030
[  211.801669]   [<ffffffff810c6af2>] lock_acquire+0xc2/0x290
[  211.807956]   [<ffffffff8155457d>] ipv4_local_port_range+0xcd/0x110
[...]

PR #579, opened 2016-07-08
Prevent deadlock on net->ipv4.ip_local_ports.lock

milabs:
Related to #578.

### DIFF
--- a/tempesta_fw/sock.c
+++ b/tempesta_fw/sock.c
@@ -1001,6 +1001,8 @@ EXPORT_SYMBOL(ss_release);
 int
 ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
 {
+	struct net *net = sock_net(sk);
+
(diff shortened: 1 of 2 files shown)


tempesta_fw/sock.c
```diff
@@ -1009,6 +1011,10 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
 	if (sk->sk_state != TCP_CLOSE)
 		return -EISCONN;
 
+	/* Prevent possible deadlock with inet_get_local_port_range() (#578) */
+	if (spin_is_locked(&net->ipv4.ip_local_ports.lock.lock))
+		return -EAGAIN;
```
krizhanovsky: Looks very hacky: locking retries are done on timeout basis requireing many code to execute for each lock try. Also timeout retries are adjusted just to do retries, while typically timers are for periodic works.

milabs:
This issue was fixed in linux kernel by 4ee3bd4a8c7463cdef0b82ebc33fc94a9170a7e0 commit.

PR #581, opened 2016-07-10
#576: fix infinite loop on work queue push()

krizhanovsky:
Fix https://github.com/tempesta-tech/tempesta/issues/576#issuecomment-230177421: do not try to push() socket close work synchronously if current context is the work queue consumer

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -477,7 +477,7 @@ tfw_http_conn_cli_dropfree(TfwHttpMsg *hmreq)
 	BUG_ON(!(TFW_CONN_TYPE(hmreq->conn) & Conn_Clnt));
 
 	if (hmreq->flags & TFW_HTTP_CONN_CLOSE)
-		ss_close(hmreq->conn->sk);
+		ss_close_sync(hmreq->conn->sk, true);
 	tfw_http_conn_msg_free(hmreq);
 }
 
--- a/tempesta_fw/sock.c
+++ b/tempesta_fw/sock.c
@@ -249,6 +249,10 @@ ss_send(struct sock *sk, SsSkbList *skb_list, bool pass_skb)
 	/*
 	 * Schedule the socket for TX softirq processing.
 	 * Only part of @skb_list could be passed to send queue.
+	 *
+	 * We can't transmit the data escaping the queueing because we have to
+	 * order transmissions and other CPUs can push data to transmit for
+	 * the socket while current CPU was servicing other sockets.
 	 */
 	if (ss_wq_push(&sw, false)) {
 		SS_WARN("Cannot schedule socket %p for transmission\n", sk);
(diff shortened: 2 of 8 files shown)


tempesta_fw/sock_clnt.c
```diff
@@ -92,7 +100,7 @@ tfw_cli_conn_release(TfwConnection *conn)
 {
 	TfwCliConnection *cli_conn = (TfwCliConnection *)conn;
 
-	del_timer_sync(&cli_conn->ka_timer);
+	del_singleshot_timer_sync(&cli_conn->ka_timer);
```
keshonok: Is this a correct use of a timer delete function? The timer is rearmed on `ss_close()` error in `tfw_sock_cli_keepalive_timer_cb()`, and for that reason it's not a single shot timer.

tempesta_fw/sock_clnt.c
```diff
@@ -57,7 +57,15 @@ tfw_sock_cli_keepalive_timer_cb(unsigned long data)
 	TfwConnection *conn = (TfwConnection *)data;
 
 	TFW_DBG("Client timeout end\n");
-	ss_close(conn->sk);
+
+	/* Close socket asynchronously to avoid deadlock on del_timer_sync(). */
+	if (ss_close(conn->sk)) {
```
milabs: What if `ss_close` fails here?

tempesta_fw/sock.c
```diff
@@ -414,42 +418,58 @@ ss_droplink(struct sock *sk)
...
+__ss_close(struct sock *sk, bool sync, bool need_drop)
 {
-	SsWork sw = {
-		.sk	= sk,
-		.action	= SS_CLOSE,
-	};
 
 	if (unlikely(!sk))
 		return SS_OK;
+	sk_incoming_cpu_update(sk);
```
milabs: It must be called few lines later withing the sync path, not here.

krizhanovsky: No, it must be done before calling `smp_processor_id() != sk->sk_incoming_cp`

tempesta_fw/sock_clnt.c
krizhanovsky: This is timer handler, so it's Ok to rearm the timer. However, since keep alive timer expired, it has sense to set lower timeout...

tempesta_fw/sock_clnt.c
krizhanovsky: Yeah, agree. Previously it was singleshot, but not for now.

tempesta_fw/http.c
```diff
@@ -477,7 +477,7 @@ tfw_http_conn_cli_dropfree(TfwHttpMsg *hmreq)
 	BUG_ON(!(TFW_CONN_TYPE(hmreq->conn) & Conn_Clnt));
 
 	if (hmreq->flags & TFW_HTTP_CONN_CLOSE)
-		ss_close(hmreq->conn->sk);
+		ss_close_sync(hmreq->conn->sk, true);
```
keshonok: I believe that we close a socket on the same CPU it was processed on. This change breaks that assumption as `tfw_http_conn_cli_dropfree()` may be called on a different CPU as well.

PR #585, opened 2016-07-13
Make a copy of IP-header while unrolling

milabs:
This fixes problem with ss_skb_fmt_src_addr() which is called from the
tfw_http_add_x_forwarded_for() function. In case of GRO, unrollong
process can take place and orignal SKB metadata will be lost.

### DIFF
--- a/tempesta_fw/ss_skb.c
+++ b/tempesta_fw/ss_skb.c
@@ -1021,6 +1021,25 @@ ss_skb_queue_coalesce_tail(SsSkbList *skb_list, const struct sk_buff *skb)
 	return 0;
 }
 
+/**
+ * Tempesta makes use of the source IP address that is kept in the IP
+ * header of the original skb @from. Copy the needed IP header contents to
+ * the new skb @to.
+ */
+static inline void
+__copy_ip_header(struct sk_buff *to, const struct sk_buff *from)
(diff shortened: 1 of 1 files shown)


tempesta_fw/ss_skb.c
```diff
@@ -1021,6 +1021,20 @@ ss_skb_queue_coalesce_tail(SsSkbList *skb_list, const struct sk_buff *skb)
...
+	const struct iphdr *ip4 = ip_hdr(from);
+	const struct ipv6hdr *ip6 = ipv6_hdr(from);
+
+	if (ip6->version == 6)
+		memcpy(to->data, ip6, sizeof(*ip6));
+	else
+		memcpy(to->data, ip4, sizeof(*ip4));
+
+	skb_reset_network_header(to);
+}
```
krizhanovsky: Why do we need to copy whole IP header? Secondly, if it's done for `tfw_http_add_x_forwarded_for()` only, then probably we can just add IP address field to TfwHttpMsg and initialize it with first skb, then `tfw_http_add_x_forwarded_for()` can take address from there.

krizhanovsky:
Ok, looks good for me. But please write good comment following our discussion which describes why do we copy the header.

PR #589, opened 2016-07-20
fix #1

### LINKED ISSUE #1: [RFC7234] Cache warnings
Tempesta MUST attach 110 warning header on [sending stale responses](https://tools.ietf.org/html/rfc7234#section-4.2.4).

Depends on #513 (cache entry freshness calculation).

sergsever:


tempesta_fw/cache.c
```diff
@@ -1150,7 +1150,7 @@ cache_req_process_node(TfwHttpReq *req, unsigned long key,
 
 	if (!tfw_cache_entry_is_live(req, ce))
 		goto out;
-
+
```
keshonok: Please remove the empty tab.

tempesta_fw/cache.c
```diff
@@ -1160,6 +1160,8 @@ cache_req_process_node(TfwHttpReq *req, unsigned long key,
 	TFW_INC_STAT_BH(cache.hits);
 
 	resp = tfw_cache_build_resp(ce);
+	 if ( tfw_cache_entry_is_live(req, ce) > ce->lifetime)
```
keshonok: This function is called just a few lines earlier. Please use it. There's not need to call it again.

Also, the conditions described in RFC7234 4.2.4. are not covered. One of those conditions is not applicable yet (related to revalidate directive as we don't process it yet), but the one related to `Age` header field probably should be taken into acccount.

Also, what about responses that are not generated from cache? The `Warning` header should be added to those as well.

tempesta_fw/http.c
```diff
@@ -683,6 +683,36 @@ tfw_http_add_hdr_via(TfwHttpMsg *hm)
 }
 
 static int
+tfw_http_add_hdr_110(TfwHttpMsg *hm)
```
keshonok: It's best to rename it to `tfw_http_add_hdr_warning_110()`.

tempesta_fw/http.h
```diff
@@ -107,11 +107,13 @@ enum {
 #define TFW_HTTP_CC_MIN_FRESH		0x00000020
 #define TFW_HTTP_CC_OIFCACHED		0x00000040
 /* Response only CC directives. */
+
 #define TFW_HTTP_CC_MUST_REVAL		0x00000100
 #define TFW_HTTP_CC_PROXY_REVAL		0x00000200
 #define TFW_HTTP_CC_PUBLIC		0x00000400
 #define TFW_HTTP_CC_PRIVATE		0x00000800
 #define TFW_HTTP_CC_S_MAXAGE		0x00001000
+#define TFW_HTTP_CC_STALE		0x00001100 /*is stale, but pass*/
```
keshonok: This is not a `Cache-Control` header field directive, so this is not the right place for the flag. It is a response flag. Perhaps `TFW_HTTP_RESP_STALE` is a better name.

tempesta_fw/http.c
```diff
@@ -748,9 +778,11 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
 		return r;
 
 	r = tfw_http_add_hdr_via(hm);
+	if (resp->cache_ctl.flags & TFW_HTTP_CC_STALE) {
+		r = tfw_http_add_hdr_110(hm);
 	if (r)
 		return r;
-
+}
```
keshonok: A missing TAB?

tempesta_fw/http.c
```diff
@@ -683,6 +683,36 @@ tfw_http_add_hdr_via(TfwHttpMsg *hm)
...
+		.eolen = 2,
+		.flags = 3 << TFW_STR_CN_SHIFT
+#undef S_Stale
+	};
+
+	r = tfw_http_msg_hdr_add(hm, &rh);
+	if (r)
+		TFW_ERR("Unable to add 110: header to msg [%p]\n", hm);
+	else
+		TFW_DBG2("Added 110: header to msg [%p]\n", hm);
```
keshonok: There's no `110` header field. The correct name is `Warning: 110`.

tempesta_fw/http.c
```diff
@@ -683,6 +683,36 @@ tfw_http_add_hdr_via(TfwHttpMsg *hm)
...
+			{ .ptr = *this_cpu_ptr(&g_buf),
+			  .len = vhost->hdr_via_len },
+		},
+		.len = SLEN(S_Stale) + vlen + vhost->hdr_via_len,
+		.eolen = 2,
+		.flags = 3 << TFW_STR_CN_SHIFT
+#undef S_Stale
+	};
+
+	r = tfw_http_msg_hdr_add(hm, &rh);
```
keshonok: All of the above is completely bogus. `vhost->hdr_via_len` is completely unrelated to the `Warning` header field. Please use `tfw_http_msg_hdr_xfrm()` to add the `Warning` header field to a response.

tempesta_fw/cache.c
```diff
@@ -1144,11 +1144,12 @@ cache_req_process_node(TfwHttpReq *req, unsigned long key,
 	TfwHttpResp *resp = NULL;
 	TDB *db = node_db();
 	TdbIter iter;
+	time_t is_live = tfw_cache_entry_is_live(req, ce);
```
keshonok: This is wrong.
`ce` is not set until later, when the record is found in the cache database. See `tfw_cache_dbce_get()`.

tempesta_fw/cache.c
```diff
@@ -1158,8 +1159,9 @@ cache_req_process_node(TfwHttpReq *req, unsigned long key,
 		ce->hdr_num, ce->hdr_len, ce->key, ce->status, ce->hdrs,
 		ce->body);
 	TFW_INC_STAT_BH(cache.hits);
-
 	resp = tfw_cache_build_resp(ce);
+	if (is_live > ce->lifetime)
+		resp->cache_ctl.flags |= TFW_HTTP_RESP_STALE;
```
keshonok: This is not a Cache-Control flag. It's a response message flag.

tempesta_fw/http.c
```diff
@@ -748,9 +748,15 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
 		return r;
 
 	r = tfw_http_add_hdr_via(hm);
+	if (resp->cache_ctl.flags & TFW_HTTP_RESP_STALE) {
+		char *s_110_n = "Warning:";
+		char *s_110_v = "110 - Response is stale.";
+		r = tfw_http_msg_hdr_xfrm(hm, s_110_n, SLEN(s_110_n),
+					  s_110_v, SLEN(s_110_v), 
+					  TFW_HTTP_HDR_RAW, 0);
 	if (r)
```
keshonok: Wrong indentation.

tempesta_fw/http.h
```diff
@@ -107,11 +107,14 @@ enum {
...
 #define TFW_HTTP_CC_OIFCACHED		0x00000040
 /* Response only CC directives. */
+
 #define TFW_HTTP_CC_MUST_REVAL		0x00000100
 #define TFW_HTTP_CC_PROXY_REVAL		0x00000200
 #define TFW_HTTP_CC_PUBLIC		0x00000400
 #define TFW_HTTP_CC_PRIVATE		0x00000800
 #define TFW_HTTP_CC_S_MAXAGE		0x00001000
+/* It is stale, but pass with a warning */
+#define TFW_HTTP_RESP_STALE		0x00001100
```
keshonok: This is not a Cache-Control flag. It's a response message flag.

tempesta_fw/http.h
```diff
@@ -107,11 +107,14 @@ enum {
 #define TFW_HTTP_CC_MIN_FRESH		0x00000020
 #define TFW_HTTP_CC_OIFCACHED		0x00000040
 /* Response only CC directives. */
+
```
keshonok: Why the empty line is added here?

tempesta_fw/http.c
```diff
@@ -748,9 +748,15 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
 		return r;
 
 	r = tfw_http_add_hdr_via(hm);
+	if (resp->cache_ctl.flags & TFW_HTTP_RESP_STALE) {
+		char *s_110_n = "Warning:";
+		char *s_110_v = "110 - Response is stale.";
```
keshonok: Perhaps, `"\"110 - Response is stale\""`? The text offered in the RFC does not contain the dot at the end, and does contain apostrophes.

keshonok: These two should better be made a `#define`.

tempesta_fw/cache.c
keshonok: From its name, `is_live` appears to be a binary variable, but it's not. Perhaps, `lifetime` is a better name here.

tempesta_fw/http.h
```diff
@@ -231,6 +231,8 @@ typedef struct {
 /* Response flags */
 #define TFW_HTTP_VOID_BODY		0x010000	/* Resp to HEAD req */
 #define TFW_HTTP_HAS_HDR_DATE		0x020000	/* Has Date: header */
+/* It is stale, but pass with a warning */
+#define TFW_HTTP_RESP_STALE		0x04000
```
keshonok: This is wrong. You missed the trailing zero. This should be `0x040000`.

tempesta_fw/http.c
```diff
@@ -748,9 +748,15 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
 		return r;
 
 	r = tfw_http_add_hdr_via(hm);
-	if (r)
-		return r;
+	if (resp->flags & TFW_HTTP_RESP_STALE) {
+#define s_110_n "Warning:"
+#define s_110_v "110 - Response is stale"
```
keshonok: Please `#undef` the definitions when they're not needed anymore.
Also, constant definitions must be in upper case according to the coding style.

Also, this is not a requirement, but I would choose a better and meaningful name. This is a `Warning` header field, but the name doesn't tell anything about that. How about `S_WARN_110_N`?

tempesta_fw/http.c
```diff
@@ -748,9 +748,15 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
 		return r;
 
 	r = tfw_http_add_hdr_via(hm);
-	if (r)
-		return r;
+	if (resp->flags & TFW_HTTP_RESP_STALE) {
+#define s_110_n "Warning:"
```
keshonok: Please make sure that `:` (colon) is required here. I am not so sure of that.

PR #592, opened 2016-07-29
fix #1

### LINKED ISSUE #1: [RFC7234] Cache warnings

sergsever:


tempesta_fw/http.c
```diff
@@ -748,9 +748,19 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
 		return r;
 
 	r = tfw_http_add_hdr_via(hm);
-	if (r)
-		return r;
-
+	if (resp->flags & TFW_HTTP_RESP_STALE) {
+#define S_WARN_110_N "Warning"
+#define S_WARN_110_V "110 - Response is stale"
+
```
keshonok: Unnecessary empty line.

keshonok:
Looks good to me.

tempesta_fw/cache.c
```diff
@@ -1144,11 +1144,12 @@ cache_req_process_node(TfwHttpReq *req, unsigned long key,
 	TfwHttpResp *resp = NULL;
 	TDB *db = node_db();
 	TdbIter iter;
+	time_t lifetime = 0;
```
krizhanovsky: `lifetime` is never used uninitialized, so there is no sense to initialize it with zero.

tempesta_fw/http.c
```diff
@@ -748,9 +748,18 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
-	if (r)
-		return r;
-
+	if (resp->flags & TFW_HTTP_RESP_STALE) {
+#define S_WARN_110_N "Warning"
+#define S_WARN_110_V "110 - Response is stale"
+		r = tfw_http_msg_hdr_xfrm(hm, S_WARN_110_N,
+ 					  SLEN(S_WARN_110_N), S_WARN_110_V,
+					  SLEN(S_WARN_110_V),
+					  TFW_HTTP_HDR_RAW, 0);
```
krizhanovsky: Why `tfw_http_msg_hdr_xfrm()` instead of `tfw_http_msg_hdr_add()`? It seems there is no problem if upstream server returns response with the warning and we retrieve it from the cache. Meantime we don't store the adjusted response in the cache and we can ovoid the header lookup overhead if we use `tfw_http_msg_hdr_add()`.

By the way, it'd be better to place `SLEN(S_WARN_110_N)` on the same line with `S_WARN_110_N`.

sergsever: With tfw_http_msg_hdr_add we should use the static initialisation of the TfwStr, that needs change after merging new TfwStr.

tempesta_fw/http.c
```diff
@@ -748,9 +748,23 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
-
+	if (resp->flags & TFW_HTTP_RESP_STALE) {
+#define S_WARN_110_N "Warning"
+#define S_WARN_110_V "110 - Response is stale"
+	/* TODO: ajust for #215 */
+	TfwStr wh = {
+	.ptr = (TfwStr []) {
+		{.ptr = S_WARN_110_N, .len = SLEN(S_WARN_110_N)},
+		{.ptr = S_WARN_110_V, .len = SLEN(S_WARN_110_V)},
+}, .len = SLEN(S_WARN_110_N) + SLEN(S_WARN_110_V), .eolen = 2,
```
krizhanovsky: Pleaxe fix the identation. It seems closing brace and `len` initialization must be on different lines

tempesta_fw/http.c
```diff
@@ -748,9 +748,23 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
 
 	r = tfw_http_add_hdr_via(hm);
-	if (r)
-		return r;
-
+	if (resp->flags & TFW_HTTP_RESP_STALE) {
+#define S_WARN_110_N "Warning"
+#define S_WARN_110_V "110 - Response is stale"
+	/* TODO: ajust for #215 */
+	TfwStr wh = {
```
keshonok: The indentation is wrong. This is an subsection of the `if{}` block.

tempesta_fw/http.c
keshonok: This header doesn't have any variable parts. It's just a constant string. Therefore, there's no need to create a complex compound string. Also, please pay attention to the semicolon, again.

tempesta_fw/http.c
```diff
@@ -748,9 +748,23 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+	.ptr = (TfwStr []) {
+		{.ptr = S_WARN_110_N, .len = SLEN(S_WARN_110_N)},
+		{.ptr = S_WARN_110_V, .len = SLEN(S_WARN_110_V)},
+}, .len = SLEN(S_WARN_110_N) + SLEN(S_WARN_110_V), .eolen = 2,
+	.flags = 3 << TFW_STR_CN_SHIFT
```
krizhanovsky: There are only 2 chunks, not 3. Also you didn't set `eolen`.

tempesta_fw/http.c
```diff
@@ -751,9 +751,16 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
 		return r;
 
 	r = tfw_http_add_hdr_via(hm);
-	if (r)
-		return r;
+	if (resp->flags & TFW_HTTP_RESP_STALE) {
+#define S_WARN_110 "Warning: 110 - Response is stale\r\n"
```
keshonok: The conventions are that headers are stored without `EOL` characters in Tempesta. A special `eolen` attribute tells the number of `EOL` bytes. So please remove `\r\n` from the string.

tempesta_fw/http.c
```diff
@@ -751,9 +751,16 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+	/* TODO: ajust for #215 */
```
keshonok: Improper indentation.

tempesta_fw/http.c
```diff
@@ -751,9 +751,16 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+		TfwStr wh = {.ptr = S_WARN_110, .len = SLEN(S_WARN_110)};
```
keshonok: `.eolen` attribute is missing.

tempesta_fw/http.c
```diff
@@ -751,9 +751,16 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
-	if (r)
-		return r;
+	if (resp->flags & TFW_HTTP_RESP_STALE) {
+#define S_WARN_110 "Warning: 110 - Response is stale"
+		/* TODO: ajust for #215 */
+		TfwStr wh = {.ptr = S_WARN_110, .len = SLEN(S_WARN_110),.eolen = 2};
+		r = tfw_http_msg_hdr_add(hm, &wh);
+		if (r)
+			return r;
+#undef S_WARN_110
```
krizhanovsky: The empty line should be after the brace

PR #594, opened 2016-08-02
Sl 477

sergsever:


tempesta_fw/http_parser.c
```diff
@@ -68,6 +68,8 @@ enum {
  */
 #define IS_CR_OR_LF(c) (c == '\r' || c == '\n')
 
+#undef isspace
+#define isspace(c) (c == ' ' || c == '\t')
```
milabs: Nope. This may looks to be correct, but I'm sure that it breaks, for example [HTTP-date](https://tools.ietf.org/html/rfc7231#section-7.1.1.1). Every occurrence of `isspace` within `http_parser` need to be verified.

sergsever: It is correct for current format of the HTTP-date. In one place of the date-processing it replaced by the IS_CR_OR_LF.

krizhanovsky:
Ok, then good to merge. `http_match.c` is subject for rework in context of TL development, so we can ignore it for now.

PR #595, opened 2016-08-03
tls: add the implementation

milabs:


krizhanovsky:
It seems `tempesta/scripts/tempesta.sh` should be updated to load/unload ttls module

tempesta_fw/http.c
```diff
@@ -408,7 +408,7 @@ tfw_http_send_504(TfwHttpMsg *hmreq)
  * the connection structure. Increment the number of users
  * of the connection structure. Initialize GFSM for the message.
  */
-static TfwMsg *
+TfwMsg *
```
krizhanovsky: Why do you remove `static` keyword for the function and functions at the below?

tempesta_fw/tls.c
```diff
@@ -19,50 +19,250 @@
...
+
+	ss_skb_queue_tail(&tls->rx_queue, skb);
+
+	r = mbedtls_ssl_handshake(&c->tls->ssl);
+	if (r == MBEDTLS_ERR_SSL_CONN_EOF) {
+		return TFW_PASS; /* more data needed */
+	} else if (r == 0) {
+		struct sk_buff *nskb;
+
+#define MAX_PAYLOAD_SIZE 1460
```
krizhanovsky: Why not to place the define to top of the file as usual?

tempesta_fw/tls.c
```diff
@@ -71,12 +271,227 @@ static TfwConnHooks tls_conn_hooks = {
...
+		.allow_none = true,
+		.allow_repeat = false,
+	},
+	{
+		"ssl_certificate_key", NULL,
+		tfw_tls_cfg_handle_crt_key,
+		.allow_none = true,
+		.allow_repeat = false,
+	},
+	{}
```
krizhanovsky: Please add appropriate description to README.md whenever your create new configuration options.

krizhanovsky:
Since the PR also contains updated and ported to Linux kernel version of mbedTLS, then all comments from https://github.com/tempesta-tech/tempesta/pull/597 are applicable to the PR as well.

tempesta_fw/http.h
```diff
@@ -353,6 +353,12 @@ bool tfw_http_parse_terminate(TfwHttpMsg *hm);
 int tfw_http_msg_process(void *conn, struct sk_buff *skb, unsigned int off);
 unsigned long tfw_http_req_key_calc(TfwHttpReq *req);
 
+/* Export HTTP connection hooks (for TLS) */
+int tfw_http_conn_init(TfwConnection *);
+void tfw_http_conn_drop(TfwConnection *);
+void tfw_http_conn_release(TfwConnection *);
+TfwMsg *tfw_http_conn_msg_alloc(TfwConnection *);
```
krizhanovsky: The HTTP functions must not be exported this way, rather GFSM state switches must be used instead. See other comments for proposed solution.

tempesta_fw/tls.c
```diff
@@ -19,50 +19,250 @@
...
+	mbedtls_ssl_free(&tls->ssl);
+	tfw_http_conn_drop(conn);
+}
+
+static TfwMsg *
+tfw_tls_conn_msg_alloc(TfwConnection *conn)
+{
+	TFW_DBG("TLS:%p %s\n", conn->tls, __func__);
+
+	return tfw_http_conn_msg_alloc(conn);
```
krizhanovsky: First of all, `conn_msg_alloc()` in TLS case is called when TLS message is arrived, that's not necessary encrypted HTTP message, but handshake message as well. In this particular case HTTP must not be bothered at all since we don't know which message we received. Rather, TSL FSM from ttls/ssl_tls.c must be activated here. Just see frang module for exmple.

tempesta_fw/tls.h
```diff
@@ -21,22 +21,27 @@
...
 #define TFW_GFSM_HTTPS_STATE(s)	((TFW_FSM_HTTPS << TFW_GFSM_FSM_SHIFT) | (s))
 enum {
 	/* HTTPS FSM initial state, not hookable. */
-	TFW_HTTPS_FSM_INIT		= TFW_GFSM_HTTPS_STATE(0),
-
-	/* TODO */
-	TFW_HTTPS_FSM_TODO_ISSUE_81	= TFW_GFSM_HTTPS_STATE(1),
-
+	TFW_HTTPS_FSM_INIT	= TFW_GFSM_HTTPS_STATE(0),
+	TFW_HTTPS_FSM_CHUNK	= TFW_GFSM_HTTPS_STATE(1),
```
krizhanovsky: We use upcall design for GFSM protocols. For example, HTTP FSM provides set of hookable states for higher layers, e.g. Frang FSM gets control when HTTP FSM reaches TFW_HTTP_FSM_REQ_MSG state. TLS lays under HTTP, i.e. it must expose its FSM states to HTTP while the last one must get control when TLS reaches particular states. Basically there should be TLS states from ttls/ssl_tls.c. Since we're going to maintaing TLS as separate module, this states should be moved to ttls/, just like Frang implements its FSM.

tempesta_fw/connection.h
```diff
@@ -87,6 +89,7 @@ typedef struct {
 	struct list_head	msg_queue;
 	spinlock_t		msg_qlock;
 	atomic_t		refcnt;
+	TfwTlsContext		*tls;
```
krizhanovsky: I'd say that TLS connection (lets name it `TfwTlsConnection`) inherits `TfwConnection`.

tempesta_fw/http.c
```diff
@@ -1401,6 +1402,9 @@ tfw_http_msg_process(void *conn, struct sk_buff *skb, unsigned int off)
 {
 	TfwConnection *c = (TfwConnection *)conn;
 
+	TFW_DBG("Add skb %p to message %p\n", skb, c->msg);
+	ss_skb_queue_tail(&c->msg->skb_list, skb);
```
krizhanovsky: Why the call is moved here from `tfw_connection_recv()`? I think TLS message can theoretically be compound from 1-byte TCP segents.

tempesta_fw/connection.c
```diff
@@ -94,6 +94,8 @@ tfw_connection_release(TfwConnection *conn)
 int
 tfw_connection_send(TfwConnection *conn, TfwMsg *msg)
 {
+	if (conn->tls)
+		return tfw_tls_send_msg(conn->tls, msg);
 	return ss_send(conn->sk, &msg->skb_list, msg->ss_flags);
```
krizhanovsky: Ok, this has sense: we can think about `TfwConnection` as 6th (presentation) OSI layer on top of TCP sockets, so the layer can be encrypted or not. I think we don't need to maintain two TfwConnection-like interfaces for encrypted and raw data.

Please update `connection.h` header comment: replace `...generic connection (at OSI level 4) management` by `...generic connection management at OSI level 6 (presentation)`

tempesta_fw/tls.c
```diff
@@ -19,50 +19,250 @@
...
+		TFW_DBG2("TLS:%p read %d bytes [%.*s]\n", tls, r, r, nskb->data);
 
-	return r;
+		skb_trim(nskb, r);
+
+#undef MAX_PAYLOAD_SIZE
+
+		r = tfw_gfsm_move(&c->msg->state, TFW_HTTPS_FSM_CHUNK, nskb, 0);
+		if (r == TFW_BLOCK)
+			return TFW_BLOCK;
```
krizhanovsky: This is were do we move to state of processing new data chunk which we've just decrypted. HTTP establishes its hook for the TLS state and must call `tfw_http_conn_msg_alloc()` on its own when it's in `TFW_HTTPS_FSM_INIT` - now the state must be initialized by GFSM rather than by `tfw_http_conn_msg_alloc()` (see similar behaviour for `TFW_FRANG_FSM_INIT` when Frang sets HTTP hooks).

tempesta_fw/tls.h
```diff
@@ -21,22 +21,27 @@
...
 	TFW_HTTPS_FSM_DONE	= TFW_GFSM_HTTPS_STATE(TFW_GFSM_STATE_LAST)
 };
 
+typedef struct {
+	mbedtls_ssl_context	ssl;
+	SsSkbList		rx_queue;
+	SsSkbList		tx_queue;
```
krizhanovsky: The queues seem like dirty workaround to be able to work with mbedTLS w/o reworking its IO and FSM internals mostly placed in ttls/ssl_tls.c. We can leave with them for the very first release, but make a TODO comment to remve them in the new design.

tls/Makefile
```diff
@@ -16,4 +16,4 @@
 # this program; if not, write to the Free Software Foundation, Inc., 59
 # Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 
-obj-m = ttls/ programs/
+obj-m = ttls/
```
krizhanovsky: This and the rest at the blow seems just the same as https://github.com/tempesta-tech/tempesta/pull/597 (update for mbedTLS). So please finish #597 first and commit it to master. After that this pull request can be reduced to Tempesta FW related stuff only.

milabs:
Improved and merged within https://github.com/tempesta-tech/tempesta/pull/603

PR #596, opened 2016-08-03
mbedtls-mini: it works (occasionally)

milabs:
Review-only branch.


tempesta_fw/Makefile
```diff
@@ -55,4 +62,7 @@ tempesta_fw-objs = \
 	vhost.o \
 	work_queue.o
 
-obj-m	+= log/ classifier/ stress/ sched/ t/
+obj-m	+= log/ classifier/ sched/
+obj-y	+= mbedtls/
```
krizhanovsky: TLS should be standalone module.

tempesta_fw/mbedtls/lk.c
```diff
@@ -0,0 +1,18 @@
+#include "lk.h"
+
+void *calloc(size_t n, size_t size)
+{
+	return kzalloc(n * size, GFP_ATOMIC);
```
krizhanovsky: This dummy functiona redefinition can work for very first version of code. We have fast pool allocation which should be used instead of poor `kmalloc()`. Moreover, we need carefully review all allocations and do our best to avoid them.

tempesta_fw/connection.c
```diff
@@ -94,6 +94,27 @@ tfw_connection_release(TfwConnection *conn)
...
 tfw_connection_send(TfwConnection *conn, TfwMsg *msg)
 {
+	if (conn->tls) {
+		struct sk_buff *skb;
+
+		while ((skb = ss_skb_dequeue(&msg->skb_list))) {
+			int r, i;
+
+			if (skb_headlen(skb)) {
+				r = mbedtls_ssl_write(&conn->tls->ctx, skb->data, skb_headlen(skb));
```
krizhanovsky: Too long lines. Pleas keep code nice looking ever in prototypes.

tempesta_fw/tls.c
```diff
@@ -19,8 +19,19 @@
...
+#include <mbedtls/certs.h>
+#include <mbedtls/debug.h>
+#include <mbedtls/error.h>
+
 #include "connection.h"
 #include "tls.h"
+#include "http_msg.h"
+
+static mbedtls_ssl_config tfw_tls_conf;
+static mbedtls_x509_crt tfw_tls_srvcert;
```
krizhanovsky: User should be able to configure path to the certificate in `tempesta_fw.conf`

tempesta_fw/tls.c
```diff
@@ -31,38 +42,169 @@
...
 tfw_tls_conn_drop(TfwConnection *conn)
 {
+	TfwTlsContext *tls = conn->tls;
+
+	BUG_ON(!tls);
+
+	printk("%s\n", __func__);
+
+	mbedtls_ssl_free(&tls->ctx);
+	kfree(tls);
```
krizhanovsky: This is constant size structure, so SLAB should be used

tempesta_fw/connection.c
krizhanovsky: Basically this is the point where and why we can't live with mbedTLS and must develop our own fork. MbedTLS and Tempesta live in different paradigms. Actually mbedTLS was never developed with performance in mind (see comment https://github.com/tempesta-tech/tempesta/issues/81#issuecomment-128968333: "Probably there are communication issues (actually there are and we saw some of them) "). In particular:

1. mbedTLS doesn't care about IO while Tempesta processes packets immediately. We shouldn't adjust Tempesta code to work with mbedTLS buffers, but rather adjust mbedTLS such that it'll be aware about skb;

2. mbedTLS handles it's own states (e.g. see `ssl->state != MBEDTLS_SSL_HANDSHAKE_OVER` in `mbedtls_ssl_write()`) and does it in poor manner. Tempesta has GFSM to handle state machines efficiently.

All in all basically you need to fully redesign at least ttls/ssl_tls.c.

tempesta_fw/tls.c
```diff
@@ -31,38 +42,169 @@
...
+	TfwTlsContext *tls = c->tls;
+	struct sk_buff *skb = alloc_skb(MAX_TCP_HEADER + len, GFP_ATOMIC);
+
+	if (unlikely(!skb))
+		return MBEDTLS_ERR_NET_SEND_FAILED;
+
+	skb_reserve(skb, MAX_TCP_HEADER);
+	skb_put(skb, len);
+
+	if (unlikely(skb_store_bits(skb, 0, buf, len)))
```
krizhanovsky: That's just performance killer!

1. you need to allocate new skb only if it's cloned/shared, however we already handle this at SS layer;

2. `mbedtls_ssl_write()` does lots of memory copies (just see `memcpy()` and `memmove()` calls in it) to encrypt the daata and after that you make one more copy here.

Consider you call `tfw_connection_send()` just to send some static response (e.g. `tfw_http_send_502()`) - why should we allocate the skb? Tempesta code has already created skbs.

tempesta_fw/Makefile
```diff
@@ -22,6 +22,13 @@ ifdef DEBUG
 	EXTRA_CFLAGS += -DDEBUG=$(DEBUG)
 endif
 
+MBEDTLS = $(obj)/mbedtls
+MBEDTLS_CFLAGS = -I$(MBEDTLS)/include -I$(MBEDTLS)/mbedtls/include \
+	-include $(MBEDTLS)/lk.h -DMBEDTLS_CONFIG_FILE='"$(MBEDTLS)/config.h"'
+export MBEDTLS MBEDTLS_CFLAGS
+
+EXTRA_CFLAGS += $(MBEDTLS_CFLAGS)
```
krizhanovsky: As it's explained at the below, this is not mbedtls, rather it's our own, very specific, fork of the library. So all the variables and functions should be named as `TTLS` (Tempesta TLS) instead of `MBEDTLS`.

PR #597, opened 2016-08-06
ttls: update to mbedTLS 2.3.0 (GPL) version

milabs:


tls/Makefile
```diff
@@ -16,4 +16,4 @@
 # this program; if not, write to the Free Software Foundation, Inc., 59
 # Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 
-obj-m = ttls/ programs/
```
krizhanovsky: I believe we don't need the testing programs and can remove `tls/programs` at all.

tls/ttls/ssl_srv.c
```diff
@@ -30,24 +30,26 @@
...
 
+#if defined(MBEDTLS_PLATFORM_C)
+#include "platform.h"
+#else
+#include <stdlib.h>
+#define mbedtls_calloc    calloc
+#define mbedtls_free      free
+#define mbedtls_time      time
+#define mbedtls_time_t    time_t
+#endif
```
krizhanovsky: I think we should split (at least imaginary) mbedTLS code to two pieces: system dependent code and crypto algorithms implementation. The first piece is supposed to be reworked to be well integrated with Tempesta infrastructure, so we can just roughly rewrite system and libc calls (e.g. just replace `malloc()` by `kmalloc()` calls w/o intermediate `mbedtls_` defines). The second part is good to leave as is with as minimum changes as possible to be able to apply patches from further mbedTLS versions. However, the crypto routines should be considered to merge woth linux/crypto.

So in this case and few similar cases at the below (e.g for pthread calls), it seems the `mbedtls_` defines are unnecessary.

tls/ttls/ttls.c
```diff
@@ -0,0 +1,100 @@
...
+
+void *calloc(size_t n, size_t size)
+{
+	return kzalloc(n * size, GFP_ATOMIC);
+}
+
+void free(void *ptr)
+{
+	kfree(ptr);
+}
```
krizhanovsky: Please mark the functions as TODO to analyze all the calls over the whole code and carefully change all of them to appropriate kernel routines (probably SLAB allocations can be used) or exclude at all (if the optimization is possible).

tls/ttls/ttls.c
```diff
@@ -0,0 +1,100 @@
+#include <linux/module.h>
```
krizhanovsky: Please always write header comments in new files

tls/ttls/ttls.c
```diff
@@ -0,0 +1,100 @@
...
+
+static void
+ttls_exit(void)
+{
+}
+
+module_init(ttls_init);
+module_exit(ttls_exit);
+
+MODULE_LICENSE("GPLv2");
```
krizhanovsky: Please add module version and developer as well

tls/ttls/ttls.h
```diff
@@ -0,0 +1,36 @@
+#ifndef __TTLS_H__
```
krizhanovsky: Please always write header comments in new files

krizhanovsky:
Please move ttls/ content to tls since there is only README.md and merge to master

PR #600, opened 2016-08-13
Fix #235:

### LINKED ISSUE #235: Link all single client connections with an appropriate peer
Currently a peer is created for each client connection even though the connections may be coming from the same client. As a result there are multiple `TfwClient{}` instances for the same peer.

One `TfwClient{}` instance should be created for each client. An incoming connection should be matched with a specific peer (`TfwClient{}`), and it should be linked with that peer. In the end, there should be a large number of client connections linked to a smaller number of peers.

Peers are identified by a specific set of attributes, such as an IP address, etc. So there's a task of maintaining a store of peers that provides for fast insert/delete/search, perhaps a hash. `Frang` deals with a very similar, if not the same task. There needs to be a single store of peers that Frang, the client code, and the rest of code can use.

Frang needs to be reworked to use a single store of peers.

krizhanovsky:
* Introduce clients hash table, now new connection just get()'s TfwClient;
* Since client can have many HTTP sessions, TfwSession is introduced to
  handle different sticky cookies for the same client;
* Now received sticky cookie is verified, so timestamp is added to cookie value;
* To keep HTTP sessions persistent sticky_secret config option is introduced;
* Config option sess_lifetime is also introduced to limit sessions lifetime;
* Sticky cookie is just a mechanism for HTTP sessions, so http_sticky is renamed
  to http_sess;
* tfw_addr_get_sk_saddr() was removed as doubling for ss_getpeername();
* TfwHttpMsg was replaced by TfwHttpReq and TfwHttpResp to make the logic explicit
  about what the arguements are;
* tfw_http_prep_hexstring() is removed since there are standard bin2hex().


tempesta_fw/classifier/frang.c
```diff
@@ -1132,17 +1091,15 @@ frang_init(void)
 {
 	int r;
 
-	TFW_DBG("Frang module init");
-	frang_mem_cache = KMEM_CACHE(frang_account_t, 0);
-	if (!frang_mem_cache) {
-		TFW_ERR("frang: can't create cache\n");
+	if (unlikely(sizeof(FrangAcc) > sizeof(TfwClassifierPrvt))) {
+		TFW_ERR("frang: seems miscompilation problem\n");
```
keshonok: It might be better to replace this with `BUILD_BUG_ON()`.

tempesta_fw/classifier/frang.c
```diff
@@ -479,7 +435,8 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
 	 * but don't enforce the policy for HTTP older than 1.1.
 	 */
 	if (TFW_STR_EMPTY(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST])) {
-		frang_msg("Host header field", &ra->addr, " is missed\n");
+		frang_msg("Host header field", &FRANG_ACC2CLI(ra)->addr,
+			  " is missed\n");
```
keshonok: is missing

tempesta_fw/client.c
```diff
@@ -4,83 +4,140 @@
...
- * Find a client corresponding to the @sk.
+ * Find a client corresponding to the @sk by IP addres.
+ * More advanced identification is possible based on User-Agent,
+ * Cookie and other HTTP headers.
  *
  * The returned TfwClient reference must be released via tfw_client_put()
  * when the @sk is closed.
+ *
+ * TODO #100: evict connections and/or clients and drop their accouning.
+ * Probably FrangAcc.last_ts should be moved to TfwClient to the purpose.
```
keshonok: `last_ts` has been removed from Frang in this patch set.

tempesta_fw/http_sess.c
```diff
@@ -0,0 +1,731 @@
...
+
+	if (tfw_cfg_check_single_val(ce))
+		return -EINVAL;
+	if (len > STICKY_KEY_MAXLEN)
+		return -EINVAL;
+
+	memset(tfw_sticky_key, 0, STICKY_KEY_MAXLEN);
+	memcpy(tfw_sticky_key, ce->vals[0], len);
+	r = crypto_shash_setkey(tfw_sticky_shash, (u8 *)tfw_sticky_key,
+				STICKY_KEY_MAXLEN);
```
keshonok: Is there a reason that `STICKY_KEY_MAXLEN` is used instead of `len`?

tempesta_fw/http_sess.c
```diff
@@ -0,0 +1,731 @@
...
+	.start = tfw_cfg_sess_start,
+	.stop = tfw_cfg_sess_stop,
+	.specs = (TfwCfgSpec[]) {
+		{
+			.name = "sticky",
+			.handler = tfw_http_sticky_cfg,
+			.allow_none = true,
+		},
+		{
+			.name = "sticky_secret",
```
keshonok: Is there the intention for the `sticky_secret` directive that it may somehow apply to all `sticky` directives? If not, then, perhaps, this should better be a part of the `sticky` directive:
`sticky name=abracadabra secret=blablabla enforce;`

tempesta_fw/http_sess.c
```diff
@@ -0,0 +1,731 @@
...
+		 * from the hash table.
+		 */
+		kmem_cache_free(sess_cache, sess);
+}
+
+/**
+ * Obtains appropriate HTTP session for the request based on Sticky cookies.
+ */
+int
+tfw_http_sess_obtain(TfwHttpReq *req)
```
keshonok: Perhaps, `tfw_http_sess_set()` is a better name? Or, `tfw_http_sess_getandset()`. `obtain` makes me think that the session handle is returned to the caller. Instead, it's actually set for the request `req` inside the function.

tempesta_fw/http_sess.c
```diff
@@ -0,0 +1,731 @@
...
+	 * possible. The only thing which we can do is to enforce the cookie.
+	 * However, we cal loose innocent clients w/ disabled cookies.
+	 * We leave this for administrator decision or more progressive DDoS
+	 * mitigation techniques.
+	 */
+
+	if (!sv.ts) {
+		/* No sticky cookie in request and no enforcement. */
+		sv.ts = jiffies;
+		if (__sticky_calc(req, &sv))
```
keshonok: The above two lines of code make up the `tfw_http_sticky_calc()` function. Why not use it instead?

tempesta_fw/http_sess.c
```diff
@@ -0,0 +1,731 @@
...
+
+	hb = &sess_hash[hash_min(key, SESS_HASH_BITS)];
+
+	spin_lock(&hb->lock);
+
+	hlist_for_each_entry_safe(sess, tmp, &hb->list, hentry) {
+		/* Collect garbage first to not to return expired session. */
+		if (sess->expires < jiffies) {
+			hash_del(&sess->hentry);
+			tfw_http_sess_put(sess);
```
keshonok: Requests may potentially stay for a long time in Tempesta due to various reasons. What if that session is still part of some request? Then the `req->sess` pointer will become stale. Is there something that I am missing here?

tempesta_fw/http.h
```diff
@@ -223,7 +225,7 @@ typedef struct {
 #define TFW_HTTP_CHUNKED		0x000004
 
 /* Request flags */
-#define TFW_HTTP_STICKY_SET		0x000100	/* Need 'Set-Cookie` */
+#define TFW_HTTP_HAS_STICKY_COOKIE	0x000100
```
keshonok: Perhaps, just `TFW_HTTP_HAS_STICKY` would be better. Additional `_COOKIE` doesn't add anything to the understanding of this constant.

tempesta_fw/http_sess.c
```diff
@@ -0,0 +1,731 @@
...
+		tfw_http_msg_clnthdr_val(hdr, TFW_HTTP_HDR_USER_AGENT,
+					 &ua_value);
+
+	addr_len = tfw_addr_sa_len(addr);
+
+	memset(desc, 0, sizeof(desc));
+	shash_desc->tfm = tfw_sticky_shash;
+	shash_desc->flags = 0;
+
+	TFW_PRINT_STICKY_COOKIE(addr, &ua_value, sv);
```
keshonok: Maybe `_DBG_PRINT_` or `_DPRINT_`? I didn't get it at first, so I may not be alone.

tempesta_fw/classifier.h
```diff
@@ -28,6 +28,11 @@
 #include "tempesta_fw.h"
 #include "connection.h"
 
+/* Size of classifier private cliet accounting data. */
+#define TFW_CLASSIFIER_ACCSZ	256
+
+typedef struct { char _[TFW_CLASSIFIER_ACCSZ]; } TfwClassifierPrvt;
```
keshonok: What guarantees that the actual data will not exceed this value? I think there should be a compile-time check for that somewhere.

tempesta_fw/http_sess.c
```diff
@@ -0,0 +1,731 @@
...
+	if (r)
+		TFW_WARN("Cannot add \"" S_F_SET_COOKIE "%.*s=%.*s\"\n",
+			 PR_TFW_STR(&tfw_cfg_sticky.name), len, buf);
+	return r;
+}
+
+/*
+ * No Tempesta sticky cookie found.
+ *
+ * Create Tempesta sticky cookie value, and store it for future use.
```
keshonok: This comment is no longer valid for this function. That's done in `tfw_http_sess_obtain()` now.

tempesta_fw/http_sess.c
krizhanovsky: `obtain` is because we search and return or create a new session, just like `tfw_client_obtain()`. I wished to express the allocation/lookup logic rather than return value.

tempesta_fw/http_sess.c
krizhanovsky: Yes, good catch! We should initialize session reference counter to 1 since the hash is the initial user of the session.

tempesta_fw/http_sess.c
krizhanovsky: I just prefer simple configuration options like `name = val`, there is nothing good in complex configuration syntax.

Issue #564, opened 2016-06-18
Application Performance Monitoring (APM)


krizhanovsky:
Tempesta must provide APM for advanced DDoS mitigation logic as well as for system administration purposes. APM must extend `/proc/tempesta/perfstat` by application response time in milliseconds:

* maximum

* minimum

* middle value

* percentiles (50, 75, 90, 95)

krizhanovsky:
[The implementation and basic test](https://github.com/tempesta-tech/tempesta/commit/7922d9a5df41ab2166f7492b91f3b906849065ab) for algorithm calculating percentilies and their trends on data stream.

keshonok:
Implemented in 4e1e7a3

Issue #565, opened 2016-06-19
Advanced Load Balancing


krizhanovsky:
Load balancing algorithms must be extended in following way.

Firstly, Round-robin algorithm must be renamed to Ratio and extended by weights, probably dynamic. E.g. if we have 2 servers and the first has two times more CPUs than the second, then we can say that the first one has weight 2 and the first one 1. I.e. the scheduler must send requests in [1, 1, 2] instead of current [1, 2] manner. Generally speaking current RR scheme is exactly Ratio where all the servers have equal weights.

The weights can be configured manually or left for dynamic calculation. I.e. default weights for all the servers is 1 and a user can configure manually the server weights or specify `dynamic` for dynamic weights calculation. Dynamic ratio depends on #564: the weights relation is exactly relation of maximum, minimum, average or particular percentile (must be configured as well, average is the default) response time for last `T` seconds (`T` should be also configurable, 5 minutes by default).

The next option for Dynamic ration is more intelligent Predictive which must consider trend in backend response time as well as the response time itself. The exact algorithm for current weight calculation is TBD.

It worth to mention that we don't implement Fastest and Least connections algorithms like [F5 does it](https://devcentral.f5.com/articles/intro-to-load-balancing-for-developers-ndash-the-algorithms), because:
1. Fastest is essentially Dynamic Ratio, but firstly causes congestion/bufferbloat problem and then degrades the server ratio. It's pretty stupid way to balance load;
2. Tempesta FW makes significant efforts to pass HTTP traffic to backend servers in persistent connections as efficiently as possible, so all the upstreams have always equal number of connections. BTW https://github.com/tempesta-tech/tempesta/pull/423 must be resolved ASAP (this is TODO for me and @keshonok).

`sched_srv_conn` from #593 must also be implemented.

Also please update/develop appropriate unit tests as for current schedulers.

krizhanovsky:
[The implementation and basic test](https://github.com/tempesta-tech/tempesta/commit/7922d9a5df41ab2166f7492b91f3b906849065ab) for algorithm calculating percentilies and their trends on data stream.

krizhanovsky:
[...]

PR #601, opened 2016-08-15
Implement APM stats for server response times. (#564)

keshonok:
Response time for each request is accumulated over a predefined time
period for each configured back end server, and used for calculation
of percentiles. Accumulated data for the latest time period is used
in the calculation. That gives a fresh impression on the performance
of a specific server.

Server's APM stats are available now by reading the file in /proc:
/proc/tempesta/servers/<server IP address>:<port>

### DIFF
--- a/tempesta_fw/apm.c
+++ b/tempesta_fw/apm.c
@@ -18,12 +18,11 @@
  * 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  * Prototype for fast precentiles calculation.
  */
+#include <linux/atomic.h>
 #include <linux/kernel.h>
 #include <linux/slab.h>
-#include <linux/atomic.h>
 #include <linux/spinlock.h>
 #include <linux/stringify.h>
-#include <linux/sort.h>
 
 #include "apm.h"
 #include "cfg.h"
@@ -238,7 +237,6 @@ tfw_stats_adjust(TfwPcntRanges *rng, int r, spinlock_t *spinlock)
 				   &rng->cnt[r][0]);
 			for (i = TFW_STATS_BCKTS / 2; i < TFW_STATS_BCKTS; ++i)
 				atomic_set(&rng->cnt[r - 1][i], cnt);
-
 		}
 		/*
 		 * Fall through to reduce the range order. The first bucket
@@ -252,10 +250,11 @@ tfw_stats_adjust(TfwPcntRanges *rng, int r, spinlock_t *spinlock)
 	 * If servers are too fast (all responses within 1ms), then there's
 	 * nothing to do here.
 	 */
-	if (r) {
-		pc.atomic = rng->ctl[r].atomic;
-		if (likely(pc.order))
-			__range_shrink_left(rng, &pc, r);
+	if (!r)
+		goto out;
+	pc.atomic = rng->ctl[r].atomic;
+	if (likely(pc.order)) {
+		__range_shrink_left(rng, &pc, r);
 	}
 out:
 	spin_unlock(spinlock);
@@ -328,7 +327,29 @@ typedef struct {
 	unsigned long	jtmistamp;
 	atomic_t	reset;
 	spinlock_t	spinlock;
-} TfwApmRBufEnt;
+} TfwApmRBEnt;
+
+/*
+ * The ring buffer contol structure.
+ * This is a supporting structure used in the calculation of percentiles.
+ * Also it keeps related data that are useful in making decisions on the
+ * recalculation of percentiles.
+ * @ctlent	- An array of pointers to the response time ranges data.
+ * @centry_ctl	- A copy of response time ranges of the current entry.
+ * @fentry_ctl	- A copy of response time ranges of the first entry.
+ * @jtmwstamp	- The start of the time window the percentiles are for.
+ * @entry_cnt	- The number of hits in the current buffer ring entry.
+ * @total_cnt	- The number of hits within the current time window.
+ */
+
+typedef struct {
+	TfwPcntCtl	**ctlent;
+	TfwPcntCtl	centry_ctl[TFW_STATS_RANGES];
+	TfwPcntCtl	fentry_ctl[TFW_STATS_RANGES];
+	unsigned long	jtmwstamp;
+	unsigned long	entry_cnt;
+	unsigned long	total_cnt;
+} TfwApmRBCtl;
 
 /*
  * The ring buffer structure.
@@ -336,7 +357,7 @@ typedef struct {
  * @rbufsz	- The size of @rbent.
  */
 typedef struct {
-	TfwApmRBufEnt	*rbent;
+	TfwApmRBEnt	*rbent;
 	int		rbufsz;
 } TfwApmRBuf;
 
@@ -373,41 +394,6 @@ typedef struct {
 } TfwApmStats;
 
 /*
- * The ring buffer list entry structure.
- * This is a supporting structure used in the calculation of percentiles.
- * @ctl		- A pointer to the response time ranges data for the entry.
- * @pcntrng	- A pointer to the stats data for the ring buffer entry.
- */
-typedef struct {
-	TfwPcntCtl	*ctl;
-	TfwPcntRanges	*pcntrng;
-} TfwApmRBLstEnt;
-
-/*
- * The ring buffer list structure. 
- * This is a supporting structure used in the calculation of percentiles.
- * The list consists of entries of the ring buffer that take part in the
- * calculation of percentiles. Also it keeps related data that are useful
- * in making decisions on the recalculation of percentiles.
- * @entry_ctl	- A copy of response time ranges of the current entry.
- * @rblstent	- The array of ring buffer entries for the calculation.
- * @rblstsz	- The current size of @rblstent.
- * @rblstmaxsz	- The maximum size of @rblstent.
- * @jtmwstamp	- The start of the time window the percentiles are for.
- * @entry_cnt	- The number of hits in the current buffer ring entry.
- * @total_cnt	- The number of hits within the current time window.
- */
-typedef struct {
-	TfwPcntCtl	entry_ctl[TFW_STATS_RANGES];
-	TfwApmRBLstEnt	*rblstent;
-	unsigned int	rblstsz;
-	unsigned int	rblstmaxsz;
-	unsigned long	jtmwstamp;
-	unsigned long	entry_cnt;
-	unsigned long	total_cnt;
-} TfwApmRBLst;
-
-/*
  * APM Data structure.
  * Note that the organization of the supporting data heavily depends
  * on the fact that there's only one party that does the calculation
@@ -421,11 +407,11 @@ typedef struct {
  * @flags	- The atomic flags (see below).
  */
 #define TFW_APM_DATA_F_REARM	0x0001		/* Re-arm the timer. */
-#define TFW_APM_TIMER_TIMEOUT	(HZ * 3)	/* The timer periodicity. */
+#define TFW_APM_TIMER_TIMEOUT	(HZ)		/* The timer periodicity. */
 
 typedef struct {
 	TfwApmRBuf		rbuf;
-	TfwApmRBLst		rblst;
+	TfwApmRBCtl		rbctl;
 	TfwApmStats		stats;
 	struct timer_list	timer;
 	unsigned long		flags;
@@ -442,9 +428,13 @@ static const TfwPcntCtl __read_mostly tfw_rngctl_init[TFW_STATS_RANGES] = {
 	{{4, 109, 349}}
 };
 
-/* A superset of percentiles for all users. */
-const unsigned int __read_mostly tfw_apm_prcntl_ith[] = {
-	1, 50, 75, 90, 95, 99
+/*
+ * A superset of percentiles for all users. Note that 1%, 50%, and
+ * 99% percentiles are used to tell the minimum, the median, and the
+ * maximum values correspondingly.
+ */
+static const Percentile __read_mostly tfw_apm_prcntl[] = {
+	{1}, {50}, {75}, {90}, {95}, {99}
 };
 
 static int tfw_apm_jtmwindow;		/* Time window in jiffies. */
@@ -471,36 +461,40 @@ typedef struct {
 } __attribute__((packed)) TfwApmRBEState;
 
 static inline void
-__tfw_apm_state_set(TfwApmRBEState *st, u16 v, u8 r, u8 b)
+__tfw_apm_state_set(TfwApmRBEState *st, u16 v, u8 i, u8 r, u8 b)
 {
 	st->v = v;
+	st->i = i;
 	st->r = r;
 	st->b = b;
 }
 
 static void
-tfw_apm_state_next(TfwApmRBLstEnt *rblstent, TfwApmRBEState *st)
+__tfw_apm_state_next(TfwPcntRanges *rng, TfwPcntCtl *ctl, TfwApmRBEState *st)
 {
-	int r, b;
+	int i = st->i, r, b;
 	unsigned short rtime;
 
+	for (r = i / TFW_STATS_BCKTS; r < TFW_STATS_RANGES; ++r) {
+		for (b = i % TFW_STATS_BCKTS; b < TFW_STATS_BCKTS; ++b, ++i) {
+			if (!atomic_read(&rng->cnt[r][b]))
+				continue;
+			rtime = ctl[r].begin + (b << ctl[r].order);
+			__tfw_apm_state_set(st, rtime, i, r, b);
+			return;
+		}
+	}
+	__tfw_apm_state_set(st, USHRT_MAX, i, r, b);
+}
+
+static void
+tfw_apm_state_next(TfwPcntRanges *rng, TfwPcntCtl *ctl, TfwApmRBEState *st)
+{
 	/* See if there're no more buckets. */
 	if (st->i >= TFW_STATS_TOTAL_BCKTS)
 		return;
 	++st->i;
-	r = st->i / TFW_STATS_BCKTS;
-	b = st->i % TFW_STATS_BCKTS;
-	for ( ; r < TFW_STATS_RANGES; ++r) {
-		for ( ; b < TFW_STATS_BCKTS; ++b, ++st->i) {
-			if (!atomic_read(&rblstent->pcntrng->cnt[r][b]))
-				continue;
-			rtime = rblstent->ctl[r].begin
-				+ (b << rblstent->ctl[r].order);
-			__tfw_apm_state_set(st, rtime, r, b);
-			return;
-		}
-	}
-	__tfw_apm_state_set(st, USHRT_MAX, r, b);
+	__tfw_apm_state_next(rng, ctl, st);
 }
 
 /*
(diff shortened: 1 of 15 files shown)


tempesta_fw/apm.c
```diff
@@ -0,0 +1,734 @@
...
+	if (r_time <= pc2.end) {
+		TfwPcntCtl pc0, pc1 = { .atomic = rng->ctl[1].atomic };
+		if (pc1.end < r_time) {
+			atomic_inc(__rng(&pc2, rng->cnt[2], r_time));
+			tfw_stats_adjust(rng, 2);
+			return;
+		}
+
+		pc0.atomic = rng->ctl[0].atomic;
+//		BUG_ON(pc0.begin != 1); /* left bound is never moved */
```
krizhanovsky: Must be uncommented?

tempesta_fw/apm.c
```diff
@@ -0,0 +1,734 @@
...
+	pc3.atomic = rng->ctl[3].atomic;
+	if (unlikely(r_time > pc3.end)) {
+		tfw_stats_extend(rng, r_time);
+		pc3.atomic = rng->ctl[3].atomic;
+	}
+	atomic_inc(__rng(&pc3, rng->cnt[3], r_time));
+	tfw_stats_adjust(rng, 3);
+}
+
+
```
krizhanovsky: Extra empty line

tempesta_fw/apm.c
```diff
@@ -0,0 +1,734 @@
...
+	int		rbentsz;
+} TfwApmRBuf;
+
+/*
+ * APM Data structure.
+ * @rbuf	The ring buffer.
+ */
+typedef struct {
+	TfwApmRBuf	rbuf;
+} TfwApmData;
```
krizhanovsky: It seems there is no sense to have single member structure

tempesta_fw/apm.c
```diff
@@ -0,0 +1,734 @@
...
+	{{1, 17, 47}},
+	{{2, 48, 108}},
+	{{4, 109, 349}}
+};
+
+static int tfw_apm_jtmwindow;		/* Time window in jiffies. */
+static int tfw_apm_jtmintrvl;		/* Time interval in jiffies. */
+static int tfw_apm_tmwscale;		/* Time window scale. */
+
+
```
krizhanovsky: extra empty line

tempesta_fw/http.h
```diff
@@ -252,6 +252,7 @@ typedef struct {
 	unsigned char	version;					\
 	unsigned int	flags;						\
 	unsigned long	content_length;					\
+	unsigned long	jtstamp;					\
```
krizhanovsky: Please describe the filed in the comment.

tempesta_fw/apm.c
```diff
@@ -0,0 +1,734 @@
...
+tfw_apm_rngmerge(TfwPcntCalc *pcalc, TfwPcntRanges **rnglst, size_t rlsz)
+{
+	int i, r, b, pcsz;
+	TfwPcntCtl pctl[TFW_STATS_RANGES];
+
+	for (i = 0, pcsz = 0; i < rlsz; ++i) {
+		spin_lock(&tfw_stats_sa_guard);
+		for (r = 0; r < TFW_STATS_RANGES; ++r)
+			pctl[r].atomic = rnglst[i]->ctl[r].atomic;
+		spin_unlock(&tfw_stats_sa_guard);
```
krizhanovsky: The spin-lock must be more granular, like per ring buffer entry.

tempesta_fw/apm.c
```diff
@@ -0,0 +1,734 @@
...
+	__tfw_apm_update(&((TfwApmData *)data)->rbuf, jtstamp, jrtime);
+}
+
+/*
+ * Create and initialize an APM ring buffer for a server.
+ */
+void *
+tfw_apm_create(void)
+{
+	int i, size, rbentsz = tfw_apm_tmwscale + 2;
```
krizhanovsky: Please comment why do we need these 2 extra entries,

tempesta_fw/apm.c
```diff
@@ -0,0 +1,734 @@
...
+	 */
+	jtmwstart = jiffies - tfw_apm_jtmwindow;
+	for (i = 0, i_rlsz = 0; i < rlsz; ++i, --imax) {
+		rbent = &rbuf->rbent[imax % rbuf->rbentsz];
+		if (rbent->jtstamp && (rbent->jtstamp >= jtmwstart))
+			rnglst[i_rlsz++] = &rbent->pcntrng;
+	}
+
+	return i_rlsz;
+}
```
krizhanovsky: The function is required to track periods of inactivity only and I believe we can just remove it. Period of inactivity is technically just zeroed range. So lets just zero previous range on timer. The only timer should do statistic calculation, the reasonable value is `HZ / 20`. Since the timer runs more frequently than ranges are changed it shall track the last statistic update timestamp and if it determines that previous range wasn't updated due to period if inactivity, the range must be zeroed (vectorized `memset()` will do this quickly). Only after the check and zeroing the timer function should do statistics recalculation.

tempesta_fw/apm.c
```diff
@@ -0,0 +1,734 @@
...
+	for (i = 0; i < plsz; ++i) {
+		pval[i] = total_cnt * pcntl[i].ith / 100;
+		if (!pval[i])
+			pcntl[p++].val = 0;
+	}
+	for (i = 0; i < pcsz; ++i) {
+		TfwPcntCalc *pc = pcalc + i;
+		cnt += atomic_read(&rnglst[pc->intrvl]->cnt[pc->range][pc->bucket]);
+		for ( ; p < plsz && pval[p] <= cnt; ++p)
+			pcntl[p].val = pc->rtmval;
```
krizhanovsky: The only one thing where the algorithm differs from the unit test is ring buffer and essentially different ranges in its items. So basically we can just add one extra loop over the ring buffer (note that I proposed to have zero ranges) to `tfw_stat_calc()` from the unit test. Lets introduce an array `ptrs` of size equal to size of the ring buffer. The array must contain items like

         struct {
                   int r; // current i'th rb entry range
                   int b; // current i'th rb entry bucket
         }

The array is initialized by zeros. On first iteration we search minimal response time for 0th bucket or 0th range over all ring buffer (rb) entries and use the bucket for current `pcnts` calculation (from original unit test `tfw_stat_calc()`). `b` and probably `r` for current rb entry is moved forward. We repeat the search of the minimal bucket at second iteration, but we consider the next bucket of just used rb entry. The process repeats using only minimal buckets. That's long process, but it shorter, than current merging. We can thik about the algorithm as generalization of merge sort.

tempesta_fw/apm.c
```diff
@@ -0,0 +1,734 @@
...
+		kfree(pcalc);
+
+	return 0;
+}
+
+int
+tfw_apm_calc(void *data, Percentile *pcntl, size_t plsz, TfwPool *pool)
+{
+	BUG_ON(!data);
+	return __tfw_apm_calc((TfwApmData *)data, pcntl, plsz, pool);
```
krizhanovsky: With [schedulers](https://github.com/tempesta-tech/tempesta/issues/565) in mind we expect high contention on statistic recalculation. So lets calculate statistic in a `HZ / 20` timer only. The timer function must save calculated statistic in one of two static buffers while `tfw_apm_calc()` will use the other buffer meantime. Since the first timer runs the timer function will find most ring buffer items zero but the current one, it still can calculate current statistics.

krizhanovsky:
Also please update the unit test with "bad" cases which the algorithm does poor calculations on.

tempesta_fw/procfs.c
```diff
@@ -151,18 +151,22 @@ tfw_srvstats_seq_show(struct seq_file *seq, void *off)
...
 
-	tfw_apm_calc_user(srv->apm, pcntl, ARRAY_SIZE(pcntl));
+	tfw_apm_stats(srv->apm, prcntl, ARRAY_SIZE(prcntl));
 
-	SPRNE("Minimal response time\t\t", pcntl[0].val);
-	SPRNE("Maximum response time\t\t", pcntl[5].val);
-	SPRNE("Median  response time\t\t", pcntl[2].val);
+	SPRNE("Minimal response time\t\t", prcntl[0].val);
+	SPRNE("Maximum response time\t\t", prcntl[5].val);
+	SPRNE("Median  response time\t\t", prcntl[2].val);
```
milabs: Maybe it'll be useful to swap those lines to show stat in `Min. Median Max.` order ?

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1039 @@
...
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ * Prototype for fast precentiles calculation.
+ */
+#include <linux/kernel.h>
+#include <linux/slab.h>
+#include <linux/atomic.h>
+#include <linux/spinlock.h>
+#include <linux/stringify.h>
+#include <linux/sort.h>
```
krizhanovsky: It seems `sort.h` isn't needed any more. Also please keep includes in alphabetical order - it's always easier to find an include in ordered sequence.

tempesta_fw/t/unit/user_space/percentiles.c
```diff
@@ -381,7 +382,7 @@ tfw_stat_calc(TfwPcntRanges *rng, Percentilie *pcnts, size_t np, bool clear)
 		for (b = 0; b < TFW_STAT_BCKTS; ++b) {
 			cnt += atomic_read(&rng->cnt[r][b]);
 			for ( ; p < np && pval[p] <= cnt; ++p) {
-				pcnts[p].ith = cnt * 100 / tot_cnt;
+//				pcnts[p].ith = cnt * 100 / tot_cnt;
```
krizhanovsky: Debugging code?

tempesta_fw/t/unit/user_space/percentiles.c
```diff
@@ -300,10 +300,11 @@ tfw_stat_adjust(TfwPcntRanges *rng, int r)
 	 * If servers are too fast (all responses within 1ms),
 	 * then there is nothing to do for us.
 	 */
-	pc.atomic = rng->ctl[r].atomic;
-	if (unlikely(!pc.order))
-		goto out;
-	__range_shrink_left(rng, &pc, r);
+	if (r) {
```
krizhanovsky: Good catch! Let's just `goto out` instead of the two `if (r)`.

tempesta_fw/t/unit/user_space/percentiles.c
```diff
@@ -264,7 +264,7 @@ tfw_stat_adjust(TfwPcntRanges *rng, int r)
 		goto out;
 
 	printf("  -- range %d has outlier %lu (avg=%lu total=%lu) at"
-	       " bucket %lu\n", r, max, sum / cnt, cnt, i_max);
+	       " bucket %lu\n", r, max, sum / cnt, sum, i_max);
```
krizhanovsky: (not for this change, but for the test) There is still no bad test cases? Or did you fix the algorithm by the fix below?

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1039 @@
...
+static const TfwPcntCtl __read_mostly tfw_rngctl_init[TFW_STATS_RANGES] = {
+	{{0, 1, 16}},
+	{{1, 17, 47}},
+	{{2, 48, 108}},
+	{{4, 109, 349}}
+};
+
+/* A superset of percentiles for all users. */
+const unsigned int __read_mostly tfw_apm_prcntl_ith[] = {
+	1, 50, 75, 90, 95, 99
```
krizhanovsky: Oh, missed the line previous time. I used 1st percentilie for test only. In real life I don't thing that anybody is interested in 1st percentile :)

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1039 @@
...
+ * @i	- The current sequential bucket number across all ranges.
+ * @r	- The current range number.
+ * @b	- The current bucket number.
+ */
+typedef struct {
+	u16	v;
+	u8	i;
+	u8	r : 4;
+	u8	b : 4;
+} __attribute__((packed)) TfwApmRBEState;
```
krizhanovsky: You're renamed `upd` to `update` but still use very 'telling names' like `v`, `i`, `r` and `b` :)

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1039 @@
...
+{
+	int ret, rdidx;
+	TfwApmData *data = apmdata;
+	TfwApmPrcntl *prcntl;
+
+	BUG_ON(!apmdata);
+
+	rdidx = (unsigned int)atomic_read(&data->stats.rdidx) % 2;
+	prcntl = &data->stats.prcntl[rdidx];
+	if (!read_trylock(&prcntl->rwlock)) {
```
krizhanovsky: Why can't we just call `read_lock()`?

tempesta_fw/t/unit/user_space/percentiles.c
krizhanovsky: Just tried to start Tempesta with Apache upstream with default CentOS configuration and got following results for one request from Chrome browser:

        [root@localhost tempesta]# cat /proc/tempesta/servers/127.0.0.1\:8080 
        Minimal response time		: 0ms
        Median  response time		: 0ms
        Maximum response time		: 0ms
        Percentiles
        50%:	3ms
        75%:	0ms
        90%:	0ms
        95%:	0ms
        [root@localhost tempesta]# cat /proc/tempesta/perfstat 
        SS pfl hits				: 6
        SS pfl misses				: 12
        Cache hits				: 0
        Cache misses				: 0
        Client messages received		: 6
        Client messages forwarded		: 6
        Client messages served from cache	: 0
        Client messages parsing errors		: 0
        Client messages filtered out		: 0
        Client messages other errors		: 0
        Client connection attempts		: 4
        Client established connections		: 4
        Client connections active		: 4
        Client RX bytes				: 2983
        Server messages received		: 6
        Server messages forwarded		: 6
        Server messages parsing errors		: 0
        Server messages filtered out		: 0
        Server messages other errors		: 0
        Server connection attempts		: 8
        Server established connections		: 8
        Server connections active		: 4
        Server RX bytes				: 13545
        [root@localhost tempesta]# cat etc/tempesta_fw.conf 
        listen 172.16.0.5:80;
        server 127.0.0.1:8080;
        
        cache 1;
        
        sticky name=__TFW__ enforce;
        sticky_secret "f00)9eR59*_/22";
        sess_lifetime 90;
        
        frang_limits {
                request_rate 20;
                request_burst 15;
                connection_rate 8;
                connection_burst 6;
                concurrent_connections 8;
                client_header_timeout 20;
                client_body_timeout 10;
                http_uri_len 1024;
                http_field_len 256;
                http_ct_required false;
                http_methods get post head;
                http_ct_vals "text/plain" "text/html";
                http_header_chunk_cnt 10;
                http_body_chunk_cnt 0;
        	http_host_required 0;
        }

I.e. 50% of requests are less than 3ms. But what about the rest of 3 requests (performance statistic shows that there are 6 requests).

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1039 @@
...
+void *
+tfw_apm_create(void)
+{
+	TfwApmData *data;
+	TfwApmRBufEnt *rbent;
+	TfwApmRBLstEnt *rblstent;
+	Percentile *percentile[2];
+	int i, size;
+	int rbufsz = tfw_apm_tmwscale + 2;
+	int rblstmaxsz = tfw_apm_tmwscale + 1;
```
krizhanovsky: Please comment why do we need these extra entries.

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1039 @@
...
+}
+
+/*
+ * Initialize a ring buffer entry.
+ */
+static inline void
+tfw_apm_rbent_init(TfwApmRBufEnt *rbent, unsigned long jtstamp)
+{
+	memset(&rbent->pcntrng, 0, sizeof(TfwPcntRanges));
+	memcpy(rbent->pcntrng.ctl, tfw_rngctl_init, sizeof(rbent->pcntrng.ctl));
```
krizhanovsky: I think we shouldn't initialize the ranges by `tfw_rngctl_init`. We've learnt the appropriate ranges and typically they aren't changed from time to time. Meantime we loose accuracy when we adjust ranges. So it's better to improve accuracy when the system learn the ranges.

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1039 @@
...
+ * Set the flag to indicate if potentially new values were calculated.
+ */
+static int
+tfw_apm_calc(TfwApmData *data)
+{
+	int ret, wridx;
+	Percentile percentile[] = {
+		{tfw_apm_prcntl_ith[0]}, {tfw_apm_prcntl_ith[1]},
+		{tfw_apm_prcntl_ith[2]}, {tfw_apm_prcntl_ith[3]},
+		{tfw_apm_prcntl_ith[4]}, {tfw_apm_prcntl_ith[5]},
```
krizhanovsky: It'd be better to use `memcpy()` - the initialization looks ugly and is unsupportable.

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1039 @@
...
+ * Return 1 if the percentile values didn't need the recalculation.
+ */
+static int
+__tfw_apm_calc(TfwApmData *data, TfwApmPrcntl *prcntl)
+{
+	TfwApmRBLst *rblst = &data->rblst;
+	int i, rblstsz = rblst->rblstsz;
+
+	if (tfw_apm_rblst_update(data))
+		return 1;
```
krizhanovsky: I believe we don't need `TfwApmRBLst` at all and can make all the calculations over the initial ring buffer, so the function should be eliminated.

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1041 @@
...
+tfw_apm_calc(TfwApmData *data)
+{
+	int nfilled, wridx, recalc;
+	PrcntlStats pstats[ARRAY_SIZE(tfw_apm_pstats)];
+	TfwApmPrcntl wrkprcntl = { pstats, ARRAY_SIZE(pstats) };
+	TfwApmPrcntl *prcntl;
+
+	memcpy(pstats, tfw_apm_pstats, sizeof(tfw_apm_pstats));
+
+	smp_mb__before_atomic();
```
krizhanovsky: Why compiler shouldn't move instructions around the point?

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1041 @@
...
+
+	memcpy(pstats[0], tfw_apm_pstats, sizeof(tfw_apm_pstats));
+	memcpy(pstats[1], tfw_apm_pstats, sizeof(tfw_apm_pstats));
+
+	rwlock_init(&data->stats.prcntl[0].rwlock);
+	rwlock_init(&data->stats.prcntl[1].rwlock);
+	atomic_set(&data->stats.rdidx, 0);
+
+	set_bit(TFW_APM_DATA_F_REARM, &data->flags);
+	smp_mb__after_atomic();
```
krizhanovsky: `setup_timer()` is a function call from external module - i don't believe compiler can replace the `set_bit` operation with the function call.

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1041 @@
...
+ */
+static void
+tfw_apm_calc(TfwApmData *data)
+{
+	int nfilled, wridx, recalc;
+	PrcntlStats pstats[ARRAY_SIZE(tfw_apm_pstats)];
+	TfwApmPrcntl wrkprcntl = { pstats, ARRAY_SIZE(pstats) };
+	TfwApmPrcntl *prcntl;
+
+	memcpy(pstats, tfw_apm_pstats, sizeof(tfw_apm_pstats));
```
krizhanovsky: `tfw_apm_pstats` is static const array, nobody changes it. It seems has no sense to copy it just to copy the copy to `prcntl->pstats`

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1041 @@
...
+ * if potentially new percentile values were calculated.
+ * Return 0 if the percentile values didn't need the recalculation.
+ * REturn -1 of the recalculation could not be performed.
+ */
+static int
+__tfw_apm_calc(TfwApmData *data, TfwApmPrcntl *prcntl, int recalc)
+{
+	int ret;
+
+	if (!spin_trylock(&tfw_sa_guard)) {
```
krizhanovsky: You update per server APM data and there is expected to have tens of the the servers, so using global lock is bad. Please move the lock to per server data structure.

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1041 @@
...
+
+/*
+ * APM Data structure.
+ * Note that the organization of the supporting data heavily depends
+ * on the fact that there's only one party that does the calculation
+ * of percentiles - the function that runs periodically on timer.
+ * If there are several different parties that do the calculation,
+ * then the data may need to be organized differently.
+ * @rbuf	- The ring buffer for the specified time window.
+ * @rblst	- The list of ring buffer entries used in the calculation.
```
krizhanovsky: The entry was removed.

tempesta_fw/apm.c
keshonok: This lock is needed to prevent range changes while the percentiles are calculated. The range changes may occur only for the current ring buffer entry, and that ring buffer entry is the same for all servers. I don't see any practical reason for having a lock per server data structure.

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1041 @@
...
+ * Readers read the latest percentiles at @prcntl[@rdidx % 2]. The writer
+ * writes the new percentiles to @prcntl[(@rdidx + 1) % 2], and then
+ * increments @rdidx. The reading and writing is protected by a rwlock.
+ * @prcntl	- The percentiles for reading/writing (flip-flop manner).
+ * @rdidx	- The current index in @prcntl for readers.
+ */
+typedef struct {
+	TfwApmPrcntl	prcntl[2];
+	atomic_t	rdidx;
+} TfwApmStats;
```
krizhanovsky: We still have to use rw lock, so it seems there is no sense to have the 2-items array with current pointer - it doesn't help anyhow,

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1041 @@
...
+ *    are unknown apriori;
+ *
+ * 5. The adjustments of buckets must be performed in a lock-less fashion
+ *    in a multi-core environment;
+ *
+ * 6. If a user asks for Nth percentile, e.g. 75th, an inaccurate value
+ *    may be returned that is in fact for a different percentile, e.g.
+ *    81st. That is very possible if there is insufficient data for an
+ *    accurate percentiles calculation.
+ *
```
krizhanovsky: extra empty line

tempesta_fw/apm.c
keshonok: In my understanding, this is not a compiler barrier on an SMP system, but a memory barrier for synchronization between CPUs. `atomic_read()` doesn't imply memory barriers. I would want to be absolutely sure that this is the entry for reading the latest calculated percentiles, as the index may be incremented on another CPU.

On the other hand, there's also `test_and_clear_bit()` here that does imply a memory barrier. Perhaps, we can put that statement first and get the memory barrier for `atomic_read()` for free? That's a bit non-obvious though.

**ADDED:** Well, as it turns out `rdidx` is not needed that much, then this barrier is automatically not needed as well.

tempesta_fw/procfs.c
```diff
@@ -137,6 +151,145 @@ tfw_perfstat_seq_open(struct inode *inode, struct file *file)
 	return single_open(file, tfw_perfstat_seq_show, PDE_DATA(inode));
 }
 
+/*
+ * Individual server statistics. Note that 1%, 50%, and 99% percentiles
+ * are used to tell the minimum, the median, and the maximum values
+ * correspondingly.
+ */
+static const PrcntlStats __read_mostly tfw_procfs_pstats[] = {
+	{1}, {50}, {75}, {90}, {95}, {99}
```
krizhanovsky: It's just a copy & paste from `tfw_apm_pstats`. Also please remove `{1}` from both of the arrays.

keshonok: It's explained in the comment what the {1} (and {99} as well) percentiles are used for.

tempesta_fw/apm.c
```diff
@@ -0,0 +1,1041 @@
...
+	kfree(data);
+}
+
+/*
+ * Initialize a ring buffer entry.
+ */
+static inline void
+tfw_apm_rbent_init(TfwApmRBEnt *rbent, unsigned long jtmistamp)
+{
+	memcpy(rbent->pcntrng.ctl, tfw_rngctl_init, sizeof(rbent->pcntrng.ctl));
```
krizhanovsky: It seems comment https://github.com/tempesta-tech/tempesta/pull/601#discussion_r76244964 is still not addressed: I think we should keep already learnt ranges...

keshonok: This function is only used when a ring buffer for a server is created and its entries are initialized. When everything is working, an entry is resent using `__tfw_apm_rbent_reset()` function.

tempesta_fw/apm.c
krizhanovsky: Ok, agreed. Just noticed that the barrier is defined in arch/x86/include/asm/barrier.h as 

        #define smp_mb__before_atomic() barrier()

and that's just 

        #define barrier() __asm__ __volatile__("": : :"memory")

in include/linux/compiler-gcc.h

tempesta_fw/procfs.c
krizhanovsky: No, the comment is wrong. Minimum and maximum have nothing with percentilies. Consider sequence

        0 100 101 102 103 ... 297 1000      (200 numbers in total)

minimum and maximums are 0 and 1000 correspondingly, while 1st percentile is 100 and 99th is 297 correspondingly  much different numbers.

tempesta_fw/apm.c
krizhanovsky: Oh, my failure - I forgot that the lock contention happens only for slow readers, which isn't likely. So please just describe the locking solution in the comment above.

tempesta_fw/apm.c
```diff
@@ -299,47 +326,51 @@ tfw_stats_adjust(TfwPcntRanges *rng, int r)
...
-tfw_stats_update(TfwPcntRanges *rng, unsigned int r_time)
+tfw_stats_update(TfwPcntRanges *rng, unsigned int r_time, spinlock_t *slock)
 {
 	TfwPcntCtl pc3, pc2 = { .atomic = rng->ctl[2].atomic };
 
+	/* Adjust min/max values. */
+	if (!tfw_stats_adj_min(rng, r_time))
+		tfw_stats_adj_max(rng, r_time);
+	/* Add to @tot_val for AVG calculation. */
+	atomic64_add(r_time, &rng->tot_val);
```
krizhanovsky: That's dangerous. While we have 64 bits, each small delta-t there are many requests on the fly processed at many CPU cores at the same time, so each small deltat-t we're adding relatively large values (request processing latency) to the counter on all the CPU cores.

Rather we should use formula:

       avg = ((avg * tot_n) + r_time) / ++tot_n

, where `tot_n` is total number of requests. Since we're working in multi core environment we should calculate the statistic per cpu to avoid races in counters. I think `min` and `max` also are good to be calculated per cpu to about the loops over atomic caunters. Thus, procfs interface can show average in following way:

        avg = (cpu0->avg * cpu0->tot_n + ... + cpuN->avg * cpuN->tot_n) / (cpu0->tot_n + ... + cpuN->tot_n)

Calculation of total `min` and `max` is trivial.

krizhanovsky:
Ok, looks good for me.

PR #603, opened 2016-08-18
tls: add the implementation (v1)

milabs:
This adds stable TLS support based on mbedTLS library.

TODO:
- [ ] FSM for TLS

### DIFF
--- a/tempesta_fw/connection.c
+++ b/tempesta_fw/connection.c
@@ -97,20 +97,7 @@ tfw_connection_recv(void *cdata, struct sk_buff *skb, unsigned int off)
 {
 	TfwConnection *conn = cdata;
 
-	if (!conn->msg) {
-		conn->msg = TFW_CONN_HOOK_CALL(conn, conn_msg_alloc);
-		if (!conn->msg) {
-			__kfree_skb(skb);
-			return -ENOMEM;
-		}
-		TFW_DBG("Link new msg %p with connection %p\n",
-			conn->msg, conn);
-	}
-
-	TFW_DBG("Add skb %p to message %p\n", skb, conn->msg);
-	ss_skb_queue_tail(&conn->msg->skb_list, skb);
-
-	return tfw_gfsm_dispatch(&conn->msg->state, conn, skb, off);
+	return tfw_gfsm_dispatch(&conn->state, conn, skb, off);
 }
 
 void
--- a/tempesta_fw/connection.c
+++ b/tempesta_fw/connection.c
@@ -89,7 +89,7 @@ tfw_connection_release(TfwConnection *conn)
 int
 tfw_connection_send(TfwConnection *conn, TfwMsg *msg)
 {
-	return ss_send(conn->sk, &msg->skb_list, msg->ss_flags);
+	return TFW_CONN_HOOK_CALL(conn, conn_send, msg);
 }
 
 int
--- a/tempesta_fw/connection.c
+++ b/tempesta_fw/connection.c
@@ -25,12 +25,7 @@
 #include "log.h"
 #include "sync_socket.h"
 
-#define TFW_CONN_MAX_PROTOS	TFW_GFSM_FSM_N
-
-static TfwConnHooks *conn_hooks[TFW_CONN_MAX_PROTOS];
-
-#define TFW_CONN_HOOK_CALL(conn, hook_name) \
-	conn_hooks[TFW_CONN_TYPE2IDX(TFW_CONN_TYPE(conn))]->hook_name(conn)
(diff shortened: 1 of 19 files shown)


tempesta_fw/connection.c
```diff
@@ -94,28 +89,15 @@ tfw_connection_release(TfwConnection *conn)
...
-		}
-		TFW_DBG("Link new msg %p with connection %p\n",
-			conn->msg, conn);
-	}
-
-	TFW_DBG("Add skb %p to message %p\n", skb, conn->msg);
-	ss_skb_queue_tail(&conn->msg->skb_list, skb);
-
-	return tfw_gfsm_dispatch(&conn->msg->state, conn, skb, off);
+	return tfw_gfsm_dispatch(&conn->state, conn, skb, off);
```
krizhanovsky: Yeah, looks like a very good solution: there is definitely connection state which we didn't have before and which differs from messages state. Also `conn_msg_alloc` seems somewhat unnatural for connectlon layer. So the hook should be removed - now we call `tfw_http_conn_msg_alloc()` from http.c only.

tempesta_fw/http.c
```diff
@@ -463,6 +457,7 @@ tfw_http_conn_msg_free(TfwHttpMsg *hm)
 static int
 tfw_http_conn_init(TfwConnection *conn)
 {
+	tfw_gfsm_state_init(&conn->state, conn, TFW_HTTP_FSM_INIT);
```
krizhanovsky: Please adjust the comment above

tempesta_fw/http.c
```diff
@@ -513,6 +508,15 @@ tfw_http_conn_drop(TfwConnection *conn)
 }
 
 /**
+ *
```
krizhanovsky: Empty comment?

tempesta_fw/sock_clnt.c
```diff
@@ -41,11 +41,9 @@
  * TfwConnection extension for client sockets.
  *
  * @conn	- The base structure. Must be the first member.
- * @ka_timer	- The keep-alive timer for the connection.
  */
 typedef struct {
 	TfwConnection		conn;
-	struct timer_list	ka_timer;
 } TfwCliConnection;
```
krizhanovsky: So `TfwCliConnection` becomes single-member structure and there is no sense to pollute name space. It's better to remove the structure.

tempesta_fw/tls.c
```diff
@@ -19,73 +19,583 @@
...
+static struct kmem_cache *tfw_tls_cache = NULL;
+
+static inline int
+ttls_ssl_handshake(TfwTlsContext *tls)
+{
+	int r;
+
+	spin_lock(&tls->lock);
+	r = mbedtls_ssl_handshake(&tls->ssl);
+	spin_unlock(&tls->lock);
```
krizhanovsky: Bad point. We introduced work queue to make sockets CPU local. So that's bad to access TLS context from many CPUs and we should use similar mechanism to ensure that TLS context are per-CPU. Probably not in this patch set, but it's required to finish #81. So please write TODO comment here.

tempesta_fw/tls.c
```diff
@@ -19,73 +19,583 @@
...
+		nskb = alloc_skb(MAX_TCP_HEADER + MAX_TLS_PAYLOAD, GFP_ATOMIC);
+		if (unlikely(!nskb))
+			return TFW_BLOCK;
+
+		skb_reserve(nskb, MAX_TCP_HEADER);
+		skb_put(nskb, MAX_TLS_PAYLOAD);
+
+		r = ttls_ssl_read(tls, nskb->data, MAX_TLS_PAYLOAD);
+		if (r == MBEDTLS_ERR_SSL_WANT_READ ||
+		    r == MBEDTLS_ERR_SSL_WANT_WRITE) {
```
krizhanovsky: Please place `{` at new line for multi-line IF statements - that's more readable.

tempesta_fw/tls.c
```diff
@@ -19,73 +19,583 @@
...
+static int
+tfw_tls_conn_init(TfwConnection *c)
+{
+	int r;
+	TfwTlsContext *tls;
+
+	tls_dbg(c, "=>");
+
+	BUG_ON(c->tls);
+	tls = c->tls = kmem_cache_alloc(tfw_tls_cache, GFP_ATOMIC);
```
krizhanovsky: Each TLS connection has `TfwTlsContext` as well as `TfwConnection`. So why to do tow SLAB calls? There is no reason to do so. Please make TLS connection inherited from TfwConnection or make more elegant (`tls` pointing to the end of function isn't such) way if you can invent such.

krizhanovsky:
good to merge

PR #607, opened 2016-08-25
http-parser: remove erroneous isdigit check when parsing day

### LINKED ISSUE #606: http_parser: parse_int_ws boundary issue
`parse_int_ws` is used incorrectly when integer parsing occurs on SKB boundaries:

skb(1) = `.... Date: Wed, 24`
skb(2) = ` Aug 2016 .........`

In that case, after parsing `24` in skb(1) the parser expects that the integer will continues in skb(2). It seems that the problem may reflect other places, not only the [Date](https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/http_parser.c#L2893) day's parsing.

milabs:
It's incorrect to check for digit before calling parse_int_ws() as there
may be delimiter at the start of chunk. Also, check for upper day limit.

Fixes #606.


krizhanovsky:
good to merge

PR #608, opened 2016-08-26
Add FSM for the TLS

milabs:


tls/platform.h
```diff
@@ -125,8 +125,8 @@ int mbedtls_platform_set_calloc_free( void * (*calloc_func)( size_t, size_t ),
                               void (*free_func)( void * ) );
 #endif /* MBEDTLS_PLATFORM_FREE_MACRO && MBEDTLS_PLATFORM_CALLOC_MACRO */
 #else /* !MBEDTLS_PLATFORM_MEMORY */
-#define mbedtls_free       free
-#define mbedtls_calloc     calloc
+#define mbedtls_free       ttls_free
+#define mbedtls_calloc     ttls_calloc
```
krizhanovsky: Why do we need the `ttls_*` calls? It seems we can just define them as `kfree()` and `kzalloc()` correspondingly.

tls/Makefile
```diff
@@ -24,7 +24,13 @@ EXTRA_CFLAGS += -Wframe-larger-than=4096
...
 
-ttls-srcs = $(wildcard $(obj)/*.c)
+ttls-srcs = $(wildcard $(src)/*.c)
 ttls-objs = $(patsubst %.c, %.o, $(ttls-srcs))
 
+# exclude some original files (see tempesta_fw/ for re-implementation)
+ttls-objs := $(filter-out	\
+	$(src)/ssl_tls.o	\
+	$(src)/ssl_srv.o	\
+	, $(ttls-objs))
```
krizhanovsky: No, that's just wrong! Firstly, we should not keep the same source code in `tls` and `tempesta_fw` at the same time. Secondly, all the code in `tempesta_fw` must obey to our coding style. Next, you should have been accurately split the logic between `tls` and `tempesta_fw`, not just copy & paste source code.

krizhanovsky:
I close the poorly designed PR in favour to the new implementation.

PR #617, opened 2016-09-29
Fix a bug with error handling when calculating sticky cookie value.

keshonok:
If the calculation of the sticky cookie value resulted in an error,
then the 502 error response was sent to the client, and the result
of zero was returned to tfw_http_req_cache_cb(). Then the request
was processed further as if nothing had happened. Potentially, the
request was sent to a server, and then another response to the same
request was sent to the client.

Also, it's important that with this change the sticky cookie module
doesn't send any error responses any more. That's the prerogative of
the code in http.c.

Also, a small refactoring and better comments in a couple of places.

### DIFF
--- a/tempesta_fw/http_sess.c
+++ b/tempesta_fw/http_sess.c
@@ -145,19 +145,17 @@ search_cookie(TfwPool *pool, const TfwStr *cookie, TfwStr *val)
 	/* Search cookie name. */
 	end = (TfwStr*)cookie->ptr + TFW_STR_CHUNKN(cookie);
 	for (chunk = cookie->ptr; chunk != end; ++chunk, --n) {
-		if (chunk->flags & TFW_STR_NAME) {
-			/*
-			 * Create temporary compound string, starting
-			 * with this chunk.
-			 * We do not use it's overall length now,
-			 * so do not set it.
-			 */
-			tmp.ptr = (void *)chunk;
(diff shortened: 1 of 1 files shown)


tempesta_fw/http_sess.c
```diff
@@ -361,7 +359,7 @@ tfw_http_sticky_notfound(TfwHttpReq *req)
 
 	/* Create Tempesta sticky cookie and store it */
 	if (tfw_http_sticky_calc(req, &sv) != 0)
-		return tfw_http_send_502(req);
+		return -1;
```
krizhanovsky: Good catch and the fix eliminates the problem.

However now `tfw_http_req_cache_cb()` sends 500 error code which is generic code since the function doesn't know exact reason for the error. Meantime previously we sent 502nd error code. 502 error code seems doubtful in this situation since we failed on sticky cookie calculation and no upstream communication was involved, so I think we shouldn't return "Bad gateway code". Now behaviour seems correct.

The design problem is that there are different subsystems (Web-cache, load balancing schedulers etc.) which can return with errors and we have to send correct error codes to clients. One way to keep design correct is to return negative HTTP error codes (i.e. `-500` in this case) such that upper layer can send correct HTTP error messages. The second way (seems what we have now) is to just make decision which error code to send at `http.c` layer depending on which subsystem fails.

PR #620, opened 2016-10-10
HTTP parser fixes and few unit test additions and cleanups

krizhanovsky:


tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -146,37 +146,40 @@ do_split_and_parse(unsigned char *str, int type)
...
 
 TEST(http_parser, parses_req_method)
 {
-	FOR_REQ("GET / HTTP/1.1\r\n\r\n")
+	FOR_REQ("GET / HTTP/1.1\r\n\r\n") {
 		EXPECT_EQ(req->method, TFW_HTTP_METH_GET);
+	}
 
-	FOR_REQ("HEAD / HTTP/1.1\r\n\r\n")
+	FOR_REQ("HEAD / HTTP/1.1\r\n\r\n") {
```
keshonok: Why do we need curly brackets here and in similar statement below? All macros like `EXPECT_EQ()` are implemented correctly with `do {} while (0)` where necessary. In all practical terms, it's a single statement that doesn't require brackets.

tempesta_fw/http_msg.c
```diff
@@ -322,8 +266,8 @@ tfw_http_msg_hdr_close(TfwHttpMsg *hm, int id)
 	*h = hm->parser.hdr;
 
 	TFW_STR_INIT(&hm->parser.hdr);
-	TFW_DBG3("store header w/ ptr=%p len=%lu flags=%x id=%d\n",
-		 h->ptr, h->len, h->flags, id);
+	TFW_DBG3("store header w/ ptr=%p len=%lu eolen=%u flags=%x id=%d\n",
+		 h->ptr, h->len, h->eolen, h->flags, id);
 
 	/* Move the offset forward if current header is fully read. */
 	if (id == ht->off)
```
ikoveshnikov: Clang static analyser:
273:9: warning: comparison of integers of different signs: int and unsigned int (-Wsign-compare)

That is not related to patch itself and, warning was here before (and in many places more) and seems not being harmful. May be we should create separated issue with static analyser issues?

krizhanovsky: Yes, the issues by results of static analyzers are appreciated. We use [Coverity](https://scan.coverity.com/projects/tempesta-tech-tempesta), but we didn't upload our source code for a while, so making the new verification has sense.

Basically we don't use Clang for kernel source code, but we still can use its static analyzing. I leave the code as is for now since I have no Clang and never used it before. However, it would be very useful if you upload the new code to Coverity and create a task for all Clang warnings.

PR #621, opened 2016-10-25
HttpMsg: fix raw header removal

ikoveshnikov:
Fix bug: removing RAW http header from `TfwHttpMsg` using `tfw_http_msg_hdr_xfrm()` function call cause memory corruption.


krizhanovsky:
Good catch! Also good for me, let's merge it.

PR #625, opened 2016-10-27
AVX2 HTTP strings processing & multiple HTTP parser fixes

krizhanovsky:
Also see stings benchmark https://github.com/natsys/blog/tree/master/fast_str .

Please note that AVX2 is required to test the feature, so if your host system has at least Haswell CPU, then use `-cpu host` key for you qemu VM to pass all CPU options to the guest. While the code is written to be compilable w/o AVX2, I found that if I try to run Tempesta FW in guest w/o `-cpu host` it crashes at first XMM register load.

I tested the code with `test_tfw_str` and `test_http_parser` unit tests as well as with manual requests and some simple traffic generators.

Please review this ASAP since we need to do performance testing ASAP.


tempesta_fw/htype.h
```diff
@@ -0,0 +1,113 @@
...
+ * this program; if not, write to the Free Software Foundation, Inc., 59
+ * Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ */
+#ifndef __HTYPE_H__
+#define __HTYPE_H__
+
+/**
+ * ASCII codes to accept HTTP token (RFC 7230 3.2.6).
+ */
+static const unsigned char token_a[] __attribute__((aligned(64))) = {
```
krizhanovsky: Yes, the copies are desired. Since it's declared as `static` each translation unit will work with it's own copy and there won't be warnings. The data structure is just 4 cache lines, moreover only 2 of them are frequently used. Meantime, TFW_LC() is very performance critical call. When you make `static const` declaration compiler can do very good optimization work efficiently inlining constants to the code. If you define the constants table in other translation unit, then you lose the optimization opportunity.

Issue #626, opened 2016-10-27
BUG at /root/tempesta/tempesta_fw/http_sess.c:485


krizhanovsky:
Steps to reproduce:
1. deply [tempesta-tech.com](https://github.com/tempesta-tech/tempesta-tech.com) at Apache running in VM, the server must listen on 127.0.0.1:8080;
2. deply Tempesta FW (current master) on the same VM with following configuration
   
   ```
   listen 172.16.0.5:80;
   server 127.0.0.1:8080;
   cache 1;
   cache_fulfill * *;
   sticky name=__TFW__ enforce;
   sticky_secret "f00)9eR59*_/22";
   sess_lifetime 90;
   ```
3. send two requests, one by one, using:
   
   ```
   wget http://172.16.0.5/img/ico/favicon.png
   ```
4. enjoy the crash:
   
   ```
   kernel BUG at /root/tempesta/tempesta_fw/http_sess.c:485!
   invalid opcode: 0000 [#1] SMP
   Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) tempesta_tls(PO) nf_conntrack_ipv4 nf_defrag_ipv4 xt_conntrack nf_conntrack ipt_REJECT nf_reject_ipv4 xt_CHECKSUM iptable_mangle bridge stp llc ip6table_filter ip6_tables iptable_filter ip_tables ebtable_nat ebtables crc32c_intel aesni_intel aes_x86_64 glue_helper xfs libcrc32c exportfs lrw gf128mul ablk_helper cryptd ppdev i2c_piix4 microcode i2c_core pcspkr serio_raw parport_pc parport acpi_cpufreq dm_mirror dm_region_hash dm_log dm_mod uinput btrfs xor raid6_pq ata_generic pata_acpi e1000 ata_piix floppy ipv6 autofs4
   CPU: 0 PID: 3 Comm: ksoftirqd/0 Tainted: P           O    4.1.27+ #1
   Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.9.1-1.fc24 04/01/2014
   task: ffff88005c82a010 ti: ffff88005c88c000 task.ti: ffff88005c88c000
   RIP: 0010:[<ffffffffa060a8b3>]  [<ffffffffa060a8b3>] tfw_http_sess_resp_process+0x43/0x50 [tempesta_fw]
[...]

PR #627, opened 2016-10-28
Obtain session information before serving from cache

ikoveshnikov:
Req->sess must be set before adjusting response

Fix issue #626


krizhanovsky:
Looks good. Please address the comment https://github.com/tempesta-tech/tempesta/issues/626#issuecomment-256946011 as well.

PR #632, opened 2016-11-03
Performance optimization

krizhanovsky:
Do not queue cache work for local CPU: 25% better performance on single connection and now we have almost linear scalability on 2K-32K connections; also remove old debug prints

### DIFF
--- a/tempesta_fw/cache.c
+++ b/tempesta_fw/cache.c
@@ -1088,6 +1088,15 @@ tfw_cache_build_resp_body(TDB *db, TfwHttpResp *resp, TdbVRec *trec,
  * (copy the list of skbs is faster than scan TDB and build TfwHttpResp).
  * TLS should encrypt the data in already prepared skbs.
  *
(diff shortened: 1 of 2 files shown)


tempesta_fw/cache.c
```diff
@@ -1210,56 +1204,81 @@ cache_req_process_node(TfwHttpReq *req, unsigned long key,
...
-	TfwCacheEntry *ce = NULL;
+	if (resp) {
+		tfw_cache_add(resp, req, action);
+	}
+	else if (req->method == TFW_HTTP_METH_PURGE) {
+		tfw_cache_purge_method(req);
+	}
+	else {
+		cache_req_process_node(req, action);
+	}
```
krizhanovsky: Usually we use curly braces in comples if-else statements: https://github.com/tempesta-tech/linux-4.1.27-tfw/blob/master/Documentation/CodingStyle#L141

Issue #409, opened 2016-01-30
[Cache] Don't store hop-by-hop headers


krizhanovsky:
Keep-Alive and Server are hop-by-hop headers and should not be cached. `hbh_hdrs` must be adjusted and the headers must be made special to be analyzed quickly in `tfw_cache_copy_resp()`. 

Relates to #407.

krizhanovsky:
See also [What Proxies Must Do](https://www.mnot.net/blog/2011/07/11/what_proxies_must_do) and particularly the referenced [spec](http://tools.ietf.org/html/draft-ietf-httpbis-p1-messaging-14#section-7.1.3)

ikoveshnikov:
Actual problem differs from mentioned above: hop-by-hop headers are forwarded as end-to-end headers

In [spec mentioned above](https://tools.ietf.org/html/draft-ietf-httpbis-p1-messaging-14#section-7.1.3) 8 headers are marked as a hop-by-hop. But the [final document RFC7230](https://tools.ietf.org/html/rfc7230#section-6.1) is more strict than a draft and close to [article What Proxies Must Do](https://www.mnot.net/blog/2011/07/11/what_proxies_must_do):
```
   A proxy or gateway MUST parse a received Connection header field before
   a message is forwarded and, for each connection-option in this field,
   remove any header field(s) from the message with the same name as the
   connection-option, and then remove the Connection header field itself
   (or replace it with the intermediary's own connection options for the
   forwarded message).
```
So, for request:
```
    GET /foo HTTP/1.1
    Host: example.net
    Connection: Keep-Alive, Foo, Bar
    Foo: abc
    Foo: def
    Keep-Alive: timeout=30
```
The Connection, Keep-Alive and both Foo headers must be removed before forwarding it.
[Co-Advisor HTTP 1.1 compatibility tests](http://coad.measurement-factory.com/cgi-bin/coad2/GraseInfoCgi?info_id=test_clause/rfc2616/hopHdr&proto=http)  follows obsoleted RFC2616, but also checks hop-by-hop headers removing from messages.

'tfw_http_set_hdr_connection()` seems to return without updating Connection header. Also other hop-by hop headers are not removed from test messages.

Used tempesta (`af32a834`) configuration:
```
server <test-server-ip>
cache_methods GET HEAD;
cache_fulfill * *;
```

krizhanovsky:
Please also store correct Server and Via headers in cache as described in #634

[...]

PR #633, opened 2016-11-04
Don't cache and don't forward hop-by-hop headers

ikoveshnikov:
Fix issue #409 

Main changes:
- Update http parser to save _Connection_ header string as list of header names tagged with flag `TFW_STR_VALUE`. Most common tokens `keep-alive` and `close` and also delimiters are not marked with this flag: first processed as special header as most common one, others don't point to any headers.  `TFW_HTTP_CONN_EXTRA` flag for whole message is set if _Connection_ contains more than just `keep-alive` or `close`
- `Keep-alive` header is set as special. That helps not to slow down performance by multiple string comparisons if `Keep-alive` is the only header set by _Connection_. Corresponding unit test was updated.
- Remove hop-by-hop headers when forwarding messages from client to server and vice-versa
- Don't cache hop-by-hop headers for server responses.

### DIFF
--- a/tempesta_fw/cache.c
+++ b/tempesta_fw/cache.c
@@ -131,27 +131,6 @@ enum {
 	TFW_CACHE_REPLICA,
 };
 
-/*
- * Non-cacheable hop-by-hop response headers in terms of RFC 2068.
- * The table is used if server doesn't specify Cache-Control no-cache
- * directive (RFC 7234 5.2.2.2) explicitly.
- *
- * Server header isn't defined as hop-by-hop by the RFC, but we don't show
- * protected server to world.
- *
- * We don't store the headers in cache and create then from scratch.
- * Adding a header is faster then modify it, so this speeds up headers
- * adjusting as well as saves cache storage.
- *
- * TODO process Cache-Control no-cache
- */
-static const int hbh_hdrs[] = {
-	[0 ... TFW_HTTP_HDR_RAW]	= 0,
-	[TFW_HTTP_HDR_SERVER]		= 1,
-	[TFW_HTTP_HDR_CONNECTION]	= 1,
-	[TFW_HTTP_HDR_KEEP_ALIVE]	= 1,
-};
-
 typedef struct {
 	int		cpu[NR_CPUS];
 	atomic_t	cpu_idx;
@@ -719,7 +698,7 @@ tfw_cache_copy_hdr(char **p, TdbVRec **trec, TfwStr *src, size_t *tot_len)
  */
 static int
 tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwHttpReq *req,
-		    size_t tot_len, const unsigned int * const hbh_hdrs_raw)
+		    size_t tot_len)
 {
 	long n;
 	char *p;
@@ -758,10 +737,7 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwHttpReq *req,
 	FOR_EACH_HDR_FIELD(field, end1, resp) {
 		n = field - resp->h_tbl->tbl;
 		/* Skip hop-by-hop headers. */
-		h = (n < TFW_HTTP_HDR_RAW && hbh_hdrs[n]) ? &empty : field;
-		if (n >= TFW_HTTP_HDR_RAW && hbh_hdrs_raw
-		    && hbh_hdrs_raw[n-TFW_HTTP_HDR_RAW])
-			h = &empty;
+		h = (field->flags & TFW_STR_HBH_HDR) ? &empty : field;
 		n = tfw_cache_copy_hdr(&p, &trec, h, &tot_len);
 		if (n < 0) {
 			TFW_ERR("Cache: cannot copy HTTP header\n");
@@ -802,8 +778,7 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwHttpReq *req,
 }
 
 static size_t
-__cache_entry_size(TfwHttpResp *resp, TfwHttpReq *req,
-		   const unsigned int * const hbh_hdrs_raw)
+__cache_entry_size(TfwHttpResp *resp, TfwHttpReq *req)
 {
 	long n;
 	size_t size = CE_BODY_SIZE;
@@ -813,16 +788,11 @@ __cache_entry_size(TfwHttpResp *resp, TfwHttpReq *req,
 	size += req->uri_path.len;
 	size += req->h_tbl->tbl[TFW_HTTP_HDR_HOST].len;
 
-
 	/* Add all the headers size */
 	FOR_EACH_HDR_FIELD(hdr, hdr_end, resp) {
 		/* Skip hop-by-hop headers. */
 		n = hdr - resp->h_tbl->tbl;
-
-		h = (n < TFW_HTTP_HDR_RAW && hbh_hdrs[n]) ? &empty : hdr;
-		if (n >= TFW_HTTP_HDR_RAW && hbh_hdrs_raw
-		    && hbh_hdrs_raw[n-TFW_HTTP_HDR_RAW])
-			h = &empty;
+		h = (hdr->flags & TFW_STR_HBH_HDR) ? &empty : hdr;
 
 		if (!TFW_STR_DUP(h)) {
 			size += sizeof(TfwCStr);
@@ -849,10 +819,10 @@ __cache_entry_size(TfwHttpResp *resp, TfwHttpReq *req,
 
 static void
 __cache_add_node(TDB *db, TfwHttpResp *resp, TfwHttpReq *req,
-		 unsigned long key, const unsigned int * const hbh_hdrs_raw)
+		 unsigned long key)
 {
 	TfwCacheEntry *ce, cdata = {{}};
-	size_t data_len = __cache_entry_size(resp, req, hbh_hdrs_raw);
+	size_t data_len = __cache_entry_size(resp, req);
 	size_t len = data_len;
 
 	/*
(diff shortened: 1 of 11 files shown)


tempesta_fw/http_msg.c
```diff
@@ -811,3 +815,105 @@ tfw_http_msg_alloc(int type)
...
+			*curr = colon;
+			hbh_hdr.len += c->len;
+			TFW_STR_CHUNKN_ADD(&hbh_hdr, 1);
+
+			id = __hdr_lookup(hm, &hbh_hdr);
+			if (id < hm->h_tbl->off) {
+				idxs[id - TFW_HTTP_HDR_RAW] = 1;
+			}
+		}
+	}
```
keshonok: This whole code block looks every complicated and has code duplication. I believe it should be structured in a different way. An outer loop is needed here. Inside that loop any unneeded chunks are skipped. When we step on a VALUE chunk then start collecting a token in a inner loop. After the inner loop has finished, then the colon is appended, the lookup is done, and the header is marked if found.

tempesta_fw/http_msg.c
```diff
@@ -811,3 +815,105 @@ tfw_http_msg_alloc(int type)
 
 	return hm;
 }
+
+/**
+ * Remove Hop-by-Hop headers in terms of RFC 7230 6.1
+ */
+int tfw_http_rm_hbh_hdrs(TfwHttpMsg *hm)
```
keshonok: As noted by @ikoveshnikov, these functions should be named `tfw_http_msg_*_hbh_hdrs()` as they reside in `http_msg.c` module.

tempesta_fw/http_msg.c
```diff
@@ -811,3 +815,105 @@ tfw_http_msg_alloc(int type)
...
+	r = tfw_http_find_hbh_hdrs(hm, hbh_heades);
+	if (unlikely(r))
+		goto out;
+
+	/* Removing must be from end to beginning */
+	for (i = hm->h_tbl->off - TFW_HTTP_HDR_RAW - 1; i >= 0; --i) {
+		if (hbh_heades[i]) {
+			r = __hdr_del(hm, i + TFW_HTTP_HDR_RAW);
+			if (unlikely(r))
+				goto out;
```
keshonok: Why not `break` from the loop?

tempesta_fw/http_parser.c
```diff
@@ -877,6 +895,12 @@ __FSM_STATE(RGen_OWS) {							\
 
 /**
  * Parse Connection header value, RFC 2616 14.10.
+ *
+ * Most likely that Connection header contains only "close" or "keep-alive"
+ * tokens. If Connection header contains other tokens, message flag
+ * TFW_HTTP_CONN_EXTRA is set to indicate that hop-by-hop headers are set
+ * and must be processed later. Each value will be placed to its chunk marked
+ * with TFW_STR_VALUE flag.
```
keshonok: Each chunk of a token other than those mentioned above is marked with `TFW_STR_VALUE` flag.

tempesta_fw/http_parser.c
```diff
@@ -907,19 +931,25 @@ __parse_connection(TfwHttpMsg *hm, unsigned char *data, size_t len)
 	 * it could be names of any headers, including custom headers.
 	 */
 	__FSM_STATE(I_ConnOther) {
-		__FSM_I_MATCH_MOVE(token, I_ConnOther);
+		msg->flags |= TFW_HTTP_CONN_EXTRA;
```
keshonok: This statement will be executed each time another chunk is seen in the input, even it it's just a single byte. Perhaps, an extra state is justified here that designates the start of a token.

tempesta_fw/cache.c
```diff
@@ -846,18 +858,29 @@ tfw_cache_add(TfwHttpResp *resp, TfwHttpReq *req, tfw_http_cache_cb_t action)
 	if (!tfw_cache_employ_resp(req, resp))
 		goto out;
 
+	if (resp->flags & TFW_HTTP_CONN_EXTRA) {
+		/* Fill hop-by-hop identifiers for RAW headers */
+		hbh_hds_raw = kcalloc(resp->h_tbl->off - TFW_HTTP_HDR_RAW,
+				      sizeof(unsigned int), GFP_ATOMIC);
```
ikoveshnikov: Using `TfwStr{}->flags` to indicate hbh header sounds like a great idea.

ikoveshnikov:
Changes: 
- Rebased to current master
- Hop-by-hop headers refactored: get rid of memory allocations and VLAs, simplify code, don't try to find hop-by-hop headers twice for same response (during caching and during adjusting)
- removed extra empty lines
- update function documentation

TODO:
- check unit tests (fail on master branch, not caused by the PR)

tempesta_fw/http_msg.c
```diff
@@ -819,3 +848,89 @@ tfw_http_msg_alloc(int type)
...
+			}
+		}
+	}
+	/*
+	 * Unset TFW_HTTP_CONN_EXTRA flag:
+	 * @tfw_http_msg_mark_hbh_hdrs may be called more than once for same
+	 * message. Since marks stay in http message until it would be
+	 * destroyed, we can reuse marks later without heavy processing.
+	 */
+	hm->flags &= ~((unsigned int)TFW_HTTP_CONN_EXTRA);
```
keshonok: This serves as an indicator that allows to NOT perform heavy marking again when it had been done already. However, why not avoid repeated calls to mark the headers altogether? That can be done by setting a flag once the marking is done.

tempesta_fw/http_msg.c
```diff
@@ -819,3 +848,89 @@ tfw_http_msg_alloc(int type)
...
+
+	/*
+	 *  Server header isn't defined as hop-by-hop by the RFC, but we
+	 * don't show protected server to world.
+	 */
+	__mark_hbh_header(&hm->h_tbl->tbl[TFW_HTTP_HDR_SERVER]);
+	__mark_hbh_header(&hm->h_tbl->tbl[TFW_HTTP_HDR_CONNECTION]);
+	__mark_hbh_header(&hm->h_tbl->tbl[TFW_HTTP_HDR_KEEP_ALIVE]);
+
+	if (!(hm->flags & TFW_HTTP_CONN_EXTRA))
```
keshonok: Please see the comment below regarding repeated calls to this function.

tempesta_fw/http_msg.c
```diff
@@ -819,3 +848,89 @@ tfw_http_msg_alloc(int type)
...
+EXPORT_SYMBOL(tfw_http_msg_mark_hbh_hdrs);
+
+/**
+ * Remove Hop-by-Hop headers in terms of RFC 7230 6.1
+ */
+int tfw_http_msg_rm_hbh_hdrs(TfwHttpMsg *hm)
+{
+	int r = 0, i;
+
+	tfw_http_msg_mark_hbh_hdrs(hm);
```
keshonok: As per above comment, this should be called only when the headers have not been marked yet.

tempesta_fw/http_msg.c
```diff
@@ -155,22 +156,50 @@ __hdr_is_singular(const TfwStr *hdr)
...
 	TfwHttpHdrTbl *ht = hm->h_tbl;
 
 	for (id = TFW_HTTP_HDR_RAW; id < ht->off; ++id) {
 		TfwStr *h = &ht->tbl[id];
 		/* There is no sense to compare against all duplicates. */
 		if (h->flags & TFW_STR_DUPLICATE)
 			h = TFW_STR_CHUNK(h, 0);
-		if (tfw_stricmpspn(hdr, h, ':'))
-			continue;
+		if (!tfw_stricmpspn(hdr, h, ':'))
```
keshonok: Frankly, this doesn't match the strings correctly. It may successfully match strings like `Keep-Alive:` and `Keep-Alive-Forever:`. Why not fix it and the enforce the correct matching?

tempesta_fw/http_msg.c
```diff
@@ -155,22 +156,50 @@ __hdr_is_singular(const TfwStr *hdr)
...
+		TfwStr *h = &ht->tbl[id];
+		/* Skip if header is already marked as hop-by-hop */
+		if (h->flags & TFW_STR_HBH_HDR)
+			continue;
+		/* There is no sense to compare against all duplicates. */
+		if (h->flags & TFW_STR_DUPLICATE)
+			h = TFW_STR_CHUNK(h, 0);
+		if (tfw_str_eq(name, h, name->len)
+		    && (':' == tfw_str_at_index(h, name->len)))
+			break;
```
keshonok: Please see the comment above. It looks to me that there's no need to introduce two new TwfStr{} strings function `tfw_str_eq()` and `tfw_str_at_index()`. At least not until that functionality is really needed.

The core of this function should be made a separate function that can be used by `__hdr_lookup()`. That separate function would return the `id` if it's found, and it can be used directly in place of this `__hdr_lookup_hdr_name()`.

tempesta_fw/http_msg.c
```diff
@@ -819,3 +848,89 @@ tfw_http_msg_alloc(int type)
...
+	__mark_hbh_header(&hm->h_tbl->tbl[TFW_HTTP_HDR_KEEP_ALIVE]);
+
+	if (!(hm->flags & TFW_HTTP_CONN_EXTRA))
+		return;
+
+	tfw_http_msg_srvhdr_val(&hm->h_tbl->tbl[TFW_HTTP_HDR_CONNECTION],
+				TFW_HTTP_HDR_CONNECTION, &conn_h);
+
+	TFW_STR_FOR_EACH_CHUNK(c, &conn_h, end) {
+		if (c->flags & TFW_STR_VALUE) {
```
keshonok: Deep nesting, more difficult to read and comprehend the logic.
```C
if (!(c->flags & TFW_STR_VALUE))
        continue;
```

tempesta_fw/str.c
```diff
@@ -318,6 +318,67 @@ tfw_stricmpspn(const TfwStr *s1, const TfwStr *s2, int stop)
 EXPORT_SYMBOL(tfw_stricmpspn);
 
 /**
+ * Generic function for comparing first @len symbols in TfwStrs.
+ * @s1,@s2 may be either plain or compound, but must be longer than @len symbols
+ * Do not use it for duplicate strings, rather call it for each duplicate
+ * substring separately.
+ *
+ * @return true if the strings are equal and false otherwise.
+ */
```
krizhanovsky: It seems both the string functions are unnecessary, see comments above: you can just create temporary compound TfwStr with ":" at the end chunk and use `tfw_stricmpspn()`.

ikoveshnikov:
redisigned in #650

Issue #630, opened 2016-11-02
[Cache] Sporadic cache transfers hangs


krizhanovsky:
The bug seems introduced by some recent changes, I just verified the scenario against tempesta-tech.com and the bug isn't reproduced.

Steps to reproduce:
1. deply [tempesta-tech.com](https://github.com/tempesta-tech/tempesta-tech.com) at Apache running in VM, the server must listen on 127.0.0.1:8080;
2. deply Tempesta FW (current master) on the same VM with following configuration
   
   ```
   listen 172.16.0.5:80;
   server 127.0.0.1:8080;
   cache 1;
   cache_fulfill * *;
   ```
3. execute wget twice with recursive retrieving option, one by one, using:
   
   ```
   $ wget -r 172.16.0.5
   ```
4. the second run can't finish retrieval:
   
   ```
   --2016-11-02 22:05:13--  http://172.16.0.5/css/bootstrap.css
   Reusing existing connection to 172.16.0.5:80.
   HTTP request sent, awaiting response... 200 OK
   Length: 127337 (124K) [text/css]
   Saving to: ‘172.16.0.5/css/bootstrap.css’
   
   172.16.0.5/css/bootstrap.cs  67%[==========================>              ]  84.35K  --.-KB/s    in 74s     
   
   2016-11-02 22:06:28 (1.14 KB/s) - Connection closed at byte 86377. Retrying.
   
   --2016-11-02 22:06:29--  (try: 2)  http://172.16.0.5/css/bootstrap.css
   Connecting to 172.16.0.5:80... connected.
   HTTP request sent, awaiting response... 200 OK
   Length: 127337 (124K) [text/css]
   Saving to: ‘172.16.0.5/css/bootstrap.css’

   172.16.0.5/css/bootstrap.cs  67%[==========================>              ]  84.35K  --.-KB/s    in 75s     

   2016-11-02 22:07:44 (1.12 KB/s) - Connection closed at byte 86377. Retrying.
   ...........................
   ```

`tcpdump` shows that Tempesta returns only partial data for the file, so it seems this is TDB issue. Need to debug how much data is stored in cache and why not whole data is returned.

keshonok:
[...]

PR #637, opened 2016-11-07
Correctly save the body of a message. (#630)

keshonok:
Remove the condition that prevented saving parts of the body.
Use BUG_ON() to enforce the correct length of the body.


krizhanovsky:
Good fix! Good to merge

Issue #629, opened 2016-11-02
Response with an empty body and missing Content-Length: header is not handled correctly


keshonok:
When there's no `Content-Length:` header field in a response, then the response is considered complete after a server closes the connection. That works fine if a response has a body. However it doesn't work correctly when there's just the response header and no response body.

keshonok:
When a CRLF is found after the set of all HTTP headers of a response, the parser moves to `Resp_BodyUnlimStart` state. As there's no body (yet), the parser sees that there's no more data and returns to the caller with the result `POSTPONE`. Then the back end server closes the connection, and in this case that means that the HTTP response has no body indeed. However, the parser has never really entered and processed the state `Resp_BodyUnlimStart`, and therefore has never moved to the next state `Resp_BodyUnlimRead`. The function `tfw_http_parse_terminate()` is based on the fact that the parser is in the state `Resp_BodyUnlimRead`, and therefore it doesn't return the expected result.

Resume: The function `tfw_http_parse_terminate()` must expect both `Resp_BodyUnlimStart` and `Resp_BodyUnlimRead` states.

ikoveshnikov:
fixed in dc2409bd84864f282092ace89b9777715d015b1a

Issue #534, opened 2016-05-26
Web-server: send large resources in chunked mode


krizhanovsky:
After #471 implementation we need to be able to send large resources in chunked encoding instead of Content-Length only. It seems the chunks length should be near to page size to perfectly work with TDB data and `skb` fragments.

#391 first points are related to the issue. Issue #498 requires `proxy_buffering` option which is related to the issue.

krizhanovsky:
Also consider https://github.com/tempesta-tech/tempesta/pull/638 and corresponding issue https://github.com/tempesta-tech/tempesta/issues/639#issuecomment-260643114 : we must be able to transform response body to chunked encoding.

krizhanovsky:
This is HTTP/1 specific and the protocol isn't crucial for the web server mode anymore

PR #638, opened 2016-11-09
http parser: fix parsing connection termination during response witho…

ikoveshnikov:
…ut known lenght

For responses without Content-Length header parser during
`TFW_HTTP_INIT_RESP_BODY_PARSING()` procedure switches its state to
`Resp_BodyUnlimStart` state. If response body is present parser will
switch to `Resp_BodyUnlimRead` state. When server will terminate
connection to indicate that end of response is reached, message will
be forwarded to client.

But if no body is present, parser will never reach `Resp_BodyUnlimRead`
state and message will not be forwarded to client.

fix issue #629

### DIFF
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -768,6 +768,13 @@ __FSM_STATE(RGen_BodyInit) {						\
 		FSM_EXIT();						\
 	}								\
 	/* Process the body until the connection is closed. */		\
+	/*								\
+	 * TODO: Currently Tempesta fully assembles response before	\
+	 * transmitting it to a client. This behaviour is considered	\
+	 * dangerous and the issue must be solved in generic way:	\
+	 * Tempesta must use chunked transfer encoding for proxied	\
+	 * responses w/o lengths. Refer issue #534 for more information	\
+	 */								\
 	__FSM_MOVE_nofixup(Resp_BodyUnlimStart);			\
 }
 
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -3208,16 +3208,19 @@ tfw_http_parse_terminate(TfwHttpMsg *hm)
 	if (hm->parser.state == Resp_BodyUnlimRead
 	    || hm->parser.state == Resp_BodyUnlimStart)
 	{
-		char c_len[20] = {0};
+		char c_len[TFW_ULTOA_BUF_SIZ] = {0};
+		size_t digs;
 		int r;
 
 		BUG_ON(hm->body.flags & TFW_STR_COMPLETE);
 		hm->body.flags |= TFW_STR_COMPLETE;
 		hm->content_length = hm->body.len;
-		sprintf(c_len, "%lu", hm->content_length);
-		r = tfw_http_msg_hdr_xfrm(hm, "Content-Length:",
-					  sizeof("Content-Length:") - 1,
-					  c_len, strlen(c_len),
+		if (!(digs = tfw_ultoa(hm->content_length, c_len,
+				       TFW_ULTOA_BUF_SIZ)))
+			return false;
+		r = tfw_http_msg_hdr_xfrm(hm, "Content-Length",
+					  sizeof("Content-Length") - 1,
+					  c_len, digs,
 					  TFW_HTTP_HDR_CONTENT_LENGTH, 0);
 		return (r == 0);
 	}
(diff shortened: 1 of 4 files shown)


tempesta_fw/http_parser.c
```diff
@@ -3195,7 +3195,8 @@ tfw_http_parse_terminate(TfwHttpMsg *hm)
 	BUG_ON(!hm);
 	BUG_ON(!(TFW_CONN_TYPE(hm->conn) & Conn_Srv));
 
-	if (hm->parser.state == Resp_BodyUnlimRead) {
+	if (hm->parser.state == Resp_BodyUnlimRead
+	    || hm->parser.state == Resp_BodyUnlimStart) {
```
keshonok: In a complex and lengthy `if` statements it's difficult then to separate the body of the `if` block from the conditions. For that reason the opening curly brace if moved to the next, separate line.

ikoveshnikov:
Changes since last review:
- Rebased to master
- fix issue #639 by adding Content-Length to message with unlimited body

tempesta_fw/http_parser.c
```diff
@@ -3199,11 +3199,27 @@ tfw_http_parse_terminate(TfwHttpMsg *hm)
...
+	    || hm->parser.state == Resp_BodyUnlimStart)
+	{
+		char c_len[20] = {0};
+		int r;
+
 		BUG_ON(hm->body.flags & TFW_STR_COMPLETE);
 		hm->body.flags |= TFW_STR_COMPLETE;
 		hm->content_length = hm->body.len;
-		return true;
+		sprintf(c_len, "%lu", hm->content_length);
```
krizhanovsky: We do not use format strings in Tempesta FW since they're slow. In fact you know the argument type and don't need to run generic format string logic. Secondly do not use length-less functions, i,e, `snprintf()` should be preferred to `sprintf()`. Please add following code (with appropriate kernel API) to str.[ch] with appropriate tests and use the function here.

        #include <stdio.h>
        
        /*
         * Quickly get number of digits - 1, e.g. returns '0' for '7' and '2' for '333'.
         * It's assumed that small numbers are likely.
         */
        static inline unsigned int
        dig_num(unsigned int a)
        {
        	if (a < 10)
        		return 0;
        	if (a < 100)
        		return 1;
        	if (a < 1000) 
        		return 2;
        	if (a < 10000)
        		return 3;
        	if (a < 100000)
        		return 4;
        	if (a < 1000000)
        		return 5;
        	if (a < 10000000)
        		return 6;
        	if (a < 100000000)
        		return 7;
        	if (a < 1000000000)
        		return 8;
        	return 9;
        }
        
        #define unlikely(c)	__builtin_expect((c), 0)
        
        static size_t
        uitoa(unsigned int ai, char *buf, size_t len)
        {
        	size_t n = dig_num(ai);
        	char *p = buf + n;
        
        	if (unlikely(n > len))
        		return 0;
        
        	do {
        		/* Compiled to only one MUL instruction with -O2. */
        		*p-- = "0123456789"[ai % 10];
                	ai /= 10;
        	} while (ai);
        
        	return n + 1;
        }
        
        #define T(a)								\
        do {									\
        	char buf[10];							\
        	size_t n = uitoa(a, buf, 10);					\
        	printf("%u -> %.*s (len=%u)\n", a, n, buf, n);			\
        } while (0)
        
        int
        main(int argc, char *argv[])
        {
        
        	T(0);
        	T(10);
        	T(0xff);
        	T(0xaff);
        	T(0x45ff);
        	T(0xaabbccff);
        	T(0xffffffff);
        
        	return 0;
        }

tempesta_fw/http_parser.c
```diff
@@ -3199,11 +3199,27 @@ tfw_http_parse_terminate(TfwHttpMsg *hm)
...
+		r = tfw_http_msg_hdr_xfrm(hm, "Content-Length:",
+					  sizeof("Content-Length:") - 1,
+					  c_len, strlen(c_len),
+					  TFW_HTTP_HDR_CONTENT_LENGTH, 0);
```
krizhanovsky: Actually, this doesn't satisfy comment https://github.com/tempesta-tech/tempesta/issues/639#issuecomment-260643114 : according to the logic we must fully read while response before we can calculate Content-Length value. However, converting the whole response body to chunked encoding requirex XFRM logic for body which we don't have for now. Also we need to consider which chunks size we should use. So please mark the code with TODO comment mentioning issue #534 - we must rework the code.

PR #641, opened 2016-11-15
Correctly handle msg->crlf.

keshonok:
- Do not add extra chunks to msg->crlf in case of a chunked body
  and trailing headers.
- In tests, do not send "Content-Length:" header together with
  "Transfer-Encoding:" header.
- Check that the parsed body length matches the value specified
  in "Content-Length:" header only for messages with linear body
  (not chunked).

### DIFF
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -1064,8 +1064,6 @@ __parse_transfer_encoding(TfwHttpMsg *hm, unsigned char *data, size_t len)
 		 * compress;
 		 */
 		__FSM_I_MATCH_MOVE(token, I_TransEncodOther);
-		if (__fsm_sz)
-			msg->flags &= ~__TFW_HTTP_CHUNKED_LAST;
 		c = *(p + __fsm_sz);
 		if (IS_WS(c) || c == ',')
 			__FSM_I_MOVE_n(I_EoT, __fsm_sz + 1);
@@ -1078,8 +1076,11 @@ __parse_transfer_encoding(TfwHttpMsg *hm, unsigned char *data, size_t len)
 	__FSM_STATE(I_EoT) {
 		if (IS_WS(c) || c == ',')
 			__FSM_I_MOVE(I_EoT);
-		if (IS_TOKEN(c))
+		if (IS_TOKEN(c)) {
+			/* "chunked" is not the last coding. */
+			msg->flags &= ~__TFW_HTTP_CHUNKED_LAST;
 			__FSM_I_MOVE_n(I_TransEncodChunked, 0);
+		}
 		if (IS_CRLF(c))
 			return __data_off(p);
 		return CSTR_NEQ;
@@ -2011,8 +2012,9 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 	/* ----------------    Header Lines    ---------------- */
 
 	/*
-	 * Start of HTTP header or end of header part of the request.
-	 * There is a switch for first character of a header name.
+	 * The start of an HTTP header or the end of the header part
+	 * of the request. There is a switch for the first character
+	 * of a header field name.
 	 */
 	__FSM_STATE(RGen_Hdr) {
 		TFW_HTTP_PARSE_CRLF();
@@ -3330,7 +3332,11 @@ tfw_http_parse_resp(void *resp_data, unsigned char *data, size_t len)
 
 	/* ----------------    Header Lines    ---------------- */
 
-	/* Start of HTTP header or end of whole request. */
+	/*
+	 * The start of an HTTP header or the end of the header part
+	 * of the response. There is a switch for the first character
+	 * of a header field name.
+	 */
 	__FSM_STATE(RGen_Hdr) {
 		TFW_HTTP_PARSE_CRLF();
 
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -707,12 +707,8 @@ __FSM_STATE(RGen_BodyInit) {						\
 		/* The alternative: remove "Content-Length" header. */	\
 		if (!TFW_STR_EMPTY(&tbl[TFW_HTTP_HDR_CONTENT_LENGTH]))	\
 			return TFW_BLOCK;				\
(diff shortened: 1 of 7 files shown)


tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -799,7 +798,6 @@ TEST(http_parser, chunked)
 	}
 
 	FOR_RESP("HTTP/1.1 200 OK\r\n"
-		 "Content-Length: 100\r\n"
```
krizhanovsky: As we discussed yesterday, the cases should not be just removed. Rather the cases must be added to `mangled_messages()` with comment referencing to the RFC why the requests and responses are not valid. The test must ensure that the messages are blocked.

tempesta_fw/http_parser.c
```diff
@@ -697,25 +698,28 @@ __FSM_STATE(RGen_HdrOtherV) {						\
...
-	 * there's NO "chunked" coding, then send 400 response		\
-	 * (Bad Request) and close the connection.			\
-	 */								\
+	if (!TFW_STR_EMPTY(&tbl[TFW_HTTP_HDR_TRANSFER_ENCODING])) {	\
+		/* The alternative: remove "Content-Length" header. */	\
+		if (!TFW_STR_EMPTY(&tbl[TFW_HTTP_HDR_CONTENT_LENGTH]))	\
+			return TFW_BLOCK;				\
+		if (msg->flags & TFW_HTTP_CHUNKED) {			\
+			/* "chunked" must be the last in the list. */	\
+			if (!(msg->flags & __TFW_HTTP_CHUNKED_LAST))	\
```
ikoveshnikov: This check is related to validation of `Transfer-Encoding` header. I think better place for it is `__parse_transfer_encoding()` function

tempesta_fw/http_parser.c
```diff
@@ -729,38 +733,35 @@ __FSM_STATE(RGen_BodyInit) {						\
...
-	 * until the connection is closed.				\
-	 */								\
-	if (!TFW_STR_EMPTY(&msg->h_tbl->tbl[TFW_HTTP_HDR_CONTENT_LENGTH])) \
+	if (!TFW_STR_EMPTY(&tbl[TFW_HTTP_HDR_TRANSFER_ENCODING])) {	\
+		/* The alternative: remove "Content-Length" header. */	\
+		if (!TFW_STR_EMPTY(&tbl[TFW_HTTP_HDR_CONTENT_LENGTH]))	\
+			return TFW_BLOCK;				\
+		if (msg->flags & TFW_HTTP_CHUNKED) {			\
+			/* "chunked" must be the last in the list. */	\
+			if (!(msg->flags & __TFW_HTTP_CHUNKED_LAST))	\
```
ikoveshnikov: Same here. Please see previous comment

tempesta_fw/http_parser.c
```diff
@@ -1031,37 +1032,40 @@ __parse_transfer_encoding(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+			msg->flags |= TFW_HTTP_CHUNKED
+				      | __TFW_HTTP_CHUNKED_LAST;
 		}, I_EoT);
 		TRY_STR_INIT();
-		__FSM_I_MOVE_n(I_TransEncodExt, 0);
+		__FSM_I_MOVE_n(I_TransEncodOther, 0);
 	}
 
-	__FSM_STATE(I_TransEncodExt) {
+	__FSM_STATE(I_TransEncodOther) {
```
ikoveshnikov: Every time we reach this state when `TFW_HTTP_CHUNKED` is already set we can invalidate message: chunked encoding is not the last encoding. And we don't need to use `__TFW_HTTP_CHUNKED_LAST`flag

tempesta_fw/t/unit/helpers.c
```diff
@@ -33,6 +33,9 @@
  */
 #include "http_msg.h"
 
+TfwConnection conn_req = { .proto.type = Conn_HttpClnt };
+TfwConnection conn_resp = { .proto.type = Conn_HttpSrv };
```
krizhanovsky: Shouldn't they be static to keep the names local for the translation unit?

tempesta_fw/t/fuzzer.c
```diff
@@ -33,14 +34,19 @@ MODULE_VERSION("0.1.2");
 MODULE_LICENSE("GPL");
```
krizhanovsky: Please don't forget to update module version for such large changes, `0.1.3` should be now.

tempesta_fw/t/fuzzer.c
```diff
@@ -119,64 +134,107 @@ static FuzzMsg expires[] = {
...
-};
+/*
+ * A function that makes sure that the header field value is compatible
+ * with the RFC. That is needed in complex cases where the correctness
+ * may depend on internal or auxiliary values.
+ * The function must return the flags of the field ORed with the result
+ * of either FUZZ_VALID or FUZZ_INVALID.
+ */
+typedef unsigned int (*fld_func)(TfwFuzzContext *ctx,
+				 int type, int fld, int val);
```
krizhanovsky: It's better to use `fld_func_t` such that it's clear that it's a type.

tempesta_fw/t/fuzzer.c
```diff
@@ -365,54 +435,63 @@ __add_field(TfwFuzzContext *ctx, char **p, char *end, int t, int n)
...
 }
 
 static unsigned int
-add_header(TfwFuzzContext *ctx, char **p, char *end, int t)
+__add_header_rand(TfwFuzzContext *ctx, int type,
+		  char **p, char *end, int t, int n)
+{
+	unsigned int rand;
+
+	get_random_bytes(&rand, sizeof(rand));
```
krizhanovsky: The fuzzer must be deterministic. I.e. if you run it and face an issue you should be able to reproduce the issue. If you use random gnerator, then it's hard to reproduce the fuzzer results. This is not ultimate fuzzing, but rather it's handy tool to efficiently generate various HTTP message variants which are hard to code all by hands. Also y we don't want to exhaust system entropy. Probably continuously incrementing static variable would be good. I reckon compiler optimizes `% 2` to `& 0x1`, but probably it's better to avoid div operation.

PR #649, opened 2016-11-22
Fix the handling of crlf and related tests

keshonok:


tempesta_fw/http.c
```diff
@@ -966,7 +966,8 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
 			 * The request is fully parsed,
 			 * fall through and process it.
 			 */
-			BUG_ON(req->content_length != req->body.len);
+			BUG_ON(!(req->flags & TFW_HTTP_CHUNKED)
+			       && (req->content_length != req->body.len));
 			;
```
krizhanovsky: We don't need nop `;` here and at the below - `BUG_ON()` is compiled only leaving trailing `;`.

Issue #634, opened 2016-11-04
[Cache] Do not copy HTTP headers


krizhanovsky:
Currently we copy HTTP server headers from TDB. We do this to be able to adjust them in `tfw_http_adjust_resp()`. However, with #409 in mind and that we should store responses with already calculated Via and Server headers instead of adjusting them for each serviced request. So in general there should be two `tfw_http_adjust_resp()` versions: one of them is executed when response is received and cached and the second one does client specific adjustments.

So after the changes we can directly send headers and status line from TDB as we do this for response body.

We shall form only one skb for small responses avoiding multiple skb fragments. However, paged fragment seems applicable and are the way how the issue should be implemented.

ikoveshnikov:
Keeping in mind implementation of #409: #650.  While setting right value for Server header hop-by-hop mark for that header should be removed. This should be done before removing hop-by-hop headers

krizhanovsky:
The issue didn't pay attention to TLS: we have to copy data for encryption, i.e. we can not directly put an skb data to a socket any more. The only thing which we can do is to optimize copying and fragmentation in `tfw_http_adjust_resp()` - the subject for #1103.

krizhanovsky:
We do need to copy cached HTTP response, however, HTTP headers are copied twice now - in `tfw_cache_build_resp()` and `tfw_tls_encrypt()`. There are also problems with HTTP headers adjustments (#1103) between the copies. Thus, to behave more efficiently in TLS and HTTP/2 we should do following:

1. Currently we store HTTP responses in our own representation, e.g. `TfwCStr` is used to write string data with prefix length. This is very close to HPACK. So to minimize memory footprint and speedup HTTP/2 cache service, we should store HTTP responses in HAPCK format. HTTP static table should only be used for generic representation.

2. Do not spend processing resources on Hufman dynamic encoding - just send HTTP responses from cache as is.

3. Invocation and the calling loop for `tfw_cache_build_resp_hdr()` in `tfw_cache_build_resp()` should be integrated with `tfw_http_adjust_resp()` (via callback for example) to generate all required HTTP/1 or HTTP/2 headers in one shot with only one memory allocation. Tightly linked with #1103.

[...]

PR #650, opened 2016-11-24
Don't cache and don't forward hop-by-hop headers

ikoveshnikov:
Fix issue #409 

- Parse Connection header during parsing HTTP message;
- Store raw headers listed in Connection header in temporary array allocated in message's pool: every raw header is marked as hop-by-hop if it's name present in that array. When connection header parsed in the first time, parser checks already parsed raw headers and mark them as hop-by-hop headers if needed. If Connection header does not contain any raw headers names no extra allocations or string comparison happen;
- Special headers have their own marking procedure which does not involve string comparisons. Some of the special headers are always hop-by-hop (Connection, we also mark Server in responses as hop-by-hop) and using generic comparison like for Raw headers will ruin performance;
- Unit tests for responses and requests with hop-by-hop marking checks; 
- little optimisation of __hdr_lookup(): reduce string comparisons.

*TBD:* 
When response contains `Connection: keep-alive`, `Keep-alive` header will be marked as hop-by-hop header and will not be saved in cache (that was the requirement in #409 ). RFC 7230 6.1 also say:
```
 A proxy or gateway MUST parse a received Connection header field before
   a message is forwarded and, for each connection-option in this field,
   remove any header field(s) from the message with the same name as the
   connection-option, and then remove the Connection header field itself
   (or replace it with the intermediary's own connection options for the
   forwarded message).
```
Currently we have no realised way to set default `Keep-Alive` header (not found in issues but here is corresponding TODO: https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/http.c#L639 ). So question is: delete `Keep-Alive` header in forwarded messages or not? But messages served from cache will have no such header. Currently i remove it from both forwarded messages: response and request, disabling that is one-line patch.

### DIFF
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -810,6 +810,39 @@ tfw_http_msg_free(TfwHttpMsg *m)
 EXPORT_SYMBOL(tfw_http_msg_free);
 
 /**
+ * Add spec header indexes to list of hop-by-hop headers.
+ */
+static inline void
+__hbh_parser_init_req(TfwHttpReq *req)
+{
+	TfwHttpHbhHdrs *hbh_hdrs = &req->parser.hbh_parser;
+
+	BUG_ON(hbh_hdrs->spec);
+	/* Connection is hop-by-hop header by RFC 7230 6.1 */
+	hbh_hdrs->spec = 0x1 << TFW_HTTP_HDR_CONNECTION;
+}
+
+/**
+ * Same as @__hbh_parser_init_req for response.
+ */
+static inline void
+__hbh_parser_init_resp(TfwHttpResp *resp)
+{
+	TfwHttpHbhHdrs *hbh_hdrs = &resp->parser.hbh_parser;
+
+	BUG_ON(hbh_hdrs->spec);
+	/*
+	 * Connection is hop-by-hop header by RFC 7230 6.1
+	 *
+	 * Server header isn't defined as hop-by-hop by the RFC, but we
+	 * don't show protected server to world.
+	 */
+	hbh_hdrs->spec = (0x1 << TFW_HTTP_HDR_CONNECTION) |
+			 (0x1 << TFW_HTTP_HDR_SERVER);
+
+}
+
+/**
  * Allocate a new HTTP message.
  * Given the total message length as @data_len, it allocates an appropriate
  * number of SKBs and page fragments to hold the payload, and sets them up
@@ -841,6 +874,10 @@ tfw_http_msg_alloc(int type)
 	INIT_LIST_HEAD(&hm->msg.msg_list);
 
 	hm->parser.to_read = -1; /* unknown body size */
+	if (type & Conn_Clnt)
+		__hbh_parser_init_req((TfwHttpReq *)hm);
+	else
+		__hbh_parser_init_resp((TfwHttpResp *)hm);
 
 	if (type & Conn_Clnt)
 		hm->destructor = tfw_http_req_destruct;
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -324,22 +324,22 @@ __tfw_http_msg_add_str_data(TfwHttpMsg *hm, TfwStr *str, void *data,
 }
 
 int
-__tfw_http_msg_grow_hdr_tbl(TfwHttpHdrTbl **ht, TfwPool * pool)
+tfw_http_msg_grow_hdr_tbl(TfwHttpMsg *hm)
 {
-	TfwHttpHdrTbl *new_ht = *ht;
-	size_t order = new_ht->size / TFW_HTTP_HDR_NUM, new_order = order << 1;
+	TfwHttpHdrTbl *ht = hm->h_tbl;
+	size_t order = ht->size / TFW_HTTP_HDR_NUM, new_order = order << 1;
 
-	new_ht = tfw_pool_realloc(pool, new_ht, TFW_HHTBL_SZ(order),
+	ht = tfw_pool_realloc(hm->pool, ht, TFW_HHTBL_SZ(order),
 			      TFW_HHTBL_SZ(new_order));
-	if (!new_ht)
+	if (!ht)
 		return -ENOMEM;
-	new_ht->size = __HHTBL_SZ(new_order);
-	new_ht->off = (*ht)->off;
-	memset(new_ht->tbl + __HHTBL_SZ(order), 0,
+	ht->size = __HHTBL_SZ(new_order);
+	ht->off = hm->h_tbl->off;
+	memset(ht->tbl + __HHTBL_SZ(order), 0,
 	       __HHTBL_SZ(order) * sizeof(TfwStr));
-	*ht = new_ht;
+	hm->h_tbl = ht;
 
-	TFW_DBG3("grow http headers table to %d items\n", new_ht->size);
+	TFW_DBG3("grow http headers table to %d items\n", ht->size);
 
 	return 0;
 }
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -576,6 +576,27 @@ tfw_http_msg_hdr_xfrm(TfwHttpMsg *hm, char *name, size_t n_len,
 }
 
 /**
+ * Remove hop-by-hop headers in the message
+ */
+int
+tfw_http_msg_del_hbh_hdrs(TfwHttpMsg *hm)
+{
+	TfwHttpHdrTbl *ht = hm->h_tbl;
+	unsigned int hid = ht->off;
+	int r = 0;
+
+	do {
+		hid--;
+		if (ht->tbl[hid].flags & TFW_STR_HBH_HDR)
+			if ((r = __hdr_del(hm, hid)))
+				return r;
+	} while (hid != 0);
+
+	return 0;
+}
+
+
+/**
  * Add a header, probably duplicated, without any checking of current headers.
  */
 int
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -166,11 +166,13 @@ __hdr_lookup(TfwHttpMsg *hm, const TfwStr *hdr)
 		/* There is no sense to compare against all duplicates. */
 		if (h->flags & TFW_STR_DUPLICATE)
 			h = TFW_STR_CHUNK(h, 0);
-		if (tfw_stricmpspn(hdr, h, ':'))
-			continue;
+		if (!tfw_stricmpspn(hdr, h, ':'))
+			break;
+	}
+
+	if (id <  ht->off) {
 		if (__hdr_is_singular(hdr))
 			hm->flags |= TFW_HTTP_FIELD_DUPENTRY;
-		break;
 	}
 
 	return id;
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -152,13 +152,9 @@ __hdr_is_singular(const TfwStr *hdr)
  * i.e. check whether the header is duplicate.
  * The lookup is performed untill ':', so header name only is enough in @hdr.
  * @return the header id.
- *
- * Certain header fields are strictly singular and may not be repeated in
- * an HTTP message. Duplicate of a singular header fields is a bug worth
- * blocking the whole HTTP message.
  */
-static unsigned int
-__hdr_lookup(TfwHttpMsg *hm, const TfwStr *hdr)
+unsigned int
+tfw_http_msg_hdr_lookup(TfwHttpMsg *hm, const TfwStr *hdr)
 {
 	unsigned int id;
 	TfwHttpHdrTbl *ht = hm->h_tbl;
@@ -172,10 +168,21 @@ __hdr_lookup(TfwHttpMsg *hm, const TfwStr *hdr)
 			break;
 	}
 
-	if (id <  ht->off) {
-		if (__hdr_is_singular(hdr))
-			hm->flags |= TFW_HTTP_FIELD_DUPENTRY;
-	}
+	return id;
+}
+
+/**
+ * Certain header fields are strictly singular and may not be repeated in
+ * an HTTP message. Duplicate of a singular header fields is a bug worth
+ * blocking the whole HTTP message.
+ */
+static inline unsigned int
+__hdr_lookup(TfwHttpMsg *hm, const TfwStr *hdr)
+{
+	unsigned int id = tfw_http_msg_hdr_lookup(hm, hdr);
+
+	if ((id <  hm->h_tbl->off) && __hdr_is_singular(hdr))
+		hm->flags |= TFW_HTTP_FIELD_DUPENTRY;
 
 	return id;
 }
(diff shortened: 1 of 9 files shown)


tempesta_fw/http_parser.c
```diff
@@ -457,6 +457,134 @@ parse_int_hex(unsigned char *data, size_t len, unsigned long *acc)
...
+static void
+mark_raw_hbh(TfwHttpMsg *hm, TfwStr *hdr)
+{
+	TfwHttpHdrTbl *ht = hm->parser.hbh_parser.raw;
+	unsigned int i;
+
+	if (!ht)
+		return;
+
+	for (i = 0; i < ht->off; ++i) {
```
krizhanovsky: Does RFC allow setting headers, mentioned in Connection, before Connection header? The code is dangerous, just consider the example:

        GET / HTTP/1.1\r\n
        Connection: keep-alive, Host, User-Agent\r\n
        Host: foo.com\r\n
        User-Agent: malicious client\r\n
        \r\n

So server will receive malformed request.

We should run the loop at least from TFW_HTTP_HDR_RAW. Probably there is some limitation which headers can be mentioned in Connection header... The other way (I think the safest one) is to remove Keep-Alive only and ignore any other mentioned headers in Connection. However, we still use current mechanism for the headers removal such that we'll be able to remove other headers which are known to be mentioned in Connection. I argue against blind headers removal.

tempesta_fw/http_parser.c
```diff
@@ -457,6 +457,134 @@ parse_int_hex(unsigned char *data, size_t len, unsigned long *acc)
...
+ * with that name and mark them as hop-by-hop.
+ */
+static int
+__hbb_parser_add_data(TfwHttpMsg *hm, char *data, unsigned long len, bool last)
+{
+	unsigned int hid;
+	TfwStr *hdr, *append;
+	int r;
+	TfwHttpHdrTbl *ht = hm->parser.hbh_parser.raw;
+	if (!ht) {
```
krizhanovsky: This isn't mentioned in our Coding Style, but [FreeBSD KNF](https://www.freebsd.org/cgi/man.cgi?query=style&sektion=9) requires split variables definitions and code. It's bit hard to read when code immediately follows variable definitions immediately.

tempesta_fw/http_parser.c
```diff
@@ -457,6 +457,134 @@ parse_int_hex(unsigned char *data, size_t len, unsigned long *acc)
...
+	hm->parser.hbh_parser.raw->size = __HHTBL_SZ(1);
+	hm->parser.hbh_parser.raw->off = 1;
+	memset(hm->parser.hbh_parser.raw->tbl, 0,
+	       __HHTBL_SZ(1) * sizeof(TfwStr));
+
+	return 0;
+}
+
+/**
+ * Add header name listed in Connection header to rable of raw headers.
```
krizhanovsky: rable -> table

tempesta_fw/http_parser.c
```diff
@@ -457,6 +457,134 @@ parse_int_hex(unsigned char *data, size_t len, unsigned long *acc)
...
+			return r;
+		ht = hm->parser.hbh_parser.raw;
+	}
+
+	if (ht->tbl[ht->off - 1].flags & TFW_STR_COMPLETE) {
+		hid = ht->off;
+		if (hid == ht->size)
+			if (__tfw_http_msg_grow_hdr_tbl(&hm->parser.hbh_parser.raw,
+							hm->pool))
+				return -ENOMEM;
```
krizhanovsky: This is true and powerful headers parser. Do we really allow that Connection mentioning more than 16 header? I believe it has sense to block the header. Meantime it will make the code faster and easier and that's good.

tempesta_fw/http_parser.c
```diff
@@ -917,8 +1069,24 @@ __parse_connection(TfwHttpMsg *hm, unsigned char *data, size_t len)
 	 * it could be names of any headers, including custom headers.
 	 */
 	__FSM_STATE(I_ConnOther) {
-		__FSM_I_MATCH_MOVE(token, I_ConnOther);
+		/* Unwind __FSM_I_MATCH_MOVE(token, I_ConnOther): */
+		__fsm_n = __data_remain(p);
```
krizhanovsky: There is MATCH_MOVE macroses which embody the code.

tempesta_fw/http_parser.c
```diff
@@ -2079,6 +2318,17 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 				__FSM_MOVE_n(RGen_OWS, 5);
 			}
 			__FSM_MOVE(Req_HdrH);
+		case 'k':
+			if (likely(__data_available(p, 11)
+				   && C4_INT_LCM(p + 1, 'e', 'e', 'p', '-')
+				   && C4_INT_LCM(p + 5, 'a', 'l', 'i', 'v')
```
krizhanovsky: The discussed bug to fix

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -877,6 +877,219 @@ TEST(http_parser, cookie)
 		"\r\n");
 }
 
+TEST(http_parser, req_hop_by_hop)
+{
```
krizhanovsky: Please add test case described above for "Connection: Host\r\n' to mangled message.

tls/ttls.c
```diff
@@ -128,7 +128,7 @@ module_exit(ttls_exit);
 
 MODULE_AUTHOR("Tempesta Technologies");
 MODULE_VERSION("2.3.0");
-MODULE_LICENSE("GPLv2");
+MODULE_LICENSE("GPL v2");
```
krizhanovsky: Doesn't it taints kernel? The change is fine if not.

krizhanovsky:
Please address the comment https://github.com/tempesta-tech/tempesta/issues/409#issuecomment-258436328 or move the requirement to #634

tempesta_fw/http_parser.c
ikoveshnikov: 1. RFC does not have any restrictions where headers listed in Connection header must be placed. So both variants are ok: before or after Connection header. More over i tried to set up custom hop-by-hop headers in ngingx and apache, both allow setting custom headers before and after the header. Here example from RFC 6455 with hop-by-hop header before Connection header:
```
GET /chat HTTP/1.1
Host: server.example.com
Upgrade: websocket
Connection: Upgrade
Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==
Origin: http://example.com
Sec-WebSocket-Protocol: chat, superchat
Sec-WebSocket-Version: 13
```

2. Table you point to stores names for only raw headers. I agree, it could be confusing, but i used data structure which suit me perfectly. In situation you show user-agent and every other spec header listed in connection header (except keep-alive) will not be removed

Obsoleted now, in new revision this array is statically allocated

ikoveshnikov:
Changes since last review:
- rebased to current master
- fixed error in current master with Transfer-Encoding header: the header now is SPEC header but it missed in `__http_msg_hdr_val()`
- Connection header is limited to have less than 16 connection tokens for raw headers. Otherwise request or response will be filtered 
- Drop message if some of connection tokens is end-to-end headers used by tempesta and parsed by TFW_HTTP_PARSE_RAWHDR_VAL macro. Comparing of connection tokens with all headers defined in RFC 7230, 7231 as end-to-end will lead to a lot of string comparisons inside of parsing of Connection header 
- Unit tests updated to have more test cases for hop-by-hop headers marking

The feature have no special requirements for #634 . (Headers listed there are not hop-by-hop)

tempesta_fw/http_msg.c
```diff
@@ -573,12 +577,62 @@ tfw_http_msg_hdr_xfrm(TfwHttpMsg *hm, char *name, size_t n_len,
...
+	/*
+	 * This function is called before hm->h_tbl fully parsed,
+	 * If header is epmty, dont touch it
+	 */
+	if ((hid > ht->off) || (TFW_STR_EMPTY(&ht->tbl[hid])))
+		return;
+
+	h = &ht->tbl[hid];
+	TFW_STR_FOR_EACH_DUP(d, h, end)
+		d->flags |= TFW_STR_HBH_HDR;
```
keshonok: Do we really need to mark each duplicate as HBH? I don't think so, but maybe there's a case to prove otherwise?

tempesta_fw/http_msg.c
```diff
@@ -573,12 +577,62 @@ tfw_http_msg_hdr_xfrm(TfwHttpMsg *hm, char *name, size_t n_len,
...
+	return 0;
+}
+
+/**
+ * Lookup for the header @hdr in already collected headers table @ht,
+ * and mark it as hop-by-hop. The lookup is performed untill ':', so header
+ * name only is enough in @hdr.
+ */
+void
+tfw_http_msg_mark_hbh_hdr(TfwHttpMsg *hm, TfwStr *hdr)
```
keshonok: I believe that this function (under a better name) belongs in `http_parser.c`.
If I understand correctly, the reason for placing it here was that `__hdr_lookup()` was located here, and it's a local (static) function to this sub-module. In my opinion though it should be the other way around: `__hdr_lookup()` should be made public (and given a proper name for that) and used from `http_parser.c`

tempesta_fw/http_parser.c
```diff
@@ -462,30 +462,51 @@ parse_int_hex(unsigned char *data, size_t len, unsigned long *acc)
...
+ * Add spec header indexes to list of hop-by-hop headers
  */
-static void
-__hbh_parser_init(TfwHttpHbhHdrs *hbh_hdrs, bool req)
+static inline void
+__hbh_parser_init_req(TfwHttpReq *req)
+{
+	TfwHttpHbhHdrs *hbh_hdrs = &req->parser.hbh_parser;
+	/* Connection is hop-by-hop header by RFC 7230 6.1 */
+	hbh_hdrs->spec |= 0x1 << TFW_HTTP_HDR_CONNECTION;
```
keshonok: As this should be performed only once per message, shouldn't this be a direct assignment, not an OR operation?

tempesta_fw/http_parser.c
```diff
@@ -462,30 +462,51 @@ parse_int_hex(unsigned char *data, size_t len, unsigned long *acc)
...
 	 * Server header isn't defined as hop-by-hop by the RFC, but we
-	 * don't show protected server to world. User-agent and Server headers
-	 * have the same id in tfw_http_hdr_t, so Server header must be marked
-	 * only for responses.
+	 * don't show protected server to world.
 	 */
 	hbh_hdrs->spec |= 0x1 << TFW_HTTP_HDR_CONNECTION;
-	if (!req)
-		hbh_hdrs->spec |= 0x1 << TFW_HTTP_HDR_SERVER;
+	hbh_hdrs->spec |= 0x1 << TFW_HTTP_HDR_SERVER;
```
keshonok: Please see the comment above regarding direct assignment vs OR operation.

tempesta_fw/http_parser.c
```diff
@@ -1348,6 +1348,7 @@ __parse_transfer_encoding(TfwHttpMsg *hm, unsigned char *data, size_t len)
 /* Main (parent) HTTP request processing states. */
 enum {
 	Req_0,
+	Req_CRLF,
```
keshonok: In my opinion, this may appear too generic. Something like **Req_0_CRLF** may visibly clarify that this is one of the initial states.

tempesta_fw/http_parser.c
```diff
@@ -2089,16 +2090,20 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
...
 		len, min(500, (int)len), data, len > 500 ? "..." : "", req);
 
 	__FSM_START(parser->state) {
 
 	/* ----------------    Request Line    ---------------- */
 
+	/* Parser internal initilizers, must be called once per message. */
 	__FSM_STATE(Req_0) {
+		__hbh_parser_init_req(req);
+		/* fall through */
```
ikoveshnikov: I have checked out all the places in code, where _fall through_ used instead of `FSM_MOVE` directives. No errors there and  _fall through_ can be replaced with `__FSM_JMP`. Since desired state is the next one _fall through_ is used.

`__FSM_STATE` saves current parser state to `__fsm_const_state`: https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/http_parser.c#L120. In all the places in parser _fall through_ is used only when we do not move our current position in message and we want to work with the same data in next stage or stages. 

Any flavour of `FSM_MOVE*` is not an option here since `parser->to_go` will be set to 0 for situation you are describing ( https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/http_parser.c#L137 ). And next stage will take *next* character, not the same one.

tempesta_fw/http_parser.c
```diff
@@ -887,8 +1081,38 @@ __FSM_STATE(RGen_OWS) {							\
...
+/* Add spec header of id @hid and @name to list of spec hop-by-hop headers */
+#define TRY_HBH_SPEC_HDR_LAMBDA(name, hid, lambda)			\
+	TRY_HBH_TOKEN(name, {						\
+		BUILD_BUG_ON(sizeof(parser->hbh_parser.spec) * CHAR_BIT	\
+			     <  TFW_HTTP_HDR_RAW);			\
+		lambda;							\
+		parser->hbh_parser.spec |= 0x1 << hid;			\
+		if (!TFW_STR_EMPTY(&msg->h_tbl->tbl[hid]))		\
+			msg->h_tbl->tbl[hid].flags |= TFW_STR_HBH_HDR;	\
+	})
```
krizhanovsky: The macro is used in only one place, so the code would be more readable if you place the macro as inline code at the appropriate place.

tempesta_fw/http_parser.c
```diff
@@ -887,8 +1081,38 @@ __FSM_STATE(RGen_OWS) {							\
...
+
+#define TRY_SPEC_HBH_HDR(name, hid) TRY_SPEC_HBH_HDR_LAMBDA(name, hid, {})
```
krizhanovsky: The macro is not used at all.

tempesta_fw/http_parser.c
```diff
@@ -1781,9 +2101,15 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 
 	/* ----------------    Request Line    ---------------- */
 
+	/* Parser internal initilizers, must be called once per message. */
 	__FSM_STATE(Req_0) {
+		__hbh_parser_init_req(req);
```
krizhanovsky: Why did you create a new parser state for the initialization routine? It seems you can call __hbh_parser_init_*() functions in `tfw_http_msg_alloc()` where base parser is initialized.

tempesta_fw/http_parser.c
```diff
@@ -3264,9 +3558,15 @@ tfw_http_parse_resp(void *resp_data, unsigned char *data, size_t len)
 
 	/* ----------------    Status Line    ---------------- */
 
+	/* Parser internal initilizers, must be called once per message. */
 	__FSM_STATE(Resp_0) {
+		__hbh_parser_init_resp(resp);
```
krizhanovsky: The same as for request

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -950,6 +982,309 @@ TEST(http_parser, cookie)
...
+			default:
+				EXPECT_FALSE(field->flags & TFW_STR_HBH_HDR);
+				break;
+			}
+		}
+	}
+
+	/* Connection header lists end-to-end spec headers */
+	FOR_REQ(REQ_HBH_START
+		"Connection: Host, Content-Length, Content-Type, Connection,"
```
krizhanovsky: I believe the request must be blocked since there is special header in `Connection`

Issue #651, opened 2016-11-24
kernel crash in __pskb_copy_fclone()


keshonok:
This occurs on current master at b2f706a. The configuration is the simplest: the back end server runs in the same virtual machine with Tempesta. The back end server is Apache and it's configured with a copy of tempesta-tech.com web site as a virtual host. All other Apache configuration options are default.
```
server 127.0.0.1:8080;
cache 0;
```

Default configuration options for Apache mean that the server closes connections every 100 requests. As far as I can see that plays a role in the kernel crash.

The crash occurs from time to time when Tempesta is stressed with a large stream of concurrent requests from the host system. It doesn't occur every time but it happens often enough. The following command is used:
```
wrk -t 128 -c 4096 -d 60s http://192.168.10.230/
```
The crash doesn't seem to occur when Tempesta is stressed with Apache's `ab` tool.

Below is the stack trace of the crash.

```
[50264.709671] general protection fault: 0000 [#1] SMP
[50264.710627] Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) tempesta_tls(PO) ppdev 9pnet_virtio aesni_intel aes_x86_64 glue_helper lrw gf128mul ablk_helper cryptd 9pnet microcode virtio_balloon i2c_piix4 i2c_core pcspkr parport_pc parport acpi_cpufreq ip_tables ata_generic pata_acpi crc32c_intel virtio_pci e1000 virtio_ring serio_raw virtio ata_piix floppy ipv6 autofs4 [last unloaded: tempesta_tls]
[50264.710627] CPU: 0 PID: 3 Comm: ksoftirqd/0 Tainted: P           O    4.1.27-ab+ #1
[50264.710627] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS Bochs 01/01/2011
[50264.710627] task: ffff88011af9a090 ti: ffff88011afa0000 task.ti: ffff88011afa0000
[50264.710627] RIP: 0010:[<ffffffff81520567>]  [<ffffffff81520567>] __pskb_copy_fclone+0x1b7/0x230
[50264.710627] RSP: 0018:ffff88011afa3728  EFLAGS: 00010212
[50264.710627] RAX: 0800000000000000 RBX: ffff88000209dc00 RCX: 6b6b6b6bd6d6d6d6
[...]

PR #652, opened 2016-11-25
Correctly lock requests from concurrent deletion. (#651)

keshonok:


keshonok:
In fact, a server connection is get'ed before a request is forwarded to the server over that connection. That's done in an atomic way, so a server connection cannot be get'ed when it is released. Equally, a server connection cannot be released while forwarding action of that request is in progress. While this patch does prevent the crash, that just a workaround that masks the real problem. The core of the problem is different, and the issue must be resolved in a different way.

PR #657, opened 2016-11-29
Proxy performance tuning; fix #655

### LINKED ISSUE #655: Broken server connections failovering
Test scenario is described in #654. The problem is in [tfw_sock_srv_do_failover()](https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/sock_srv.c#L299): we get the connection, such that destructor isn't called from `tfw_connection_put()` while it points to `tfw_srv_conn_release()` which calls `tfw_sock_srv_connect_try_later()`. So if a server connection dies, it's never restored.

krizhanovsky:


tempesta_fw/sock_srv.c
```diff
@@ -170,29 +169,53 @@ tfw_sock_srv_connect_try(TfwSrvConnection *srv_conn)
...
+		srv_conn->timeout = timeouts[srv_conn->attempts];
+		TFW_DBG_ADDR("Cannot establish connection",
+			     &srv_conn->conn.peer->addr);
+	} else {
+		srv_conn->timeout = timeouts[ARRAY_SIZE(timeouts) - 1];
+		if (srv_conn->attempts == ARRAY_SIZE(timeouts)
+		    || !(srv_conn->attempts % 60))
+		{
+			char addr_str[TFW_ADDR_STR_BUF_SIZE] = { 0 };
+			tfw_addr_fmt_v6(&srv_conn->conn.peer->addr.v6.sin6_addr,
```
ikoveshnikov: I didn't get, why unconditionally format string as ipv6 address? If we connect to server by IPv4, string will be filled with garbage. Better to fill `addr_str` with `tfw_addr_fmt_v6` or `tfw_addr_fmt_v4` chosen by value of `addr.family`

tls/ttls.c
```diff
@@ -128,7 +128,7 @@ module_exit(ttls_exit);
 
 MODULE_AUTHOR("Tempesta Technologies");
 MODULE_VERSION("2.3.0");
-MODULE_LICENSE("GPLv2");
+MODULE_LICENSE("GPL");
```
ikoveshnikov: I have nothing against change of license but if this change is to fix tainting of kernel misspelled `GPLv2` token can be replaced with `GPL v2` without change of sense. Here is list of supported license identifiers: http://lxr.free-electrons.com/source/include/linux/module.h#L172

scripts/tfw_lib.sh
```diff
@@ -26,23 +26,80 @@ popd > /dev/null
 declare -r TFW_NAME=`basename $0` # program name (comm name in ps)
 declare -r TFW_NETDEV_PATH="/sys/class/net/"
 declare -r TFW_SCRIPTS="$TFW_ROOT/scripts"
+declare -r CPUS_N=$(grep -c processor /proc/cpuinfo)
 
-declare TFW_DEVS=$(ls $TFW_NETDEV_PATH)
+declare TFW_DEVS=$(ifconfig | grep -o '^[a-z0-9]\+')
```
krizhanovsky: There is no sense to play with queues of down interface, so we fetch up interfaces only.

Issue #488, opened 2016-05-19
SS, classifier: TCP buffering & DDoS mitigation


krizhanovsky:
Synchronous Sockets break TCP buffering: if the system is overloaded, then each TCP segment is processed as it goes to CPU and there is no context which can reduce TCP window. I.e. TCP windows don't work at all. It means that under overload segments are just dropped by NIC and TCP sender just retransmits by timers. This behavior significantly degrades TCP performance.

The possible solution is TCP QoS - manage all TCP connections by clusters, i.e. send zero (or smoothly decreasing value by some more complicated algorithm) window for low-scored clients and provide more service for highly-score client clusters. This mostly relates to DDoS mitigation classification and in fact just an other side of #100: we can close client connection at all ("evict" the connection) or reduce TCP window for the client.

NIC packet drops must be introduced as system stress condition. Probably some more network performance events should be processed as a stress condition triggering connections eviction and/or TCP window shrinking.

One more thing to think about. All work is done is softirq, i.e. we spend more time to process each packet under heavy load, but we don't do anything while system is idle. Small traffic bursts are mitigated by dev queue and Tempesta frees the queue slower than vanilla Linux. However, large traffic bursts lead to TCP buffer overflows in normal system and Tempesta behaves there better since it processes TCP segments faster than usual user-space process. Also more work in softiq means more data structures accessed with larger memory footprint, so CPU cache starvation is more probable. Thus, all cleanups (e.g. connections eviction) and other things (e.g. traffic classification) must be done in designated kernel threads running of designates CPU cores to mitigate cache pressure. The threads must be awakened on system stress events (like VMM exhausting or NIC packet drops).

krizhanovsky:
From #100: the most urgent thing is to keep security accounting data for a client for some time after the last client connection is closed. This is very important to track client security limits properly for `Connection: closed` connections. See https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/client.c#L89

krizhanovsky:
The consequence of the issue also appears on simple test with configuration as
```
listen 192.168.100.4:80;
server 127.0.0.1:9090 conns_n=1;
cache 0;
server_queue_size 1;
```
In this case we get a lot of error responses:
```
# ./wrk -c 4096 -t 8 -d 30 http://192.168.100.4:80/
Running 30s test @ http://192.168.100.4:80/
  8 threads and 4096 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
[...]

Issue #391, opened 2016-01-07
TCP synchronous sending


krizhanovsky:
Current implementation of ```ss_send()``` and cache logic are unsatisfactory:

1. ```tfw_cache_build_resp()``` can produce very large response. Many of the ```skb``` are simply dropped by Qdisc the others will be sent using TSQ tasklet. So there is no sense to do large TDB scans, but rather the scans should be done by small portions when there is room in Qdisc. This also reduces bufferbloat;

2. RX softirq is responsible also for HTTP message retransmissions (requests to a upstream server or responses to client), so we must lock 2 sockets. Giving that requests and responses can came at the same time 2 cpus can try to lock the same sockets in different orders leading to deadlock. Kernel threads using SS interface also suffer from the issue (a socket can be locked by softirq and a kernel thread, see #337). This also bad for performance since different CPUs compete for the same sockets. Thus transmission action must be scheduled to proper CPU and performed from TX softirq;

3. RX softirq can do millions TDB blocks transitions to process one request to large stored file - this leads to significant packet drops from other clients. Moreover, concurrent requests to large files may lead to efficient DoS;

4. There is no sense to replicate multi-gigabyte files among NUMA nodes as well as to bother with work queue to fetch small records from remote node. Large files should be transferred by smaller blocks in work queue as currently. Bloom filter can be used to mark that a record is on the remote node (any new large record insertion should update the filters on all the nodes).

PR #660, opened 2016-12-05
Enforce the correct order of responses. Handle non-idempotent requests.

keshonok:


tempesta_fw/vhost.c
```diff
@@ -247,11 +303,169 @@ tfw_vhost_match(TfwStr *arg)
...
+		return -EINVAL;
+	}
+	if (ce->val_n != 3) {
+		TFW_ERR("%s: Invalid number of arguments.\n", cs->name);
+		return -EINVAL;
+	}
+
+	/* The method: one of GET, PUT, POST, etc. in form of a bitmask. */
+	in_method = (char *)ce->vals[0];
+	ret = tfw_cfg_map_enum(tfw_method_enum, in_method, &method);
```
ikoveshnikov: By the RFC 7231 4.2: 
```
Of the request methods defined by this specification, the GET, HEAD, OPTIONS, and 
TRACE methods are defined to be safe.
...
Of the request methods defined by this specification, PUT, DELETE, and safe request
methods are idempotent.
```

Maybe we should add new enum with methods that could be non-idempotent?

tempesta_fw/http.c
```diff
@@ -382,6 +370,396 @@ tfw_http_send_504(TfwHttpReq *req)
...
+ * The request is added to the server connection (forwarding) queue.
+ * If forwarding is on hold at this moment, then the request will be
+ * forwarded later. Otherwise, if the queue is drained, then forward
+ * the request to the server immediately. If the queue is not drained,
+ * then forward all stalled requests to the server.
+ *
+ * Forwarding to a server is considered to be on hold after
+ * a non-idempotent request is forwarded to the server. The hold
+ * is removed when the holding non-idempotent request is followed
+ * by another request from the same client, which enables pipelining.
```
ikoveshnikov: Is that suits RFC 7230 6.3.2?
```
A user agent SHOULD NOT pipeline requests after a non-idempotent method, 
until the final response status code for that method has been received, unless 
the user agent has a means to detect and recover from partial failure conditions 
involving the pipelined sequence.
```
As i got it, we cannot remove hold untill request was answered

keshonok: Not really. As it's said in the cited excerpt from the RFC, a user agent *should not* pipeline after a non-idempotent method. However, it a user agent has a means to detect and recover from from partial failure conditions 
involving the pipelined sequence, it *may* still pipeline requests after a non-idempotent method. So if another message is received from a client before a response comes to a non-idempotent request, that means that the client knows what it's doing and can recover from failure if necessary. Effectively, that removes the requirement of waiting for a response to a non-idempotent request and makes it possible to continue the pipelining.

tempesta_fw/http.c
```diff
@@ -786,20 +1369,13 @@ tfw_http_req_cache_service(TfwHttpReq *req, TfwHttpResp *resp)
 {
 	if (tfw_http_adjust_resp(resp, req))
 		goto resp_err;
-
-	if (tfw_cli_conn_send(req->conn, (TfwMsg *)resp))
-		goto resp_err;
-
+	tfw_http_resp_fwd(req, resp);
```
krizhanovsky: We should use lightweight logic here as previously, see comment for `tfw_http_req_process()`.

tempesta_fw/http.c
```diff
@@ -1121,32 +1750,37 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
...
+	/*
+	 * Responses from cache don't have @resp->conn.
+	 *
+	 * FIXME: The same check is performed in tfw_http_popreq()
+	 * which happens just a bit earlier. Is there a way to avoid
+	 * it here? The condition is considered rare, and there's no
+	 * need to check for it in the regular path. The real issue
+	 * here is that APM stats can't handle response times that
+	 * are >= USHORT_MAX. So for now don't count the requests
+	 * that are re-sent after a server connection is restored.
```
krizhanovsky: I didn't get what the issue is. APM count delays for particular server. So if we reschedule a request to other server (or ever different connection of the server), then we should use the new, current, request timestamp for APM accounting.

TBD. Probably APM should treat connection error on special way. I.e. if the server drops connections and/or doesn't respond to sent requests at all, then we should adjust some different counters. Meantime I think we can just add maximum possible delay (probably compile-time constant) for request which were sent, but never replied.

tempesta_fw/sched/tfw_sched_rr.c
```diff
@@ -101,32 +101,49 @@ tfw_sched_rr_add_conn(TfwSrvGroup *sg, TfwServer *srv, TfwConnection *conn)
...
+ * in attempt to increase throughput. However, if all live connections
+ * contain non-idempotent requests, then re-run the algorithm and get
+ * the first live connection as it is usually done.
  */
 static TfwConnection *
 tfw_sched_rr_get_srv_conn(TfwMsg *msg, TfwSrvGroup *sg)
 {
-	int c, s, i;
-	TfwConnection *conn;
+	unsigned long idx;
```
krizhanovsky: Please don't rename simple and widespread name to something more complex and long. `i` is widespread identifier for `index` and is frequently used as a loop iterator.

tempesta_fw/sched/tfw_sched_rr.c
```diff
@@ -101,32 +101,49 @@ tfw_sched_rr_add_conn(TfwSrvGroup *sg, TfwServer *srv, TfwConnection *conn)
...
-			conn = srv_cl->conns[i];
-			if (tfw_connection_get_if_nfo(conn))
+			idx = atomic64_inc_return(&srv_cl->rr_counter);
+			conn = srv_cl->conns[idx % srv_cl->conn_n];
+			if (unlikely(tfw_connection_restricted(conn))
+			    || unlikely(tfw_server_queue_full(conn)))
+				continue;
+			if (skipnip && tfw_connection_hasnip(conn)) {
+				if (likely(tfw_connection_live(conn)))
+					nipconn++;
```
krizhanovsky: The good question and the answer should be commented for the function. RR scheduler must be the fastest one. Usually the optimistic approach gives us the fastest solution: we're optimistic in assumption that there are not too many non-idempotent request or there are enough server connections. So we can save two CMPXCHG instructions on getting and putting the saved connection in quick path.

tempesta_fw/sched/tfw_sched_rr.c
```diff
@@ -101,32 +101,49 @@ tfw_sched_rr_add_conn(TfwSrvGroup *sg, TfwServer *srv, TfwConnection *conn)
...
- * round-robin manner.
+ * On each subsequent call the function returns the next server in the
+ * group. Parallel connections to the same server are also rotated in
+ * the round-robin manner.
+ *
  * Dead connections and servers w/o live connections are skipped.
+ * Initially, connections with non-idempotent requests are also skipped
+ * in attempt to increase throughput. However, if all live connections
+ * contain non-idempotent requests, then re-run the algorithm and get
+ * the first live connection as it is usually done.
```
krizhanovsky: This is non trivial thing. Please add this to README/Wiki as unique property of RR scheduler.

tempesta_fw/sock.c
```diff
@@ -433,7 +433,7 @@ ss_droplink(struct sock *sk)
 int
 __ss_close(struct sock *sk, int flags)
 {
-	if (unlikely(!sk))
+	if (unlikely(!(sk && ss_sock_live(sk))))
```
krizhanovsky: Why can't we close for example half-open or half-closed sockets? I believe we should be able to close such sockets.

tempesta_fw/sock_srv.c
```diff
@@ -281,15 +305,18 @@ tfw_sock_srv_connect_complete(struct sock *sk)
...
-	if (r) {
+	if ((r = tfw_connection_new(conn))) {
 		TFW_ERR("conn_init() hook returned error\n");
 		return r;
 	}
 
 	/* Let schedulers use the connection hereafter. */
 	tfw_connection_revive(conn);
 
+	/* Repair the connection is necessary. */
```
krizhanovsky: "is" -> "if"

tempesta_fw/connection.h
```diff
@@ -77,22 +77,34 @@ enum {
...
+typedef struct tfw_connection_t {
 	SsProto			proto;
 	TfwGState		state;
 	struct list_head	list;
 	struct list_head	msg_queue;
+	struct list_head	nip_queue;				/*srv*/
 	spinlock_t		msg_qlock;
+	unsigned long		flags;					/*srv*/
 	atomic_t		refcnt;
+	atomic_t		qsize;					/*srv*/
```
keshonok: The reason this variable is made atomic is because it's checked in schedulers which run lock-less. The check is performed using `tfw_server_queue_full()` function in `server.h`. I will add a comment regarding this.

krizhanovsky: But we can read any integer atomically. It seems schedulers just need to read the variable, so it still doesn't have to have `atomic_t` type.

tempesta_fw/http.c
```diff
@@ -813,48 +1047,46 @@ tfw_http_req_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
...
-	 * scheduled much faster.
+	 * TODO #593: check whether req->sess->srv_conn is alive. If not,
+	 * then get a new connection for req->sess->srv_conn->peer from
+	 * an appropriate scheduler. That eliminates the long generic
+	 * scheduling work flow. When the first request in a session is
+	 * scheduled by the generic logic, TfwSession->srv_conn must be
+	 * initialized to point at the appropriate TfwConnection, so that
+	 * all subsequent session hits are scheduled much faster.
 	 */
 	srv_conn = tfw_sched_get_srv_conn((TfwMsg *)req);
```
krizhanovsky: Please consider this [article about QoS](https://www.igvita.com/2008/05/13/load-balancing-qos-with-haproxy/). It has sense to add new configuration option for server group which specifies the queue size. If all the server connections are busy (`tfw_server_queue_full()` returns true) we should put a request to the new queue of configured size. If `srv_conn->qsize` decrements, then we can schedule next request from head of the new queue. If the new queue is full, then just do as currently - send error response to a client (however, it should be 50x response instead of 40x response.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -145,18 +145,19 @@ tfw_sched_hash_get_srv_conn(TfwMsg *msg, TfwSrvGroup *sg)
 	msg_hash = tfw_http_req_key_calc((TfwHttpReq *)msg);
 	for (tries = 0; tries < __HLIST_SZ(TFW_SG_MAX_CONN); ++tries) {
 		for (ch = sg->sched_data; ch->conn; ++ch) {
+			if (unlikely(tfw_connection_restricted(ch->conn))
+			    || unlikely(tfw_server_queue_full(ch->conn))
+			    || unlikely(!tfw_connection_live(ch->conn)))
+				continue;
```
keshonok: The presence of non-idempotent requests in a forwarding queue does not prohibit sending though the connection. It's just it may be a little while before this particular message is sent out, as Tempesta has to wait for responses to non-idempotent requests in the queue before forwarding more requests.

The goal for a hash scheduler is to always send over the same hash-bound connection. So how do you suggest we skip connections with non-idempotent requests in them? Do you want the algorithm to "wait out" looping in hopes that the resulting connection gets available?

tempesta_fw/sock.c
keshonok: The goal here is to avoid calling `close()` on opposite courses. One is a close() that comes from Tempesta, and the other is a close() that comes from the other side of the connection. Both would call `connection_drop()`, and that should not be happening. So if the socket that we want to close is not in established state, then it must be that it's being closed already (apparently on a close request from the other side).

tempesta_fw/http.c
```diff
@@ -382,6 +370,396 @@ tfw_http_send_504(TfwHttpReq *req)
...
+		tfw_http_req_nonidemp_delist(srv_conn, req);
+		list_del_init(&req->msg.fwd_list);
+		atomic_dec(&srv_conn->qsize);
+		spin_unlock(&srv_conn->msg_qlock);
+		TFW_DBG2("%s: Forwarding error: conn=[%p] req=[%p]\n",
+			 __func__, srv_conn, req);
+		tfw_http_send_500(req);
+		TFW_INC_STAT_BH(clnt.msgs_otherr);
+		return;
+	}
```
keshonok: Yes, that's correct, and I did see that. The logic here was to have a special case for just one request which is not a rare case. That would avoid the extra logic related to sending a bunch of requests, like loop set up, etc, etc. Do you think this special case is unnecessary?

tempesta_fw/http.c
```diff
@@ -777,6 +1288,78 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+	list_for_each_entry_safe(req, tmp, &out_queue, msg.seq_list) {
+		list_del_init(&req->msg.seq_list);
+		resp = (TfwHttpResp *)req->resp;
+		/*
+		 * If the client connection is dead, then discard all
+		 * @req and @resp in the @out_queue. Remaining requests
+		 * from the client in the @seq_queue will be handled at
+		 * the time the client connection is released.
+		 */
+		if (!tfw_connection_live(cli_conn)) {
```
keshonok: Indeed. Thanks. Still, a client connection may become dead before/while we attempt to forward a bunch of responses to the connection. The only check on whether the connection is alive is made on the socket level in `ss_do_send()` which is the time the SKB are about to be submitted to Linux TCP/IP stack. However before that happens, the request's data is added to the work queue, the work is scheduled on another CPU, and then data is processed in a separate softirq thread. All of that unnecessarily takes resources. Is that okay? May be it would be useful to use the same `tfw_connection_put_to_death()` on client connections as well?

tempesta_fw/sched/tfw_sched_hash.c
ikoveshnikov: Yes, i got it, you right. Speeding up by skipping connections with non-idempotent requests in them is not a priority for hash scheduler. Didn't think in that way.

tempesta_fw/http.c
```diff
@@ -382,6 +370,396 @@ tfw_http_send_504(TfwHttpReq *req)
...
+ * Tell if the server connection's forwarding queue is on hold.
+ * It's on hold it the request that was sent last was non-idempotent.
+ */
+static inline bool
+tfw_http_conn_on_hold(TfwConnection *srv_conn)
+{
+	TfwHttpReq *req = (TfwHttpReq *)srv_conn->msg_sent;
+
+	BUG_ON(!(TFW_CONN_TYPE(srv_conn) & Conn_Srv));
+	return (req && tfw_http_req_is_nonidempotent(req));
```
keshonok: This particular function specifically checks if it's allowed to continue forwarding. So the checks are not equivalent or duplicate.

The presence of non-idempotent requests in the connection does not mean we have to stop forwarding. Those requests may be somewhere farther in the queue, and the forwarding may continue until one of those is sent out. When a non-idempotent requests is sent out (`conn->req_sent` is non-idempotent), then forwarding must be paused until a response is received to that request.

tempesta_fw/sock_srv.c
```diff
@@ -366,37 +359,59 @@ static const SsHooks tfw_sock_srv_ss_hooks = {
...
-	return ss_close_sync(conn->sk, true);
+	/*
+	 * If the connection is closed already, then simply release its
+	 * resources. Otherwise, use synchronous closing to ensure that
+	 * the job is enqueued.
+	 */
+	if (atomic_read(&conn->refcnt) == TFW_CONN_DEATHCNT)
+		tfw_connection_release(conn);
+	else
+		ret = ss_close_sync(conn->sk, true);
```
krizhanovsky: Yeah, looks good for me.

etc/tempesta_fw.conf
```diff
@@ -4,25 +4,33 @@
 
 # TAG: sched.
 #
-# Specifies the scheduler used to distribute load among servers
+# Specifies the scheduler used to distribute the load among servers within
+# a group.
```
krizhanovsky: Space at the end of the line.

tempesta_fw/connection.h
```diff
@@ -54,63 +54,149 @@ enum {
...
  * @list	- member in the list of connections with @peer;
- * @msg_queue	- queue of messages to be sent over the connection;
- * @msg_qlock	- lock for accessing @msg_queue;
  * @refcnt	- number of users of the connection structure instance;
  * @timer	- The keep-alive/retry timer for the connection;
  * @msg		- message that is currently being processed;
  * @peer	- TfwClient or TfwServer handler;
  * @sk		- an appropriate sock handler;
+ * @destructor	- called when a connection is destroyed;
+ * @forward	- called when a request is forwarded to server;
```
krizhanovsky: There is no `@forward` member in the definition.

tempesta_fw/http.c
```diff
@@ -383,18 +369,636 @@ tfw_http_send_504(TfwHttpReq *req, const char *reason)
...
+__tfw_http_req_nip_delist(TfwSrvConn *srv_conn, TfwHttpReq *req)
+{
+	BUG_ON(list_empty(&req->nip_list));
+	list_del_init(&req->nip_list);
+	if (list_empty(&srv_conn->nip_queue))
+		clear_bit(TFW_CONN_B_HASNIP, &srv_conn->flags);
+}
+
+/*
+ * Put @req on the list of non-idempotent requests in @srv_conn.
```
krizhanovsky: Space at the end of the line

tempesta_fw/server.h
```diff
@@ -105,18 +112,37 @@ struct tfw_scheduler_t {
...
+
+/*
+ * max_recns can be the maximum value for the data type to mean
+ * the unlimited number of attempts, which is the value that should
+ * never be reached. UINT_MAX seconds is more than 136 years. It's
+ * safe to assume that it's not reached in a single run of Tempesta.
+ */
+static inline bool
+tfw_srv_conn_need_resched(TfwSrvConn *srv_conn)
+{
```
krizhanovsky: Many spaces at the end of line

tempesta_fw/apm.c
```diff
@@ -528,14 +528,14 @@ static inline void
 __tfw_apm_state_next(TfwPcntRanges *rng, TfwApmRBEState *st)
 {
 	int i = st->i, r, b;
-	unsigned short rtime;
+	unsigned short rtt;
```
keshonok: Yes, indeed it's "round trip time".

tempesta_fw/connection.h
```diff
@@ -197,15 +346,25 @@ tfw_connection_put(TfwConnection *conn)
...
+static inline void
+tfw_cli_conn_put(TfwCliConn *cli_conn)
+{
+	tfw_connection_put((TfwConn *)cli_conn);
+}
+static inline void
+tfw_srv_conn_put(TfwSrvConn *srv_conn)
+{
+	tfw_connection_put((TfwConn *)srv_conn);
+}
```
keshonok: The use of inline functions is intentional as these provide type checking. I consider this feature important as these conversions from one type of connection to another type may yield a bug that would be very difficult to discover. Macros do not provide that feature, otherwise I would agree with you. Would you still like me to replace these with macros?

Also, the missing empty lines mentioned in the comments above are also intentional. These one-line functions are visibly and visually grouped together as they are basically the same. Would you suggest a different solution to this type of grouping? Or would you just like me break this type of grouping and separate all functions with an empty line?

tempesta_fw/sched/tfw_sched_rr.c
```diff
@@ -90,43 +90,67 @@ tfw_sched_rr_add_conn(TfwSrvGroup *sg, TfwServer *srv, TfwConnection *conn)
...
 	TfwRrSrv *srv_cl;
+	TfwSrvConn *srv_conn;
 
 	BUG_ON(!sl);
-
+rerun:
 	for (s = 0; s < sl->srv_n; ++s) {
-		i = atomic64_inc_return(&sl->rr_counter) % sl->srv_n;
-		srv_cl = &sl->srvs[i];
+		idxval = atomic64_inc_return(&sl->rr_counter);
```
keshonok: `i` was an index. `idxval` is NOT an index. It's a value that is used to calculate the index, hence the name.

`i` is not used for any other purpose in this function. The compiler will do the right thing either way, but the intention was to make the code more readable than it would be in the variant below:
```C
        for (s = 0; s < sl->srv_n; ++s) {
                i = (unsigned long)atomic64_inc_return(&sl->rr_counter)
                    % sl->srv_n;
        }
```

Perhaps, you'd like this variant more?
```C
        for (s = 0; s < sl->srv_n; ++s) {
                i = (unsigned long)
                    atomic64_inc_return(&sl->rr_counter) % sl->srv_n;
        }
```

tempesta_fw/http.c
```diff
@@ -383,18 +369,636 @@ tfw_http_send_504(TfwHttpReq *req, const char *reason)
...
+{
+	TfwServer *srv = (TfwServer *)srv_conn->peer;
+	TfwHttpReq *req_sent = (TfwHttpReq *)srv_conn->msg_sent;
+
+	if (req_sent && tfw_http_req_is_nip(req_sent)
+	    && likely(!(srv->sg->flags & TFW_SRV_RETRY_NIP)))
+	{
+		BUG_ON(list_empty(&req_sent->nip_list));
+		srv_conn->msg_sent =
+			/* list_is_first(&req_sent->fwd_list, fwd_queue); */
```
keshonok: There's `list_is_last()` function in the Linux kernel sources, but there's no `list_is_first()`. This is basically an implementation of `list_is_first()`, and so I tried to explain that. Would you like me to remove this comment?

tempesta_fw/http.c
```diff
@@ -383,18 +369,636 @@ tfw_http_send_504(TfwHttpReq *req, const char *reason)
...
+	     req = tmp, tmp = list_next_entry(tmp, fwd_list))
+	{
+		if (tfw_http_req_evict_timeout(srv_conn, srv, req, equeue))
+			continue;
+		if (tfw_http_req_evict_retries(srv_conn, srv, req, equeue))
+			continue;
+		if (!tfw_http_req_fwd_send(srv_conn, srv, req, equeue))
+			continue;
+		req_resent = req;
+		if (unlikely(first))
```
keshonok: How do you mean - move `first` to `for()` loop's condition? I am not sure I understand.

tempesta_fw/http.c
```diff
@@ -796,122 +1501,272 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+{
+	TfwHttpReq *req, *tmp;
+
+	list_for_each_entry_safe(req, tmp, ret_queue, msg.seq_list) {
+		BUG_ON(!req->resp);
+		if (tfw_cli_conn_send(cli_conn, (TfwMsg *)req->resp)) {
+			ss_close_sync(cli_conn->sk, true);
+			return;
+		}
+		__tfw_http_resp_pair_free(req);
```
keshonok: I am not sure I understand. The response (and the request) is not needed after it's sent out. If for some reason a special flag is set that tells that the SKB data is needed, then the SKBs are copied. In that case the response and it's SKB data can be freed. If SKB data is not needed (the usual case), then the full chain of SKBs is simply detached from the response. When the response is freed there's no SKBs to free.

tempesta_fw/http.c
```diff
@@ -796,122 +1501,272 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
...
+	 * Doing ss_close_sync() on client connection's socket is safe
+	 * as long as @req that holds a reference to the connection is
+	 * not freed.
+	 */
+	spin_lock(&cli_conn->seq_qlock);
+	if (unlikely(list_empty(seq_queue))) {
+		spin_unlock(&cli_conn->seq_qlock);
+		TFW_DBG2("%s: The client's request missing: conn=[%p]\n",
+			 __func__, cli_conn);
+		ss_close_sync(cli_conn->sk, true);
```
keshonok: There's nothing wrong with queueing. However, without the SYNC flag queueing is NOT guaranteed. We have to close the connection at this point as the pairing may get broken.

With `ss_close_sync()` the connection may get closed either right away, or be queued for a later execution. Either way, closing is guaranteed. If it's queued, then scheduled earlier responses will be sent out before the connection is closed, but that's will not happen if the connection is closed right away in the same thread.

Note that this code may (and does, in fact) get executed multiple times, which causes multiple calls to close the connection. That's why there's a check for a duplicate code in `ss_send()` in the code path that gets executed immediately in the function's calling thread. Is there a better way?

tempesta_fw/sock.c
```diff
@@ -552,10 +552,15 @@ __ss_close(struct sock *sk, int flags)
...
-	 * liveness.
+	 * The socket is owned by current CPU, so there's no need to check
+	 * if it's live. However, in some cases this may be called multiple
+	 * times on the same socket. Do it only once for the socket.
 	 */
 	bh_lock_sock(sk);
+	if (unlikely(!ss_sock_live(sk))) {
+		bh_unlock_sock(sk);
+		return SS_OK;
+	}
```
keshonok: No, this is slightly different, and it's in a different place than in that patch. The comment above this code block explains why that was introduced. Also, please see [this comment in http.c](https://github.com/tempesta-tech/tempesta/pull/660/files#r101928722).

tempesta_fw/connection.h
krizhanovsky: How many bugs concerning type safety can you remember? I'd say that typically we have errors in algorithms (e.g. our traditional sockets synchronization), sometimes use after free or other broken pointers. However, I probably can remember only a few bugs with invalid types and typically they are easily fixed somewhere around fix compilation of the new code. Type safety looks like

          if (NULL == a)
               ....

-- the code become less readable, but it doesn't solve real bugs which take a lot of effort to fix. Meantime, C++ type safety takes enormous effort from the library developers and common programmers, but I never felt issues with unsafe typization in C...

Also making the single-line functions and violate the coding style because the function aren't "real" function, but actually type verifiers, isn't a good way I think. Thus I vote for macros.

tempesta_fw/http.c
```diff
@@ -383,18 +369,636 @@ tfw_http_send_504(TfwHttpReq *req, const char *reason)
...
+	return tfw_http_conn_resend(srv_conn, false, equeue);
+}
+
+/*
+ * Handle the complete re-forwarding of requests in a server connection
+ * that is being repaired, after the first request had been re-forwarded.
+ * The connection is not scheduled until all requests in it are re-sent.
+ */
+static void
+__tfw_http_conn_fwd_repair(TfwSrvConn *srv_conn, struct list_head *equeue)
```
keshonok: I tried to do that, but it lead to really, really long function names. If the HTTP client and server logic is separated in the future, then it will be clear which function is what simply by the location of a function. Only external functions will need to be named explicitly.

I would leave the names as they are for now. What do you think?

tempesta_fw/http.c
krizhanovsky: Maybe just say that this is implementation of `list_is_first()`?.. Now it looks like commented out line of code

tempesta_fw/http.c
krizhanovsky: Yeah, my comment is wrong. Never mind :)

tempesta_fw/http.c
krizhanovsky: Yeah, if you've already worked on the names, then I don't see any sense to put significant effort to renaming for now. Let's leave it for further code reworkings.

tempesta_fw/http.c
krizhanovsky: Yes, you're right.

tempesta_fw/sched/tfw_sched_rr.c
krizhanovsky: Well, I'd say that `i` is usually anything beginning with "i", like "index", "iterator", "idxval" and so on. Just a short name when the code and meaning of the variable is clear. If you feel much more comfortable with `idxval` here, I wouldn't mind.

tempesta_fw/http.c
```diff
@@ -390,6 +389,26 @@ tfw_http_send_504(TfwHttpReq *req, const char *source, const char *reason)
...
 /*
+ * SKB data is needed for calculation of a cache key from fields of
+ * a request. It's also needed when a request may need to be re-sent.
+ * In all other cases it can just be passed to the network layer.
+ */
+static inline void
+tfw_http_req_init_ss_flags(TfwSrvConn *srv_conn, TfwHttpReq *req)
+{
+	TfwSrvGroup *sg = ((TfwServer *)(srv_conn->peer))->sg;
+	if (tfw_cache_msg_cacheable(req) || (req->retries < sg->max_refwd))
```
krizhanovsky: I think there is no need to optimize this rare case and we can just always copy skb. It's bad, but we'll address this in #391 and #488.

krizhanovsky:
Good to merge

keshonok:
Implemented in #660 (merge commit c40924b).

Issue #658, opened 2016-11-29
Client friendly requests blocking


krizhanovsky:
Currently HTTP parser can return TFW_BLOCK in many cases. Meantime when TFW_BLOCK is returned from `tfw_http_msg_process()`, then the message is just silently dropped. So if a client uses poor browser, then they typically just see hang pages loading. System administrator should be able to specify the behaviour: return HTTP error message to a client of just silently block malformed and malicious message (requests and responses). This could be done by a new configuration option like

        block_action <drop|reply> [nolog]

`drop` is just current behaviour while `reply` sends error response to a client. Optional argument `nolog` requires don't do logging the reason for blocking. By default Tempesta FW must write to the log why particular message was blocked. I assume it could be done by replacing statements like 

        return TFW_BLOCK;

by

        return tfw_block("Content-Length with Transfer-Encoding in the same request");

, where `tfw_block()` checks configuration variable and does the logging.

krizhanovsky:
There is difference between malicious and malformed requests: we should spend as little as possible resources (i.e. silently block) for the first ones and provide reasonable user-friendly responses for the second ones. So `TFW_DROP` should be introduced to handle malformed requests while `TFW_BLOCK` should be used to filter our malicious requests. Also there should possibility to configure actions for the two different request types in different ways.

krizhanovsky:
Currently we just send HTTP error codes with empty body, so a user agent (tested with Firefox at least) just shows blank page. Thus, some predefined HTML body must be sent as well. Further we should allow a user to configure the error page body as it's done in all Web servers.

krizhanovsky:
The task seems somewhat more complicated than it was expected initially. The problem is if `connection_recv()` returns `TFW_BLOCK` then `ss_tcp_data_ready()` calls `ss_linkerror()` to immediately close the client connection. (Note that `ss_tcp_state_change()` doesn't call `ss_linkerror()` if `ss_tcp_process_data()` returns bad code and this must be fixed.) Meantime, if a client has pipelined many HTTP requests, then they're can be on the fly while we're experiencing the blocking logic, so the connection can be closed before transmission of responses for previous good requests and it's unclear how to send the HTTP error response.

[...]

PR #662, opened 2016-12-11
Resistance to misbehaving upstream server

krizhanovsky:
Workaround for #651: introduce TFW_HTTP_MSG_SENT flag such that unsent requests aren't paired with inavalid responses; drop connections if unpaired responses
detected.

Fix tfw_connection_unlink_to_sk(): do the actions in reverse order as they are
done in tfw_connection_link_to_sk().

Fix multiple memory leaks in error paths in HTTP request/response processing
(HTTP messages weren't freed).

Partial #658: print warning on parser failures, print reasons for error HTTP
responses.

Optimizations: don't copy request skbs if no caching introduced; don't put
work for known dead socket to work queue (races arepossible, but
we just do the extra work in worst case)

Many cleanups.

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -187,7 +187,13 @@ tfw_http_prep_302(TfwHttpMsg *resp, TfwHttpReq *req, TfwStr *cookie)
 static inline void
 __init_req_ss_flags(TfwHttpReq *req)
 {
-	((TfwMsg *)req)->ss_flags |= SS_F_KEEP_SKB;
+	/*
+	 * We need skb data only for calculating cache key by the request
+	 * fields. In all other cases we can just pass skb data to network
+	 * layer.
+	 */
+	if (tfw_cache_msg_cacheable(req))
+		((TfwMsg *)req)->ss_flags |= SS_F_KEEP_SKB;
 }
 
 static inline void
@@ -265,7 +271,7 @@ tfw_http_send_200(TfwHttpReq *req)
  * HTTP 403 response: Access is forbidden.
  */
 int
-tfw_http_send_403(TfwHttpReq *req)
+tfw_http_send_403(TfwHttpReq *req, const char *reason)
 {
 	TfwStr rh = {
 		.ptr = (TfwStr []){
@@ -278,7 +284,7 @@ tfw_http_send_403(TfwHttpReq *req)
 		.flags = 4 << TFW_STR_CN_SHIFT
 	};
 
-	TFW_DBG("Send HTTP 404 response to the client\n");
+	TFW_DBG("Send HTTP 404 response: %s\n", reason);
 
 	return tfw_http_send_resp(req, &rh, __TFW_STR_CH(&rh, 1));
 }
@@ -289,7 +295,7 @@ tfw_http_send_403(TfwHttpReq *req)
  * HTTP 404 response: Tempesta is unable to find the requested data.
  */
 int
-tfw_http_send_404(TfwHttpReq *req)
+tfw_http_send_404(TfwHttpReq *req, const char *reason)
 {
 	TfwStr rh = {
 		.ptr = (TfwStr []){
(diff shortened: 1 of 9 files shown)


tempesta_fw/http.c
```diff
@@ -877,16 +887,19 @@ tfw_http_req_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
 		spin_unlock(&srv_conn->msg_qlock);
 		goto send_500;
 	}
+	mb();
```
keshonok: What exactly this memory barrier is for?

tempesta_fw/http.c
```diff
@@ -1153,28 +1173,32 @@ tfw_http_resp_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
...
-	TfwMsg *msg;
+	TfwHttpReq *req;
 	TfwConnection *conn = hmresp->conn;
 
 	spin_lock(&conn->msg_qlock);
-	if (unlikely(list_empty(&conn->msg_queue))) {
+
+	req = list_first_entry_or_null(&conn->msg_queue, TfwHttpReq,
+				       msg.msg_list);
+	if (unlikely(!req || !(req->flags & TFW_HTTP_MSG_SENT))) {
```
keshonok: I'd say that a (write) memory barrier is justified before checking for `TFW_HTTP_MSG_SENT` flag. That would ensure that all writes are committed before the check is performed.

tempesta_fw/http.c
```diff
@@ -956,13 +970,15 @@ tfw_http_req_process(TfwConnection *conn, struct sk_buff *skb, unsigned int off)
 		case TFW_BLOCK:
 			TFW_DBG2("Block invalid HTTP request\n");
 			TFW_INC_STAT_BH(clnt.msgs_parserr);
+			tfw_http_conn_msg_free((TfwHttpMsg *)req);
```
keshonok: This is not essential, but the rest of the code has the following sequence is case of an error:
* send an error message if necessary;
* free the erroneous message;
* increment statistics counters;

New changes make this a different sequence which is contrasting.

tempesta_fw/http.c
```diff
@@ -1153,28 +1173,32 @@ tfw_http_resp_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
  * Request messages that were forwarded to a backend server are added
  * to and kept in @msg_queue of the connection @conn for that server.
  * If a paired request is not found, then the response is deleted.
+ *
+ * If thre isn't paired client request, then it seems upsream server is
```
keshonok: It appears that the comment is garbled. Perhaps, "*there isn't* a paired client request"?
Or, perhaps: `If a paired client request is missing, then...`

tempesta_fw/http.c
```diff
@@ -1211,10 +1235,11 @@ tfw_http_resp_gfsm(TfwHttpMsg *hmresp, struct sk_buff *skb, unsigned int off)
 	req = tfw_http_popreq(hmresp);
 	if (unlikely(!req)) {
 		TFW_INC_STAT_BH(serv.msgs_filtout);
+		tfw_http_conn_msg_free(hmresp);
 		return TFW_STOP;
```
keshonok: It appears that in fact this should be `TFW_BLOCK` as this is about a server connection and the paired request is missing.

tempesta_fw/http.h
```diff
@@ -234,6 +234,7 @@ typedef struct {
 #define TFW_HTTP_HAS_HDR_DATE		0x020000	/* Has Date: header */
 /* It is stale, but pass with a warning */
 #define TFW_HTTP_RESP_STALE		0x040000
+#define TFW_HTTP_MSG_SENT		0x080000
```
keshonok: This is the section for response flags. I believe that it should be moved to section for request flags (and renamed to `TFW_HTTP_REQ_SENT).

tempesta_fw/http.c
krizhanovsky: It guarantees that the flag is set exactly after finishing `tfw_connection_send()` which can be inlined. I didn't use atomic operation for the flag manipulations, so I used explicit memory barrier.

tempesta_fw/http.c
krizhanovsky: Yeah, good note. It seems I really forgot paired barrier. However, there are read operations only, so I believe read barrier is the right one here.

Issue #593, opened 2016-07-31
Sticky sessions scheduler


krizhanovsky:
Depends on #235.

Actually foundation for [Sticky cookie](https://github.com/tempesta-tech/tempesta/issues/29) is application layer client identification, i.e. `hash(client IP, User-Agent)` identifies a 'virtual' client ('virtual' because there could be different clients behind the same proxy as well as couple of different browsers on the same device). 

The identifier is used for scheduling the client requests to appropriate server (see [Nginx sticky module](https://github.com/lusis/nginx-sticky-module) and [HAProxy sticky sessions](http://blog.haproxy.com/2012/03/29/load-balancing-affinity-persistence-sticky-sessions-what-you-need-to-know/)) or for security issues. The second use case (HMAC key in particular) is the subject for #235. While the first application, sticky sessions scheduling, wasn't developed at all.

So hash scheduler must be extended such that it can use Tempesta sticky cookie value to decide which upstream server the message should be scheduled to. This will guarantee that HTTP session will be always processed by the same upstream server. No one current scheduler can guarantee HTTP sessions persistence, so Tempesta can break user experience.

Overall requests scheduling must have important property: if Tempesta receives first request within a new HTTP session, it must schedule the request, and correspondingly the whole session, to upstream using some other scheduling algorithm like round-robin or one of [advanced load balancing algorithms](https://github.com/tempesta-tech/tempesta/issues/565). Actually hash scheduler using URI and Host values is also suitable for initial scheduling. See comment in `tfw_http_req_cache_cb()`.

[...]

PR #666, opened 2016-12-20
Sticky Sessions implementation

ikoveshnikov:
Fix issue #593 

Schedule HTTP request to the same servers for session preservation

### DIFF
--- a/tempesta_fw/http.h
+++ b/tempesta_fw/http.h
@@ -27,6 +27,7 @@
 #include "gfsm.h"
 #include "msg.h"
 #include "str.h"
+#include "server.h"
 #include "vhost.h"
 
 /**
@@ -244,7 +245,8 @@ typedef struct {
  * @users	- the session use counter;
  * @ts		- timestamp for the client's session;
  * @expire	- expiration time for the session;
- * @srv_conn	- upstream server connection servicing the session;
+ * @conns	- upstream server connections servicing the session;
+ * @conns_lock	- protects @conns
  */
 typedef struct {
 	unsigned char		hmac[SHA1_DIGEST_SIZE];
@@ -252,7 +254,8 @@ typedef struct {
 	atomic_t		users;
 	unsigned long		ts;
 	unsigned long		expires;
-	TfwConnection		*srv_conn;
+	struct list_head	conns;
+	rwlock_t		conns_lock;
 } TfwHttpSess;
 
 /**
@@ -412,5 +415,10 @@ void tfw_msg_add_data(void *handle, TfwMsg *msg, char *data, size_t len);
 int tfw_http_sess_obtain(TfwHttpReq *req);
 int tfw_http_sess_resp_process(TfwHttpResp *resp, TfwHttpReq *req);
 void tfw_http_sess_put(TfwHttpSess *sess);
+extern bool tfw_cfg_sticky_sessions;
+extern bool tfw_cfg_sticky_sessions_failover;
+TfwConnection *tfw_http_sess_get_conn(TfwHttpReq *req, TfwSrvGroup *sg);
+int tfw_http_sess_save_conn(TfwHttpReq *req, TfwSrvGroup *sg,
+			    TfwConnection *conn);
 
 #endif /* __TFW_HTTP_H__ */
--- a/tempesta_fw/http_sess.c
+++ b/tempesta_fw/http_sess.c
@@ -82,9 +82,10 @@ typedef struct {
   *
   * @list	- list of connections;
   * @sg		- target server group;
-  * @conn	- connection can be established to server group differ from @sg,
-  *		in this case connection is established to server from back-up
-  *		group.
+  * @conn	- last used connection;
+  *
+  * Connection may be established to server in backup group which differs from
+  * @sg.
   */
 typedef struct {
 	struct list_head	list;
@@ -533,10 +534,10 @@ tfw_http_sess_put(TfwHttpSess *sess)
 	 * from the hash table.
 	 */
 	if (atomic_dec_and_test(&sess->users)) {
-		TfwHttpSessConns *conn, *tmp;
-		list_for_each_entry_safe(conn, tmp, &sess->conns, list) {
-			list_del(&conn->list);
-			kmem_cache_free(sess_conn_cache, sess);
+		TfwHttpSessConns *sess_conn, *tmp;
+		list_for_each_entry_safe(sess_conn, tmp, &sess->conns, list) {
+			list_del(&sess_conn->list);
+			kmem_cache_free(sess_conn_cache, sess_conn);
 		}
 		kmem_cache_free(sess_cache, sess);
 	}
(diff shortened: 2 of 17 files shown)


tempesta_fw/http_sess.c
```diff
@@ -593,6 +628,76 @@ tfw_http_sess_obtain(TfwHttpReq *req)
...
+
+	list_for_each_entry(s_conn, &sess->conns, list) {
+		if (s_conn->sg == sg) {
+			conn = s_conn->conn;
+			break;
+		}
+		else if (s_conn->conn &&
+			 (((TfwServer *)s_conn->conn->peer)->sg == sg))
+		{
+			backup_conn = s_conn->conn;
```
keshonok: But the search still continues? I am not sure I understand the logic. I would appreciate an explanation.

ikoveshnikov: If previously scheduling to `sg` failed, session could be bound with backup server connection instead. So connection used last time to schedule request to group `sg` may belong to server in other group. We need to keep scheduling session for backup server even when primary server group is back online.

Actually `backup_conn` is not used to speed up search. When the session is scheduled to group `sg` in the first time, we can reuse existing connection to `sg`, if group `sg` is a backup group for some other group

tempesta_fw/t/unit/test_http_sticky.c
```diff
@@ -515,6 +516,302 @@ TEST(http_sticky, req_have_cookie_enforce)
...
+	EXPECT_NOT_NULL((conn_m));
+	EXPECT_EQ(((TfwServer *)conn_m->peer)->sg, sg_m);
+
+	test_conn_release_all(sg_m);
+	test_conn_release_all(sg_b);
+	test_sg_release_all();
+	sticky_sess_deny_sched();
+}
+
+TEST(http_sticky_sched, resuse_conn)
```
keshonok: reuse_conn

tempesta_fw/t/unit/test_http_sticky.c
```diff
@@ -566,8 +863,26 @@ TEST_SUITE(http_sticky)
...
+
+	/* Disable failovering for next tests */
+	sticky_sess_deny_sched();
+	BUG_ON(!tfw_cfg_sticky_sessions);
+
+	TEST_RUN(http_sticky_sched, add_conns);
+	TEST_RUN(http_sticky_sched, sched_one_sg);
+	TEST_RUN(http_sticky_sched, sched_one_sg_offline_srv);
+	TEST_RUN(http_sticky_sched, sched_backup_sg);
+	TEST_RUN(http_sticky_sched, resuse_conn);
```
keshonok: reuse

tempesta_fw/server.h
```diff
@@ -42,12 +42,14 @@ typedef struct tfw_scheduler_t TfwScheduler;
...
  * @list	- member pointer in the list of servers of a server group;
  * @sg		- back-reference to the server group;
+ * @sched_data	- private scheduler data for the server;
  * @apm		- opaque handle for APM stats;
  */
 typedef struct {
 	TFW_PEER_COMMON;
 	struct list_head	list;
 	TfwSrvGroup		*sg;
+	void			*sched_data;
```
ikoveshnikov: We use the same member name `sched_data` in `TfwSrvGroup`. Both have the same meaning and usage, so i think it is a good idea to have the same names for them. Should i rename both to `schref`?

tempesta_fw/sched/tfw_sched_rr.c
```diff
@@ -101,42 +104,73 @@ tfw_sched_rr_add_conn(TfwSrvGroup *sg, TfwServer *srv, TfwConnection *conn)
...
+ * Parallel connections to the server are rotated in the round-robin manner.
+ */
+static TfwConnection *
+tfw_sched_rr_srv_get_conn(TfwMsg *msg, TfwServer *srv)
+{
+	TfwRrSrv *srv_cl = srv->sched_data;
+
+	/* For @srv without connections srv_cl will be NULL */
+	if (!srv_cl)
+		return NULL;
```
ikoveshnikov: I don't see a reason for removing unit tests for schedules about scheduling for a server group without connections. I will add `unlikely` here as a compromise.

tempesta_fw/sched.c
```diff
@@ -39,7 +40,138 @@
...
+		if (!tfw_cfg_sticky_sessions_failover) {
+			return ERR_PTR(-EINVAL);
+		}
+		else {
+			char addr_str[TFW_ADDR_STR_BUF_SIZE] = { 0 };
+
+			tfw_addr_ntop(&srv->addr, addr_str, sizeof(addr_str));
+			TFW_WARN("sched: Unable to reschedule request to the "
+				 "same server %s, schedule from server group\n",
+				 addr_str);
```
ikoveshnikov: This warning is shown not because of scheduler fault, but because target server is unreachable. 
If Tempesta cannot reach backend server it shows warning like this one:
"Cannot establish connection with <server> in 60 tries, keep trying...". These two warning are very close to each other and it is more easy for user to understand the reason of high warnings count.

More over, at that exact moment i cannot tell which schedule group will be used to schedule request. E.g., last time request was scheduled to the server from the backup group, but now that server is offline. I will try to find a new server in main group, but the whole group also can be offline, so i will try to schedule to backup group by calling the same function again.

tempesta_fw/http.h
```diff
@@ -412,5 +415,10 @@ void tfw_msg_add_data(void *handle, TfwMsg *msg, char *data, size_t len);
 int tfw_http_sess_obtain(TfwHttpReq *req);
 int tfw_http_sess_resp_process(TfwHttpResp *resp, TfwHttpReq *req);
 void tfw_http_sess_put(TfwHttpSess *sess);
+extern bool tfw_cfg_sticky_sessions;
+extern bool tfw_cfg_sticky_sessions_failover;
+TfwConnection *tfw_http_sess_get_conn(TfwHttpReq *req, TfwSrvGroup *sg);
+int tfw_http_sess_save_conn(TfwHttpReq *req, TfwSrvGroup *sg,
+			    TfwConnection *conn);
```
ikoveshnikov: As for me, it is "just good enough". Moving all the session related types and functions lead to circular dependencies: `TfwHttpReq` require `TfwHttpSess`, but all the session related functions require `TfwHttpReq`. Both approaches using forward declarations or moving only part of session related staff to separate header don't look good for me. How do you think?

tempesta_fw/sched.c
keshonok: No, that's not quite what I meant. I meant to be a bit more specific about which scheduler and group *were* used for scheduling to the server, not which group will be used.

tempesta_fw/server.h
keshonok: Not necessarily. I just made a suggestion. That's entirely your decision as you're the author of this code.

ikoveshnikov:
Over engineered and does not suit requirements

Issue #76, opened 2015-03-11
Requests scheduling to massive farm of backend servers


krizhanovsky:
Currently `TFW_SCHED_MAX_SERVERS` is defined as 64 backend server at maximum, which is not enough for virtualized environments (virtual hosting or clouds), so it should be extended at least to 64K and all scheduling algorithms also must be updated accordingly to process such number of backend servers.

We still do not expect too many servers per one site group among which client requests are scheduled, but we expect  a lot of independent sites.

**CRUCIAL NOTE:** it's quite atypical to have 64K servers in the same server group. Virtualized environment means many small small sites behind Tempesta FW, i.e. the sense of the issue is HTTP scheduler, which must schedule a request among thousands of server groups. In this case most of server groups can have only one server. Meantime there could be really powerful installations with hundreds upstream servers. Thus 2-tier schedulers (ration, hash etc.) still must have dynamic arrays for connections and servers, but probably we don't need to introduce special data structures able to efficiently handle thousands of servers in the same server group.

vdmit11:
Well, the `TFW_SCHED_MAX_SERVERS` was added for a purpose.
A fixed-size array of servers has certain advantages over a linked list:
 - You can use binary search (although linked list may be transofrmed to a skip list, but this is more complicated).
 - You can allocate per-CPU arrays easily. The small value of `TFW_SCHED_MAX_SERVERS` allows to do that statically. Dynamic per-CPU linked lists are not that easy.
 - All the memory is packed together which is good for caching, hence the performance is better.

Of course, we can re-allocate the array as needed and so on,
but for me it looks easier to have two separate modules for the two cases:
 - A module for small groups of servers that are online most of the time.
 - Another module for large groups of servers that go offline all the time.

These two cases require different implementations and involve different optimizations, so I think we really need  separate modules.

krizhanovsky:
```
    ....to have two separate modules for the two cases:

    A module for small groups of servers that are online most of the time.
    Another module for large groups of servers that go offline all the time.
```

This is not a different logic, this is just a different cases, so they should be treated in the same code base. Probably, you just can allocate array for small server set or use hash table or trees to handle thousands of servers. But the different containers should be processed by the same logic.

Or please make an example of logic which is fundamentally different for the cases.

krizhanovsky:
The system should dynamically establish new connections to busy upstream servers and also dynamically shrink redundant connections.

krizhanovsky:
[...]

PR #670, opened 2016-12-28
Advanced load balancing - ratio scheduler

keshonok:


tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,467 @@
...
+	for (s = 0, n = 0; s < sl->srv_n; ++s) {
+		do {
+			idx = atomic64_inc_return(&sl->counter);
+			cl = rated->list[idx % rated->listsz];
+			if (likely(cl != cl_down))
+				break;
+		} while (++n < rated->listsz);
+
+		if (unlikely(n == rated->listsz))
+			continue;
```
ikoveshnikov: When we proceed to the next server `n` must be set to `0`

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,467 @@
...
+	BUG_ON(!sl);
+
+	rcu_read_lock();
+	rated = rcu_dereference(sl->rated);
+
+	for (s = 0, n = 0; s < sl->srv_n; ++s) {
+		do {
+			idx = atomic64_inc_return(&sl->counter);
+			cl = rated->list[idx % rated->listsz];
+			if (likely(cl != cl_down))
```
ikoveshnikov: I didn't get logic here, especially with `cl_down = cl;`, I would appreciate an explanation.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,467 @@
...
+/**
+ * On each subsequent call the function returns the next available
+ * connection to one of the servers in the group. Connections to a
+ * server are rotated in pure round-robin manner. The server is chosen
+ * according to its current weight. Servers with more weight are chosen
+ * more often than servers with less weight. 
+ * Dead connections and servers w/o live connections are skipped.
+ */
+static TfwConnection *
+tfw_sched_ratio_get_srv_conn(TfwMsg *msg, TfwSrvGroup *sg)
```
ikoveshnikov: You have asked about other ways to implement scheduler, here is my suggestion. Add `int repeats; atomic64_t s_counter;` to `TfwRatioSrv`. Choosing server will be done in this way:
```
for (s = 0; s < sl->srv_n; ++s) {
    i = (uint64_t)atomic64_inc_return(&sl->counter) % sl->srv_n;
    srv_cl = &sl->srvs[i];

    /*  
     * Make less than srv_cl->repeats requests to one server in a row, 
     * and then move to next one 
     */
    uint64_t continue_with_server =  
        (uint64_t)atomic64_inc_return(& srv_cl->s_counter) %  (srv_cl->repeats + 1);
    it (!continue_with_server)
       continue;
    
    /* Get connection for server. */
    ...
}
return NULL;
```
That seems to be naive, but that does the trick. I think with this approach we don't need to have rcu or make `TfwRatioSrv->repeats` atomic, since it updated quite rare.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,467 @@
...
+	sl = kzalloc(sizeof(TfwRatioSrvList), GFP_KERNEL);
+	BUG_ON(!sl);
+
+	for (i = 1; i < TFW_RATIO_EPOOLSZ; ++i)
+		atomic_set(&sl->epool[i].free, 1);
+	rcu_assign_pointer(sl->rated, &sl->epool[0]);
+	sg->sched_data = sl;
+
+	atomic_set(&sl->rearm, 1);
+	setup_timer(&sl->timer, tfw_sched_ratio_tmfn, (unsigned long)sg);
```
krizhanovsky: I think we should setup the timer if there is dynamic scheduling configuration only.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,467 @@
...
+		TFW_DBG3("%s: No available RCU entry\n", __func__);
+		timeout = 1;
+		goto rearm;
+	}
+
+	tfw_sched_fill_rated(sg, ratio, this_rated);
+
+	prev_rated = sl->rated;
+	rcu_assign_pointer(sl->rated, this_rated);
+	call_rcu(&prev_rated->rcu, tfw_sched_rated_put);
```
krizhanovsky: Hm, very interesting... Really, we use faster pool allocations for ratios instead of kmalloc(). But allocations would happen in the timer only, i.e. there shouldn't be too many kmalloc() calls. So why do we need the pool and why can't we just use kmalloc() for the new ratios and kfree() called by `call_rcu()`?

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,467 @@
...
+ * Calculate server ratios based on specific dynamic weights.
+ * Return 1 if no recalculation is required.
+ * Return 0 if the ratios were recalculated.
+ */
+static int
+tfw_sched_ratio_dynamic(TfwSrvGroup *sg, int *ratio)
+{
+	size_t s;
+	TfwRatioSrvList *sl = sg->sched_data;
+	int recalc = 0, weight[sl->srv_n];
```
krizhanovsky: With #76 in mind `sl->srv_n` can be very large, so we can easily overrun the stack.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,467 @@
...
+			if (likely(cl != cl_down))
+				break;
+		} while (++n < rated->listsz);
+
+		if (unlikely(n == rated->listsz))
+			continue;
+
+		for (c = 0; c < cl->conn_n; ++c) {
+			idx = atomic64_inc_return(&cl->counter);
+			srv_conn = cl->conns[idx % cl->conn_n];
```
krizhanovsky: With #76 in mind `conns` can be dynamically allocated rather declared with static size as currently. It seems original RR algorithm is the most straightforward to satisfy #76 requirements, since it basically doesn't scan all the servers/connection as for example hash scheduler.

BTW the function is pretty complex, so a good comments are required describing the complex loops, e.g. the do-while.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,467 @@
...
+/*
+ * The size of scale that is used to range back-end servers.
+ * It's most effective in calculations when it's a power of two. 
+ */
+#define TFW_RATIO_SCALE_64	(1 << 6)
+#if TFW_SG_MAX_SRV > TFW_RATIO_SCALE_64
+#define TFW_RATIO_SCALE		TFW_SG_MAX_SRV
+#else
+#define TFW_RATIO_SCALE		TFW_RATIO_SCALE_64
+#endif
```
krizhanovsky: The constants are useless in sense of #76.

tempesta_fw/apm.c
```diff
@@ -825,31 +833,42 @@ tfw_apm_prcntl_fn(unsigned long fndata)
...
+	rdidx = (unsigned int)atomic_read(&data->stats.rdidx) % 2;	\
+	asent = &data->stats.asent[rdidx];				\
+									\
+	fn_lock(&asent->rwlock);					\
+	memcpy(pstats->val, asent->pstats.val,				\
+	       pstats->psz * sizeof(pstats->val[0]));			\
+	fn_unlock(&asent->rwlock);					\
+	pstats->seq = rdidx;						\
+									\
+	return (seq != rdidx);
```
krizhanovsky: Is there a way to know whether the actual values were changed? E.g there can be 1000 of very busy servers, but after some time they go to some "stable" state, when their response times remain *almost* the same. "Almost" is crucial, because if a server response time changes for exampe for 1/100 or even 1/1000, then there is no sense to trigger upper LB logic about the irrelevant changes.

Meantime, real life farms of 1000 servers are usually always have issues with several servers, so there the real (probably) case is that 99% of servers are in the "stable" state, while the rest of the servers face issues and their weights must be adjusted. In sense of #76 it would be good to be able to catch the exact set of servers with changed pstats.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,467 @@
...
+
+static inline void
+tfw_sched_fill_rated(TfwSrvGroup *sg, int *ratio, TfwRatioRated *rated)
+{
+	size_t s, i, n;
+	TfwRatioSrvList *sl = sg->sched_data;
+
+	for (s = 0, i = 0; s < sl->srv_n; ++s)
+		for (n = 0; n < ratio[s]; ++n, ++i)
+			rated->list[i] = &sl->srvs[s];
```
krizhanovsky: `TfwRatioSrv->list` has size 64 (1 << 6), which is just 2 times larger than `TFW_SG_MAX_SRV`, so it seems the array can be overrun on large ratios.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,467 @@
...
+	size_t s;
+	unsigned long calc[sz], sum = 0, unit;
+
+	for (s = 0; s < sz; ++s) {
+		calc[s] = ((1 << 16) * TFW_RATIO_SCALE) / weight[s];
+		sum += calc[s];
+	}
+	unit = ((1 << 16) * TFW_RATIO_SCALE) / sum;
+	for (s = 0; s < sz; ++s)
+		ratio[s] = (calc[s] * unit) / (1 << 16) ? : 1;
```
krizhanovsky: I implemented the unit test in https://github.com/tempesta-tech/tempesta/commit/b7ccb83723298a4571514629f51b5156415c00ac . Please use the algorithm for the issue implementation.

keshonok:
Replaced by #718.

PR #671, opened 2016-12-28
Initial wiki notes for schedulers

ikoveshnikov:
Here is just initial description of schedulers for wiki.

In my opinion, wiki should contain only those things that do not changed frequently. Copying code comments is waste of paper: one day documentation will be far away from code and will be useless. 
Instead i suggest documenting in wiki the next things:
* Tempesta configuration file. User edit it, so good explanation of options is wanted. Some option can affect each other, and that can be quite complicated.
* Complicated logic of some subsystems/features. That more for developers. Some subsystems like an iceberg: api is pretty and clear, but the devil is in the detail.

That is not final document: there is a lot of references to other subsystem/modules. Just need a feedback, which sections to add, what to cover. Better to see in rich diff format.

Please don't pay a lot of attention to looks, still working on it (:

### DIFF
--- a/doc/scheduling.md
+++ b/doc/scheduling.md
@@ -1,0 +1,180 @@
+# Scheduling in TempestaFW: User Guide
+
+TempestaFW uses various schedulers to distribute load among known servers.
+
+TempestaFW implements two types of schedulers: **group** and **server** schedulers.
+**Group schedulers** are used to classify request and distribute it to corresponding
+server group. **Server schedulers** are used to distribute request to specific
+server in chosen [Server Group]().
+
+
+## Scheduling Requests among Server Groups
+
+
+### HTTP Scheduler
+
+**HTTP Scheduler** inspects headers of an HTTP request and chooses target [Server Group]()
+according to scheduling rules. Scheduling rules are defined by user in TempestaFW
+configuration and map headers patterns to server groups. By default no rules
+are defined. Syntax is:
+```
+sched_http_rules {
+	match <SRV_GROUP> <FIELD> <OP> <ARG> [backup=<BACKUP_SRV_GROUP>];
+	...
+}
+```
+`SRV_GROUP` and `BACKUP_SRV_GROUP` -- the reference to a previously defined
+server groups. `BACKUP_SRV_GROUP` is optional parameter and can be skipped, while
+main `SRV_GROUP` option is mandatory.
+
+If whole `SRV_GROUP` group is offline, request
+will be scheduled to `BACKUP_SRV_GROUP`. Backup server groups are very handy for
+[A/B testing](https://en.wikipedia.org/wiki/A/B_testing) purposes: in that case
+`SRV_GROUP` can represent unstable/test service, and `BACKUP_SRV_GROUP` -- stable
+production service; if test service is under the maintenance users will be
+forwarded to stable service.
+
+`FIELD` -- an HTTP request field, following are supported:
+* **uri** Only a part of URI is looked at that contains the path and the query
+string if any. (e.g. `/abs/path.html?query&key=val#fragment`);
+* **host** The host part from URI in HTTP request line, or the value of `Host`
+header. Host part in URI takes priority over the `Host` header value;
+* **hdr_host** The value of `Host` header;
+* **hdr_conn**  The value of `Connection` header;
+* **hdr_raw** The contents of any other HTTP header field as specified by
+`ARG`. `ARG` must include contents of an HTTP header starting with the header
+field name. The `suffix` `OP` is not supported for this `FIELD`. Processing
+of `hdr_raw` may be slow because it requires walking over all headers of an
+HTTP request.
+
+`OP` -- a string comparison operator:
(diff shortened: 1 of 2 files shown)


doc/scheduling.md
```diff
@@ -0,0 +1,180 @@
+# Scheduling in TempestaFW: User Guide
```
krizhanovsky: Does Markdown provide support for contents? It'd be nice to have contents at begin of the file.

doc/scheduling.md
```diff
@@ -0,0 +1,180 @@
...
+## Scheduling Requests inside Server Groups
+
+**Server Schedulers** are bound to exact server groups, see
+[Setting scheduler]() section in [Server Group]() description.
+
+User can override default server scheduler for all groups using `sched`
+configuration option:
+```
+sched <SCHED_NAME>;
+```
```
krizhanovsky: It's unclear where the option should be specified. It'd be useful to provide a configuration example where some scheduler is specified.

doc/scheduling.md
```diff
@@ -0,0 +1,180 @@
...
+
+
+### Hash Scheduler
+
+Chooses server to schedule request to by hashed key value. Key is built using
+_uri_, _request method_ and _host header_ of the request. In most cases load will
+be distributed among all the servers in group, but situations when single server
+pulls all the load are also possible. Although it is quite improbable, such
+condition is quite stable: it cannot be fixed by adding/removing servers and
+restarting Tempesta FW.
```
krizhanovsky: Rendezvous hashing should be mentioned here - this is important feature for end users.

doc/scheduling.md
```diff
@@ -0,0 +1,180 @@
...
+Hash scheduler have an advantages though. It schedules requests of same resources
+to the same servers. That can reduce load of backend servers.
+
+Scheduler name for use in configuration file: `hash`.
+
+
+## Scheduling HTTP sessions to the same server
+
+None of the schedulers is responsible for forwarding requests from the same HTTP
+session to the same server. [Sticky Sessions]() option should be used instead.
```
krizhanovsky: This is Wiki, so it should be friendly to beginners, so please describe when a user should choose the scheduler, when it's important and useful. Also it's good to say in RR section that RR is very fast and basically provides good enough load distribution. Hash section already has "Hash scheduler have an advantages though..."

doc/scheduling.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Troubleshooting
+
+- TempestaFW overloads some of my servers while others in the same group are
+almost unused.
+
+Hash Server scheduler does not guarantee fair load of backend servers and
+can cause situations, when some servers pull most of the load. Round-robin
+scheduler should be used instead. Other configuration options can also affect
+schedulers behaviour, see [Effects caused to schedulers](#Effects caused to schedulers by other configuration options)
```
krizhanovsky: This case is too generic and the suggestion can be false. What if each Nth request is extremely heavy? In this case RR will do pretty bad scheduling. Troubleshooting section must be concentrated on solution of exact small issues, rather than suggest how to fix generally non-working installation.

doc/scheduling.md
```diff
@@ -0,0 +1,180 @@
...
+
+- Requests are still forwarded to servers from backup group while main is
+back online.
+
+This happens only if [Sticky Sessions]() are enabled. TempestaFW will stop
+forwarding requests to backup servers as soon as all the HTTP sessions, served
+by backup servers will be closed.
```
krizhanovsky: This seems just a back side of proper working of the scheduler. I think it should be moved to description of the scheduler.

doc/scheduling.md
```diff
@@ -0,0 +1,180 @@
...
+
+- Request is scheduled to a wrong server group or dropped while all the servers
+are reachable for TempestaFW.
+
+If request is scheduled to a wrong group it is most likely that there is an
+error in HTTP scheduler rules: request suit rule that is closer to the top
+of the list. Put more precise rules before more generic.
+
+Dropping of the message means that dropped request didn't match any rule.
```
krizhanovsky: This seems like a suggestion how to fix some abstract issue due to misconfiguration. I believe documentation like Troubleshooting or FAQ should be written in response to real users experience: when there is some mismatch between the system operation and user expectations.

doc/scheduling_devgd.md
```diff
@@ -0,0 +1,80 @@
+# Scheduling in TempestaFW: Developer Guide
```
krizhanovsky: I didn't review the file since it doesn't affect end user experience. If you'd imagine that the doc would be useful for you when you join the project, then I'm OK with it. Feel free to add any useful information for new comers (experienced core developers however). Just please place the developer docs separately from user guide.

doc/scheduling.md
ikoveshnikov: Github wiki is more than simple markdown pages, see example here: https://github.com/opencomputeproject/onie/wiki   

Seems that it have separate repo, would know more in next couple of days

ikoveshnikov:
Outdated by wiki pages

Issue #254, opened 2015-09-10
Dirty shutdown


krizhanovsky:
Test case:
1. start Tempesta with enabled Frang
2. send some request
3. shutdown Tempesta

Tempesta doesn't clean data structures on shutdown, so following trace is produced and the system hangs:

```
    .......
    [sync_sockets]   ss_send: sk ffff8800382ce740, sk->sk_socket           (null), state (Established)
    [sync_sockets]   ss_send:213 entail skb=ffff88005bc61440 data_len=0 len=467
    [sync_sockets]   ss_send:231 sk=ffff8800382ce740 is_queue_empty=0 tcp_send_head(sk)=ffff88005bc61440 sk->sk_state=1
    [tempesta]   Free msg: ffff8800382cd040
    [tempesta]   free skb ffff880059feb800: truesize=2271 sk=          (null), destructor=          (null) users=1 type=Conn_Clnt
    [tempesta]   Free msg: ffff88004634b040
    [tempesta]   free skb ffff88003c937c48: truesize=2368 sk=          (null), destructor=          (null) users=1 type=Conn_Srv
    [tempesta] got state via sysctl: stop
    [tempesta] stopping all modules...
    [tempesta]   stopping modules...
    [tempesta]   mod_stop(): frang
    [tempesta]   spec cleanup: 'frang_limits'
    [tempesta]   spec cleanup: 'http_methods'
    [tempesta]   spec cleanup: 'http_ct_vals'
    [tempesta]   mod_stop(): tfw_sched_http
    [tempesta]   spec cleanup: 'sched_http_rules'
    [tempesta]   spec cleanup: 'match'
    [tempesta]   mod_stop(): sock_clnt
    [tempesta]   spec cleanup: 'listen'
    [tempesta]   mod_stop(): sock_srv
    [sync_sockets]   Close socket ffff880046388840 (account=1)
    [sync_sockets]   __ss_do_close: sk ffff880046388840, sk->sk_socket           (null), state (Established)
    [tempesta]   Free msg:           (null)
    [sync_sockets]   Close socket ffff880046388140 (account=1)
    [sync_sockets]   __ss_do_close: sk ffff880046388140, sk->sk_socket           (null), state (Established)
    [tempesta]   Free msg:           (null)
    [sync_sockets]   Close socket ffff880038310880 (account=1)
    [sync_sockets]   __ss_do_close: sk ffff880038310880, sk->sk_socket           (null), state (Established)
[...]

Issue #668, opened 2016-12-27
Bad socket cpu locality on new Linux 4.8.15


krizhanovsky:
After moving to [Linux 4.8.15](https://github.com/tempesta-tech/linux-4.8.15-tfw) there are many warnings

        [   94.378748] net_ratelimit: 406 callbacks suppressed
        [   94.382632] [sync_sockets] Warning: Bad socket cpu locality on <state change>: sk=ffff880098718a40 old_cpu=0 curr_cpu=2
        [   94.387874] [sync_sockets] Warning: Bad socket cpu locality on <close>: sk=ffff880098718a40 old_cpu=0 curr_cpu=2
        [   94.392556] [sync_sockets] Warning: Bad socket cpu locality on <state change>: sk=ffff880098927540 old_cpu=0 curr_cpu=1
        [   94.392557] [sync_sockets] Warning: Bad socket cpu locality on <close>: sk=ffff880098927540 old_cpu=0 curr_cpu=1
        [   94.439161] [sync_sockets] Warning: Bad socket cpu locality on <state change>: sk=ffff8800b06da0c0 old_cpu=3 curr_cpu=2
        [   94.443579] [sync_sockets] Warning: Bad socket cpu locality on <close>: sk=ffff8800b06da0c0 old_cpu=3 curr_cpu=2
        [   94.481354] [sync_sockets] Warning: Bad socket cpu locality on <state change>: sk=ffff8800af915540 old_cpu=2 curr_cpu=0
        [   94.483552] [sync_sockets] Warning: Bad socket cpu locality on <close>: sk=ffff8800af915540 old_cpu=2 curr_cpu=0
        [   94.560506] [sync_sockets] Warning: Bad socket cpu locality on <state change>: sk=ffff880098718a40 old_cpu=2 curr_cpu=3
        [   94.564996] [sync_sockets] Warning: Bad socket cpu locality on <close>: sk=ffff880098718a40 old_cpu=2 curr_cpu=3
        [   99.383667] net_ratelimit: 20 callbacks suppressed
        [   99.385786] [sync_sockets] Warning: Bad socket cpu locality on <state change>: sk=ffff880098619500 old_cpu=1 curr_cpu=3
        [   99.389106] [sync_sockets] Warning: Bad socket cpu locality on <close>: sk=ffff880098619500 old_cpu=1 curr_cpu=3
[...]

PR #676, opened 2017-01-29
Fix #116 and #254: clean restart

### LINKED ISSUE #116: Client connections are not closed when Tempesta is stopped
Loaded Tempesta has a handful of listening sockets open, a handful of connections to backend servers, and a potentially huge number of client connections at any given time. When Tempesta is stopped, listening sockets, as well as connections to backend servers are closed. However, client connections are not closed at that time. Sockets will eventually be closed by the kernel, but not closing connections by Tempesta leads to memory leak as Tempesta internal data linked to each socket is not released.

The issue is somewhat complicated by the fact that currently Tempesta doesn't keep track of client connections. A solution need to be developed that takes the minimal amount of resources, both CPU and memory.

krizhanovsky:


tempesta_fw/cfg.c
```diff
@@ -1517,61 +1530,69 @@ tfw_cfg_start_mods(const char *cfg_text, struct list_head *mod_list)
...
-	MOD_FOR_EACH_REVERSE(mod, mod_list) {
+	MOD_FOR_EACH_REVERSE(mod, &tfw_cfg_mods) {
 		spec_cleanup(mod->specs);
 	}
 
 	return ret;
 }
 
+int
+tfw_cfg_start(void)
```
ikoveshnikov: Function is small and pretty, but it's a new public function. More over, we have restrictions here: it cannot be called twice without cleanup. That's your opinion, should we document it anyway?

tempesta_fw/sock.c
```diff
@@ -78,6 +85,63 @@ skb_sender_cpu_clear(struct sk_buff *skb)
...
+static void
+__active_guard_exit(unsigned long val, int cpu)
+{
+	atomic64_sub(val, per_cpu_ptr(&__ss_active, cpu));
+}
+
+#define ss_active_guard_enter(v)	__active_guard_enter(v,		\
+							     smp_processor_id())
+#define ss_active_guard_exit(v)		__active_guard_exit(v,		\
+							    smp_processor_id())
```
ikoveshnikov: Just matter of taste, but i like more formatting as below:
```
#define ss_active_guard_enter(v) \
	__active_guard_enter(v,	smp_processor_id())
#define ss_active_guard_exit(v) \
	__active_guard_exit(v, smp_processor_id())
```

tempesta_fw/sock_srv.c
```diff
@@ -471,6 +438,10 @@ tfw_sock_srv_start(void)
 static void
 tfw_sock_srv_stop(void)
 {
+	/*
+	 * Connections list is read-only at run time for now, so no need
+	 * to synchronize the list access or disable softirqs.
+	 */
```
keshonok: Don't we need to wait until the connections are closed by Tempesta (which is different from having the sockets closed).

tempesta_fw/sock.c
```diff
@@ -1049,7 +1130,14 @@ ss_connect(struct sock *sk, struct sockaddr *uaddr, int uaddr_len, int flags)
 	if (sk->sk_state != TCP_CLOSE)
 		return -EISCONN;
 
-	return sk->sk_prot->connect(sk, uaddr, uaddr_len);
+	if (ss_active_guard_enter(SS_SOFTIRQ_VAL))
```
keshonok: Logically, this should be `SS_LISTEN_VAL` as this is the connection establishing phase, and the connection is not live.

tempesta_fw/cfg.c
krizhanovsky: Original Cfg code mixes start/stop/restart activity with configuration. I did the first step in separating start/stop logic from configuration by moving sysctl routines out of cfg.c. I believe further steps must be taken to clean configuration code. The modules list must be handled outside cfg.c. So we control start/stop consistency **outside** cfg.c. I though it's obvious that calling start functions twice is wrong. Who in earth might wish to do so?

tempesta_fw/work_queue.h
```diff
@@ -51,10 +51,10 @@ int __tfw_wq_push(TfwRBQueue *wq, void *ptr, bool sync);
 int tfw_wq_pop(TfwRBQueue *wq, void *buf);
 
 static inline int
-tfw_wq_push(TfwRBQueue *wq, void *ptr, int cpu, struct irq_work *work,
+tfw_wq_push(TfwRBQueue *q, void *ptr, int cpu, struct irq_work *work,
```
krizhanovsky: I just touched `tfw_wq_push()` at some place and made the change for name consistency. However I didn't spend time for full names cleanup - we can always do it background. Generally shorter names are preffered if it's still clear what do they mean.

tempesta_fw/sock.c
```diff
@@ -50,8 +50,15 @@ static const char *ss_statename[] = {
 };
 #endif
 
+#define SS_LISTEN_VAL		0x0000000000000001UL
+#define SS_LISTEN_VAL_MASK	0x000000000000ffffUL
+#define SS_SOFTIRQ_VAL		0x0000000000010000UL
```
krizhanovsky: > is it correct that they way it's used, the maximum number of these is the number of active CPUs in the system?

Basically, yes. I create a comment. Actually `__ss_active` and the constants implement a counter of current SS down- and upcalls on the fly. This is different from connections counting.

tempesta_fw/sock.c
```diff
@@ -78,6 +85,63 @@ skb_sender_cpu_clear(struct sk_buff *skb)
...
+	}
+}
+
+static void
+__active_guard_exit(unsigned long val, int cpu)
+{
+	atomic64_sub(val, per_cpu_ptr(&__ss_active, cpu));
+}
+
+#define ss_active_guard_enter(v)	__active_guard_enter(v,		\
```
krizhanovsky: Actually, guard is almost synonym for `lock`, so there could be many guards. `SS_CALL_GUARD_*` also explicitly says that the guard is for `SS_CALL()`, SS upcall interface.

tempesta_fw/sock.c
```diff
@@ -78,6 +85,63 @@ skb_sender_cpu_clear(struct sk_buff *skb)
...
+	SS_CALL(cb, sk);						\
+})
+
+#define SS_CALL_GUARD_EXIT(cb, sk)					\
+do {									\
+	if (unlikely(sk->sk_incoming_cpu == TFW_SK_CPU_INIT))		\
+		break;							\
+	SS_CALL(cb, sk);						\
+	__active_guard_exit(SS_SOFTIRQ_VAL, sk->sk_incoming_cpu);	\
+} while (0)
```
krizhanovsky: The answer is "simplicity". All the macros are just several lines of code. Meantime, if a function is used, then the function must work with variadic arguments as well as accept address of the function, i.e. you have to pass full `((SsProto *)(sk)->sk_user_data)->hooks->callback_name` to each call of the function, instead of just pass name of the callback. At least I din't figure out easy and nice way to program it using functions. Frankly speaking I also don't see an issue with the macro...

tempesta_fw/sock.c
```diff
@@ -424,16 +488,20 @@ ss_do_close(struct sock *sk)
...
-/*
- * This function is for internal Sync Sockets use only. It's called
- * under the socket lock taken by the kernel, and in the context of
- * the socket that is being closed.
+/**
+ * This function is for internal Sync Sockets use only. It's called under the
+ * socket lock taken by the kernel, and in the context of the socket that is
+ * being closed.
+ *
+ * This is unintentional connection closing, usually due to some data errors.
```
krizhanovsky: Yes, we've lived for a long time with two callbacks `connection_drop()` and `connection_error()` which are basically just did the same thing (they actually called exactly the same code for either server and client connections). I commented purpose of the callbacks in `ss_hooks` definition. In short:

* connection_error - SS upcall letting upper layer to know about some errors in a TCP connection;

* connection_drop - is just a cleanup callback which is called when a TCP connection is already closed and we can do cleanup at Tempesta layer.

Please also note, that we have a consideration that TfwConnection lives longer than underlying TCP socket.

> `ss_droplink()` calls `connection_error()` callback now. That seems wrong. 

The comment explains the usage. "Intentional" means that we, Tempesta FW, intentionally closes connection. Basically, we don't care about other side intentions. If a client closes connection, then we just drop all associated resources with the connection, the same way as if there is some real connection error, e.g. TCP RST. The same for server: what's the difference why the server connection dies? We just move to failovering process regardless whether the server resets the connection or there were some error at TCP/IP layer. The different story is *intentional* closing, when Tempesta FW closes a connection - very different logic must be called in this case.

tempesta_fw/sock.c
```diff
@@ -483,7 +551,7 @@ __ss_close(struct sock *sk, int flags)
 	ss_do_close(sk);
 	bh_unlock_sock(sk);
 	if (flags & SS_F_CONN_CLOSE)
-		SS_CALL(connection_drop, sk);
+		SS_CALL_GUARD_EXIT(connection_drop, sk);
```
krizhanovsky: Hm, it seems I didn't get you.

`connection_drop()` is called on intentional, our side, closing only (`ss_tx_action()` and `__ss_close()`). `ss_droplink()` is called on unintentional, other side in this case, closing. In any case `ss_do_close()` is called under socket lock. So `connection_drop()` and `connection_error()` callbacks will be called concurrently in your case. Basically, client and server functions registered on the SS hooks are designed to be run in concurrently: we unlink TfwConnection from `sk->sk_user_data` before destroying it and carefully track TfwConnection reference counter to survive concurrent TfwConnection put.

About which issue exactly are you talking about? Do you see exact code places with some weakness?

tempesta_fw/sock.c
krizhanovsky: Logically you're right and I change the constant. However, technically there is no difference. I used LISTEN constant just to get a barrier which guarantees that after closing listening socket no new connection appears and we can safely traverse all established connections and close them all. So at this point the only guarding, which prevents calling the callback on shutdown phase, has sense and exact constant value doesn't matter.

tempesta_fw/sock.c
```diff
@@ -1162,17 +1256,122 @@ ss_tx_action(void)
...
 		sock_put(sk); /* paired with tfw_wq_push() */
 	}
 }
 
+#define HANDLE_TOO_LONG_WAIT(t0, job)					\
+do {									\
+	if (t0 + HZ * 5 < jiffies) {					\
+		SS_WARN("cpu %d: pending " job " for 5s (__ss_active=%#lx)\n",\
+			cpu, atomic64_read(active));			\
+		return;	/* give up - everything should be finished now */\
```
krizhanovsky: Yes, the problem is with #668: sometimes we just can't accurately account sockets. I added comment to the issue https://github.com/tempesta-tech/tempesta/issues/668#issuecomment-276735774 and adjust the comment in the macro.

tempesta_fw/sock_clnt.c
```diff
@@ -241,6 +244,27 @@ static const SsHooks tfw_sock_clnt_ss_hooks = {
...
+	 */
+	return ss_close(conn->sk);
+}
+
+static int
+tfw_cli_conn_close_all(TfwClient *cli)
+{
+	TfwConnection *conn;
+
+	tfw_peer_for_each_conn(cli, conn, list, __cli_clonn_close_cb);
```
krizhanovsky: Yes, this is somewhat tricky. Actually `tfw_peer_for_each_conn()` returns a value, so if the work queue is overflown, then `tfw_cli_conn_close_all()` returns an error and we continue our closing loop in `tfw_sock_clnt_stop_all()`. I tried to make the code generic for server and client connections, but it seems returning from `tfw_peer_for_each_conn()` makes a reader wondering. Peer locks are also an issue - it's very desired to keep locking inside peer.h. I reworked the code for explicit returning.

tempesta_fw/sock_srv.c
```diff
@@ -197,6 +203,10 @@ tfw_sock_srv_connect_try_later(TfwSrvConnection *srv_conn)
 	 */
 	static const unsigned long timeouts[] = { 1, 10, 100, 250, 500, 1000 };
 
+	/* Don't rearm reconnection timer if we're about to shutdown. */
+	if (unlikely(!ss_active()))
```
krizhanovsky: The problem is with this timer only, not all timers in the system. The problem is that we should not initiate new connection attempts during shoutdown process, because we need to traverse all the existing connections and close all of them. If the connection traversing and closing processes races with a new connection establishing, then we can leave the connection open which leads to a crash. Meantime it's OK to delete other timers at appropriate module stopping.

tempesta_fw/sock_srv.c
```diff
@@ -298,13 +308,34 @@ tfw_sock_srv_connect_complete(struct sock *sk)
...
+ * associated resources before SS put()'s the socket.
+ */
+static void
+tfw_sock_srv_connect_drop(struct sock *sk)
+{
+	TfwConnection *conn = sk->sk_user_data;
+
+	tfw_connection_drop(conn);
+	tfw_connection_put(conn);
+}
```
krizhanovsky: Generally I explained the design decisions above.

We don't need to unlink Tempesta from sk, since it must be guaranteed that TfwConnection live longer than sk. BTW I strongly dislike current set of link/unlink calls: it's really hard to manage the links. I thought about aggregation of some of the calls, but didn't find a quick solution, so I left things as they are. Please think about the code simplification as well.

> Stats are not collected;

fixed

`tfw_connection_get_if_nfo()` doesn't return a connection if it's reference count is zero. Also `tfw_connection_put()` don't destruct a connection if its reference counter isn't zero. However, freed connections stay in schedulers' data structures and we could get a crash. This is why we must stop sock_srv module _after_ sock_clnt to guarantee that there is no scheduling requests when all server connections are dead.

tempesta_fw/sock_srv.c
krizhanovsky: Yes, we do this in `ss_synchronize()` called from `tfw_cfg_stop()` (it surely must be moved from Cfg module as well in future). In particular we're waiting until all connection active counters are zeroed on all the CPUs.

tempesta_fw/sock.c
keshonok: I have spent quite some time trying to think of different scenarios that may occur here. This is tricky.

So, the case of asynchronous close() is covered by a check for `SOCK_DEAD`. A close() is asynchronous unless a specific set of criteria are satisfied, and one criteria in the set is that we're in SoftIRQ. 

In the case of synchronous close() [this should really be called *immediate* here as `ss_close_sync()` may enqueue a close() the same way as an asynchronous `ss_close()`, and so it would behave the same way as an asynchronous close()] the point is that all calls into SS/Tempesta are serialized with a socket lock. SS is not called by the kernel when the socket is dead (SOCK_DEAD). It means that if we're here in SS and we're doing a synchronous/immediate close(), then the socket is alive at this moment.

Is that correct? Perhaps it's best to document this.

Unless somehow an immediate socket close() is called twice in succession by Tempesta in one SoftIRQ, we're safe.

Can it be called twice? I am not sure yet. Perhaps, that must be construed as a bug. Right now, there's no protection against that, so it might be a good idea to put a trap for it.

tempesta_fw/peer.h
```diff
@@ -43,29 +43,42 @@ static inline void
...
+	int r = 0;							\
+	spin_lock(&(p)->conn_lock);					\
+	list_for_each_entry(conn, &(p)->conn_list, member) {		\
+		r = (cb)(conn);						\
+		if (unlikely((r)))					\
+			break;						\
+	}								\
+	spin_unlock(&(p)->conn_lock);					\
+	r;								\
+})
```
krizhanovsky: Unfortunately, we need to pass `member` to the macro and this is the only reason why I used macro instead of inline function. We don't know where exactly TfwConnection embodies list meber, so we can't get connection from a list member to pass the pointer to the callback function. Including connection.h to peer.h looks ever stronger ugliness than using the macro. Meantime, I still don't see an issue with the macro: it's small, it's does generic logic and it uses common GCC practices.

tempesta_fw/sock_srv.c
```diff
@@ -450,10 +418,10 @@ tfw_sock_srv_connect_srv(TfwServer *srv)
 static int
 tfw_sock_srv_disconnect_srv(TfwServer *srv)
 {
-	TfwSrvConnection *srv_conn;
+	TfwConnection *conn;
+
+	tfw_peer_for_each_conn(srv, conn, list, tfw_sock_srv_disconnect);
```
krizhanovsky: Yeah, fixed:

        while (tfw_sg_for_each_srv(tfw_sock_srv_disconnect_srv))
                schedule();

tempesta_fw/sock.c
krizhanovsky: > It means that if we're here in SS and we're doing a synchronous/immediate close(), then the socket is alive at this moment.

Yes. Tempesta layer can close a socket only if it has corresponding TfwConnection. When we link a socket with TfwConnectionn we increment the socket reference count in `tfw_connection_link_to_sk()`, so if Tempesta closes a socket, then it refers it and the socket is alive.

> Can it be called twice? I am not sure yet. Perhaps, that must be construed as a bug.

Yes, but we have reference counters which should protect the socket and TfwConnetion.

tempesta_fw/sock.c
```diff
@@ -1162,17 +1275,126 @@ ss_tx_action(void)
...
+	might_sleep();
+	while (1) {
+		for_each_online_cpu(cpu)
+			acc += atomic64_read(per_cpu_ptr(&__ss_act_cnt, cpu));
+		BUG_ON(acc < 0);
+		if (!(acc & SS_M_ACT_NEWCONN))
+			break;
+		schedule();
+		if (acc == acc_old) {
+			HANDLE_TOO_LONG_WAIT(t0, "listeners");
```
krizhanovsky: Replaced the message string by "listening sockets". Actually this phase synchronize listening sockets, not just new connections.

tempesta_fw/procfs.h
```diff
@@ -49,20 +48,30 @@ typedef struct {
...
-	u64	rx_bytes;
-} TfwPeerStat;
+/*
+ * @msgs_fromcache	- The number of messages served from cache.
+ * @online		- The number of clients online.
+ */
+typedef struct {
+	TFW_STAT_COMMON;
+	u64	msgs_fromcache;
+	u64	online;
```
krizhanovsky: This is online _clients_, not connections. And we have client specific structure, so `clnt.clinet_online` doesn't have too much sense.

PR #679, opened 2017-02-08
Testing framework

ikoveshnikov:
Here is framework for functional, stress and performance testing of TempestaFW. `Readme` provide more information.


tempesta_fw/t/functional/README.md
```diff
@@ -0,0 +1,197 @@
...
+## Recommended configuration
+
+Running tests during development process can cause crashes to TempestaFW.
+Since TempestaFW is implemented as a set of kernel modules it is not convenient
+to run testing framework on the same host. It is recommended to run testing
+framework on a separated host.
+
+Recommended test-beds:
+
+- Fully distributed. 4 different hosts with their own roles are used. This
```
krizhanovsky: I think in most cases the test suite will be run on the same VM, just to verify that Tempesta FW passes all the tests. The distributed test is certainly useful, but for development tasks only. Thus, it's better to use single VM installation as a default configuration.

tempesta_fw/t/functional/README.md
```diff
@@ -0,0 +1,197 @@
...
+Testing framework has a lot of options described in `tests_config.ini'.
+You can create default tests configuration by calling:
+
+```sh
+$ ./run_tests.py -d
+```
+Or run next command to show an example of configuration:
+```sh
+$ ./run_tests.py -e
+```
```
krizhanovsky: The option is good, but a user still needs to read the README to configure the test suite. And that's not comfortable at all. It's better to add a default configuration file with comments for all the options.

tempesta_fw/t/functional/helpers/tf_cfg.py
```diff
@@ -0,0 +1,142 @@
...
+                                            'port': '22',
+                                            'dir': '/root/tempesta',},
+                               'Server': {'ip': '127.0.0.1',
+                                          'mac': 'ff:ff:ff:ff:ff:ff',
+                                          'hostname': 'localhost',
+                                          'user': 'root',
+                                          'port': '22',
+                                          'nginx': 'nginx',
+                                          'workdir': '/root/nginx',
+                                          'resourses': '/srv/http/'}
```
krizhanovsky: It's better to use `/var/www/html/` as a default directory for Apache HTTPD or Nginx

tempesta_fw/t/functional/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+`resourses` - absolute path to directory with sample web pages. Must be
+reachable by `nginx`.
+
+
+### Run tests
+
+To run all the tests simply run:
+```sh
+$ ./run_tests.py
```
krizhanovsky: Probably we need the file only. Now `run_all_tests.sh` is just broken, so it's better to remove it at all

PR #694, opened 2017-03-06
First version of the HTTP/2 implementation

sysprg:
This is first phase of the HTTP/2 implementation


krizhanovsky:
There should be some Makefile which at least build the code. Also it seems you generated hftestdata.h using some script, so there is no sense to commit the large auto-generated file to the repository. Instead it's better to commit the generation script and run it on build.

tempesta_fw/http2/hpack_helpers.h
```diff
@@ -0,0 +1,166 @@
...
+
+#endif
+
+#ifdef Platform_64bit
+	#define HPACK_LIMIT 63
+	#define HPACK_LAST 1
+#else
+	#define HPACK_LIMIT 28
+	#define HPACK_LAST 15
+#endif
```
krizhanovsky: The above looks like a bunch of unused and doubling of Linux routines code.

tempesta_fw/http2/rotate.h
```diff
@@ -0,0 +1,266 @@
+/**
+ *		Tempesta FW
+ *
+ * Conversion between little and big endian numbers.
```
krizhanovsky: The comment is just the same as for netconv.h, so is it just one more duplication of a code available in Linux?

tempesta_fw/http2/hindex.h
```diff
@@ -0,0 +1,76 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 59
+ * Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ */
+
```
krizhanovsky: Actually, it's tricky to build Linux kernel module code in a subdirectory within code in a parent directory. Thus, it's assumed that http2 content will be moved to tempesta_fw/ to be built as part of tempesta_fw.ko module. So please don't introduce too many small source code files and preferably name them as they will be in tempesta_fw/.

tempesta_fw/http2/huffman.c
```diff
@@ -0,0 +1,641 @@
...
+				shift = state[i].shift;
+				offset = state[i].offset;
+				if (shift >= 0) {
+					* dst++ = (char) offset;
+					current -= shift;
+					goto Root;
+				}
+				else {
+					current += shift;
+				#ifdef HT_BALANCED_TREE
```
sysprg: But this macro definition is generated from the analysis of the Huffman tree, if new symbols appear in the standard or the existing ones are changed, then another branch of code can become active. All definitions from hfstate.h are generated by hgen/hgen.c

tempesta_fw/http2/huffman.c
```diff
@@ -0,0 +1,641 @@
...
+							break;
+						}
+					}
+					else {
+						goto End;
+					}
+				}
+			}
+			current += HT_NBITS - HT_MBITS;
+			{
```
sysprg: Usually yes, but in this case we have an critical loop. I will need to enter new variables. And two almost identical blocks of code will become too different. In addition, I can forget in one of them a variable from another, which will lead to tricky errors. And if (instead of itroducing new variables) you make the variables with same names are common, then this will make analysis of the code (by the compiler) more difficult. The compiler does not always understand that a "same" variable in two different pieces of code is not linked with itself by value. Compiler does not understand that the second block of the code never uses the values ​​of the variables generated in the first code fragment. An explicit block sometimes helps the compiler to better optimize the critical loop.

tempesta_fw/http2/huffman.c
```diff
@@ -0,0 +1,641 @@
...
+#define Write4()		     \
+	Write2();		     \
+	dst[2] = (char) (aux >> 40); \
+	dst[3] = (char) (aux >> 32)
+#define WriteAux()		     \
+	Write4();		     \
+	dst[4] = (char) (aux >> 24); \
+	dst[5] = (char) (aux >> 16); \
+	dst[6] = (char) (aux >> 8);  \
+	dst[7] = (char) aux
```
sysprg: And tomorrow someone decides to enable a strict alignment control in the processor (by already exisiting flag in EFLAGS) and re-compile kernel with it, because Intel can write in the recommendations for the next processor that "you should never use unaligned access to memory" ... :)

tempesta_fw/http2/hpack_helpers.h
sysprg: No, in this case it's the contants for decoding a specific representation of integers in HPACK. Here we can kill the code for 32-bit processors, but the constants themselves are not a repetition of constants from other h-files. They are specific for splitting an integer into 7-bit blocks in this particular encoding method. I can write ((64 / 7) * 7) and ((1 << (64 - (64 / 7) * 7)) - 1) for more clarity, but it's not just (64 - 1).

krizhanovsky: I mostly meant `GET_INT16()` and `GET_INT32()` macros.

tempesta_fw/http2/huffman.c
krizhanovsky: Hm, I see... You set `Incomplete_Path` in hgen.c with printing "Bug" comment. Does it really mean some problems? How probable the standard changes? Since standard changes are typically means broken existing services, I'd assume that the changes aren't probable, but probably I'm missing something...

tempesta_fw/http2/huffman.c
krizhanovsky: It seems I didn't get you. It seems that the issue is with `shift` and `i` variables only. Why can't we just declare them global for the whole function? It seems we can just remove the block braces and move the two variables to begin of the function and that's all required changes. Why performance of the function will be affected?

tempesta_fw/http2/huffman.c
krizhanovsky: And the changes will mean for us a lot of work. However, the scenario is very unlikely and we should not use the complex and slow code just to avoid improbable issues.

In general, there a lot of amazing open source software which is highly portable. However, the software can not utilize full power of a particular OS and a hardware. Tempesta FW targets maximum performance by full utilization of a target platform. Our README.md clearly limits ever x86-64 range of compatible hardware (typically you can not run Tempesta FW on a CPU older than Haswell) and we do a lot of work to get maximum from the hardware.

tempesta_fw/http2/huffman.c
sysprg: I'm just trying to write the code so that in the future it can be reused again without major changes. It's just #ifdef, it does not generate unnecessary code. Let's say some other tree appears in the new version of the standard. I can delete these branches, but no one can restore them quickly if another version gets another tree. "Bug" in this context is not a generator error, but a path in a tree that does not lead to the generation of a valid character at the output. Now the decoder can not only decode symbols, but also detect all errors (deviations from the standard) in the input stream.

tempesta_fw/http2/huffman.c
sysprg: The values ​​of the variables "I" and "shift" in the first block are not related to the values ​​of the variables "I" and "shift" in the second block. If we declare them externally, then we force the compiler to allocate the same registers for two different sets of variables. Probably, SSA-analysis in the latest versions of GCC will be able to split the false dependency on the data by using analysis of the post-dominantors tree, since it will see that these variables are re-assigned anew outside the loop. And it himself will create two new variables for the second block in the new registers. But I tried to write code in such manner that it does not much depend on the features of optimization algorithms in the compiler. One of the main general rules is to minimize the scope of all independent variables.

tempesta_fw/http2/huffman.c
krizhanovsky: I believe the code will be just dead for near future. Meantime we have to review and support the code, so we do not maintain unused code. Moreover, we do not develop code for "possible future needs": we don't know the future until it happens, so requirements and an environment can change. In general, we develop code when it's needed only.

tempesta_fw/http2/huffman.c
krizhanovsky: Ok, I see. The braces looks strange and in general it's better to keep code simple and clear. If you make a code "strange" for performance reasons, then it must be commented, so a reader won't be wondering. However, I believe GCC does a good job in the local variables optimizations. We ever recently found that it rearranges the variables definitions, so that they are placed on the stack w/o gaps.At least I never seen explicit code localization for performance reasons in the kernel....

tempesta_fw/http2/huffman.c
krizhanovsky: Ok, this is the case for https://github.com/tempesta-tech/tempesta/issues/695 . Please write a good comment about the optimization.

Issue #696, opened 2017-03-09
kernel BUG at /root/ab/tempesta-ab/tempesta_fw/http.c:1321


keshonok:
The setup:
* Tempesta's current master at c40924b;
* Tempesta kernel 4.8.15 at [5b3aa7f](https://github.com/tempesta-tech/linux-4.8.15-tfw/commit/5b3aa7fe30ce46bb8f112a9544548da926cb3714);
* Both Tempesta and Nginx run on the same physical test server with 12 CPU cores and 64Gb RAM;
* Simplest Tempesta configuration:
```
server 127.0.0.1:8080
```

The configuration means that Tempesta bypasses cache on all requests. All requests are forwarded to the back-end server without exceptions.

The test is a simple `wrk -t 8 -c 1024 -d 600s http://192.168.0.1`. It runs on another physical machine with 8 CPU cores and 32Gb RAM. It is connected with the test server via 10G interface.

```
[  632.352381] kernel BUG at /root/ab/tempesta-ab/tempesta_fw/http.c:1321!
[  632.359233] invalid opcode: 0000 [#1] SMP
[  632.363473] Modules linked in: tfw_sched_rr(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) tempesta_tls(O) joydev hid_generic usbhid hid mlx4_en iTCO_wdt iTCO_vendor_support intel_rapl evdev ehci_pci sb_edac edac_core x86_pkg_temp_thermal ast intel_powerclamp igb coretemp ehci_hcd ttm snd_pcm crct10dif_pclmul crc32_pclmul crc32c_intel ghash_clmulni_intel snd_timer aesni_intel drm_kms_helper ptp snd aes_x86_64 glue_helper lrw gf128mul ablk_helper xhci_pci xhci_hcd drm ahci cryptd pps_core mlx4_core soundcore libahci usbcore pcspkr i2c_algo_bit libata fb_sys_fops sg i2c_i801 syscopyarea sysfillrect i2c_smbus sysimgblt mei_me ioatdma i2c_core mei lpc_ich dca usb_common mfd_core wmi ipmi_si ipmi_msghandler acpi_power_meter acpi_pad tpm_tis tpm_tis_core tpm button autofs4 ext4 crc16 jbd2 mbcache sd_mod mpt3sas raid_class scsi_transport_sas scsi_mod shpchp
[...]

PR #700, opened 2017-03-14
Do not share @req->reason with @req->resp. (#696)

keshonok:
HTTP responses don't use @msg.seq_list. No need to init it.

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -555,8 +555,9 @@ tfw_http_req_error(TfwSrvConn *srv_conn, TfwHttpReq *req,
 {
 	tfw_http_req_delist(srv_conn, req);
 	list_add_tail(&req->fwd_list, equeue);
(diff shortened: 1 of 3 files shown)


tempesta_fw/http.c
```diff
@@ -2573,7 +2574,12 @@ static TfwConnHooks http_conn_hooks = {
 int __init
 tfw_http_init(void)
 {
-	int r = tfw_gfsm_register_fsm(TFW_FSM_HTTP, tfw_http_msg_process);
+	int r;
+	/* Make sure @req->httperr doesn't take too much space. */
+	BUILD_BUG_ON(sizeof(((TfwHttpMsg *)0)->httperr)
```
ikoveshnikov: Please use  [FIELD_SIZEOF](http://lxr.free-electrons.com/source/include/linux/kernel.h#L77) macro

Issue #685, opened 2017-02-20
Sticky cookies load balancing


krizhanovsky:
Sticky cookie scheduler (#593) and the whole http_sess must be extended by the following:

* sticky cookie must be configurable per vhost and/or server group to be able to set different sticky cookies for different serviced domain;

* sticky cookie must set configurable expire values;

* we should be able to use upstream native session cookies, so the cookie name should be specified with some new configuration argument (Nginx uses "learn" for the option), see more info at http://httpd.apache.org/docs/2.4/mod/mod_proxy_balancer.html#stickyness_implementation

* the sessions must be stored in persistent TDB table, just like we did this for client accounting in #1178

### Notes on persistent data structures

I add the notes here since I added the last bullet about TDB usage as the sessions storage with several doubts, at least that currently TDB is unfinished and buggy. There are two main reasons to use TDB instead of plain hash tables:

1. performance - HTries is significantly [faster](https://github.com/tempesta-tech/blog/blob/master/htrie/results) than hash and will be optimized further

2. data persistency (in normal conditions, not crash safety) - we need to dynamically deploy (#786) crucial security fixes for the kernel and system reboot is required, so it makes sense not to lose sessions and client accounting as well as the web cache. During one of the pilot projects we've discussed that typically there are monthly security updates.

The notes about persistent data structures are good to add to the Wiki, like [High avaliability](https://github.com/tempesta-tech/tempesta/wiki/High-availability) section, when the issues is completed.

ikoveshnikov:
There was a discussion, that locks in sticky sessions scheduler might be reworked in a better way. https://github.com/tempesta-tech/tempesta/pull/713#discussion_r108073676

ikoveshnikov:
Sticky cookie module should process a request before it's assigned to any vhost, see #1043, #899 (this behaviour is implemented in #1067 ).  With per-vhost cookie it means that we don't know in advance which cookie must be stored in the request. So the fast string matching algorithm should be used to search among all possible sticky cookie names.

ikoveshnikov:
There was a discussion in the chat about #685 and #1075 implementation. According to it there is no reasons to introduce new data structures, it's better to reuse/update existing ones. We can also put vhost name into cookie name, like: `__tfw_<vhost>` instead just `__tfw`. As soon sticky cookies can assign vhost to the request, they're close to the `http_chain`, so code used for http_chains can be reused. Please, revise the notes in the chat on implementing the issue.

[...]

PR #713, opened 2017-03-22
Sticky sessions implementation

### LINKED ISSUE #593: Sticky sessions scheduler

ikoveshnikov:
Change log:

* fix #593 implement sticky sessions scheduler
* fix setting unlimited sticky cookie life time
* improve stability of unit tests

comment https://github.com/tempesta-tech/tempesta/issues/76#issuecomment-287572183 not implemented in the patch, instead i work on it in separate branch since some discussion may be needed.


tempesta_fw/http_sess.c
```diff
@@ -720,6 +725,18 @@ tfw_http_sticky_secret_cfg(TfwCfgSpec *cs, TfwCfgEntry *ce)
 	return 0;
 }
 
+static int
+tfw_http_sticky_sess_lifetime_cfg(TfwCfgSpec *cs, TfwCfgEntry *ce)
+{
+	int r = tfw_cfg_set_int(cs, ce);
+
+	/* @sess_lifetime value of 0 means unlimited. */
+	if (!r && !tfw_cfg_sticky.sess_lifetime)
```
ikoveshnikov: `0` in terms of `sess_lifetime` means `unlimited`, but setting actually zero to the parameter makes session to be removed right after it was created. As discussed before, `UINT_MAX` value should represent `unlimited` value for the code, but for the user it should remain as `0`.

I can rewrite condition as follows if you find that more meaningful:
```
if (!r && !*(cs->dest))
    *(cs->dest) = UINT_MAX;
```

tempesta_fw/http_sess.c
```diff
@@ -736,12 +753,16 @@ TfwCfgMod tfw_http_sess_cfg_mod = {
 			.allow_none = true,
 		},
 		{
+			/* Value is parsed as int, set max to INT_MAX*/
 			.name = "sess_lifetime",
 			.deflt = "0",
-			.handler = tfw_cfg_set_int,
+			.handler = tfw_http_sticky_sess_lifetime_cfg,
 			.dest = &tfw_cfg_sticky.sess_lifetime,
```
ikoveshnikov: In `tfw_http_sticky_sess_lifetime_cfg` i use `tfw_cfg_set_int` for parsing int value and check it for min-max borders. To use that i need to set `.dest` either here or in `tfw_http_sticky_sess_lifetime_cfg`.

tempesta_fw/http_sess.c
keshonok: Yes, that would make this function to NOT depend on the specific variable.
That would also make it generic for getting a number value for a directive, and setting is to max value if the number is zero (also a default value).

tempesta_fw/http_sess.c
```diff
@@ -83,10 +84,10 @@ static TfwCfgSticky tfw_cfg_sticky;
...
 static char tfw_sticky_key[STICKY_KEY_MAXLEN];
 
-SessHashBucket sess_hash[SESS_HASH_SZ] = {
-	[0 ... (SESS_HASH_SZ - 1)] = {
-		HLIST_HEAD_INIT,
-	}
+static SessHashBucket sess_hash[SESS_HASH_SZ] = {
+//	[0 ... (SESS_HASH_SZ - 1)] = {
+//		HLIST_HEAD_INIT,
+//	}
```
krizhanovsky: Don't use C++ style comments. Don't leave dead (just commented out) code. Why did you remove the initialization? I know that it's just zeroying, but using the initialization is still a good practice. At least use the list initializer in `tfw_http_sess_init()` near to spinlock initialization.

tempesta_fw/http_sess.c
```diff
@@ -565,7 +566,7 @@ tfw_http_sess_obtain(TfwHttpReq *req)
 			goto found;
 	}
 
-	if (!(sess = kmem_cache_alloc(sess_cache, GFP_ATOMIC))) {
+	if (!(sess = kmem_cache_zalloc(sess_cache, GFP_ATOMIC))) {
```
krizhanovsky: Don't use __GFP_ZERO w/o need, that's relatively expensive thing to make the kernel write zeroes to free pages. Instead, you can just write 3 NULLs  to the new pointer fields.

tempesta_fw/http_sess.h
```diff
@@ -0,0 +1,64 @@
...
+ */
+struct tfw_http_sess_t {
+	unsigned char		hmac[SHA1_DIGEST_SIZE];
+	struct hlist_node	hentry;
+	atomic_t		users;
+	unsigned long		ts;
+	unsigned long		expires;
+	TfwStickyConn		st_conn;
+};
+
```
krizhanovsky: Extra empty line. Why do you define TfwHttpSess ouside of the file?

tempesta_fw/sched.c
```diff
@@ -71,6 +70,166 @@ tfw_sched_get_srv_conn(TfwMsg *msg)
...
+ * Find an outgoing connection for client with tempesta sticky cookie.
+ * @sess is not null when calling the function.
+ *
+ * Reuse req->sess->st_conn.srv_conn if it is alive. If not,
+ * then get a new connection for req->sess->srv_conn->peer from
+ * an appropriate scheduler. That eliminates the long generic
+ * scheduling work flow.
+ */
+static inline TfwSrvConn *
+__get_sticky_srv_conn(TfwMsg *msg, TfwHttpSess *sess)
```
krizhanovsky: Basically, `sched.c` isn't an appropriate place for Sticky cookie scheduling. `sched.c` is the generic scheduling logic. `__get_sticky_srv_conn()`, `__try_conn()` and maybe some other functions are good candidates for http_sess.c or a new file `sched/tfw_sched_sticky.c`. Also keep in mind #685 which is going to extend the logic even more.

tempesta_fw/sched.c
```diff
@@ -71,6 +70,166 @@ tfw_sched_get_srv_conn(TfwMsg *msg)
...
+{
+	TfwStickyConn *st_conn = &sess->st_conn;
+	TfwSrvConn *srv_conn;
+
+	read_lock(&st_conn->lock);
```
krizhanovsky: Why do you use read lock for `__try_conn()` which basically calls lock-free routines only?

tempesta_fw/sched.c
```diff
@@ -71,6 +70,166 @@ tfw_sched_get_srv_conn(TfwMsg *msg)
...
+		srv_conn = backup_sg->sched->sched_sg_conn(msg, backup_sg);
+	}
+
+	if (unlikely(!srv_conn))
+		TFW_DBG2("sched: Unable to select server from group '%s'\n",
+			 sg->name);
+
+	return srv_conn;
+}
+EXPORT_SYMBOL(tfw_sched_get_sg_srv_conn);
```
krizhanovsky: Please define the function before the first usage of it, i.e. before `__get_sticky_srv_conn()`.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -45,22 +45,25 @@
...
 
 typedef struct {
-	TfwSrvConn	*srv_conn;
-	unsigned long	hash;
-} TfwConnHash;
+	size_t			conn_n;
+	TfwServer		*srv;
+	TfwSrvConn		*conn[TFW_SRV_MAX_CONN];
+	unsigned long		hash[TFW_SRV_MAX_CONN];
+} TfwHashSrv;
```
krizhanovsky: Comment https://github.com/tempesta-tech/tempesta/pull/666/files#r94275356 is still not addressed. I saw your PR https://github.com/tempesta-tech/tempesta/pull/714/files, but it's somehow already closed.

tempesta_fw/http_sess.c
ikoveshnikov: Didn't want C++ comments to appear here. I use IDE with integrated clang code analysis and clang gets stuck on this initialization and the whole IDE hangs every time i make changes to this file.  I will be twice more careful on sending patches next time.

tempesta_fw/sched.c
ikoveshnikov: If session is under failovering, connection stored in sessions private data is no more valid. Even if original server is suddenly back online.

krizhanovsky: Isn't it better to use reference counting to make the session private data valid even under failovering?

tempesta_fw/http_sess.h
ikoveshnikov: In last PR i was advised by @keshonok  to move all the session related definitions from `http.h` to separate header. But  `TfwHttpSess` is required for `TfwHttpReq` in `http.h` and most functions in this header require definitions from `http.h`, so circular dependency take place here.

As for me, advice was very valuable: http.h is too big and placing there more and more definitions only breaks it to loosely connected parts.

PR #718, opened 2017-04-03
Ratio algorithm for dynamic load balancing

keshonok:
Also, includes dynamic memory allocation for server and connection data
so that the maximum number of servers or connections it's not limited by
constants.


tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,942 @@
...
+	 * Adjust the sum of ratios that is changed in this procedure.
+	 */
+	for (si = 0; si < ratio->srv_n; ++si) {
+		if (srvdata[si].oratio == 1) {
+			has_one_val = 1;
+			break;
+		}
+	}
+	if (has_one_val) {
+		unsigned int orsum = ratio->schdata.orsum;
```
ikoveshnikov: `ratio->schdata.orsum` is `unsigned long`, not `unsigned int`

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,942 @@
...
+        return NULL;
+}
+
+/*
+ * Return an entry to the RCU pool.
+ */
+static inline void
+__tfw_sched_ratio_rpool_put(TfwRatio *ratio)
+{
+        atomic_set(&ratio->free, 1);
```
ikoveshnikov: spaces instead of tabs, same for the next function

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,942 @@
...
+
+	for (si = 0; si <= nr_cpu_ids; ++si, ++ratio) {
+		smp_mb__before_atomic();
+		if (atomic_read(&ratio->free)) {
+			atomic_set(&ratio->free, 0);
+			return ratio;
+		}
+	}
+
+        return NULL;
```
ikoveshnikov: spaces instead of tabs

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,942 @@
...
+ * weights like { 10, 1 } make more sense than weights like { 1000, 10 }.
+ * Requests are distributed proportionally in both cases, but significant
+ * bursts are possible in the first case.
+ *
+ * TODO: The algorithm may and should be improved.
+ */
+static inline bool
+tfw_sched_ratio_is_srv_turn(TfwRatio *ratio, size_t csidx)
+{
+	unsigned int headsum2, tailsum2;
```
ikoveshnikov: both should be type of `unsigned long`

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,942 @@
...
+}
+
+/**
+ * Same as @tfw_sched_ratio_sched_sg_conn(), but schedule a connection
+ * to a specific server in a group.
+ */
+static TfwSrvConn *
+tfw_sched_ratio_sched_srv_conn(TfwMsg *msg, TfwServer *srv)
+{
+        int skipnip = 1, nipconn = 0;
```
ikoveshnikov: spaces instead of tabs

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,942 @@
...
+ *
+ * Ratio scheduler must be the fastest scheduler. Also, it's essential
+ * to maintain a completely fair distribution of requests to servers
+ * according to servers weights.
+ */
+static TfwSrvConn *
+tfw_sched_ratio_sched_sg_conn(TfwMsg *msg, TfwSrvGroup *sg)
+{
+	size_t srv_tried_n = 0;
+        int skipnip = 1, nipconn = 0;
```
ikoveshnikov: spaces instead of tabs

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,942 @@
...
+		size_t left_sdidx = srvdata[left].sdidx;
+		srvdata[left++].sdidx = srvdata[right].sdidx;
+		srvdata[right--].sdidx = left_sdidx;
+	}
+
+	return 1;
+}
+
+/*
+ *  * Get a free for use entry from the RCU pool.
```
ikoveshnikov: Extra asterisks

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,942 @@
...
+	 * connections were under restrictions for one reason or another.
+	 * The idea is that the conditions for server's connections may
+	 * change any time, and so the next time one or more connections
+	 * to the same server will not be restricted.
+	 * Perhaps, though, it makes sense to skip these servers that
+	 * were restricted, and go directly to the next server. Getting
+	 * the next server reqires a lock, so perhaps it makes sense to
+	 * to skip these repetitive servers while under the lock.
+	 */
+	while (srv_tried_n < ratio->srv_n) {
```
ikoveshnikov: Actually, condition `while (srv_tried_n < ratio->srv_n)` will never mean, that we try all possible servers. And you described that in a comment above. Better to rename `srv_tried_n` to `tries`:
```
tries = ratio->srv_n + 1;
while(--tries) {
    ...
```
At least, this code doesn't lie that it tries all the server. Even more: nature of this counter is almost the same as for `tries` counter in hash scheduler.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -245,11 +200,105 @@ tfw_sched_hash_get_srv_conn(TfwMsg *msg, TfwServer *srv)
...
+	TfwSrvConn *srv_conn;
+	TfwHashSrv *hsrv;
+	TfwHashSrvList *sl;
+
+	/*
+	 * Validate the number of servers in the group, and the number
+	 * of connections for each server.
+	 */
+	si = 0;
+	list_for_each_entry(srv, &sg->srv_list, list) {
```
keshonok: I'd rather leave the validation check. I've moved it to a separate function. It's called only once for each group at configuration processing time, and I consider it a good safety measure.

As for the comment on the new behaviour, it's here: https://github.com/tempesta-tech/tempesta/blob/bab5e182a4d4192a308e1510600832d5bdceb905/tempesta_fw/server.h#L116
I have extended it a little bit to be more visible and clear.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,942 @@
...
+	     tfw_sched_ratio_srvdata_cmp, tfw_sched_ratio_srvdata_swap);
+
+	/*
+	 * Do actions described in step 3 in the function's description.
+	 * Select the part of the array that omits entries from step 1
+	 * if there are any. Those are entries at the start and at the
+	 * end of the array. Reverse the sequence of server descriptor
+	 * indices in that part of the array.
+	 */
+	if (has_one_val) {
```
keshonok: Should be `if (!has_one_val) {`

tempesta_fw/t/unit/test_sched_ratio.c
```diff
@@ -0,0 +1,350 @@
...
+#include "sock_srv.c"
+
+#ifdef module_init
+#undef module_init
+#undef module_exit
+#define module_init(func)
+#define module_exit(func)
+#endif
+
+#include "../../sched/tfw_sched_ratio.c"
```
keshonok: Yes, that was the reason I didn't add unit tests for the new functionality. I didn't feel like they would be useful enough. These kind of tests would better be done with functional testing.

I will open a task for this, thanks.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -245,12 +200,114 @@ tfw_sched_hash_get_srv_conn(TfwMsg *msg, TfwServer *srv)
...
+
+	for (si = 0; si < sl->srv_n; ++si) {
+		if (sl->srvs[si].conn)
+			kfree(sl->srvs[si].conn);
+		if (sl->srvs[si].hash)
+			kfree(sl->srvs[si].hash);
+	}
+
+	kfree(sl);
+	sg->sched_data = NULL;
```
keshonok: This sounds like a good idea. Will do that.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,1176 @@
...
+ */
+static TfwRatioSrvDesc *
+tfw_sched_ratio_next_srv(TfwRatioPool *rpool, TfwRatio *ratio)
+{
+	size_t csidx;
+	TfwRatioSrvData *srvdata = ratio->srvdata;
+	TfwRatioSchData *schdata = &ratio->schdata;
+
+	/* Start with server that has the highest ratio. */
+	spin_lock(&schdata->lock);
```
krizhanovsky: This is a bad thing. We have to use the lock as described in the comment, but we still want to have RR (ratio) scheduler as the most quick one. However, probably that would be an issue now if we for example try it to run on 4 processors machine (e.g. 40 and more CPU cores). Thus, let's introduce 2 versions of the function: this one for dynamic ratios and the old one, from RR scheduler, for static ratios.

tempesta_fw/server.c
```diff
@@ -249,25 +229,14 @@ tfw_sg_release_all(void)
 	TfwServer *srv, *srv_tmp;
 	TfwSrvGroup *sg, *sg_tmp;
 
-	write_lock(&sg_lock);
```
krizhanovsky: Please comment that we don't need the lock because the function is called at shutdown time, when there are no users any more.

tempesta_fw/sock_srv.c
```diff
@@ -451,10 +459,8 @@ tfw_sock_srv_connect_srv(TfwServer *srv)
 	 * is locked, and spews lots of warnings. LOCKDEP doesn't know
 	 * that parallel execution can't happen with the same socket.
 	 */
-	list_for_each_entry(srv_conn, &srv->conn_list, list)
-		tfw_sock_srv_connect_try_later(srv_conn);
-
-	return 0;
+	return tfw_peer_for_each_conn(srv, srv_conn, list,
+				      __tfw_sock_srv_connect_try_later_cb);
```
krizhanovsky: Why do we need to call `tfw_sock_srv_connect_try_later()` under `conn_lock`?

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -245,12 +200,114 @@ tfw_sched_hash_get_srv_conn(TfwMsg *msg, TfwServer *srv)
...
+	int ret = -ENOMEM;
+	size_t size, ci;
+	unsigned int sum_conn_n;
+	TfwServer *srv;
+	TfwSrvConn *srv_conn;
+	TfwHashSrv *hsrv;
+	TfwHashSrvList *sl;
+
+	if (tfw_sched_hash_validate_grp(sg))
+		return -EINVAL;
```
keshonok: The purpose of the validation is to confirm that the number of entries in the lists match the recorded list sizes. It's convenient so that there's no need to check for that on each loop iteration when memory is allocated, partitioned, and assigned. The entries are iterated by traversing a list, so there's need to match that against the number of entries. 

It's done just once per whole SG (server group). Are you suggesting that it's better if it's done on each loop iteration (list traversal)?

tempesta_fw/sock_srv.c
```diff
@@ -521,17 +528,15 @@ tfw_srv_conn_free(TfwSrvConn *srv_conn)
...
 	int i;
 	TfwSrvConn *srv_conn;
 
-	for (i = 0; i < conns_n; ++i) {
+	for (i = 0; i < srv->conn_n; ++i) {
 		if (!(srv_conn = tfw_srv_conn_alloc()))
 			return -ENOMEM;
-		tfw_connection_link_peer((TfwConn *)srv_conn,
-					 (TfwPeer *)srv);
-		tfw_sg_add_conn(srv->sg, srv, srv_conn);
```
keshonok: You can't simply add a server or a connection to a scheduler. Scheduler's data is a set of very specific data structures that are built for efficiency. In both cases we need to build a new set of data for a scheduler, and then make a scheduler seamlessly use it via RCU. At least, that's my understanding. Did you have something different in mind?

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -245,12 +200,114 @@ tfw_sched_hash_get_srv_conn(TfwMsg *msg, TfwServer *srv)
...
+
+	sum_conn_n = 0;
+	hsrv = sl->srvs;
+	list_for_each_entry(srv, &sg->srv_list, list) {
+		size = sizeof(hsrv->conn[0]) * srv->conn_n;
+		if (!(hsrv->conn = kzalloc(size, GFP_KERNEL)))
+			goto cleanup;
+		size = sizeof(hsrv->hash[0]) * srv->conn_n;
+		if (!(hsrv->hash = kzalloc(size, GFP_KERNEL)))
+			goto cleanup;
```
keshonok: It's interesting, that for that we need to go over each server and count the total number of connections (to all servers). Nearly the same thing is done for validation.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,1176 @@
...
+	unsigned int	riter;
+	unsigned long	crsum;
+	unsigned long	orsum;
+} TfwRatioSchData;
+
+/**
+ * Historic (past) data unit for an individual upstream server.
+ *
+ * @x		- count of timer function invocations.
+ * @y		- RTT from APM in msecs.
```
keshonok: The primary reason was that derivative members become too long, like, say, `cnt_avg_rtt_avg`, etc. In calculations these names are used multiple times in the same statement which adds up to very long statements that are difficult to read and leads to ugly formatting. I believe the comments here are sufficient to understand the actual meaning of the calculations, and the algorithm can be easily correlated with the original in `slr.c`.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,1176 @@
...
+ * for each server in a group. Ratios are calculated on those RTT values.
+ * However that way the ratios do not represent the real weight of each
+ * server because a bigger RTT value mean that a server is less favorable
+ * and has a lesser, NOT bigger weight.
+ *
+ * Based on ratios calculated from RTT values, the algorithm here assigns
+ * a correct ratio to each server in the group.
+ * 1. If the minimal ratio is 1, then fill the entries with minimal ratio
+ *    with values from an entry with the maximum ratio. Fill the entries
+ *    with maximum ratio with values from an entry with minimal ratio.
```
keshonok: Yes, that's exactly what is means. I will rephrase it.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,1176 @@
...
+tfw_sched_ratio_srvdata_cmp(const void *lhs, const void *rhs)
+{
+	unsigned int lhs_ratio = ((const TfwRatioSrvData *)lhs)->oratio;
+	unsigned int rhs_ratio = ((const TfwRatioSrvData *)rhs)->oratio;
+
+	if (lhs_ratio > rhs_ratio)
+		return -1;
+	if (lhs_ratio < rhs_ratio)
+		return 1;
+	return 0;
```
keshonok: Technically, that's not equivalent as `oratio` is an unsigned integer. One extreme example would be:
```C
lhs = UINT_MAX;
rhs = 1;
return rhs - lhs;
```

nmmmnu:
or you can use something like this:

https://github.com/nmmmnu/HM3/blob/master/include/sgn.h

C++, but template can be removed easily. It will give you +1 / - 1

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,1176 @@
...
+{
+	size_t si, max_val_idx = 0;
+	TfwRatioSrvDesc *srvdesc = rpool->srvdesc;
+	TfwRatioSrvData *srvdata = ratio->srvdata;
+
+	/* Collect server weights from the configuration. */
+	for (si = 0; si < rpool->srv_n; ++si) {
+		srvdata[si].sdidx = si;
+		srvdata[si].weight = srvdesc[si].srv->weight;
+	}
```
keshonok: `tfw_sched_ratio_calc()` is common to static or dynamic weights. However the initial set up is different. Here in the case of static weights the weights are taken from a server instance (structure). In case of dynamic weights the weights are RTT received from APM. How do you propose to eliminate this loop?

tempesta_fw/sched/tfw_sched_ratio.c
nmmmnu: If you need +/-1, you can use something like SGN on top of it

https://github.com/nmmmnu/HM3/blob/master/include/sgn.h

C++, but template can be removed easily. It will give you +1 / - 1

keshonok: After a little research, I believe that a simple rewrite as shown below should be pretty efficient and without unnecessary branching in GCC with optimization:
```C
return (rhs_ratio < lhs_ratio) ? -1 : (rhs_ratio > lhs_ratio);
```

nmmmnu: -O2 removes the jump on your case:
https://godbolt.org/g/HKBmuQ

here is the code without jump at all in -O0
https://godbolt.org/g/sPhovh

#include <stdio.h>

static int calc(int const a, int const b){
	return (a < b) - (a > b);
}

int main(int argc, char **argv){
	printf("%d\n", calc(argv[1][0], argv[2][0]));
}

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,1176 @@
...
+ * Get a free for use entry from the RCU pool.
+ */
+static TfwRatio *
+tfw_sched_ratio_rpool_get(TfwRatioPool *rpool)
+{
+	int si;
+	TfwRatio *ratio = rpool->rpool;
+
+	for (si = 0; si <= nr_cpu_ids; ++si, ++ratio) {
+		smp_mb();
```
keshonok: That's a good question. The initial idea was to pair atomic store/load operations on `ratio->free` so that the current state of the variable is visible on all CPUs.

Now that I look at it again it appears that all atomic operations on a specific `ratio->free` variable are always executed on the same CPU: The timer function calls `tfw_sched_ratio_rpool_get()` and `tfw_sched_ratio_rpool_put()`. It also calls `call_rcu()` which calls `tfw_sched_ratio_rpool_put()`. Then the callback is executed on the same CPU as the timer function is. That leads to the conclusion that memory barriers are not needed at all around `ratio-->free`. Is that a correct line of thought?

Also, perhaps this function should better be rewritten using cmpxchg? That would make it clearer and cleaner.

tempesta_fw/sched/tfw_sched_hash.c
krizhanovsky: At the below you do loop with relatively heavy body, so additional check for two integers (index and the size of the array) won't hurt performance. We always keep loop bodies as small as possible and move conditions outside of the loop, but only if the condition hasn't been checked for _each_ iteration. There is absolutely no sense in splitting a loop to two loops.

tempesta_fw/sched/tfw_sched_hash.c
krizhanovsky: Yeah, I missed that we need `srv->conn_n`, so probably it'd be really hard to use only one allocation. However, these two particular calls of `kzalloc()` depend on the same data and must be merged.

tempesta_fw/sched/tfw_sched_ratio.c
krizhanovsky: Well, any names like `cnt` or `rtt` are much better than abstract `x` and `y`. Comments are good, but it's better to make the code readable as is, especially with that comments are about the structure definition and the actual code as very far from the definition, so you have to keep two windows open to read the code.

tempesta_fw/sched/tfw_sched_ratio.c
krizhanovsky: Yeah, that's not so simple. I've checked Hacker's Delight and it seems we need to verify the sign bit, so the subtraction requires more operations and becomes much complex. `? :` is essentially a condition, so I believe @nmmmnu's `(a < b) - (a > b)` is the best for the function.

tempesta_fw/sched/tfw_sched_ratio.c
krizhanovsky: I'd prefer copy & paste rather than extra overhead. Probably you'll find other opportunities to reduce amount of copy&paste'd code, e.g. moving smaller pieces of code to separate functions. Or maybe some defines... Performance is the top priority. We've recently discussed ugly defines which reduces copy&paste, but still have good performance.

tempesta_fw/sched/tfw_sched_ratio.c
krizhanovsky: Yeah, definitely. Atomic introduce implicit memory barriers and the pattern `atomic_read()` followed by `atomic_set()` introduces a race when we put() the `ratio` between the atomics, so I belive CMPXCHG is better here.

tempesta_fw/sock_srv.c
krizhanovsky: I meant something like if we reconfigure just number of connections for existing server groups, then probably we'd need the specific call to add a connection. However, I'm not sure whether we should introduce separate workflow for the case. Maybe it's easier just to reconfigure the server groups without significant overhead. If you think that we don't need the call any more, then I'm fine with it.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,1176 @@
...
+			} else if (srvdata[si].oratio == 1) {
+				right = si - 1;
+				break;
+			}
+	}
+	while (left < right) {
+		size_t left_sdidx = srvdata[left].sdidx;
+		srvdata[left++].sdidx = srvdata[right].sdidx;
+		srvdata[right--].sdidx = left_sdidx;
+	}
```
keshonok: I will put a thorough comment with an example in the function's description.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -0,0 +1,1176 @@
...
+	 */
+	if (!has_one_val) {
+		left = 0;
+		right = rpool->srv_n - 1;
+	} else {
+		for (si = 0; si < rpool->srv_n; ++si)
+			if (srvdata[si].oratio == max_ratio) {
+				left = si + 1;
+			} else if (srvdata[si].oratio == 1) {
+				right = si - 1;
```
keshonok: If we're here after a call to `tfw_sched_ratio_calc()`, then it's guaranteed that not all calculated ratios are equal. There's a comment on that on line 344 above. That's the precondition for going forward with these manipulations. So, if we're right here in this code block, then there are at least two servers, and one of them has ratio of 1, and there's another (different) server with maximum ratio.

tempesta_fw/sched/tfw_sched_ratio.c
keshonok: Do you mean, the old one, from RR scheduler, for static **equal** ratios?

tempesta_fw/apm.c
```diff
@@ -871,59 +865,140 @@ tfw_apm_stats(void *apmdata, TfwPrcntlStats *pstats)
...
+	 * Note that if server needs a recalculation of stats values,
+	 * it makes sense only if there were updates to server's stats
+	 * data. If there's no updates then a recalculation will lead
+	 * to the same (insufficient) result.
+	 */
+	for_each_online_cpu(cpu) {
+		TfwApmWqItem wq_item;
+		TfwRBQueue *wq = &per_cpu(tfw_apm_wq, cpu);
+
+		while (!tfw_wq_pop(wq, &wq_item)) {
```
keshonok: I am not sure I completely understand you suggestion. Are you suggesting *saving*/*storing* just one RTT sample per millisecond from Tempesta to APM's per-CPU array? Otherwise, an array of just 512 items (which is significantly less than 2k items in WQ now) will not be able to hold all RTT data until the timer function wakes up.

The major point was to avoid concurrency while updating data and calculating stats values for any particular server. That way the data is consistent at the time stats values are calculated. Concurrency caused data inconsistency due to incomplete updates, and data inconsistency caused the stats values calculation failures. So the point was that samples for any particular server should be submitted in a serialized manner (without concurrency), and stats values calculation should be part of that sequence.

PR #723, opened 2017-04-10
Correctly account for active connections when Tempesta is stopped on configuration error

keshonok:


tempesta_fw/sock.c
```diff
@@ -790,6 +790,8 @@ ss_tcp_state_change(struct sock *sk)
 			 */
 			ss_do_close(sk);
 			sock_put(sk);
+			if (!lsk)
+				ss_active_guard_exit(SS_V_ACT_LIVECONN);
```
krizhanovsky: It seems we need this only if we have established(-ing) upstream connections. However, at first glance the code looks very strange (how can we have active connections if there is no a listener?), so please write a comment for the code.

PR #725, opened 2017-04-20
Installer script

ikoveshnikov:
Script for easy installing Tempesta from latest github releases.

### DIFF
--- a/pkg/scripts/tempesta_installer.sh
+++ b/pkg/scripts/tempesta_installer.sh
@@ -101,6 +101,11 @@ tfw_install()
 	mkdir -p $DOWNLOAD_DIR
 
 	tfw_install_deps
+	# TempestaFW is shipped as DKMS module. By default, dkms framework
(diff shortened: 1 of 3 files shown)


pkg/scripts/tempesta_installer.sh
```diff
@@ -0,0 +1,169 @@
...
+}
+
+tfw_install_packages()
+{
+	repo="${1}"
+	shift
+	files=("${@}")
+
+	echo "Downloading latest packages from github.com/$GITHUB_USER/$repo ..."
+	mkdir -p $DOWNLOAD_DIR/$repo
```
krizhanovsky: Ok, but just mention in the Wiki that there will be `tfw_downloads` directory when the script finishes and that a user should have enough disk space (how much?) to download the packages. (One might think to run it from /tmp, which is sometimes is mounted in ramfs and is limited in size.)

PR #727, opened 2017-04-20
Linux packaging

ikoveshnikov:
* add linux kernel packaging for debian 9: `/pkg/linux/debian`
* move `/pkg/debian` to `/pkg/tempesta/debian`
* add packaging readme


ikoveshnikov:
As discussed, i put debian packaging for linux kernel into `/pkg/`. That doesn't seems to be a good solution: 1.5K files and 0.5M lines added by this patch. Storing TempestaFW packaging in `/pkg/` doesn't
burden the repo and significantly simplify packaging.

Isn't it better to move linux packaging to separate repo, like i have done it originally: https://github.com/ikoveshnikov/linux-tfw-package ?

PR #734, opened 2017-05-23
Speed up hash scheduler

ikoveshnikov:
Note: patch highly rely on changes in ab-alb-2 branch, so pull request is done to that branch. Changes:
* functional tests is updated to support ratio `scheduler`
* Hash scheduler uses sorted array of connections, this improves performance (O(lg n)).
* It is possible now to _get_ connection as soon as it found. No reruns of scheduling are needed.
* Calculation of connection hashes was improved to avoid collisions.

Drawbacks:
* Assigning server group to hash scheduler is rather slow: creation of sorted array is implemented in naive and not optimal way.


tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -175,109 +211,127 @@ tfw_sched_hash_get_sg_conn(TfwMsg *msg, TfwSrvGroup *sg)
...
-	TfwHashSrvList *sl;
+	TfwHashConnList *cl;
 
 	if (unlikely(!sg->srv_n || list_empty(&sg->srv_list)))
 		return -EINVAL;
 
-	size = sizeof(TfwHashSrvList) + sizeof(TfwHashSrv) * sg->srv_n;
+	list_for_each_entry(srv, &sg->srv_list, list) {
+		conn_n += srv->conn_n;
+	}
```
krizhanovsky: No need braces for single line loop body.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -175,109 +211,127 @@ tfw_sched_hash_get_sg_conn(TfwMsg *msg, TfwSrvGroup *sg)
...
-		hash = hsrv->hash;
-		list_for_each_entry(srv_conn, &srv->conn_list, list) {
-			if (unlikely(ci++ == srv->conn_n))
+		TfwSrvConn *conn;
+		unsigned long srv_hash = __calc_srv_hash(srv);
+
+		size = sizeof(TfwHashConnList)
+				+ sizeof(TfwHashConn) * srv->conn_n;
+		if (!(srv->sched_data = kzalloc(size, GFP_KERNEL)))
+			return -ENOMEM;
```
krizhanovsky: You can calculate total required amount of memory in first loop over `sg->srv_list`, so the whole memory region can be allocated at once and just set pointers to required `srv->sched_data`'s. This will simplify `tfw_sched_hash_del_grp()` to just one `kfree()` call as it was before.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -175,109 +211,127 @@ tfw_sched_hash_get_sg_conn(TfwMsg *msg, TfwSrvGroup *sg)
...
+
+		list_for_each_entry(conn, &srv->conn_list, list) {
+			unsigned long hash;
+retry:
+			seed += 1;
+			hash = hash_long(srv_hash ^ seed, BITS_PER_LONG);
+
+			ret = __add_conn(cl, conn, hash);
+			if (ret) {
+				if (ret == -EEXIST)
```
krizhanovsky: Only `-EEXIST` is returned from `__add_conn()` as error code, so no need for `ret` variable.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -175,109 +211,127 @@ tfw_sched_hash_get_sg_conn(TfwMsg *msg, TfwSrvGroup *sg)
...
+
+		size = sizeof(TfwHashConnList)
+				+ sizeof(TfwHashConn) * srv->conn_n;
+		if (!(srv->sched_data = kzalloc(size, GFP_KERNEL)))
+			return -ENOMEM;
+
+		list_for_each_entry(conn, &srv->conn_list, list) {
+			unsigned long hash;
+retry:
+			seed += 1;
```
krizhanovsky: I think it's better to initialize `seed` by a relatively large prime and increment it by smaller prime. That will get us better bits mixture generating more uniform distribution of hash values.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -175,109 +211,127 @@ tfw_sched_hash_get_sg_conn(TfwMsg *msg, TfwSrvGroup *sg)
...
+
+static int
+__add_conn(TfwHashConnList *cl, TfwSrvConn *conn, unsigned long hash)
+{
+	ssize_t idx;
+	TfwHashConn new_hcon = {hash, conn};
+
+	idx = __bsearch(hash, cl);
+	if (idx < 0) {
+		/* @cl is empty. */
```
krizhanovsky: It seems you can just check `cl->conn_n` before calling `__bsearch()` - that will make the code simpler. Probably it has sense to check `cl->conn_n` outside of `__basearch()`.

tempesta_fw/t/unit/sched_helper.c
```diff
@@ -73,6 +73,12 @@ test_start_sg(TfwSrvGroup *sg, const char *sched_name, unsigned int flags)
 	kernel_fpu_end();
 
 	sg->flags = flags;
+	/* Adjust servers weights for static ratio scheduler. */
+	if (!strcmp(sched_name, "ratio")
+	    && (sg->flags & TFW_SG_F_SCHED_RATIO_STATIC))
```
keshonok: At configuration time, server weights are set/adjusted for any variation of Ratio scheduler, not just static. For Dynamic and Predictive variations these server weights are initial weights before any dynamic data is collected.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -138,34 +141,67 @@ __find_best_conn(TfwSrvConn **best_conn, TfwHashSrv *srv_cl,
...
+				: ULONG_MAX;
+		unsigned long r_diff = (l_idx >= 0)
+				? (cl->conns[r_idx].hash - best_hash)
+				: ULONG_MAX;
+		ssize_t best_idx = (l_diff <= r_diff) ? l_idx : r_idx;
+
+		conn = cl->conns[best_idx].conn;
+		if (likely(!tfw_srv_conn_restricted(conn)
+			   && !tfw_srv_conn_queue_full(conn)
+			   && tfw_srv_conn_get_if_live(conn)))
```
keshonok: This condition is repeated twice. Perhaps, it makes sense to make an inline function with it? Like `tfw_sched_hash_conn_suitable()`?

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -175,115 +211,123 @@ tfw_sched_hash_get_sg_conn(TfwMsg *msg, TfwSrvGroup *sg)
...
+		scl->conns[scl->conn_n] = *hconn;
+		++scl->conn_n;
+	}
 }
 
 static int
 tfw_sched_hash_add_grp(TfwSrvGroup *sg)
 {
-	int ret = -EINVAL;
-	size_t size, si, ci;
```
keshonok: `si` and `ci` were used to verify that that data in `sg` is self-consistent. That was for protection against possible future changes in `sock_srv.c` by someone who didn't quite realize that `sg` must be completely formed before adding it to a scheduler. This verification is pretty cheap calculation-wise. It is really so unnecessary that it required the removal?

PR #740, opened 2017-05-30
Support for new HTTP methods

### LINKED ISSUE #730: Full support of HTTP methods
Currently Tempesta FW supports only GET, POST, HEAD and PURGE methods. However, Tempesta FW must be able to service more exotic applications, e.g. requiring PROPFIND methods, so more extensive set of methods must be supported. At least. methods from [RFC 7231 4.1](https://tools.ietf.org/html/rfc7231#section-4.1) and [RFC 7237](https://tools.ietf.org/html/rfc7237) must be implemented. The new methods, which we don't know how to process, must be just forwarded to an upstream.

ikoveshnikov:
* fix #730 
* fix parsing split Purge requests (in segments: <PURG> <E...>)

### DIFF
--- a/tempesta_fw/cache.c
+++ b/tempesta_fw/cache.c
@@ -173,20 +173,28 @@ static TfwStr g_crlf = { .ptr = S_CRLF, .len = SLEN(S_CRLF) };
 
 /*
  * The mask of non-cacheable methods per RFC 7231 4.2.3.
- * Currently none of the non-cacheable methods are supported.
- * Note: POST method is cacheable but not supported at this time.
+ * Safe methods that do not depend on a current or authoritative response
+ * are defined as cacheable: GET, HEAD, and POST.
+ * Note: caching of POST method responces is not supported at this time.
  */
-static unsigned int tfw_cache_nc_methods = (1 << TFW_HTTP_METH_POST);
+static unsigned int tfw_cache_nc_methods =
+		~((1 << TFW_HTTP_METH_GET) | (1 << TFW_HTTP_METH_HEAD));
 
 static inline bool
 __cache_method_nc_test(tfw_http_meth_t method)
 {
+	BUILD_BUG_ON(sizeof(tfw_cache_nc_methods) * BITS_PER_BYTE
+		     < _TFW_HTTP_METH_COUNT);
+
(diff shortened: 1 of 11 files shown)


tempesta_fw/http_parser.c
```diff
@@ -186,6 +186,18 @@ do {									\
 #define __FSM_MATCH_MOVE(alphabet, to)	__FSM_MATCH_MOVE_f(alphabet, to, \
```
krizhanovsky: Please update header comment to copyright in 2017.

tempesta_fw/http_parser.c
```diff
@@ -2092,15 +2169,101 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
...
+				if (likely(__data_available(p, 6))
+				    && (*(p + 4) == 'C') && (*(p + 5) == 'K'))
+				{
+					req->method = TFW_HTTP_METH_UNLOCK;
+					__FSM_MOVE_nofixup_n(Req_MUSpace, 6);
+				}
+				__FSM_MOVE_nofixup_n(Req_MethUnlo, 4);
 			}
-			TFW_PARSER_BLOCK(Req_Method); /* Unsupported method */
+			/* TODO: unknown */
```
krizhanovsky: What's the sense of the TODO comment? We're falling to unknown method and validate it's alphabet - this seems the right logic, what we need to do more?

tempesta_fw/cache.c
```diff
@@ -1399,7 +1407,9 @@ tfw_cache_cfg_method(TfwCfgSpec *cs, TfwCfgEntry *ce)
 	unsigned int i, method;
 	const char *val;
 
-	BUILD_BUG_ON(sizeof(cache_cfg.methods) * 8 < _TFW_HTTP_METH_COUNT);
+	BUILD_BUG_ON(sizeof(cache_cfg.methods) * BITS_PER_BYTE
+		     < _TFW_HTTP_METH_COUNT);
```
keshonok: My understanding of `BUILD_BUG_ON()` is that it's evaluated at compilation time, so using it once per a condition is sufficient.

PR #741, opened 2017-05-30
Fix a memory leak in implementation of pool

keshonok:
The implementation keeps a handful number of pages cached.
The cache was not cleared when Tempesta was unloaded.

### DIFF
--- a/tempesta_fw/pool.c
+++ b/tempesta_fw/pool.c
@@ -254,6 +254,15 @@ tfw_pool_init(void)
 void
 tfw_pool_exit(void)
 {
+	int i;
+
+	for_each_online_cpu(i) {
+		unsigned int pgn = per_cpu(pg_next, i);
+		unsigned long *pgc = (unsigned long *)per_cpu_ptr(pg_cache, i);
+		while (pgn--)
+			free_page(pgc[pgn]);
+	}
+
 	free_percpu(pg_cache);
 }
 


tempesta_fw/pool.c
```diff
@@ -254,6 +254,17 @@ tfw_pool_init(void)
 void
 tfw_pool_exit(void)
 {
+	int icpu;
+
+	for_each_online_cpu(icpu) {
+		unsigned int pgn = per_cpu(pg_next, icpu);
+		while (pgn--) {
+			unsigned long *pgc = (unsigned long *)
+					     per_cpu_ptr(pg_cache, icpu);
```
krizhanovsky: Usually following formatting looks better if there is not enough space:

```C
        unsigned long *pgc;
        pgc = (unsigned long *)per_cpu_ptr(pg_cache, icpu);
```

nmmmnu:
Strange but this code looks fishy :) Something is not right

for_each_online_cpu(icpu) {
unsigned int pgn = per_cpu(pg_next, icpu);
while (pgn--) {
unsigned long *pgc = (unsigned long *)
    per_cpu_ptr(pg_cache, icpu);
free_page(pgc[pgn]);
}
}

correct me but:

*unsigned long *pgc = (unsigned long *) per_cpu_ptr(pg_cache, icpu);*

does not depends of while loop.

correct version must be something like this:

for_each_online_cpu(icpu) {
unsigned int pgn = per_cpu(pg_next, icpu);
unsigned long *pgc = (unsigned long *) per_cpu_ptr(pg_cache, icpu);

while (pgn--)
free_page(pgc[pgn]);
}

sure modern optimizer will pull this outside of the loop.

krizhanovsky:
@nmmmnu thank you for good remark!

PR #742, opened 2017-06-01
Fix __read_mostly section collision

ikoveshnikov:
const __read_mostly is not legal and causes section type conflicts.
That's because the read.mostly section is not read only.

GCC 7 raises build error on this line.

### DIFF
--- a/tempesta_fw/vhost.c
+++ b/tempesta_fw/vhost.c
@@ -24,7 +24,7 @@
 #include "str.h"
 
 /* Mappings for match operators. */
-static TfwCfgEnum const __read_mostly tfw_match_enum[] = {
+static const TfwCfgEnum tfw_match_enum[] = {
 	{ "*",		TFW_HTTP_MATCH_O_WILDCARD },
 	{ "eq",		TFW_HTTP_MATCH_O_EQ },
 	{ "prefix",	TFW_HTTP_MATCH_O_PREFIX },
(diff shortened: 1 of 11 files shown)


nmmmnu:
[0 ... _TFW_HTTP_VER_COUNT] = NULL,

this is GNU extension why not do it without this?

some like solution below.

Also string literals, e.g. "1.0 " and others are of type "const char *".
Are you going to modify them in some way?

Also, at least for C++ (not sure for C), accessing static variable is done
by lock.
You can just pull this static variable s_http_version into .h file or make
it global constant into .c file.
Because it is static, it is global anyway.

also seems like global variables are guaranteed to be zeroing so you can do
just -
note this will be initialized in compile time and will be generally faster
than current solution.

const char *s_http_version[] = { [TFW_HTTP_VER_09] = "0.9 " ,
[TFW_HTTP_VER_10] = "1.0 " , [TFW_HTTP_VER_11] = "1.1 " , [TFW_HTTP_VER_20]
= "2.0 " };
https://gist.github.com/nmmmnu/07c3acbb2729dce9d74867fdb7489baa

nmmmnu:
`         .ptr = (void *)("\0   0.9 1.0 1.1 2.0 " + hm->version * 4),
`
Man this is just awful code.  :)
But there is one more problem, if new protocol is added in the future, how you will handle it.
but you can do something like:

```
       char x[4][] = {
               "\n\n\n\n",
              "0.9",
              "1.0",
              //...
       }

```
at least is more readable

tempesta_fw/vhost.c
```diff
@@ -105,7 +105,7 @@ static TfwAddr		tfw_capuacl[TFW_CAPUACL_ARRAY_SZ];
  * Note that @loc_dflt in the default vhost serves as global
  * default caching policy.
  */
-static char const __read_mostly s_hdr_via_dflt[] =
+static char const s_hdr_via_dflt[] =
```
keshonok: Perhaps, `const char`, for consistency? Not that it really matters.

PR #744, opened 2017-06-05
Functional test: reboot under high load

### LINKED ISSUE #738: Functional test for stress restart
Current test t/functional/regression/test_shutdown.py verifies shutdown in very mercy conditions. Meantime, it should restart Tempesta (`tempesta.sh --restart`) many times under heavy client load. Servers (there should be at least several servers as well) also must send responses. Cache and proxy modes must be tested.

ikoveshnikov:
fix #738

### DIFF
--- a/tempesta_fw/t/functional/regression/test_reboot.py
+++ b/tempesta_fw/t/functional/regression/test_reboot.py
@@ -4,7 +4,8 @@ Test TempestaFW reeboot under heavy load.
 
 from __future__ import print_function
 import unittest
-import threading
+from threading import Thread
+from time import sleep
 from helpers import deproxy, tf_cfg, tempesta, remote, control
 from testers import stress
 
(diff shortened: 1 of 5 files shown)


tempesta_fw/t/functional/regression/test_reboot.py
```diff
@@ -0,0 +1,82 @@
...
+        # Set defconfig for Tempesta.
+        self.tempesta.config.set_defconfig(config)
+        self.configure_tempesta()
+        control.servers_start(self.servers)
+        self.wrk.prepare()
+
+        for i in range(self.restart_cycles):
+            self.tempesta.start()
+            threading.Timer(self.duration, self.tempesta.stop).start ()
+            control.client_run_blocking(self.wrk)
```
krizhanovsky: It seems you restart wrk on each iteration. Giving that there is only 1 second delay, we won't have significant queues sizes, so probably the conditions are still relatively mercy. I'd suggest to run heavy wrk, wait few seconds, and start the restarting loop with random delay in range [1..20] seconds.

PR #773, opened 2017-07-25
http_parser: bring return code into __FSM_EXIT macro

milabs:
Please, review.


tempesta_fw/http_parser.c
```diff
@@ -238,9 +237,8 @@ do {									\
 		__msg_hdr_chunk_fixup(data, len);			\
 		parser->_i_st = to;					\
 		__fsm_const_state = to;					\
-		r = TFW_POSTPONE;					\
 		finish;							\
```
ikoveshnikov: `finish` can be a conditional statement that changes `r` to `TFW_BLOCK`. This used at least once, so change here or where the `__FSM_I_MATCH_MOVE_finish` macro is used is needed.

milabs: Right, now I see the problem

PR #784, opened 2017-08-02
Fix possible but unlikely  TfwSrvConn.qsize overflow

ikoveshnikov:
The overflow is very unlikely but still may happen.

`server_queue_size 0;` in TempestaFW configuration means "unlimited queue size" and sets
`TfwSrvGroup.max_qsize` to `UINT_MAX`. New requests can be pushed into `TfwSrvConn.fwd_queue`
until queue size is smaller than `max_qsize`. Queue size may overflow the maximum allowed size, but
overflowing  `UINT_MAX` cause unsigned integer overflow, so queue size will have wrong value.

Could happen if `server_queue_size` is set to `0` and size of fwd_queue has reached `UINT_MAX`.
The issue is very unlikely, almost impossible.

### DIFF
--- a/tempesta_fw/sock_srv.c
+++ b/tempesta_fw/sock_srv.c
@@ -945,7 +945,7 @@ tfw_cfgop_setup_srv_group(void)
 	BUG_ON(!tfw_cfg_sg);
 	BUG_ON(!tfw_cfg_sched);
 
-	tfw_cfg_sg->max_qsize = tfw_cfg_queue_size ? : UINT_MAX;
+	tfw_cfg_sg->max_qsize = tfw_cfg_queue_size ? : INT_MAX;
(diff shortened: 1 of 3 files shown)


tempesta_fw/sock_srv.c
```diff
@@ -945,7 +945,7 @@ tfw_cfgop_setup_srv_group(void)
 	BUG_ON(!tfw_cfg_sg);
 	BUG_ON(!tfw_cfg_sched);
 
-	tfw_cfg_sg->max_qsize = tfw_cfg_queue_size ? : UINT_MAX;
+	tfw_cfg_sg->max_qsize = tfw_cfg_queue_size ? : INT_MAX;
```
krizhanovsky: Please add a comment which we've discussed in the chat.

Issue #787, opened 2017-08-07
[Cache]: Invalidating of cache entries by sending PURGE doesn't work


ikoveshnikov:
The issue was introduced when dependency to the request method (`req->meth`) was injected into `tfw_http_req_key_calc()`. So when trying to purge cache entry TempestaFW actually tries to remove cached reply to `PURGE` request to origin server.

Steps to reproduce:
1. configure TempestaFW to support cache and PURGE, e.g.:
```
server 192.168.122.12:8080;

cache 2;
cache_fulfill * *;
cache_methods GET HEAD;
cache_purge;
cache_purge_acl 192.168.122.1;
```
2. send GET to origin to populate cache on TempestaFW
3. send GET once again to see that request is served from cache
4. send PURGE to the same resource. You will get 404
5. send GET once again. Request is still served from cache

ikoveshnikov:
Fixed in 86585a4cd6ab525d8b099c1b2e3523f16688d83f

Issue #788, opened 2017-08-07
[Cache] Unable to serve from cache if cache entry was expired or invalidated


ikoveshnikov:
The issue is here: https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/cache.c#L852

`__cache_add_node()` must revalidate or remove already existing entry to the `key` given instead of unconditional addition of a new cache entry. The new cache entry will have the same key as the expired one, so the new entry will be added into one collision chain with expired entry. In most cases it will be added to the end of the chain.

This make impossible to get not expired cache entry when serving new requests. And response to every request of that resource will be added into cache. That leads to memory leakage.

Steps to reproduce:
1. Configure backend to have small expire timeout and disable caching on backend. E.g. for nginx:
```
expires 15; # 15 seconds
```
2. Send GET to TempestaFW. It will be forwarded to origin server, response will populate the cache
3. Send GET once again, it will be served from cache
4. Wait 15-20 secconds, cache entry will expire
5. Send GET, It will be forwarded to origin server, response will populate the cache
6. Send GET once again, It will be forwarded to origin server, response will populate the cache. That is the issue, the request must be served from cache

Note: there is a few more tasks on cache revalidation: #515 #518

Issue #782, opened 2017-08-01
Kernel BUG at tempesta_fw/sock_srv.c:520 while running regression.test_reboot.RebootUnderLoadTest on QEMU


intelfx:
Simply running all tests (`./run_tests.py`) with `tests_config.ini` copied from the sample on QEMU triggers a chain of two BUGs:
- one `kernel BUG at /root/tempesta/tempesta_fw/sock_srv.c:520!` and
- consequently a `BUG tfw_srv_conn_cache (Tainted: G      D    O   ): Objects remaining in tfw_srv_conn_cache on __kmem_cache_shutdown()`.

This semi-reliably happens while running `regression.test_reboot.RebootUnderLoadTest`.

Configuration:
- QEMU, 4C SMP on 4C8T CPU
- both the client, the servers and Tempesta on the same host
- intelfx/tempesta@4f010fef4356a6f4e35dca03d5d270d8dd1a6b38 (branch [`crash/sock_srv.c-520`](https://github.com/intelfx/tempesta/tree/crash/sock_srv.c-520))

Log excerpt (http://ix.io/yOS):
```
[23398.445468] [tdb] Start Tempesta DB
[23398.451660] [tempesta] Initializing Tempesta FW kernel module...
[23398.455965] [tempesta] Registering new scheduler: hash
[23398.458404] [tempesta] Registering new scheduler: http
[23398.460842] [tempesta] Registering new scheduler: ratio
[23401.937333] [tempesta] Starting all modules...
[23401.944078] [tdb] Opened table /opt/tempesta/db/filter.tdb: size=16777216 rec_size=20 base=ffff8e7161800000
[23401.945307] [tempesta] Open listen socket on: 0.0.0.0
[23401.946025] [tempesta] modules are started
[23412.007301] [tempesta] Stopping all modules...
[23412.020273] [tdb] Close table 'filter0.tdb'
[23412.021137] ------------[ cut here ]------------
[23412.021884] kernel BUG at /root/tempesta/tempesta_fw/sock_srv.c:520!
[23412.022709] invalid opcode: 0000 [#1] SMP
[...]

PR #789, opened 2017-08-07
Fix PURGE requests

ikoveshnikov:
Fix invalidating responses with PURGE requests, issue #787. Corresponding functional test is provided.

* Dependency to request method in request key generation is removed. This have to positives outcomes:
    * All cached responses for the same resource is added to TDB with the same key, and stored in the same collision chain. That makes much easier to process PURGE request.
    * The dependency actually broke `hash` scheduler: different methods for the same resource may be scheduled to different servers. So this is fixed.
* Fix getting the next record in collision chain in `htrie`

### DIFF
--- a/tempesta_db/core/htrie.c
+++ b/tempesta_db/core/htrie.c
@@ -869,6 +869,12 @@ next_bckt:
 		if (*b) {
 			read_lock_bh(&(*b)->lock);
 			r = TDB_HTRIE_BCKT_1ST_REC(*b);
+
+			if (r && tdb_live_rec(dbh, r) && r->key == key) {
+				read_unlock_bh(&_b->lock);
+				/* Unlock the bucket by tdb_rec_put(). */
+				return r;
+			}
 		}
 		read_unlock_bh(&_b->lock);
 		_b = *b;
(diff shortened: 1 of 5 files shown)


ikoveshnikov:
Although corresponding functional test is provided, it fails due to several issues: first it fails when trying to serve from cache request which was previously invalidated - issue #788 , then Tempesta is crushed due to issue #782. 

It also can fail due to mismach of `Date:` header, the situation is discussed with @intelfx during his work on #782 . This happen quite frequent for me. Here is a workaround, but i can't provide it as final solution:
```diff
--- a/tempesta_fw/t/functional/helpers/deproxy.py
+++ b/tempesta_fw/t/functional/helpers/deproxy.py
@@ -136,8 +136,8 @@ class HeaderCollection(object):
         return headers
 
     def __eq__(self, other):
-        h_self = set([(hed.lower(), val) for hed, val in self.items()])
-        h_other = set([(hed.lower(), val) for hed, val in other.items()])
+        h_self = set([(hed.lower(), val) for hed, val in self.items() if hed.lower() != 'date'])
+        h_other = set([(hed.lower(), val) for hed, val in other.items() if hed.lower() != 'date'])
         return h_self == h_other
 
     def __ne__(self, other):
```

PR #790, opened 2017-08-07
Introduce SKB dumping functions that are helpful in debugging.

keshonok:


krizhanovsky:
This is already done in https://github.com/tempesta-tech/tempesta/commit/54ecbed4edc748f37b2bcf8e887484660ee4f205#diff-6892ca70f781b6c0fb03c6b20fc4cd0eR1145 . I doubt if skb consistency is usable by others. Anyway, if it is, then it should be done as an incremental patch to the pointed out patch.

Issue #797, opened 2017-08-08
run_tests.py: mirror logs (esp. failure tracebacks and exceptions) into a file by default


intelfx:
The functional test framework should mirror test logs (including exceptions) into a file for convenience.

krizhanovsky:
There must be a configuration option where write the messages to.

Issue #795, opened 2017-08-08
run_tests.py: negative selection + execution status persistence and/or "start from given test"


intelfx:
It would be nice for `run_tests.py` to support:

* negative selection (`./run_tests.py sessions -sessions.test_cookies`)
* persistence (`./run_tests.py` *[writes a status file]* ⇒ crash ⇒ `./run_tests.py --continue`)
* "start from given point" (`./run_tests.py --start-from flacky_net`)

krizhanovsky:
As #794 occurred it has sense to skip the failing test_hash_scheduler, all tests for hash scheduler or even all tests for all schedulers, but there is no such opportunity. It's very inconvenient to account all available tests (BTW what's the best way to do this?) and point all of them out in the test command line. Need to develop all the types of the negative selection of tests.

intelfx:
Follow-ups:

* accept filename-style testcase names (`./run_tests.py sessions/test_cookies.py`) to allow globs
* add a mode to list all matched tests without executing them (`./run_tests.py --list-only [ARGUMENTS...]`)
* accept directories (`./run_tests.py sessions`) — currently they are accepted, but actually skipped as empty modules (looks like sub-modules are not taken into account by unittest framework)

Issue #793, opened 2017-08-08
run_tests.py requires dummy tests_config.ini


krizhanovsky:
As described in the [Wiki](https://github.com/tempesta-tech/tempesta/tree/master/tempesta_fw/t/functional#configuration) `./run_tests.py -d` generates default config file and it actually does so:

        $ touch tests_config.ini
        $ ./run_tests.py -d
        Default configuration saved to /root/tempesta.759/tempesta_fw/t/functional/helpers/../tests_config.ini

However, if there is no tests_config.ini, then the script is unable to generate it as well as to show help:

        $ rm tests_config.ini
        $ ./run_tests.py -d
        Tests configuration error! Configuration file "tests_config.ini" not found. Exiting.
        $ ./run_tests.py -h
        Tests configuration error! Configuration file "tests_config.ini" not found. Exiting.

ikoveshnikov:
#783 introduces a better way to start only specified tests: `run_tests.py` becomes the entry point to run specified tests, so just disable configuration checking on module import: https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/t/functional/helpers/tf_cfg.py#L109  It was needed to use testcase modules as stand-alone programs.

PR #799, opened 2017-08-15
Further fixes to functional tests

### LINKED ISSUE #792: [Tests]: base_message_chain() should provide chain that suit given method
`testers.functional.base_message_chain()` provide base message chain for specified URI and method. Currently the function just generates common GET chain and just gives possibility to redefine URI and method. Instead it should generate common chain for the given method: i.e. HEAD chain must have empty body in responses, POST chain have non-empty body in requests and empty body in responses.

Also generation of cache and proxy chains should be moved from cache tests to the same helper module.

intelfx:
Fixes #792, #793, #795, #797 and various non-reported problems.


tempesta_fw/t/functional/cache/test_purge.py
```diff
@@ -60,10 +59,9 @@ def proxy_chain(method, uri):
...
 
 def purge_chain(uri):
-    chain = functional.base_message_chain(uri=uri, method='PURGE')
-    chain.no_forward()
+    chain = functional.base_message_chain(uri=uri, method='PURGE', forward=False)
     headers = [
         'Connection: keep-alive',
         'Content-Length: 0']
-    chain.response = deproxy.Response.create(200, headers, date=True, body='')
+    chain.response = deproxy.Response.create(200, headers, date=chain.response.headers["Date"], body='')
```
ikoveshnikov: The logic should be moved into `chains.base()`.

tempesta_fw/t/functional/helpers/deproxy.py
```diff
@@ -135,9 +144,43 @@ def from_stream(rfile, no_crlf=False):
...
+        return len(self.headers.get('date', [])) == 1
+
     def __eq__(self, other):
-        h_self = set([(hed.lower(), val) for hed, val in self.items()])
-        h_other = set([(hed.lower(), val) for hed, val in other.items()])
+        h_self = self._as_dict_lower()
+        h_other = other._as_dict_lower()
+
+        if self.is_expected and other.is_expected:
+            print("HeaderCollection: comparing two is_expected=True\n" +
```
ikoveshnikov: Why not just raise framework error on these two cases?

tempesta_fw/t/functional/selftests/test_headercollection.py
```diff
@@ -101,13 +101,22 @@ def test_is_equal(self):
...
         self.assertFalse(self.headers != reorderd)
 
+        same_keys_reorderd = deproxy.HeaderCollection()
+        same_keys_reorderd.add('C', 'zxcv')
+        same_keys_reorderd.add('A', 'uiop')
+        same_keys_reorderd.add('A', 'jkl;')
+        same_keys_reorderd.add('A', 'qwerty')
+        same_keys_reorderd.add('B', 'asdf')
+        self.assertTrue(self.headers != same_keys_reorderd)
+        self.assertFalse(self.headers == same_keys_reorderd)
```
ikoveshnikov: Please also add test, where some singular header will be repeated twice. That header collection must not be equal to original one.

tempesta_fw/t/functional/helpers/chains.py
```diff
@@ -0,0 +1,140 @@
...
+    # Build a request
+    #
+
+    request_headers = [
+        'Host: %s' % tf_cfg.cfg.get('Tempesta', 'ip'),
+        'User-Agent: curl/7.53.1',
+        'Connection: keep-alive',
+        'Accept: */*'
+    ]
+    request = deproxy.Request.create(method, request_headers, uri=uri)
```
intelfx: I don't quite understand what you want me to do — the GET chain, as basic as it gets, is still valid for every method.

tempesta_fw/t/functional/cache/test_purge.py
intelfx: It is already there, I just forgot to drop the obsolete helpers from this testcase.

tempesta_fw/t/functional/helpers/deproxy.py
intelfx: Because there are selftests on HeaderCollection that do not call `set_expected()`. I did not figure out how to solve this cleanly, so I'll just remove the check altogether. I could monkey-patch HeaderCollection inside selftests, but it is not re-imported...

tempesta_fw/t/functional/selftests/test_headercollection.py
intelfx: OK

tempesta_fw/t/functional/helpers/deproxy.py
intelfx: Nevermind, done.

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -309,7 +309,7 @@ def stop(self):
         tf_cfg.dbg(3, '\tStoping Nginx on %s' % self.get_name())
         pid_file = os.path.join(self.workdir, self.config.pidfile_name)
         config_file = os.path.join(self.workdir, self.config.config_name)
-        cmd = '[ -f %s ] && kill -s TERM $(cat %s)' % (pid_file, pid_file)
+        cmd = '[[ -e %s ]] && pid=$(cat %s) && kill -s TERM $pid && while [[ -e /proc/$pid ]]; do sleep 0.5; done' % (pid_file, pid_file)
```
intelfx: OK

tempesta_fw/t/functional/helpers/deproxy.py
```diff
@@ -382,12 +430,12 @@ def update(self):
 
     @staticmethod
     def create(status, headers, version='HTTP/1.1', date=False,
-               srv_version=None, body=None):
+               srv_version=None, body=None, body_void=False):
```
intelfx: This way we'll skip sanity check in `helpers.HttpMessage.read_sized_body()` if an empty body is incorrectly passed together with a non-zero `Content-Length: ` HTTP header. I'd leave this an explicit parameter.

ikoveshnikov: Ok, I see the point now.

intelfx:
Force-pushed to add missing `.gitignore` entries as requested.

Issue #529, opened 2016-05-26
[RFC 7232] ETag


krizhanovsky:
Tempesta must process [Etag](https://tools.ietf.org/html/rfc7232#section-2.3) header and store its value in `TfwCacheEntry`.

ikoveshnikov:
Closed  in 21bc2f1fdbde214259c83966e5c9598002107210

Issue #519, opened 2016-05-25
[RFC 7234] Process conditional requests for cache entries revalidation


krizhanovsky:
Tempesta must process `If-Modified-Since` header, see [RFC 7234 4.3.2](https://tools.ietf.org/html/rfc7234#section-4.3.2). `If-Match` and `If-None-Match` must be parsed as well since they affect `If-Modified-Since` processing, but their processing is subject for upcoming issues.

Depends on #1 in sending 111'st warning and #529 (ETag).

ikoveshnikov:
Closed in 21bc2f1fdbde214259c83966e5c9598002107210

PR #801, opened 2017-08-16
Handle conditional requests

ikoveshnikov:
* #529 : ETag is added to cache entry to allow handling validation and range requests
* #519 : Process Conditional (If-Modified-Since and If-None-Match) requests, respond with 304 or 412 code

More unit and functional tests to come in a couple of days.

### DIFF
--- a/tempesta_fw/str.c
+++ b/tempesta_fw/str.c
@@ -304,9 +304,10 @@ EXPORT_SYMBOL(tfw_strcat);
  * headers until ':', so plain C is Ok for now.
  *
  * Returns:
- *   0 - strings match;
- *   1 - strings match and @stop is found;
- *  -1 - strings do not match;
+ *   0		- strings match;
+ *   INT_MAX	- strings match and @stop is found;
+ *   -1		- strings do not match, @s1 < @s2;
+ *   1		- strings do not match, @s1 > @s2;
  */
 static inline int
 __cstricmpspn(const unsigned char *s1, const unsigned char *s2, int n, int stop,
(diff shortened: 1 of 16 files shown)


tempesta_fw/str.c
```diff
@@ -582,6 +594,43 @@ tfw_str_to_cstr(const TfwStr *str, char *out_buf, int buf_size)
...
+ *
+ * Return compaund TfwStr starting at next string value.
+ */
+TfwStr
+tfw_str_next_str_val(const TfwStr *str)
+{
+	TfwStr r_str = { 0 }, *chunk, *end;
+	bool skip = true;
+
+	if (!str || TFW_STR_DUP(str) || TFW_STR_PLAIN(str))
```
krizhanovsky: I guess `TFW_STR_DUP`'licated string passes as an argument is a bug.

tempesta_fw/t/functional/cache/test_conditional.py
```diff
@@ -0,0 +1,161 @@
...
+from __future__ import print_function
+from helpers import tf_cfg, deproxy
+from testers import functional
+
+__author__ = 'Tempesta Technologies, Inc.'
+__copyright__ = 'Copyright (C) 2017 Tempesta Technologies, Inc.'
+__license__ = 'GPL2'
+
+# TODO: Check that all headers that must be present in 304 response are there.
+# Not implemented since some of them (at least Vary) affect cache behaviour.
```
krizhanovsky: Please add the item to #810

PR #808, opened 2017-08-19
Fix #768: Set limit for digits count in 'chunk-size' field of chunked request.

### LINKED ISSUE #768: Chunked Transfer Coding - chunk-size decoding issue
As per RFC 7230 (4.1):

~~~
     chunked-body   = *chunk
                      last-chunk
                      trailer-part
                      CRLF

     chunk          = chunk-size [ chunk-ext ] CRLF
                      chunk-data CRLF
     chunk-size     = 1*HEXDIG
     last-chunk     = 1*("0") [ chunk-ext ] CRLF

     chunk-data     = 1*OCTET ; a sequence of chunk-size octets
~~~

Tempesta uses [parse_int_hex](https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/http_parser.c#L483) to decode the length. But that function doesn't track the length of the size field itself. So, it's possible to use any amount of leading zeroes ('0') and although it's allowed by RFC this can be used to cause kind of DOS attack:

~~~
000...000100500
bla-bla...
0000....0000....000...
~~~

The script to check the issue:
~~~
import socket

s = socket.socket()
s.connect(('tempesta-tech.com', 80))

s.send("POST / HTTP/1.1\n")
s.send("Host: tempesta-tech.com\n")
s.send("Transfer-Encoding: chunked\n")
s.send("\n")
s.send("5\nHELLO\n")

while True:
    s.send('0')
~~~

aleksostapenko:


tempesta_fw/http_parser.c
```diff
@@ -489,7 +490,12 @@ parse_int_hex(unsigned char *data, size_t len, unsigned long *acc)
 			return CSTR_NEQ;
 		if (unlikely(*acc > (UINT_MAX - 16) / 16))
 			return CSTR_BADLEN;
+		if (unlikely(*d_count >= (sizeof(*acc) - 1))) {
```
ikoveshnikov: You save half a byte on every read operation. Condition above allows you to store only 3 bytes. I doubt that it was a desired behaviour.

ikoveshnikov:
Please, also add unit test to `test_http_parser.c`.

tempesta_fw/http_parser.c
aleksostapenko: Corrected

aleksostapenko: Corrected to `>= sizeof(int) * 2`

tempesta_fw/http.h
```diff
@@ -260,6 +261,7 @@ typedef struct {
  */
 typedef struct {
 	unsigned short	to_go;
+	unsigned short	_d_count;
```
aleksostapenko: Corrected.

tempesta_fw/http_parser.c
```diff
@@ -489,7 +490,12 @@ parse_int_hex(unsigned char *data, size_t len, unsigned long *acc)
...
+			TFW_DBG_ADDR("Too many digits in chunk-size field", addr);
```
aleksostapenko: Debug printout and extra argument `addr` had been removed.

PR #815, opened 2017-08-23
Fix #768:  1. Replace 'int' with 'long' for chunks size. 2. Add check for 'signed long' overflow.

### LINKED ISSUE #768: Chunked Transfer Coding - chunk-size decoding issue

aleksostapenko:


tempesta_fw/http_parser.c
```diff
@@ -1006,7 +1007,7 @@ __FSM_STATE(RGen_BodyChunk) {						\
 }									\
 __FSM_STATE(RGen_BodyReadChunk) {					\
 	BUG_ON(parser->to_read < 0);					\
-	__fsm_sz = min_t(int, parser->to_read, __data_remain(p));	\
+	__fsm_sz = min_t(long, parser->to_read, __data_remain(p)); \
```
ikoveshnikov: Broken alignment of trailing `\` character.

aleksostapenko: Corrected

PR #827, opened 2017-08-30
Inherit unspecified directives from group 'default' (Fix #813)

### LINKED ISSUE #813: [Config] Overriding defaults for server groups
Default config file states that server group settings applied outside any server group overrides defaults for explicit server groups:
https://github.com/tempesta-tech/tempesta/blob/9df1d21d3ea109f8d8b504a28510ff7d348575bd/etc/tempesta_fw.conf#L59-L62

But `sock_srv.c` acts in a different way: settings defined outside any group affects only and only to a implicit server group. Wiki pages suit real behaviour and require updating. Also note, that defauts override can be stated after groups with default settings:
```
srv_group some_name {
    #no defaults override
}

#here can be server group settings that override defaults

srv_group some_name_2 {
    #no defaults override
}
```

keshonok:


tempesta_fw/sock_srv.c
```diff
@@ -680,12 +705,12 @@ tfw_cfgop_sticky_sess(TfwCfgSpec *cs, TfwCfgEntry *ce, unsigned int *use_sticky)
...
 		return -EINVAL;
 	}
-	if (!ce->val_n) {
+	if (tfw_cfg_is_dflt_value(ce)) {
+		*use_sticky = 0;
+	} else if (!ce->val_n) {
 		*use_sticky = TFW_SRV_STICKY;
 	} else if (!strcasecmp(ce->vals[0], "allow_failover")) {
 		*use_sticky = TFW_SRV_STICKY | TFW_SRV_STICKY_FAILOVER;
-	} else if (!strcasecmp(ce->vals[0], TFW_CFG_DFLT_VAL)) {
```
ikoveshnikov: As i see, `TFW_CFG_DFLT_VAL` is not used anywhere now, and can be replaces with `NULL`.

tempesta_fw/sock_srv.c
```diff
@@ -596,6 +596,26 @@ static int tfw_cfg_cns_retries, tfw_cfg_out_cns_retries;
...
+	bool sticky_sess	: 1;
+} __attribute__((packed)) tfw_cfg_is_set;
+
+/* Please keep the condition in these two macros in sync. */
+#define TFW_CFGOP_HAS_DFLT(ce, v)				\
+	(tfw_cfg_is_dflt_value(ce) && tfw_cfg_is_set.v)
+#define TFW_CFGOP_INHERIT(ce, v)				\
+({								\
+	if (tfw_cfg_is_dflt_value(ce) && tfw_cfg_is_set.v) {	\
+		tfw_cfg_##v = tfw_cfg_out_##v;			\
```
ikoveshnikov: Seems, that line duplicates initialisation of parameters `tfw_cfg_##v` in `tfw_cfgop_begin_srv_group()` and not needed.

As a point of future improvement: if default value is passed to given cfg entry (`tfw_cfg_is_dflt_value(ce) == true`), but global default value is not set yet (`tfw_cfg_is_set.v == false`), we can parse that value and save both to global default server group option (`tfw_cfg_out_##v`) and server group option for the current group (`tfw_cfg_##v`). That will save time for the cases when a lot of server group with default settings are defined in config file.

tempesta_fw/sock_srv.c
```diff
@@ -1226,8 +1257,14 @@ tfw_cfgop_sched(TfwCfgSpec *cs, TfwCfgEntry *ce, TfwScheduler **arg_sched,
 static int
 tfw_cfgop_in_sched(TfwCfgSpec *cs, TfwCfgEntry *ce)
 {
-	tfw_cfg_schref = &tfw_cfg_schref_predict;
+	if (TFW_CFGOP_HAS_DFLT(ce, sched)) {
+		tfw_cfg_sched = tfw_cfg_out_sched;
+		tfw_cfg_schref = tfw_cfg_out_schref;
+		tfw_cfg_sg_flags = tfw_cfg_out_sg_flags;
```
ikoveshnikov: Same here, this 3 lines duplicates `tfw_cfgop_begin_srv_group()`.

PR #835, opened 2017-09-03
Fix parsing responses not allowed to have a body

### LINKED ISSUE #816: BUG on invalid response
The bug is reported by @bartolootrit . An adjusted deproxy test returns an invalid response as

```
HTTP/1.1 204 No Content
Content-type: text/html
Connection: keep-alive
Content-Length: 138
Last-Modified: Mon, 12 Dec 2016 13:59:39 GMT
Server: Deproxy Server
Date: Wed, 23 Aug 2017 11:56:18 GMT
<html>
<head>
    <title>An Example Page</title>
</head>
<body>
    Hello World, this is a very simple HTML document.
</body>
</html>
```

making Tempesta crash at https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/http.c#L2478

ikoveshnikov:
Fix #816 

* Tag all responses not allowed to have a body with VOID_BODY flag
* Block responses with Content-Length header if it is not allowed by RFC 7230 Section 3.3.2

### DIFF
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -963,17 +963,12 @@ __FSM_STATE(RGen_BodyInit) {						\
 #define TFW_HTTP_INIT_RESP_BODY_PARSING()				\
 __FSM_STATE(RGen_BodyInit) {						\
 	TfwStr *tbl = msg->h_tbl->tbl;					\
-	TfwHttpResp *resp = (TfwHttpResp *)msg;				\
 									\
 	TFW_DBG3("parse response body: flags=%#x content_length=%lu\n",	\
 		 msg->flags, msg->content_length);			\
 									\
 	/* There's no body. */						\
-	/* TODO: Add (req == CONNECT && resp == 2xx) */			\
-	if (resp->status - 100U < 100U || resp->status == 204		\
-	    || resp->status == 304 || msg->flags & TFW_HTTP_VOID_BODY)	\
-	{								\
-		/* There is no body. */					\
+	if (msg->flags & TFW_HTTP_VOID_BODY) {				\
 		msg->body.flags |= TFW_STR_COMPLETE;			\
 		FSM_EXIT(TFW_PASS);					\
 	}								\
@@ -1219,7 +1214,7 @@ done:
 }
 
 /**
- * Parse Content-Length header value, RFC 2616 14.13.
+ * Parse Content-Length header value, RFC 7230 section 3.3.2.
  */
 static int
 __parse_content_length(TfwHttpMsg *msg, unsigned char *data, size_t len)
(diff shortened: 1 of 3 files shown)


tempesta_fw/http_parser.c
```diff
@@ -4180,6 +4185,13 @@ tfw_http_parse_resp(void *resp_data, unsigned char *data, size_t len)
 			/* Status code is fully parsed, move forward. */
 			resp->status = parser->_acc;
 			parser->_acc = 0;
+			/* Some responses are not allowed to have a body. */
+			/* TODO: Add (req == CONNECT && resp == 2xx) */
+			if (resp->status - 100U < 100U || resp->status == 204
```
keshonok: I'd say it's unclear now what's the reason for these checks. I would add an explicit reference to 7230 section 3.3.3 in the comment.

tempesta_fw/http_parser.c
```diff
@@ -1219,16 +1214,26 @@ __parse_connection(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
-	 * According to RFC 7230 section 3.3.2:
-	 *
+	 * A server MUST NOT send a Content-Length header field in any response
+	 * with a status code of 1xx (Informational) or 204 (No Content).  A
+	 * server MUST NOT send a Content-Length header field in any 2xx
+	 * (Successful) response to a CONNECT request
+	 */
+	if (TFW_CONN_TYPE(msg->conn) & Conn_Srv) {
+		TfwHttpResp *resp = (TfwHttpResp *)msg;
+
```
keshonok: Usually it's best to keep the code dense in cases like this, and don't add an extra empty line. It's just three lines of code. The first statement in the block is treated as an assignment in this case. Anyway, it's a matter of choice, preference, taste, etc.

tempesta_fw/http_parser.c
```diff
@@ -4180,6 +4185,13 @@ tfw_http_parse_resp(void *resp_data, unsigned char *data, size_t len)
 			/* Status code is fully parsed, move forward. */
 			resp->status = parser->_acc;
 			parser->_acc = 0;
+			/* Some responses are not allowed to have a body. */
+			/* TODO: Add (req == CONNECT && resp == 2xx) */
```
keshonok: This comment is actually part of the processing according to 7230 section 3.3.2. Right now we don't have a link to the paired request at this stage. But I think it doesn't have to be that way. We should be able to 'pop' the paired request as soon as we start processing the response.

Now we only have a link from `req` to `resp` (`req->resp`), as that is dictated by the logic of the current code. When developing the code I actually has a common `pair` member of `TfwHttpMsg` which was pointing to the paired request from the response, and to the paired response from the request. There was no need for that at the time, so a pointer in one direction only had been left. Perhaps this idea can be re-introduced to cover cases like this in the comment, where there's the need to check both the paired request's request type, and the response's status code.

ikoveshnikov: Yes, sounds reasonable. But currently there's still no need in paired request since CONNECT method is not supported yet. I will leave that for future updates.

Issue #717, opened 2017-04-01
Frang: limit HTTP error codes (password crackers blocking)


krizhanovsky:
We have several requests for protection against passwords crackers. The problem can be generalized to block a client sending requests which leads to N specific HTTP error codes in a specified time frame T (in seconds). The syntax for the new rule should be as:

        http_resp_code_block <list of error codes> <limit> <time>

e.g.

        http_resp_code_block 401 403 3 10

craig:
Or maybe causing <limit> error codes in a timeframe? Otherwise it might be easy to circumvent by doing a 200 request every <limit -1> requests.

krizhanovsky:
Yeah, good point! In fact we do so for other Frang limits, so we should do the same in this limit.

PR #837, opened 2017-09-05
Frang: limit HTTP error codes (password crackers blocking) #717

bartolootrit:


tempesta_fw/http.c
```diff
@@ -2420,6 +2423,20 @@ tfw_http_resp_cache(TfwHttpMsg *hmresp)
...
+	 * needed now. Save previous (NULL) value for the sake of safety
+	*/
+	state = &hmresp->conn->state;
+	prev_state_obj = state->obj;
+	state->obj = req;
+	prev_resp = req->resp;
+	req->resp = hmresp;
+	tfw_gfsm_move(state, TFW_HTTP_FSM_RESP_MSG_FWD, NULL, 0);
+	state->obj = prev_state_obj;
+	req->resp = prev_resp;
```
krizhanovsky: Yeah, bad stuff about existing GFSM code. I created #838 issue to extend GFSM in proper way.

tempesta_fw/classifier/frang.c
```diff
@@ -502,14 +546,70 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
 }
 
+static unsigned short
+frang_resp_next_idx(unsigned short ab, const unsigned short len)
+{
+	return ++ab >= len ? 0 : ab;
+}
+
+static void
+frang_resp_keep_frame(FrangRespCodeStat *stat, const unsigned long tc)
```
ikoveshnikov: As i see, the `frang_req_limit()` has pretty the same meaning and mostly the same should be done in your patch. Look, in the patch you may recount current rate of responses (with given status code) twice: first during response processing, then during processing the next request. I suggest solely increment number of responses for _current part_ of current time frame during response processing and evaluate limit overflow during request processing. 

With this approach you don't need scale factor and variable-sized `FrangRespCodeStat.buf`, and structure like `FrangAcc.FrangRates` will cover all your needs. The only difference is sampling rate: `FrangAcc.FrangRates` is sampled `FRANG_FREQ` times per second (since limits are per-second) while your data can be sampled `FRANG_FREQ` times per `FrangHttpRespCodeBlock.tf` (since limit is per user defined time frame).

tempesta_fw/classifier/frang.c
```diff
@@ -502,14 +546,70 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+{
+	const FrangHttpRespCodeBlock *conf = frang_cfg.http_resp_code_block;
+	const unsigned short ai = (tc % conf->tf) / conf->sf;
+
+	/*
+	 * FrangRespCodeStat is filled with zero in the beginning
+	 * because of TfwClient allocation with __GFP_ZERO flag
+	 */
+	if (tc - stat->tb < conf->tf)
```
ikoveshnikov: Don't see how the comment above is connected with the condition. I actually don't see any need in the comment.

tempesta_fw/classifier/frang.c
krizhanovsky: >  suggest solely increment number of responses for current part of current time frame during response processing and evaluate limit overflow during request processing.

We discussed this with @bartolootrit . The problem is that we limit number of bad responses in _time_ _frame_. Consider we had a limit for 10 responses per 10 seconds and a user generates 2 bad responses per second leading to blocking at 6th second. Next the user stops sending and we have to update the statistic on next request - we must allow 2 more bad responses on 7th second. So it seems we have to recalculate the statistic either on requests and responses.

tempesta_fw/classifier/frang.c
```diff
@@ -993,6 +1137,100 @@ frang_free_ct_vals(TfwCfgSpec *cs)
...
+
+	cb->sf = cb->tf / FRANG_RESP_CODE_TIME_FRAME
+		+ (cb->tf % FRANG_RESP_CODE_TIME_FRAME != 0);
+	/*
+	 * (cb->tf - 1) / cb->sf is a maximum value of ai in frang_resp_handler()
+	 */
+	cb->stat_buf_len = ((cb->tf - 1) / cb->sf) + 1;
+
+
+	return frang_register_fsm_resp();
```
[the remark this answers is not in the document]
krizhanovsky: The good point in the current code is that we don't register the GFSM hook if the feature wasn't configured and we don't need to use the hook. It's better for runtime. Meantime, `frang_register_fsm_resp()` prints its own error messages, so the user sees the reason for the failure, but, yes, only at configuration phase.

tempesta_fw/classifier/frang.c
```diff
@@ -868,6 +974,44 @@ frang_http_req_handler(void *obj, struct sk_buff *skb, unsigned int off)
...
+{
+	TfwHttpReq *req = (TfwHttpReq *)obj;
+	TfwHttpResp *resp = (TfwHttpResp *)req->resp;
+	FrangAcc *ra = (FrangAcc *)req->conn->sk->sk_security;
+	FrangRespCodeStat *stat = &ra->resp_code_stat;
+	const FrangHttpRespCodeBlock *conf = frang_cfg.http_resp_code_block;
+	unsigned long tc;
+
+	if (!frang_resp_code_range(resp->status)
+	    || !test_bit(FRANG_HTTP_CODE_BIT_NUM(resp->status), conf->codes))
```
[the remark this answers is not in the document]
bartolootrit: frang_resp_handler() will be called if frang_register_fsm_resp() is called in frang_set_rsp_code_block() after successful allocation of FrangHttpRespCodeBlock

tempesta_fw/classifier/frang.c
```diff
@@ -993,6 +1137,100 @@ frang_free_ct_vals(TfwCfgSpec *cs)
...
+
+	return 0;
+}
+
+static int
+frang_parse_ushort(const char *s, unsigned short *out)
+{
+	int n;
+	if(tfw_cfg_parse_int(s, &n) || n < 1 || n > USHRT_MAX) {
+		TFW_ERR_NL("frang: http_resp_code_block: \"%s\" isn't a valid value\n", s);
```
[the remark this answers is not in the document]
bartolootrit: The coding style says: "However, never break user-visible strings such as printk messages, because that breaks the ability to grep for them". 
It will be fixed in the next commit though.

tempesta_fw/classifier/frang.c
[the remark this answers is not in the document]
ikoveshnikov: Yes, sorry for the comment. Marked the case here to check once again before submitting review, but forgot to remove the comment at the end.

tempesta_fw/classifier/frang.c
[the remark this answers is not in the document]
ikoveshnikov: I meant call `frang_register_fsm_resp()` in `frang_init()` only if `frang_cfg.http_resp_code_block` is non-null. For me it looks that `frang_init()` is the best place to register GFSM, but i don't insist.

tempesta_fw/classifier/frang.c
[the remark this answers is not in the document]
krizhanovsky: That's true, so we try to keep error messages as small as possible and/or break lines on ungrepable points, `"%s"` in this case.

tempesta_fw/classifier/frang.c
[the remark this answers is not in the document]
krizhanovsky: > I meant call frang_register_fsm_resp() in frang_init() only if frang_cfg.http_resp_code_block is non-null. 

Can we check the configuration option already in `frang_init()`? If so, then it will be the best solution. I supposed that `frang_init()` is called on the module loading, before reading the configuration file, so we don't know whether the option is switched on or off.

tempesta_fw/http.c
```diff
@@ -2420,6 +2423,20 @@ tfw_http_resp_cache(TfwHttpMsg *hmresp)
 	}
 
 	/*
+	 * FSM needs TfwHttpReq to record data. It needs TfwHttpResp as well.
+	 * It will be added to the request later in tfw_http_resp_fwd(), but it's
+	 * needed now. Save previous (NULL) value for the sake of safety
```
krizhanovsky: We just had a conversation about the place where Frang is called for responses. Actually, your initial attempt to hook `tfw_http_resp_fwd` is the most obvious. So please add a note to the comment why do we decided to hook exactly this function.

tempesta_fw/classifier/frang.c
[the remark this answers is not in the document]
ikoveshnikov: Sorry, both you guys, for the some reason i've messed up everything altogether: both init and start of Tempesta's module. I was wrong in this case, nothing is wrong here.

tempesta_fw/classifier/frang.c
```diff
@@ -502,14 +537,38 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+	return tb - ta + (ta > tb ? FRANG_RESP_TIME_PERIOD : 0)
+		< frang_cfg.http_resp_code_block->tf;
+}
+
+static int
+frang_bad_resp_limit(FrangAcc *ra)
+{
+	FrangRespCodeStat *stat = ra->resp_code_stat;
+	long cnt = 0;
+	const int ts = (jiffies / HZ) % FRANG_RESP_TIME_PERIOD;
```
ikoveshnikov: You have added a few restrictions which only complicates code. Try to keep them as simpler as possible. Having `ts` of the same type as `jiffiies` wouldn't require `FRANG_RESP_TIME_PERIOD` macro. You don't need to check if the requests are made in the same year,  keep track of integer overflow and all that stuff.

I've looked though the latest changes (FrangRates history approach) and i find usage of `FrangRespCodeStat.ts` (and `i` from line below) to be wrong. Let's look to the idea behind the `FrangRespCodeStat.ts` member and the `FrangAcc.FrangRespCodeStat[FRANG_FREQ])` circular buffer. Instead of working with continuous time we measure time with small discrete quantum. Size of each quantum is `(timeframe in seconds) /   FRANG_FREQ`. The `FrangRespCodeStat.ts` is time measured in such quantum.

On processing response we get current system time in jiffies and translate to that quantum. Then we get the number of the quantum in current time frame. This tells which exactly array member of  `FrangAcc.FrangRespCodeStat[FRANG_FREQ])` must be used. If timestamp of the current `FrangRespCodeStat` doesn't equals to current timestamp, we can simply overwrite it since it out of current time frame. Finally we increment number of responses for the current `FrangRespCodeStat`.

On processing request we also get current system time in jiffies and translate to that quantum. That is needed to find beginning of the current time frame. Then we traverse over the `FrangAcc.FrangRespCodeStat[FRANG_FREQ])` circular buffer to get number of responses in the current time frame (time stamp is bigger or equal to the beginning of current time frame).

What is needed to be done:
- `FrangRespCodeStat.ts` must be type of `unsigned long` (same as `jiffies`)
- introduce function to convert jiffies to current quantum depending of time frame size. The conversion is used 3 times: `frang_bad_resp_limit()`, `frang_resp_handler()` and `frang_req_limit()` (the latest uses 1 second time frames). 
`ts = jiffies * FRANG_FREQ / (HZ * time_frame_size)`
- current member in circular buffer is directly connected to the time stamp: 
`i = ts % (FRANG_FREQ * time_frame_size)`
- introduce helper function (or macro) for collecting current rate (twice:  `frang_bad_resp_limit()`, and `frang_req_limit()`) to compare timestamp of buffer member to start of current timestamp:
`if (buff[i].ts + FRANG_FREQ * time_frame_size >= ts) ...`

tempesta_fw/classifier/frang.c
```diff
@@ -502,14 +537,38 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+	int i = 0;
+
+	for (; i < FRANG_FREQ; ++i) {
+		if (frang_resp_time_in_frame(stat[i].ts, ts))
+			cnt += stat[i].cnt;
+		if (cnt > frang_cfg.http_resp_code_block->limit)
```
ikoveshnikov: I see an optimisation effort here, but the condition is called for every `stat[X]`. Either move it to previous `if` statement body or after the cycle end. The array is small, the optimisation is not required here.

tempesta_fw/classifier/frang.c
```diff
@@ -69,20 +69,54 @@ typedef struct {
 	unsigned int	req;
 } FrangRates;
 
+/* The least power of two greater than a year in seconds */
+#define FRANG_RESP_TIME_PERIOD 33554432
```
ikoveshnikov: See review comment in `frang_bad_resp_limit()`.

tempesta_fw/classifier/frang.c
```diff
@@ -69,20 +69,54 @@ typedef struct {
...
+ * Response code record.
+ *
+ * @cnt	- Amount of responses in a time frame part;
+ * @ts	- Response time in seconds. Four bytes could be used to store enough
+ *   number of seconds in the assumption that there won't be delays between user
+ *   responses longer than a year;
+ */
+typedef struct {
+	long	cnt;
+	int	ts;
```
ikoveshnikov: See review comment in `frang_bad_resp_limit()`.

tempesta_fw/classifier/frang.c
```diff
@@ -868,6 +933,48 @@ frang_http_req_handler(void *obj, struct sk_buff *skb, unsigned int off)
...
+	int ts;
+	int i;
+
+	if (!frang_resp_code_range(resp->status)
+	    || !test_bit(FRANG_HTTP_CODE_BIT_NUM(resp->status), conf->codes))
+		return TFW_PASS;
+
+	spin_lock(&ra->lock);
+
+	ts = (jiffies / HZ) % FRANG_RESP_TIME_PERIOD;
```
ikoveshnikov: See review comment in `frang_bad_resp_limit()`.

tempesta_fw/classifier/frang.c
```diff
@@ -69,20 +69,54 @@ typedef struct {
...
+ *
+ * @cnt	- Amount of responses in a time frame part;
+ * @ts	- Response time in seconds. Four bytes could be used to store enough
+ *   number of seconds in the assumption that there won't be delays between user
+ *   responses longer than a year;
+ */
+typedef struct {
+	long	cnt;
+	int	ts;
+} __attribute__((packed)) FrangRespCodeStat;
```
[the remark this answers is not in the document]
bartolootrit: Without packing:
sizeof(FrangRespCodeStat) == 16; sizeof(resp_code_stat) == 128; sizeof(FrangAcc) == 264

With packing:
sizeof(FrangRespCodeStat) == 12; sizeof(resp_code_stat) == 96; sizeof(FrangAcc) == 232

It's needed to store @ts as int because @cnt could be greater than 2G and FrangAcc should be <= TFW_CLASSIFIER_ACCSZ == 256 bytes.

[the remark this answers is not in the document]
ikoveshnikov: `int ts` cause hacks, better to increase `TFW_CLASSIFIER_ACCSZ`. `TfwClassifierPrvt` is allocated once per client it's not a problem.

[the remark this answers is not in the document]
krizhanovsky: I agree with @bartolootrit : there is no sense to loose the 4 bytes for each item. Frang is a DDoS mitigation module, so we have to spend system resources for malicious users. We can expect millions of users (keep in mind influence of IoT), so while RAM is chap nowadays we'll use CPU caches ineficiantly if we increase the data structure significantly. Moreover, we're going to introduce many other classification algorithms and will need to increase the accounting data structure anyway, so it's better to keep it as tiny as possible.

tempesta_fw/classifier/frang.c
```diff
@@ -262,6 +295,12 @@ frang_conn_close(struct sock *sk)
 }
 
 static int
+frang_time_in_frame(const unsigned long tcur, const unsigned long tprev)
+{
+	return tprev + FRANG_FREQ > tcur;
```
ikoveshnikov: The function works wrong in `frang_bad_resp_limit()` and right in `frang_req_limit()`.  Currently the function checks whether the difference between `tpev` and `tcur` is less than one second. The function must be extended with third argument: time frame in seconds, which must be used as multiplier  for `FRANG_FREQ`.

bartolootrit:
```
ts = frang_resp_quantum();
i = ts % FRANG_FREQ;
if (ts != stat[i].ts) {
    stat[i].ts = ts;
    stat[i].cnt = 0;
}
ts = frang_resp_quantum() == ((jiffies / HZ) % FRANG_RESP_TIME_PERIOD) * FRANG_FREQ / frang_cfg.http_resp_code_block->tf;
```
In other words:
```
ts = (seconds % period) / (time_frame / FRANG_FREQ) == int32_seconds / (time_frame / FRANG_FREQ)
```
Let tf = 80, FRANG_FREQ = 8, and (jiffies / HZ) % FRANG_RESP_TIME_PERIOD == 100
0th second: ts0 = floor(100 / (80/8)) = 10
10th second: ts10 = floor(110 / (80/8)) = 11
80th second: ts80 = floor(180 / (80/8)) = 18

ts80 - ts0 == 8 >= FRANG_FREQ
80th second is in the other time frame than 0th second.

ikoveshnikov: Yes, i see, it was mine mistake. Sorry for that. The patch set can be merged into master.

krizhanovsky:
@bartolootrit thank you very much for you valuable contribution!

PR #839, opened 2017-09-08
Fix #825: Close client connection if error response is failed.

### LINKED ISSUE #825: Handling errors in constructing messages to clients
We have various situations, when we build 500, 502, 200 and other responses to clients. However when building of the response is failed (mostly due to insufficient memory) we do nothing. This breaks response-request pairing for pipelined requests. We have no choice but close the client connection in this case.

aleksostapenko:


tempesta_fw/http.c
```diff
@@ -272,6 +278,12 @@ tfw_http_send_resp(TfwHttpReq *req, TfwStr *msg, const TfwStr *date)
 	tfw_http_resp_fwd(req, (TfwHttpResp *)hmresp);
 
 	return 0;
+mem_err:
+	TFW_DBG2("%s: Memory allocation error: conn=[%p]\n",
+		 __func__, req->conn);
+	ss_close_sync(req->conn->sk, true);
+	tfw_http_conn_msg_free((TfwHttpMsg *)req);
+	return -ENOMEM;
```
ikoveshnikov: Also, exactly the same situation is processed here and in zap_error(), it should be moved into separate function to avoid code duplication.

ikoveshnikov:
Seems that it's the best solution possible since the issue can happen on any processing stage.  I think that `tfw_http_send_*()` in the patch shouldn't return error code on errors. It should have the same behaviour as `tfw_http_resp_fwd()`: send response or process the error if needed.

I have also noticed, that generated 302 and 304 replies are still affected by the issue.

tempesta_fw/http.c
```diff
@@ -660,24 +673,29 @@ tfw_http_req_zap_error(struct list_head *equeue)
...
-			break;
-		case 504:
-			tfw_http_send_504(req, req->httperr.reason);
-			break;
-		default:
-			TFW_WARN("Unexpected response error code: [%d]\n",
-				 req->httperr.status);
-			tfw_http_send_500(req, req->httperr.reason);
-			break;
+		if (!ret) {
```
krizhanovsky: Why do we stop sending error responses after first failure? If there is not enough memory, then memory reclamation process is running and we can keep trying to do our job.

tempesta_fw/http.c
```diff
@@ -660,24 +673,29 @@ tfw_http_req_zap_error(struct list_head *equeue)
...
+				ret = tfw_http_send_404(req, req->httperr.reason);
+				break;
+			case 500:
+				ret = tfw_http_send_500(req, req->httperr.reason);
+				break;
+			case 502:
+				ret = tfw_http_send_502(req, req->httperr.reason);
+				break;
+			case 504:
+				ret = tfw_http_send_504(req, req->httperr.reason);
```
krizhanovsky: The lines are bit wider than 80 characters and there is no need to place `switch` inside the `if` statement: just more statistic accounting before sending the responses and use `continue` statement in if.

tempesta_fw/http.c
```diff
@@ -260,10 +266,10 @@ tfw_http_send_resp(TfwHttpReq *req, TfwStr *msg, const TfwStr *date)
...
 	if (!(hmresp = tfw_http_msg_alloc_err_resp()))
-		return -ENOMEM;
+		goto mem_err;
 	if (tfw_http_msg_setup(hmresp, &it, msg->len)) {
 		tfw_http_msg_free(hmresp);
-		return -ENOMEM;
+		goto mem_err;
 	}
 
 	tfw_http_prep_date(date->ptr);
```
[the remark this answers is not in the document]
keshonok: The error returned from `tfw_http_msg_write()` is of different nature. It tells that there's an _internal_ error as msg (or skb) fragments are pre-allocated to fit a complete prepared message. Perhaps, a kernel message is justified in this case along with returning an error. Other that that, I agree that it's still an error while sending a response, and it should be treated equally with other errors in `tfw_http_send_resp()`.

tempesta_fw/http.c
keshonok: > Why do we stop sending error responses after first failure? 

As I tried to explain in the comment above, once a response cannot be sent, it's irrelevant whether subsequent responses can be sent or not. The pairing is broken, and the connection must be closed for that reason. And by design any subsequent responses will not be sent to a client anyway.

Or, perhaps I didn't understand the question?

tempesta_fw/http.c
keshonok: I believe a short comment would help to understand the consequences of closing a client connection here.

As soon there's an error sending a response, the pairing of requests and responses gets broken, and the connection must be closed to avoid wrong pairings with subsequent responses.

At the same time, as soon as a prepared response is not linked with a paired request, sending to that client stops starting with that request, because that creates a "hole" in the chain of requests -- a request without a response. Subsequent responses cannot be sent to the client until that hole is closed, which at this point will never happen.

tempesta_fw/http.c
```diff
@@ -660,24 +673,29 @@ tfw_http_req_zap_error(struct list_head *equeue)
...
+				ret = tfw_http_send_504(req, req->httperr.reason);
+				break;
+			default:
+				TFW_WARN("Unexpected response error code: [%d]\n",
+					 req->httperr.status);
+				ret = tfw_http_send_500(req, req->httperr.reason);
+				break;
+			}
+		} else {
+			ss_close_sync(req->conn->sk, true);
```
keshonok: I am not sure I undestand. `tfw_http_send_XXX()` functions ultimately call `tfw_http_send_resp()` which handles errors. It appears that with this code you're doing the same job twice.

And yes, like noted by @ikoveshnikov, you actually close a wrong client connection - from the next request, if there's any. And that request may be from a different connection.

So, which place to handle errors is ultimately better? If it's inside `tfw_http_send_resp()`, then it looks like `tfw_http_send_XXX()` functions don't need to return an error code at all. Or, perhaps, cases like with HTTP response codes 302 and 304 may prompt handling these errors outside of `tfw_http_send_resp()`.

It appears that more analysis is needed for a solution that handles all cases in an elegant way.

tempesta_fw/http.c
krizhanovsky: @keshonok thank you for the review.

As I understand for example `tfw_http_conn_repair()`, `equeue` can contain requests from different client connections, so in the code we face problems on sending error messages to different client connections and the error handling for the connections must not depend on each other.

keshonok: > can contain requests from different client connections

Ah, yes, of course. Didn't understand it at first. You are correct, each erroneous request must be processed independently, and processing in this function must not stop after first error.

tempesta_fw/http.c
```diff
@@ -236,9 +236,15 @@ tfw_http_prep_304(TfwHttpMsg *resp, TfwHttpReq *req, void *msg_it,
...
  * Set current date in the header of an HTTP error response, and set
  * the "Connection:" header field if it was present in the request.
+ * If memory allocation error occurred, then client connection
+ * should be closed, because response-request pairing for
+ * pipelined requests is violated.
  *
  * NOTE: This function expects that the last chunk of @msg is CRLF.
  */
+
+static void tfw_http_conn_msg_free(TfwHttpMsg *hm);
```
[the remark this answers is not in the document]
aleksostapenko: Corrected

tempesta_fw/http.c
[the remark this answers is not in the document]

tempesta_fw/http.c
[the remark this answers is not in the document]
aleksostapenko: Since this is another error, related to the inability to send a response to the client with information about the initial error - then yes, we should. Corrected.

tempesta_fw/http.c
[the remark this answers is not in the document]

tempesta_fw/http.c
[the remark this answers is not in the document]
aleksostapenko: Corrected: error handling logic has been removed from the `tfw_http_req_zap_error` function.

tempesta_fw/http.c
aleksostapenko: Comment added.

tempesta_fw/http.c
aleksostapenko: Corrected: error handling logic has been removed from the tfw_http_req_zap_error function.

tempesta_fw/http.c
aleksostapenko: I inserted error handling logic into `tfw_http_req_zap_error` - from the considerations that if we faced with memory allocation error, it is better not to try sending remain error responses and just to close all those connections - to speed up memory reclamation. But now I see that this is wrong way: closing the connection immediately without trying to send the client an error message and possibly to leave connection alive - is not justified.
So, I removed error handling logic from `tfw_http_req_zap_error`.
I left the error code returning from the functions `tfw_http_send_*` since they are used in unit tests:
```
        EXPECT_EQ(tfw_http_send_502(mock.req, "sticky calculation"), 0);
```
But maybe in case of unit tests - it is worth to replace functions `tfw_http_send_*` with helper functions which return values.

aleksostapenko: @keshonok thank you for review.

tempesta_fw/http.c
```diff
@@ -259,27 +311,46 @@ tfw_http_send_resp(TfwHttpReq *req, TfwStr *msg, const TfwStr *date)
...
+	/*
+	 * Pairing of requests and responses gets broken here. As soon
+	 * as request is not linked with any response, sending to that
+	 * client stops starting with that request, because that creates
+	 * a "hole" in the chain of requests -- a request without a response.
+	 * Subsequent responses cannot be sent to the client until that
+	 * hole is closed, which at this point will never happen. To solve
+	 * this situation there is no choice but to close client connection.
+	 */
+	tfw_http_req_conn_close(req);
```
ikoveshnikov: The function name is too generic, better to rename to `tfw_http_resp_build_error()` or smth like that and move the comment to the function description.

tempesta_fw/http.c
```diff
@@ -223,23 +223,75 @@ tfw_http_prep_304(TfwHttpMsg *resp, TfwHttpReq *req, void *msg_it,
...
+/*
+ * Close the client connection and free unpaired request. This function
+ * is needed for cases when we cannot prepare response for this request.
+ */
+void
+tfw_http_req_conn_close(TfwHttpReq *req)
+{
+	ss_close_sync(req->conn->sk, true);
+	tfw_http_conn_msg_free((TfwHttpMsg *)req);
+	TFW_INC_STAT_BH(clnt.msgs_otherr);
```
ikoveshnikov: Trailing white space. I noted the issue third time in your patches. Hard to find while reviewing on github.

tempesta_fw/http.c
```diff
@@ -259,27 +311,46 @@ tfw_http_send_resp(TfwHttpReq *req, TfwStr *msg, const TfwStr *date)
...
+		TFW_DBG2("%s: Response skb allocation error: conn=[%p]\n",
+			 __func__, req->conn);
+		goto err_setup;
 	}
-
 	tfw_http_prep_date(date->ptr);
-	tfw_http_msg_write(&it, hmresp, msg);
-
+	if (tfw_http_msg_write(&it, hmresp, msg)) {
+		TFW_DBG2("%s: Respnse allocation error: conn=[%p]\n",
```
ikoveshnikov: Misprint in resp**o**nse

tempesta_fw/http.c
```diff
@@ -318,16 +389,16 @@ tfw_http_send_403(TfwHttpReq *req, const char *reason)
 	};
 
 	TFW_DBG("Send HTTP 403 response: %s\n", reason);
-
-	return tfw_http_send_resp(req, &rh, __TFW_STR_CH(&rh, 1));
+
```
ikoveshnikov: Trailing white space.

tempesta_fw/http.c
```diff
@@ -259,27 +311,46 @@ tfw_http_send_resp(TfwHttpReq *req, TfwStr *msg, const TfwStr *date)
 		msg->len += crlf->len - crlf_len;
 	}
 
-	if (!(hmresp = tfw_http_msg_alloc_err_resp()))
-		return -ENOMEM;
+	if (!(hmresp = tfw_http_msg_alloc_err_resp())) {
+		TFW_DBG2("%s: Response message allocation error: conn=[%p]\n",
```
ikoveshnikov: I don't think, that we need so precise debugging for memory errors. It seems reasonable to put just single `TFW_DBG2` after `err_setup` label.

tempesta_fw/cache.c
```diff
@@ -1388,34 +1388,39 @@ tfw_cache_purge_invalidate(TfwHttpReq *req)
...
 	default:
-		return tfw_http_send_403(req, "purge: invalid option");
+		tfw_http_send_403(req, "purge: invalid option");
+		return;
 	}
 
-	return ret
-		? tfw_http_send_404(req, "purge: processing error")
-		: tfw_http_send_200(req);
+	tfw_http_send_200(req);
```
ikoveshnikov: I personally like the old approach in this switch-case block, when the error processing was divided from handling different purge methods.

tempesta_fw/http.c

tempesta_fw/http.c

tempesta_fw/http.c

tempesta_fw/http.c

tempesta_fw/http.c

tempesta_fw/cache.c
aleksostapenko: Corrected.

tempesta_fw/http.c
```diff
@@ -223,23 +223,85 @@ tfw_http_prep_304(TfwHttpMsg *resp, TfwHttpReq *req, void *msg_it,
...
+void
+tfw_http_resp_build_error(TfwHttpReq *req)
+{
+	ss_close_sync(req->conn->sk, true);
+	spin_lock(&((TfwCliConn *)req->conn)->seq_qlock);
+	if (likely(!list_empty(&req->msg.seq_list)))
+		list_del_init(&req->msg.seq_list);
+	spin_unlock(&((TfwCliConn *)req->conn)->seq_qlock);
+	tfw_http_conn_msg_free((TfwHttpMsg *)req);
+	TFW_INC_STAT_BH(clnt.msgs_otherr);
```
keshonok: I am not convinced that removal of a request from `seq_list`, and then freeing it, is necessary here. This will work just fine, yes, but it looks like unnecessary complication (and extra unnecessary code).

My understanding is this. When a error response cannot be sent out due to an error while preparing it, a "hole" in the `seq_list`  will cause a stop in sending responses to that client. No subsequent responses will be sent to it, starting with the response to this request `req`. Then the erroneous client connection is closed. That causes the connection drop and the invocation of `tfw_http_conn_cli_drop()`, which takes care of all requests that are still in the queue, including this one.

In other words, I believe, that one the first statement in this function is required to handle the situation properly. I am not sure of the last statement, as from that point the client connection should be considered closed.

aleksostapenko: My suspicions about request leak in this case were caused by comments and code from  `tfw_http_conn_cli_drop()`:
```
/*                                                                                                                                                                 
 * Drop client connection's resources.                                                                                                                             
 *                                                                                                                                                                 
 * Desintegrate the client connection's @seq_list. Requests that have                                                                                              
 * a paired response can be freed. Move those to @zap_queue for doing                                                                                              
 * that without the lock. Requests without a paired response have not                                                                                              
 * been answered yet. They are held in the lists of server connections                                                                                             
 * until responses come. Don't free those requests.
```
. . .
. . .

```
       spin_lock(&cli_conn->seq_qlock);
        list_for_each_entry_safe(req, tmp, seq_queue, msg.seq_list) {
                if (req->resp)
                        list_move_tail(&req->msg.seq_list, &zap_queue);
                else
        }
```
As it seems - unpaired request will be removed from `seq_queue` but will not be freed - to wait the paired response. But in our case response has already been arrived - so the request will be lost.
And as we have to explicitly free request - we also have to remove it from `seq_queue` before that, otherwise - we will catch `BUG_ON` in
```
tfw_http_req_destruct(void *msg)
{
        TfwHttpReq *req = msg;

        BUG_ON(!list_empty(&req->msg.seq_list));
. . .
```

keshonok: Yes, you're absolutely correct, I missed that. Never underestimate the value of a good comment!

I have another question though. Did you consider just passing a NULL `resp` to `tfw_http_resp_fwd()`, and then handling this special case by closing the client connection? In my view, that may provide a cleaner path for handling this rare error. It might be easier to follow and understand, because it follows the same flow of execution as in the normal case. Locks are natural there (already exist), and all there's to do there is to handle a NULL `resp` in an appropriate place.

It appears that the whole needed flow is already there. For instance, you could do this:
```C
BUG_ON(list_empty(&req->msg.seq_list));
if (!resp) {
        ss_close_sync(cli_conn->sk, true);
        tfw_http_conn_msg_free((TfwHttpMsg *)req);
        TFW_INC_STAT_BH(serv.msgs_otherr);
        return;
}
req->resp = (TfwHttpMsg *)resp;
```

Note that the case where `seq_list` is empty is handled before this piece of code in `tfw_http_resp_fwd()`, and it will correctly handle a NULL `resp`.

There's one caveat here though. The same caveat exists with your original code that handles this error outside of `tfw_http_resp_fwd()`.

Let's say there are 10 CPUs (from 1 to 10), and there are 10 requests in `seq_list`.  Let's say this kind of error occurs on CPU №10. Now, let's say the order in which other 9 `seq_list` entries are filled with a `resp` is: 9, 8, 7, 6, 5, 4, 3, 2, 1. Until the first entry is filled with a `resp`, all of these responses are not sent out to the client.

If CPU 10 comes last after all of these, then all 9 responses are sent out to the client. If CPU 10 comes first, then none of the 9 already received and prepared responses are sent out. You get the picture.

So the question is, can we handle this special case in `tfw_http_resp_fwd()` so that we surely send all preceding (ready) responses to the client before we close the connection? Perhaps, with a special flag? I believe that for proper efficiency, this is a valid case to think about...

ikoveshnikov: As for me, passing `NULL` response to `tfw_http_resp_fwd()` as a part of error processing doesn't seem to be a clear and nice. But it's a matter of taste mostly.

Currently we have traditional approach with separate normal and error processing:
```C
/* Some actions to prepare response (resp) */
if (error_on_prepare_resp)
    goto error;

tfw_http_resp_fwd(req, resp);
return;

error:
kfree(resp);
tfw_http_resp_build_error(req);
```
But if `tfw_http_resp_build_error()` is moved to `tfw_http_resp_fwd()`, error processing will be implemented in other way:
```C
/* Some actions to prepare response (resp) */
if (error_on_prepare_resp) {
    kfree(resp);
    resp = NULL;
}

tfw_http_resp_fwd(req, resp);
return;
```
In the second case error processing is implicitly made in `tfw_http_resp_fwd(req, resp)` call and that makes the call to be mandatory. That's why i don't really like the approach. But as i said, it's matter of taste.

Concerning  second problem you have mentioned (drops of received but not forwarded responses). It's a hard problem actually. There is only one reason why Tempesta fails to prepare response on it's own: lack of memory. If single memory error happen, it's very likely that more are coming. It seems reasonable to drop one clients to save memory for others. Tempesta's behaviour is not tested thoroughly on random memory allocation errors which can potentially cause very different errors.

keshonok: I tend to agree with @ikoveshnikov that handling errors at the place they arise is more traditional, explicit, and better understood.

Per request from @ikoveshnikov, here's a sample comment that describes the case where errors are handled _outside_ of `tfw_http_resp_fwd()`. A comment like this would help to describe the intricacies of the case and help to remember the choice that has been made.

*_As a consequence of closing a client connection on error of preparing a response, it's possible that some already prepared responses will not be sent to the client. That depends on the order in which CPUs close the connection and call tfw_http_resp_fwd(). This is the intended behaviour. The goal is to free some memory at the cost of dropping a few clients, so that Tempesta can continue working._*

krizhanovsky: @keshonok, this is an interesting case, very strange however. Basically, we can simplify it to a race between preparing response to 10th request and 1st one - all other responses are just blocked by unfinished 1st request. The problem happens only on absence of memory, i.e. response for the 1st request must be received and processed just before the problem happens and next the CPU must stuck a little bit giving 1th CPU to face memory starvation. The case should be rare and insignificant in case of race between only 2 messages. The case with all 10 messages is even less probable. Next, we typically need some memory for network transmission (keep in mind that skb clones etc), so its likely that all the 9 messages still won't be delivered. Last, all the messages are synchronized via socket work queue to send data over the socket and close it, so if we're lucky we can deliver 9 responses before closing the socket due to an error.

Memory starvation is very serious thing introducing OOM killing of various processes and failures on most of allocation calls. It's better to choose some unlucky connections/clients and completely free the resources to get enough memory instead of trying to do anything under heavy memory pressure, which leads to long scannings for free memory pages etc.

aleksostapenko: Added specified comment for 'tfw_http_resp_build_error' function.

PR #845, opened 2017-09-22
Add new stress tests.

aleksostapenko:
1. Add pipelined requests during stress testing.
2. Add fault injection via SystemTap.

### DIFF
--- a/tempesta_fw/t/functional/helpers/control.py
+++ b/tempesta_fw/t/functional/helpers/control.py
@@ -114,17 +114,13 @@ class Wrk(Client):
         self.script = ''
 
     def set_script(self, script):
-        self.script = script
+        self.script = ''.join([script, '.lua'])
 
     def append_script_option(self):
         if not self.script:
             return
-        path = ''.join([os.path.dirname(os.path.realpath(__file__)),
-                        '/../wrk/', self.script, '.lua'])
-        script_path = os.path.abspath(path)
-        assert os.path.isfile(script_path), \
-               'No script found: %s !' % script_path
-        self.options.append('-s %s' % script_path)
+
+        self.options.append('-s %s' % get_script_path(self.script, 'wrk'))
 
     def form_command(self):
         self.options.append('-d %d' % self.duration)
@@ -177,6 +173,14 @@ class Ab(Client):
 #-------------------------------------------------------------------------------
 # Client helpers
 #-------------------------------------------------------------------------------
+
+def get_script_path(script, dir_name):
+    path = ''.join([os.path.dirname(os.path.realpath(__file__)),
+                    '/../', dir_name, '/', script])
+    script_path = os.path.abspath(path)
+    assert os.path.isfile(script_path), \
+        'No script found: %s !' % script_path
+    return script_path
 
 def client_run_blocking(client):
     """ User must call client.prepare() before calling the function and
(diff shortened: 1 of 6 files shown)


tempesta_fw/t/functional/systemtap/resp_alloc_err.stp
```diff
@@ -0,0 +1,6 @@
+probe module("tempesta_fw").function("tfw_http_msg_alloc_err_resp").return
+{
+	if ($return != 0) {
+		$return = 0
+	}
```
krizhanovsky: Make sense. There is nothing special about preventing memory leakage: `tfw_pool_destroy()` is quite straightforward and short function and you can just reimplement it in the hook, basically it just frees the allocated pages.

aleksostapenko: Yes, I've missed this approach. Corrected.

intelfx:
@krizhanovsky We discussed this briefly on Slack — with the conclusion being that @aleksostapenko does the _stress_ test (which does not concern itself with correctness) and I do the _functional_ test once I'm done with CI.

tempesta_fw/t/functional/regression/test_stress_pipeline.py
```diff
@@ -0,0 +1,45 @@
...
+import sys
+import unittest
+from helpers import control, tempesta
+from testers import stress
+
+__author__ = 'Tempesta Technologies, Inc.'
+__copyright__ = 'Copyright (C) 2017 Tempesta Technologies, Inc.'
+__license__ = 'GPL2'
+
+class Pipeline(stress.StressTest):
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -178,6 +174,14 @@ def parse_out(self, stdout, stderr):
 # Client helpers
 #-------------------------------------------------------------------------------
 
+def get_script_path(script, dir_name):
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -123,7 +123,7 @@ def append_script_option(self):
                         '/../wrk/', self.script, '.lua'])
         script_path = os.path.abspath(path)
         assert os.path.isfile(script_path), \
-               'No script found: %s !' % script_path
+            'No script found: %s !' % script_path
```
intelfx: This is an unrelated change, please revert this hunk.

tempesta_fw/t/functional/regression/test_stress_pipeline.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def run_test(self, ka_reqs):
+        for s in self.servers:
+            s.config.set_ka(ka_reqs)
+        self.generic_test_routine('cache 0;\n')
+
+    def test_unlimited_ka(self):
+        self.run_test(sys.maxsize)
+
+    @unittest.expectedFailure
```
intelfx: Please remove this decorator: the test actually passes.

tempesta_fw/t/functional/helpers/control.py
intelfx: This is an unrelated change. Please revert this hunk.

tempesta_fw/t/functional/regression/test_stress_pipeline.py
```diff
@@ -0,0 +1,52 @@
...
+    def test_low_ka(self):
+        """Low keep_alive value, make the server to close after the limit
+        is exhausted connection; thus Tempesta must generate 502 response.
+        """
```
intelfx: The commentary is misleading. This test spawns more than one server, so failovering will complete successfully without causing Tempesta to respond with a 502.

tempesta_fw/t/functional/regression/test_stress_pipeline.py
```diff
@@ -0,0 +1,52 @@
...
+        self.run_test(50)
+
+class PipelineFaultInjection(Pipeline):
```
intelfx: This test does nothing as implemented now. If you have time, please override `create_servers()` to create just a single nginx instance with a small `keepalive_requests` value (such that failovering will happen and will fail) and override the assert_* methods as needed, inverting those error checks that are expected to fail (or, better yet, add a parameter of `expect_errors=False` or something like that to the assert_* methods). Finally, on top of that create another test with fault injection. Later we will modify the Lua script to actually validate the responses.

If you inherit your failure-mode test from `Pipeline`, there will be no point in keeping the `test_unlimited_ka()` method in these tests, so delete it with `delattr()` in global scope.

tempesta_fw/t/functional/regression/test_stress_pipeline.py
aleksostapenko: Corrected.

tempesta_fw/t/functional/regression/test_stress_pipeline.py
aleksostapenko: Corrected

tempesta_fw/t/functional/regression/test_stress_pipeline.py
aleksostapenko: Corrected.

tempesta_fw/t/functional/helpers/control.py
aleksostapenko: Corrected.

PR #855, opened 2017-11-03
Remove TFW_STOP/SS_STOP flags (#658).

aleksostapenko:


tempesta_fw/http.c
```diff
@@ -2303,8 +2302,13 @@ tfw_http_req_process(TfwConn *conn, struct sk_buff *skb, unsigned int off)
...
 		 * So, save the needed request flag for later use as it
 		 * may not be accessible later through @req->flags.
+		 * If the connection must be closed, it also should be marked
+		 * with @Conn_Stop flag - to left it alive for sending responses
+		 * and, at the same time, to stop passing data for processing
+		 * from the lower layer.
 		 */
-		req_conn_close = req->flags & TFW_HTTP_CONN_CLOSE;
+		if((req_conn_close = req->flags & TFW_HTTP_CONN_CLOSE))
+			TFW_CONN_TYPE(req->conn) |= Conn_Stop;
```
krizhanovsky: It has sense to remove `req_conn_close` variable and just use `TFW_CONN_TYPE(req->conn) | Conn_Stop` check instead. That simplifies the code for one variable.

tempesta_fw/http.c
```diff
@@ -2369,7 +2373,7 @@ tfw_http_req_process(TfwConn *conn, struct sk_buff *skb, unsigned int off)
 		 * from this point on.
 		 */
 		if (req_conn_close)
-			return TFW_STOP;
+			return TFW_BLOCK;
```
krizhanovsky: I think it's better to replace this by

     if (req->flags & TFW_HTTP_CONN_CLOSE)
         break;

since `TFW_BLOCK ` means blocked (typically malicious) request and it's better to keep value from the previous call .

tempesta_fw/sock.c
```diff
@@ -746,21 +746,22 @@ ss_tcp_process_skb(struct sock *sk, struct sk_buff *skb, int *processed)
...
+			break; /* connection stopped */
+		}
 		if (r < 0) {
 			TFW_DBG("[%d]: Processing error: sk %p r %d\n",
 			        smp_processor_id(), sk, r);
 			goto out; /* connection dropped */
-		} else if (r == SS_STOP) {
-			TFW_DBG("[%d]: Stop processing: sk %p\n",
-			        smp_processor_id(), sk);
-			break;
```
krizhanovsky: I think that wasn't a good piece of code. The code takes place when we completely read a client request with `Connection: close`, but client sends us more data. We have to close the socket, i.e. send FIN and go through normal TCP close process with appropriate sequence numbers and ACKs. If we just `break` here, then we improperly update `tp->copied_seq`: we update it for the FIN, but not for other date after the request with `Connection: close`. Thus, connection closing can go with errors.

Please add a functional test which sends data like
```
GET / HTTP/1.1\r\n
Host: foo.com\r\n
\r\n
[many arbitrary data to fit couple skbs]
```

(One case with the tail data and one without the data). The result should be verified with tcpdump which must show that we have correct number of packets and there are proper FIN sequences. Ideally this should be more or less full connection closing test.

@intelfx and @ikoveshnikov please note the test scenarios: typically we have to develop many tests verifying correct operation of TCP.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -117,7 +117,7 @@ do_split_and_parse(unsigned char *str, int type)
 
 	r = split_and_parse_n(str, type, len, chunks);
 	if (++chunks > len)
-		return TFW_STOP;
+		return 1;
```
krizhanovsky: The value now looks strange. Please add a comment that we return any non `TFW_*` constant value just to stop copping a message to more pieces.

tempesta_fw/http.c
```diff
@@ -2075,7 +2074,7 @@ tfw_http_req_set_context(TfwHttpReq *req)
 static inline void
 tfw_http_req_mark_error(TfwHttpReq *req, int status, const char *msg)
 {
-	TFW_CONN_TYPE(req->conn) |= Conn_Suspected;
+	TFW_CONN_TYPE(req->conn) |= Conn_Stop;
 	req->flags |= TFW_HTTP_SUSPECTED;
```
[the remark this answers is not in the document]
aleksostapenko: It seems to me that the current name is correct, as the `TFW_HTTP_SUSPECTED` flag is only for request (not for connection) and is set in `tfw_http_req_mark_error()` function only (i.e. in `Suspect` case). For `Connection: close` case - the `TFW_HTTP_CONN_CLOSE` request flag is responsible. These flags separately cover each case, unlike the `Conn_Stop` connection flag - which covers both (`Suspect` and `Connection: close`) cases.

tempesta_fw/http.c
```diff
@@ -1801,9 +1802,7 @@ __tfw_http_resp_fwd(TfwCliConn *cli_conn, struct list_head *ret_queue)
 	list_for_each_entry_safe(req, tmp, ret_queue, msg.seq_list) {
 		BUG_ON(!req->resp);
 		tfw_http_resp_init_ss_flags((TfwHttpResp *)req->resp, req);
-		if (tfw_cli_conn_send(cli_conn, (TfwMsg *)req->resp) ||
-		    req->flags & TFW_HTTP_SUSPECTED)
```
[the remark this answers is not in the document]
aleksostapenko: This check had been moved into `tfw_http_resp_init_ss_flags()` function, where also exists a similar check for `TFW_HTTP_CONN_CLOSE` flag. As a result of these checks - the `SS_F_CONN_CLOSE` flag will be set in `TfwMsg->ss_flags` and connection will be closed in `ss_tx_action()` via `__sk_close_locked()`.

tempesta_fw/http.c
aleksostapenko: The request may be released in `tfw_http_req_cache_cb` (if it will be served from cache) - so, connection descriptor can be inaccessible through `req->conn` after `tfw_cache_process`.

tempesta_fw/http.c
```diff
@@ -536,7 +536,8 @@ tfw_http_req_init_ss_flags(TfwSrvConn *srv_conn, TfwHttpReq *req)
 static inline void
 tfw_http_resp_init_ss_flags(TfwHttpResp *resp, const TfwHttpReq *req)
 {
-	if (req->flags & TFW_HTTP_CONN_CLOSE)
+	if (req->flags & TFW_HTTP_CONN_CLOSE ||
+	    req->flags & TFW_HTTP_SUSPECTED)
```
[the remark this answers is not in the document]

tempesta_fw/t/unit/test_http_parser.c
aleksostapenko: Corrected.

tempesta_fw/http.c
aleksostapenko: Corrected partially (reason described in previous comment).

tempesta_fw/sock.c
keshonok: I am not sure I understand this. All skb that come here are counted in `tp->copied_seq`, so it's updated properly in all cases. That happens before they are passed to Tempesta for processing.

In this particular case the socket was not closed because we still had to send a response back the the client - when it came from a server. That's when a close would occur, according to the `Connection: close` header. That's was the whole purpose of `SS_STOP` compared to `SS_BLOCK`. `SS_BLOCK` would cause an immediate close, while `SS_STOP` would not.

All of that created a window for a misbehaving client that could continue sending more requests despite `Connection: close`. The window was between receiving that request and sending a response to it. In that case Tempesta would continue processing those newer requests. That was the deficiency of the old solution which I described in the comment to the issue.

Still, those new skb would be properly accounted for from TCP seq's point of view.

Closing would happen after

aleksostapenko: >All skb that come here are counted in `tp->copied_seq`
If we `break` from loop in case of `TFW_STOP` (or ` Conn_Stop` flag detected) - then skbs which remain in `skb_list` (i.e. "unrolled" skbs) - will not be counted in `tp->copied_seq`.

keshonok: Skb in `skb_list` were part of a GRO skb. And that GRO skb had been fully counted before it was unrolled. I don't think there's need to count the data twice.

krizhanovsky: The problem is with https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/sock.c#L718 : when we account or not the TCP sequence numbers, so we need not to exit (break) before accounting all akb pieces. `skb_unroll()` doesn't know about TCP anything.

keshonok: Duh! Yes, indeed. Thanks for pointing that out.

It appears though, that other error code paths are also affected. That includes returning on error in `ss_skb_unroll()`, and breaking out of the loop on `(r < 0)`. `tp->copied_seq` would not be updated properly in these cases, and a potential FIN would not be counted as well.

In these cases the connection is closed immediately, and perhaps an improper `tp->copied_seq` doesn't really matter. But perhaps not, and then this should better be fixed as well in this patch.

krizhanovsky: Yeah, good point: the other cases also treated in wrong way. But there is a difference: stopped connections still must be treated in graceful way (i.e. closed by FIN) while blocked connections should be reseted. We never cared about blocked connections, so this is why they are closed in current way. However, there is TODO comment https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/sock.c#L495 and we actually should reset the bad connections. Created issue https://github.com/tempesta-tech/tempesta/issues/861 . @aleksostapenko keep in mind the issue during implementation of #860.

tempesta_fw/sock.c
```diff
@@ -751,7 +755,7 @@ ss_tcp_process_skb(struct sock *sk, struct sk_buff *skb, int *processed)
 			continue;
 		}
 
-		r = SS_CALL(connection_recv, conn, skb, offset);
+		r = SS_CALL(connection_recv, conn, skb, off);
```
keshonok: Is there the need for an extra `off` variable, and extra assignments? Can we just do `offset = 0` after this line?

aleksostapenko: In case of `Conn_Stop` flag set (https://github.com/tempesta-tech/tempesta/blob/8a40fda1111ad7e4ab7c6a2740e482e48e2f8986/tempesta_fw/sock.c#L731) - we will go to the next iteration, `offset` will remain in incorrect state. On the other hand, we must check `Conn_Stop` flag for connection - before `SS_CALL()`.

keshonok: Yep, got it. Thanks!

keshonok: On the other hand, once we have that condition, doesn't it mean that we never get to proceed past it (to `SS_CALL`), so `offset` will not be used anyway, and its value won't matter? We will just loop until we account for and free all skb in `skb_list`, and never execute `SS_CALL()`.

keshonok: On the other hand, once we have that condition, wouldn't it mean that we never get to proceed past it (to `SS_CALL()`), so that `offset` would not be used anyway, and its value would be irrelevant? We would just loop until we account for and free all skb in `skb_list`, and we would never execute `SS_CALL()`. In the end, wouldn't that mean, that we could just zero `offset` after it's first use (where you do it already) without the need for an extra `off` variable?

keshonok: Sorry for (possible) latest comments, my browser got stuck in reload/repost mode. I have nothing to add to the discussion.

PR #859, opened 2017-11-10
Corrections for some functional tests.

aleksostapenko:


tempesta_fw/t/functional/testers/stress.py
```diff
@@ -116,13 +116,13 @@ def assert_tempesta(self):
                          msg=(msg % 'requests'))
         self.assertEqual(self.tempesta.stats.srv_msg_parsing_errors, 0,
                          msg=(msg % 'responses'))
-        if not self.cl_msg_errors:
+        if not self.tfw_msg_errors:
```
ikoveshnikov: I believe simple 
```
if self.tfw_msg_errors:
   return
```
is better here.

aleksostapenko: Corrected.

PR #864, opened 2017-11-22
Cleanup: Remove dead code: tfw_str_del_chunk()

keshonok:
The function is unused since the commit 08a10df55e37 when the original
__field_finish() was removed from http_parser.c, and the parser was
fixed to not create unnecessary empty chunks at the end of a current
chunk of processed data.

There was no need in this function since September 2015, so I believe
it's safe to remove it.

### DIFF
--- a/tempesta_fw/str.c
+++ b/tempesta_fw/str.c
@@ -97,29 +97,6 @@ tfw_ultoa(unsigned long ai, char *buf, unsigned int len)
 }
 EXPORT_SYMBOL(tfw_ultoa);
 
-void
-tfw_str_del_chunk(TfwStr *str, int id)
-{
-	unsigned int cn = TFW_STR_CHUNKN(str);
-
-	if (unlikely(TFW_STR_PLAIN(str)))
-		return;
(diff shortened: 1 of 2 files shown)


krizhanovsky:
`str.[ch]` is a string library and `tfw_str_del_chunk()` can be useful in further code. If we just remove the function, then it's likely that we'll just forget about it and reinvent it when we need it. Meantime the bad thing about the function is that it isn't covered by `t/unit/test_tfw_str.c` - it'd be good to write a unit test for the function.

keshonok:
Uh, that's fine. Just thought that, perhaps, it was worth removing since it wasn't used in two years. But I can understand the point that this should be considered a library function.

My understanding is that the two opposite functions here are `tfw_str_del_chunk()` and `tfw_str_add_compound()`. As far as I can see none of these are really **_covered_** by unit tests. The second one is just _used_ in `str` tests.

krizhanovsky:
Created issues https://github.com/tempesta-tech/tempesta/issues/865 for the tests

keshonok:
The function in question is a library function, hence it's should not be removed.

PR #867, opened 2017-11-27
Fix #860: Add functional tests for TCP connection closing.

### LINKED ISSUE #860: Functional test for connection closing
The test idea is described in comment https://github.com/tempesta-tech/tempesta/pull/855#discussion_r148932522 . The test also should be usable for #861 (TCP RST).

aleksostapenko:


tempesta_fw/t/functional/helpers/analyzer.py
```diff
@@ -0,0 +1,123 @@
...
+        self.packets = []
+        self.dump_file = '/tmp/tmp_packet_dump'
+        cmd = 'timeout %s tcpdump -i any %s-w - tcp port %s || true'
+        count_flag = ('-c %s ' % count) if count else ''
+        self.cmd = cmd % (timeout, count_flag, port)
+        self.err_msg = ' '.join(["Can't %s sniffer on", host])
+        self.node_side_close = node_close
+
+    def sniff(self):
+        stdout, stderr = self.node.run_cmd(self.cmd, timeout=None,
```
intelfx: Uh. I really dislike this tossing of data around: reading it completely in memory, then writing to a temporary file and finally reading it in memory again via scapy. How big is this dump expected to be on average, by the way?

It would be a lot nicer to feed the stream directly to scapy. https://github.com/secdev/scapy/issues/196 suggests you can pass an opened file object to `sniff(offline=...)` and it will parse the stream. Could you please try if scapy is happy with a file-like object returned by `paramiko.client.SSHClient.exec_command()`? 

This will probably require adding a new interface to `helpers.remote.Node` (it should relate to `helpers.remote.Node.exec_command()` just like `subprocess.Popen()` relates to `subprocess.run()`, i. e. run a command asynchronously and return its streams).

intelfx: Even if that fails somewhy, at least use StringIO to avoid having a temporary file.

tempesta_fw/t/functional/tcp_connection/test_connection_close.py
```diff
@@ -0,0 +1,78 @@
...
+
+__author__ = 'Tempesta Technologies, Inc.'
+__copyright__ = 'Copyright (C) 2017 Tempesta Technologies, Inc.'
+__license__ = 'GPL2'
+
+
+class CloseConnection(functional.FunctionalTest):
+    """Regular connection closing."""
+
+    def stop_and_close(self):
```
ikoveshnikov: It's just the same code as `functional.FunctionalTest.tearDown()` which is called after every test. Isn't it better to override `tearDown()` method and append `sniffer.stop()` there if it necessary?

aleksostapenko: We need to close the connection to check (via `tcpdump`) the correctness of its closing (and in case of failure - to throw an exception in the `test*` function of class); so before stopping `tcpdump` and analyzing its output, we need to close `Deproxy` (and its client and server connections also)  - in `test*` function.

tempesta_fw/t/functional/helpers/analyzer.py
```diff
@@ -0,0 +1,123 @@
...
+    def __init__(self, node, host, count=0,
+                 timeout=30, port=80,
+                 node_close=True):
+        self.node = node
+        self.port = port
+        self.thread = None
+        self.captured = 0
+        self.packets = []
+        self.dump_file = '/tmp/tmp_packet_dump'
+        cmd = 'timeout %s tcpdump -i any %s-w - tcp port %s || true'
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/t/functional/helpers/analyzer.py
aleksostapenko: Unfortunately, `scapy.sniff(offline=file_obj)` interface does not support `StringIO` objects or `paramiko` file objects. I mentioned this in comment to `sniff()` function.

tempesta_fw/t/functional/helpers/deproxy.py
```diff
@@ -413,6 +409,10 @@ def parse_firstline(self, stream):
         except:
             raise ParseError('Invalid Status code!')
 
+    def get_length(self):
+        firstline = ' '.join([self.version, self.status, self.reason])
+        return len('\r\n'.join([firstline, str(self)]))
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/t/functional/tcp_connection/test_connection_close.py
ikoveshnikov: Ok, I see the point. Then the function should be documented since the approach differs from other tests

tempesta_fw/t/functional/tcp_connection/test_connection_close.py
```diff
@@ -0,0 +1,78 @@
...
+                                         expected_response = response_403)
+        return [chain_200, chain_403]
+
+    def create_sniffer(self):
+        self.sniffer = analyzer.AnalyzerCloseRegular(self.tempesta.node,
+                                                     self.tempesta.host,
+                                                     timeout=10)
+
+
+
```
ikoveshnikov: Too lot of empty lines

tempesta_fw/t/functional/helpers/analyzer.py
```diff
@@ -0,0 +1,128 @@
...
+            f.write(stdout)
+
+    def start(self):
+        self.thread = Thread(target=self.sniff)
+        self.thread.start()
+
+    def stop(self):
+        if self.thread:
+            self.thread.join()
+            if os.path.exists(self.dump_file):
```
ikoveshnikov: Missing dump file is a framework error, isn't it? If so we can raise a error here to fail the test and report framework error. See `helpers/error.py`. The test will be marked as `Errored` instead of `Failed`.

tempesta_fw/t/functional/helpers/deproxy.py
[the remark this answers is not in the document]
aleksostapenko: Also changed the update logic for `self.msg` - due to out of sync problem between textual and logical message representations during receiving more than one message in a row.

tempesta_fw/t/functional/tcp_connection/test_connection_close.py
aleksostapenko: Corrected.

tempesta_fw/t/functional/tcp_connection/test_connection_close.py
aleksostapenko: Added comment into `stop_and_close(self)` function.

tempesta_fw/t/functional/helpers/analyzer.py
aleksostapenko: Corrected.

tempesta_fw/t/functional/helpers/deproxy.py
```diff
@@ -212,20 +212,26 @@ def __init__(self, message_text=None, body_parsing=True, body_void=False):
...
+        self.__set_str_msg()
 
     def __parse(self, stream):
         self.parse_firstline(stream)
         self.parse_headers(stream)
         self.body = ''
         self.parse_body(stream)
 
+    def __set_str_msg(self):
+        self.msg = '\r\n'.join([self.get_firstline(), str(self)])
```
ikoveshnikov: The patch set fixes quite a lot of dirty code in the test framework. I would like to ask for one more cleanup.
For both request and response you have to join firstline and the rest of the message in many places: in `__set_str_msg()`, in `[Request|Responce].update()`. But in the same time `__str__()` method ignores firstline when it shouldn't.

To make the code looks better we should join the firstline to the rest of the message in `__str__()` method. After that we won't need to pass the first line to `update()` method, so both Request and Response won't need to override the `update()` method. `__set_str_msg()`  will be changed to simple `self.msg = str(self)`.

aleksostapenko: Corrected.

Issue #758, opened 2017-07-02
Test for memory leakages


krizhanovsky:
Need to develop a long running test for catching memory leaks like https://github.com/tempesta-tech/tempesta/issues/691 . This should be part of long running tests, so depends on [CI (#484)](https://github.com/tempesta-tech/tempesta/issues/484).

PR #870, opened 2017-12-08
Add tests against memory leaks

vladtcvs:
#758

### DIFF
--- a/tempesta_fw/t/functional/leaks/test_leak.py
+++ b/tempesta_fw/t/functional/leaks/test_leak.py
@@ -1,0 +1,144 @@
+"""
+Testing for memory leaks
+"""
+
+__author__ = 'Tempesta Technologies, Inc.'
+__copyright__ = 'Copyright (C) 2017 Tempesta Technologies, Inc.'
+__license__ = 'GPL2'
+
+import unittest
+import re
+from time import sleep
+
+from testers import stress
+from helpers import tf_cfg, control, tempesta, remote
+
+def drop_caches():
+    """ Drop caches """
+    remote.tempesta.run_cmd("echo 3 > /proc/sys/vm/drop_caches")
+    sleep(1)
+
+def file_exists(remote_file):
+    """ Check existance of file on Tempesta host """
+    check_cmd = "if [ -e %s ]; then echo -n yes; fi" % remote_file
+    [stdout, stderr] = remote.tempesta.run_cmd(check_cmd)
+    if stdout != "yes":
+        return False
+    return True
+
+def has_kmemleak():
+    """ Check presence of kmemleak """
+    return file_exists("/sys/kernel/debug/kmemleak")
+
+def has_meminfo():
+    """ Check presence of meminfo """
+    return file_exists("/proc/meminfo")
+
+def read_kmemleaks():
+    """ Get amount of kmemleak records """
+    kmemleakfile = "/sys/kernel/debug/kmemleak"
+    if not has_kmemleak():
+        tf_cfg.dbg(1, "kmemleak file does not exists")
+        return -1
+    cmd = "cat %s | grep \"unreferenced object\" | wc -l" % kmemleakfile
+    [stdout, stderr] = remote.tempesta.run_cmd(cmd)
+    return int(stdout)
+
+def get_memory_line(name):
+    """ Get value from /proc/meminfo """
+    if not has_meminfo():
+        return -1
+    [stdout, stderr] = remote.tempesta.run_cmd("cat /proc/meminfo")
+    line = re.search("%s:[ ]+([0-9]+)" % name, stdout)
+    if line:
+        return int(line.group(1))
+    return -1
+
+def slab_memory():
+    """ Get amount of slab used memory """
+    drop_caches()
+    slabmem = get_memory_line("Slab")
+    return slabmem
+
+def used_memory():
+    """ Measure total memory usage """
+    drop_caches()
+    totalmem = get_memory_line("MemTotal")
+    if totalmem == -1:
+        return -1
+    freemem = get_memory_line("MemFree")
+    if freemem == -1:
+        return -1
+    return totalmem - freemem
+
+class LeakTest(stress.StressTest):
+    """ Leaks testing """
+    config = 'cache 0;\n'
+    backend_connections = 10
+    memory_leak_thresold = 32*1024 # in kib
+
+    def assert_tempesta(self):
+        """ We don't check that traffic is parsed correctly. Only detect leaks. """
+
+    def assert_clients(self):
+        """ Check only traffic size. We don't need to check responses. """
+        cl_req_cnt = 0
+        cl_conn_cnt = 0
+        for client in self.clients:
+            req, err = client.results()
+            cl_req_cnt += req
+            cl_conn_cnt += client.connections * self.pipelined_req
+        exp_min = cl_req_cnt
+        exp_max = cl_req_cnt + cl_conn_cnt
+        self.assertTrue(
+            self.tempesta.stats.cl_msg_received >= exp_min and
+            self.tempesta.stats.cl_msg_received <= exp_max,
+            "Wrong cl_msg_received"
+        )
+
+    def create_servers(self):
+        """ Overrirde to create needed amount of upstream servers. """
+        port = tempesta.upstream_port_start_from()
+        self.servers = [control.Nginx(listen_port=port)]
+        for server in self.servers:
+            server.config.set_ka(self.backend_connections)
+
+    def test_kmemleak(self):
+        """ Detecting leaks with kmemleak """
+        if not has_kmemleak():
+            self.tempesta = None
+            self.servers = None
+            return unittest.TestCase.skipTest(self, "No kmemleak")
+
+        kml1 = read_kmemleaks()
+        self.generic_test_routine(self.config)
+        kml2 = read_kmemleaks()
+        self.assertTrue(kml1 == kml2)
+
+    def test_slab_memory(self):
+        """ Detecting leaks with slab memory measure """
+        if not has_meminfo():
+            self.tempesta = None
+            self.servers = None
+            return unittest.TestCase.skipTest(self, "No meminfo")
+
+        used1 = slab_memory()
+        self.generic_test_routine(self.config)
+        self.tearDown()
+        used2 = slab_memory()
+        tf_cfg.dbg(2, "used %i kib of slab memory=%s kib - %s kib" % (used2 - used1, used2, used1))
+        self.assertTrue(used2 - used1 < self.memory_leak_thresold)
+
+    def test_used_memory(self):
+        """ Detecting leaks with total used memory measure """
+        if not has_meminfo():
+            self.tempesta = None
+            self.servers = None
+            return unittest.TestCase.skipTest(self, "No meminfo")
+
+        used1 = used_memory()
+        self.generic_test_routine(self.config)
+        self.tearDown()
+        used2 = used_memory()
+        tf_cfg.dbg(2, "used %i kib of memory = %s kib - %s kib" % (used2 - used1, used2, used1))
+        self.assertTrue(used2 - used1 < self.memory_leak_thresold)
--- a/tempesta_fw/classifier.h
+++ b/tempesta_fw/classifier.h
@@ -29,7 +29,11 @@
 #include "connection.h"
 
 /* Size of classifier private cliet accounting data. */
+#ifdef CONFIG_DEBUG_LOCK_ALLOC
+#define TFW_CLASSIFIER_ACCSZ	512
+#else
 #define TFW_CLASSIFIER_ACCSZ	256
+#endif
 
 typedef struct { char _[TFW_CLASSIFIER_ACCSZ]; } TfwClassifierPrvt;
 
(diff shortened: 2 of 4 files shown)


tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,110 @@
...
+        backend_connections = 10
+        memory_leak_thresold = 16*1024 # in kib
+
+        def read_kmemleaks(self):
+                """Get amount of kmemleak records"""
+                kmemleakfile = "/sys/kernel/debug/kmemleak"
+                if not os.path.isfile(kmemleakfile):
+                        return -1
+                echocmd = "cat %s | grep \"unreferenced object\" | wc -l" % kmemleakfile
+                proc = subprocess.Popen(["sh", "-c", echocmd], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
ikoveshnikov: The test is going to work only if TempestFW and test framework are running on the same host. In all other cases the test will provide useless data with possible false positives. Please see `remote.py` to see, how to run the command on target host (`control.py/Tempesta`, `control.py/Nginx` and many other classes use that approach)

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,110 @@
...
+                result = proc.stdout.read()
+                return int(result)
+
+        def assert_tempesta(self):
+                """Memory leaks detecting"""
+
+        def assert_clients(self):
+                """ Do nothing """
```
ikoveshnikov: I think the reason of multiple assertion disabling must be documented in class doc string. So you won't need the same comment in every `assert_*()` method. As long as i know, it's common for python to form functions with void body like below:
```
def func():
    pass
```

Also, please, use only one style for the comments. I mean spaces after quotes.

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,110 @@
...
+                kmemleakfile = "/sys/kernel/debug/kmemleak"
+                if not os.path.isfile(kmemleakfile):
+                        return -1
+                echocmd = "cat %s | grep \"unreferenced object\" | wc -l" % kmemleakfile
+                proc = subprocess.Popen(["sh", "-c", echocmd], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                result = proc.stdout.read()
+                return int(result)
+
+        def assert_tempesta(self):
+                """Memory leaks detecting"""
```
ikoveshnikov: We need to check at least some counters, just to check, that traffic was actually generated and forwarded. same for `assert_clients()`.

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,110 @@
...
+import psutil
+from time import sleep
+
+import unittest
+from testers import stress
+from helpers import tf_cfg, control, tempesta
+
+
+class LeakTest(stress.StressTest):
+        """ Leaks testing """
```
ikoveshnikov: Please, run `pylint2` tool before sending patches. the tool reports of multiple warnings and errors on this file.

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,110 @@
...
+                port = tempesta.upstream_port_start_from()
+                self.servers = [control.Nginx(listen_port=port, ka=self.backend_connections)]
+
+        def test_kmemleak(self):
+                """Detecting leaks with kmemleak"""
+                kml1 = self.read_kmemleaks()
+                self.generic_test_routine(self.config)
+                kml2 = self.read_kmemleaks()
+                if kml1 == -1 or kml2 == -1:
+                        return
```
ikoveshnikov: I think, the test must report warning if `kmemleack` is not accessible.

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,110 @@
...
+                self.assertTrue(kml1 == kml2)
+
+        def drop_caches(self):
+                """Drop caches"""
+                dcfile = open("/proc/sys/vm/drop_caches", "w")
+                dcfile.write("3")
+                dcfile.close()
```
ikoveshnikov: Same as above, the function may (and should) be run on remote node. I see many problems of the same time below, all of them should be fixed.

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,110 @@
...
+                lines = meminfo.readlines()
+                slabmem = -1
+                for line in lines:
+                        if line.startswith("Slab:"):
+                                split = line.split(" ")
+                                split = filter(None, split)
+                                if len(split) != 3:
+                                        break
+                                mem = split[1]
+                                slabmem = int(mem)
```
ikoveshnikov: The single `re.search()` call may help to get rid of the whole loop.

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,110 @@
...
+                return slabmem
+
+        def test_slab_memory(self):
+                """Detecting leaks with slab memory measure"""
+                used1 = self.slab_memory()
+                self.generic_test_routine(self.config)
+                self.tearDown()
+                used2 = self.slab_memory()
+                if used1 == -1 or used2 == -1:
+                        return
```
ikoveshnikov: Same as above, print the warning if required data is not accessible. If the tests is not reasonable without it, throw a test error exception.

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,110 @@
...
+"""
+
+__author__ = 'Tempesta Technologies, Inc.'
+__copyright__ = 'Copyright (C) 2017 Tempesta Technologies, Inc.'
+__license__ = 'GPL2'
+
+import os.path
+import subprocess
+
+import psutil
```
ikoveshnikov: Pylint2: `E: 12, 0: Unable to import 'psutil' (import-error)`. It's a new library required for the tests. Update to Readme is required.

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -326,11 +326,11 @@ def stop(self):
 
 class Nginx(object):
 
-    def __init__(self, listen_port, workers=1):
+    def __init__(self, listen_port, workers=1, ka=100):
```
ikoveshnikov: I don't see clear reason in this change. `Nginx.Config` class has method to configure keep alive settings, which requires _two_ parameters. Here we pass only one of them.

tempesta_fw/t/functional/leaks/test_leak.py
vladtcvs: fixed

tempesta_fw/t/functional/leaks/test_leak.py
vladtcvs: fixed

tempesta_fw/t/functional/leaks/test_leak.py
vladtcvs: fixed

tempesta_fw/t/functional/helpers/control.py
vladtcvs: fixed

tempesta_fw/t/functional/leaks/test_leak.py
vladtcvs: fixes

tempesta_fw/t/functional/leaks/test_leak.py
vladtcvs: fixes

tempesta_fw/t/functional/leaks/test_leak.py
vladtcvs: fixed

tempesta_fw/t/functional/leaks/test_leak.py
vladtcvs: ашчув

tempesta_fw/t/functional/leaks/test_leak.py
vladtcvs: Hm, in original assert_tempesta() there is no such checks...

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -258,13 +259,17 @@ def start(self):
         cmd = '%s/scripts/tempesta.sh --start' % self.workdir
         env = { 'TFW_CFG_PATH': self.config_name }
         self.node.run_cmd(cmd, timeout=30, env=env, err_msg=(self.err_msg % 'start'))
+        self.state = "started"
```
ikoveshnikov: I have tried such approach, it also has an issue. Imagine, you've interrupted (repeated `^C`) a test just in the middle. Then TempestaFW and Nginx will remain alive. If you try to start a new test it will be failed on `setUp()` due to errors on starting TempestaFW and Nginx. Without saving the state you can simply run the same test once again to cleanup the environment.

The problem @intelfx highlighted - exceptions in `tearDown()`, which makes clenup procedure  interrupted if  one of it's steps failed. Every `node.run_cmd()` call may throw an exception and storing the state actually doesn't solve the problem. I even think, that it doesn't provide any advantages.

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,138 @@
...
+            return -1
+        cmd = "cat %s | grep \"unreferenced object\" | wc -l" % kmemleakfile
+        [stdout, stderr] = self.tempesta.node.run_cmd(cmd)
+        return int(stdout)
+
+    def get_memory_line(self, name):
+        """ Get value from /proc/meminfo """
+        [stdout, stderr] = self.tempesta.node.run_cmd("cat /proc/meminfo")
+        line = re.search("%s:[ ]+([0-9]+)" % name, stdout)
+        if len(line.groups()) == 0:
```
ikoveshnikov: simple `if line:` is enough here.

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,138 @@
...
+        cl_conn_cnt = 0
+        for client in self.clients:
+            req, err = client.results()
+            cl_req_cnt += req
+            cl_conn_cnt += client.connections * self.pipelined_req
+        exp_min = cl_req_cnt
+        exp_max = cl_req_cnt + cl_conn_cnt
+        self.assertTrue(
+            self.tempesta.stats.cl_msg_received >= exp_min and
+            self.tempesta.stats.cl_msg_received <= exp_max
```
ikoveshnikov: Please provide assert message, it'll be easier to get, what is wrong with the test

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+
+    def create_clients(self):
+        """ Override to set desired list of benchmarks and their options. """
+        self.clients = [control.Wrk()]
```
ikoveshnikov: The function is absolutely the same as in the base class.

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,138 @@
...
+        self.assertTrue(kml1 == kml2)
+
+    def test_slab_memory(self):
+        """ Detecting leaks with slab memory measure """
+        used1 = self.slab_memory()
+        if used1 == -1:
+            tf_cfg.dbg(1, "slab memory can't be measured")
+            return
+        self.generic_test_routine(self.config)
+        self.tempesta.stop()
```
ikoveshnikov: Ah, I see. You save state of TempestaFW in order not to stop already stopped system during `tearDown()`. It's also possible to set `self.tempesta = None` for the same aim. I believe, we should also stop all servers before measuring free memory since servers can be started on the same node.

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -241,7 +241,6 @@ def clients_parallel_load(client, count=None):
 
 
 class Tempesta(object):
-
```
ikoveshnikov: Better to leave this line here.

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -325,7 +324,6 @@ def stop(self):
 #-------------------------------------------------------------------------------
 
 class Nginx(object):
-
```
ikoveshnikov: Better to leave this line here.

tempesta_fw/t/functional/leaks/test_leak.py
```diff
@@ -0,0 +1,144 @@
...
+        return -1
+    freemem = get_memory_line("MemFree")
+    if freemem == -1:
+        return -1
+    return totalmem - freemem
+
+class LeakTest(stress.StressTest):
+    """ Leaks testing """
+    config = 'cache 0;\n'
+    backend_connections = 10
```
ikoveshnikov: Variable name doesn't correctly represent it's sense. `keepalive_requests` or something like that is more suitable naming.

Issue #628, opened 2016-10-31
Custom characters set for URI and HTTP headers


krizhanovsky:
`str_simd.c` uses number of ASCII characters table to match URI and HTTP header fields. The tables should be made customizable by a user. I.e. user should be able to describe which character sets they allow for particular HTTP message field and Tempesta FW and startup compiles the sets to the tables and use them for the SIMD processing. The filtering is very useful to prevent some types of Web attacks, e.g. [SQL Injections](https://websec.wordpress.com/2010/03/19/exploiting-hard-filtered-sql-injections/).

Must be added separate alphabet for POST arguemnts, so XSS and HTTP parameter fragmentation attacks can be prevented by excluding `<` and/or `*` from the alphabet.

The validator is very close to ModSecurity's [validateByteRange](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#validateByteRange).

Please update https://github.com/tempesta-tech/tempesta/wiki/Web-security and create an issue for a functional test.

krizhanovsky:
For now we do not parse POST body and just skip it, so #902 must use alphabet checking implemented in the issue to really verify POST body.

PR #872, opened 2017-12-15
Add user-defined headers before forward response to client

### LINKED ISSUE #511: HTTP: transform custom HTTP headers
Web-server mode (#471) requires setting of custom HTTP headers. Cache-Control, Expires and other similar server headers can be generated on this basis only. The feature is also useful for various proxy cache use cases.

The new configuration option must be added:

```
    hdr_add <header name> <header value>
```

Configuration example from #471:

```
    vhost tempesta-tech {
            location "?" {
                proxy_pass heavy;
                cache_fulfill;
                hdr_add Cache-Control "no-cache";
            }
            index   index.html;
            root_db /opt/tempesta/db/www_tt.tdb;
            hdr_add Expires "Thu, 15 Apr 2020 20:00:00 GMT"
    }
```

Moreover, proxy mode also requires adjustment of arbitrary HTTP headers, see for example [Nginx's proxy_set_header](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header).

Don't forget to develop appropriate functional test and update Wiki.

ikoveshnikov:
fix #511


etc/tempesta_fw.conf
```diff
@@ -469,6 +469,20 @@
...
+#
+# Syntax:
+#   hdr_add <name> <value>;
+#
+# Example:
+#   hdr_add Expires "Thu, 15 Apr 2020 20:00:00 GMT"
+#
+# Default:
+#   None.
+
```
krizhanovsky: Don't forget to add the description to the Wiki

tempesta_fw/vhost.c
```diff
@@ -74,6 +75,8 @@ static size_t		tfw_capolicy_sz = 0;	/* Current size. */
  */
 #define TFW_NIPDEF_ARRAY_SZ	(64)
```
krizhanovsky: Unnecessary empty line

tempesta_fw/http.c
```diff
@@ -1762,6 +1789,10 @@ tfw_http_adjust_resp(TfwHttpResp *resp, TfwHttpReq *req)
 	if (r < 0)
 		return r;
 
+	r = tfw_http_add_loc_hdrs(hm, req);
+	if (r < 0)
+		return r;
```
krizhanovsky: Not only requests are must be adjustable - an example from the issue `hdr_add Expires "Thu, 15 Apr 2020 20:00:00 GMT"` is about responses.

tempesta_fw/http.c
```diff
@@ -1709,6 +1709,33 @@ tfw_http_add_x_forwarded_for(TfwHttpMsg *hm)
...
+
+	/* TODO #862: req->location must be the full set of options. */
+	if (!loc || !loc->usr_hdrs_sz)
+		loc = req->vhost->loc_dflt;
+	if (!loc || !loc->usr_hdrs_sz)
+		loc = (tfw_vhost_get_default())->loc_dflt;
+	BUG_ON(!loc);
+
+	for (i = 0; i < loc->usr_hdrs_sz; ++i) {
+		r = tfw_http_msg_hdr_add(hm, loc->usr_hdrs[i]);
```
krizhanovsky: Well, the issue isn't clear: it's titled "transform custom HTTP headers", but specifies `hdr_add` only. There is reference to [Nginx's `proxy_set_header`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header) which adds a header in all cases except Host and Connection headers when it actually replaces them. That's absolutely non-flexible. It's better to allow a user to delete and replace any headers as well as add it, so please add header deletion and substitution as well:

        hdr_del <header name>
        hdr_sub <header name> <header value>

tempesta_fw/http.c
ikoveshnikov: Actually only responses are adjusted in the patch: `tfw_http_add_loc_hdrs()` function is called in the `tfw_http_adjust_resp()` as a part of response processing. Is it required to adjust _requests_ before forwarding them to backend servers?

krizhanovsky: Yeah, my bad - of course I was commenting `tfw_http_adjust_resp()` :)

We absolutely have to be able to adjust requests before forwarding them to a backend. `X-Forwarded-For` is the most trivial example which we hard code for now. Other examples could be security tokens passed to most powerful WAFs or removing some danger headers from suspicious requests.

etc/tempesta_fw.conf
```diff
@@ -486,6 +486,66 @@
...
+# Example:
+#   req_hdr_add X-Forwarded-Host "mydomain.com"
+#
+# Default:
+#   None.
+
+# TAG: req_hdr_set
+#
+#  Replace or remove a user-defined header from HTTP request message before
+# forwarding it to backend server.
```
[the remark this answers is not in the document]
ikoveshnikov: The only difference between `req_hdr_add` and `req_hdr_set` is ability of `_add` to append user defined string to the existing header. In this case original value of header will be saved, but a user defined string will be added to the header. If the feature is not required, `req_hdr_add` directive can be easily removed.

E.g. for the configuration `resp_hdr_add Set-Cookie "ck=djnfsdnfkdn"` the header
`Set-Cookie: c=fvfkm` will be transformed into `Set-Cookie: c=fvfkm, ck=djnfsdnfkdn`

tempesta_fw/http_msg.c
```diff
@@ -79,7 +140,6 @@ __http_msg_hdr_val(TfwStr *hdr, unsigned id, TfwStr *val, bool client)
 		[TFW_HTTP_HDR_X_FORWARDED_FOR] = SLEN("X-Forwarded-For:"),
 		[TFW_HTTP_HDR_KEEP_ALIVE] = SLEN("Keep-Alive:"),
 		[TFW_HTTP_HDR_TRANSFER_ENCODING] = SLEN("Transfer-Encoding:"),
-		[TFW_HTTP_HDR_USER_AGENT] = SLEN("User-Agent:"),
```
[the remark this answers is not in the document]
ikoveshnikov: It looks confusing to init the same array member twice, since
https://github.com/tempesta-tech/tempesta/blob/81b45c2eb6b7682d934ea821c41dad6b2d24387c/tempesta_fw/http.h#L186

The `User-agent` case is checked in a few lines below:
https://github.com/tempesta-tech/tempesta/blob/81b45c2eb6b7682d934ea821c41dad6b2d24387c/tempesta_fw/http_msg.c#L100

tempesta_fw/http_msg.c
```diff
@@ -26,6 +26,67 @@
...
+		[TFW_HTTP_HDR_COOKIE]	= TfwStr_string("Cookie:"),
+		[TFW_HTTP_HDR_ETAG]	= TfwStr_string("ETag:"),
+		[TFW_HTTP_HDR_CONNECTION] = TfwStr_string("Connection:"),
+		[TFW_HTTP_HDR_X_FORWARDED_FOR] = TfwStr_string("X-Forwarded-For:"),
+		[TFW_HTTP_HDR_KEEP_ALIVE] = TfwStr_string("Keep-Alive:"),
+		[TFW_HTTP_HDR_TRANSFER_ENCODING] = TfwStr_string("Transfer-Encoding:"),
+	#undef TfwStr_string
+	};
+	DEFINE_TFW_STR(user_agent, "User-Agent:");
+	DEFINE_TFW_STR(if_none_match, "If-None-Match:");
```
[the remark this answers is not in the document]
ikoveshnikov: According to `tfw_http_hdr_t`:
https://github.com/tempesta-tech/tempesta/blob/81b45c2eb6b7682d934ea821c41dad6b2d24387c/tempesta_fw/http.h#L186
https://github.com/tempesta-tech/tempesta/blob/81b45c2eb6b7682d934ea821c41dad6b2d24387c/tempesta_fw/http.h#L189
So the headers can't be listed in the array due to duplicated indexes.

tempesta_fw/http_msg.c
```diff
@@ -26,6 +26,67 @@
...
+
+	int i;
+
+	for (i = 0; i < TFW_HTTP_HDR_RAW; ++i)
+		if (!tfw_stricmpspn(hdr, &spec_hdrs[i], ':'))
+			return i;
+	if (!tfw_stricmpspn(hdr, &user_agent, ':'))
+		return TFW_HTTP_HDR_USER_AGENT;
+	if (!tfw_stricmpspn(hdr, &if_none_match, ':'))
+		return TFW_HTTP_HDR_IF_NONE_MATCH;
```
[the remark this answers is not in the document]
ikoveshnikov: Actually there is `__tfw_http_msg_find_hdr()` function for binary search. But it requires alphabetically sorted array. That's not true for the array since original `tfw_http_hdr_t` is not sorted. The function is called only at configuration processing, so i've decided to keep it as clean as possible

etc/tempesta_fw.conf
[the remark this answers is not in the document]
krizhanovsky: So `resp_hdr_set` deletes a header, adds it to a HTTP message or replaces a header if it's already present, is it correct? If so, then please describe the difference in https://github.com/tempesta-tech/tempesta/wiki/Modify-HTTP-Messages , preferably with some examples.

tempesta_fw/http_msg.c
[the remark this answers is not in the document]
krizhanovsky: Yes, this why I wrote "It's better to use bit different code for requests and responses. The function is used in tfw_cfgop_mod_hdr() only and there you have msg_type." It's can be figured out why the code looks exactly this, but from first glance it looks strange and it's better to rework it in more straightforward way.

tempesta_fw/http_msg.c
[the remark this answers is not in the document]
krizhanovsky: Ok, then just point out this in a comment.

My concern is that we have several library-like files (str.[ch], ss_skb.[ch], http_msg.[ch]) which contain logic useful in various situations. The function also looks like very useful in future, so it's necessary either to limit it's usage or write in in very solid way.

tempesta_fw/http_msg.c
```diff
@@ -64,6 +90,58 @@ __tfw_http_msg_find_hdr(const TfwStr *hdr, const TfwStr array[], size_t size)
 }
 EXPORT_SYMBOL(__tfw_http_msg_find_hdr);
 
+/**
+ * Get header id in header table for header @hdr.
+ *
+ * TODO: return error if header is not allowed in @msg_type.
```
krizhanovsky: Good topic to discuss and think about. Really, having `Server` header in a request doesn't look normal. However, there are Web applications setting custom headers. Moreover, there are proprietary soft (e.g. AntiBViruses) actively working with their clouds using HTTP and the soft sometimes sends quite strange traffic.

Maybe this is a subject for machine learning (to learn the normal set of HTTP headers). Maybe we should provide a configuration option to allow a user to specify which headers are allowed, like we're going to do this for HTTP strings alphabets in #628.

tempesta_fw/http_msg.c
```diff
@@ -26,14 +26,40 @@
...
+			{ .ptr = (void *)val,	.len = v_len },
+		},
+		.len = n_len + SLEN(S_DLM) + v_len,
+		.eolen = 2,
+		.flags = (val ? 3 : 2) << TFW_STR_CN_SHIFT
+	};
+
+	return tfw_strdup(pool, &tmp_hdr);
+}
+EXPORT_SYMBOL(tfw_http_msg_make_hdr);
```
krizhanovsky: There is no need to export the symbol. Please never export symbols in vain and for unit tests (I fixed several such exports)!

tempesta_fw/http_msg.c
ikoveshnikov: This function checks only headers added by end user. The `(resp|req)_hdr_set` gives user opportunity to make very strange requests and responses and even broke network interactions between agents and server. But user must know what he does. 

>    Maybe we should provide a configuration option to allow a user to specify which headers are allowed

It's already here: user can block specific headers in network by with `(resp|req)_hdr_set <SomeHeader>`; all the headers <SomeHeader> will be removed from forwarded traffic.

tempesta_fw/http_msg.c
```diff
@@ -26,14 +26,39 @@
 #include "http_msg.h"
 #include "ss_skb.h"
 
+/**
+ * Build TfwStr repesented HTTP header.
```
aleksostapenko: `repesented` -> `represented`

tempesta_fw/http_msg.c
```diff
@@ -64,6 +89,73 @@ __tfw_http_msg_find_hdr(const TfwStr *hdr, const TfwStr array[], size_t size)
...
+		TfwStrDefV("X-Forwarded-For:",	TFW_HTTP_HDR_X_FORWARDED_FOR),
+	};
+
+	BUILD_BUG_ON(ARRAY_SIZE(resp_hdrs) != TFW_HTTP_HDR_RAW);
+
+	return __tfw_http_msg_spec_hid(hdr, resp_hdrs);
+}
+
+/**
+ * Get header id in requestf header table for header @hdr.
```
aleksostapenko: `requestf` -> `request`

tempesta_fw/http_msg.c
```diff
@@ -532,49 +613,37 @@ __hdr_sub(TfwHttpMsg *hm, char *name, size_t n_len, char *val, size_t v_len,
 
 /**
  * Transform HTTP message @hm header with identifier @hid.
- * Raw header transforers must provide the header name by @name and @n_len.
- * If @val is NULL, then the header will be deleted from @hm.
+ * @hdr must be compaund string and containt two or three parts:
```
aleksostapenko: `containt` -> `contain`

tempesta_fw/http_msg.c
```diff
@@ -45,7 +70,7 @@ __tfw_http_msg_find_hdr(const TfwStr *hdr, const TfwStr array[], size_t size)
 
 	while (start < end) {
 		size_t mid = start + (end - start) / 2;
-		const TfwStr *sh = &array[mid];
+		const TfwStr *sh = (array + mid * member_sz);
```
aleksostapenko: It seems that brackets are excess here.

tempesta_fw/http_msg.c
```diff
@@ -26,14 +26,39 @@
...
+		.len = n_len + SLEN(S_DLM) + v_len,
+		.eolen = 2,
+		.flags = (val ? 3 : 2) << TFW_STR_CN_SHIFT
+	};
+
+	return tfw_strdup(pool, &tmp_hdr);
+}
+
 /**
  * Find @hdr in @array. @array must be in lowercase and sorted in alphabetical
```
aleksostapenko: Is the lowercase requirement still relevant for this function?

tempesta_fw/vhost.c
```diff
@@ -310,6 +319,28 @@ tfw_vhost_match(TfwStr *arg)
 	return &tfw_vhost_dflt;
 }
 
+/**
+ * Find a headers modification description according to target message type
+ * and current location.
+ *
+ * @req		- http request to deturmine location;
+ * @msg_type	- Target message type (TFW_HTTP_MSG_REQ or TFW_HTTP_MSG_RESP).
```
aleksostapenko: Arguments description doesn't seem to reflect current function signature.

tempesta_fw/vhost.c
```diff
@@ -99,8 +104,12 @@ static TfwLocation tfw_location_dflt = {
 	.capo_sz = 0,
 	.nipdef = tfw_nipdef_dflt,
 	.nipdef_sz = 0,
+	.mod_hdrs = { {0, tfw_hmods_req_dflt}, {0, tfw_hmods_resp_dflt}}
 };
 
+/* Pool for 'hdr_add' directive (TfwLocation->usr_hdrs) */
```
aleksostapenko: Seems like obsolete comment: `TfwLocation->usr_hdrs`.

PR #874, opened 2017-12-19
 Add wrk mixed requests tests

### LINKED ISSUE #765: Stress tests with real life requests
Current our functional tests use mostly simple GET requests, but all the implemented (and at least several not implemented!) request types must be tested. Many HTTP headers (also which we parse and not aware about) must be tested. All the requests must be run with significant load.

Keep in mind that we're developing our own high-peformance fuzzer [TfwBomber](https://github.com/tempesta-tech/tempesta/issues/375), so the test for the issue must be relatively simple and take not so much time to execute as a part of whole functional tests run. The real fuzzing is left for TfwBomber.

vladtcvs:
GET requests

HEAD GET requests

GET POST requests

POST requests with small body

POST requests with big body

fix #765


tempesta_fw/t/functional/wrk/post_big.lua
```diff
@@ -0,0 +1,885 @@
+wrk.method  = "POST"
+wrk.headers = {["Content-Type"]="text/plain"}
+wrk.body    = [[
+    I
+Мой дядя самых честных правил,
```
ikoveshnikov: Unexpected but funny. Usually i saw that some repeated pattern was used for that. It's said that it's more helpful to catch garbaging of recieve-responce buffers. That do you think here?

tempesta_fw/t/functional/wrk/post_small.lua
```diff
@@ -0,0 +1,3 @@
+wrk.method  = "POST"
+wrk.headers = {["Content-Type"]="text/plain"}
+wrk.body    = "some not very long text\n"
```
ikoveshnikov: RFC 7230 3.3.2: `... Content-Length header field is normally sent in a POST request ...`
Ether `Transfer-Encoding` or `Content-Length` should be set for this request or Request will be treated as request without body. Look for the TempestaFW log, you should receive multiple parsing errors for this test.
See RFC 7230 3.3.3 for more details.

tempesta_fw/t/functional/wrk/post_big.lua
vladtcvs: I think, in cases of some memory leaks or smth. else, non-repeating text can be more useful, because it's easier to identify data.

tempesta_fw/t/functional/wrk/get_post.lua
```diff
@@ -0,0 +1,13 @@
+init = function(args)
+    local r = {}
+    r[1] = wrk.format("GET", "/")
+    r[2] = wrk.format("GET", "/", {["Accept"] = "text/plain"})
```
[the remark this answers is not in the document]
vladtcvs: Added another test, with real request

tempesta_fw/t/functional/wrk/head_get.lua
```diff
@@ -0,0 +1,10 @@
+init = function(args)
+        local r = {}
+        r[1] = wrk.format("HEAD", "/")
```
ikoveshnikov: As i have noticed recently in the chat, `wrk` handles responses to `HEAD` in a wrong way. It expects body in such response. That lead to very frequent connection drops by `wrk` and the responses are missed in the `wrk` statistics. I believe we should mark tests expected to fail rather than just ignore statistics errors.

Anyway `wrk` is still an only option here, may be we can prepare a patch and send it to upstream is some nearest future.

vladtcvs: > Anyway wrk is still an only option here, may be we can prepare a patch and send it to upstream is some nearest future.

I think, this is the right way

PR #875, opened 2017-12-22
Exception catching in tearDown

### LINKED ISSUE #846: Defensive programming in functional tests
This is something I overlooked when I was cleaning up the functional test suite. Our tests' cleanup routines abort on the first error met, possibly leaving the node in a dirty state (i. e. modules still loaded, nginx still running etc.).

This must be fixed in a general manner — e. g. catch all external exceptions and have a special `error.bug()` that logs the error but does not raise an exception.

vladtcvs:
use try except for catching exceptions in stop functions

fix #846

### DIFF
--- a/tempesta_fw/t/functional/helpers/control.py
+++ b/tempesta_fw/t/functional/helpers/control.py
@@ -4,8 +4,9 @@ from __future__ import print_function
 import abc
 import re
 import os
+import asyncore
 import multiprocessing.dummy as multiprocessing
-from . import tf_cfg, remote, error, nginx, tempesta
+from . import tf_cfg, remote, error, nginx, tempesta, deproxy, stateful
 
 __author__ = 'Tempesta Technologies, Inc.'
 __copyright__ = 'Copyright (C) 2017 Tempesta Technologies, Inc.'
@@ -14,6 +15,7 @@ __license__ = 'GPL2'
 #-------------------------------------------------------------------------------
 # Clients
 #-------------------------------------------------------------------------------
+
 
 class Client(object):
     __metaclass__ = abc.ABCMeta
@@ -96,7 +98,7 @@ class Client(object):
         copied to remote node, present in command line as parameter and
         removed after client finish.
         """
-        os.path.join(self.workdir, filename)
+        full_name = os.path.join(self.workdir, filename)
         self.files.append((filename, content))
         self.options.append('%s %s' % (option, full_name))
         self.cleanup_files.append(full_name)
@@ -238,10 +240,7 @@ def clients_parallel_load(client, count=None):
 #-------------------------------------------------------------------------------
 # Tempesta
 #-------------------------------------------------------------------------------
-
-
-class Tempesta(object):
-
+class Tempesta(stateful.Stateful):
     def __init__(self):
         self.node = remote.tempesta
         self.workdir = self.node.workdir
@@ -251,8 +250,9 @@ class Tempesta(object):
         self.stats = tempesta.Stats()
         self.host = tf_cfg.cfg.get('Tempesta', 'hostname')
         self.err_msg = ' '.join(["Can't %s TempestaFW on", self.host])
+        self.stop_procedures = [self.stop_tempesta, self.remove_config]
 
-    def start(self):
+    def run_start(self):
         tf_cfg.dbg(3, '\tStarting TempestaFW on %s' % self.host)
         self.stats.clear()
         self.node.copy_file(self.config_name, self.config.get_config())
@@ -261,11 +261,12 @@ class Tempesta(object):
         self.node.run_cmd(cmd, timeout=30, env=env,
                           err_msg=(self.err_msg % 'start'))
 
-    def stop(self):
-        """ Stop and unload all TempestaFW modules. """
+    def stop_tempesta(self):
         tf_cfg.dbg(3, '\tStoping TempestaFW on %s' % self.host)
         cmd = '%s/scripts/tempesta.sh --stop' % self.workdir
         self.node.run_cmd(cmd, timeout=30, err_msg=(self.err_msg % 'stop'))
+
+    def remove_config(self):
         self.node.remove_file(self.config_name)
 
     def reload(self):
@@ -297,6 +298,8 @@ class TempestaFI(Tempesta):
             self.modules_dir = '/lib/modules/$(uname -r)/custom/'
         else:
             self.stap_msg = 'Cannot %s stap %s kernel.'
+        self.stop_procedures = [self.letout, self.letout_finish,
+                                self.stop_tempesta, self.remove_config]
 
     def inject_prepare(self):
         if self.module_stap:
@@ -320,23 +323,16 @@ class TempestaFI(Tempesta):
         if self.module_stap:
             self.node.run_cmd('rm -r %s' % self.modules_dir)
 
-    def start(self):
-        Tempesta.start(self)
+    def run_start(self):
+        Tempesta.run_start(self)
         self.inject_prepare()
         self.inject()
-
-    def stop(self):
-        self.letout()
-        self.letout_finish()
-        Tempesta.stop(self)
-
 
 #-------------------------------------------------------------------------------
 # Server
 #-------------------------------------------------------------------------------
 
-class Nginx(object):
-
+class Nginx(stateful.Stateful):
     def __init__(self, listen_port, workers=1):
         self.node = remote.server
         self.workdir = tf_cfg.cfg.get('Server', 'workdir')
@@ -348,11 +344,12 @@ class Nginx(object):
         self.err_msg = "Can't %s Nginx on %s"
         self.active_conns = 0
         self.requests = 0
+        self.stop_procedures = [self.stop_nginx, self.remove_config]
 
     def get_name(self):
         return ':'.join([self.ip, str(self.config.port)])
 
-    def start(self):
+    def run_start(self):
         tf_cfg.dbg(3, '\tStarting Nginx on %s' % self.get_name())
         self.clear_stats()
         # Copy nginx config to working directory on 'server' host.
@@ -364,10 +361,9 @@ class Nginx(object):
         self.node.run_cmd(cmd, ignore_stderr=True,
                           err_msg=(self.err_msg % ('start', self.get_name())))
 
-    def stop(self):
+    def stop_nginx(self):
         tf_cfg.dbg(3, '\tStoping Nginx on %s' % self.get_name())
         pid_file = os.path.join(self.workdir, self.config.pidfile_name)
-        config_file = os.path.join(self.workdir, self.config.config_name)
         cmd = ' && '.join([
             '[ -e \'%s\' ]' % pid_file,
             'pid=$(cat %s)' % pid_file,
@@ -376,6 +372,10 @@ class Nginx(object):
         ])
         self.node.run_cmd(cmd, ignore_stderr=True,
                           err_msg=(self.err_msg % ('stop', self.get_name())))
+
+    def remove_config(self):
+        tf_cfg.dbg(3, '\tRemoving Nginx config for %s' % self.get_name())
+        config_file = os.path.join(self.workdir, self.config.config_name)
         self.node.remove_file(config_file)
 
     def get_stats(self):
@@ -419,6 +419,11 @@ def servers_start(servers):
     threads = __servers_pool_size(len(servers))
     pool = multiprocessing.Pool(threads)
     pool.map(Nginx.start, servers)
+
+def servers_force_stop(servers):
+    threads = __servers_pool_size(len(servers))
+    pool = multiprocessing.Pool(threads)
+    pool.map(Nginx.force_stop, servers)
 
 def servers_stop(servers):
     threads = __servers_pool_size(len(servers))
--- a/tempesta_fw/t/functional/helpers/deproxy.py
+++ b/tempesta_fw/t/functional/helpers/deproxy.py
@@ -8,7 +8,7 @@ import sys
 import time
 import calendar # for calendar.timegm()
 from  BaseHTTPServer import BaseHTTPRequestHandler
-from . import error, tf_cfg, tempesta
+from . import error, tf_cfg, tempesta, stateful
 
 
 __author__ = 'Tempesta Technologies, Inc.'
@@ -450,7 +450,7 @@ class Response(HttpMessage):
 #-------------------------------------------------------------------------------
 MAX_MESSAGE_SIZE = 65536
 
-class Client(asyncore.dispatcher):
+class Client(asyncore.dispatcher, stateful.Stateful):
 
     def __init__(self, host=None, port=80):
         asyncore.dispatcher.__init__(self)
@@ -460,10 +460,22 @@ class Client(asyncore.dispatcher):
         self.tester = None
         if host is None:
             host = 'Tempesta'
-        addr = tf_cfg.cfg.get(host, 'ip')
-        tf_cfg.dbg(4, '\tDeproxy: Client: Connect to %s:%d.' % (addr, port))
+        self.addr = tf_cfg.cfg.get(host, 'ip')
+        self.port = port
+        self.stop_procedures = [self.__stop_client]
+        self.orig_addr = ''
+
+    def __stop_client(self):
+        tf_cfg.dbg(4, '\tStop deproxy client')
+        self.close()
+        self.addr = self.orig_addr
+
+    def run_start(self):
+        self.orig_addr = self.addr
+        tf_cfg.dbg(3, '\tStarting deproxy client')
+        tf_cfg.dbg(4, '\tDeproxy: Client: Connect to %s:%d.' % (self.addr, self.port))
         self.create_socket(socket.AF_INET, socket.SOCK_STREAM)
-        self.connect((addr, port))
+        self.connect((self.addr, self.port))
 
     def clear(self):
         self.request_buffer = ''
@@ -579,7 +591,7 @@ class ServerConnection(asyncore.dispatcher_with_send):
                 pass
 
 
-class Server(asyncore.dispatcher):
+class Server(asyncore.dispatcher, stateful.Stateful):
 
     def __init__(self, port, host=None, conns_n=None, keep_alive=None):
         asyncore.dispatcher.__init__(self)
@@ -593,21 +605,16 @@ class Server(asyncore.dispatcher):
         if host is None:
             host = 'Client'
         self.ip = tf_cfg.cfg.get('Client', 'ip')
-        tf_cfg.dbg(4, '\tDeproxy: Server: Start on %s:%d.' % (self.ip, port))
-        self.setup()
+        self.stop_procedures = [self.__stop_server]
 
-    def setup(self):
+    def run_start(self):
+        tf_cfg.dbg(4, '\tDeproxy: Server: Start on %s:%d.' % (self.ip, self.port))
         self.create_socket(socket.AF_INET, socket.SOCK_STREAM)
         self.set_reuse_addr()
         self.bind((self.ip, self.port))
         self.listen(socket.SOMAXCONN)
 
-    def restart(self):
-        asyncore.dispatcher.__init__(self)
-        self.tester.servers.append(self)
-        self.setup()
-
-    def stop(self):
+    def __stop_server(self):
         tf_cfg.dbg(4, '\tDeproxy: Server: Stop on %s:%d.' % (self.ip,
                                                              self.port))
         self.close()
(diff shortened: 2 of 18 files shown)


tempesta_fw/t/functional/helpers/control.py
```diff
@@ -363,20 +400,39 @@ def start(self):
         cmd = ' '.join([tf_cfg.cfg.get('Server', 'nginx'), '-c', config_file])
         self.node.run_cmd(cmd, ignore_stderr=True,
                           err_msg=(self.err_msg % ('start', self.get_name())))
+        self.state = "started"
+
+    def force_stop(self):
```
ikoveshnikov: both `Nginx` and `Tempesta` classes has the same "stateful" approach. (Seems quite the same for client classes.) Isn't it better to derive them from some common  class? 
As i see this, being statefull is very daemon specific: you can't wrap everything  in one big try-catch block, you have to divide the full procedure into small steps and have a separate try-catch block for every action. You've also noticed that and do it in the right way. We can make a base class for `stop()` functions that runs array of stop steps and define that array in every derived class.

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -430,4 +486,22 @@ def servers_get_stats(servers):
     pool = multiprocessing.Pool(threads)
     pool.map(Nginx.get_stats, servers)
 
+def stop_old():
```
ikoveshnikov: As we discussed in the chat, that stop should clean up test environment of a given test. Normally that means calling `tearDown()` function. So `./run_tests.sh -C` must take the last test from previous run and call `tearDown()` function of it.

tempesta_fw/t/functional/testers/functional.py
```diff
@@ -55,13 +55,24 @@ def setUp(self):
...
-            self.client.close()
-        if self.tempesta:
-            self.tempesta.stop()
-        if self.tester:
-            self.tester.close_all()
+        try:
+            asyncore.close_all()
+            if self.client:
+                self.client.close()
+        except:
```
ikoveshnikov: Seems that `client` class must be stopped just in the same way as `Nginx` and `Tempesta` classes. So it's also statefull.

tempesta_fw/t/functional/helpers/control.py
vladtcvs: agree

tempesta_fw/t/functional/helpers/control.py
vladtcvs: we can not run tearDown(), because tearDown() expects, that test structures were initialized before

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -15,6 +16,43 @@
...
+
+        if self.state != STATE_ERROR:
+            self.state = STATE_STOPPED
+
+    def stop(self):
+        """ Try to stop object """
+        if self.state != STATE_STARTED:
+            tf_cfg.dbg(2, "Not started")
+            return
+        self.force_stop()
```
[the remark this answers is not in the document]
vladtcvs: No, if we have error state, we shouldn't perform stop, but we still can run force_stop

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -423,7 +497,7 @@ def servers_start(servers):
 def servers_stop(servers):
     threads = __servers_pool_size(len(servers))
     pool = multiprocessing.Pool(threads)
-    pool.map(Nginx.stop, servers)
+    pool.map(Nginx.force_stop, servers)
```
ikoveshnikov: I can't agree here. `stop()` tracks state while `force_stop()`. Why is stateless version is used here?

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -236,12 +274,44 @@ def clients_parallel_load(client, count=None):
...
+        self.client = client
+        self.stop_procedures = [self.stop_client]
+
+    def run_start(self):
+        tf_cfg.dbg(3, '\tStarting deproxy client')
+        self.client.tempesta_connect()
+
+    def stop_client(self):
+        tf_cfg.dbg(3, '\tStop deproxy client')
+        asyncore.close_all()
```
ikoveshnikov: This is very internal for `deproxy.py`. Also `asyncore.close_all()` stops network activity for  all classes that uses `python.asyncore()`. May be current file structure is outdated with introducing statedull clasess and `control.Tempesta` should be moved to `tempesta.Tempesta`, `control.Nginx` to `nginx.Nginx` and so on? What is your opinion?

tempesta_fw/t/functional/run_tests.py
```diff
@@ -110,6 +119,47 @@ def usage():
...
+# For the sake of simplicity, Unconditionally discover all tests and filter them
+# afterwards instead of importing individual tests by positive filters.
+loader = unittest.TestLoader()
+tests = []
+shell.testsuite_flatten(tests, loader.discover('.'))
+
+
+# the default value of fs.nr_open
+nofile = 1048576
+resource.setrlimit(resource.RLIMIT_NOFILE, (nofile, nofile))
```
ikoveshnikov: Somewhere above root privileges are checked. I think these two code fragments should be grouped together to make cleaner why we request root privileges.

tempesta_fw/t/functional/testers/functional.py
```diff
@@ -44,24 +47,50 @@ def create_servers_helper(self, count, start_port=None, keep_alive=None,
...
-            self.tester.close_all()
+        try:
+            if self.client_state:
+                self.client_state.force_stop()
+        except:
+            tf_cfg.dbg(1, 'Unknown exception in stoping client')
+
+        try:
+            if self.tempesta:
+                self.tempesta.force_stop()
```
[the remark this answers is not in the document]
vladtcvs: it is force_stop() because if we use tearDown() in "-C" case, we don't start objects

tempesta_fw/t/functional/testers/functional.py
```diff
@@ -44,24 +47,50 @@ def create_servers_helper(self, count, start_port=None, keep_alive=None,
...
+        except:
+            tf_cfg.dbg(1, 'Unknown exception in stoping Tempesta')
+
+        try:
+            if self.tester_state:
+                self.tester_state.force_stop()
+        except:
+            tf_cfg.dbg(1, 'Unknown exception in stop tester')
```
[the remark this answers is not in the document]
vladtcvs: If we have an error, we should continue stopping other parts

tempesta_fw/t/functional/helpers/control.py
vladtcvs: I'm agree with this

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -15,6 +16,43 @@
 # Clients
 #-------------------------------------------------------------------------------
 
+STATE_STARTED = "started"
+STATE_STOPPED = "stopped"
+STATE_ERROR = "error"
+
+class Stateful:
```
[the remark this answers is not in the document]
vladtcvs: fixed

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -15,6 +16,43 @@
...
+    state = STATE_STOPPED
+    stop_procedures = []
+    # should be overrided
```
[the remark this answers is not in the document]
vladtcvs: fixed

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -15,6 +16,43 @@
...
+    def run_start(self):
```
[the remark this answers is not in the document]
vladtcvs: fixed

tempesta_fw/t/functional/helpers/control.py
vladtcvs: fixed

tempesta_fw/t/functional/sessions/test_cookies.py
```diff
@@ -26,8 +26,8 @@ def test(self):
 
 class TestCookiesSupport(TestNoCookiesSupport):
 
-    def create_tester(self, message_chain):
-        self.tester = cookies.TesterUseCookies(message_chain, self.client,
+    def create_tester(self):
+        self.tester = cookies.TesterUseCookies(self.client,
```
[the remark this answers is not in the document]
vladtcvs: ашчув

tempesta_fw/t/functional/helpers/control.py
vladtcvs: but i think, this should be done in separate PR

tempesta_fw/t/functional/run_tests.py
vladtcvs: fixed

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -236,12 +275,45 @@ def clients_parallel_load(client, count=None):
...
 
-class Tempesta(object):
+    def stop_deproxy(self):
+        self.tester.close_all()
 
+#-------------------------------------------------------------------------------
+# Tempesta
+#-------------------------------------------------------------------------------
+
+
```
ikoveshnikov: These two lines are extra.

ikoveshnikov:
Tried to run tests and got this:
```
1 ik@ikpc ../tempesta/tempesta_fw/t/functional (git)-[vlts-846-exceptions] 
% ./run_tests.py -v                                                  
  File "./run_tests.py", line 10, in <module>
    from helpers import tf_cfg, remote, shell, control
  File "/mnt/Storage/Projects/tempesta/tempesta/tempesta_fw/t/functional/helpers/control.py", line 299, in <module>
    class DeproxyTester(object, Stateful):
TypeError: Error when calling the metaclass bases
    Cannot create a consistent method resolution
order (MRO) for bases object, Stateful
```

tempesta_fw/t/functional/helpers/control.py
vladtcvs: fixed

tempesta_fw/t/functional/testers/functional.py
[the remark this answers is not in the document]
vladtcvs: fixed

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -15,6 +16,46 @@
...
+    def run_start(self):
+        """ Shoul be overrided """
+        pass
+
+    def start(self):
+        """ Try to start object """
+        if self.state != STATE_STOPPED:
+            return
+        self.state = STATE_BEGIN_START
```
ikoveshnikov: It's possible to initialize `self.state` with `ERROR` state. In this case special processing for the `STATE_BEGIN_START` won't be needed and one less state to worry about. What do think?

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -297,6 +371,7 @@ def __init__(self, stap_script, mod=False, mod_name='stap_tempesta'):
             self.modules_dir = '/lib/modules/$(uname -r)/custom/'
         else:
             self.stap_msg = 'Cannot %s stap %s kernel.'
+        self.stop_procedures = [self.stop_tempesta, self.remove_config]
```
ikoveshnikov: Already defined in base `Tempesta` class. no need to redefine here.

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -320,23 +395,28 @@ def letout_finish(self):
...
         self.inject()
 
-    def stop(self):
-        self.letout()
-        self.letout_finish()
-        Tempesta.stop(self)
+    def stop_tempesta(self):
+        letout_fail = None
+        try:
+            self.letout()
```
ikoveshnikov: These to functions (`self.letout(); self.letout_finish()`) should be moved to `self.stop_procedures[]`. It will do just what you want.

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -320,23 +395,28 @@ def letout_finish(self):
...
+            self.letout_finish()
+        except Exception as exc:
+            letout_fail = exc.message
+        Tempesta.stop_tempesta(self)
+        if letout_fail != None:
+            raise Exception("letout finish failed: %s" % letout_fail)
```
ikoveshnikov: The whole patch was about not throwing exceptions during stop process to let all classes make theirs best to cleanup environment. This line breaks all your efforts. Take your time, don't hurry.

tempesta_fw/t/functional/testers/functional.py
```diff
@@ -44,24 +47,70 @@ def create_servers_helper(self, count, start_port=None, keep_alive=None,
...
+        # Close client connection before stopping the TempestaFW.
+        if self.client_state:
+            self.client_state.force_stop()
+
+        if self.tempesta:
+            self.tempesta.force_stop()
+
+        if self.tester_state:
+            self.tester_state.force_stop()
+
```
ikoveshnikov: So many empty lines are not required in the function.

tempesta_fw/t/functional/testers/functional.py
```diff
@@ -44,24 +47,70 @@ def create_servers_helper(self, count, start_port=None, keep_alive=None,
...
+        if self.tester_state:
+            self.tester_state.stop()
+
+        try:
+            control.finish_all_deproxy()
+        except:
+            print ('Unknown exception in stopping deproxy')
+
+        if self.client_state.state == control.STATE_ERROR:
```
ikoveshnikov:
```
for proc in [client, server, ....]:
    if proc.state == STATE_ERROR:
        raise Exception("Error during stopping %s" % proc.__class__.__name__)
```
will do the trick without copy-paste. What do you think?

tempesta_fw/t/functional/helpers/control.py
vladtcvs: I think it is a bad idea.

vladtcvs: It can lead to unexpected behaviour

vladtcvs: Testing framework it is a bad place for some 'hacks' and optimizations. It must be straightforward

vladtcvs: When we begin start procedure, we are not in error state

vladtcvs: So we must not set state to error

tempesta_fw/t/functional/helpers/control.py
vladtcvs: for some reason, without whis explicit definition, overriding worked bad and original functions called

tempesta_fw/t/functional/testers/functional.py
vladtcvs: empty lines make code more readable

tempesta_fw/t/functional/testers/functional.py
vladtcvs: It is possible

tempesta_fw/t/functional/helpers/control.py
vladtcvs: fixed

tempesta_fw/t/functional/helpers/control.py
vladtcvs: fixed

tempesta_fw/t/functional/testers/functional.py
vladtcvs: fixed

tempesta_fw/t/functional/regression/test_shutdown.py
```diff
@@ -40,38 +43,41 @@ def init(self):
         defconfig = 'cache 0;\n'
         self.tempesta.config.set_defconfig(defconfig)
         self.create_servers()
+        for server in self.servers:
+            server.start("Server")
         self.configure_tempesta()
-        self.tempesta.start()
+        self.tempesta.start("Tempesta")
```
ikoveshnikov: May be it's better to move `obj` from `Stateful.start()` to `Stateful.__init__()`? Then  the object name will be available while `Stateful` class instance is alive.

tempesta_fw/t/functional/sessions/test_cookies.py
```diff
@@ -39,15 +37,13 @@ class TestNoEnforcedCookiesSupport(TestNoCookiesSupport):
         'sticky_secret "f00)9eR59*_/22";\n'
         '\n')
 
-    def create_tester(self, message_chain):
-        self.tester = cookies.TesterIgnoreEnforcedCookies(
-            message_chain, self.client, self.servers)
+    def create_tester(self):
+        self.tester = cookies.TesterIgnoreEnforcedCookies(self.client, self.servers)
```
ikoveshnikov: Too long line.

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -235,13 +237,13 @@ def clients_parallel_load(client, count=None):
     client.parse_out(stdout, stderr)
     client.cleanup()
 
+def finish_all_deproxy():
+    asyncore.close_all()
```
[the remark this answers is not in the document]
vladtcvs: fixed

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -320,23 +326,20 @@ def letout_finish(self):
...
+        Tempesta.run_start(self)
         self.inject_prepare()
         self.inject()
 
-    def stop(self):
-        self.letout()
-        self.letout_finish()
-        Tempesta.stop(self)
+    def stop_tempesta(self):
+        Tempesta.stop_tempesta(self)
```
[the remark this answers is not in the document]
vladtcvs: fixed

tempesta_fw/t/functional/helpers/deproxy.py
```diff
@@ -593,21 +605,16 @@ def __init__(self, port, host=None, conns_n=None, keep_alive=None):
...
 
-    def setup(self):
+    def run_start(self):
+        tf_cfg.dbg(4, '\tDeproxy: Server: Start on %s:%d.' % (self.ip, self.port))
         self.create_socket(socket.AF_INET, socket.SOCK_STREAM)
         self.set_reuse_addr()
         self.bind((self.ip, self.port))
         self.listen(socket.SOMAXCONN)
 
-    def restart(self):
```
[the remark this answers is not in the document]
vladtcvs: Stateful.restart() will be called

tempesta_fw/t/functional/helpers/deproxy.py
```diff
@@ -665,9 +672,9 @@ def empty():
         return MessageChain(Request(), Response())
 
 
-class Deproxy(object):
+class Deproxy(stateful.Stateful):
 
-    def __init__(self, message_chains, client, servers, register=True):
+    def __init__(self, client, servers, register=True, message_chains=[]):
```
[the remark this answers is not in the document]
vladtcvs: Fixed

tempesta_fw/t/functional/helpers/stateful.py
```diff
@@ -0,0 +1,52 @@
...
+            except Exception as exc:
+                tf_cfg.dbg(1, 'Exception in stoping Tempesta: %s' % str(exc))
+                self.state = STATE_ERROR
+
+        if self.state != STATE_ERROR:
+            self.state = STATE_STOPPED
+
+    def stop(self, obj=""):
+        """ Try to stop object """
+        if self.state != STATE_STARTED and self.state != STATE_BEGIN_START:
```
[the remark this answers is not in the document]
vladtcvs: We can try to stop

tempesta_fw/t/functional/sessions/test_cookies.py
vladtcvs: fixed

tempesta_fw/t/functional/regression/test_shutdown.py
vladtcvs: In my variant we can for debug purposes change names in the place

tempesta_fw/t/functional/helpers/remote.py
```diff
@@ -71,7 +71,8 @@ def run_cmd(self, cmd, timeout=DEFAULT_TIMEOUT, ignore_stderr=False,
                 if not err_msg:
                     err_msg = ("Error running command '%s' on %s" %
                                (cmd, self.host))
-                error.bug(err_msg, stdout=stdout, stderr=stderr)
+                raise Exception("Remote error: %s, stdout = %s, stderr = %s" %
```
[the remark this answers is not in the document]
vladtcvs: As i remember, there was finishing of test with erorr.bug()

tempesta_fw/t/functional/helpers/deproxy.py
```diff
@@ -480,7 +492,7 @@ def handle_connect(self):
         pass
 
     def handle_close(self):
-        self.close()
+        pass
```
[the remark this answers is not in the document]
vladtcvs: fixed

PR #877, opened 2017-12-28
Fix #672: Add health monitoring for backend servers.

### LINKED ISSUE #672: Servers failovering by HTTP error codes
Currently server failovering takes into action when we face network layer failure, i.e. if the server is unresponsive. However, we need to failover server also it they return failure return codes (i.e. 404 or 500). So we have to analyze HTTP response codes and initiate server failovering.

A user should be able to specify the behaviour using the new configuration options:

        server_failover_http <code> <number> <time>

, where `<code>` is HTTP response code, `<number>` is the number of the response codes in `<time>` frame in seconds when the server is considered dead. For example:

        server_failover_http 404 3 10;
        server_failover_http 500 10 2;
        server_failover_http 402 10 2;

I.e. the option can be repeated many times. `*` is allowed to specify error codes like `4*` or `5*`, i.e. all error codes greater or equal to 400 or 500 correspondingly.

The feature should be implemented on top of APM extension. The server faulures (network and application layers) are crucial for service monitoring, so APM accounting should be extended by failures on TCP/IP and HTTP layers.

Probably `<time>` specification can be omitted and we can just use time frame configured for APM.

As always, don't forget to update README/Wiki.

aleksostapenko:


tempesta_fw/apm.c
```diff
@@ -1058,6 +1215,149 @@ tfw_apm_del_srv(TfwServer *srv)
...
+tfw_apm_hm_set_srv(const char *name, TfwServer *srv)
+{
+	TfwApmHMCtl *hmctl;
+	TfwApmHM *hm;
+	unsigned long now;
+
+	BUG_ON(!srv->apmref);
+	hmctl = &((TfwApmData *)srv->apmref)->hmctl;
+	list_for_each_entry(hm, &tfw_hm_list, list) {
+		if(!strcasecmp(name, hm->name)) {
```
krizhanovsky: Please place a space between `if` and opening brace. Also it's better to write such loops like
```C
if (strcasecmp(name, hm->name))
     continue;
// the if body
```

tempesta_fw/http.c
```diff
@@ -2088,6 +2154,11 @@ static void
 tfw_http_cli_error_resp_and_log(bool reply, bool nolog, TfwHttpReq *req,
 				int status, const char *msg)
 {
+	if (!req->conn) {
+		tfw_http_msg_free((TfwHttpMsg *)req);
+		return;
+	}
```
krizhanovsky: Later it'll be unclear what requests without a connection are, so please write a comment, or, better since there are  a few such places, write a small function for the requests deletion.

tempesta_fw/str.c
```diff
@@ -668,6 +669,34 @@ tfw_str_next_str_val(const TfwStr *str)
...
+		len += c->len;
+		p += c->len;
+	}
+	BUG_ON(len != str->len);
+	*crc32 = crc32(0, data, len);
+	kfree(data);
+
+	return 0;
+}
+EXPORT_SYMBOL(tfw_str_crc32_calc);
```
krizhanovsky: Please add an appropriate unit test to t/unit/test_tfw_str.c

tempesta_fw/str.c
```diff
@@ -668,6 +669,34 @@ tfw_str_next_str_val(const TfwStr *str)
...
+ */
+int
+tfw_str_crc32_calc(const TfwStr *str, u32 *crc32)
+{
+	const TfwStr *c, *end;
+	char *p, *data;
+	unsigned long len = 0;
+
+	BUG_ON(str->len && !str->ptr);
+	data = p = kmalloc(str->len, GFP_ATOMIC);
```
ikoveshnikov: Why not use `tfw_str_to_cstr()` here?

tempesta_fw/apm.c
```diff
@@ -372,6 +373,103 @@ tfw_stats_update(TfwPcntRanges *rng, unsigned int r_time)
...
+typedef struct {
+	DECLARE_BITMAP(codes, 512);
+	struct list_head	list;
+	char			*name;
+	char			*url;
+	int			urlsz;
+	char			*req;
+	unsigned long		reqsz;
+	u32			crc32;
+	unsigned short		tmt;
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/apm.c
```diff
@@ -372,6 +373,103 @@ tfw_stats_update(TfwPcntRanges *rng, unsigned int r_time)
...
+typedef struct {
+	TfwApmHM		*hm;
+	TfwApmHMStats		*hmstats;
+	atomic64_t		rcount;
+	unsigned long		jtmstamp;
+	struct timer_list	timer;
+	atomic_t		rearm;
+} TfwApmHMCtl;
+
+/* Entry for configureation of separate health monitors. */
```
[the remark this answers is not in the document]
aleksostapenko: Corrected

tempesta_fw/apm.c
aleksostapenko: Corrected.

tempesta_fw/http.c
```diff
@@ -742,6 +742,55 @@ tfw_http_error_resp_switch(TfwHttpReq *req, int status, const char *reason)
...
+	TfwHttpMsg *hm;
+
+	if (!(hm = (TfwHttpMsg *)tfw_pool_new(TfwHttpReq, TFW_POOL_ZERO)))
+		return NULL;
+
+	ss_skb_queue_head_init(&hm->msg.skb_list);
+	INIT_LIST_HEAD(&hm->msg.seq_list);
+	INIT_LIST_HEAD(&((TfwHttpReq *)hm)->fwd_list);
+	INIT_LIST_HEAD(&((TfwHttpReq *)hm)->nip_list);
+	hm->destructor = tfw_http_req_destruct;
```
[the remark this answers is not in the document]

tempesta_fw/http.c
aleksostapenko: Corrected. Replaced these pointer checks with `TFW_HTTP_HMONITOR` request flag verification, according to https://github.com/tempesta-tech/tempesta/pull/877#discussion_r160961463.

tempesta_fw/str.c
aleksostapenko: Removed this function. Used `tfw_str_to_cstr()` instead.

tempesta_fw/apm.c
```diff
@@ -372,6 +373,103 @@ tfw_stats_update(TfwPcntRanges *rng, unsigned int r_time)
 	return;
 }
 
+/* Time granularity for HTTP codes accounting during health monitoring. */
+#define HM_FREQ		10
+
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/apm.c
```diff
@@ -1025,10 +1174,11 @@ int
 tfw_apm_add_srv(TfwServer *srv)
 {
 	TfwApmData *data;
+	bool hm = test_bit(TFW_SRV_B_HMONITOR, (unsigned long *)&srv->flags);
 
 	BUG_ON(srv->apmref);
 
-	if (!(data = tfw_apm_create()))
+	if (!(data = tfw_apm_create(hm)))
 		return -ENOMEM;
```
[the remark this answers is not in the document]
aleksostapenko: Removed `hm` argument from `tfw_apm_create()` due to reconfiguration changes.

tempesta_fw/apm.c
```diff
@@ -1058,6 +1215,149 @@ tfw_apm_del_srv(TfwServer *srv)
...
+	unsigned int i, sum = 0;
+	TfwApmHMStats *hmstats = ((TfwApmData *)apmref)->hmctl.hmstats;
+	TfwApmHMHistory *history = NULL;
+	TfwApmHMCfg *cfg = NULL;
+	unsigned long ts;
+
+	BUG_ON(!hmstats);
+	for (i = 0; i < tfw_hm_codes_cnt; ++i) {
+		if (hmstats[i].hmcfg->code == status ||
+		    hmstats[i].hmcfg->code == (int)(status / 100))
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/apm.c
```diff
@@ -1154,15 +1454,319 @@ tfw_handle_apm_stats(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+		return -EINVAL;
+
+	if (ce->attr_n) {
+		TFW_ERR_NL("Unexpected attributes\n");
+		return -EINVAL;
+	}
+
+	if (tfw_http_parse_status(ce->vals[0], &code)) {
+		TFW_ERR_NL("Unable to parse http code value: '%s'\n",
+			   ce->vals[0] ? ce->vals[0] : "No value specified");
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/http.c
```diff
@@ -1179,6 +1235,13 @@ tfw_http_conn_resched(struct list_head *sch_queue, struct list_head *equeue)
 					     " an available back end server");
 			continue;
 		}
+
+		if (!req->conn && sch_conn->peer != prev_conn->peer) {
```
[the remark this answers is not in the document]
aleksostapenko: Corrected. `TFW_HTTP_HMONITOR` request flag introduced.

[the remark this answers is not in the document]

tempesta_fw/str.c
aleksostapenko: Yes, missed that handy function. Thanks.
Corrected.

tempesta_fw/apm.c
```diff
@@ -1151,14 +1496,318 @@ tfw_cfgop_apm_stats(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+}
+
+static int
+tfw_cfgop_finish_apm_hm(TfwCfgSpec *cs)
+{
+	BUG_ON(!tfw_hm_entry);
+	BUG_ON(list_empty(&tfw_hm_list));
+	if (!tfw_hm_entry->codes && !tfw_hm_entry->crc32) {
+		TFW_ERR_NL("Response codes and crc32 values are not"
+			   " configured for '%s'\n", cs->name);
```
keshonok: The error message is confusing and doesn't report the actual reason for the error. Both of these configuration options are marked as `allow_none` which means that they can be skipped. So the actual error here, I believe, is that at least one of these (or both, for that matter) must be configured.

tempesta_fw/apm.c
```diff
@@ -1151,14 +1496,318 @@ tfw_cfgop_apm_stats(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+	{
+		.name		= "request",
+		.deflt		= "\"GET / HTTTP/1.0\r\n\r\n\"",
+		.handler	= tfw_cfgop_apm_hm_request,
+		.allow_none	= false,
+		.allow_repeat	= false,
+	},
+	{
+		.name		= "request_url",
+		.deflt		= NULL,
```
keshonok: I believe, that If the `request` option has a default value, then it's logical that this option should have a default value as well.

tempesta_fw/apm.c
```diff
@@ -1151,14 +1496,318 @@ tfw_cfgop_apm_stats(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+
+	BUG_ON(!tfw_hm_entry);
+	if (tfw_cfg_check_single_val(ce))
+		return -EINVAL;
+
+	size = strlen(ce->vals[0]);
+	tfw_hm_entry->req = (char *)__get_free_pages(GFP_KERNEL,
+						     get_order(size));
+	if (!tfw_hm_entry->req) {
+		TFW_ERR_NL("Can't allocate memory for helth"
```
keshonok: `helth` -> `health`

tempesta_fw/apm.c
```diff
@@ -1151,14 +1496,318 @@ tfw_cfgop_apm_stats(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+
+static int
+tfw_cfgop_begin_apm_hm(TfwCfgSpec *cs, TfwCfgEntry *ce)
+{
+	TfwApmHM *hm;
+	size_t size;
+
+	if (tfw_cfg_check_val_n(ce, 1))
+		return -EINVAL;
+
```
keshonok: My understanding is that these two checks are congenerical -  about a very similar thing, simple, and logically in the same realm. Usually there's no need for an empty line between checks like these as they are not logically separated. There's lot of examples in the code for this.

tempesta_fw/apm.c
```diff
@@ -1151,14 +1496,318 @@ tfw_cfgop_apm_stats(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
 
+static int
+tfw_cfgop_apm_server_failover(TfwCfgSpec *cs, TfwCfgEntry *ce)
+{
+	TfwApmHMCfg *hm_entry;
+	int code, limit, tframe;
+
+	if (tfw_cfg_check_val_n(ce, 3))
+		return -EINVAL;
+
```
keshonok: My understanding is that these two checks are congenerical -  about a very similar thing, simple, and logically in the same realm. Usually there's no need for an empty line between checks like these as they are not logically separated. There's lot of examples in the code for this.

tempesta_fw/apm.c
```diff
@@ -1151,14 +1496,318 @@ tfw_cfgop_apm_stats(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+}
+
+static int
+tfw_cfgop_apm_hm_timeout(TfwCfgSpec *cs, TfwCfgEntry *ce)
+{
+	int timeout;
+
+	if (tfw_cfg_check_single_val(ce))
+		return -EINVAL;
+
```
keshonok: My understanding is that these two checks are congenerical -  about a very similar thing, simple, and logically in the same realm. Usually there's no need for an empty line between checks like these as they are not logically separated. There's lot of examples in the code for this.

tempesta_fw/http_msg.c
```diff
@@ -868,17 +869,18 @@ tfw_http_msg_alloc(int type)
...
+	if (intent != HTTP_MSG_LIGHT) {
+		hm->h_tbl = (TfwHttpHdrTbl *)tfw_pool_alloc(hm->pool, TFW_HHTBL_SZ(1));
+		if (unlikely(!hm->h_tbl)) {
+			TFW_WARN("Insufficient memory to create message\n");
+			tfw_pool_destroy(hm->pool);
+			return NULL;
+		}
+		hm->h_tbl->size = __HHTBL_SZ(1);
+		hm->h_tbl->off = TFW_HTTP_HDR_RAW;
+		memset(hm->h_tbl->tbl, 0, __HHTBL_SZ(1) * sizeof(TfwStr));
```
keshonok: These changes leave a strange feeling of being half-done. On one hand, the new argument allows for creating of special messages without the header table that is required for the parser. On the other hand, the parser is still initialized regardless of that in the code that follows. Then, there's `tfw_http_msg_alloc_err_resp()` that basically does the same thing but it's still there and not removed.

Now the new argument is required for all allocations, and it looks somewhat ugly.

I believe this (and related) code should be refactored. The original top-level interface should stay the same in the form of `tfw_http_msg_alloc(int type)`. A different interface should be introduced for messages that are not going through the parser AND are not modified - which are the messages that are fully prepared from scratch by Tempesta. Perhaps, internally, these two different sets of interfaces may use the same common code - either by calling some inline functions or with macros. `tfw_http_msg_alloc_err_resp()` would be superseded with the new interface.

Perhaps, you could call it `tfw_http_msg_alloc_light()` - nothing better comes to mind right now.

tempesta_fw/procfs.c
```diff
@@ -206,6 +208,24 @@ tfw_srvstats_seq_show(struct seq_file *seq, void *off)
 			atomic64_read(&srv->sess_n));
 	seq_printf(seq, "Total schedulable connections\t: %zd\n",
 			srv->conn_n - rc);
+
+	seq_printf(seq, "HTTP health monitor is enabled\t: %d\n", hm);
+	if (hm) {
```
keshonok: Perhaps, `hm_stats` should be declared within this block as it's not used elsewhere.

tempesta_fw/server.h
```diff
@@ -64,9 +65,15 @@ typedef struct {
 	atomic64_t		refcnt;
 	unsigned int		weight;
 	unsigned int		flags;
+	unsigned int		hm_flags;
```
keshonok: Correct me if I am wrong, but my understanding is that both atomic and non-atomic ops can be used on the same variable. There's only a handful of flags, and it's a 32-bit variable. So, perhaps, a separate `hm_flags` is unnecessary?

tempesta_fw/http.c
```diff
@@ -831,8 +879,13 @@ tfw_http_req_fwd_send(TfwSrvConn *srv_conn, TfwServer *srv,
...
 			 __func__, srv_conn, req);
-		tfw_http_req_error(srv_conn, req, equeue, 500,
-				   "request dropped: forwarding error");
+		if (req->flags & TFW_HTTP_HMONITOR) {
+			tfw_http_req_delist(srv_conn, req);
+			tfw_http_msg_free((TfwHttpMsg *)req);
+		}
+		else
+			tfw_http_req_error(srv_conn, req, equeue, 500,
+					   "request dropped: forwarding error");
```
keshonok: Curly braces are required here according to the coding standard (because the other branch has them).

tempesta_fw/http.c
```diff
@@ -2104,6 +2176,11 @@ static void
...
 				int status, const char *msg)
 {
+	if (req->flags & TFW_HTTP_HMONITOR) {
+		tfw_http_msg_free((TfwHttpMsg *)req);
+		return;
+	}
+
 	if (reply) {
 		TfwCliConn *cli_conn = (TfwCliConn *)req->conn;
 		tfw_connection_unlink_msg(req->conn);
```
keshonok: Not visible here, but curly braces are required here in the `else` branch.

tempesta_fw/http.c
```diff
@@ -2174,6 +2256,42 @@ tfw_srv_client_block(TfwHttpReq *req, int status, const char *msg)
...
+	return true;
+}
+
+static inline void
+tfw_http_drop_hmonitor_resp(TfwHttpReq *req, TfwHttpResp *resp)
+{
+	tfw_connection_unlink_msg(resp->conn);
+	tfw_apm_update(((TfwServer *)resp->conn->peer)->apmref,
+		       resp->jrxtstamp,
+		       resp->jrxtstamp - req->jtxtstamp);
```
keshonok: This line fits perfectly in the preceding line of code.

tempesta_fw/http.c
```diff
@@ -831,8 +879,13 @@ tfw_http_req_fwd_send(TfwSrvConn *srv_conn, TfwServer *srv,
 	if (tfw_connection_send((TfwConn *)srv_conn, (TfwMsg *)req)) {
 		TFW_DBG2("%s: Forwarding error: conn=[%p] req=[%p]\n",
 			 __func__, srv_conn, req);
-		tfw_http_req_error(srv_conn, req, equeue, 500,
-				   "request dropped: forwarding error");
+		if (req->flags & TFW_HTTP_HMONITOR) {
+			tfw_http_req_delist(srv_conn, req);
+			tfw_http_msg_free((TfwHttpMsg *)req);
```
keshonok: So, if we're unable to send a health monitoring request, then it's simply dropped without any trace or action upon that? Is that the intended behaviour?

tempesta_fw/sock_srv.c
```diff
@@ -946,6 +959,90 @@ tfw_cfgop_lookup_sg_cfg(const char *name)
...
+
+	if (hname) {
+		if (orig_hm)
+			health->flags |= TFW_SRV_HEALTH_DISABLE;
+		health->flags |= TFW_SRV_HEALTH_ENABLE;
+		health->name = (char *)(health + 1);
+		memcpy(health->name, hname, size);
+		srv = orig_srv ? : new_srv;
+	} else {
+		srv = orig_srv;
```
keshonok: It's difficult to understand the variety of combinations of function's arguments without comments. `orig_srv` may be `NULL`. The code suggest it's definitely NOT `NULL` when `!hname`. Why is that? A proper comment would be helpful.

tempesta_fw/sock_srv.c
```diff
@@ -946,6 +959,90 @@ tfw_cfgop_lookup_sg_cfg(const char *name)
...
+	health->srv = srv;
+	list_add_tail(&health->list, &tfw_cfg_health_lst);
+
+	return 0;
+}
+
+static int
+tfw_cfg_srv_set_health(void)
+{
+	TfwSrvHealth *hth;
```
keshonok: There're some common shortcuts that are understood in a specific way. For instance, nth usually mean n-th consecutive number, or something like that. Of course you can name variables any way you like in a small function, but it's best if the name resemble an unambiguous meaning.

There's a number of "rules" for creating short names, one of which is removing all vowels: `default` -> `dflt`, etc. With that, it would be `health` -> `hlth`. Or just leave it `health`, it's sufficiently short in this particular case. Or even just `h` - as long as it can't be confused with h-th number or something. :-)

tempesta_fw/http.c
```diff
@@ -2174,6 +2256,42 @@ tfw_srv_client_block(TfwHttpReq *req, int status, const char *msg)
...
+		*out = HTTP_STATUS_4XX;
+		break;
+	case '5':
+		*out = HTTP_STATUS_5XX;
+		break;
+	default:
+		return false;
+	}
+	return true;
+}
```
keshonok: Perhaps, this can be made a simple static const array to avoid unnecessary run-time branching?

tempesta_fw/http.c
```diff
@@ -3185,12 +3357,9 @@ tfw_cfgop_resp_body(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
 	}
 
-	if (tfw_cfgop_parse_http_status(ce->vals[0], &code))
-	{
+	if (tfw_cfgop_parse_http_status(ce->vals[0], &code)) {
 		TFW_ERR_NL("Unable to parse 'response_body' value: '%s'\n",
-			   ce->vals[0]
-			   ? ce->vals[0]
-			   : "No value specified");
+			   ce->vals[0]);
```
keshonok: Can't this message be made clearer? What the 'value'? Perhaps, 'HTTP status code'?

tempesta_fw/http.c
```diff
@@ -742,6 +743,53 @@ tfw_http_error_resp_switch(TfwHttpReq *req, int status, const char *reason)
...
+
+	if (!tfw_apm_hm_srv_status_alive(resp->status, srv->apmref)) {
+		body_cstr = tfw_pool_alloc(resp->pool, resp->body.len + 1);
+		if (unlikely(!body_cstr)) {
+			TFW_ERR("Can't allocate memory for hmonitor"
+				" crc32 verification\n");
+			return;
+		}
+		tfw_str_to_cstr(&resp->body, body_cstr, resp->body.len + 1);
+		crc32 = crc32(0, body_cstr, resp->body.len);
```
keshonok: Frankly, this looks like extra unnecessary work. The whole body of a response is copied to memory allocated from pool just to calculate CRC32 on it.

This is a response to a health monitoring request, so I guess it's relatively small, and because it comes from a server it probably comes in minimal number of chunks (i.e. it's not sent one byte at a time). As far as I can see `crc32()` function is perfectly suited for calculating the checksum over data that is located in different chunks (the `seed` argument), so it should be straightforward to write a function that goes over string's chunks and calculates CRC32.

tempesta_fw/http.c
```diff
@@ -742,6 +743,53 @@ tfw_http_error_resp_switch(TfwHttpReq *req, int status, const char *reason)
...
+		tfw_pool_free(resp->pool, body_cstr, resp->body.len + 1);
+		if (!tfw_apm_hm_srv_crc32_alive(crc32, srv->apmref))
+			return;
+	}
+
+	tfw_srv_mark_alive(srv);
+}
+
+static inline void
+tfw_http_srv_hm_update(TfwServer *srv, TfwHttpReq *req)
```
keshonok: This comment is for the whole `http.c` module. The new function names are very diverse and it's difficult to attribute a function to new health monitoring functionality. Why not make them have a common prefix, like `tfw_http_hm_XXX`?

Forexample, this function would be named `tfw_http_hm_srv_update()`.

tempesta_fw/http.c
```diff
@@ -1162,23 +1215,38 @@ tfw_http_conn_collect(TfwSrvConn *srv_conn, struct list_head *sch_queue,
...
 	TfwSrvConn *sch_conn;
 	TfwHttpReq *req, *tmp;
+	TfwServer *srv = (TfwServer *)prev_conn->peer;
 
 	/*
 	 * Process the complete queue and re-schedule all requests
-	 * to other servers/connections.
+	 * to other servers/connections. Health monitoring requests
+	 * must be re-scheduled to the same server (other servers
+	 * may not have enabled health monitor).
```
keshonok: Requests in a connection are rescheduled (which is different from resending) when all attempts to reconnect have failed (there's a configurable number of attempts). Attempts to reconnect may take time which may overlap with health monitoring. That means that by the time a health monitoring request is rescheduled, either the relevant requests (to a configured URL) or another health monitoring request may have went to the server through another connection. Perhaps it's best to check for that before rescheduling a health monitoring request.

On the other hand, rescheduling and sending these "outdated" health monitoring requests won't hurt. If that's the idea, then perhaps it's best to document it.

tempesta_fw/http.c
```diff
@@ -1162,23 +1215,38 @@ tfw_http_conn_collect(TfwSrvConn *srv_conn, struct list_head *sch_queue,
...
 	 */
 	list_for_each_entry_safe(req, tmp, sch_queue, fwd_list) {
-		if (!(sch_conn = tfw_sched_get_srv_conn((TfwMsg *)req))) {
+		if (req->flags & TFW_HTTP_HMONITOR) {
+			sch_conn = srv->sg->sched->sched_srv_conn((TfwMsg *)req,
+								  srv);
+			if (!sch_conn) {
+				list_del_init(&req->fwd_list);
+				tfw_http_msg_free((TfwHttpMsg *)req);
```
keshonok: So, a health monitoring request still cannot be sent to the server. We just remove it and do nothing. Just like when no health monitoring requests are sent out at all. So what's the difference? Is that the intended behaviour?

tempesta_fw/http.c
```diff
@@ -742,6 +743,53 @@ tfw_http_error_resp_switch(TfwHttpReq *req, int status, const char *reason)
...
+				      " for bad responses is exceeded",
+				      &srv->addr);
+		}
+		return;
+	}
+
+	if (!tfw_srv_suspended(srv))
+		return;
+
+	if (!tfw_apm_hm_srv_status_alive(resp->status, srv->apmref)) {
```
keshonok: It's an extra validation, but perhaps it makes sense to check that the response has plain (not chunked) body. Just in case, as the body is then used in crc32 calculation.

tempesta_fw/http.c
```diff
@@ -742,6 +743,53 @@ tfw_http_error_resp_switch(TfwHttpReq *req, int status, const char *reason)
 	tfw_http_send_resp(req, code);
 }
 
+static inline void
+tfw_http_srv_hmonitor(TfwHttpResp *resp)
+{
+	u32 crc32;
+	char *body_cstr;
```
keshonok: Perhaps it's best if these two are be moved to the code block where they are used.

tempesta_fw/sock_srv.c
```diff
@@ -842,6 +842,19 @@ static struct {
...
+	char			*name;
+	TfwServer		*srv;
+	unsigned int		flags;
+} TfwSrvHealth;
+
+/* Types of configuration for server health monitor. */
+#define TFW_SRV_HEALTH_ENABLE		0x0001
+#define TFW_SRV_HEALTH_DISABLE		0x0002
+
+static struct list_head tfw_cfg_health_lst = LIST_HEAD_INIT(tfw_cfg_health_lst);
```
keshonok:
```C
static LIST_HEAD(tfw_cfg_health_lst);
```

tempesta_fw/apm.c
```diff
@@ -1151,14 +1496,318 @@ tfw_cfgop_apm_stats(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
-		.allow_repeat = false,
+		.name		= "apm_stats",
+		.deflt		= "window=300 scale=5",
+		.handler	= tfw_cfgop_apm_stats,
+		.cleanup	= tfw_cfgop_cleanup_apm,
+		.allow_none	= true,
+		.allow_repeat	= false,
+	},
+	{
+		.name		= "server_failover_http",
```
keshonok: The major thing about this patch is that this option should not be managed/processed by the APM module. I believe that architecturally is belongs to HTTP module. APM is about collecting, calculating, and managing various stats data, whereas the stats data itself my come from different sources and for different purposes. So this is a new and distinct stats characteristics, it's collected by the APM module, and the APM may tell the current state of various stats data, in this case - the health of a particular server.

In other words, these configuration options, I believe, belong to `socks_srv.c`, and half of the code should be in `http.c`, especially the parts that send health check http requests.

tempesta_fw/apm.c
```diff
@@ -1058,6 +1213,196 @@ tfw_apm_del_srv(TfwServer *srv)
...
+
+	BUG_ON(!srv->apmref);
+	hm = ((TfwApmData *)srv->apmref)->hmctl.hm;
+	BUG_ON(!hm);
+	if(!strcasecmp(name, hm->name))
+		return true;
+
+	return false;
+}
+
```
krizhanovsky: extra empty line

tempesta_fw/t/functional/regression/test_health_monitor.py
```diff
@@ -0,0 +1,227 @@
...
+        self.stages[0].check_transition(messages)
+        self.current_stage = self.stages.pop(0)
+        self.stages_processed += 1
+        return True
+
+    def get_server_path(self):
+        return 'default/%s:%s' % (self.servers[0].ip, self.servers[0].port)
+
+
+class Stage(object):
```
krizhanovsky: I know this isn't C, but please keep forward declaration. Having most experience in C it's much easier to read code which firstly defines all the staff and after that uses it.

tempesta_fw/t/functional/regression/test_health_monitor.py
```diff
@@ -0,0 +1,227 @@
...
+def make_502_expected():
+    response = deproxy.Response.create(
+        status=502,
+        headers=['Content-Length: 0', 'Connection: keep-alive'],
+        date=deproxy.HttpMessage.date_time_string()
+    )
+    return response
+
+
+class TestHealthMonitor(functional.FunctionalTest):
```
krizhanovsky: Please comment in the code each test case from https://github.com/tempesta-tech/tempesta/pull/877#discussion_r159697039

tempesta_fw/apm.c
aleksostapenko: @keshonok  Thank you for review.

aleksostapenko: Corrected.

tempesta_fw/apm.c
aleksostapenko: Corrected.

tempesta_fw/apm.c
aleksostapenko: Corrected.

tempesta_fw/apm.c
aleksostapenko: Corrected (and two cases below too).

tempesta_fw/http_msg.c
aleksostapenko: Done.

tempesta_fw/http.c
```diff
@@ -1162,23 +1215,38 @@ tfw_http_conn_collect(TfwSrvConn *srv_conn, struct list_head *sch_queue,
...
+				continue;
+			}
+		} else if (!(sch_conn = tfw_sched_get_srv_conn((TfwMsg *)req))) {
 			TFW_DBG("Unable to find a backend server\n");
 			__tfw_http_req_error(req, equeue, 502,
 					     "request dropped: unable to find"
 					     " an available back end server");
 			continue;
+			tfw_http_srv_hm_update((TfwServer *)sch_conn->peer,
+					       req);
```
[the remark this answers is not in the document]

tempesta_fw/apm.c
```diff
@@ -1058,6 +1213,196 @@ tfw_apm_del_srv(TfwServer *srv)
 	srv->apmref = NULL;
 }
 
+void
+tfw_apm_hm_srv_rcount_update(TfwStr *uri_path, void *apmref)
+{
+	TfwApmHMCtl *hmctl = &((TfwApmData *)apmref)->hmctl;
+	TfwApmHM *hm = READ_ONCE(hmctl->hm);
+	tfw_str_eq_flags_t flags = TFW_STR_EQ_CASEI;
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/http_sess.c
```diff
@@ -102,7 +102,7 @@ tfw_http_sticky_send_302(TfwHttpReq *req, StickyVal *sv)
 	TfwHttpMsg *hmresp;
 	char buf[sizeof(*sv) * 2];
 
-	if (!(hmresp = tfw_http_msg_alloc(Conn_Srv)))
+	if (!(hmresp = tfw_http_msg_alloc(Conn_Srv, HTTP_MSG_DEFAULT)))
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/apm.c
aleksostapenko: Corrected.

tempesta_fw/t/functional/regression/test_health_monitor.py
aleksostapenko: Corrected.

tempesta_fw/t/functional/regression/test_health_monitor.py
aleksostapenko: Done.

tempesta_fw/procfs.c
aleksostapenko: Corrected.

tempesta_fw/http.c

tempesta_fw/http.c

tempesta_fw/http.c

tempesta_fw/http.c
aleksostapenko: Warning added.

tempesta_fw/sock_srv.c
aleksostapenko: This has been redesigned.

tempesta_fw/sock_srv.c
aleksostapenko: This code has been redesigned.

tempesta_fw/http.c

tempesta_fw/http.c

tempesta_fw/http.c

tempesta_fw/http.c
aleksostapenko: Warning added.

tempesta_fw/http.c
aleksostapenko: This code has been redesigned.

tempesta_fw/sock_srv.c
aleksostapenko: This code has been redesigned (removed `tfw_cfg_health_lst`).

tempesta_fw/sock_srv.c
```diff
@@ -1234,6 +1332,13 @@ tfw_cfgop_server(TfwCfgSpec *cs, TfwCfgEntry *ce, TfwCfgSrvGroup *sg_cfg)
 				return -EINVAL;
 			}
 			has_weight = true;
+		} else if (!strcasecmp(key, "health")) {
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/server.h
```diff
@@ -255,6 +262,33 @@ tfw_srv_conn_need_resched(TfwSrvConn *srv_conn)
...
 
+/*
+ * Put server into alive or suspended (exluded from processing) state.
+ */
+static inline void
+tfw_srv_mark_alive(TfwServer *srv)
+{
+	if (test_bit(TFW_SRV_B_HMONITOR, (unsigned long *)&srv->hm_flags))
+		clear_bit(TFW_SRV_B_SUSPEND, (unsigned long *)&srv->hm_flags);
+}
```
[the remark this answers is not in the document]
aleksostapenko: Thank you for helpful comment, `test_bit(TFW_SRV_B_HMONITOR, (unsigned long *)&srv->hm_flags)` is not needed here at all. I left this check in `tfw_srv_suspended()` to avoid spinlock (or read/writelock) synchronization in `tfw_http_hm_control()`.

tempesta_fw/server.h
aleksostapenko: In current implementation of `set_bit/clear_bit` functions for x86, in cases of constant bit index argument - per-byte instructions `orb` and `andb` are used. So if for bit index we use immediate value less then 32 - we will not go beyond the `int` boarders. But yes, this is not very safe approach.
Corrected.

tempesta_fw/apm.c
```diff
@@ -1058,6 +1213,197 @@ tfw_apm_del_srv(TfwServer *srv)
...
+
+	rtime = (long)hm->tmt - (jiffies - READ_ONCE(hmctl->jtmstamp))
+					/ HZ;
+	stats->rtime = rtime < 0 ? 0 : rtime;
+
+	return stats;
+}
+
+bool
+tfw_apm_get_hm(const char *name, void **res_hm)
```
krizhanovsky: Why not to return `hm`? Why do we need the pointer and a boolean return value?

tempesta_fw/apm.c
```diff
@@ -1154,15 +1454,319 @@ tfw_handle_apm_stats(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+		.spec_ext	= &(TfwCfgSpecChild ) {
+			.begin_hook = tfw_apm_hm_cfg_begin,
+			.finish_hook = tfw_apm_hm_cfg_finish
+		},
+		.allow_none	= true,
+		.allow_repeat	= true,
+		.cleanup	= tfw_apm_hm_cfg_cleanup,
 	},
-	{}
+	{ 0 }
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/str.c
```diff
@@ -668,6 +669,26 @@ tfw_str_next_str_val(const TfwStr *str)
...
+
+	BUG_ON(str->len && !str->ptr);
+	TFW_STR_FOR_EACH_CHUNK(c, str, end) {
+		crc = crc32(crc, c->ptr, c->len);
+		len += c->len;//!!!
+	}
+	BUG_ON(len != str->len);
+	return crc;
+}
+EXPORT_SYMBOL(tfw_str_crc32_calc);
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/str.c
```diff
@@ -668,6 +669,26 @@ tfw_str_next_str_val(const TfwStr *str)
...
+tfw_str_crc32_calc(const TfwStr *str)
+{
+	const TfwStr *c, *end;
+	unsigned long len = 0;
+	u32 crc = 0;
+
+	BUG_ON(str->len && !str->ptr);
+	TFW_STR_FOR_EACH_CHUNK(c, str, end) {
+		crc = crc32(crc, c->ptr, c->len);
+		len += c->len;//!!!
```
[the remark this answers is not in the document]
aleksostapenko: Yes, forgot to remove this debug code. Thanks!
Corrected.

tempesta_fw/t/functional/regression/test_health_monitor.py
```diff
@@ -0,0 +1,256 @@
...
+        config = (
+            'sched_http_rules {\n'
+            'match srv_grp1 * * * ;\n'
+            '}\n'
+            'server_failover_http 404 300 10;\n'
+            'health_check h_monitor1 {\n'
+            'request "GET / HTTP/1.1\r\nHost: example.com\r\n\r\n";\n'
+            'request_url "/page.html";\n'
+            'resp_code 200;\n'
+            'resp_crc32 3456;\n'
```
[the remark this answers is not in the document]
aleksostapenko: Done. New functional test `TestHealthMonitorCRCOnly` (for separate CRC32 only verification) is added.

tempesta_fw/http.c
```diff
@@ -2879,8 +3044,10 @@ tfw_http_req_key_calc(TfwHttpReq *req)
 
 	req->hash = tfw_hash_str(&req->uri_path);
```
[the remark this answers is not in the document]

tempesta_fw/apm.c
```diff
@@ -1058,6 +1213,197 @@ tfw_apm_del_srv(TfwServer *srv)
...
+void
+tfw_apm_hm_enable_srv(TfwServer *srv, void *hmref)
+{
+	TfwApmHMCtl *hmctl;
+	unsigned long now;
+	TfwApmHM *hm = hmref;
+
+	BUG_ON(!srv->apmref);
+	BUG_ON(!hm);
+	BUG_ON(test_bit(TFW_SRV_B_HMONITOR, &srv->hm_flags));
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/apm.c
```diff
@@ -1058,6 +1213,197 @@ tfw_apm_del_srv(TfwServer *srv)
...
+
+	/* Set new health monitor for server. */
+	hmctl = &((TfwApmData *)srv->apmref)->hmctl;
+	WRITE_ONCE(hmctl->hm, hm);
+
+	/* Init server's health control fields. */
+	atomic64_set(&hmctl->rcount, 0);
+	clear_bit(TFW_SRV_B_SUSPEND, &srv->hm_flags);
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/apm.c
```diff
@@ -1058,6 +1213,197 @@ tfw_apm_del_srv(TfwServer *srv)
...
+}
+
+bool
+tfw_apm_hm_srv_alive(int status, TfwStr *body, void *apmref)
+{
+	TfwApmHM *hm = READ_ONCE(((TfwApmData *)apmref)->hmctl.hm);
+
+	BUG_ON(!hm);
+	if (hm->codes && test_bit(HTTP_CODE_BIT_NUM(status), hm->codes))
+		return true;
```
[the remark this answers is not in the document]
aleksostapenko: But in case of default `auto` monitor - it seems that we haven't any acceptable CRC32 value to examine.

tempesta_fw/apm.c
aleksostapenko: Corrected.

tempesta_fw/server.h
```diff
@@ -64,9 +65,15 @@ typedef struct {
 	atomic64_t		refcnt;
 	unsigned int		weight;
 	unsigned int		flags;
+	unsigned long		hm_flags;
```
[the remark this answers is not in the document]
aleksostapenko: Done: flags have been merged.

tempesta_fw/http.c
```diff
@@ -742,6 +742,38 @@ tfw_http_error_resp_switch(TfwHttpReq *req, int status, const char *reason)
...
 }
 
+static void
+tfw_http_hm_control(TfwHttpResp *resp)
+{
+	TfwServer *srv = (TfwServer *)resp->conn->peer;
+
+	if (!test_bit(TFW_SRV_B_HMONITOR, &srv->hm_flags))
+		return;
+
```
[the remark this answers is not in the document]
aleksostapenko: Functions `tfw_http_hm_control()` and `tfw_srv_mark_suspended()` have been redesigned in accordance with discussion in chat.

tempesta_fw/server.h
```diff
@@ -255,6 +262,31 @@ tfw_srv_conn_need_resched(TfwSrvConn *srv_conn)
...
+}
+
+/*
+ * Tell if server is suspended.
+ */
+static inline bool
+tfw_srv_suspended(TfwServer *srv)
+{
+	return test_bit(TFW_SRV_B_HMONITOR, &srv->hm_flags)
+		&& test_bit(TFW_SRV_B_SUSPEND, &srv->hm_flags);
```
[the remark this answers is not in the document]
aleksostapenko: Corrected: https://github.com/tempesta-tech/tempesta/pull/877#discussion_r168401906.

tempesta_fw/http.c
```diff
@@ -742,6 +742,54 @@ tfw_http_error_resp_switch(TfwHttpReq *req, int status, const char *reason)
...
+	if (!tfw_apm_hm_srv_limit(resp->status, srv->apmref))
+		return false;
+
+	do {
+		old_flags = cmpxchg(&srv->flags, flags,
+				    flags | TFW_SRV_F_SUSPEND);
+		if (likely(old_flags == flags)) {
+			TFW_WARN_ADDR("server has been suspended: limit"
+				      " for bad responses is exceeded",
```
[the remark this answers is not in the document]

tempesta_fw/apm.c
```diff
@@ -1058,6 +1213,194 @@ tfw_apm_del_srv(TfwServer *srv)
...
+
+	if (hm->crc32 && body->len &&
+	    tfw_str_crc32_calc(body) == hm->crc32)
+		return true;
+
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/apm.c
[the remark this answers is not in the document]
aleksostapenko: Done (the last approach).

tempesta_fw/sock_srv.c
```diff
@@ -681,7 +694,7 @@ tfw_sock_srv_grace_shutdown_srv(TfwSrvGroup *sg, TfwServer *srv)
 
 	tfw_server_get(srv);
 	__tfw_sg_del_srv(sg, srv, false);
-	srv->flags |= TFW_CFG_F_DEL;
+	set_bit(TFW_CFG_B_DEL, &srv->flags);
```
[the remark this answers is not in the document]
aleksostapenko: Corrected, but with `TFW_CFG_M_ACTION` - leave the same as didn't find acceptable interface for bit mask checking.

tempesta_fw/apm.c
```diff
@@ -372,6 +373,111 @@ tfw_stats_update(TfwPcntRanges *rng, unsigned int r_time)
...
+ *
+ * @hm		- pointer to settings for specific health monitor;
+ * @hmstats	- pointer to array of stat entries for all monitored HTTP codes;
+ * @rcount	- current count of health monitoring requests (in @hm->tmt);
+ * @jtmstamp	- time in jiffies of last @timer call (for procfs);
+ * @timer	- timer for sending health monitoring request;
+ * @rearm	- flag for gracefull stopping of @timer;
+ */
+typedef struct {
+	TfwApmHM		*hm;
```
ikoveshnikov: Seems at least `hmstats`, `hm` are server group-wide options and should be stored there. We can also use one timer per server group: send health check message to all server in group. What do you think? not the comment below.

tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -128,11 +128,12 @@ __bsearch(const unsigned long hash, TfwHashConnList *cl)
...
 
 static inline int
-__is_conn_suitable(TfwSrvConn *conn)
+__is_conn_suitable(TfwSrvConn *conn, bool hmonitor)
 {
 	return !tfw_srv_conn_restricted(conn)
 		&& !tfw_srv_conn_queue_full(conn)
-		&& tfw_srv_conn_get_if_live(conn);
+		&& tfw_srv_conn_get_if_live(conn)
+		&& (hmonitor || !tfw_srv_suspended((TfwServer *)conn->peer));
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/sched/tfw_sched_ratio.c
```diff
@@ -863,6 +864,14 @@ tfw_sched_ratio_sched_srv_conn(TfwMsg *msg, TfwServer *srv)
 	if (unlikely(!srvdesc))
 		goto done;
 
+	/*
+	 * Bypass the suspend checking if connection is needed for
+	 * helth monitoring of backend server.
+	 */
+	if (!(((TfwHttpReq *)msg)->flags & TFW_HTTP_HMONITOR)
+	    && tfw_srv_suspended(srv))
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/sock_srv.c
```diff
@@ -1978,27 +2096,50 @@ tfw_cfgop_update_sg_cfg(TfwCfgSrvGroup *sg_cfg)
...
+}
+
 static int
 tfw_cfgop_start_sg_cfg(TfwCfgSrvGroup *sg_cfg)
 {
 	int r;
-	TfwSrvGroup *sg = sg_cfg->parsed_sg;
+	void *hm = NULL;
+
+	if (sg_cfg->hm_name && !(hm = tfw_apm_get_hm(sg_cfg->hm_name)))
```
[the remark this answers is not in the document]

tempesta_fw/sock_srv.c
```diff
@@ -1956,19 +2045,48 @@ tfw_cfgop_sg_start_sched(TfwCfgSrvGroup *sg_cfg, TfwSrvGroup *sg)
...
+	list_for_each_entry(srv, &sg->srv_list, list) {
+		/*
+		 * Server was not found in new configuration, so
+		 * later (in 'tfw_cfgop_update_sg_srv_list()') it
+		 * will be stopped and removed.
+		 */
+		if (!(srv->flags & TFW_CFG_M_ACTION))
+			continue;
+
+		tfw_cfgop_update_srv_health(srv, sg_cfg->hm_name, hm);
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/apm.c
aleksostapenko: We cannot make `hmstats` group-wide - since this will break the essence of the task (server-wide HTTP health statistics and server-wide suspend/alive decisions).
As for `hm` - yes, theoretically we can use it group-wide, but for now I'd prefer leave it server-wide for three reasons:
1. Changing current implementation (originally grounded on a server-wide approach) will require significant code changes and re-testing.
2. In current approach all health monitoring functionality is concentrated in APM module and the there is no need to spreading it between other modules (e.g `sock_srv` etc.)
3. Current solution is more flexible and give more granularity in HM management if it will be needed in future, and I do not see significant advantages in group-wide approach except of saving memory on `hm` pointers for each server.

tempesta_fw/t/functional/regression/test_health_monitor.py
```diff
@@ -0,0 +1,288 @@
...
+        self.stages_processed += 1
+        return True
+
+    def get_server_path(self):
+        return '%s/%s:%s' % (self.sg_name, self.servers[0].ip,
+                             self.servers[0].port)
+
+
+class TestHealthMonitor(functional.FunctionalTest):
+    """ Test for health monitor functionality with stress option.
```
ikoveshnikov: I wouldn't say it's 'regression' test. It's definitely test for a independent new feature - health monitoring. Please move the test to it's own directory. We'll need more tests about HTTP availability in future.

Issue #838, opened 2017-09-05
GFSM revision for HTTP responses


krizhanovsky:
https://github.com/tempesta-tech/tempesta/pull/837/files#diff-0d15717c2e20453f4271befe0024c2d4R2430 does messy hack to pass request state to GFSM handler. The problem is that GFSM doesn't account requirements of HTTP response handling and this must be fixed.

aleksostapenko:
The hack referred to above is most likely the reason of crash described in #868. The problem arise not regularly - during functional `Pipeline` test:
```
[  596.963619] ------------[ cut here ]------------
[  596.964054] kernel BUG at /home/test_user/nfs_client/tempesta/tempesta_fw/http.c:1802!
[  596.964596] invalid opcode: 0000 [#1] SMP
[  596.964596] Modules linked in: tfw_sched_ratio(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) tempesta_tls(O) rpcsec_gss_krb5 nfsv4 dns_resolver nfs fscache kvm_intel kvm irqbypass ppdev crct10dif_pclmul crc32_pclmul ghash_clmulni_intel bo\
chs_drm ttm aesni_intel aes_x86_64 lrw drm_kms_helper gf128mul glue_helper ablk_helper cryptd sg joydev parport_pc evdev parport drm snd_pcm snd_timer snd soundcore serio_raw button pcspkr nfsd auth_rpcgss oid_registry nfs_acl lockd grace sunrpc ip_tables x_tables\
 autofs4 ext4 crc16 jbd2 fscrypto mbcache sr_mod cdrom sd_mod ata_generic ata_piix psmouse floppy e1000 crc32c_intel i2c_piix4 libata scsi_mod
[  596.964596] CPU: 1 PID: 4559 Comm: nginx Tainted: G           O    4.9.35+ #1
[  596.964596] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS Ubuntu-1.8.2-1ubuntu1 04/01/2014
[  596.964596] task: ffff8de7da468000 task.stack: ffffb082037ac000
[  596.964596] RIP: 0010:[<ffffffffc08e6087>]  [<ffffffffc08e6087>] tfw_http_resp_fwd+0x187/0x2d0 [tempesta_fw]
[  596.964596] RSP: 0018:ffff8de7ffd039a8  EFLAGS: 00010246
[...]

PR #884, opened 2018-01-16
Early bind of response and request

### LINKED ISSUE #871: Kernel panic for some responses
```
[   84.295653] [tempesta] Warning: Parser error: state=Resp_HttpVer input=0x2e('.3

[   84.295653] Date: Wed, 1') data_len=286 off=0
[   84.900866] [tempesta] Warning: response dropped: processing error: ::ffff:7f00:1
[   85.218876] [tempesta] Warning: Parser error: state=Resp_HttpVer input=0x72('r>nginx/1.10.3</') data_len=43 off=0
[   86.835532] BUG: unable to handle kernel paging request at ffff9cf46ba92318
```

ikoveshnikov:
fix #871. 
Also done:
- Http parser can look now inside request when parses response, that was required for some methods.
- remove the hack described in #838, but GFSM revision is still required.


tempesta_fw/http.c
```diff
@@ -1346,15 +1417,10 @@ tfw_http_conn_msg_alloc(TfwConn *conn)
...
-		TfwHttpReq *req;
-		TfwSrvConn *srv_conn = (TfwSrvConn *)conn;
-
-		spin_lock(&srv_conn->fwd_qlock);
-		req = list_first_entry_or_null(&srv_conn->fwd_queue,
-					       TfwHttpReq, fwd_list);
-		spin_unlock(&srv_conn->fwd_qlock);
-		if (req && (req->method == TFW_HTTP_METH_HEAD))
-			hm->flags |= TFW_HTTP_VOID_BODY;
+		if (unlikely(tfw_http_resp_bind_req(hm))) {
```
keshonok: While this comment does apply to this specific line of code, this is more of a general comment.

My concern is that I don't completely understand the implication of a possible break in communication between Tempesta and a back end server. The code works the way that it *removes* the paired request from the forwarding queue as soon as the response starts coming in. If the back end is unable to transmit a full HTTP response due to some reason, wouldn't it attempt to re-send the response again later? It the paired request is already removed from the forwarding queue, that will break the pairing of requests and responses.

I would say that early binding should not remove the corresponding request from the forwarding queue just yet. It's simply a linkage. When the response is parsed, only then the request should be removed.

What do you think?

tempesta_fw/http.h
```diff
@@ -453,10 +455,14 @@ typedef struct {
 	time_t			date;
 	time_t			last_modified;
 	unsigned long		jrxtstamp;
+	TfwHttpMsg		*req;
```
keshonok: An alternative is moving this to `TfwHttpMsg{}/TFW_HTTP_MSG_COMMON`, and also move `TfwHttpReq{}->resp` there. Then, there it could be either a union of `{ req, resp }`, or simply a single member, say, `->pair`.

tempesta_fw/http.h
```diff
@@ -453,10 +455,14 @@ typedef struct {
...
 } TfwHttpResp;
 
 #define TFW_HTTP_RESP_STR_START(r)	__MSG_STR_START(r)
 #define TFW_HTTP_RESP_STR_END(r)	((&(r)->body) + 1)
+#define TFW_HTTP_RESP_BOUND_REQ(r)	(TfwHttpReq *)(((TfwHttpResp *)r)->req)
+#define TFW_HTTP_RESP_BIND_REQ(resp, req)	\
+	(((TfwHttpResp *)resp)->req = (TfwHttpMsg *)req)
```
keshonok: First, it appears that there's no need in the outer braces. Secondly, though - it is macro really needed when it's only used once? And then, I am not sure this even makes a good case for a macro...

tempesta_fw/http.h
```diff
@@ -453,10 +455,14 @@ typedef struct {
 	time_t			date;
 	time_t			last_modified;
 	unsigned long		jrxtstamp;
+	TfwHttpMsg		*req;
 } TfwHttpResp;
 
 #define TFW_HTTP_RESP_STR_START(r)	__MSG_STR_START(r)
 #define TFW_HTTP_RESP_STR_END(r)	((&(r)->body) + 1)
+#define TFW_HTTP_RESP_BOUND_REQ(r)	(TfwHttpReq *)(((TfwHttpResp *)r)->req)
```
keshonok: This appears to be a cut-n-paste bug. The macro is the same as `TFW_HTTP_REQ_BOUND_RESP()` above. One of these two should be different.

On another thought, I wouldn't make any of these a macro. The statement short enough, and it is totally clear so  there's no need to explain it through a different name (the name of a macro).

tempesta_fw/http.c
```diff
@@ -487,7 +487,7 @@ tfw_http_send_resp(TfwHttpReq *req, resp_code_t code)
 		msg.len += crlf->len - crlf_len;
 	}
 
-	if (!(hmresp = tfw_http_msg_alloc_err_resp()))
+	if (!(hmresp = tfw_http_msg_alloc_err_resp(req)))
```
keshonok: Isn't `resp->req` irrelevant at this stage, other that for a `BUG_ON()`?

At this stage the paired request is definitely not in the forwarding queue. From there, the response and the paired request go together through the code function call chain where necessary (as function arguments). This can be viewed as loose binding.

With that in mind, perhaps, only one of the two ways of binding should be used once the response is parsed. One is direct binding, where the link to the paired request is kept in the response's structure. This comes for free now. The other is loose binding where the paired request comes as an argument in function call chain in the code.

I don't know how easy or difficult would be to do that. But using both of these ways of binding seems to be a little confusing now (and, perhaps, excessive).

tempesta_fw/http.c
ikoveshnikov: Yes, I've missed that scenario. 

> I would say that early binding should not remove the corresponding request from the forwarding queue just yet. It's simply a linkage. When the response is parsed, only then the request should be removed.

I'm agree here. I'll fix that.

tempesta_fw/http.h
ikoveshnikov: I see the point here, this will also make unneeded all that `BOUND_REQ` macro.

tempesta_fw/http.c
ikoveshnikov: Idea was to use `resp->req` linkage as a primary linkage. Response just can't exist if there is no request for it.  But the previous loose binding was kept untouched in the patch. Now i see how it's confusing, so I plan to keep only first one.

ikoveshnikov:
@keshonok Thank for your review!

tempesta_fw/classifier/frang.c
```diff
@@ -956,8 +956,9 @@ frang_resp_code_range(const int n)
 static int
 frang_resp_handler(void *obj, struct sk_buff *skb, unsigned int off)
 {
-	TfwHttpReq *req = (TfwHttpReq *)obj;
-	TfwHttpResp *resp = (TfwHttpResp *)req->resp;
+	TfwConn *conn = (TfwConn *)obj;
+	TfwHttpResp *resp = container_of(conn->msg, TfwHttpResp, msg);
+	TfwHttpReq *req = TFW_HTTP_RESP_BOUND_REQ(resp);
```
krizhanovsky: We can fix #838 by always keeping requests in a pair with responses and vise versa, as it's done here, or pass request and response along with L3/L4 data (skb, off) to GFSM callbacks. The design problem with GFSM is that it operates with skb and off, i.e. on L3/L4 layers only, while Frang requires HTTP requests and responses. So we definitelly need `TfwFsmData`. However, now probably we don't need both the `req` and `resp` in it and can replace it with the union. I think it's easier for a GFSM hook to get both the `req` and `resp` and check them for NULL rather than try to guess what's passed and get a pair. Opinions?

tempesta_fw/http.c
krizhanovsky: Yes, please address the comment, e.g. `tfw_http_resp_fwd()` doesn't need to accept `req` as an argument at the below.

By the way, @keshonok thank you for the meaningful review.

tempesta_fw/http_parser.c
```diff
@@ -4109,6 +4109,22 @@ tfw_http_init_parser_resp(TfwHttpResp *resp)
...
+ * Adjust parser for response according to it's request.
+ */
+static void
+tfw_http_adj_parser_resp(TfwHttpResp *resp)
+{
+	TfwHttpReq *req = TFW_HTTP_RESP_BOUND_REQ(resp);
+
+	/* Can happen only in unittests. */
+	if (unlikely(!req))
+		return;
```
krizhanovsky: We never adjust production code to satisfy tests. Tests are for production code and production code isn't for test. Please remove the check and adjust the test to handle the case.

tempesta_fw/http.c
```diff
@@ -1494,19 +1520,9 @@ tfw_http_conn_cli_drop(TfwCliConn *cli_conn)
...
 	 */
 	spin_lock(&cli_conn->seq_qlock);
-	list_for_each_entry_safe(req, tmp, seq_queue, msg.seq_list) {
-		if (req->resp)
-			list_move_tail(&req->msg.seq_list, &zap_queue);
-		else
-			list_del_init(&req->msg.seq_list);
-	}
+	list_for_each_entry_safe(req, tmp, seq_queue, msg.seq_list)
+		list_del_init(&req->msg.seq_list);
```
keshonok: There's the description to this function which describes the logic of actions that were removed. The description should be modified to conform to the current logic.

tempesta_fw/http.c
```diff
@@ -1866,11 +1884,14 @@ tfw_http_resp_fwd(TfwHttpReq *req, TfwHttpResp *resp)
...
 	}
 	BUG_ON(list_empty(&req->msg.seq_list));
-	req->resp = (TfwHttpMsg *)resp;
+	resp->flags |= TFW_HTTP_RESP_READY;
+	WARN_ON((TfwHttpReq *)resp->req != req);
 	/* Move consecutive requests with @req->resp to @ret_queue. */
 	list_for_each_entry(req, seq_queue, msg.seq_list) {
-		if (req->resp == NULL)
+		if (req->resp == NULL
+		    || !(req->resp->flags & TFW_HTTP_RESP_READY)) {
```
keshonok: It looks to me that either you don't need curly brackets at all as this is a single statement body, or the opening bracket should be on the next line. :-) I think the latter is better. The current variant is no better than no brackets at all.

tempesta_fw/http.c
```diff
@@ -1328,6 +1328,38 @@ tfw_http_req_destruct(void *msg)
...
+tfw_http_popreq(TfwHttpMsg *hmresp)
+{
+	TfwHttpReq *req;
+	TfwSrvConn *srv_conn = (TfwSrvConn *)hmresp->conn;
+	struct list_head *fwd_queue = &srv_conn->fwd_queue;
+
+	spin_lock(&srv_conn->fwd_qlock);
+	req = list_first_entry_or_null(fwd_queue, TfwHttpReq, fwd_list);
+	while (req && req->pair)
+		req = list_next_entry(req, fwd_list);
```
keshonok: I feel a little uneasy about this search loop. I think the code should be made more robust. The concerns are:

1. It disregards `msg_sent` which serves the role of a stop guard for the list of sent messages. If there's a bug somewhere (req/resp get unsynchronized), this loop may cross it and go over unsent messages.
2. It disregards the end of the queue as well. It may get the queue head which is not a `TfwHttpReq{}` structure.
3. There's no case where `req` is not found (see [2] above). The latest `req` that was looked up will be paired with the response.

All of these are situations that should not happen in the course of normal work. That means, one or more of them will definitely happen, sooner or later. :-)

tempesta_fw/cache.c
```diff
@@ -1614,24 +1616,26 @@ cache_req_process_node(TfwHttpReq *req, tfw_http_cache_cb_t action)
...
 static void
-tfw_cache_do_action(TfwHttpReq *req, TfwHttpResp *resp,
-		    tfw_http_cache_cb_t action)
+tfw_cache_do_action(TfwHttpMsg *msg, tfw_http_cache_cb_t action)
 {
-	if (resp) {
-		tfw_cache_add(resp, req, action);
+	TfwHttpReq *req;
+
+	if (TFW_CONN_TYPE(msg->conn) == Conn_Srv) {
```
keshonok: This is incorrect. The check should be
```C
        if (TFW_CONN_TYPE(msg->conn) & Conn_Srv) {
        }
```

tempesta_fw/cache.c
```diff
@@ -1643,13 +1647,19 @@ tfw_cache_ipi(struct irq_work *work)
...
+tfw_cache_process(TfwHttpMsg *msg, tfw_http_cache_cb_t action)
 {
 	int cpu;
 	unsigned long key;
 	TfwWorkTasklet *ct;
 	TfwCWork cw;
+	TfwHttpResp *resp = NULL;
+	TfwHttpReq *req = (TfwHttpReq *)msg;
+
+	if (TFW_CONN_TYPE(msg->conn) == Conn_Srv) {
```
keshonok: This is incorrect. The check should be
```C
        if (TFW_CONN_TYPE(msg->conn) & Conn_Srv) {
        }
```

tempesta_fw/http.c
```diff
@@ -1494,19 +1520,9 @@ tfw_http_conn_cli_drop(TfwCliConn *cli_conn)
...
-	}
+	list_for_each_entry_safe(req, tmp, seq_queue, msg.seq_list)
+		list_del_init(&req->msg.seq_list);
 	spin_unlock(&cli_conn->seq_qlock);
-
-	list_for_each_entry_safe(req, tmp, &zap_queue, msg.seq_list) {
-		BUG_ON(!list_empty_careful(&req->fwd_list));
-		BUG_ON(!list_empty_careful(&req->nip_list));
-		__tfw_http_resp_pair_free(req);
-	}
```
krizhanovsky: A client sends us several requests and the client connection immediately terminated. All of the requests are in `fwd_queue`'s and are being sent to servers, reforwarded and so on until all of the receive responses. Then `tfw_http_resp_fwd()` finds empty `seq_queue` of the client connection and calls `ss_close_sync()` for each of the requests. All in all, we do a lot of unnecessary work.

It's OK to leave the dead requests in `fwd_queue`'s to minimize lock contention, but it's easy and cheap to:
1. set some flag for the reuests to make them evicted from `fwd_queue` right they're going to be send to a server or rescheduled;
2. we don't need to call `ss_close_sync()` in `tfw_http_resp_fwd()` since all `connection_error` and `connection_drop` callbacks follow `ss_do_close()` call in sock.c, so `tfw_http_conn_cli_drop()` is always called on closed socket.
3. `tfw_http_conn_cli_drop()` should call `tfw_connection_put()` just like `tfw_sock_srv_connect_drop()` and `tfw_sock_srv_connect_failover()` do. When `tfw_http_resp_fwd()` calls `tfw_http_conn_msg_free()` for the last request referencing the client connection, the connection is freed.

tempesta_fw/http.c
```diff
@@ -1866,10 +1884,11 @@ tfw_http_resp_fwd(TfwHttpReq *req, TfwHttpResp *resp)
 		return;
 	}
 	BUG_ON(list_empty(&req->msg.seq_list));
-	req->resp = (TfwHttpMsg *)resp;
+	resp->flags |= TFW_HTTP_RESP_READY;
+	WARN_ON((TfwHttpReq *)resp->req != req);
```
aleksostapenko: Looks like redundant WARN_ON here (exactly the same assignment is made several lines above). Or maybe `WARN_ON(req->resp != resp);` should be here.

tempesta_fw/http_msg.c
```diff
@@ -845,15 +846,16 @@ EXPORT_SYMBOL(tfw_http_msg_free);
...
+tfw_http_msg_alloc_err_resp(TfwHttpReq *req)
 {
-	TfwHttpMsg *hm;
+	TfwHttpResp *resp;
 
-	if (!(hm = (TfwHttpMsg *)tfw_pool_new(TfwHttpResp, TFW_POOL_ZERO)))
+	if (!(resp = (TfwHttpResp *)tfw_pool_new(TfwHttpResp, TFW_POOL_ZERO)))
 		return NULL;
-	ss_skb_queue_head_init(&hm->msg.skb_list);
+	tfw_http_msg_pair((TfwHttpMsg *)req, (TfwHttpMsg *)resp);
```
aleksostapenko: Not sure that it is worth to combine response allocation and request pairing: in https://github.com/tempesta-tech/tempesta/pull/877 interface `tfw_http_msg_alloc_err_resp()` is replaced by light version of general message allocation interface. So, as a result - pairing functionality will still have to be taken out separately (or to be added to general `__tfw_http_msg_alloc` interface, which is undesirable).

ikoveshnikov: The reason to combine allocation and pairing is to force pairing of request and responses. In the patch response without paired request is a bug. To avoid creating responses without paired request a new interface is introduced.

tempesta_fw/http.c
ikoveshnikov: > we don't need to call ss_close_sync() in tfw_http_resp_fwd() since all connection_error and connection_drop callbacks follow ss_do_close() call in sock.c, so tfw_http_conn_cli_drop() is always called on closed socket.

I'm agree with the comment: https://github.com/tempesta-tech/tempesta/blob/f287f9f910131f576d66b9e3f3884537e92eff04/tempesta_fw/http.c#L2089-L2091
Need fo close client connection if it's open and `seq_queue` is empty.

> tfw_http_conn_cli_drop() should call tfw_connection_put() just like tfw_sock_srv_connect_drop() and tfw_sock_srv_connect_failover() do. 

`tfw_http_conn_cli_drop()` is a `TfwConnHooks`, while both
 `tfw_sock_srv_connect_drop()` and `tfw_sock_srv_connect_failover()` are `SsHooks`. Not  `tfw_http_conn_cli_drop()`  but `tfw_sock_clnt_drop()` and `tfw_sock_clnt_error()` must put the client connection. It's already done this way, so nothing to do at this point.

tempesta_fw/http_msg.c
```diff
@@ -945,14 +971,21 @@ __tfw_http_msg_alloc(int type, bool full)
...
+	BUILD_BUG_ON(FIELD_SIZEOF(TfwHttpMsg, flags) * BITS_PER_BYTE
+		     < _TFW_HTTP_FLAGS_NUM);
 
 	if (full) {
 		hm->h_tbl = (TfwHttpHdrTbl *)tfw_pool_alloc(hm->pool,
 							    TFW_HHTBL_SZ(1));
 		if (unlikely(!hm->h_tbl)) {
-			TFW_WARN("Insufficient memory to create message\n");
+			TFW_WARN("Insufficient memory to create %s message\n",
+				 ((type & Conn_Clnt) ? "request" : "response"));
```
krizhanovsky: Not  big deal, but usually it helpful to have different messages: having one in dmesg you can quickly understang which code exactly produced the message. Also, as it's in your code, good to have as much of the message in the same line for easier `grep`'ing the code.

tempesta_fw/http.c
```diff
@@ -941,12 +942,31 @@ tfw_http_req_zap_error(struct list_head *eq)
...
+ */
+static inline bool
+tfw_http_req_evict_dropped(TfwSrvConn *srv_conn, TfwHttpReq *req,
+			   struct list_head *eq)
+{
+	if (unlikely(req->flags & TFW_HTTP_F_REQ_DROP)) {
+		TFW_DBG2("%s: Eviction: req=[%p] client disconnected\n",
+			 __func__, req);
+		tfw_http_req_err(srv_conn, req, eq, 403,
+				 "request evicted: client disconnected");
```
krizhanovsky: Why do we need to pass the request to the error queue? `tfw_http_req_zap_error()` just does nothing on `req->flags & TFW_HTTP_F_REQ_DROP`, i.e. the `req` remains unfreed and we have a memory leakage. Moreover, we don't need to bother with the error queue - there is nothing to do with `TFW_HTTP_F_REQ_DROP` requests and it seems we can just free them right now.

PR #889, opened 2018-01-21
We know what we do with sched_data, so disable senseless warnings

krizhanovsky:


tempesta_fw/sched/tfw_sched_hash.c
```diff
@@ -244,14 +244,13 @@ static void
 tfw_sched_hash_del_grp(TfwSrvGroup *sg)
 {
 	TfwServer *srv;
-	TfwHashConnList *cl = rcu_dereference(sg->sched_data);
+	TfwHashConnList *cl = rcu_dereference_check(sg->sched_data, 1);
 
+	rcu_assign_pointer(sg->sched_data, NULL);
```
ikoveshnikov: Resetting of group sched data pointer is done afrer resetting  per-server pointer because both type of pointers points to one memory chunk. That chunk is allocated for group sched data pointer while per-server pointers point somewhere inside that chunk. Logically per-server sched data pointers must be reset before resetting group sched data pointer. 

Just a remark on original order. No real changes here since all pointers are protected by rcu.

Issue #853, opened 2017-10-31
http_match for Referer header


krizhanovsky:
Several DDoS attacks can be mitigated by filtering `Referer` HTTP header, so need to add the header to list of Special headers and make sure that it's quickly processed by `http_match`.

Update Wiki for DDoS use cases. Also please create a task for appropriate test for the next milestone.

krizhanovsky:
@aleksostapenko you didn't update neither https://github.com/tempesta-tech/tempesta/wiki/Scheduling-and-Load-Balancing#http-scheduler-for-server-groups nor https://github.com/tempesta-tech/tempesta/wiki/DDoS-mitigation#iframes-on-a-busy-site . Close the issue after the Wiki updates.

aleksostapenko:
Wiki updated.

PR #891, opened 2018-01-22
Add http scheduler matching for Referer header. 

aleksostapenko:


tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2312,116 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
...
+static int
+__req_parse_referer(TfwHttpMsg *hm, unsigned char *data, size_t len)
+{
+	int r = CSTR_NEQ;
+	unsigned char *off_ptr = data;
+	__FSM_DECLARE_VARS(hm);
+
+	__FSM_START(parser->_i_st) {
+
+	__FSM_STATE(Req_Uri) {
```
keshonok: I do understand the original intent, but I also do think that it's bad practice to reuse the top-level states as interior states. As the code evolves, this may lead to very difficult bugs if, say, the values for top-level states are changed in a certain way. I am not saying that it will happen, perhaps not as this is still a small code base, but it's best to exclude the very possibility of a bug like this.

I believe it's best to introduce **interior** states for this function.

tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2312,116 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		if (likely(c == '/')) {
+			__FSM_I_MOVE_offset_fn(Req_UriAbsPath, 1, off_ptr);
+		}
```
keshonok: Unnecessary curly braces.

tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2312,116 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
...
+			   && C4_INT_LCM(p, 'h', 't', 't', 'p')
+			   && *(p + 4) == ':'
+			   && *(p + 5) == '/'
+			   && *(p + 6) == '/'))
+		{
+			__FSM_I_MOVE_nofixup_n(Req_UriAuthorityStart, 7);
+		}
+		if (likely(TFW_LC(c) == 'h')) {
+			__FSM_I_MOVE_nofixup_n(Req_UriSchH, 1);
+		}
```
keshonok: Unnecessary curly braces.

tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2312,116 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		/* Skip URI scheme 'http://' token. */
+		off_ptr = p;
+		if (likely(isalnum(c) || c == '.' || c == '-')) {
+			__FSM_I_MOVE_offset_fn(Req_UriAuthority, 1, off_ptr);
+		} else if (likely(c == '/')) {
+			__FSM_I_MOVE_offset_fn(Req_UriAbsPath, 1, off_ptr);
+		} else if (c == '[') {
+			__FSM_I_MOVE_offset_fn(Req_UriAuthorityIPv6, 1,
+					       off_ptr);
+		}
```
keshonok: Unnecessary curly braces.

tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2312,116 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		return CSTR_NEQ;
+	}
+
+	__FSM_STATE(Req_UriAuthority) {
+		if (likely(isalnum(c) || c == '.' || c == '-')) {
+			__FSM_I_MOVE_offset_fn(Req_UriAuthority, 1, off_ptr);
+		}
```
keshonok: Unnecessary curly braces.

tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2312,116 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
...
+	}
+
+	__FSM_STATE(Req_UriAuthorityIPv6) {
+		if (likely(isxdigit(c) || c == ':')) {
+			__FSM_I_MOVE_offset_fn(Req_UriAuthorityIPv6, 1,
+					       off_ptr);
+		} else if(c == ']') {
+			__FSM_I_MOVE_offset_fn(Req_UriAuthorityEnd, 1,
+					       off_ptr);
+		}
```
keshonok: Unnecessary curly braces.

tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2312,116 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		return CSTR_NEQ;
+	}
+
+	__FSM_STATE(Req_UriAuthorityEnd) {
+		if (likely(c == '/')) {
+			__FSM_I_MOVE_offset_fn(Req_UriAbsPath, 1, off_ptr);
+		} else if (c == ':') {
+			__FSM_I_MOVE_offset_fn(Req_UriPort, 1, off_ptr);
+		}
```
keshonok: Unnecessary curly braces.

tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2312,116 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		__FSM_JMP(I_EoL);
+	}
+
+	__FSM_STATE(Req_UriPort) {
+		if (likely(isdigit(c))) {
+			__FSM_I_MOVE_offset_fn(Req_UriPort, 1, off_ptr);
+		} else if (likely(c == '/')) {
+			__FSM_I_MOVE_offset_fn(Req_UriAbsPath, 1, off_ptr);
+		}
```
keshonok: Unnecessary curly braces.

tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2312,116 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		__FSM_I_MATCH_MOVE_offset(uri, Req_UriAbsPath, off_ptr);
+		p += __fsm_sz;
+		__FSM_JMP(I_EoL);
+	}
+
+	__FSM_STATE(I_EoL) {
+		__msg_hdr_chunk_fixup(off_ptr, p - off_ptr);
+		if (IS_CRLFWS(c)) {
+			return __data_off(p);
+		}
```
keshonok: Unnecessary curly braces.

tempesta_fw/http_parser.c
```diff
@@ -297,19 +339,30 @@ __FSM_STATE(st) {							\
 	TFW_PARSER_BLOCK(st);						\
 }
 
+#define __FSM_I_TX_COND_nofixup(st, condition, st_next)			\
+__FSM_STATE(st) {							\
+	if (likely(condition))						\
+		__FSM_I_MOVE_nofixup_n(st_next, 1);			\
+	return CSTR_NEQ;						\
+}
+
```
keshonok: I am not sure this is the right place for this macro, as well as for `__FSM_I_TX_nofixup()` and `__FSM_I_TX_LC_nofixup()` below. All interior macros are grouped together in a section of the source file. I would say that these macros should be defined there as well.

tempesta_fw/http_parser.c
```diff
@@ -3278,6 +3449,17 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len)
 				__FSM_MOVE_n(RGen_OWS, 7);
 			}
 			__FSM_MOVE(Req_HdrP);
+		case 'r':
+			if (likely(__data_available(p, 8)
+				   && C4_INT_LCM(p + 1, 'e', 'f', 'e', 'r')
+				   && TFW_LC(*(p + 5)) == 'e'
+				   && TFW_LC(*(p + 6)) == 'r'
+				   && *(p + 1) == ':'))
```
keshonok: Perhaps, this line should look like this? Is this code tested at all?
```C
&& *(p + 7) == ':'))
```

tempesta_fw/http_parser.c
```diff
@@ -297,19 +339,30 @@ __FSM_STATE(st) {							\
 	TFW_PARSER_BLOCK(st);						\
 }
 
+#define __FSM_I_TX_COND_nofixup(st, condition, st_next)			\
+__FSM_STATE(st) {							\
+	if (likely(condition))						\
+		__FSM_I_MOVE_nofixup_n(st_next, 1);			\
```
keshonok: Analogous to `__FSM_MOVE_nofixup()`, perhaps it makes sense to introduce `__FSM_I_MOVE_nofixup()` as well?

tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2312,116 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		if (likely(__data_available(p, 7)
+			   && C4_INT_LCM(p, 'h', 't', 't', 'p')
+			   && *(p + 4) == ':'
+			   && *(p + 5) == '/'
+			   && *(p + 6) == '/'))
+		{
+			__FSM_I_MOVE_nofixup_n(Req_UriAuthorityStart, 7);
+		}
+		if (likely(TFW_LC(c) == 'h')) {
+			__FSM_I_MOVE_nofixup_n(Req_UriSchH, 1);
```
keshonok: Analogous to `__FSM_MOVE_nofixup()`, perhaps it makes sense to introduce `__FSM_I_MOVE_nofixup()` as well?

tempesta_fw/sched/tfw_sched_http.c
```diff
@@ -213,6 +214,7 @@ tfw_sched_http_cfg_arg_tbl[_TFW_HTTP_MATCH_F_COUNT] = {
 	[TFW_HTTP_MATCH_F_WILDCARD]	= TFW_HTTP_MATCH_A_WILDCARD,
 	[TFW_HTTP_MATCH_F_HDR_CONN]	= TFW_HTTP_MATCH_A_STR,
 	[TFW_HTTP_MATCH_F_HDR_HOST]	= TFW_HTTP_MATCH_A_STR,
+	[TFW_HTTP_MATCH_F_HDR_REFERER]	= TFW_HTTP_MATCH_A_STR,
```
keshonok: Is there a description anywhere what exactly is matched? What exactly is given as a string to match? I didn't find that information in the description of the issue #853, or in the comments here.

tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2312,116 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
...
+ * simplifications according to RFC 7231 chapter 5.5.2.
+ */
+static int
+__req_parse_referer(TfwHttpMsg *hm, unsigned char *data, size_t len)
+{
+	int r = CSTR_NEQ;
+	unsigned char *off_ptr = data;
+	__FSM_DECLARE_VARS(hm);
+
+	__FSM_START(parser->_i_st) {
```
ikoveshnikov: Is it required to fully parse `Referer` header? We parse it for frang and ddos mitigation. How the attack work: attacker places  link to the victim on popular web site in such a way that every visitor send request to victim. It's expected that `referer` header is constructed correctly, just need to check that value is a part of URI alphabet. 

More important is how we are going to use the header in frang. It's seems that string match is the only way here. From this point of view, it's may be sufficient to start a new chunk at the beginning of uri and mark it with `TFW_STR_VALUE` flag.

comment from @krizhanovsky  is required here.

tempesta_fw/http_parser.c
```diff
@@ -282,6 +282,48 @@ do {									\
...
+	}								\
+	goto to;							\
+} while (0)
+
+/*
+ * Macros below is like __FSM_MOVE_nf(), but for interior
+ * states and for creation chunk's @ptr not from beginning @data
+ * but from some offset.
+ */
+#define __FSM_I_MOVE_offset_fn(to, n, off_ptr)				\
```
keshonok: The name of the macro doesn't really follow the naming scheme that already exists in the source file. There are macros that end with `_nf()`, and those mean that the move forward is for `n` bytes, and the work is on the field given as an argument.

With that in mind, this `_fn()` ending looks confusing.

tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2312,116 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
 	return r;
 }
 
+/**
+ * Function for parsing 'Referer' header value. This is a partial
+ * copy from parsing procedure for request start line URI with some 
+ * simplifications according to RFC 7231 chapter 5.5.2.
+ */
+static int
+__req_parse_referer(TfwHttpMsg *hm, unsigned char *data, size_t len)
```
keshonok: It appears that only part of the contents of the 'Referer: ' header field is stored in the TfwStr{} string for this header field. I believe this is the first header field parsed this way. This should be commented and reasons explained.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -1498,6 +1498,63 @@ TEST(http_parser, if_none_match)
...
+		ht = req->h_tbl;
+		tfw_http_msg_clnthdr_val(&ht->tbl[TFW_HTTP_HDR_REFERER],
+					 TFW_HTTP_HDR_REFERER,
+					 &h_referer);
+		EXPECT_TRUE(tfw_str_eq_cstr(&h_referer, s_referer2,
+					    strlen(s_referer2), 0));
+	}
+
+	FOR_REQ("GET /foo HTTP/1.1\r\n"
+		"Referer: http://[2001:0db8:11a3:09d7:1f34:8a2e:07a0:765d]:8080"
```
ikoveshnikov: Try to add spaces before and after referrer link. It seems that the header  is not parsed correctly if any.

tempesta_fw/http_parser.c
krizhanovsky: Yes, there were several discussions in hacker forums (like antichat) how to disable Rerefer header for iFrame attacks and basically there is no good solution since Referer is the fundamental security header and browsers don't allow to mangle it. So there is no much sense to introduce the parsing overhead for the header which is safe for iFrame attack and hasn't much sense to be verified for other attacks.

tempesta_fw/http_parser.c
krizhanovsky: I appreciate the effort for strong input validation, this is what we generally do in Tempesta FW. But this particular header is introduced to the set of the special headers just to block particular and very narrow set of attacks which can not mangle the header value, so no need to introduce more parser complexity to validate the header values. So please treat the header value just as an opaque data, with alphabet validation only as we do this for other headers.

tempesta_fw/http_parser.c
aleksostapenko: @keshonok Thanks for review.
This code has been redesigned according to https://github.com/tempesta-tech/tempesta/pull/891#discussion_r165703517

tempesta_fw/http_parser.c
aleksostapenko: Corrected.

tempesta_fw/http_parser.c
aleksostapenko: Done.

tempesta_fw/t/unit/test_http_parser.c
ikoveshnikov: Comment is still not addressed, try the patch below. The test fails with it:
```
[  203.687589] tfw_test: TEST_RUN(http_parser, referer)
[  203.689449] tfw_test: === request: [GET /foo HTTP/1.1
[  203.693618] tfw_test: /home/user/tempesta/tempesta/tempesta_fw/t/unit/test_http_parser.c:1518: error: 
[  203.696479] tfw_test: FAIL:
[  203.697361] tfw_test:   test__http_parser__referer()
[  203.699459] tfw_test:   can't parse request (code=-2):
```
The patch:
```diff
diff --git a/tempesta_fw/t/unit/test_http_parser.c b/tempesta_fw/t/unit/test_http_parser.c
index 17d1ce1..c3c27c4 100644
--- a/tempesta_fw/t/unit/test_http_parser.c
+++ b/tempesta_fw/t/unit/test_http_parser.c
@@ -1504,8 +1504,8 @@ TEST(http_parser, referer)
        TfwStr h_referer;
 
        const char *s_referer1 =
-               "http://tempesta-tech.com:8080"
-               "/cgi-bin/show.pl?entry=tempesta";
+               "  http://tempesta-tech.com:8080"
+               "/cgi-bin/show.pl?entry=tempesta  ";
        const char *s_referer2 =
                "/cgi-bin/show.pl?entry=tempesta";
        const char *s_referer3 =
@@ -1513,8 +1513,8 @@ TEST(http_parser, referer)
                ":8080/cgi-bin/show.pl?entry=tempesta";
 
        FOR_REQ("GET /foo HTTP/1.1\r\n"
-               "Referer: http://tempesta-tech.com:8080"
-               "/cgi-bin/show.pl?entry=tempesta\r\n"
+               "Referer:   http://tempesta-tech.com:8080"
+               "/cgi-bin/show.pl?entry=tempesta  \r\n"
                "\r\n")
        {
                ht = req->h_tbl;
```

aleksostapenko: Corrected.

tempesta_fw/http_parser.c
```diff
@@ -2251,6 +2261,37 @@ __req_parse_host(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		if (IS_CRLF(*(p + __fsm_sz)))
+			return __data_off(p + __fsm_sz);
+		return CSTR_NEQ;
+	}
+	__FSM_STATE(Req_I_EoT) {
+		if (IS_WS(c))
+			__FSM_I_MOVE(Req_I_EoT);
+		if (IS_CRLF(c)) {
+			return __data_off(p);
+		}
```
ikoveshnikov: Extra curly braces

Issue #893, opened 2018-01-24
[CFG] Repeated parsing of invalid configuration may cause dead loop


ikoveshnikov:
Happens on multiple starts or restarts when at least one server is defined in configuration and successfully processed.

E.g. Start or restart TempestaFW with configuration below will lead to dead loop on back end servers cleanup.
```
$ cat etc/tempesta_fw.conf 
sched ratio;
server 192.168.122.1:8080;

location prefix "/" {
jsnvkjsfnvkjf;
}
```

Multiple starts:
```
[  137.280176] [tdb] Start Tempesta DB
[  137.299532] [tempesta] Initializing Tempesta FW kernel module...
[  137.313294] [tempesta] Registering new scheduler: hash
[  137.319645] [tempesta] Registering new scheduler: http
[  137.328238] [tempesta] Registering new scheduler: ratio
[  137.332705] [tempesta] Prepearing for the configuration processing.
[  137.334386] [tempesta] ERROR: don't know how to handle: jsnvkjsfnvkjf
[  137.336079] [tempesta] ERROR: configuration parsing error:
[  137.336079]    5: jsnvkjsfnvkjf;
[  137.336079]       ^^^^^^^^^^^^^^
[  142.527026] [tempesta] Prepearing for the configuration processing.
[  142.528117] [tempesta] ERROR: location: Duplicate entry: 'location prefix /'
[  142.529515] [tempesta] ERROR: configuration parsing error:
[  142.529515]    4: location prefix "/" {
[  142.529515]       ^^^^^^^^^^^^^^^^^^^^^
[  146.876336] [tempesta] Prepearing for the configuration processing.
[  146.877553] [tempesta] ERROR: location: Duplicate entry: 'location prefix /'
[  146.879882] [tempesta] ERROR: configuration parsing error:
[  146.879882]    4: location prefix "/" {
[  146.879882]       ^^^^^^^^^^^^^^^^^^^^^
[  151.944005] [tempesta] Warning: pending for server callbacks to complete for 5s
[  156.948045] [tempesta] Warning: pending for server callbacks to complete for 5s
[  161.952042] [tempesta] Warning: pending for server callbacks to complete for 5s
[  166.956048] [tempesta] Warning: pending for server callbacks to complete for 5s
[  171.960006] [tempesta] Warning: pending for server callbacks to complete for 5s
[  176.964006] [tempesta] Warning: pending for server callbacks to complete for 5s
[...]

PR #894, opened 2018-01-24
Fix hang on configuration parse error

ikoveshnikov:
Fix first part of #893

### DIFF
--- a/tempesta_fw/cfg.c
+++ b/tempesta_fw/cfg.c
@@ -703,9 +703,7 @@ parse_cfg_entry(TfwCfgParserState *ps)
 	}
 
 	FSM_STATE(PS_EXIT) {
-		/* Cleanup entry on error */
-		if (ps->err)
-			entry_reset(&ps->e);
+		/* Cleanup of entry is done in tfw_cfg_parse_mods() */
 		TFW_DBG3("pfsm: exit\n");
 	}
 }
@@ -1376,7 +1374,7 @@ print_parse_error(const TfwCfgParserState *ps)
 	 * by @spec_finish_handling() function.
 	 */
 	if (!ps->e.line) {
-		TFW_ERR_NL("configuration parsing error");
+		TFW_ERR_NL("configuration parsing error\n");
 		return;
 	}
 
@@ -1426,7 +1424,7 @@ tfw_cfg_parse_mods(const char *cfg_text, struct list_head *mod_list)
 	do {
 		parse_cfg_entry(&ps);
 		if (ps.err) {
-			TFW_ERR("syntax error\n");
+			TFW_ERR_NL("syntax error\n");
 			goto err;
 		}
 		if (!ps.e.name)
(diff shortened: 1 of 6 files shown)


tempesta_fw/sock_srv.c
```diff
@@ -1791,9 +1791,30 @@ tfw_cfgop_grace_time(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+	 * parsing has failed and no real 'srv_group' directives hve met.
+	 */
+	srv_mod = tfw_sock_srv_mod;
+	INIT_LIST_HEAD(&srv_mod.list);
+	INIT_LIST_HEAD(&mod_list);
+	list_add(&srv_mod.list, &mod_list);
+	cfg_spec = tfw_cfg_spec_find(srv_mod.specs, "srv_group");
+	BUG_ON(!cfg_spec);
+	cfg_spec->__called_now = true;
+	cfg_spec->__called_cfg = true;
```
keshonok: Frankly, I believe that these are supposed to be flags that are *internal* to `cfg.c`. If this is the only way to make things work correctly in this case (a necessary evil), then I think it's best to wrap setting of these flags into a function with a meaningful name specifically for a case like this.

tempesta_fw/cfg.c
```diff
@@ -747,27 +747,51 @@ spec_start_handling(TfwCfgSpec specs[])
...
+				   "doesn't allow reconfiguration!\n",
+				   spec->name, parent->name);
+		}
 	}
 }
 
+static void
+spec_start_handling(TfwCfgSpec specs[])
+{
+	__spec_start_handling(NULL, specs);
```
keshonok: Just a thought here.
So, an error in `specs[]` is found, where `spec->allow_reconfig` is set whereas the parent directive doesn't allow reconfiguration. Shouldn't this be a real error? This looks like a coding error (not the external configuration file error). So why this coding error is just corrected on the fly instead of hard stopping and allowing the developer to fix it right away?

ikoveshnikov: I was too concerned about reconfiguration and not breaking running system that i've missed that the check will happen on the very first start, whatever the configuration would be.

ikoveshnikov: Replaced the error report with BUG_ON. Since we have tests that use directives marked as `allow_reconfig` the coding error won't make it's way to production. 
Other such directives should be thoroughly tested since BUG_ON may fire during reconfiguration of running system. That happens if the affected directive is added to the configuration only during reconfiguration and not present in startup configuration.

tempesta_fw/cfg.c
```diff
@@ -748,28 +746,46 @@ spec_start_handling(TfwCfgSpec specs[])
 		BUG_ON(!spec->handler);
 		if (spec->handler == &tfw_cfg_handle_children)
 			BUG_ON(!spec->cleanup);
+		BUG_ON(parent && spec->allow_reconfig && !parent->allow_reconfig);
```
krizhanovsky: Is it absolutely obvious, that the system is going to crash and we have to crash right now? Basically it's discouraged to use `BUG_ON()` construction. Can we use `WARN_ON()` instead, abort the configuration and Tempesta start procedure?

Issue #520, opened 2016-05-25
[Frang] negative cache entries rate limit


krizhanovsky:
We store negative cache entries in the cache, so a DoS attack is possible when a user just generates any URI and efficiently exhaust Web-cache memory by sending the requests to Tempesta. So Frang should provide a rate limit for number of negative cache entries per time unit.

`tfw_cache_mgr` (#515) must clear the entries out.

While [random URLs](https://blog.sucuri.net/2014/02/layer-7-ddos-blocking-http-flood-attacks.html), producing tons of negative responses and cache entries, is a popular attack, we also should be able to limit per-client rate of cache bypassing requests. The limit is good to have, but it could be harmful, so #488 solves the problem in more gentle way by QoS.

Please update [the Wiki page](https://github.com/tempesta-tech/tempesta/wiki/DDoS-mitigation).

krizhanovsky:
Actually duplicate of #717, added the use case for [Web cache bypass](https://github.com/tempesta-tech/tempesta/wiki/DDoS-mitigation#web-cache-bypass) DDoS attack to the Wiki.

PR #896, opened 2018-01-25
Fix #838, fix #842: make GFSM get more data on L4-L7

### LINKED ISSUE #838: GFSM revision for HTTP responses

### LINKED ISSUE #842: Frang: wrong handle of errors during init.
The error is in https://github.com/tempesta-tech/tempesta/blob/e88059d183bb45d63ed62351a83923c0a8038aaa/tempesta_fw/classifier/frang.c#L1118

The function returns non-null `r` on errors. But if error happens during one of `tfw_gfsm_register_hook()` calls, `r` equals to `0`, so the `frang_init()` function returns successful error code (`0`).

krizhanovsky:
1. GFSM #838: use TfwFsmData as L4-L7 data descriptor for GFSM - pass skb and
   off as well as request and response, probably NULL, to each call.
2. GFSM #842: actually hooks are unregisterable for a while;
3. Frang: use explicit uint type conversion instead of FRANG_RESP_TIME_PERIOD
4. Frang: no need to keep per-quantum counter in 64-bit integer;
5. Adjust levels of debug messages on SS layer according to log.h comment;
6. Small cleanups all around.

### DIFF
--- a/tempesta_fw/classifier/frang.c
+++ b/tempesta_fw/classifier/frang.c
@@ -558,7 +558,7 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
 static unsigned int
 frang_resp_quantum(void)
 {
-	return jiffies / HZ * FRANG_FREQ / frang_cfg.http_resp_code_block->tf;
+	return jiffies * FRANG_FREQ / (frang_cfg.http_resp_code_block->tf * HZ);
 }
 
 static int
--- a/tempesta_fw/classifier/frang.c
+++ b/tempesta_fw/classifier/frang.c
@@ -24,7 +24,7 @@
  * Or, that singular header fields may not be duplicated in an HTTP header.
  *
  * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
- * Copyright (C) 2015-2017 Tempesta Technologies, Inc.
+ * Copyright (C) 2015-2018 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
@@ -57,7 +57,7 @@
 
 MODULE_AUTHOR(TFW_AUTHOR);
 MODULE_DESCRIPTION("Tempesta static limiting classifier");
-MODULE_VERSION("0.2.0");
+MODULE_VERSION("0.3.0");
 MODULE_LICENSE("GPL");
 
 /* We account users with FRANG_FREQ frequency per second. */
@@ -78,14 +78,13 @@ typedef struct {
  *   responses longer than 68 years;
  */
 typedef struct {
-	long		cnt;
+	unsigned int	cnt;
 	unsigned int	ts;
 } __attribute__((packed)) FrangRespCodeStat;
 
 #define FRANG_HTTP_CODE_MIN 100
 #define FRANG_HTTP_CODE_MAX 599
 #define FRANG_HTTP_CODE_BIT_NUM(code) ((code) - FRANG_HTTP_CODE_MIN)
-#define FRANG_RESP_TIME_PERIOD (1 << (sizeof(int) * 8 - 1))
 
 /**
  * Response code block setting
@@ -174,6 +173,17 @@ do {									\
 #define frang_limmsg(lim_name, curr_val, lim, addr)			\
 	frang_msg(lim_name " exceeded", (addr), ": %ld (lim=%ld)\n",	\
 		  (long)curr_val, (long)lim)
+
+#ifdef DEBUG
+#define frang_dbg(fmt_msg, addr, ...)					\
+do {									\
+	char abuf[TFW_ADDR_STR_BUF_SIZE] = {0};				\
+	tfw_addr_fmt_v6(&(addr)->v6.sin6_addr, 0, abuf);		\
+	TFW_DBG("frang: " fmt_msg, abuf, ##__VA_ARGS__);		\
+} while (0)
+#else
+#define frang_dbg(...)
+#endif
 
 static int
 frang_conn_limit(FrangAcc *ra, struct sock *unused)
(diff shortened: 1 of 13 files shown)


tempesta_fw/tls.c
```diff
@@ -109,6 +110,7 @@ tfw_tls_msg_process(void *conn, struct sk_buff *skb, unsigned int off)
 		return TFW_PASS; /* more data needed */
 	} else if (r == 0) {
 		struct sk_buff *nskb;
+		TfwFsmData data = {};
```
keshonok: Just a small note. My understanding is that an empty initializer list is a C++ feature, and also a GNU C extension, but it's not in standard C syntax. There are samples of both `{}` and `{0}` (or `{ 0 }`) throughout Tempesta's code. Perhaps it's best to bring it to one common style.
https://stackoverflow.com/questions/17589533/is-an-empty-initializer-list-valid-c-code/17589839#17589839

tempesta_fw/gfsm.c
```diff
@@ -3,31 +3,63 @@
...
+ * coroutines, i.e. a subroutine can:
+ *   1. call other subroutine;
+ *   2. yield control to other subroutine.
+ *
+ * GFSM is a voluntary (co-operative) data-driven scheduler for the subroutines.
+ * While traditional preemptive (time-sharing) scheduler cares about fair time
+ * slice sharing among processes, GFSM cares about processing hot (in sence of
+ * CPU caches) data by all subroutines, who, and only who, are interested in
+ * the data. I.e. when a packet arrives, all subroutines interested in
+ * processing the packet are called immediatelly while the packet data resides
```
aleksostapenko: `immediatelly` -> `immediately`

tempesta_fw/gfsm.c
```diff
@@ -3,31 +3,63 @@
...
  *
+ * A subroutine yields it's control flow by moving to the next state using
+ * tfw_gfsm_move().
+ *
  * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
- * Copyright (C) 2015 Tempesta Technologies, Inc.
+ * Copyright (C) 2015-2018 Tempesta Technologies, Inc.
  *
- * This program is free software; you can redistribute it and/or modify it
+ * This program is free software; you cana redistribute it and/or modify it
```
aleksostapenko: `cana` -> `can`

tempesta_fw/gfsm.c
```diff
@@ -125,9 +157,20 @@ tfw_gfsm_switch(TfwGState *st, int state, int prio)
...
 
+/**
+ * TODO #77 (User-kernel space transport): user-space processing is an
+ * asynchronous operation which can be called at different states of different
+ * FSMs, so GFSM must introdufe SLEAL logic:
+ *   1. an FSM (responsible for user-space message mapping) can return
+ *      TFW_STEAL;
+ *   2. getting the return code from gfsm_move() current FSM must finish
+ *      the message processing logic and return with stored current state;
+ *   3. when a user-space program finish GFSM must unwing the call stack and
```
aleksostapenko: `unwing` -> `unwind`

tempesta_fw/gfsm.c
```diff
@@ -125,9 +157,20 @@ tfw_gfsm_switch(TfwGState *st, int state, int prio)
 	return fsm_next;
 }
 
+/**
+ * TODO #77 (User-kernel space transport): user-space processing is an
+ * asynchronous operation which can be called at different states of different
+ * FSMs, so GFSM must introdufe SLEAL logic:
```
aleksostapenko: `introdufe` -> `introduce` and maybe `SLEAL` -> `STEAL`.

tempesta_fw/gfsm.h
```diff
@@ -129,32 +130,63 @@ enum {
 	TFW_POSTPONE	= SS_POSTPONE,
 };
 
+/**
+ * An input data for a subroutine (FSM) to process.
+ *
+ * Typically an FSM works solely on L4 or L7, however, in futuer, there could be
```
aleksostapenko: `futuer` -> `future`.

tempesta_fw/classifier/frang.c
```diff
@@ -541,18 +551,21 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+/**
+ * Monotonically increasing time quantums. The configured time frame is devided
+ * by FRANG_FREQ slots to get the quantums granularity.
+ */
 static unsigned int
 frang_resp_quantum(void)
 {
-	return ((jiffies / HZ) % FRANG_RESP_TIME_PERIOD) * FRANG_FREQ
-		/ frang_cfg.http_resp_code_block->tf;
+	return jiffies / HZ * FRANG_FREQ / frang_cfg.http_resp_code_block->tf;
```
aleksostapenko: In current implementation if `frang_cfg.http_resp_code_block->tf == 1`, then `ts % FRANG_FREQ` will always give 0.
It seems that `jiffies * FRANG_FREQ / (frang_cfg.http_resp_code_block->tf * HZ)` would produce better granularity.

tempesta_fw/http.c
```diff
@@ -2588,20 +2598,14 @@ tfw_http_resp_cache(TfwHttpMsg *hmresp)
...
-	prev_resp = req->resp;
-	req->resp = hmresp;
-	tfw_gfsm_move(state, TFW_HTTP_FSM_RESP_MSG_FWD, NULL, 0);
-	state->obj = prev_state_obj;
-	req->resp = prev_resp;
+	data.skb = NULL;
+	data.off = 0;
+	data.req = (TfwMsg *)req;
+	data.resp = (TfwMsg *)hmresp;
+	tfw_gfsm_move(&hmresp->conn->state, TFW_HTTP_FSM_RESP_MSG_FWD, &data);
```
ikoveshnikov: The hook was added in #717, recently closed #520 required the same for cached responses, according to [Wiki](https://github.com/tempesta-tech/tempesta/wiki/DDoS-mitigation#web-cache-bypass) the feature is required. Sorry but I got confused, is it required to support the #520? Now it's not supported.
If i's required, then either the same gfsm hook must be called in `tfw_http_req_cache_service()` or this hook must be moved to  `tfw_http_resp_fwd()`

krizhanovsky: The idea was that if we limit number of negative server response for a client, then it's enough for both of the issues, #520 and #717. The only difference between them is that in #717'th case there is typically the same URI in all client requests while in #520'th URIs are different. There is no need to care about Web cache filling if a client hitted the limit - just let a garbage collector (subject for TDBv2) collect all negative and useless cache entries.

PR #897, opened 2018-01-26
Long body in responses and requests

### LINKED ISSUE #619: Static test for unlimited request and response bodies
Processing of unlimited bodies is stopped by connection closing, so unit test for the cases isn't trivial. So static Python test must be developed to test the cases for request and response.

vladtcvs:
fix #619


tempesta_fw/t/functional/helpers/control.py
```diff
@@ -108,19 +108,71 @@ def set_user_agent(self, ua):
 class Wrk(Client):
     """ wrk - HTTP benchmark utility. """
 
+    class ScriptGenerator(object):
```
ikoveshnikov: Why you have decided to define class inside `Wrk` class? It's requred only in some cases, better to move the definition into separate file like it's done for `tempesta.py` and `nginx.py` which provide advanced control options for `Tempesta` and `Nginx` classes. I admit i'm not good in python and there can be much and much better implementations of current code, but i can't  agree here.

etc/tempesta_fw.conf
```diff
@@ -92,6 +92,14 @@
...
 #   None.
 
+cache 0;
+
+server 10.55.1.1:80;
+
+block_action error reply;
+
+block_action attack reply;
+
```
krizhanovsky: Seems a mistaken change: the file is used by end users, so never put any testing stuff here, especially, specific for your local installation.

vladtcvs: fixed

tempesta_fw/t/functional/helpers/deproxy.py
```diff
@@ -491,6 +532,8 @@ def handle_read(self):
                            '<<<<<\n%s>>>>>'
                            % self.response_buffer))
             raise
+        if len(self.response_buffer) > 0:
+            raise ParseError('Garbage after response end:\n```\n%s\n```\n' % self.response_buffer)
```
ikoveshnikov: Ok, now it's makes the job. But a either a `TODO` comment or rework is required here. Currently both client and response don't support message pipelining because of tests structure: messages from TempestaFW are validated one-by one (The test suite was developed before pipelining was supported). We may configure Client client with the full chain list and give it some instructions about when to use the pipelining. When the Server receives a new message it can save it to the right slot in chain list. Not all tests allow pipelining.

tempesta_fw/t/functional/helpers/wrk.py
```diff
@@ -0,0 +1,57 @@
...
+import os
+
+class ScriptGenerator(object):
+    """ Generate lua script """
+    request_type = "GET"
+    uri = "/"
+    headers = []
+    body = ""
+    filename = None
+    def __luaencode(self, value):
```
ikoveshnikov: Method can be a simple function. But why it's actually required? It does just nothing.

tempesta_fw/t/functional/helpers/wrk.py
```diff
@@ -0,0 +1,57 @@
...
+        config.write("    path = \"%s\",\n" % self.__luaencode(self.uri))
+        config.write("    headers = {\n")
+        for header in self.headers:
+            config.write("        [\"%s\"] = \"%s\",\n" % (header[0], header[1]))
+        config.write("    },\n")
+        config.write("    body = \"%s\",\n" % self.body)
+        config.write("}\n")
+        config.write("local req\n")
+        config.write("init = function()\n")
+        config.write("    req = wrk.format(r.method, r.path, r.headers, r.body)\n")
```
ikoveshnikov: If i'm not mistaken just setting 
```
wrk = {
    method  = "GET",
    path    = "/",
    headers = {},
    body    = nil,
}
```
will make the same job for you. In the same time `wrk` docs says that `request()` function is heavy and should be avoided:
https://github.com/wg/wrk/blob/7594a95186ebdfa7cb35477a8a811f84e2a31b62/SCRIPTING#L81-L84

tempesta_fw/t/functional/long_body/test_long_request.py
```diff
@@ -0,0 +1,68 @@
...
+import body_generator
+import os
+
+from testers import stress
+from helpers import tf_cfg, control, tempesta, remote, wrk
+
+class RequestTestBase(stress.StressTest):
+    """ Test long request """
+    config = "cache 0;\n"
+    root = "/tmp/long_body/"
```
ikoveshnikov: There is variable `client.workdir` in `tf_cfg` which points to `/tmp/client` by default. Use it as root dir for your directories and files.

tempesta_fw/t/functional/long_body/test_long_request.py
```diff
@@ -0,0 +1,68 @@
...
+    scriptdir = root + "scripts/"
+    wrk = None
+    clients = []
+    generator = None
+
+    def create_clients_with_body(self, length):
+        """ Create wrk client with long request body """
+        self.generator = wrk.ScriptGenerator()
+        self.generator.set_body(body_generator.generate_body(length))
+        if not os.path.exists(self.root):
```
ikoveshnikov: Use `remote.node.mkdir()` instead: first it always care about on which node should it make the directory, secondly it already throws an exception for test suit on errors.

tempesta_fw/t/functional/helpers/deproxy.py
```diff
@@ -414,6 +431,31 @@ def parse_firstline(self, stream):
         except:
             raise ParseError('Invalid Status code!')
 
+    def parse_body(self, stream):
+        """ RFC 7230 3.3.3 """
+        # 3.3.3 1
+        if self.method == "HEAD":
```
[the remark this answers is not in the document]
vladtcvs: independent flag can generate inconsistency. We don't have 302 and CONNECT tests now, and when we write such tests, we should add this cases here

vladtcvs:
problem with breaking other tests fixed

tempesta_fw/t/functional/helpers/deproxy.py
vladtcvs: Now we don't have pipelined functional test. So I've added  "TODO"

tempesta_fw/t/functional/helpers/wrk.py
vladtcvs: It should care about escaping, if we write test with "\"" and so on in content

tempesta_fw/t/functional/helpers/wrk.py
vladtcvs: fixed, but we need 
```
wrk.method = "GET",
wrk.path = ....
```

instead of `wrk = {...}`

tempesta_fw/t/functional/long_body/test_long_request.py
vladtcvs: wrk scripts copy code fixed

tempesta_fw/t/functional/helpers/wrk.py
vladtcvs: fixed

tempesta_fw/t/functional/helpers/wrk.py
vladtcvs: Now we don't have tests with content, which should be escaped. So now it does nothing

tempesta_fw/t/functional/long_body/test_long_request.py
vladtcvs: fixed

tempesta_fw/t/functional/helpers/chains.py
```diff
@@ -197,6 +217,34 @@ def base(uri='/', method='GET', forward=True, date=None):
                                       server_response=backend_resp)
     return copy.copy(base_chain)
 
+def response_403(date=None, connection=None):
```
ikoveshnikov: There is `response_500()` in the begging of the file, and ` response_403` here. Better to group code somehow.

tempesta_fw/t/functional/helpers/chains.py
```diff
@@ -33,6 +33,27 @@ def make_502_expected():
...
+    date = deproxy.HttpMessage.date_time_string()
+    headers = [
+        'Date: %s' % date,
+        'Content-Length: 0',
+        'Connection: keep-alive'
+    ]
+    response = deproxy.Response.create(status=500, headers=headers)
+    return response
+
+def response_502():
```
ikoveshnikov: There already  is https://github.com/tempesta-tech/tempesta/blob/390410117b8dd155c722b1f0f434437d43bb18ae/tempesta_fw/t/functional/helpers/chains.py#L28 That is the difference?

tempesta_fw/t/functional/helpers/chains.py
```diff
@@ -197,6 +217,34 @@ def base(uri='/', method='GET', forward=True, date=None):
...
+def response_403(date=None, connection=None):
+    if date is None:
+        date = deproxy.HttpMessage.date_time_string()
+    if connection is None:
+        resp = deproxy.Response.create(status=403,
+                                       headers=['Content-Length: 0'],
+                                       date=date,
+                                       body='')
+    else:
+        resp = deproxy.Response.create(status=403,
```
ikoveshnikov: Copy-paste. It's better to use variable to store list of headers and add `Connection:` header if `connection is not None`, then call `deproxy.Response.create()` only once

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -127,18 +127,28 @@ def __init__(self, threads=-1, uri='/', ssl=False):
...
             return
-        path = ''.join([os.path.dirname(os.path.realpath(__file__)),
-                        '/../wrk/', self.script, '.lua'])
-        script_path = os.path.abspath(path)
-        assert os.path.isfile(script_path), \
-               'No script found: %s !' % script_path
+        script_path = self.workdir + "/" + self.script
+
+        if self.copy_script:
+            local_path = ''.join([self.local_scriptdir, self.script])
```
ikoveshnikov: Wrk is always called on the same node as test frame work. Why need to copy file? We can just reference it. Which is old behaviour.

tempesta_fw/t/functional/helpers/chains.py
vladtcvs: fixed

tempesta_fw/t/functional/helpers/chains.py
vladtcvs: it's the same. fixed

tempesta_fw/t/functional/helpers/control.py
vladtcvs: It's for universal work with scripts from wrk/ and generated scripts

tempesta_fw/t/functional/helpers/chains.py
vladtcvs: fixed

vladtcvs: fixed

tempesta_fw/t/functional/long_body/test_long_response.py
```diff
@@ -0,0 +1,65 @@
...
+class ResponseTestBase(stress.StressTest):
+    """ Test long response """
+    config = "cache 0;\n"
+    filename = "long.bin"
+    uri = "/" + filename
+    fullname = ""
+
+    def create_content(self, length):
+        """ Create content file """
+        content = body_generator.generate_body(length)
```
ikoveshnikov: It's good to test a lot of different kind of response. I think, we even need to go further: we need to prepare a back end configuration that will be capable to respond to POST, to serve a different type of resources, small and big ones. That will be a great enhancement for tests, isn't it?
Wee need to create a separate issue for that for one of the next milestones. What do you think?

tempesta_fw/t/functional/helpers/wrk.py
```diff
@@ -0,0 +1,45 @@
...
+        config += "wrk.method = \"%s\"\n" % self.request_type
+        config +="wrk.path = \"%s\"\n" % self.__luaencode(self.uri)
+        config += "wrk.headers = {\n"
+        for header in self.headers:
+            name = self.__luaencode(header[0])
+            value = self.__luaencode(header[1])
+            config += "    [\"%s\"] = \"%s\",\n" % (name, value)
+        config += "},\n"
+        config += "wrk.body = \"%s\",\n" % self.__luaencode(self.body)
+        remote.client.copy_file(filename, config)
```
ikoveshnikov: We've discussed this line in the chat, it should be fixed.

tempesta_fw/t/functional/helpers/deproxy.py
```diff
@@ -512,6 +552,10 @@ def handle_read(self):
                            '<<<<<\n%s>>>>>'
                            % self.response_buffer))
             raise
+        if len(self.response_buffer) > 0:
```
ikoveshnikov: Please check with `test_pairing.py` tests, i expect problems here. Reach me in chat if you catch an error there, we'll discuss  a solution.

tempesta_fw/t/functional/long_body/test_long_response.py
vladtcvs: I think it's a good idea to make backend more configurable

tempesta_fw/t/functional/helpers/wrk.py
vladtcvs: fixed

PR #898, opened 2018-01-26
Fix #802: Add 'Age:' header for stored responses.

### LINKED ISSUE #802: [Cache] Generate Age header
RFC 7234 Section 4:
```
 When a stored response is used to satisfy a request without
   validation, a cache MUST generate an Age header field (Section 5.1),
   replacing any present in the response with a value equal to the
   stored response's current_age; see Section 4.2.3.
```

aleksostapenko:


tempesta_fw/cache.c
```diff
@@ -1571,6 +1572,23 @@ tfw_cache_build_resp(TfwCacheEntry *ce)
...
+	time_t age = tfw_cache_entry_age(ce);
+
+	if (!(digs = tfw_ultoa(age, cstr_age, TFW_ULTOA_BUF_SIZ)))
+		return false;
+
+	if (tfw_http_msg_hdr_xfrm(hmresp, "Age", sizeof("Age") - 1,
+				  cstr_age, digs, TFW_HTTP_HDR_RAW, 0))
+		return false;
+
+	return true;
```
ikoveshnikov: We have many same functions that add or replace one header, all of them return error code, not bool value. See `tfw_http_adjust_req()` for example. All the functions have the same logic,  so the return value  should be the same.

tempesta_fw/cache.c
```diff
@@ -1545,6 +1545,7 @@ tfw_cache_build_resp(TfwCacheEntry *ce)
 				  ce->status_len, &resp->s_line))
 		goto err;
 
+	resp->h_tbl->off = ce->hdr_num;
```
ikoveshnikov: Good catch!

tempesta_fw/t/functional/cache/age.py
```diff
@@ -0,0 +1,35 @@
...
+__copyright__ = 'Copyright (C) 2018 Tempesta Technologies, Inc.'
+__license__ = 'GPL2'
+
+class CacheAgeTester(deproxy.Deproxy):
+    """Tester helper for Age: header verification."""
+
+    def __init__(self, *args, **kwargs):
+        deproxy.Deproxy.__init__(self, *args, **kwargs)
+
+    def recieved_response(self, response):
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

PR #913, opened 2018-02-16
Fix #844: Avoid frang and sticky processing for whitelist requests.

### LINKED ISSUE #844: Whitelisting from nftables
Frang and especially sticky cookies modules may significantly impact user experience and the site availability for search engines, so we need a whitelist for IP addresses which should not be impacted by Frang and sticky cookies. Bot the modules must introduce a new options to bypass traffic with specified label. 

The new configuration option must be defined as

        whilelist_mark <mark>

, where <mark> is an integer value set in `skb->mark` by [nftables](https://wiki.nftables.org/wiki-nftables/index.php/Setting_packet_metainformation).

Please create an issue for appropriate functional test (for the next milestone) and add corresponding description to https://github.com/tempesta-tech/tempesta/wiki/Handling-clients

aleksostapenko:


tempesta_fw/http.c
```diff
@@ -742,6 +748,23 @@ tfw_http_error_resp_switch(TfwHttpReq *req, int status, const char *reason)
...
+
+	if (!tfw_wl_marks.mrks || !(TFW_CONN_TYPE(conn) & Conn_Clnt))
+		return;
+
+	for (i = 0; i < tfw_wl_marks.sz; ++i) {
+		if (tfw_wl_marks.mrks[i] == skb->mark) {
+			msg->flags |= TFW_HTTP_WHITELIST;
+			break;
+		}
+	}
```
krizhanovsky: Please use `bsearch()`.

tempesta_fw/http.c
```diff
@@ -3228,6 +3255,43 @@ tfw_cfgop_resp_body(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+		return -EINVAL;
+	}
+	if (ce->attr_n) {
+		TFW_ERR_NL("Unexpected attributes\n");
+		return -EINVAL;
+	}
+
+	tfw_wl_marks.sz = ce->val_n;
+	if (!(tfw_wl_marks.mrks = kmalloc(ce->val_n, GFP_KERNEL)))
+		return -ENOMEM;
```
keshonok: Is there an understanding of how many marks can be specified? If there can be a lot, then it would be convenient to allow repetitive use of this option, where just a subset of marks is specified in each line, for better readability.

Right now `allow_repeat` is set to `false` (which is the default), and the code can't handle multiple specifications of this option.

tempesta_fw/http.c

tempesta_fw/http.c
```diff
@@ -1645,6 +1674,9 @@ tfw_http_msg_create_sibling(TfwHttpMsg *hm, struct sk_buff **skb,
 	if (unlikely(!shm))
 		return NULL;
 
+	/* New message created, so maybe it should be in whitelist. */
+	tfw_http_mark_wl_new_msg(hm->conn, shm, *skb);
```
ikoveshnikov: This a pipelined request case. originally previous message and the new one were in the same skb, so the skb mark is the same for both. No need to start the search once again: just copy whitelist flag from the previous message.

tempesta_fw/http.c
```diff
@@ -3403,6 +3436,47 @@ tfw_cfgop_resp_body(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
 }
 
+static int
+tfw_cfgop_whitelist_mark(TfwCfgSpec *cs, TfwCfgEntry *ce)
+{
+	unsigned int i;
+	const char *val;
+
+	if (!ce->val_n) {
+		TFW_ERR_NL("Invalid number of arguments: %zu\n", ce->val_n);
```
ikoveshnikov: It will always say `Invalid number of arguments: 0`. May be a message like `("%s: At least one argument is required", ce->name)` will be more user friendly?

etc/tempesta_fw.conf
```diff
@@ -699,4 +699,21 @@
 # Default:
 #   All limits are disabled (the values are set to zero/false/empty).
 
+#
+# TAG: whilelist_mark
```
ikoveshnikov: whi**T**elist_mark. Same misprint in lines below. I was surprised to get `ERROR: don't know how to handle: 'whilelist_mark` after copying name from configuration example(:

tempesta_fw/http.c
aleksostapenko: Yes, this question was the subject of discussion and it was decided to use one directive since the number of marks should be supposedly small.

tempesta_fw/http.c

tempesta_fw/http.c

etc/tempesta_fw.conf
aleksostapenko: Yeah - copy-paste=(
Thanks! Corrected.

Issue #888, opened 2018-01-21
Fault on running Tempesta FW w/ configured, but not loaded Frang


krizhanovsky:
Config file, Frang limits are configured:
```
listen 192.168.100.4:80;
server 127.0.0.1:8080; #apache
cache 0;

frang_limits {
	http_ct_vals "text/plain" "text/html";
	client_header_timeout 1;
	client_body_timeout 1;
	request_rate 20;
	request_burst 15;
	http_host_required false;
}
```

However, Tempesta FW is started w/o Frang module:
```
# ./scripts/tempesta.sh --restart
```
which leads to the system crash:
```
[ 3716.047307] [tempesta] ERROR: don't know how to handle: 'frang_limits'
[ 3716.048160] [tempesta] ERROR: configuration parsing error:
[ 3716.048160]    6: frang_limits {
[ 3716.048160]       ^^^^^^^^^^^^^^
[ 3716.051545] general protection fault: 0000 [#1] SMP DEBUG_PAGEALLOC
[ 3716.052146] Modules linked in: tfw_sched_ratio(O) tfw_sched_http(O) tfw_sched_hash(O) tempesta_fw(O) tempesta_db(O) tempesta_tls(O) binfmt_misc bochs_drm ttm drm_kms_helper crct10dif_pclmul ghash_clmulni_intel drm aesni_intel aes_x86_64 lrw gf128mul glue_helper ablk_helper cryptd ppdev sg fb_sys_fops syscopyarea sysfillrect sysimgblt serio_raw parport_pc parport pcspkr button ip_tables x_tables autofs4 ext4 crc16 jbd2 fscrypto mbcache sr_mod sd_mod cdrom ata_generic ata_piix libata psmouse scsi_mod e1000 i2c_piix4
[ 3716.052146] CPU: 2 PID: 1044 Comm: sysctl Tainted: G           O    4.9.35-kdump #131
[ 3716.052146] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.10.2-2.fc27 04/01/2014
[ 3716.052146] task: ffff97eaba9ca6c0 task.stack: ffffbcdb408b4000
[ 3716.052146] RIP: 0010:[<ffffffffc07bc29e>]  [<ffffffffc07bc29e>] tfw_sg_release+0xe/0x80 [tempesta_fw]
[ 3716.052146] RSP: 0018:ffffbcdb408b7c88  EFLAGS: 00010202
[...]

PR #920, opened 2018-02-20
Add JS challenge

### LINKED ISSUE #536: JavaScript challenge for HTTP DDoS mitigation
If DDoS bot is able to process cookies and do redirects sent by sticky cookie, then JavaScript is useful to challenge a bot. So `http_sticky` module should be extended to `http_challenge` module which should implement JavaScript challenge as well.

The module shall do basically the same as sticky cookie, but using JavaScript instead of HTTP redirects: send 200 response with HTML document containing JavaScript code which sets encrypted/hashed Cookie and requests the same URI.

There is [reference implementation](http://kyprizel.github.io/testcookie-nginx-module/) by @kyprizel.

There is a good discussion about [PhantomJS detection](https://blog.shapesecurity.com/2015/01/22/detecting-phantomjs-based-visitors/), but all the methods aren't infalliable,

ikoveshnikov:
Add JavaScript challenge (fix #536) 
start frang on demand (#888)


tempesta_fw/http_sess.h
```diff
@@ -78,6 +78,19 @@ struct tfw_http_sess_t {
 	TfwStickyConn		st_conn;
 };
 
+enum {
+	/* Internal error, may be any number < 0. */
+	TFW_HTTP_SESS_FAILURE = -1,
+	/* Session successfuly obtained. */
+	TFW_HTTP_SESS_SUCCESS = 0,
+	/* Can't obtain session: new client; redirect message sent. */
```
krizhanovsky: => "a redirection message was sent"

tempesta_fw/http.c
```diff
@@ -1998,6 +2182,19 @@ tfw_http_req_cache_cb(TfwHttpReq *req, TfwHttpResp *resp)
 	tfw_http_req_zap_error(&equeue);
 	goto conn_put;
 
+send_503:
+	/*
+	 * Requested resource can't be challenged. Don't break response-request
+	 * queue on client side by dropping the request.
+	 */
+	HTTP_SEND_RESP(req, 503, "request dropped: can't send JS challenge.");
+	TFW_INC_STAT_BH(clnt.msgs_otherr);
```
krizhanovsky: This and all the below are security events, so it seem `msgs_filtout` counter must be incremented instead.

PR #927, opened 2018-02-26
Better verbosity for tests

ikoveshnikov:
Running tests on a new testbed configuration raises question, how effective tests are. I had been running tests for a long time before I realised that the load was extremely low and tests was ineffective.

Changes:
- Get requests rate from `wrk`  output and show it on 2-nd level of debug
- Echo full `wrk` output on debug level 3
- Warn on `wrk` socket errors
- Add basic tests for predict and dynamic ratio schedulers

At least now it's easier to look though tests results to check for performance regressions.

### DIFF
--- a/tempesta_fw/t/functional/helpers/control.py
+++ b/tempesta_fw/t/functional/helpers/control.py
@@ -152,6 +152,14 @@ class Wrk(Client):
         m = re.search(r'Requests\/sec:\s+(\d+)', stdout)
         if m:
             self.rate = int(m.group(1))
+        m = re.search(r'(Socket errors:.+)\n', stdout)
+        if m:
+            tf_cfg.dbg(0, ("WARNING! Socket errors on wrk. "
+                          "Too many concurrent connections?"))
+            err_m = re.search(r'\w+ (\d)+, \w+ (\d)+, \w+ (\d)+, \w+ (\d)+',
+                              m.group(1))
+            self.errors += (int(err_m.group(1)) + int(err_m.group(2))
+                            + int(err_m.group(3)) + int(err_m.group(4)))
         return True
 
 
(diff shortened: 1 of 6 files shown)


tempesta_fw/t/functional/helpers/control.py
```diff
@@ -89,7 +91,9 @@ def prepare(self):
         return True
 
     def results(self):
-        return self.requests, self.errors
+        if (self.rate == 0):
+            self.rate = self.requests / self.duration
```
vladtcvs: As I understood, `rate == 0` used as an uninitialized value. An if we see `rate==0`, we initialize it with actual value. But `rate == 0` is also possible in (vary rare) case of `self.requests == 0`. I think, it's better to use `rate==-1` for uninitialized value

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -145,6 +149,17 @@ def parse_out(self, stdout, stderr):
         m = re.search(r'Non-2xx or 3xx responses: (\d+)', stdout)
         if m:
             self.errors = int(m.group(1))
+        m = re.search(r'Requests\/sec:\s+(\d+)', stdout)
+        if m:
+            self.rate = int(m.group(1))
+        m = re.search(r'(Socket errors:.+)\n', stdout)
+        if m:
+            tf_cfg.dbg(0, ("WARNING! Socket errors on wrk. "
```
vladtcvs: Debug level 0? Shouldn't we print only results of test in debug level 0?

ikoveshnikov: Wrk errors can mean a lot: network errors, HTTP parsing errors, and more. Unfortunately we can't get better statistics than `connect`, `read`, `write` or  `timeout` errors. Since normally none of them are acceptable, it seems good point to raise the test failure. Allowing to treat such errors as ordinary one can be also helpful.
I've updated the patch to reflect this behaviour. Any suggestions?

PR #948, opened 2018-03-11
Fix 945 - mixed_tests

### LINKED ISSUE #945: mixed_requests.test_mixed.RarelyUsedRequests: HTTP client detected errors
The issue appears after #936 fix: the test still fails:
```
# ./run_tests.py mixed_requests.test_mixed.RarelyUsedRequests
File tests_resume.json not found
Not resuming: File tests_resume.json not found

Running functional tests...

test (mixed_requests.test_mixed.RarelyUsedRequests) ... 
	Init test case...
	Starting Nginx on 127.0.0.1:8000
	Starting TempestaFW on localhost
	Running 1 HTTP clients on localhost
Running 10s test @ http://127.0.0.1/
  10 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    14.67ms   22.67ms 122.82ms   84.67%
    Req/Sec   622.10    793.05     9.01k    89.25%
  49251 requests in 10.02s, 6.96MB read
  Socket errors: connect 0, read 8848, write 0, timeout 0
  Non-2xx or 3xx responses: 31472
Requests/sec:   4916.48
Transfer/sec:    711.45KB
 
WARNING! Socket errors on wrk. Too many concurrent connections?
	Client: errors: 40320, requests: 49251, rate: 4916
	Clients in total: errors: 40320, requests: 49251, rate: 4916
FAIL
	Stoping TempestaFW on localhost
	Stoping Nginx on 127.0.0.1:8000
	Removing Nginx config for 127.0.0.1:8000

======================================================================
FAIL: test (mixed_requests.test_mixed.RarelyUsedRequests)
  File "/root/tempesta/tempesta_fw/t/functional/mixed_requests/mixed_test.py", line 17, in test
    self.generic_test_routine(self.config)
  File "/root/tempesta/tempesta_fw/t/functional/testers/stress.py", line 171, in generic_test_routine
    self.assert_clients()
  File "/root/tempesta/tempesta_fw/t/functional/testers/stress.py", line 113, in assert_clients
    self.assertEqual(err, 0, msg='HTTP client detected errors')
AssertionError: HTTP client detected errors

Ran 1 test in 11.003s

FAILED (failures=1)
```

The result is the same for the following patch  (I also tried to replace `OPTIONS` with `GET` or `PATCH` and the results are the same):
```diff
diff --git a/tempesta_fw/t/functional/wrk/mixed.lua b/tempesta_fw/t/functional/wrk/mixed.lua
index f68100a..7734276 100644
--- a/tempesta_fw/t/functional/wrk/mixed.lua
+++ b/tempesta_fw/t/functional/wrk/mixed.lua
@@ -3,14 +3,6 @@
 init = function(args)
    local r = {}
    r[1] = wrk.format("OPTIONS", "/")
-   r[2] = wrk.format("GET", "/")
-   r[3] = wrk.format("HEAD", "/")
-   r[4] = wrk.format("POST", "/")
-   r[5] = wrk.format("PUT", "/")
-   r[6] = wrk.format("PATCH", "/")
-   r[7] = wrk.format("DELETE", "/")
-   r[8] = wrk.format("TRACE", "/")
-   r[9] = wrk.format("CONNECT", "/")
    req = table.concat(r)
 end
```

I tried to execute OPTIONS request using curl and it works fine with Tempesta FW configured as a proxy:
```
# curl -v -X OPTIONS http://192.168.100.4:80/
*   Trying 192.168.100.4...
* TCP_NODELAY set
* Connected to 192.168.100.4 (192.168.100.4) port 80 (#0)
> OPTIONS / HTTP/1.1
> Host: 192.168.100.4
> User-Agent: curl/7.52.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Sat, 10 Mar 2018 18:59:06 GMT
< Allow: GET,HEAD,POST,OPTIONS,HEAD,HEAD
< Content-Length: 0
< Connection: 
< Content-Type: text/html
< Server: Tempesta FW/0.5.0
< Via: 1.1 tempesta_fw (Tempesta FW 0.5.0)
< 
* Curl_http_done: called premature == 0
* Connection #0 to host 192.168.100.4 left intact
```

So it seems there is some problem with the wrk script. Also please make error message more developer friendly: `HTTP client detected errors` - what can I do with the error message and how to debug the problem? What errors exactly? Which requests produced the errors?

vladtcvs:
fix #945


vladtcvs:
Also parly fixes 702 - we can just don't count 502 errors. But it is only a temporary solution, because we should can detect other 502 errors, not related to backend connection errors

tempesta_fw/t/functional/mixed_requests/mixed_test.py
```diff
@@ -7,6 +7,12 @@ class MixedRequests(stress.StressTest):
     script = None
     clients_num = int(tf_cfg.cfg.get('General', 'concurrent_connections'))
 
+    def create_servers(self):
+        port = tempesta.upstream_port_start_from()
+        server = control.Nginx(listen_port=port)
+        server.config.set_return_code()
+        self.servers = [server]
```
ikoveshnikov: Better to call `create_servers()` of the base class and then  set return code for each server: if base class function will changed, you won't need to update this test.

tempesta_fw/t/functional/wrk/connect.lua
```diff
@@ -0,0 +1,2 @@
+wrk.method = "CONNECT"
```
ikoveshnikov: `Connect` is very specific method: after 200 response connection between client and server works as a pipe, not as HTTP connection. We don't support the method. What the test checks?

vladtcvs: the test checks not failing of tempesta under multiple connect requests

tempesta_fw/t/functional/mixed_requests/mixed_test.py
vladtcvs: fixed

Issue #537, opened 2016-05-27
Fast access logging for analytics


krizhanovsky:
Now all information from config file parsing errors to clients blocking is written to dmesg. Instead following logs on top of TempestaDB must be introduced:
- access log - `TfwHttpReq` must be passed as an argument to `tfw_log_access()` function.
- security log - all security event must be logged there (e.g. block client IP, Frang rate limiting etc.). It seems there could be several logging functions which write events at different layers. Since we log blocking events, the events might overflow the log due to DoS attack. At the moment they're logged using `net_warn_ratelimited()`, so we're good with the system ratelimiter. We should either leave the security log in the kernel log or implement our own rate limiter.
- error log - all HTTP processing events must go there (e.g. invalid URI), records in the log also must be done in `TfwHttpReq` context;
- probably stress (performance) events.

The 2 modes of logging must be implemented:
1. TDB tables with automatic eviction of old records (ring buffer) to fit in RAM
1. logs transmission to a remote server using TCP synchronous sockets (not to lose records under peak load)

The logs should be configured by independent configuration options:

```
    log_access /opt/tempesta/db/log_access.tdb <variables>
    log_error /opt/tempesta/db/log_error.tdb <variables>
    log_security 192.168.1.1:5000 <variables>
```
`variables` is the list of variables to log:
1. remote_addr - remote user IP
2. time - local time (miliseconds since epoch)
3. method - request method
4. uri
5. resp_status - response status
6. body_sent - response bytes sent
7. resp_time - response time
8. values of any special header with `srvhdr_` or `clntcdr_` prefixes and `-` changed to `_`, e.g. `srvhdr_set_cookie` or `clnt_user_agent`
9. cache miss/hit
10. origin IP

In general, **Tempesta DB should provide streaming data processing (#515 and #516) foundation** for the logging application. Probably we need to keep all request and response headers with the metadata (e.g. timings, chunks information, TCP/IP and TLS layers etc) for a relatively short sliding window. Such data is extremely useful to query to debug immediate performance issues and DDoS attacks.

*TBD a possible application is to batch events in front of a time series DB, e.g. ClickHouse, InfluxDB or TimescaleDB.*

We need to support per-vhost logging, i.e. thousands of vhosts having several logs each. This could be done using either secondary index #733 or we have to be able to scale to thousands of TDB tables.

[...]

PR #952, opened 2018-03-13
Fix #903: add logging for all error responses.

### LINKED ISSUE #903: Poor handling of HTTP error codes
Use configuration from https://github.com/tempesta-tech/tempesta/wiki/Web-security#dns-rebinding and send a request like:
```
$ strace -s512 wget http://192.168.100.4/
```
(no `Host` header). In the strace output you'll see
```
HTTP/1.1 502 Bad Gateway\r\n
Date: Mon, 05 Feb 2018 10:09:34 GMT\r\n
Content-Length: 0\r\n
Connection: keep-alive\r\n\r\n
```
, no `Server` header, which is very wished. Moreover, Tempesta FW doesn't log anything in dmesg - all HTTP error responses must be logged with net ratelimiting.

aleksostapenko:


tempesta_fw/http.c
```diff
@@ -834,16 +856,24 @@ tfw_http_enum_resp_code(int status)
...
+	__tfw_http_send_resp(req, code);
+}
 
-	tfw_http_send_resp(req, code);
+/* Common interface for sending error responses. */
+void
+tfw_http_send_resp(TfwHttpReq *req, int status, const char *reason)
+{
+	if (!(tfw_blk_flags & TFW_BLK_ERR_NOLOG))
+		TFW_WARN_ADDR(reason, &req->conn->peer->addr);
```
krizhanovsky: Please add the `status` to the warning message, to get something like:
```
[tempesta] Warning: request dropped: processing error, status 502: ::ffff:c0a8:6401
```
For more verbose logging we have #537, but now we still need something more administration-friendly.

Issue #941, opened 2018-03-06
Paired request missing


ikoveshnikov:
A few messages ` Warning: Paired request missing, HTTP Response Splitting attack?` may appear in long term tests. Tempesta drops the server connection on such situations. To reproduce run e.g. `sched.test_ratio.RatioStressTest` with relatively long duration. I used duration of 10 hours, first message appeared after 5 hours of testing. Total amount of messages was ~10.

In the `sched.test_ratio.RatioStressTest` failovering of backend server connections is quite aggressive: every 100 request. In the same time `sched.test_ratio.RatioPerfTest` which is more relaxed because of disabled failovering shows no such errors.

Server connections failovering must be revised.
Tested on 5abfde71530da4edd2b34c5a31de350d1f1180e8

krizhanovsky:
It seems one probable scenario is described in comment https://github.com/tempesta-tech/tempesta/pull/884#discussion_r168962872 :

A client sends us several requests and the client connection immediately terminated. All of the requests are in fwd_queue's and are being sent to servers, reforwarded and so on until all of the receive responses. Then `tfw_http_resp_fwd()` finds empty seq_queue of the client connection and calls `ss_close_sync()` for each of the requests. All in all, we do a lot of unnecessary work.

It's OK to leave the dead requests in fwd_queue's to minimize lock contention, but it's easy and cheap to:

1. set some flag for the reuests to make them evicted from fwd_queue right they're going to be send to a server or rescheduled;
2. we don't need to call `ss_close_sync()` in `tfw_http_resp_fwd()` since all connection_error and connection_drop callbacks follow `ss_do_close()` call in sock.c, so `tfw_http_conn_cli_drop()` is always called on closed socket.
3. `tfw_http_conn_cli_drop()` should call `tfw_connection_put()` just like `tfw_sock_srv_connect_drop()` and `tfw_sock_srv_connect_failover()` do. When `tfw_http_resp_fwd()` calls `tfw_http_conn_msg_free()` for the last request referencing the client connection, the connection is freed.

krizhanovsky:
We didn't see the problem for a while. The optimization proposal is moved to #687 as the right place for the performance improvement, not a bug.

PR #957, opened 2018-03-15
Split skb if can't allocate a new response

ikoveshnikov:
If sibling response can't be allocated or paired with any request, then splitting skb is required to send client exactly one response, but not response and rest of data in skb.

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -1818,14 +1818,21 @@ tfw_http_msg_create_sibling(TfwHttpMsg *hm, struct sk_buff **skb,
 
 	/* The sibling message belongs to the same connection. */
 	shm = (TfwHttpMsg *)tfw_http_conn_msg_alloc(hm->conn);
-	if (unlikely(!shm))
+	if (unlikely(!shm)) {
+		/*
+		 * Split skb to proceed with current message @hm, or rest of
+		 * data in this connection will be attached to the message.
+		 */
+		if (type == Conn_Srv)
+			*skb = ss_skb_split(*skb, split_offset);
 		return NULL;
+	}
 
 	/*
 	 * New message created, so it should be in whitelist if
 	 * previous message was (for client connections).
 	 */
-	if (TFW_CONN_TYPE(hm->conn) & Conn_Clnt)
+	if (type == Conn_Clnt)
 		shm->flags |= hm->flags & TFW_HTTP_F_WHITELIST;
 
 	/*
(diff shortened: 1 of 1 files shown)


tempesta_fw/http.c
```diff
@@ -1818,14 +1818,21 @@ tfw_http_msg_create_sibling(TfwHttpMsg *hm, struct sk_buff **skb,
...
 	/* The sibling message belongs to the same connection. */
 	shm = (TfwHttpMsg *)tfw_http_conn_msg_alloc(hm->conn);
-	if (unlikely(!shm))
+	if (unlikely(!shm)) {
+		/*
+		 * Split skb to proceed with current message @hm, or rest of
+		 * data in this connection will be attached to the message.
+		 */
+		if (type == Conn_Srv)
+			*skb = ss_skb_split(*skb, split_offset);
```
krizhanovsky: Yeah, makes sense. But what about 2 pipelined requests? It seems we send then as one request to the server and the server returns us 2 responses - we can't find a request for the 2nd response and get #941...

tempesta_fw/http.c
```diff
@@ -1818,14 +1818,21 @@ tfw_http_msg_create_sibling(TfwHttpMsg *hm, struct sk_buff **skb,
...
 		return NULL;
+	}
 
 	/*
 	 * New message created, so it should be in whitelist if
 	 * previous message was (for client connections).
 	 */
-	if (TFW_CONN_TYPE(hm->conn) & Conn_Clnt)
+	if (type == Conn_Clnt)
```
krizhanovsky: `type` was unused in the function and I think it's better to remove the argument: `hm->conn->proto.type` is quick after all.

tempesta_fw/http.c
ikoveshnikov: If we can't allocate a sibling message, it means that we have to drop the connection. For server connection we can use message already parsed from the connection. It's normal response and we can forward it to the client. Just need to set correct framing. And the patch is about it.

Situation is little different with client connections. As I said the connection is closed, so we drop the already processed request without further forwarding to backend server. There is no sense in splitting skb for client connections.
https://github.com/tempesta-tech/tempesta/blob/4935217018d37315f24166e38d936a0fbeb3d6bc/tempesta_fw/http.c#L2718-L2732

That's not a reason for #941

krizhanovsky: Ok, has sense. Please add a short comment that we do not care about requests there and just drop client connections

krizhanovsky: Ok, has sense. Please add a short comment that we do not care about requests there and just drop client connections

PR #958, opened 2018-03-15
Warn user on config parsing errors and many fixes to functional tests

ikoveshnikov:
As we debugged some tests with @vladtcvs it turned out that many errors during configuration parsing may not be shown to end user due to rate limiting. Show all warnings and errors during start/stop


tempesta_fw/t/functional/reconf/reconf_stress.py
```diff
@@ -83,16 +83,22 @@ def stress_reconfig_generic(self, configure_func, reconfigure_func):
...
         self.tempesta.get_stats()
 
-        r_thread.join()
+        self.r_thread.join()
         self.assert_clients()
 
+    def tearDown(self):
+        # Wait for reconfig thread if it's not finished (exception was thrown
+        # during stress_reconfig_generic()
+        self.r_thread.join()
```
vladtcvs: need to check if self.r_thread is initialized.

PR #966, opened 2018-03-19
fix wrk command

vladtcvs:
If number of CPU is more than number of connection, create only number_of_connections threads

### DIFF
--- a/tempesta_fw/t/functional/helpers/control.py
+++ b/tempesta_fw/t/functional/helpers/control.py
@@ -162,6 +162,8 @@ class Wrk(Client):
         # count for remote node.
         if self.threads == -1:
             self.threads = remote.get_max_thread_count(self.node)
+        if self.threads > self.connections:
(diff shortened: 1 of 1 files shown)


tempesta_fw/t/functional/helpers/control.py
```diff
@@ -162,6 +162,8 @@ def form_command(self):
         # count for remote node.
         if self.threads == -1:
             self.threads = remote.get_max_thread_count(self.node)
+        if self.threads > self.connections:
+            self.threads = self.connections
```
ikoveshnikov: If `self.connections` is not multiple of `self.threads`, then `wrk` will establish only `threads * (connections // threads)` connections. The other `connections % threads` connections will be ignored. Is it worth to print warning in this case, what do you think?

vladtcvs: Good idea, I will add such warning

PR #967, opened 2018-03-19
Warn about number of connections

vladtcvs:


vladtcvs:
https://github.com/tempesta-tech/tempesta-test/pull/6

PR #968, opened 2018-03-20
Test control

vladtcvs:


tempesta_fw/t/functional/tempesta_test.py
```diff
@@ -0,0 +1,89 @@
+import unittest
+from helpers import shell, remote
+
```
ikoveshnikov: 1. Missing copyright. 
2. Need to document in Readme.md, why the file is needed, file docstring is required. Just to make clear that the file is not intended to be started manually

tempesta_fw/t/functional/tempesta_test.py
```diff
@@ -0,0 +1,89 @@
...
+class TempestaTestEngine(object):
+
+    def __init__(self, v_level=1):
+        all_tests = []
+        loader = unittest.TestLoader()
+        shell.testsuite_flatten(all_tests, loader.discover('.'))
+        self.v_level = v_level
+        self.tests = [TempestaTest(t) for t in all_tests]
+        remote.connect()
```
ikoveshnikov: How about nofile limit? Should it be defined here?  https://github.com/tempesta-tech/tempesta/blob/43e35e9742b7f82adb1411d94a5bf7ff4d310c17/tempesta_fw/t/functional/run_tests.py#L153

tempesta_fw/t/functional/tempesta_test.py
```diff
@@ -0,0 +1,89 @@
...
+        for t in self.tests:
+            if t.state != TEST_DISABLED:
+                t.reset()
+
+    def run_tests(self):
+        pending = [t.test for t in self.tests if t.state == TEST_PENDING]
+        testsuite = unittest.TestSuite(pending)
+        testRunner = unittest.runner.TextTestRunner(verbosity=self.v_level,
+                                        resultclass=self.__resultclass())
+        testRunner.run(testsuite)
```
ikoveshnikov: How can we care about kernel crashes? E.g. crash happens during one tests and all the following testa are failed.

Issue #959, opened 2018-03-18
Dead loop in tfw_sg_wait_release()


krizhanovsky:
Test
```
test (mixed_requests.test_mixed.GetPostRequests) ... FAIL
```
causes dead loop with
```
[ 1926.684030] [tempesta] Warning: pending for server callbacks to complete for 5s, 1 server groups still exist
[ 1931.688017] [tempesta] Warning: pending for server callbacks to complete for 5s, 1 server groups still exist
[ 1936.692039] [tempesta] Warning: pending for server callbacks to complete for 5s, 1 server groups still exist
...
```

The reason for the leaving server group must be fixed. Also, it really has sense to crash instead of infinitely wait.

ikoveshnikov:
Then a response is received, corresponding request is removed from servers's forward queue. If the request is not the first one in client's `seq_queue`, it won't be forwarded to the client right away.
When the client disconnects, `tfw_http_conn_cli_drop()` is called, but requests-responses pairs are not freed. That is the reason, why server connections are not restored after closing by backend servers and not destroyed on shutdown.

PR #969, opened 2018-03-20
Fix releasing of requests and responses on client disconnects

ikoveshnikov:
See explanation in https://github.com/tempesta-tech/tempesta/issues/959#issuecomment-374545478.

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -1753,13 +1753,11 @@ tfw_http_conn_release(TfwConn *conn)
 }
 
 /*
- * Deq the request from @seq_queue and free the request
- * and the paired response.
+ * Free the request and the paired response.
  */
 static inline void
 tfw_http_resp_pair_free(TfwHttpReq *req)
 {
-	list_del_init(&req->msg.seq_list);
 	tfw_http_conn_msg_free(req->pair);
 	tfw_http_conn_msg_free((TfwHttpMsg *)req);
 }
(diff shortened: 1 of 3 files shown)


tempesta_fw/http.c
```diff
@@ -1803,6 +1801,14 @@ tfw_http_conn_cli_drop(TfwCliConn *cli_conn)
 	list_for_each_entry_safe(req, tmp, seq_queue, msg.seq_list) {
 		req->flags |= TFW_HTTP_F_REQ_DROP;
 		list_del_init(&req->msg.seq_list);
+		/*
+		 * Response is processed and removed from fwd_queue, need to be
+		 * destroyed.
+		 */
+		if (req->resp && (req->resp->flags & TFW_HTTP_F_RESP_READY)) {
+			tfw_http_resp_pair_free(req);
+			TFW_INC_STAT_BH(serv.msgs_otherr);
```
krizhanovsky: Why this is a server error? It seems that a server successfully respond to the request and we're here just because the client drops the connection?

tempesta_fw/http_msg.c
```diff
@@ -948,6 +948,9 @@ tfw_http_msg_free(TfwHttpMsg *m)
 	if (!m)
 		return;
 
+	/* Check that the paired response was destroyed before request. */
+	if (m->conn && (TFW_CONN_TYPE(m->conn) & Conn_Clnt))
+		WARN_ON_ONCE(m->pair);
```
krizhanovsky: Usually we don't write assertions this way, but place the whole condition into the assertion.

tempesta_fw/http.c
ikoveshnikov: Here is output of performance statistics before the fix:
```
. . .
Server messages received		: 20
Server messages forwarded		: 16
Server messages parsing errors		: 0
Server messages filtered out		: 0
Server messages other errors		: 3
. . .
```
In this stats we see, that one response have disappeared (`received != forwarded + p_errors + filtered + o_errors)`. Other errors is the most suitable counter, which is used for the same situation in other places in code, e.g. https://github.com/tempesta-tech/tempesta/blob/43e35e9742b7f82adb1411d94a5bf7ff4d310c17/tempesta_fw/http.c#L2209-L2214

PR #982, opened 2018-03-26
Return 1 for failed unit tests

vladtcvs:


tempesta_fw/t/unit/run_all_tests.sh
```diff
@@ -34,3 +34,10 @@ rmmod tfw_fuzzer
 
 echo -e "\n @@@ UNIT TEST OUTPUT SUMMARY (see dmesg for full log):\n"
 dmesg | grep tfw_test
+
+res=`dmesg | grep tfw_test | tail -n 10 | grep "finish - all passed" | wc -l`
```
ikoveshnikov: You can use `-q`  option for grep. It suppresses output and returns `0` if match found `1` otherwise. It's better than counting lines.

PR #983, opened 2018-03-27
Fix #918: release connection if its' reference count is zero.

### LINKED ISSUE #918: [Cfg] can not cleanup after unsuccessful start
### Conditions

Load too many server groups using the script from #76 to exhaust system memory on Tempesta FW start (it's good to use VM with small memory to reproduce the problem quickly). Need fixed #908 to avoid deadlock on stopping modules.

### Description

In the conditions described above `sock_srv` start terminates with `-ENOMEM` error code and the system starts to stop all modules, but hangs in `tfw_sg_wait_release()`. It seems groups unregistering doesn't work properly.

```
[60281.321763] [tdb] Start Tempesta DB
[60281.923959] [tempesta] Initializing Tempesta FW kernel module...
[60282.036189] [tempesta] Registering new scheduler: hash
[60282.064816] [tempesta] Registering new scheduler: http
[60282.096349] [tempesta] Registering new scheduler: ratio
[60282.123114] [tempesta] Prepearing for the configuration processing.
[60475.161529] [tempesta] Configuration processing is completed.
[60475.985783] [tdb] Opened table /opt/tempesta/db/filter.tdb: size=16777216 rec_size=20 base=ffff93f4b9c00000
[60476.015806] TCP: request_sock_TCP: Possible SYN flooding on port 9090. Sending cookies.  Check SNMP counters.
[60541.289607] [tempesta] ERROR: Unable to start module 'sock_srv': -12
[60541.295307] [tempesta] Stopping all modules...
[60546.793950] [tdb] Close table 'filter0.tdb'
[60546.798267] [tempesta] modules are stopped
[60546.801731] [tempesta] Warning: Configuration parsing has failed. Clean up...
[60554.508073] [tempesta] Warning: pending for server callbacks to complete for 5s, 17 server groups still exist
[60559.512278] [tempesta] Warning: pending for server callbacks to complete for 5s, 17 server groups still exist
[60564.516044] [tempesta] Warning: pending for server callbacks to complete for 5s, 17 server groups still exist
[60569.520133] [tempesta] Warning: pending for server callbacks to complete for 5s, 17 server groups still exist
.......
```

aleksostapenko:


tempesta_fw/sock_srv.c
```diff
@@ -446,7 +446,8 @@ tfw_sock_srv_disconnect(TfwConn *conn)
 	 * attached to the connection are released. If the connection is
 	 * closed already, then force the release of attached resources.
 	 */
-	if (atomic_read(&conn->refcnt) != TFW_CONN_DEATHCNT)
+	refcnt = atomic_read(&conn->refcnt);
+	if (refcnt && refcnt != TFW_CONN_DEATHCNT)
```
keshonok: Could you please describe the case when the reference count is zero? It's not obvious from the patch. A good comment would be helpful.
The patch introduces extra work in a potentially frequently executed code, while it appears that it's for a very limited case. Can it be done differently? Like, calling `tfw_connection_put_to_death()` beforehand in that particular case?

aleksostapenko: @keshonok Thank you for review.
Fix has been reworked and extended. Comment added.

tempesta_fw/connection.h
```diff
@@ -185,7 +185,8 @@ enum {
 	TFW_CONN_B_QFORWD,	/* Need to forward requests in the queue. */
 	TFW_CONN_B_HASNIP,	/* Has non-idempotent requests. */
 
-	TFW_CONN_B_DEL		/* Remove connection */
+	TFW_CONN_B_DEL,		/* Remove connection */
+	TFW_CONN_B_ACTIVE	/* Connection has been got at least once. */
```
ikoveshnikov: `Connection has been got at least once.` is not actually true. The flag means that the connection itself got reference to server. And it's not about of  connection's reference counter. `Connection was in use or at least scheduled to be established.`

tempesta_fw/sock_srv.c
```diff
@@ -304,10 +304,17 @@ tfw_srv_conn_release(TfwSrvConn *srv_conn)
 	 * in deferred context after a short pause (in a timer
 	 * callback). The only reason not to start new reconnect
 	 * attempt is removing server from the current configuration.
+	 *
+	 * TFW_CONN_B_ACTIVE bit is checked here to see if the connection has
+	 * already got a server's reference; if so, then server's reference must
+	 * be put. If bit is not set, this means that connection didn't have
+	 * time to get server (due to some error in sock_srv start
```
ikoveshnikov: "didn't have time to get server" -> "didn't have time _or was never scheduled_ to reach server"
Using of `get` in this sentence is confusing since talk is about getting/putting reference counters.

tempesta_fw/sock_srv.c
```diff
@@ -483,7 +502,7 @@ tfw_sock_srv_connect_srv(TfwServer *srv)
 	 * that parallel execution can't happen with the same socket.
 	 */
 	list_for_each_entry(srv_conn, &srv->conn_list, list) {
-		tfw_server_get(srv);
+		tfw_sock_srv_conn_activate(srv, srv_conn);
 		tfw_sock_srv_connect_try_later(srv_conn);
```
ikoveshnikov: Seems like `tfw_sock_srv_connect_try_later()` is always called after `tfw_sock_srv_conn_activate()`. Let's move it there.

tempesta_fw/sock_srv.c
```diff
@@ -2035,7 +2056,10 @@ tfw_cfgop_update_sg_srv_list(TfwCfgSrvGroup *sg_cfg)
 
 	TFW_DBG2("Update server list for group '%s'\n", sg_cfg->orig_sg->name);
 
-	__tfw_sg_for_each_srv(sg_cfg->orig_sg, __tfw_cfgop_update_sg_srv_list, sg_cfg);
+	if ((r = __tfw_sg_for_each_srv(sg_cfg->orig_sg,
```
ikoveshnikov: Line is too log, is was also warned a couple of time that the next formatting is better:
```
r = __tfw_sg_for_each_srv(sg_cfg->orig_sg, __tfw_cfgop_update_sg_srv_list,
                          sg_cfg);
if (r)
    return r;
```

tempesta_fw/sock_srv.c
```diff
@@ -539,6 +558,7 @@ tfw_srv_conn_alloc(void)
 	INIT_LIST_HEAD(&srv_conn->fwd_queue);
 	INIT_LIST_HEAD(&srv_conn->nip_queue);
 	spin_lock_init(&srv_conn->fwd_qlock);
+	atomic_set(&srv_conn->refcnt, TFW_CONN_DEATHCNT);
```
ikoveshnikov: Need to comment why the refcounter is initialised with this value. Some thing like "Force releasing of taken server's reference counter on connection removing"

tempesta_fw/sock_srv.c
aleksostapenko: There is also `tfw_sock_srv_connect_try_later()` call here https://github.com/tempesta-tech/tempesta/blob/f4ef380a113a4d194a13ab3ae5bb4f719dac9797/tempesta_fw/sock_srv.c#L309, and it seems that we needn't increase server reference count in this path.

tempesta_fw/connection.h
aleksostapenko: Corrected.

tempesta_fw/sock_srv.c

tempesta_fw/sock_srv.c
aleksostapenko: Done.

Issue #878, opened 2017-12-29
DSL for test scenarios


krizhanovsky:
The task is motivated by https://github.com/tempesta-tech/tempesta/pull/851#discussion_r159002053 : each functional test must be developed in Python, in imperative way. It'd be much more efficient to develop tests in special purpose domain specific language (DSL), e.g. MariaDB [test suite](https://github.com/tempesta-tech/mariadb/tree/trunk/mysql-test) uses SQL-based shell-like language for the tests. The test are much more readable and can be quickly analysed for which scenarios are implemented.

Since the test system is in-house infrastructure project with different development & release flow, it has sense to consider it to move into separate project.

krizhanovsky:
The issue was originally in 0.8 milestone, however in 0.5 we faced significant setbacks in the tests development: many issues required significant changes in the kernel of the test suite. So the issue becomes crucial and first to do in 0.6 to arrange the test development process.

ikoveshnikov:
Need to create backend configuration capable to serve different types of resources:
- small files, around ten or hundred bytes, making several responses to fit single skb
- large files, which won't fit single skb
- successful post requests (not just overriding response code, but full featured response)
- failed post requests

ikoveshnikov:
All tests creates only one TfwClient, since only one source IP is used, many TfwClients are required for Frang tests

krizhanovsky:
Probably the most simple way to emulate many source IP addresses is to rewrite the addresses on ingress and egress directions by iptables. Mangling packets in both the directions is required to keep TCP sessions. You can generate the rules on the fly, e.g. by some hash on source port.

krizhanovsky:
Probably the most simple way to emulate many source IP addresses is to rewrite the addresses on ingress and egress directions by iptables. Mangling packets in both the directions is required to keep TCP sessions. You can generate the rules on the fly, e.g. by some hash on source port.

krizhanovsky:
Minutes from today call:

1. configuration templates must be explicitly defied in test .py files as a variables or members of a dictionary. Global test definitions (e.g. IP addresses, directories etc) must be handled using standard template engine (e.g. Template).
2. The template arguments must be used to define 'workstation' or 'CI' workload. E.g. (I took syntax from http://search.cpan.org/~samtregar/HTML-Template-Expr/Expr.pm)
```
keepalive_requests <TMPL_VAR EXPR="100 * keepalive_requests_mul">
```
[...]

PR #987, opened 2018-03-28
In progress: Example of new test description

vladtcvs:
fix: #878


vladtcvs:
See https://github.com/tempesta-tech/tempesta-test/pull/5

PR #994, opened 2018-04-01
Fix #938

### LINKED ISSUE #938: http_parser: missing the obs-text
This is an awesome work, thanks a lot! I am playing with the code of this repo. But I am confused by the HTTP parser here. `tempesta` use SIMD to boost up the parser, the most important primitive here is the `__tfw_strspn_simd`, but I find a problem I am trying to run the following tests, according to the comments in the code, the  `tfw_match_ctext_vchar` SHOULD match `0xff`(`ctext = .... | obs-text`), while it doesn't. The reason here might the usage of `_mm256_shuffle_epi8` when the most significant bit = 1, the shuffled result would be `0x00`, so when the input char = 0b1xxxxxxx, we not accept it, which means we lost the `obs-text`?
 
``` c
void test0() {
    char buffer[256];
    int i;
    for (i=0; i< 256; i++)
        buffer[i] = 'A';

    size_t res = tfw_match_ctext_vchar(buffer, 256);

    printf("%d\n", res);
}

void test1() {
    char buffer[256];
    int i;
    for (i=0; i< 256; i++)
        buffer[i] = 0xff;

    size_t res = tfw_match_ctext_vchar(buffer, 256);

    printf("%d\n", res);
}

int main() {
    tfw_str_init_const();

    test0();
    test1();
}
```
Output:

```
256
0
```

krizhanovsky:
See PoC and benchmarks at https://github.com/natsys/blog/commit/d0b016d0bd30f3fb6a4c76393c82e5d3973512ba

### DIFF
--- a/tempesta_fw/str_simd.c
+++ b/tempesta_fw/str_simd.c
@@ -3,7 +3,7 @@
  *
  * x86-64 SIMD routines for HTTP strings processing.
  *
- * Copyright (C) 2016 Tempesta Technologies, Inc.
+ * Copyright (C) 2016-2018 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
@@ -45,9 +45,12 @@
  * @QETOKEN	- `token` with double quotes and equal sign;
  * @NCTL	- ASCII VCHAR (RFC RFC 5234, Apendix B.1.) plus SP and HTAB,
  *		  used to accept HTTP header values;
- * @CTVCH	- union of ctext and VCHAR, example usage is User-Agent;
  * @XFF		- ASCII characters for HTTP X-Forwarded-For header (RFC 7239);
  * @CO		- cookie-octet as defined in RFC 6265 4.1.1 plus DQUOTE;
+ * @ZERO	- ASCII zero upper bound for matching 0 < v < SP;
+ * @SP		- ASCII SP low bound for matching 0 < v < SP;
+ * @HTAB	- ASCII HTAB;
+ * @DEL		- ASCII DEL;
  */
 static struct {
 	__m128i A128;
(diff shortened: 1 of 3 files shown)


tempesta_fw/str_simd.c
```diff
@@ -143,6 +146,13 @@ tfw_str_init_const(void)
 		0xf8, 0xfc, 0xfc, 0xfc, 0xfc, 0xfc, 0xfc, 0xfc,
 		0xfc, 0xfc, 0xfc, 0xf4, 0xd8, 0xfc, 0xfc, 0x7c);
 
+	/* ASCII zero character '\0' to use in strict signed comparison. */
```
aleksostapenko: It seems that it is not ASCII zero or `\0`, but `0x80` value.

tempesta_fw/str_simd.c
```diff
@@ -1042,6 +1066,89 @@ __tfw_strspn_avx2_128(const char *str, __m256i sm)
...
+__tfw_match_ctext_vchar32(const char *str)
+{
+	unsigned int r;
+
+	__m256i v = _mm256_lddqu_si256((void *)str);
+
+	__m256i sub = _mm256_sub_epi8(v, __C.ZERO256);
+	__m256i x = _mm256_cmpeq_epi8(v, __C.DEL256);
+	x |= _mm256_cmpgt_epi8(__C.SP256, sub);
+	x ^= _mm256_cmpeq_epi8(v, __C.HTAB256);
```
aleksostapenko: Also there is `_mm256_xor_si256()` function. Is it applicable here?

tempesta_fw/str_simd.c
krizhanovsky: `0x80` is used to go from necessary unsigned comparison to signed comparison which we only have in SIMD. I described the step from Hacker's Delight in the blog post http://natsys-lab.blogspot.ru/2016/10/http-strings-processing-using-c-sse42.html . But, yes, `0 - 0x80` would be simpler and more readable.

tempesta_fw/str_simd.c
krizhanovsky: Yes, this is exactly what the compiler does this for us.

Issue #996, opened 2018-04-03
Workaround for 50x error codes


krizhanovsky:
Tempesta generates a lot of 50x error responses during tests due to #940 and https://github.com/tempesta-tech/tempesta/issues/488#issuecomment-376980897 . So a temporal workaround is required to ignore the error responses to make CI going. The issue must be closed when the original #940 and #488 are fixed and the workaround is removed, probably in 0.7 only.

krizhanovsky:
The workaround for 0.6 is implemented in https://github.com/tempesta-tech/tempesta-test/pull/12 .

PR #997, opened 2018-04-03
Temporary don't measure 502 and 504 errors

vladtcvs:
fix: #996


krizhanovsky:
The only one note: you specified "Fix #996", but the issue shouldn't be closed after the merge, instead it should remain open and moved to 0.7.

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -167,6 +167,7 @@ def form_command(self):
         threads = self.threads if self.connections > 1 else 1
         self.options.append('-t %d' % threads)
         self.options.append('-c %d' % self.connections)
+        self.options.append('--timeout 5')
```
ikoveshnikov: As discussed in chat, estimated request timeout depends on number of concurrent connections. Default of `2` is great for <100 connections, while some requests may be served more than 2 seconds with >1000 connections. But there is no straight dependency and exact required value is not clear. I believe the value must be synchronized with `server_forward_timeout` which is `60` by default.

Note that increasing serving time is not Tempesta issue, it depends on backends and wrk connections too. E.g. overall wrk performance decreases while connection count increases. 

Please, also add a comment for this change. If another benchmarker will be used in tests (TfwBomber is expected), the issue must be taken into account.

tempesta_fw/t/functional/helpers/control.py
```diff
@@ -198,6 +199,10 @@ def parse_out(self, stdout, stderr):
                               m.group(1))
             self.errors += (int(err_m.group(1)) + int(err_m.group(2))
                             + int(err_m.group(3)) + int(err_m.group(4)))
+            self.statuses["connect_error"] = int(err_m.group(1))
```
ikoveshnikov: Hard to say something on this change. Wrk-specific counters are exposed for use in generic code. It's not good, but workaround is required and a new tests structure is planned for the nearest future. So ok, but add a comment and link it with #996

Issue #919, opened 2018-02-20
Waiting for port free in tests


vladtcvs:
When we run multiple tests, sometimes we run next test before ports, used in previous test, releases. So we have fails of running nginx or other server.

Reproducing: run_tests.py

`error: [Errno 98] Address already in use`

krizhanovsky:
Can be easily done with sysctl, see for example http://www.linuxbrigade.com/reduce-time_wait-socket-connections/

PR #999, opened 2018-04-04
Set tcp fin timeout

vladtcvs:
fix: #919

### DIFF
--- a/tempesta_fw/t/functional/run_tests.py
+++ b/tempesta_fw/t/functional/run_tests.py
@@ -154,6 +154,17 @@ if root_required:
 
 
 remote.connect()
+
+# configuring TCP for faster reuse the same TCP ports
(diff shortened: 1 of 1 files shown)


tempesta_fw/t/functional/run_tests.py
```diff
@@ -154,6 +154,17 @@ def usage():
 
 
 remote.connect()
+
+# configuring TCP
```
krizhanovsky: Since you're modifying tcp sysctls, it's obvious that you configure TCP. It's better to write a more programmer-friendly comment explaining that you make TCP sockets recycle quicker to make next tests run using the same TCP ports.

vladtcvs: fixed

vladtcvs:
Rebased above master branch.

PR #1000, opened 2018-04-04
Fix #166: transition to standard skb linkage inside TempstaFW.

### LINKED ISSUE #166: Using SKB headroom for SsSkbCb
Hello,

Is it possible to make TFW as a LKM without needing of kernel prepatching? As I see it, the patch (apart from other things) extends `skb->cb` so `SsSkbCb` resides on it. But taking into account that the SKBs has some headroom groving with increasing of the OSI's layer seems that it is possible to avoid SKB's `cb` area to extend.

```
#define TFW_MAGIC   0x54454d5045535441ULL

typedef struct {
    struct sk_buff  *next;
    struct sk_buff  *prev;
    unsigned long   magic;
} SsSkbCb;

#define TFW_SKB_CB(s)       ((SsSkbCb *)(s)->head)

static inline bool ss_skb_passed(const struct sk_buff *skb)
{
    const SsSkbCb *scb = TFW_SKB_CB(skb);
    return (scb->magic == TFW_MAGIC && (scb->next && scb->prev));
}
```

Any reasons why is it can't be implemented?

aleksostapenko:


linux-4.9.35.patch
```diff
@@ -1,5 +1,5 @@
 diff --git a/Documentation/kernel-parameters.txt b/Documentation/kernel-parameters.txt
```
krizhanovsky: Please remove the patch from the master and generate a new patch from https://github.com/tempesta-tech/linux-4.14.32-tfw .

tempesta_fw/ss_skb.c
```diff
@@ -236,14 +236,11 @@ __it_next_data(struct sk_buff *skb, int i, TfwStr *it)
...
-	nscb->prev = skb;
-	scb->next = nskb;
-	if (nscb->next)
-		TFW_SKB_CB(nscb->next)->prev = nskb;
+	nskb->next = skb->next;
+	nskb->prev = skb;
+	skb->next = nskb;
+	if (nskb->next)
+		nskb->next->prev = nskb;
 }
```
krizhanovsky: The function must be removed since it does just the same as standard `list_add()` except it breaks the standard list behaviour using NULL as a tail pointer. The point of #166 is that our internal skb lists mut be the same as in the kernel.

tempesta_fw/ss_skb.c
```diff
@@ -253,11 +250,11 @@ __skb_insert_after(struct sk_buff *skb, struct sk_buff *nskb)
 static inline void
 __skb_skblist_fixup(SsSkbList *skb_list)
```
krizhanovsky: I believe we can and should remove `SsSkbList` now. Regarding `__skb_skblist_fixup()` the only place where we use it is `skb_fragment()` and it seems that we used the function just because we had no circular list, so the function seems also must be removed.

tempesta_fw/ss_skb.c
```diff
@@ -236,14 +236,11 @@ __it_next_data(struct sk_buff *skb, int i, TfwStr *it)
 static inline void
 __skb_insert_after(struct sk_buff *skb, struct sk_buff *nskb)
```
krizhanovsky: I believe all our own list stuff like `ss_skb_queue_head_init()`, `ss_skb_queue_empty()`, `ss_skb_queue_tail()` and so on must be removed now.

tempesta_fw/ss_skb.h
```diff
@@ -71,39 +71,60 @@ ss_skb_queue_empty(const SsSkbList *list)
...
+ */
 static inline void
-ss_skb_unlink(SsSkbList *list, struct sk_buff *skb)
+ss_skb_init_from_frag_list(SsSkbList *list, struct sk_buff *skb)
+{
+	WARN_ON_ONCE(skb->next || skb->prev);
+
+	skb->next = skb_shinfo(skb)->frag_list;
+	list->first = list->last = skb;
+}
```
krizhanovsky: Well, `skb_shinfo(skb)->frag_list` is just a pointer to `skb` - the proof that we don't need `SsSkbList` any more. Since the function will work just with plain `skb` pointers, just 2 lines in size and is used only once, there is no need to keep the function and the code can be directly mode to `ss_skb_unroll()`.

tempesta_fw/ss_skb.h
```diff
@@ -71,39 +71,60 @@ ss_skb_queue_empty(const SsSkbList *list)
...
+ * no need to set @next pointer for each skb as these pointers are
+ * properly set in @frag_list).
+ */
+static inline void
+ss_skb_move_from_frag_list(SsSkbList *list, struct sk_buff *skb)
 {
-	SsSkbCb *scb = TFW_SKB_CB(skb);
+	skb->prev = list->last;
+	list->last = skb;
+}
```
krizhanovsky: The same as above for `ss_skb_init_from_frag_list()`.

tempesta_fw/ss_skb.h
```diff
@@ -71,39 +71,60 @@ ss_skb_queue_empty(const SsSkbList *list)
...
+		list->first = skb->next;
+	skb->next = skb->prev = NULL;
 }
 
 static inline struct sk_buff *
 ss_skb_next(struct sk_buff *skb)
 {
-	return TFW_SKB_CB(skb)->next;
+	return skb->next;
 }
```
krizhanovsky: These changes and at the below (`ss_skb_next()`, `ss_skb_peek_tail()` and so on) - please remove the stuff and use standard `skbuff` and `list` API.

linux-4.9.35.patch
aleksostapenko: Done.

tempesta_fw/ss_skb.c
aleksostapenko: `SsSkbList` is replaced with simple `struct sk_buff *skb_head` pointer to NULL-terminated list with additional tail pointer (`skb_head->prev`).

tempesta_fw/ss_skb.h
aleksostapenko: Done.

tempesta_fw/ss_skb.h
aleksostapenko: Some functions had been removed (https://github.com/tempesta-tech/tempesta/pull/1000#discussion_r180079268) but we cannot remove all our skb list API, as standard kernel skbuff list API is not applicable for our internal skb queue.

tempesta_fw/ss_skb.c
aleksostapenko: Our current internal implementation is based and largely tied with NULL-terminated lists, so we cannot use full circular lists without significant redesign of code which work  with our internal skb queue.

tempesta_fw/ss_skb.c
```diff
@@ -236,28 +236,25 @@ __it_next_data(struct sk_buff *skb, int i, TfwStr *it)
...
 {
-	SsSkbCb *lscb = TFW_SKB_CB(skb_list->last);
+	struct sk_buff *last = skb_head->prev;
 
-	if (lscb->next)
-		skb_list->last = lscb->next;
-	WARN_ON_ONCE(TFW_SKB_CB(skb_list->last)->next);
+	if (last->next)
+		skb_head->prev = last->next;
+	WARN_ON_ONCE(skb_head->prev->next);
```
krizhanovsky: I still don't see the reason for the function and why can't we move to double linked list if we need `last` pointer or accurate NULL terminated list like `skb->frag_list`: current `prev` meaning the last item as well as double linked list with NULLs for the last and first items look ugly and error prone.

tempesta_fw/ss_skb.h
```diff
@@ -59,35 +58,28 @@ ss_skb_queue_tail(struct sk_buff **skb_head, struct sk_buff *skb)
...
-	/* Tail skb - set tail pointer to the new tail skb. */
-	if ((*skb_head)->prev == skb)
-		(*skb_head)->prev = skb->prev;
-
-	/* If this is head skb, set head pointer to the new head skb. */
+	}
+	/* If this is head skb and not last, set head to the next skb. */
 	if (*skb_head == skb)
 		*skb_head = skb->next;
-	else
```
krizhanovsky: `*skb_head == skb` should be in the else branch - there is no sense to make the check after `*skb_head = NULL`

tempesta_fw/main.c
```diff
@@ -341,7 +341,7 @@ static struct ctl_table tfw_sysctl_tbl[] = {
 		.mode		= 0644,
 		.proc_handler	= tfw_ctlfn_state_io,
 	},
-	{ 0 }
+	{}
```
ikoveshnikov: Although GCC complies the code, the statement is valid for C++ and prohibited for C.  `{ 0 }` is more correct for C.

tempesta_fw/filter.c
```diff
@@ -298,7 +300,8 @@ tfw_filter_stop(void)
 	if (tfw_runstate_is_reconfig())
 		return;
 	if (ip_filter_db) {
-		nf_unregister_hooks(tfw_nf_ops, ARRAY_SIZE(tfw_nf_ops));
+		nf_unregister_net_hooks(&init_net, tfw_nf_ops,
+					ARRAY_SIZE(tfw_nf_ops));
```
krizhanovsky: Why `init_net` and why don't you add the hooks to all network namespaces as other security modules do this using `register_pernet_subsys()`?

tempesta_fw/cfg.c
```diff
@@ -1612,8 +1612,8 @@ tfw_cfg_read_file(const char *path, size_t *file_size)
 	do {
 		TFW_DBG3("read by offset: %d\n", (int)offset);
 		read_size = min((size_t)(buf_size - offset), PAGE_SIZE);
-		bytes_read = vfs_read(fp, out_buf + offset, read_size, \
-				      &offset);
+		bytes_read = kernel_read(fp, out_buf + offset, read_size, \
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/filter.c
aleksostapenko: Corrected.

Issue #857, opened 2017-11-09
[Func tests] Check required binaries before runnng tests


ikoveshnikov:
As functional tests becoming more complicated and complete they require more and more external tools. Functional tests suite must check all the required tools before start to prevent false tests failures. Some required tools are not documented in readme for functional tests, e.g. curl and systemtap.

krizhanovsky:
Guys, it seems the testing system becomes quite large and complex. What about to move it to separate repository? @intelfx , @ikoveshnikov any opinions?

intelfx:
That's gonna introduce some pretty complex problems with correlating revisions of code and tests. I intended to test (at least smoke-test) every incoming revision (not just the latest of each PR), and this will make it impossible.

ikoveshnikov:
I think, that the question depends on CI only. There is no strong dependencies between the code and the functional tests. I know only one dependency - Tempesta's version string, since it's evaluated during parsing of Tempesta's replies.

The functional tests are not aware about implementation  at all. If a test passes on one revision of Tempesta and fails on the other, that means that the tested feature is broken or not implemented on the latter one. You just can't break functional tests by changing the Tempesta's code.

But there is a problem, that you can't say that the test itself is error-free, until the feature is fully implemented (or a bug is fixed). So yes, there is some correlating issue.

It's usual approach to separate functional tests from the code. I'm for the separation, but as i said it depends on CI system.

krizhanovsky:
Guys, we  can require, somewhere in installation docs, that the test repository/directory must reside in the same directory with tempest source code and built module, so technically tests just move few directories up.

vladtcvs:
https://github.com/tempesta-tech/tempesta-test/pull/30

vladtcvs:
Fixed in tests PR#30

PR #1001, opened 2018-04-05
Move TCP config to helpers/prepare.py

vladtcvs:


tempesta_fw/t/functional/helpers/prepare.py
```diff
@@ -0,0 +1,13 @@
+__author__ = 'Tempesta Technologies, Inc.'
+__copyright__ = 'Copyright (C) 2018 Tempesta Technologies, Inc.'
+__license__ = 'GPL2'
+
+from . import remote
+
+def configure_tcp():
+    """ configuring TCP for faster reuse the same TCP ports """
```
ikoveshnikov: Need to comment why the long reuse is the issue. It'll be hard to remember in a few weeks.

Also `configure_tcp()` is just one single step in whole preparing procedure, and #857 is also required. Please, add high level prepare function and add TODO for #857

ikoveshnikov:
Previously I've added code into `run_tests.py` which checks whether root privileges are required. What do you think, the check is extra now, and root privileges are always required? 
Do we have any other code in `run_tests.py` which better to move into `prepare`?

vladtcvs:
I think, we should place running as root into remote.py methods. In this way we can run run_tests.py itself not from root, but remote commands will run as root

tempesta_fw/t/functional/helpers/prepare.py
vladtcvs: No need to write TODO. I will add this checks when will do #857

tempesta_fw/t/functional/helpers/prepare.py
```diff
@@ -0,0 +1,21 @@
+__author__ = 'Tempesta Technologies, Inc.'
+__copyright__ = 'Copyright (C) 2018 Tempesta Technologies, Inc.'
+__license__ = 'GPL2'
+
+from . import remote
+
+def configure_tcp():
+    """ Configuring TCP for faster reuse the same TCP ports.
+    We start a lot of backends and clients in stress tests.
+    And in next stress test we sometimes failing using the
```
ikoveshnikov: Better to avoid `we` or `I` if possible.
'We start a lot of backends and clients in stress tests. And in next stress test we sometimes failing using the same ports.' -> 'A lot of sockets are created in tests and bound to specific ports. Release them quicker to reuse the ports in the next test case.'

PR #1002, opened 2018-04-05
Temporary workaround: don't assert on wrk socket errors

vladtcvs:


tempesta_fw/t/functional/helpers/control.py
```diff
@@ -122,7 +122,7 @@ class Wrk(Client):
     If FAIL_ON_SOCK_ERR is set assert that none of such errors happened during
     test, otherwise print warning and count the errors as usual errors.
     """
-    FAIL_ON_SOCK_ERR=True
+    FAIL_ON_SOCK_ERR=False
```
ikoveshnikov: OK, it's not a fix but a workaround. Don't really think it's a good solution, see https://github.com/tempesta-tech/tempesta/pull/997#discussion_r179499076
I assume that the whole `FAIL_ON_SOCK_ERR` thing should be reworked. `wrk` can report on 4 different types of socket errors: connect, read write, and timeout. The message 'Socket errors on wrk. Too many concurrent connections?' showed for _any_ of this errors is incorrect: it applies only for the first one. This issue is not yours, it's mine actually, but now we have time to figure out how to fix it better.

`connect` errors - Basically it means that the test environment is configured with errors.
It seems, that `timeout` errors are caused by different `server_forward_timeout` tempesta configuration and `--timeout ` wrk settings.
`read` and `write` error must be checked further.

PR #1005, opened 2018-04-11
Fix #628: Custom characters set for URI and HTTP headers

### LINKED ISSUE #628: Custom characters set for URI and HTTP headers

krizhanovsky:
PoC and benchmarks at https://github.com/natsys/blog/commit/0fddf424c7d76093131f18458cf3085311c9246c

### DIFF
--- a/tempesta_fw/cfg.c
+++ b/tempesta_fw/cfg.c
@@ -1104,6 +1104,93 @@ tfw_cfg_parse_uint(const char *s, unsigned int *out_uint)
 }
 EXPORT_SYMBOL(tfw_cfg_parse_uint);
 
+/**
+ * Borrowed from linux/lib/kstrtox.c because the function isn't exported by
+ * the kernel.
+ */
+#define KSTRTOX_OVERFLOW	(1U << 31)
+
+static unsigned int
+_parse_integer(const char *s, unsigned int base, unsigned long *p)
+{
+	unsigned long res;
+	unsigned int rv;
+
+	res = 0;
+	rv = 0;
+	while (1) {
+		unsigned int c = *s;
+		unsigned int lc = c | 0x20; /* don't tolower() this line */
+		unsigned int val;
+
+		if ('0' <= c && c <= '9')
+			val = c - '0';
+		else if ('a' <= lc && lc <= 'f')
+			val = lc - 'a' + 10;
+		else
+			break;
+
+		if (val >= base)
+			break;
+		/*
+		 * Check for overflow only if we are within range of
+		 * it in the max base we support (16)
+		 */
+		if (unlikely(res & (~0ull << 60))) {
+			if (res > div_u64(ULLONG_MAX - val, base))
+				rv |= KSTRTOX_OVERFLOW;
+		}
+		res = res * base + val;
+		rv++;
+		s++;
+	}
+	*p = res;
+
+	return rv;
+}
+
+int
+tfw_cfg_parse_intvl(const char *s, unsigned long *i0, unsigned long *i1)
+{
+	unsigned long *v = i0;
+	unsigned int r;
+	int base;
+
+	while (*s) {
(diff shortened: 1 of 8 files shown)


tempesta_fw/http.c
```diff
@@ -3887,6 +4044,48 @@ static TfwCfgSpec tfw_http_specs[] = {
 		.allow_none = true,
 		.cleanup = tfw_cfgop_cleanup_whitelist_mark,
 	},
+	{
+		.name = "http_uri_brange",
```
ikoveshnikov: Just a notice, no fix is required. This class of directives is not about http processing, it's more about parsing and string operations. But I can't say that moving the directives  to `http_parser.c` or `str_simd.c` will make huge difference.

tempesta_fw/cfg.c
```diff
@@ -1104,6 +1104,93 @@ tfw_cfg_parse_uint(const char *s, unsigned int *out_uint)
...
+		}
+		if (v == i1 && *i0 >= *i1) {
+			TFW_ERR("Interval bound crossing\n");
+			return -EINVAL;
+		}
+
+		s += r;
+	}
+	if (v == i1 && !*v) {
+		TFW_ERR_NL("Zero interval left bound in '%s'\n", s);
```
aleksostapenko: It seems that 's' will always be `0` in `TFW_ERR_NL` output here.

tempesta_fw/http.c
```diff
@@ -3863,6 +3863,163 @@ tfw_cfgop_cleanup_whitelist_mark(TfwCfgSpec *cs)
...
+	if (ce->attr_n) {
+		TFW_ERR_NL("Unexpected attributes\n");
+		return -EINVAL;
+	}
+
+	TFW_CFG_ENTRY_FOR_EACH_VAL(ce, i, val) {
+		unsigned long i0 = 0, i1 = 0;
+
+		if (tfw_cfg_parse_intvl(val, &i0, &i1)) {
+			TFW_ERR_NL("Cannot parse brange_uri interval: '%s'\n",
```
aleksostapenko: This function is called for different branges, but in case of error it will always produce `...brange_uri...` log output.

tempesta_fw/t/unit/test_tfw_str.c
```diff
@@ -178,6 +183,95 @@ TEST(cstr, simd_match_ctext_vchar)
...
+		    || i == 0x7E || i == 0x98)
+			a[i] = 1;
+	}
+
+	/*
+	 * Set custom alphabet to URI character set plus x98, but use xff
+	 * matcher just because it's different from URI and CTEXT+VCHAR to
+	 * enforce custom matching.
+	 */
+	tfw_init_custom_xff(a, true);
```
aleksostapenko: Custom `xff` alphabet is set here but not disabled after the test, so tests from `test_http_parser.c` (which started after `tfw_str` tests) with `X-Forwarded-For` header - are failed.

PR #1007, opened 2018-04-23
Fix #635

### LINKED ISSUE #635: AVX2 memcpy(), memcmp() & Ko
Linux kernel uses dummy x86 assembly for `memcpy()`, `memcmp()` etc. Meantime Tempesta FW and Tempesta DB copy small data using the routines. Example of the data is HTTP headers - usually they small, but some of them (like Cookie) can reach enormous size and we have to copy them to be able to adjust.

Actually with #634 in mind we should not copy HTTP headers, but we still need to copy some small data.

krizhanovsky:
Please read https://github.com/tempesta-tech/tempesta/issues/635#issuecomment-383745388 for some explanations. The SIMD functions must be written in assembly since gcc requires `stdlib.h`. For now we use `tls/dummy_headers/stdlib.h` and it's good to remove the dummy headers, so `str_simd.c` also should be rewritten in assembly soon. You can just compile the PoCs with `-S` to get assembly code which I used. I din't use `vzeroupper` instruction since we save and restore FPU state in softirq and each SIMD function in our code initializes the vector registers.

The change is questionable and I'm curious for you opinions. The change introduces ~1600-line kernel patch with simple replacement for `mem{cpy,set,cmp}()` functions by their `_fast` versions. Meantime performance improvements are negligible for pure Linux networking and say some user space daemons, e.g. HAProxy + Nginx - perf shows interrupts (due to poor e1000 in particular) and user-space daemon functions. We get something only for Tempesta FW caching where we have no heavy user space logic and do some our internal `mem{cpy,set,cmp}()` operations. So probably there is no sense to introduce the large kernel patch. I just was confused by large zeroings and copyings for skbs, e.g. the data structure initialization or cloning. However, probably we can optimize the calls and there is no sense to pay extra if we have a faster implementation....

@vladtcvs could you do a performance study on our hardware testbed to learn whether the patch gives us more performance improvement with good adapters coalescing network interrupts?


ikoveshnikov:
Please, replace another `memcmp` occurrence with `memcmp_fast`: https://github.com/tempesta-tech/tempesta/blob/f571516baea08b7713e597edbba8c7ca20b4d109/tempesta_fw/str.c#L418-L420

Issue #883, opened 2018-01-14
Functional test for HTTP tables


krizhanovsky:
Need a test for HTTP tables. 

One special case must be covered by the test. HTTP tables (#731) make possible to load balance traffic just using `mark`. So test a scenario of several, say 10, server groups and distribute traffic among them by value of `mark` (from 1 to 10 for example).

aleksostapenko:
Resolved in https://github.com/tempesta-tech/tempesta-test/pull/22.

PR #1017, opened 2018-05-15
Fix #731: HTTP tables introduction.

### LINKED ISSUE #731: Multi-layer firewall
This is a first step towards [Tempesta Language](https://github.com/tempesta-tech/tempesta/issues/102) and should be considered as a foundation for it. As in [the TL task](https://github.com/tempesta-tech/tempesta/issues/102) we should be able to specify multi-layer, IP addresses and HTTP header values only for now, firewall rule. For example, to filter our all clients with particular IP and a value of User-Agent, Referer or the sticky cookie.

Since we have no compiler and a runtime for now `tdbq` or config file must be used, so the issue depends on [hot configuration reloading](https://github.com/tempesta-tech/tempesta/issues/51) and [TDB DML & queries](https://github.com/tempesta-tech/tempesta/issues/279). The rules must be stored in TDB by IP address as a key (just use ANY ADDR if no IP was specified for a rule) and be represented as a smaller table indexed by HTTP header identifiers. The filtering algorithms should be like the following (for AND operator, concatinating IP rule and HTTP rule, OR is trivial since it's essencially just a 2 independent rules):

1. at filter (IP) layer, when an IP packet arrives,

aleksostapenko:


tempesta_fw/sched/tfw_sched_http.c
```diff
@@ -85,46 +112,77 @@ MODULE_DESCRIPTION("Tempesta HTTP scheduler");
...
  *
- * The search is based on contents of an HTTP request and match rules
- * that specify which Vhost the request should be forwarded to.
+ * The search is based on contents of an HTTP request and
+ * match http chain rules that specify which virtual host
+ * or other http chain the request should be forwarded to.
  */
 static TfwVhost *
-tfw_sched_http_sched_vhost(TfwMsg *msg)
+tfw_sched_http_sched_vhost(TfwMsg *msg, bool *block)
```
ikoveshnikov: For me no target vhost found means that request must be filtered. In current `http_chain` definition `block` is defined as default action. So rule `block` must be used if no target vhost found and variable `vhost` must be set to true. That means, that `block` variable duplicates return value. But the function works different: it doesn't set variable `block` to true, if no vhost found and no explicit `block` action encountered.

I see, why you've introduced `block` variable. Request without target vhost can be filtered in two ways: blocking of entire client connection or dropping a single request. I believe, HTTP scheduler is wrong place to make a that decision, it must be done in `http.c`. But let's look to the problem. Some clients may use proxies to reach TempestaFW, one client connection can multiplicate a lot of end clients. Some of them malicious, but most of them - not. We don't want to block good clients for others deeds, so dropping of single requests seems more reasonable for me. We can still block connections with lot of malicious requests using `http_resp_code_block` frang limit.

tempesta_fw/http.c
```diff
@@ -2596,12 +2600,23 @@ tfw_http_req_add_seq_queue(TfwHttpReq *req)
 static int
 tfw_http_req_set_context(TfwHttpReq *req)
 {
-	if (!(req->vhost = tfw_vhost_match((TfwMsg *)req)))
-		return -EINVAL;
+	bool block = false;
```
ikoveshnikov: See my comment for `tfw_sched_http_sched_vhost()`

tempesta_fw/sched.c
```diff
@@ -60,7 +60,7 @@ tfw_sched_get_vhost(TfwMsg *msg)
 	list_for_each_entry_rcu(sched, &sched_list, list) {
 		if (!sched->sched_vhost)
 			break;
-		if ((vhost = sched->sched_vhost(msg))) {
+		if ((vhost = sched->sched_vhost(msg, block))) {
```
ikoveshnikov: It was said in one of the last reviews by @krizhanovsky : http (vhost) scheduler isn't one of possible schedulers and it's API is far away of hash and rr schedulers. Foreach cycle must be replaced with direct http (vhost) scheduler call. This wasn't done in this PR, please add issue for it.

tempesta_fw/sched/tfw_sched_http.c
```diff
@@ -33,24 +35,49 @@
...
+ *   }
+ *   vhost ws2 {
+ *       proxy_pass webapp_site2;
+ *       ...
+ *   }
+ *   http_chain base {
+ *       ...
+ *       mark != 1 -> storage;
+ *   }
+ *   http_chain {
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/http_match.h
```diff
@@ -61,20 +82,41 @@ typedef enum {
...
+
 typedef struct {
 	struct list_head	list;
 	tfw_http_match_fld_t	field; /* Field of a HTTP message to compare. */
+	bool			inv;   /* Comparison inversion (inequality) flag.*/
 	tfw_http_match_op_t 	op;    /* Comparison operator. */
+	TfwHttpAction		act;   /* Rule action. */
 	TfwHttpMatchArg 	arg;   /* A value to be compared with the field.
 					  note: the @arg has variable length. */
 } TfwHttpMatchRule;
```
krizhanovsky: Please arrange members of this structure and the nested structures to minimize memory gaps (e.g. integer enum types should be adjacent where possible, probably it has sense to replace the `bool` fields with integer flags).

tempesta_fw/sched/tfw_sched_http.c
krizhanovsky: It seems a request with not found vhost is dropped in `tfw_http_req_set_context()` with `cannot find vhost for request` and I think it has sense: a vhost isn't configured for the request - maybe it's just an error, not an attack. Anyway we do not pass the request (which also could be malicious) to a backend, so we're good.

tempesta_fw/sched/tfw_sched_http.c
```diff
@@ -195,276 +243,404 @@ tfw_sched_http_cfg_arg_tbl[_TFW_HTTP_MATCH_F_COUNT] = {
...
+		   type == TFW_HTTP_MATCH_A_NUM ||
+		   type == TFW_HTTP_MATCH_A_WILDCARD)
+	{
+		rule->inv = invert;
+		r = tfw_http_rule_init(rule, field, op, type, arg, arg_size - 1);
+		if (r)
+			return r;
+	} else {
+		BUG();
+		// TODO: parsing rules of other types
```
krizhanovsky: Please don't use C++-style for persistent comments.

Why don't you parse the other types?

tempesta_fw/http_match.h
```diff
@@ -37,6 +57,7 @@ typedef enum {
 	TFW_HTTP_MATCH_F_HOST,
 	TFW_HTTP_MATCH_F_METHOD,
 	TFW_HTTP_MATCH_F_URI,
+	TFW_HTTP_MATCH_F_MARK,
 	_TFW_HTTP_MATCH_F_COUNT
 } tfw_http_match_fld_t;
```
krizhanovsky: Comment from #731 
> Currently not all special headers are handled in tfw_http_match_fld_t, e.g. it doesn't handle X-Forwarded-For, User-Agent or Cookie.

Isn't  addressed

tempesta_fw/sock.c
```diff
@@ -364,9 +365,12 @@ ss_do_send(struct sock *sk, struct sk_buff **skb_head, int flags)
 
 		ss_skb_init_for_xmit(skb);
 
-		TFW_DBG3("[%d]: %s: entail skb=%p data_len=%u len=%u\n",
-		         smp_processor_id(), __func__,
-		         skb, skb->data_len, skb->len);
+		/* Propagate mark of message head skb.*/
+		skb->mark = mark;
```
krizhanovsky: Please extend #883 by a requirement to test the mark iteroperability between Tempesta FW and iptables/nftables in both the directions (or just write the test on your own). Also add an example with iptables/nftables to https://github.com/tempesta-tech/tempesta/wiki/HTTP-tables

tempesta_fw/cfg.c
```diff
@@ -202,23 +212,59 @@ entry_reset(TfwCfgEntry *e)
...
+	}
+
+	TFW_DBG3("set name: %.*s\n", len, name);
+
+	if (!check_identifier(name, len))
 		return -EINVAL;
 
-	if (!check_identifier(name_src, name_len))
+	if (!rule) {
+		e->name = e->ftoken;
```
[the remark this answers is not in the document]
aleksostapenko: `e->ftoken` is temporally assigned to local `name` variable (not to `e->name`) for uniform checking of identifier.

tempesta_fw/http.c
```diff
@@ -2596,12 +2600,23 @@ tfw_http_req_add_seq_queue(TfwHttpReq *req)
...
+	}
+	if (block) {
+		TFW_INC_STAT_BH(clnt.msgs_filtout);
+		tfw_client_block(req, 500, "request has been filtered out"
+				 " via http table");
+	} else {
+		TFW_INC_STAT_BH(clnt.msgs_otherr);
+		tfw_client_drop(req, 500, "cannot find vhost for request");
+	}
+	return -EINVAL;
```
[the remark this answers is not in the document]

tempesta_fw/http_match.c
```diff
@@ -345,47 +349,65 @@ static const match_fn match_fn_tbl[_TFW_HTTP_MATCH_F_COUNT] = {
...
+	BUG_ON(rule->arg.len < 0 || rule->arg.len >= TFW_HTTP_MATCH_MAX_ARG_LEN);
 
 	field = rule->field;
 	match_fn = match_fn_tbl[field];
 	BUG_ON(!match_fn);
 
-	return match_fn(req, rule);
+	matched = match_fn(req, rule);
+	if (rule->inv)
+		matched = !matched;
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/sched.c
aleksostapenko: Done.

tempesta_fw/sched/tfw_sched_http.c
```diff
@@ -5,7 +5,8 @@
  *
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/http_match.c
```diff
@@ -1,39 +1,36 @@
...
+ *  TFW_HTTP_MATCH_F_URI    => match_uri
  *  etc...
  * Each such match_*() function takes TfwHttpReq and TfwHttpMatchRule and
  * returns true if the given request matches to the given rule.
  * Such approach allows to keep the code structured and eases adding new
- * @field/@op combinations.
+ * @field types.
  * Currently that is implemented with a multi-dimensional array of pointers
  * (the match_fn_tbl). However the code is critical for performance, so perhaps
  * this may be optimized to a kind of jump table.
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/http_match.c
```diff
@@ -345,47 +349,65 @@ static const match_fn match_fn_tbl[_TFW_HTTP_MATCH_F_COUNT] = {
...
 	BUG_ON(!req || !rule);
-	BUG_ON(rule->field < 0 || rule->field >= _TFW_HTTP_MATCH_F_COUNT);
-	BUG_ON(rule->op < 0 || rule->op >= _TFW_HTTP_MATCH_O_COUNT);
-	BUG_ON(rule->arg.type < 0 || rule->arg.type >= _TFW_HTTP_MATCH_A_COUNT);
-	BUG_ON(rule->arg.len <= 0 || rule->arg.len >= TFW_HTTP_MATCH_MAX_ARG_LEN);
+	BUG_ON(rule->field <= 0 || rule->field >= _TFW_HTTP_MATCH_F_COUNT);
+	BUG_ON(rule->op <= 0 || rule->op >= _TFW_HTTP_MATCH_O_COUNT);
+	BUG_ON(rule->act.type <= 0 || rule->act.type >= _TFW_HTTP_MATCH_ACT_COUNT);
+	BUG_ON(rule->arg.type <= 0 || rule->arg.type >= _TFW_HTTP_MATCH_A_COUNT);
+	BUG_ON(rule->arg.len < 0 || rule->arg.len >= TFW_HTTP_MATCH_MAX_ARG_LEN);
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/http_match.h
```diff
@@ -27,6 +27,26 @@
 #include "pool.h"
 #include "http.h"
 
+/**
+ * HTTP chain. Contains list of rules for matching.
+ */
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/http_match.h
```diff
@@ -27,6 +27,26 @@
...
+	struct list_head list;
+	struct list_head mark_list;
+	struct list_head match_list;
+	const char *name;
+	TfwPool *pool;
+} TfwHttpChain;
+
+/**
+ * HTTP table. Contains list of HTTP chains.
+ */
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/http_match.c
```diff
@@ -394,77 +416,139 @@ tfw_http_match_req(const TfwHttpReq *req, const TfwHttpMatchList *mlst)
...
+{
+	size_t len;
+	const char *arg = *arg_out;
+
+	BUG_ON(!arg);
+	len = strlen(arg);
+	*op_out = TFW_HTTP_MATCH_O_EQ;
+	*size_out = len + 1;
+
+	if (arg[len - 1] == '*') {
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/http_match.h
aleksostapenko: Corrected.

tempesta_fw/sched/tfw_sched_http.c
aleksostapenko: `Addr` and `method` types are not used in TempestaFW.
Corrected.

tempesta_fw/http_match.h
aleksostapenko: Done.

tempesta_fw/sched/tfw_sched_http.c
aleksostapenko: Done: connection dropped in case of `block` action; request dropped in case of `vhost` miss.

etc/tempesta_fw.conf
```diff
@@ -259,8 +259,18 @@
 #       The value of the "Host" header field.
 #   - hdr_conn
 #       The value of the "Connection" header field.
+#   - hdr_ctype
```
ikoveshnikov: Not the issue of this PR, but it seems that all that `hdr_ctype`, `hdr_agent` and Co is not very user friendly. In the same time we have `hdr_raw`. It would better to provide single directive for all headers:
```
http_chain NAME {
	[ hdr "HDR_NAME" == (!=) ARG ] -> ACTION [ = VAL];
	...
}
```
Like it's done for `hdr_raw` now. And give a notice that comparing of the `Content-Type`, `User-Agent`, .... headers is accelerated, while processing of other headers is slow.

We have `tfw_http_msg_req_spec_hid()` function which can convert header name to it's internal representation in spec headers. So i believe that the change would be easy to implement.

tempesta_fw/http.c
```diff
@@ -2738,21 +2737,25 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
...
-
-		r = tfw_gfsm_move(&conn->state, TFW_HTTP_FSM_REQ_MSG, &data_up);
-		TFW_DBG3("TFW_HTTP_FSM_REQ_MSG return code %d\n", r);
-		/* Don't accept any following requests from the peer. */
-		if (r == TFW_BLOCK) {
+		if (block) {
 			TFW_INC_STAT_BH(clnt.msgs_filtout);
-			tfw_client_block(req, 403, "parsed request"
-				       " has been filtered out");
+			tfw_client_block(req, 500, "request has been filtered"
```
ikoveshnikov: I'd prefer `403 Forbidden` here. It's not server or Tempesta fault that client has requested forbidden location. Think of it from `http_resp_code_block` directive perspective. It worth to block clients that cause many 403 errors, while legitimate users have margin for mistake,

tempesta_fw/http.c
```diff
@@ -2738,21 +2737,25 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
...
-		if (tfw_http_req_set_context(req)) {
-			TFW_INC_STAT_BH(clnt.msgs_otherr);
-			tfw_client_drop(req, 500, "cannot find"
-				      "Vhost for request");
-			return TFW_BLOCK;
+		/* Assign the right Vhost for current request. */
+		if ((req->vhost = tfw_http_tbl_vhost((TfwMsg *)req, &block))) {
+			req->location = tfw_location_match(req->vhost,
+							   &req->uri_path);
+			r = tfw_gfsm_move(&conn->state, TFW_HTTP_FSM_REQ_MSG,
```
ikoveshnikov: The hook is not called if vhost is not found. That is not right. Need to run the hook even if vhost is not found, default vhost and default  location will be used as context.

Currently only frang is behind that hook, it still may want to block the client even if no vhost found.

tempesta_fw/sock.c
aleksostapenko: Wiki has been updated with iptables example.

etc/tempesta_fw.conf
krizhanovsky: I created issue #1033

tempesta_fw/http.c
krizhanovsky: Yeah, good catch - this must be fixed before the merge.

tempesta_fw/http.c

tempesta_fw/http.c

tempesta_fw/http_match.c
```diff
@@ -394,77 +434,169 @@ tfw_http_match_req(const TfwHttpReq *req, const TfwHttpMatchList *mlst)
...
+
+	if (!(arg_out = kzalloc(len + 1, GFP_KERNEL))) {
+			TFW_ERR_NL("HTTP tables: unable to allocate rule"
+				   " argument.\n");
+		return NULL;
+	}
+
+	*op_out = TFW_HTTP_MATCH_O_EQ;
+	if (len > 1 && arg[len - 1] == '*' && arg[len - 2] != '\\')
+			*op_out = TFW_HTTP_MATCH_O_PREFIX;
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

PR #1024, opened 2018-05-23
Fix skb leakage on HTTP request processing errors.

krizhanovsky:
tfw_http_resp_process() calls ss_skb_queue_purge() on bad_msg label, however
tfw_http_req_process() on TFW_BLOCK returned from the parser calls
tfw_client_drop() which not always free the request. (The ingress skb is
linked with the request in tfw_http_msg_process(), so we have to call
ss_skb_queue_purge() even on not parsed request.)

There is a case when we do not free skb on HTTP request processing errors.
tfw_http_cli_error_resp_and_log() -> tfw_http_req_mark_error()
-> __tfw_http_send_resp(): on bad return code from tfw_strcpy_desc()
just returns w/o calling tfw_http_resp_build_error().

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -2259,7 +2259,6 @@ tfw_http_resp_fwd(TfwHttpResp *resp)
 	}
 	if (!req_retent) {
 		spin_unlock_bh(&cli_conn->seq_qlock);
-		tfw_http_resp_pair_free(resp->req);
 		return;
 	}
 	__list_cut_position(&ret_queue, seq_queue, req_retent);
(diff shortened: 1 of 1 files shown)


tempesta_fw/http.c
```diff
@@ -2259,6 +2259,7 @@ tfw_http_resp_fwd(TfwHttpResp *resp)
 	}
 	if (!req_retent) {
 		spin_unlock_bh(&cli_conn->seq_qlock);
+		tfw_http_resp_pair_free(resp->req);
```
keshonok: Are you convinced that this is the right thing to do?

Here's my understanding. `req_retent` is not set when there're no consecutive responses (from the start of `seq_queue`) that are ready to be sent out to a client. That's a perfectly valid case. The current `resp` has been marked as ready and it's waiting for its turn until the 'slots' before it are filled. Freeing it (and the paired request) would be incorrect. Moreover, if there's need to delete req/resp pair for some reason, the request needs to be removed from  `seq_queue` first.

krizhanovsky: Yeah, that was just wrong mess in the patch. Thanks for noticing.

PR #1030, opened 2018-05-30
Fix missed connection header

### LINKED ISSUE #1029: Tempesta may send empty 'Connection: ' header
Found by @vladtcvs . If backend treats request as erroneous and closes connection with Tempesta, connection between client and Tempesta is not closed and client receives empty `Connection: ` header. 

Here is an example:
Send request directly to backend:
```
% curl -v -X TRACE http://192.168.122.12:8080/
*   Trying 192.168.122.12...
* TCP_NODELAY set
* Connected to 192.168.122.12 (192.168.122.12) port 8080 (#0)
> TRACE / HTTP/1.1
> Host: 192.168.122.12:8080
> User-Agent: curl/7.60.0
> Accept: */*
> 
< HTTP/1.1 405 Not Allowed
< Server: nginx/1.10.3
< Date: Wed, 30 May 2018 18:04:39 GMT
< Content-Type: text/html
< Content-Length: 173
< Connection: close
< 
<html>
<head><title>405 Not Allowed</title></head>
<body bgcolor="white">
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx/1.10.3</center>
</body>
</html>
* Closing connection 0
``` 

And send the same request to the the Tempesta:
```
% curl -v -X TRACE http://192.168.122.12/
*   Trying 192.168.122.12...
* TCP_NODELAY set
* Connected to 192.168.122.12 (192.168.122.12) port 80 (#0)
> TRACE / HTTP/1.1
> Host: 192.168.122.12
> User-Agent: curl/7.60.0
> Accept: */*
> 
< HTTP/1.1 405 Not Allowed
< Date: Wed, 30 May 2018 18:08:03 GMT
< Content-Type: text/html
< Content-Length: 173
< Connection:                                           <------- Empty header
< Server: Tempesta FW/0.6.0
< Via: 1.1 tempesta_fw (Tempesta FW 0.6.0)
< 
<html>
<head><title>405 Not Allowed</title></head>
<body bgcolor="white">
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx/1.10.3</center>
</body>
</html>
* Connection #0 to host 192.168.122.12 left intact      <------- Connection is not closed 
```

In the second case client receives `Connection: `header, but its empty, and connection is not closed.

I suppose, that Tempesta should close client connection if backend responds with non-200 code and closes connection.

ikoveshnikov:
Fix #1029 

Also found that `TFW_HTTP_MSG_HDR_DEL` macro works incorrect: instead of removing the header, a new header with empty header value is added.

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -2133,8 +2133,23 @@ static int
 tfw_http_adjust_resp(TfwHttpResp *resp)
 {
 	TfwHttpReq *req = resp->req;
-	int r, conn_flg = req->flags & __TFW_HTTP_MSG_M_CONN_MASK;
 	TfwHttpMsg *hm = (TfwHttpMsg *)resp;
+	int r, conn_flg;
+
+	/*
+	 * If request violated backend rules, backend may respond with 4xx code
+	 * and close connection to Tempesta. Don't encourage client to send
+	 * more such requests and cause performance degradation, close the
+	 * client connection.
+	 */
+	if ((resp->flags & TFW_HTTP_F_CONN_CLOSE) && (resp->status / 100 == 4)) {
+		req->flags = (req->flags & ~TFW_HTTP_F_CONN_KA)
+				| TFW_HTTP_F_CONN_CLOSE;
+		conn_flg = TFW_HTTP_F_CONN_CLOSE;
+	}
+	else {
+		conn_flg = req->flags & __TFW_HTTP_MSG_M_CONN_MASK;
+	}
 
 	r = tfw_http_sess_resp_process(resp);
 	if (r < 0)
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -704,7 +704,7 @@ tfw_http_msg_hdr_xfrm(TfwHttpMsg *hm, char *name, size_t n_len,
 		},
 		.len = n_len + SLEN(S_DLM) + v_len,
 		.eolen = 2,
-		.flags = 3 << TFW_STR_CN_SHIFT
+		.flags = (val ? 3 : 2) << TFW_STR_CN_SHIFT
 	};
 
 	BUG_ON(!val && v_len);


tempesta_fw/http.c
```diff
@@ -2133,8 +2133,21 @@ static int
...
 	TfwHttpReq *req = resp->req;
-	int r, conn_flg = req->flags & __TFW_HTTP_MSG_M_CONN_MASK;
 	TfwHttpMsg *hm = (TfwHttpMsg *)resp;
+	int r, conn_flg;
+
+	/*
+	 * Request violated backend rules, so backend closed connection to
+	 * Tempesta. Close client connection after forwarding the response.
+	 */
+	if ((resp->flags & TFW_HTTP_F_CONN_CLOSE) && (resp->status != 200)) {
```
krizhanovsky: I believe not only 200 should be treated as "good" response, at least `304 Not Modified`, `204 No Content`, `206 Partial Content` and I think quite a few others should be treated w/o connection closing. Please check an RFC and/or a reference implementation (e.g. Nginx or Apache) about the cases when a client connection is closed. I assume it should be only for client error (4xx) codes. Please add a comment here about the standard statements.

tempesta_fw/http_msg.c
```diff
@@ -671,7 +671,7 @@ tfw_http_msg_hdr_xfrm_str(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid,
 		return __hdr_add(hm, hdr, hid);
 	}
 
-	if (!s_val)
+	if (!s_val || !s_val->len)
 		return __hdr_del(hm, hid);
```
krizhanovsky: I believe the standard doesn't prohibit empty header values.

tempesta_fw/http.c
krizhanovsky: I forgot to mention that the scenario is pretty trivial, so I wonder why don't we have it in our functional tests? Please also write a functional test which checks all the status codes and connection resets according to the standard.

PR #1036, opened 2018-07-10
Fix #1033: Change header name format in HTTP tables configuration.

### LINKED ISSUE #1033: HttpTables: better header names
Just a copy of https://github.com/tempesta-tech/tempesta/pull/1017#discussion_r194045665 :

Not the issue of this PR, but it seems that all that `hdr_ctype`, `hdr_agent` and Co is not very user friendly. In the same time we have `hdr_raw`. It would better to provide single directive for all headers:
```
http_chain NAME {
	[ hdr "HDR_NAME" == (!=) ARG ] -> ACTION [ = VAL];
	...
}
```
Like it's done for `hdr_raw` now. And give a notice that comparing of the `Content-Type`, `User-Agent`, .... headers is accelerated, while processing of other headers is slow.

We have `tfw_http_msg_req_spec_hid()` function which can convert header name to it's internal representation in spec headers. So i believe that the change would be easy to implement.

If a user specifies a header name which we don't have in special headers, then we shall just search it case-insensitively among raw headers.

aleksostapenko:


tempesta_fw/cfg.h
```diff
@@ -125,6 +125,7 @@
  */
 typedef struct {
 	const char *fst;
+	const char *fst_ext;
```
krizhanovsky: Please write a comment for the field above.

etc/tempesta_fw.conf
```diff
@@ -259,28 +259,14 @@
...
-#       of hdr_raw may be slow as it requires walking over all headers of an
-#       HTTP request.
+#   - hdr
+#       The content of specific HTTP header. In this case HDR_NAME field is
+#       used to specify the name of header; the value of header should be
+#       specified in ARG. Matching of special headers: "X-Forwarded-For",
+#       "If-None-Match", "Referer", "Cookie", "User-Agent", "Content-Type",
+#       "Connection", "Host" - is accelerated; proccessing of other headers
+#       may be slow as it requires walking over all headers of an HTTP
+#       request. Also, suffix OP is not supported for not special headers.
```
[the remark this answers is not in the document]
aleksostapenko: Done. Wiki updated.

tempesta_fw/cfg.h
aleksostapenko: Done.

PR #1037, opened 2018-07-10
Tempesta TLS performance optimizations

### LINKED ISSUE #214: Frang: rework limits for number of chunks
This is a continuation of #187. Currently Frang count number of skb's and uses `req->chunk_cnt` field. However, skb is relatively abstract structure which can contain many data fragments thanks to GRO. Packets from Slow HTTP would be typically skb with only one fragment while packets from massive data streams typically contain highly fragmented skbs. Meantime, we already have time limits for assembling different parts of HTTP messages, so number of chunks should be **orthogonal** to this limit and implemented via counting TfwStr chunks (note that special HTTP headers have +1 chunk splitting header name and value). `req->chunk_cnt` should be eliminated.

### LINKED ISSUE #614: TLS performance extensions
Task #81 was implemented by porting mbedTLS as is, i.e. with all buffering and copies. See comments in https://github.com/tempesta-tech/tempesta/pull/603, https://github.com/tempesta-tech/tempesta/pull/596 and https://github.com/tempesta-tech/tempesta/pull/595 . The module must use in-place encryption and decryption.

Modern Linux kernels implement almost all crypto algorithms (maybe just except ECDSA) required for TLS. So it has sense to port mbed TLS handshake only to the kernel and use standard Linux crypto for encryption. Moreover, Linux 4.13 implements TLS, so moving to recent kernel is required. Only TLS handshake and infrastructure code must be taken from mbed TLS, while Linux code must be used for symmetric crypto. [Full TLS (#769)](https://github.com/tempesta-tech/tempesta/issues/769) must be kept in mind during the library adjustments.

Also current TLS code is based on mbedTLS 2.2 and must be updated to latest GPL version (currently [2.6](https://tls.mbed.org/download/start/mbedtls-2.6.0-gpl.tgz)), which has many security fixes.

It seems special hooks at handshake code are required to implement [SSL/TLS Renegotiation DoS protection (#832)](https://github.com/tempesta-tech/tempesta/issues/832).

Good reference article about TLS DDoS mitigation [SSL computational DoS mitigation](https://vincent.bernat.im/en/blog/2011-ssl-dos-mitigation).

krizhanovsky:
Fix #614:
1. Fix the memory leak in TLS;
1. Remvoed rx and tx copying queues in TLS;
1. remove copyings `skb_{copy,store}_bits()`, all crypto is done in-place now;
1.  fix multiple allocations, calloc(), malloc(), free() wrappers are almost eliminated. At least from crucial workflos for now
1. fully rewritten mbedTLS state machine for TLS handshakes, so that now it works with chunked data in zero-copy fashion;
2. unification with linux/crypt - we use the crypt, but RSA and ECDSA which seems incomplete in the kernel
3. encrypt TLS records with dynamic size calculated in `tcp_write_xmit()` (see https://github.com/tempesta-tech/linux-4.14.32-tfw/pull/3): we encrypt several skbs, but not more than a minimum between TCP congestion window and the receiver's advertised window;
4. Address #391 point 12: now we can use skb->head as paged fragments;
5. Introduced unified logging as `lib/log.h` and unified return values
6.  `ss_skb_process()`: process skb chains, replace parser->to_go by return value variable, so ss_skb_process() returns how much data was actually processed (seems buggy);
7. Fix #214
8.  Address RFC7525 by removing weak algorithms.
9. Many cleanups and coding style adjustments.


tempesta_fw/log.h
```diff
@@ -23,80 +23,19 @@
...
+#define BANNER	"fw"
+#include "lib/log.h"
+/* TODO remvoe the defefines after moving to unified logging. */
+#define TFW_ERR(...)		T_ERR(__VA_ARGS__)
+#define TFW_ERR_NL(...)		T_ERR_NL(__VA_ARGS__)
+#define TFW_WARN(...)		T_WARN(__VA_ARGS__)
+#define TFW_WARN_NL(...)	T_WARN_NL(__VA_ARGS__)
+#define TFW_LOG(...)		T_LOG(__VA_ARGS__)
+#define TFW_LOG_NL(...)		T_LOG_NL(__VA_ARGS__)
+#define TFW_ERR(...)		T_ERR(__VA_ARGS__)
```
aleksostapenko: The same define of `TFW_ERR` repeated twice.

tempesta_fw/ss_skb.c
```diff
@@ -97,6 +98,28 @@ ss_skb_alloc_pages(size_t len)
...
+ * segmentation. The allocated payload space will be filled with data.
+ */
+int
+ss_skb_alloc_data(struct sk_buff **skb_head, size_t len)
+{
+	int i_skb, nr_skbs = DIV_ROUND_UP(len, SS_SKB_MAX_DATA_LEN);
+	struct sk_buff *skb;
+
+	for (i_skb = 0; i_skb < nr_skbs; ++i_skb) {
+		skb = ss_skb_alloc_pages(min(len, SS_SKB_MAX_DATA_LEN));
```
aleksostapenko: The variable `len` does not decrease, so it seems that the last skb always will be allocated with `MAX_SKB_FRAGS` quantity of frags, even if remaining data will fits entirely in the linear part of skb.

tempesta_fw/ss_skb.c
```diff
@@ -267,6 +290,10 @@ __extend_pgfrags(struct sk_buff *skb_head, struct sk_buff *skb, int from, int n)
 		struct sk_buff *nskb;
 		unsigned int e_size = 0;
 
+		/* Going out of the @skb is prohibied by the caller. */
```
aleksostapenko: `of` -> `if`

tempesta_fw/msg.c
```diff
@@ -0,0 +1,105 @@
...
+
+		if (c_size < f_room) {
+			/*
+			 * The chunk fits in the SKB fragment with room
+			 * to spare. Stay in the same SKB fragment, swith
+			 * to next chunk of the string.
+			 */
+			c_off = 0;
+		} else {
+			frag = ss_skb_frag_next(&it->skb, &it->frag);
```
aleksostapenko: There is a problem in `ss_skb_frag_next()` which is a consequence  of mine #1000 PR: `!*skb ` is not relevant anymore after making skb list cyclic. Maybe it is worth to add `skb_head` pointer member into `TfwMsgIter` structure for comparison purpose.

tempesta_fw/msg.h
```diff
@@ -48,4 +48,21 @@ typedef struct {
 	size_t			len;
 } TfwMsg;
 
+/**
+ * Iterator for @skb fragments.
+ *
+ * @frag	- current fragment index or @skb->head if -1;
+ * @skb		- current skb fragment;
```
aleksostapenko: `current skb fragment;` -> `current skb;`

tempesta_fw/:w
```diff
@@ -0,0 +1,722 @@
+/**
```
ikoveshnikov: Mistakenly committed file.

tempesta_fw/http.c
```diff
@@ -2654,256 +2680,266 @@ tfw_http_hm_drop_resp(TfwHttpResp *resp)
...
 		 */
-		if (!req->vhost) {
-			tfw_http_send_resp(req, 404, "request dropped: cannot"
-					   " find appropriate virtual host");
-			TFW_INC_STAT_BH(clnt.msgs_otherr);
-		} else if (tfw_cache_process((TfwHttpMsg *)req,
-					     tfw_http_req_cache_cb))
-		{
+		hmsib = tfw_http_msg_create_sibling((TfwHttpMsg *)req, &skb,
+						    parsed);
```
aleksostapenko: Looks like `off + parsed` should be passed as 3rd argument into `tfw_http_msg_create_sibling()`.

tls/tls_srv.c
```diff
@@ -0,0 +1,2331 @@
...
+				return TTLS_ERR_BAD_HS_CLIENT_HELLO;
+			}
+			TTLS_HS_FSM_MOVE(TTLS_CH_HS_EXTLEN);
+		}
+		TTLS_HS_FSM_MOVE(TTLS_CH_HS_COMP);
+	}
+
+	/* Read extensins length. */
+	T_FSM_STATE(TTLS_CH_HS_EXTLEN) {
+		/* Use the first 2 bytes for toal number of extensions. */
```
aleksostapenko: Maybe `total` instead of `toal` should be.
In previous comment: `extensins` -> `extensions`.

tls/tls_srv.c
```diff
@@ -0,0 +1,2331 @@
...
+	      ttls_get_ciphersuite_name(tls->sess.ciphersuite));
+
+	/*
+	 * First write extensions, then the total length.
+	 *
+	 * RFC 7366: "If a server receives an encrypt-then-MAC request extension
+	 * from a client and then selects a stream or Authenticated Encryption
+	 * with Associated Data (AEAD) ciphersuite, it MUST NOT send an
+	 * encrypt-then-MAC response extension back to the client."
+	 * We don't support other siphersuites, so we don't send EtM extension.
```
aleksostapenko: `siphersuites` -> `ciphersuites`

tls/tls_srv.c
```diff
@@ -0,0 +1,2331 @@
...
+			T_FSM_EXIT();
+		CHECK_STATE(128);
+		T_FSM_JMP(TTLS_SERVER_KEY_EXCHANGE);
+	}
+	T_FSM_STATE(TTLS_SERVER_KEY_EXCHANGE) {
+		if ((r = ttls_write_server_key_exchange(tls, &sgt, &p)))
+			T_FSM_EXIT();
+		CHECK_STATE(1024);
+		/*
+		 * RFC 5246 Certificate Request is optional, so don't requiest
```
aleksostapenko: `requiest` -> `request`

tls/tls_internal.h
```diff
@@ -0,0 +1,416 @@
...
+ * 		  already knows because of state of the renegotiation flag, so
+ * 		  no indicator is required);
+ * @extended_ms	- use Extended Master Secret (RFC 7627)?
+ * @new_session_ticket - use NewSessionTicket?
+ * @resume	- session resume indicator;
+ * @cli_exts	- client extension presence;
+ * @curves	- supported elliptic curves;
+ * @randbytes	- random bytes;
+ * @finished	- temporal buffer for chunks of Finished message,
+ *		  @randbytes were used in prvious messages, so we can reuse it
```
aleksostapenko: `prvious` -> `previous`

tls/tls_internal.h
```diff
@@ -0,0 +1,416 @@
...
+	TTLS_SERVER_CHANGE_CIPHER_SPEC	= __TTLS_FSM_ST(11),
+	TTLS_SERVER_FINISHED		= __TTLS_FSM_ST(12),
+	TTLS_HANDSHAKE_WRAPUP		= __TTLS_FSM_ST(13),
+	TTLS_HANDSHAKE_OVER		= __TTLS_FSM_ST(14),
+	TTLS_SERVER_NEW_SESSION_TICKET	= __TTLS_FSM_ST(15),
+	TTLS_SERVER_HELLO_VERIFY_REQUEST_SENT = __TTLS_FSM_ST(16),
+};
+
+/*
+ * Extend the common FSM for hanshake parsing.
```
aleksostapenko: `hanshake` -> `handshake`

tls/tls_internal.h
```diff
@@ -0,0 +1,416 @@
...
+ */
+#define TTLS_HS_FSM_FINISH()						\
+	T_FSM_FINISH(r, tls->state);					\
+	*read += p - buf;						\
+	io->rlen += p - buf;
+
+/* Move to @st if we have at least @need bytes. */
```
aleksostapenko: Seems that comment is outdated: `@need` parameter is not present in the macro.

tls/ttls.c
```diff
@@ -1,90 +1,2790 @@
...
+
+static void
+ttls_handshake_free(TlsHandshake *hs)
+{
+	if (!hs)
+		return;
+
+	/*
+	 * Free only the linked list wrapper, not the keys themselves
+	 * since the belong to the SNI callback.
```
aleksostapenko: Maybe `the` -> `they`.

tls/ttls.c
```diff
@@ -1,90 +1,2790 @@
...
+	tls->hs->calc_finished(tls, hash, tls->conf->endpoint ^ 1);
+	if (crypto_memneq(&hs->finished[TTLS_HS_HDR_LEN], hash, TLS_HASH_LEN)) {
+		T_DBG("bad hash in finished message\n");
+		ttls_send_alert(tls, TTLS_ALERT_LEVEL_FATAL,
+				    TTLS_ALERT_MSG_DECODE_ERROR);
+		return TTLS_ERR_BAD_HS_FINISHED;
+	}
+
+	/*
+	 * Calculate final message checksum before goint to
```
aleksostapenko: `goint` -> `going`

tempesta_fw/tls.c
```diff
@@ -22,319 +22,548 @@
...
-static inline int
-ttls_ssl_write(TfwTlsContext *tls, const unsigned char *data, size_t size)
-{
-	int r;
+	BUG_ON(off >= skb->len);
 
+	/*
+	 * Perform TLS handshake if necessary and decrypt the TLS message
+	 * in-place by chunks. Add skb to the list to build scatterlist if it
+	 * it contains end of current message.
```
aleksostapenko: Double `it` word.

tempesta_fw/tls.c
```diff
@@ -22,319 +22,548 @@
...
+		 * record. Typically, handshake messages fit the same skb and
+		 * all the messages are processed in one ss_skb_process() call.
+		 * Collect all skb chunks of data record in skb_list.
+		 */
+		spin_unlock(&tls->lock);
+		return TFW_PASS;
+	case T_OK:
+		/*
+		 * A complete TLS message decrypted and ready for upper
+		 * layer protocols processing - fall throught.
```
aleksostapenko: `throught` -> `through`

tempesta_fw/tls.c
```diff
@@ -22,319 +22,548 @@
...
+	/*
+	 * Possibly there are other TLS message in the @skb - create
+	 * an skb sibling and process it on the next iteration.
+	 * If a part of incomplete TLS message leaves at the end of the
+	 * @skb, then store the skb in the TLS context for next FSM
+	 * shot.
+	 *
+	 * Many sibling skbs can be produced by TLS and HTTP layers
+	 * together - don't coalesce them: we process messages at once
+	 * and it hase sense to work with sparse skbs in HTTP
```
aleksostapenko: `hase` -> `has`

tempesta_fw/tls.c
```diff
@@ -22,319 +22,548 @@
...
+	 * adjustment logic to have some room to place a new fragments.
+	 * The logic is simple because each layer works with messages
+	 * from previous layer not crossing skb boundaries. The drawback
+	 * is that we produce a lot of skbs causing pressure on the
+	 * memory allocator.
+	 *
+	 * Split @skb before calling HTTP layer to chop it and not let HTTP
+	 * to read after end of the message.
+	 */
+	if (parsed < skb->len) {
```
aleksostapenko: It seems that `off + parsed < skb->len` condition should be checked here.

tls/ttls.h
```diff
@@ -1,49 +1,991 @@
...
+int ttls_conf_alpn_protocols(ttls_config *conf, const char **protos);
+
+/**
+ * \brief	Get the name of the negotiated Application Layer Protocol.
+ *		This function should be called after the handshake is
+ *		completed.
+ *
+ * \param ssl	SSL context
+ *
+ * \return	Protcol name, or NULL if no protocol was negotiated.
```
aleksostapenko: `Protcol` -> `Protocol`

tls/ttls.h
```diff
@@ -1,49 +1,991 @@
...
+ * @aad_buf	- temporary buffers for associated authentication data;
+ * @msgtype	- record header: message type;
+ * @hstype	- record header: handhsake type;
+ * @msglen	- record header: message length;
+ * @hslen	- current handshake message length, including the handshake
+ *		  header. For egress data used as length of @__msg (copied
+ *		  data);
+ * @rlen	- read bytes of the message body so far;
+ * @skb_list	- list of skbs attachd to the current I/O context;
+ * @off		- data offset within @skb_list, can be afther the 1st skb;
```
aleksostapenko: `attachd` -> `attached`
`afther` -> `after`

tls/md.c
```diff
@@ -1,436 +1,229 @@
...
-	if (ctx->md_ctx != NULL)
-		ctx->md_info->ctx_free_func(ctx->md_ctx);
-
-	if (ctx->hmac_ctx != NULL)
-	{
-		mbedtls_zeroize(ctx->hmac_ctx, 2 * ctx->md_info->block_size);
-		mbedtls_free(ctx->hmac_ctx);
+	md_ctx->tfm = crypto_alloc_shash(md_info->alg_name, 0, 0);
+	if (IS_ERR(md_ctx->tfm)) {
+		T_ERR("cannot initizlize hash driver %s."
```
aleksostapenko: `initizlize` -> `initialize`

tempesta_fw/tls.c
```diff
@@ -22,319 +22,548 @@
...
+	 * Perform TLS handshake if necessary and decrypt the TLS message
+	 * in-place by chunks. Add skb to the list to build scatterlist if it
+	 * it contains end of current message.
+	 */
 	spin_lock(&tls->lock);
-	r = mbedtls_ssl_write(&tls->ssl, data, size);
-	spin_unlock(&tls->lock);
+next_msg:
+	ss_skb_queue_tail(&tls->io_in.skb_list, skb);
+	msg_skb = tls->io_in.skb_list;
```
aleksostapenko: I guess, `skb_list` processing can be made without `msg_skb` variable: here and in `tfw_tls_chop_skb_rec()` - we have access to `tls->io_in.skb_list`.

tempesta_fw/tls.c
```diff
@@ -22,319 +22,548 @@
...
+static int
+tfw_tls_chop_skb_rec(TlsCtx *tls, struct sk_buff *skb, unsigned int off,
+		     TfwFsmData *data)
 {
-	int r;
+	size_t prolog = ttls_payload_off(&tls->xfrm);
+
+eaten_skb:
+	if (likely(skb->len > prolog)) {
+		off += prolog;
```
aleksostapenko: It seems that case with `skb->len > prolog` and `skb->len == prolog + off` may appear here, i.e. skb's space is exhausted but we will continue to deal with it.

lib/log.h
```diff
@@ -0,0 +1,140 @@
...
+} while (0)
+
+#define T_ERR(...)	__CALLSTACK_MSG(KERN_ERR __BNR "ERROR: " __VA_ARGS__)
+#define T_WARN(...)	__CALLSTACK_MSG(KERN_WARNING __BNR		\
+					"Warning: " __VA_ARGS__)
+#define T_LOG(...)	pr_info(__BNR __VA_ARGS__)
+/* Non-limited printing. */
+#define T_ERR_NL(...)	T_ERR(__VA_ARGS__)
+#define T_WARN_NL(...)	T_WARN(__VA_ARGS__)
+#define T_LOG_NL(...)	T_LOG(__VA_ARGS__)
```
ikoveshnikov: Non limited printing is normally USER level warnings like errors or warnings  while configuration parsing. There is no need to print call stacks here. `TFW_(ERR|WARN|LOG)` messages are also used for USER-level only messages and callstacks are not really required there and it sometimes annoying to search though system log and search between callstacks while some of them are not internal errors.

tempesta_fw/http_parser.c
```diff
@@ -1131,8 +1214,7 @@ __parse_connection(TfwHttpMsg *hm, unsigned char *data, size_t len)
 	int r = CSTR_NEQ;
 	__FSM_DECLARE_VARS(hm);
 
-	BUILD_BUG_ON(sizeof(parser->hbh_parser.spec) * CHAR_BIT
-		     < TFW_HTTP_HDR_RAW);
+	BUILD_BUG_ON(sizeof(parser->hbh_parser.spec) * 8 < TFW_HTTP_HDR_RAW);
```
ikoveshnikov: Why the change is required? Just to fit single line? Not very critical since we support only x86_64, but I prefer to have CHAR_BITS or BITS_PER_BYTE here.

tempesta_fw/sock_clnt.c
```diff
@@ -189,7 +189,20 @@ tfw_sock_clnt_new(struct sock *sk)
...
 	tfw_connection_link_peer(conn, (TfwPeer *)cli);
+
 	ss_set_callbacks(sk);
+	if (TFW_CONN_TYPE(conn) & TFW_FSM_HTTPS)
+		/*
+		 * Probably, that's not beautiful to introduce an alternate
+		 * upcall beside GFSM and SS, but that's efficient and I didn't
+		 * find a simple and better solution.
+		 */
+		conn->sk->sk_write_xmit = tfw_tls_encrypt;
```
ikoveshnikov: Not an issue, but better to use simply `sk->sk_write_xmit` as everywhere else.

tempesta_fw/tls.c
krizhanovsky: Yeah, now we modify `tls->io_in.skb_list` in `tfw_tls_chop_skb_rec()`: there was a bug that `tfw_tls_msg_process()` gets `off` from TCP layer as an overlap of sequence numbers, but we use the offset as TLS header for HTTP processing.

tempesta_fw/tls.c
krizhanovsky: Now we chop the possible TCP seqnos overlap at the begin of `tfw_tls_msg_process()` because we can not link such overlapping skb at the middle of the skb chain for HTTP processing (there is no way to specify offsets for each skb in the list). So `prolog` went away and `off` is TLS header and IV.

tempesta_fw/http_parser.c
krizhanovsky: I have no idea :) Looks like a mess.

tempesta_fw/tls.c
krizhanovsky: Actually ttls_recv() initializes IO context after each processed TLS record so tfw_tls_msg_process() must store skb_list for further chopping in tfw_tls_chop_skb_rec(). Since ttls_recv() zeroes almost full IO, tfw_tls_chop_skb_rec() doesn't need to zero it. Fixed.

tempesta_fw/http.h
```diff
@@ -251,6 +250,8 @@ enum {
 	TFW_HTTP_B_WHITELIST,
 	/* Client was disconnected, drop the request. */
 	TFW_HTTP_B_REQ_DROP,
+	/* Request was received on TLS connection. */
+	TFW_HTTP_B_REQ_TLS,
```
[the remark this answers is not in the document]
krizhanovsky: Oops. Fixed.

tls/tls_srv.c
```diff
@@ -0,0 +1,2331 @@
...
+
+	diff  = (unsigned int) r;
+	diff |= peer_pmslen ^ 48;
+	diff |= peer_pms[0] ^ ver[0];
+	diff |= peer_pms[1] ^ ver[1];
+	T_DBG("client key exchange message diff=%x\n", diff);
+
+	tls->hs->pmslen = 48;
+	/* mask = diff ? 0xff : 0x00 using bit operations to avoid branches */
+	mask = - ((diff | - diff) >> (sizeof(unsigned int) * 8 - 1));
```
ikoveshnikov: `diff | -diff` seems to be invalid operation here. This operation always result as `0xffff_ffff `.

ikoveshnikov: `diff | -diff` seems to be invalid operation here. This operation always result as `0xffff_ffff `.

krizhanovsky: Seems not:
```C
$ cat t.c 
#include <stdio.h>

int
main(int argc, char *argv[])
{
	unsigned int diff = 0;
	unsigned char mask;

	mask = -((diff | -diff) >> (sizeof(unsigned int) * 8 - 1));
	printf("mask = %x\n", mask);

	diff = 1;
	mask = -((diff | -diff) >> (sizeof(unsigned int) * 8 - 1));
	printf("mask = %x\n", mask);

	return 0;
}
$ gcc t.c && ./a.out 
mask = 0
mask = ff
```

krizhanovsky:
Fixed in https://github.com/tempesta-tech/tempesta/pull/1143

PR #1046, opened 2018-07-25
Describe case of absent proxy_pass

vladtcvs:


etc/tempesta_fw.conf
```diff
@@ -698,9 +698,9 @@
 #   Directive proxy_pass is mandatory for "vhost" and "location" directive
 #   blocks. For explicit default vhost directive (and included locations)
 #   only proxy_pass directives with 'default' server groups is allowed to be
-#   specified (or proxy_pass directive may be absent at all). For implicit
-#   default vhost - proxy_pass directive is not allowed (since default vhost
-#   must point to default server group).
+#   specified (or proxy_pass directive may be absent at all. In this case it
```
ikoveshnikov: The the expression contradicts itself. It claims that `proxy_pass` directive may be absent for explicit 'default' group. In the same time the next sentence says that the `proxy_pass` directive is created automatically. 

 According to the sources, `proxy_pass` directive is always created for the 'default' vhost and can't be absent: https://github.com/tempesta-tech/tempesta/blob/10d8307a6ab1bafe68f4e13aba64fdc886a38fc5/tempesta_fw/vhost.c#L1854-L1859

In the same time, `proxy_pass` directive is not added implicitly for the `location` directive.

etc/tempesta_fw.conf
```diff
@@ -698,9 +698,9 @@
...
+#   is created automatically). For implicit default vhost - proxy_pass directive
+#   is not allowed (since default vhost must point to default server group).
```
ikoveshnikov: 'For implicit default vhost - proxy_pass directive is not allowed (since default vhost must point to default server group).' -> 'For the implicit "defalt" vhost only `default` value for `proxy_pass` directive is allowed. It's created automatically if "default" server group is defined, explicit `proxy_pass` directive is not allowed in this case.'

etc/tempesta_fw.conf
aleksostapenko: Locations of default `vhost` do not create their own implicit `default` `proxy_pass` directives, but instead inherit the `proxy_pass` of parent `default` `vhost`. But I think, this is implementations details and we shouldn't expose them to user. So, maybe next changes would be appropriate:
'(or `proxy_pass` directive may be absent at all)' -> '(or explicit `proxy_pass` directive may be absent at all, in which case the implicit `proxy_pass` will be created automatically with `default` value)'

krizhanovsky:
@vladtcvs is on another project now, so @ikoveshnikov could you please finish and merge the PR.

etc/tempesta_fw.conf
```diff
@@ -982,36 +1031,38 @@
...
-# CRC32 is a hex number - calculated CRC32 checksum for
-# expected response body.
+# CRC32 is a hex number - calculated CRC32 checksum for expected response body.
 #
-# TIMEOUT is a timeout in seconds after which new health
-# monitoring request (specified in 'request' directive) will
-# be send to backend server (if there were no client requests
-# satisfying condition, given in 'request_url' directive).
+# TIMEOUT is a timeout in seconds after which new health monitoring request
+# (specified in 'request' directive) will be send to backend server (if there
```
aleksostapenko: `send` -> `sent`

Issue #962, opened 2018-03-19
Drop client connections on response blocking


krizhanovsky:
The issue is based on the comment https://github.com/tempesta-tech/tempesta/pull/960#pullrequestreview-104860832 : if `tfw_http_resp_process()` falls to `bad_msg` label, then the error/attack treatment is delegated to `tfw_http_srv_error_resp_and_log()`. If error responses are configured (`reply` parameter is true), them everything is fine. However, if we just block the response (`reply` is false), then current request and responses are just dropped and we continue to process next client requests. This way we break request-response sequence which we must not do.

If we block a response (as well as request), then we have to terminate client connection in respect to pipelined requests.

PR #1047, opened 2018-07-27
Fix #535: Block clients which requests have no session cookie.

### LINKED ISSUE #535: Sticky cookie extension for  L7 DDoS mitigation
Firstly, sticky cookie module must be extended by stronger than `enforce` requirement: if a client receiving the redirect tries to access the resource more than specified number of tries without set cookie, then its IP address must be blocked like Frang does it.

Please update https://github.com/tempesta-tech/tempesta/wiki/DDoS-mitigation#http-cookie-challenge and create a functional test issue.

aleksostapenko:
~~Fix for #962 is also included in this PR.~~

### DIFF
--- a/etc/tempesta_fw.conf
+++ b/etc/tempesta_fw.conf
@@ -725,7 +725,7 @@
 # Tempesta sticky cookie.
 #
 # Syntax:
-#   sticky [name=<COOKIE_NAME>] [enforce]
+#   sticky [name=<COOKIE_NAME>] [enforce] [max_misses=<LIMIT>]
 #
 # Default:
 #   Tempesta sticky cookie is not used.
@@ -739,6 +739,12 @@
 # found in an HTTP request, a client get an HTTP 302 response that
 # redirects the client to the same URI, and prompts that Tempesta
 # sticky cookie is set in the request.
+#
+# max_misses - sets the maximum count of requests without cookie.
+# This option is applicable only if 'enforce' mode is enabled.
+# If configured limit is exhausted, the corresponding client will
+# be blocked. LIMIT is non-negative integer (default value is 0,
+# which means that limit verification is disabled).
 #
 # NOTE: line breaks inside the directive are not allowed. See 'js_challenge'.
 #
--- a/etc/tempesta_fw.conf
+++ b/etc/tempesta_fw.conf
@@ -740,8 +740,8 @@
 # redirects the client to the same URI, and prompts that Tempesta
 # sticky cookie is set in the request.
 #
-# max_misses - sets the maximum count of redirected requests (with
-# invalid cookie or without cookie at all). This option is applicable
+# max_misses - sets the maximum count of redirected requests (with no
+# cookie or with incorrect cookie value). This option is applicable
 # only if 'enforce' mode is enabled. If configured limit is exhausted,
 # the corresponding client will be blocked. LIMIT is non-negative
 # integer (default value is 0, which means that limit verification
--- a/etc/tempesta_fw.conf
+++ b/etc/tempesta_fw.conf
@@ -725,7 +725,7 @@
 # Tempesta sticky cookie.
 #
 # Syntax:
-#   sticky [name=<COOKIE_NAME>] [enforce] [max_misses=<LIMIT>]
+#   sticky [name=<COOKIE_NAME>] [enforce] [max_misses=<LIMIT>] [timeout=<TMT>]
 #
 # Default:
 #   Tempesta sticky cookie is not used.
@@ -740,11 +740,20 @@
 # redirects the client to the same URI, and prompts that Tempesta
 # sticky cookie is set in the request.
 #
-# max_misses - sets the maximum count of requests without cookie.
-# This option is applicable only if 'enforce' mode is enabled.
-# If configured limit is exhausted, the corresponding client will
-# be blocked. LIMIT is non-negative integer (default value is 0,
-# which means that limit verification is disabled).
+# max_misses - sets the maximum count of redirected requests (with
+# invalid cookie or without cookie at all). This option is applicable
+# only if 'enforce' mode is enabled. If configured limit is exhausted,
+# the corresponding client will be blocked. LIMIT is non-negative
+# integer (default value is 0, which means that limit verification
+# is disabled).
+
+# timeout - sets the maximum period between the first redirected
+# request and succsessful cookie verification from certain client.
+# This option is applicable only if 'enforce' mode is enabled and
+# 'max_misses' parameter has value greater then zero. If configured
+# timeout is expired, the corresponding client will be blocked.
+# TMT is non-negative integer (default value is 0, which means that
+# timeout tracking is disabled).
 #
 # NOTE: line breaks inside the directive are not allowed. See 'js_challenge'.
 #
--- a/tempesta_fw/client.h
+++ b/tempesta_fw/client.h
@@ -42,7 +42,6 @@ typedef struct {
 	struct hlist_node	hentry;
 	spinlock_t		*hb_lock;
 	atomic_t		conn_users;
-	atomic_t		miss_count;
 	TfwClassifierPrvt	class_prvt;
 } TfwClient;
 
--- a/tempesta_fw/client.h
+++ b/tempesta_fw/client.h
@@ -32,7 +32,6 @@
  * @conn_users		- connections reference counter.
  * 			  The client is released, when the counter reaches zero:
  * 			  no connections to the server - no client for us :)
- * @miss_count		- count of cookie misses in requests;
  * @class_prvt		- private client accounting data for classifier module.
  *			  Typically it's large and vastes memory in vain if
  *			  no any classification logic is used;
--- a/tempesta_fw/client.h
+++ b/tempesta_fw/client.h
@@ -32,6 +32,7 @@
  * @conn_users		- connections reference counter.
  * 			  The client is released, when the counter reaches zero:
  * 			  no connections to the server - no client for us :)
+ * @miss_count		- count of cookie misses in requests;
  * @class_prvt		- private client accounting data for classifier module.
  *			  Typically it's large and vastes memory in vain if
  *			  no any classification logic is used;
@@ -41,6 +42,7 @@ typedef struct {
 	struct hlist_node	hentry;
 	spinlock_t		*hb_lock;
 	atomic_t		conn_users;
+	atomic_t		miss_count;
 	TfwClassifierPrvt	class_prvt;
 } TfwClient;
 
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -2341,85 +2341,92 @@ tfw_http_req_mark_error(TfwHttpReq *req, int status)
 }
 
 /**
- * Functions define logging and response behaviour during detection of
- * malformed or malicious messages. Mark client connection in special
- * manner to delay its closing until transmission of error response
- * will be finished.
+ * Function defines logging and response behaviour during detection of
+ * malformed or malicious messages. If reply to client must be sent,
+ * mark client connection in special manner to delay its closing until
+ * transmission of error response will be finished; otherwise, close
+ * connection immediately.
  */
 static void
-tfw_http_cli_error_resp_and_log(bool reply, bool nolog, TfwHttpReq *req,
-				int status, const char *msg)
+tfw_http_error_resp_and_log(bool reply, bool nolog, TfwHttpReq *req,
+			    int status, const char *msg, bool close)
 {
 	if (!nolog)
 		TFW_WARN_ADDR(msg, &req->conn->peer->addr);
 
 	if (reply) {
-		TfwCliConn *cli_conn = (TfwCliConn *)req->conn;
-		tfw_connection_unlink_msg(req->conn);
-		spin_lock(&cli_conn->seq_qlock);
-		list_add_tail(&req->msg.seq_list, &cli_conn->seq_queue);
-		spin_unlock(&cli_conn->seq_qlock);
+		if (!close) {
+			TfwCliConn *cli_conn = (TfwCliConn *)req->conn;
+			tfw_connection_unlink_msg(req->conn);
+			spin_lock(&cli_conn->seq_qlock);
+			list_add_tail(&req->msg.seq_list, &cli_conn->seq_queue);
+			spin_unlock(&cli_conn->seq_qlock);
+		}
 		tfw_http_req_mark_error(req, status);
+		return;
 	}
-	else {
-		tfw_http_conn_req_clean(req);
-	}
-}
 
-static void
-tfw_http_srv_error_resp_and_log(bool reply, bool nolog, TfwHttpReq *req,
-				int status, const char *msg)
-{
-	if (!nolog)
-		TFW_WARN_ADDR(msg, &req->conn->peer->addr);
+	if (close)
+		ss_close_sync(req->conn->sk, true);
 
-	if (reply)
-		tfw_http_req_mark_error(req, status);
-	else
-		tfw_http_conn_req_clean(req);
+	tfw_http_conn_req_clean(req);
 }
 
 /**
- * Wrappers for calling tfw_http_cli_error_resp_and_log() and
- * tfw_http_srv_error_resp_and_log() functions in client/server
- * connection contexts depending on configuration settings:
- * sending response error messages and logging.
+ * Wrappers for calling tfw_http_error_resp_and_log() function in client/server
+ * connection contexts depending on configuration settings: sending response
+ * error messages and logging.
  *
- * NOTE: tfw_client_drop() and tfw_client_block() must be called
- * only from client connection context before a request was fully parsed.
- * Otherwise tfw_srv_client_drop() and tfw_srv_client_block() must be used
- * only from server connection context.
+ * NOTE: tfw_client_drop() and tfw_client_block() are intented to be called
+ * only from client connection context during request processing and before
+ * request was attached to connection's @seq_queue - these functions do this
+ * work on their own; also, these functions do not close client connection,
+ * since during request processing connection can always be closed from the
+ * caller (tfw_http_req_process() procedure) via appropriate return value.
+ * Functions tfw_client_drop_close() and tfw_client_block_close() must be
+ * used in all other cases, including server connection contexts, as they
+ * don't operate with @seq_queue and are responsible for client connection
+ * closing.
  */
+static inline void
+__tfw_client_drop(TfwHttpReq *req, int status, const char *msg, bool close)
+{
+
+	tfw_http_error_resp_and_log(tfw_blk_flags & TFW_BLK_ERR_REPLY,
+				    tfw_blk_flags & TFW_BLK_ERR_NOLOG,
+				    req, status, msg, close);
+}
+
+static inline void
+__tfw_client_block(TfwHttpReq *req, int status, const char *msg, bool close)
+{
+	tfw_http_error_resp_and_log(tfw_blk_flags & TFW_BLK_ATT_REPLY,
+				    tfw_blk_flags & TFW_BLK_ATT_NOLOG,
+				    req, status, msg, close);
+}
+
 static inline void
 tfw_client_drop(TfwHttpReq *req, int status, const char *msg)
 {
-	tfw_http_cli_error_resp_and_log(tfw_blk_flags & TFW_BLK_ERR_REPLY,
-					tfw_blk_flags & TFW_BLK_ERR_NOLOG,
-					req, status, msg);
+	__tfw_client_drop(req, status, msg, false);
 }
 
 static inline void
 tfw_client_block(TfwHttpReq *req, int status, const char *msg)
 {
-	tfw_http_cli_error_resp_and_log(tfw_blk_flags & TFW_BLK_ATT_REPLY,
-					tfw_blk_flags & TFW_BLK_ATT_NOLOG,
-					req, status, msg);
+	__tfw_client_block(req, status, msg, false);
 }
 
 static inline void
-tfw_srv_client_drop(TfwHttpReq *req, int status, const char *msg)
+tfw_client_drop_close(TfwHttpReq *req, int status, const char *msg)
 {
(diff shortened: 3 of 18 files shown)


tempesta_fw/http_sess.c
```diff
@@ -425,6 +433,13 @@ tfw_http_sticky_notfound(TfwHttpReq *req)
 	if (!tfw_cfg_sticky.enforce)
 		return TFW_HTTP_SESS_SUCCESS;
 
+	if (tfw_cfg_sticky.max_misses &&
+	    atomic_inc_return(&cli->miss_count) > tfw_cfg_sticky.max_misses)
```
ikoveshnikov: You can set ` tfw_cfg_sticky.max_misses` to `INT_MAX` for case when `max_misses` check is disabled. The approach is used for some other directives.

krizhanovsky:
The implementation is correct in context of #535, but I agree with @ikoveshnikov that the new limit is just a duplication of redirect limitation.

> It also becomes hard to predict when the ip address blocking may be triggered, since the counter is saved in TfwClient instance which is destroyed when all client from that IP closes connections to Tempesta server.

A fix for this particular issue is requested in https://github.com/tempesta-tech/tempesta/issues/488#issuecomment-360432734

I don't see why a bot can not always send `resource?attempt=1`. So I propose to use cryptographically strong (relatively) hash function and current timestamp to prevent repetition attacks (just like we do this in the cookie itself), so we need to send something like `resource?attempt=<attempt_no | jiffies | hash(attempt_no,jiffies,secret)` and check the attempt number and the timestamp. Block the user if timestamp is older that configured number of seconds (please add the one more option to the configuration directive).

ikoveshnikov:
Seems like authorizing URI format of `/<HASH>/<ORIGINAL_URI>` can cause some troubles. E.g. some sites with user generated content and forums may use large decimal integers to address resources such as topics, articles, comments. 

If the discussed option is switched on, that integer resource URI can be mistakenly treated as invalid hash value and the client would be blocked. We still have to add literal marker before hash to prevent such errors. The resulting authorizing URI format should be `/<MARKER>_<HASH>/<ORIGINAL_URI>`. `<MARKER>` should be literal value not expected for protected resource, e.g. same as sticky cookie name which is `__tfw` by default.

krizhanovsky:
Yeah, agree. The marker should be configurable and I think there is no sense to handle two markers, for the cookie and the URL, so let's ask a user for the same marker and use it from the config file.

tempesta_fw/http.c
```diff
@@ -2341,85 +2350,92 @@ tfw_http_req_mark_error(TfwHttpReq *req, int status)
...
 	if (!nolog)
 		TFW_WARN_ADDR(msg, &req->conn->peer->addr);
 
 	if (reply) {
-		TfwCliConn *cli_conn = (TfwCliConn *)req->conn;
-		tfw_connection_unlink_msg(req->conn);
-		spin_lock(&cli_conn->seq_qlock);
-		list_add_tail(&req->msg.seq_list, &cli_conn->seq_queue);
-		spin_unlock(&cli_conn->seq_qlock);
+		if (!close) {
```
ikoveshnikov: Hm, not sure about that. There are `BUG_ON(list_empty(&req->msg.seq_list));` checks in `tfw_http_resp_fwd()` which are called in  `tfw_http_req_mark_error()` function. I expect to see a crash if `reply == true && close == true`.

May be you need to set TFW_HTTP_B_CONN_CLOSE flag for a request if `close == true`? Tempesta closes connection to clients when forwarding a response to request withe the flag set. See `tfw_http_resp_init_ss_flags()`.

tempesta_fw/http_sess.c
```diff
@@ -59,15 +60,29 @@
  * @name		- name of sticky cookie;
  * @name_eq		- @name plus "=" to make some operations faster;
  * @sess_lifetime	- session lifetime in seconds;
+ * @max_misses		- maximum count of requsts with invalid cookie;
+ * @timeout		- maximum time to wait the request with valid cookie;
```
ikoveshnikov: Please, add the unit name for the member: jiffies or second. Now it's required to look though the code to identify it.

tempesta_fw/http_sess.c
```diff
@@ -59,15 +60,29 @@
...
 	u_int		enabled : 1,
 			enforce : 1;
 } TfwCfgSticky;
 
+
+/**
+ * Temporal storage for calculated redirection mark value.
+ */
+typedef struct {
+	unsigned int	att_no;
```
ikoveshnikov: Please, add members definitions.

tempesta_fw/http.c
```diff
@@ -438,6 +442,7 @@ tfw_http_prep_redirect(TfwHttpMsg *resp, unsigned short status, TfwStr *cookie,
 		ret |= tfw_http_msg_write(&it, resp, &proto);
```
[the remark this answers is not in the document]
ikoveshnikov: The comment shows important and existing issue, so it was moved to a separate issue #1058 . The fix probably will be backported to 0.5, so it's better to fix it in a new PR.

tempesta_fw/http.c
aleksostapenko: Functions `tfw_client_drop()` and `tfw_client_block()` are intended to be called only in client connection context, and only during request processing - before request was attached to connection's `seq_queue`. I've mentioned this in comment:
https://github.com/tempesta-tech/tempesta/blob/02aa662140907ffa58fa1f52f899c18aa21bc72a/tempesta_fw/http.c#L2389.
Functions `tfw_client_drop_close()` and `tfw_client_block_close()` are intended to be called in all other cases - when request has already been attached to connection's `seq_queue` - in server connection context the request is found right at the beginning of the response processing:
https://github.com/tempesta-tech/tempesta/blob/02aa662140907ffa58fa1f52f899c18aa21bc72a/tempesta_fw/http.c#L3349
So  the request is in the `seq_queue` and `BUG_ON` in `tfw_http_resp_fwd()` should not be triggered. 
The same behavior was present before - but in `tfw_http_srv_error_resp_and_log` function, which I removed to optimize the code.

> May be you need to set TFW_HTTP_B_CONN_CLOSE flag for a request if close == true

This behavior is already implemented in `tfw_http_req_mark_error()` via `TFW_HTTP_F_SUSPECTED` flag.

tempesta_fw/http_parser.c
```diff
@@ -170,20 +183,29 @@ do {									\
...
 	__fsm_n = __data_remain(p);					\
 	__fsm_sz = tfw_match_##alphabet(p, __fsm_n);			\
 	if (unlikely(__fsm_sz == __fsm_n)) {				\
 		/* Continue field processing on next skb. */		\
 		BUG_ON(!(field)->ptr);					\
-		__msg_field_fixup(field, data + len);			\
+		if (fixup_pos)						\
+			__msg_field_fixup_pos(field, p, __fsm_sz);	\
+		else							\
+			__msg_field_fixup(field, data + len);		\
```
keshonok: I don't think this is the right change. The role of `__msg_field_fixup()` is to cut off (stop) a chunk at specified address. Same is true for `__msg_field_finish()`, just with an extra action of setting `TFW_STR_COMPLETE` flag.

Under the circumstances it appears that `p + __fsm_sz` is the same end address and `data + len`. Thus there's no need for `fixup_pos` argument, and there's no need for the whole `__FSM_MATCH_MOVE_fixup_pos()` macro.

aleksostapenko: @keshonok Thank you for review.
The restriction of `__msg_field_fixup()` usage is that it always takes `data` pointer (i.e. the beginning of currently parsed part of message) as the beginning data chunk in `tfw_http_msg_add_str_data()`. This is suitable when we don't need to split the part of currently parsed message into separate `TfwStr` chunks (except of `TFW_POSTPONE` of course - but in this case the start `data` pointer is already set to the right place for next `TfwStr` chunk). But if we split the message via `__msg_field_fixup()` - then separate `TfwStr` chunks will point to the same `data` pointer in the real message, which is wrong behavior.  This can happen when we get to `Req_UriAbsPath` from `Req_UriMark`, and `req->uri_path` already has some part of message. So we need to explicitly specify the beginning pointer for the next chunk - via `__msg_field_fixup_pos()`.
Similar behavior we have with `__msg_hdr_chunk_fixup()`, but for headers parsing (e.g. in `__req_parse_cookie()` function).

tempesta_fw/http_parser.c
```diff
@@ -170,20 +183,29 @@ do {									\
...
 		__FSM_EXIT(TFW_POSTPONE);				\
 	}								\
 } while (0)
 
+#define __FSM_MATCH_MOVE_f(alphabet, to, field)				\
+	__FSM_MATCH_MOVE_fixup_pos(alphabet, to, field, false)
+
+#define __FSM_MATCH_MOVE_pos_f(alphabet, to, field)			\
+	__FSM_MATCH_MOVE_fixup_pos(alphabet, to, field, true)
+
```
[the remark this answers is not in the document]
aleksostapenko: Answered above.

tempesta_fw/http.c
```diff
@@ -429,6 +430,9 @@ tfw_http_prep_redirect(TfwHttpMsg *resp, unsigned short status, TfwStr *cookie,
 
 	ret = tfw_http_msg_write(&it, resp, rh);
 	ret = tfw_http_msg_write(&it, resp, &h_common_1);
+
+	if (rmark->len)
+		ret |= tfw_http_msg_write(&it, resp, rmark);
```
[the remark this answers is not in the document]
aleksostapenko: But, as I understand, discussion above has came to the conclusion, that redirection mark should be put before the whole URI - not before the URI path.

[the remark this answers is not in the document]
ikoveshnikov: The comment was about relative URI path, no host was listed there. With your case client will get
`Location: /HASH/http://example.com/some_uri` redirection. No browsers will reach your resource by this link. Please, try the real browser against your code, it should show the issue.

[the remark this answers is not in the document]
aleksostapenko: Yeah, you are right. I will correct this.

tempesta_fw/http_sess.c
```diff
@@ -402,51 +438,246 @@ tfw_http_sticky_add(TfwHttpResp *resp)
...
+	TFW_DBG("http_sess: calculate redirection mark: ts=%#lx(now=%#lx),"
+		" att_no=%#x\n", mv->ts, jiffies, mv->att_no);
+
+	if ((r = crypto_shash_init(shash_desc)))
+		return r;
+	r = crypto_shash_update(shash_desc, (u8 *)&mv->att_no, sizeof(mv->att_no));
+	if (r)
+		return r;
+	return crypto_shash_finup(shash_desc, (u8 *)&mv->ts, sizeof(mv->ts),
+				  mv->hmac);
```
krizhanovsky: It's better to use `crypto_shash_digest()` instead of the 3 calls.  Please update `__sticky_calc()` as well.

tempesta_fw/http_sess.c
```diff
@@ -402,51 +438,246 @@ tfw_http_sticky_add(TfwHttpResp *resp)
...
+ * - Secret key.
+ */
+static int
+__redir_hmac_calc(TfwHttpReq *req, RedirMarkVal *mv)
+{
+	int r;
+	char desc[sizeof(struct shash_desc)
+		  + crypto_shash_descsize(tfw_sticky_shash)]
+		  CRYPTO_MINALIGN_ATTR;
+	struct shash_desc *shash_desc = (struct shash_desc *)desc;
```
krizhanovsky: There is standard macro `SHASH_DESC_ON_STACK` for the 2 declarations. Please update `__sticky_calc()` as well.

tempesta_fw/http_sess.c
```diff
@@ -402,51 +438,246 @@ tfw_http_sticky_add(TfwHttpResp *resp)
...
+
+	return 1;
+}
+
+#define sess_warn(check, addr, fmt, ...)				\
+do {									\
+	char abuf[TFW_ADDR_STR_BUF_SIZE] = {0};				\
+	tfw_addr_fmt_v6(&(addr)->v6.sin6_addr, 0, abuf);		\
+	TFW_WARN("http_sess: %s for %s" fmt, check, abuf, ##__VA_ARGS__); \
+} while (0)
```
krizhanovsky: This is a copypaste from `frang_msg()`, piece of code from `tfw_sock_srv_connect_try_later()`, and probably we have mode similar warnings with `tfw_addr_fmt_v6()` calls. I thinks it'd be better to define a common log macro for this.

tempesta_fw/http_sess.c
```diff
@@ -402,51 +438,246 @@ tfw_http_sticky_add(TfwHttpResp *resp)
...
+		tr = c->ptr;						\
+	for ( ; c < end; ++c) {						\
+		for ( ; tr < (unsigned char *)c->ptr + c->len; ++tr) {	\
+			if (count++ == sizeof((obj)->f) * 2)		\
+				goto end_##f;				\
+			(obj)->f = ((obj)->f << 4) + hex_to_bin(*tr);	\
+		}							\
+	}								\
+end_##f:								\
+	count;								\
```
krizhanovsky: It seems the result value of the statement is never used

tempesta_fw/http_sess.c
```diff
@@ -402,51 +438,246 @@ tfw_http_sticky_add(TfwHttpResp *resp)
...
+/*
+ * Calculate HMAC value for redirection mark.
+ *
+ * HMAC value is based on:
+ * - Number of attempts for session establishing;
+ * - Timestamp of first attempt;
+ * - Secret key.
+ */
+static int
+__redir_hmac_calc(TfwHttpReq *req, RedirMarkVal *mv)
```
krizhanovsky: `req` isn't used

tempesta_fw/client.h
```diff
@@ -32,6 +32,7 @@
  * @conn_users		- connections reference counter.
  * 			  The client is released, when the counter reaches zero:
  * 			  no connections to the server - no client for us :)
+ * @miss_count		- count of cookie misses in requests;
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/http.c
```diff
@@ -418,6 +418,7 @@ tfw_http_prep_redirect(TfwHttpMsg *resp, unsigned short status, TfwStr *cookie,
 
 	/* Add variable part of data length to get the total */
 	data_len = rh->len + h_common_1.len;
+	data_len += rmark->len ? rmark->len : 0;
```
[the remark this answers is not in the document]

tempesta_fw/http.c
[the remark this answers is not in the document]

tempesta_fw/http_sess.c
aleksostapenko: Done.

tempesta_fw/http_sess.c
aleksostapenko: Done.

tempesta_fw/http_sess.c
```diff
@@ -867,13 +1112,30 @@ static int
...
+			TFW_ERR_NL("%s: invalid value for 'max_misses'"
+				   " attribute: '%s'\n", cs->name, val);
+			return -EINVAL;
+		} else if (!strcasecmp(key, "timeout") &&
+			   tfw_cfg_parse_uint(val, &tfw_cfg_sticky.timeout))
+		{
+			TFW_ERR_NL("%s: invalid value for 'timeout' attribute:"
+				   " '%s'\n", cs->name, val);
+			return -EINVAL;
+		}
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/http_sess.c
```diff
@@ -136,22 +151,60 @@ tfw_http_sticky_redirect_allied(TfwHttpReq *req)
...
-tfw_http_sticky_send_redirect(TfwHttpReq *req, StickyVal *sv)
+tfw_http_sticky_send_redirect(TfwHttpReq *req, StickyVal *sv, RedirMarkVal *mv)
 {
 	unsigned long ts_be64 = cpu_to_be64(sv->ts);
-	TfwStr chunks[3], cookie = { 0 };
+	TfwStr c_chunks[3], m_chunks[4], cookie = { 0 }, rmark = { 0 };
 	DEFINE_TFW_STR(s_eq, "=");
 	TfwHttpResp *resp;
-	char buf[sizeof(*sv) * 2];
+	char c_buf[sizeof(*sv) * 2], m_buf[sizeof(*mv) * 2];
```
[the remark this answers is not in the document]
aleksostapenko: This function was introduced to expose the size of `RedirMarkVal` structure for `http_parser`. I think that inside of `http_sess.c` it would be more appropriate to depend on the size of the variables than on the structure itself.

tempesta_fw/http_sess.c
aleksostapenko: Yes, but I choose Frang-like approach, since if we used `0` as working value of `tfw_cfg_sticky.max_misses`, then we would block the first redirected request (i.e. we would block all requests) which doesn't make sense, I think.

tempesta_fw/http_sess.c
aleksostapenko: Yes, but `gcc` does not allow labels at the end of compound statement.

tempesta_fw/http_parser.c
aleksostapenko: Done.

tempesta_fw/http_sess.c
aleksostapenko: Function `crypto_shash_finup()` is the short-hand itself (for `crypto_shash_update()` and `crypto_shash_final()`), so in fact we have 4 calls here (including two calls of `crypto_shash_update()` function - for two input values), but `crypto_shash_digest()` function is suitable for one input value only. The same situation in `__sticky_calc()`.

etc/tempesta_fw.conf
```diff
@@ -740,6 +740,12 @@
 # redirects the client to the same URI, and prompts that Tempesta
 # sticky cookie is set in the request.
 #
+# max_misses - sets the maximum count of requests without cookie.
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/http_parser.c
```diff
@@ -210,6 +232,12 @@ do {									\
 	TFW_STR_CURR(&msg->parser.hdr)->flags |= flag;		  	\
 } while (0)
 
+#define __FSM_I_field_chunk_flags(field, flag)				\
+do {									\
+	TFW_DBG3("parser: add chunk flags: %u\n", flag);		\
+	TFW_STR_CURR(field)->flags |= flag;		  	\
+} while (0)
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/http_parser.c
```diff
@@ -727,6 +759,31 @@ enum {
...
+	if (__fsm_n > 0) {						\
+		if (chunk->len == str->len) {				\
+			lambda;						\
+			TRY_STR_INIT();					\
+			__FSM_I_MOVE_fixup_f(state, __fsm_n, field, 0);	\
+		}							\
+		__msg_field_fixup_pos(field, p, __fsm_n);		\
+		finish;							\
+		return CSTR_POSTPONE;					\
+	}
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/http_sess.c
aleksostapenko: Done.

tempesta_fw/http_sess.c
aleksostapenko: Done.

tempesta_fw/http_sess.c
aleksostapenko: Corrected.

tempesta_fw/http_sess.c
krizhanovsky: Can you then just use empty statement `;` instead?

tempesta_fw/http.c
```diff
@@ -2341,85 +2351,92 @@ tfw_http_req_mark_error(TfwHttpReq *req, int status)
...
-}
 
-static void
-tfw_http_srv_error_resp_and_log(bool reply, bool nolog, TfwHttpReq *req,
-				int status, const char *msg)
-{
-	if (!nolog)
-		TFW_WARN_ADDR(msg, &req->conn->peer->addr);
+	if (close)
+		ss_close_sync(req->conn->sk, true);
```
ikoveshnikov: If `reply == false` and `close == true`, the client connection will be closed and all requests in it will be discarded. Probably that is not a desired behaviour for ` tfw_client_drop_close()`: we would like to normally respond to all the previous requests, and close the connection after that.  I've already done some improvements  here: https://github.com/tempesta-tech/tempesta/pull/1012/files#diff-0d15717c2e20453f4271befe0024c2d4R2484 But there still some questions to discus. I would suggest to discuss it in the chat and probably postpone the change for the future PR.

tempesta_fw/sock_srv.c
```diff
@@ -237,12 +237,10 @@ tfw_sock_srv_connect_try_later(TfwSrvConn *srv_conn)
 		if (srv_conn->recns == ARRAY_SIZE(tfw_srv_tmo_vals)
 		    || !(srv_conn->recns % 60))
 		{
-			char addr_str[TFW_ADDR_STR_BUF_SIZE] = { 0 };
-			tfw_addr_fmt_v6(&srv_conn->peer->addr.v6.sin6_addr,
-					0, addr_str);
-			TFW_WARN("Cannot establish connection with %s in %u"
-				 " tries, keep trying...\n",
-				 addr_str, srv_conn->recns);
+			TFW_WARN_MOD_ADDR6(sock_srv, "cannot establish"
```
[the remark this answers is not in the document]

tempesta_fw/http_sess.c
aleksostapenko: Yes, missed that - thanks.
Corrected.

tempesta_fw/http.c
aleksostapenko: Changes had been reverted.

PR #1055, opened 2018-08-14
Preserve correct fsm state when moving to a next state

### LINKED ISSUE #1032: GFSM move state function doesn't restore original FSM id
`tfw_gfsm_move()` function moves FSM descriptor to the new state and calls all the registered hooks. When hook is called, FSM descriptor is switched to the registred hook: FSM id and state are set from registered hook: 
 https://github.com/tempesta-tech/tempesta/blob/fb2529973da3f4ca03b4b7914ce4a1c15896a146/tempesta_fw/gfsm.c#L242

If the hook function doesn't move the state of it's FSM to `TFW_GFSM_HTTP_STATE(TFW_GFSM_STATE_LAST)`, the FSM descriptor is still at the hook FSM id, not on the FSM id the caller used.

Here is how the issue reproduces. Frang must be configured. `tfw_gfsm_move()` can be called multiple times during during `tfw_http_msg_process()` function. The `tfw_http_msg_process()` function is one of FSM hooks. When the  `tfw_gfsm_move()`  is called, registered frang hooks are called. If frang is not configured, it has no impact to HTTP FSM, but when configured and message is not fully parsed, FSM descriptor is switched to FRANG FSM, and the FSM stays in this state without returning to the HTTP FSM. So the next `tfw_gfsm_move()` call will have incorrect state.

There are one more issue in `tfw_gfsm_move()`. If highest priority hook's returned `TFW_POSTPONE`,  but there is no lower priority hooks, `TFW_PASS` is returned to caller instead of `TFW_POSTPONE`

ikoveshnikov:
fix #1032

### DIFF
--- a/tempesta_fw/gfsm.c
+++ b/tempesta_fw/gfsm.c
@@ -280,7 +280,7 @@ EXPORT_SYMBOL(tfw_gfsm_move);
 void
 tfw_gfsm_debug_state(TfwGState *st, const char *msg)
 {
-	TFW_DBG("%s: curr=%d:  on_stack=%d fsm_id=%d prio=%d state=%d",
+	TFW_DBG("%s: curr=%d: on_stack=%d fsm_id=%d prio=%d state=%d\n",
 		msg, st->curr, !!(FSM_STATE(st) & TFW_GFSM_ONSTACK), FSM(st),
 		PRIO(st), TFW_GFSM_STATE(st));
 }
--- a/tempesta_fw/gfsm.c
+++ b/tempesta_fw/gfsm.c
@@ -80,6 +80,9 @@
 #define __GFSM_FSM(_s)		(((_s) >> TFW_GFSM_FSM_SHIFT)	\
 				 & TFW_GFSM_FSM_MASK)
 #define FSM(s)			__GFSM_FSM(FSM_STATE(s))
+#define __GFSM_PRIO(_s)		(((_s) >> TFW_GFSM_PRIO_SHIFT)	\
+				 & TFW_GFSM_PRIO_MASK)
+#define PRIO(s)			__GFSM_PRIO(FSM_STATE(s))
 #define GFSM_HOOK_N		(TFW_GFSM_PRIO_N * TFW_GFSM_STATE_N)
 #define SET_STATE(s, x)						\
 do {								\
@@ -269,6 +272,20 @@ done:
 	return r;
 }
 EXPORT_SYMBOL(tfw_gfsm_move);
+
+#ifdef DEBUG
+/**
+ * Helper function to debug current fsm state.
+ */
+void
+tfw_gfsm_debug_state(TfwGState *st, const char *msg)
+{
+	TFW_DBG("%s: curr=%d:  on_stack=%d fsm_id=%d prio=%d state=%d",
+		msg, st->curr, !!(FSM_STATE(st) & TFW_GFSM_ONSTACK), FSM(st),
+		PRIO(st), TFW_GFSM_STATE(st));
+}
+EXPORT_SYMBOL(tfw_gfsm_debug_state);
+#endif
 
 /**
  * Register a hook which will be called with priority @prio when FSM @fsm_id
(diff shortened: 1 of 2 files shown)


tempesta_fw/gfsm.c
```diff
@@ -236,7 +240,7 @@ tfw_gfsm_move(TfwGState *st, unsigned short state, const TfwFsmData *data)
 		 * rather than fixed priority levels to avoid spinning in vain.
 		 */
 		if (!(hooks[p] & mask))
-			return TFW_PASS;
+			goto done;
 
 		/* Switch context to other FSM. */
 		fsm = tfw_gfsm_switch(st, state, p);
```
aleksostapenko: It seems that part of the problem is remain in `tfw_gfsm_switch()`:
https://github.com/tempesta-tech/tempesta/blob/97184f60a78b90ed0703b8c4c4edcefd8290c794/tempesta_fw/gfsm.c#L145
if we have more than one hooks registered for certain FSM state, we get invalid behavior in both cases - whether the previous hook gained `TFW_GFSM_STATE_LAST` or not: `FSM(st)` will access to the FSM of previous hook (or `-1`).
Maybe
```
int fsm_curr = state >> TFW_GFSM_FSM_SHIFT;
int fsm_next = fsm_hooks[fsm_curr][shift].fsm_id;
```
will be more appropriate in this place.

tempesta_fw/gfsm.c
```diff
@@ -236,7 +240,7 @@ tfw_gfsm_move(TfwGState *st, unsigned short state, const TfwFsmData *data)
 		 * rather than fixed priority levels to avoid spinning in vain.
 		 */
 		if (!(hooks[p] & mask))
```
aleksostapenko: Just a refinement: as I understand, for now we should avoid to register hooks with priorities other then `TFW_GFSM_HOOK_PRIORITY_ANY`, since if we register e.g. three hooks for one state with priorities 0, 4, 5 - then hooks with 4 and 5 priorities will never be called.

tempesta_fw/gfsm.c
```diff
@@ -260,11 +265,28 @@ tfw_gfsm_move(TfwGState *st, unsigned short state, const TfwFsmData *data)
 			r = TFW_POSTPONE;
```
aleksostapenko: Not regarding this PR, but losely related with GFSM question: if one of HTTP FSM hooks returned `TFW_POSTPONE`
and e.g. in `tfw_http_req_process()`:
https://github.com/tempesta-tech/tempesta/blob/97184f60a78b90ed0703b8c4c4edcefd8290c794/tempesta_fw/http.c#L2681
we get `data_off == skb_len`, then `tfw_http_req_process()` will return `TFW_POSTPONE` too and lower-level processing will close connection. Is this valid behavior?

ikoveshnikov: No, but this is covered in #1012

tempesta_fw/gfsm.c
ikoveshnikov: Correct. At the moment we must plan the hook priorities at programming stage. It's not good to walk over all unused priorities at hot path. We need to extend `tfw_gfsm_register_hook()` to build dense array of hooks.

tempesta_fw/gfsm.c
ikoveshnikov: Good catch! Fixed.

tempesta_fw/gfsm.c
```diff
@@ -260,11 +265,28 @@ tfw_gfsm_move(TfwGState *st, unsigned short state, const TfwFsmData *data)
...
 EXPORT_SYMBOL(tfw_gfsm_move);
 
+#ifdef DEBUG
+/**
+ * Helper function to debug current fsm state.
+ */
+void
+tfw_gfsm_debug_state(TfwGState *st, const char *msg)
+{
+	TFW_DBG("%s: curr=%d:  on_stack=%d fsm_id=%d prio=%d state=%d",
```
krizhanovsky: Extra space after `:` and missed `\n` at the end of the format string.

tempesta_fw/gfsm.c
krizhanovsky: Yeah, really good catch - it takes time to realize the issue.

PR #1060, opened 2018-09-04
Fix #1058: Protocol/port persistence support during redirection.

### LINKED ISSUE #1058: Sticky cookie module can redirect user to insecure protocol or other server
As noticed in https://github.com/tempesta-tech/tempesta/pull/1047#discussion_r210718508 `Location: ` header in redirect response is filled with _http_ protocol. The module doesn't honour original client protocol. E.g.:

1. A client opens secure connection (_httpS_);
2. The client sends a request;
3. Sticky cookie module redirects the client to the same resource, but protocol is hardcoded to be _http_;
4. The client opens a new insecure connection and requests the same resource.

This means that the sticky module breaks secure connections. Thus I mark the issue as crucial.

How to reproduce:
1. Configure tempesta to use both HTTP and HTTPS:
```
server 127.0.0.1:8080;
listen 80;
listen 443 proto=https;
tls_certificate /etc/tfw-root.crt;
tls_certificate_key /etc/tfw-root.key;
```
2. Use curl to send a request:
```
curl -vkL -b cookies.txt https://192.168.122.12/
```
3. Use wireshark to see the packet flow.

Not only protocol selection is broken. If non-default port is used, it's not added to the `Location:` header. E.g. `GET http://natsys-lab.com:8080/cgi-bin/show.pl HTTP/1.1\r\n` request will be redirected to `http://natsys-lab.com/cgi-bin/show.pl`. Nobody guaranties `natsys-lab.com:8080` and `natsys-lab.com` to be the same resource.

aleksostapenko:


tempesta_fw/http.h
```diff
@@ -484,6 +485,7 @@ struct tfw_http_req_t {
 	TfwHttpCond		cond;
 	TfwStr			userinfo;
 	TfwStr			host;
+	TfwStr			port;
```
ikoveshnikov: I've tried to see, why `req->host` doesn't include port part, but it still unclear for me. We should add port subcomponent into `req->host`. This will improve our code a little bit:
- there is frang check that host in uri and 'Host: ' header are equal. The check fails since `req->host` doesn't include port component.
- the http chains rules can use host + port for routing the requests, not host name only
- code in  `tfw_http_prep_redirect()` would be a little bit simpler.

tempesta_fw/http.c
```diff
@@ -391,6 +407,7 @@ tfw_http_prep_redirect(TfwHttpMsg *resp, unsigned short status, TfwStr *cookie,
 	static TfwStr c_len_crlf = {
 		.ptr = S_REDIR_C_LEN, .len = SLEN(S_REDIR_C_LEN) };
 	TfwStr host, *rh, *cookie_crlf = &crlf, *r_end;
```
ikoveshnikov: Not an issue of this PR, but there is one issue more. According to [RFC 7230](https://tools.ietf.org/html/rfc7230#section-5.4):
>  When a proxy receives a request with an absolute-form of
   request-target, the proxy MUST ignore the received Host header field
   (if any) and instead replace it with the host information of the
   request-target.  

That means `req->host` must take precedence over `&req->h_tbl->tbl[TFW_HTTP_HDR_HOST]`, but in a few lines below we use 'Host :' header as primary value and `req->host` as fallback value.

tempesta_fw/http.c
```diff
@@ -434,9 +452,13 @@ tfw_http_prep_redirect(TfwHttpMsg *resp, unsigned short status, TfwStr *cookie,
 	 * See RFC 1945 9.3 and RFC 7231 7.1.2.
 	 */
 	if (host.len) {
-		static TfwStr proto = { .ptr = S_HTTP, .len = SLEN(S_HTTP) };
 		ret |= tfw_http_msg_write(&it, resp, &proto);
 		ret |= tfw_http_msg_write(&it, resp, &host);
+		if (req->port.len) {
+			DEFINE_TFW_STR(s_col, ":");
+			ret |= tfw_http_msg_write(&it, resp, &s_col);
+			ret |= tfw_http_msg_write(&it, resp, &req->port);
```
ikoveshnikov: Imagine that uri is provided in absolute form (with host information). in this case:
```
req->host == 'example.com'
req->port == '8080'
val(req->h_tbl->tbl[TFW_HTTP_HDR_HOST]) == 'example.com:8080'

=>
host == val(req->h_tbl->tbl[TFW_HTTP_HDR_HOST]) == 'example.com:8080'
```
That means extra `req->port` is added to redirect uri leading to `example.com:8080:8080`.  As for me, `req->host` should contain port subcomponent, and `req->port` wouldn't be needed at all.

ikoveshnikov: Please, add an issue to cover all possible cases in functional tests

tempesta_fw/http.c
```diff
@@ -391,6 +407,7 @@ tfw_http_prep_redirect(TfwHttpMsg *resp, unsigned short status, TfwStr *cookie,
 	static TfwStr c_len_crlf = {
 		.ptr = S_REDIR_C_LEN, .len = SLEN(S_REDIR_C_LEN) };
 	TfwStr host, *rh, *cookie_crlf = &crlf, *r_end;
+	DEFINE_TFW_STR_PROTO(proto, req->conn);
```
ikoveshnikov: It's also possible to create statically defined array `proto[2]` to contain both `http://` and `https://`, and use `TFW_CONN_PROTO(conn) == TFW_FSM_HTTPS` as the array index.

tempesta_fw/http.h
aleksostapenko: I supposed, that we may need a separate value in future for possible checks, but yes - I agree that in context of single-parsed host/port value for `Host` header - it will be more convenient to have host/port value in the single field for URI too. 
I will change this part.

tempesta_fw/http.c

tempesta_fw/http.c

tempesta_fw/http.h
aleksostapenko: Done.

tempesta_fw/http.c
```diff
@@ -343,6 +344,21 @@ unsigned long tfw_hash_str(const TfwStr *str);
 #define S_REDIR_KEEP	S_CRLF S_F_CONNECTION S_V_CONN_KA S_CRLF
 #define S_REDIR_CLOSE	S_CRLF S_F_CONNECTION S_V_CONN_CLOSE S_CRLF
 #define S_REDIR_C_LEN	S_F_CONTENT_LENGTH "0" S_CRLFCRLF
+
+#define DEFINE_TFW_STR_PROTO(name, conn)	\
+	TfwStr name;				\
+	name.skb = NULL;			\
```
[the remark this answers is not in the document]
aleksostapenko: This code removed.

tempesta_fw/http.c
```diff
@@ -343,6 +344,21 @@ unsigned long tfw_hash_str(const TfwStr *str);
...
+	name.eolen = name.flags = 0;		\
+	switch (TFW_CONN_PROTO(conn)) {		\
+	case TFW_FSM_HTTPS:			\
+		name.ptr = S_HTTPS;		\
+		name.len = SLEN(S_HTTPS);	\
+		break;				\
+	default:				\
+		name.ptr = S_HTTP;		\
+		name.len = SLEN(S_HTTP);	\
```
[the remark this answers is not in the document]
aleksostapenko: This code removed.

tempesta_fw/http.c

PR #1069, opened 2018-09-27
Correct link to patchset

craig:


README.md
```diff
@@ -14,7 +14,7 @@ performance characteristics in comparison with TCP servers on top of common
 Socket API or even kernel sockets.
 
 We do our best to keep the kernel modifications as small as possible. Current
-[patch](https://github.com/tempesta-tech/tempesta/blob/master/linux-4.9.35.patch)
+[patch](https://github.com/tempesta-tech/linux-4.14.32-tfw)
```
ikoveshnikov: Thank you for notice! But we should keep the link to the patch located in the current repository. You're right, we use [tempesta-tech/linux-4.14.32-tfw](https://github.com/tempesta-tech/linux-4.14.32-tfw) repo to develop the kernel patch required for Tempesta. But the patch itself is small and can be rather easily ported to other kernel versions, e.g. Debian doesn't use the vanilla kernel. That's why the link is for the patch located in the current repository. 

So the correct link will be `https://github.com/tempesta-tech/tempesta/blob/master/linux-4.14.32.patch`. Could you please fix this?

PR #1072, opened 2018-09-28
 Fix #806: Reduce count of IPI during work_queue processing.

### LINKED ISSUE #806: Reduce raised IPIs
Currently each task put into a work queue for a different CPU raises IPI to wake up ksoftirqd on the CPU. It seems IPIs are relatively expensive (?) and there is sense to raise IPI only if ksoftirqd on the designated CPU is inactive. It seems we can just wakeup appropriate ksoftirqd just like `wakeup_softirqd()` does it.

Also consider integration with [Flow Director](https://software.intel.com/en-us/articles/setting-up-intel-ethernet-flow-director) - probably we can get some benefits from it in our intra-CPU communication at all. However, the main point for the intra-CPU transport is that we have to communicate via pair of sockets, client and server, and the pairs aren't stable. (Are they?)

aleksostapenko:


tempesta_fw/cache.c
```diff
@@ -1743,10 +1742,22 @@ static void
...
+	TfwRBQueue *wq = &ct->wq;
 	TfwCWork cw;
 
-	while (!tfw_wq_pop(&ct->wq, &cw))
+	while (!tfw_wq_pop(wq, &cw))
 		tfw_cache_do_action(cw.msg, cw.action);
+
+	set_bit(TFW_QUEUE_B_IPI, &wq->flags);
+
+	smp_mb__after_atomic();
```
krizhanovsky: OK, here we set the IPI flag before checking the queue size and we check queue size before the flag in `tfw_wq_push()` - the synchronization seems safe. We need the barrier to make the queue size and the flag dependant. However, it seems the [barrier pairing rule](https://github.com/tempesta-tech/linux-4.14.32-tfw/blob/master/Documentation/memory-barriers.txt#L932) is broken here: strictly speaking, there is no guarantee that the queue operation won't be reordered with the flag testing in `tfw_wq_push()`. I see that `__tfw_wq_push()` uses at least atomic set, but it seems we still need an `_after_atomic()` barrier in `tfw_wq_push()` now.

tempesta_fw/sock.c
```diff
@@ -1358,11 +1362,24 @@ ss_tx_action(void)
...
-		raise_softirq(NET_TX_SOFTIRQ);
+	if (budget) {
+		set_bit(TFW_QUEUE_B_IPI, &wq->flags);
+
+		smp_mb__after_atomic();
+
+		if (!ss_close_q_sz())
+			return;
+
+		clear_bit(TFW_QUEUE_B_IPI, &wq->flags);
```
krizhanovsky: This is a copy of part of `tfw_wq_tasklet()` - maybe to make a macro with the queue size function as a parameter or a function accepting the queue size callback?

tempesta_fw/sock.c
```diff
@@ -208,13 +211,14 @@ ss_turnstile_push(long ticket, SsWork *sw, int cpu)
 	cn->ticket = ticket;
 	memcpy(&cn->sw, sw, sizeof(*sw));
 	spin_lock_bh(&cb->lock);
-	list_add(&cn->list, &cb->head);
+	list_add_tail(&cn->list, &cb->head);
 	cb->size++;
 	if (cb->turn > ticket)
 		cb->turn = ticket;
 	spin_unlock_bh(&cb->lock);
```
krizhanovsky: OK, we don't need a barrier after spinlock operation, but having a comment here would be good in case if we remove the spinlock.

tempesta_fw/cache.c
aleksostapenko: Corrected.

tempesta_fw/sock.c
aleksostapenko: Done.

tempesta_fw/sock.c
aleksostapenko: Done.

tempesta_fw/work_queue.h
```diff
@@ -77,6 +86,11 @@ tfw_wq_push(TfwRBQueue *q, void *ptr, int cpu, struct irq_work *work,
 	long ticket = __tfw_wq_push(q, ptr);
 	if (unlikely(ticket))
 		return ticket;
+	/*
+	 * The atomic operation is 'atomic64_cmpxchg()' in
+	 * '__tfw_wq_push()' above.
```
krizhanovsky: Why? `atomic64_cmpxchg()` is called very rarely, only on slow path. In fast path there are just atomic reads and writes: in good case there is write barrier, but in bad case there is no any barriers.

aleksostapenko: Fast path take place only when the queue is full and `tfw_wq_push` exits with ticket - before IPI generation. In case of insertion new element into the queue - the `atomic64_cmpxchg()` must be called.

tempesta_fw/work_queue.h
```diff
@@ -38,13 +38,37 @@ typedef struct {
 	long			last_head;
 	atomic64_t		head ____cacheline_aligned;
 	atomic64_t		tail ____cacheline_aligned;
+	unsigned long		flags;
 } TfwRBQueue;
 
+enum {
+	/* Enable IPI generation. */
+	TFW_QUEUE_B_IPI = 0
```
ikoveshnikov: Probably we don't need `_B_` suffix meaning `bit` here. For other flags we historically used `_B_` and `_F_` suffixes to distinguish between flags and bits, but now we tend to use common bitmap API instead of 'manual' bit operations. Removing of `_B_` suffix will synchronise the code with upcoming patches.

aleksostapenko: Corrected.

Issue #1074, opened 2018-10-05
cannot find appropriate virtual host


krizhanovsky:
I get `Warning: request dropped: cannot find appropriate virtual host, status 404` for the configuration at the below:
```
listen 80;
server 127.0.0.1:8080;
cache 1;
cache_fulfill * *;
sticky enforce;
js_challenge resp_code=503 delay_min=1000 delay_range=1000 delay_limit=3000 /root/master/etc/js_challenge.html;
```
Apache HTTPD with tempesta-tech.com is running in the same VM on 8080. The traffic is (I use Chrome for the request):
```
20:28:44.834334 IP 192.168.100.1.40380 > 192.168.100.4.80: Flags [P.], seq 917067469:917067944, ack 621472777, win 229, options [nop,nop,TS val 1259675188 ecr 3762659891], length 475: HTTP: GET / HTTP/1.1
E.....@.@.l...d...d....P6.V.%
.	...........
K..4.E.3GET / HTTP/1.1
Host: debian
Connection: keep-alive
Cache-Control: max-age=0
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/63.0.3239.132 Safari/537.36
Upgrade-Insecure-Requests: 1
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
Accept-Encoding: gzip, deflate
Accept-Language: en-US,en;q=0.9,ru;q=0.8
Cookie: _ga=GA1.1.1129566486.1538759973; _gid=GA1.1.1088249423.1538759973


20:28:44.843466 IP 192.168.100.4.80 > 192.168.100.1.40380: Flags [.], ack 917067944, win 235, options [nop,nop,TS val 3762659903 ecr 1259675188], length 0
E..4..@.@.8...d...d..P..%
.	6.X......;.....
.E.?K..4
20:28:44.843737 IP 192.168.100.4.80 > 192.168.100.1.40380: Flags [P.], seq 621472777:621472910, ack 917067944, win 235, options [nop,nop,TS val 3762659904 ecr 1259675188], length 133: HTTP: HTTP/1.1 404 Not Found
E.....@.@.8...d...d..P..%
.	6.X............
.E.@K..4HTTP/1.1 404 Not Found
Date: Fri, 05 Oct 2018 17:28:44 GMT
Content-Length: 0
[...]

Issue #598, opened 2016-08-07
Frang: sessions rate limit


krizhanovsky:
Depends on #235.

To fight against non-cookie aware DDoS bots more efficiently Frang must implement new sessions rate limits: concurrent sessions from the peer, new sessions per time unit for the peer with burst control.

krizhanovsky:
It seems the limits can be implemented using connection and request rate limiting if an administrator doesn't want to enforce sticky cookies.

avbelov23:
Need drop the session, send a redirect request and require a new cookie for clients working through a proxy instead of closing the connection

krizhanovsky:
Comment https://github.com/tempesta-tech/tempesta/issues/598#issuecomment-463230068 is essentially about https://github.com/tempesta-tech/tempesta/issues/1045

krizhanovsky:
A good user experience in JS challenge requires soft and hard limits in #1102 , so mark the issue as crucial.

ikoveshnikov:
We account not only HTTP sessions but also endpoint client hidden  behind anonymous HTTP proxy. It makes sense to limit number of concurrent clients behind a proxy. Just to avoid unlimited population of the `client` database. Bit in the same time, each client behind proxy should receive it's own HTTP session and we don't need to introduce special limit for concurrent clients behind proxy.

Still it can be suspicious when multiple different clients behind proxy share the same sessions, no matter whether they're behind the same proxy or not.

PR #1077, opened 2018-10-10
Fix unstable JS challange passing

ikoveshnikov:
Related to JS challenge issue described in #1074 

- Synchronize computation of delay in JS Challange between JS and  Tempesta code
- Don't close client connection if unchallenged resources was requested before JS challenge is passed. Blocking of clients that makes too many requests with incorrect cookie  values is subject for #598 .

@krizhanovsky  @aleksostapenko Please, test this fix in your environment. I could reproduce the issue, but I couldn't reproduce the exact behaviour you had.

### DIFF
--- a/tempesta_fw/http_sess.c
+++ b/tempesta_fw/http_sess.c
@@ -127,10 +127,10 @@ static struct kmem_cache *sess_cache;
  * @body	- body (html with JavaScript code);
  * @delay_min	- minimal timeout client must wait before repeat the request,
  *		  in jiffies;
- * @delay_limit	- maximum timeout between client's requests to pass the
- *		  challenge, in jiffies;
- * @delay_range	- allowed time range for client to make a repeated request,
- *		  in msecs;
+ * @delay_limit	- maximum timeout between JS challenge generation and client's
+ *		  request to pass the challenge, in jiffies;
+ * @delay_range	- time interval starting after @delay_min for a client to make
+ *		  a repeated request, in msecs;
  * @st_code	- status code for response with JS challenge;
  */
 typedef struct {
@@ -860,6 +860,11 @@ tfw_http_sess_check_jsch(StickyVal *sv, TfwHttpReq* req)
 		return 0;
 
 	cur_time = jiffies;
+	/*
+	 * When a client calculates it's own random delay, it uses range value
+	 * encoded as msecs, we have to use the same, to have exactly the same
+	 * calculation results. See etc/js_challenge.js.tpl .
+	 */
 	min_time = sv->ts + tfw_cfg_js_ch->delay_min
 			+ msecs_to_jiffies(sv->ts % tfw_cfg_js_ch->delay_range);
 	max_time = sv->ts + tfw_cfg_js_ch->delay_limit;
--- a/tempesta_fw/http_sess.c
+++ b/tempesta_fw/http_sess.c
@@ -121,18 +121,23 @@ static struct kmem_cache *sess_cache;
 /**
  * JavaScript challenge.
  *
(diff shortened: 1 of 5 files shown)


tempesta_fw/http_sess.c
```diff
@@ -121,18 +121,23 @@ static struct kmem_cache *sess_cache;
...
- * @delay_min	- minimal timeout to make a client wait;
- * @delay_range	- allowed time range to recieve and accept client's session;
- * @delay_limit	- maximum difference between current time and a timestamp
- *		  specified in the sticky cookie;
+ * @delay_min	- minimal timeout client must wait before repeat the request,
+ *		  in jiffies;
+ * @delay_limit	- maximum timeout between client's requests to pass the
+ *		  challenge, in jiffies;
+ * @delay_range	- allowed time range for client to make a repeated request,
+ *		  in msecs;
```
aleksostapenko: Perhaps `maximum range for random timeout added to delay_min` or something like that would be more appropriate description for `delay_range` field.

tempesta_fw/http_sess.c
```diff
@@ -1261,32 +1272,28 @@ tfw_cfg_op_jsch_parse_resp_code(TfwCfgSpec *cs, const char *val)
...
 
-	if (!limit) {
+	if (!tfw_cfg_js_ch->delay_limit) {
 		tfw_cfg_js_ch->delay_limit = def_limit;
-
-		return;
 	}
-	if (limit < min_limit) {
+	else if (tfw_cfg_js_ch->delay_limit < min_limit) {
 		TFW_WARN_NL("%s: delay limit is too low, "
```
aleksostapenko: Why not just exit with configuration error here? It seems that `min_limit` is too small value for `tfw_cfg_js_ch->delay_limit` anyway, and will lead to rejections of legitimate clients.

tempesta_fw/http_sess.c
```diff
@@ -847,23 +852,30 @@ tfw_http_sess_remove(TfwHttpSess *sess)
...
-	time_t cur_time, min_time, max_time;
+	unsigned long cur_time, min_time, max_time;
 
 	if (!tfw_cfg_js_ch)
 		return 0;
 
 	cur_time = jiffies;
 	min_time = sv->ts + tfw_cfg_js_ch->delay_min
-			+ sv->ts % tfw_cfg_js_ch->delay_range;
+			+ msecs_to_jiffies(sv->ts % tfw_cfg_js_ch->delay_range);
```
aleksostapenko: I see, that this unusual timeout calculation is intended for consistency with calculations in `js_challenge.js.tpl`, but an explanatory comment would be appropriate here.

tempesta_fw/http_sess.c
ikoveshnikov: That was a requirement of the JS Challenge task: https://github.com/tempesta-tech/tempesta/issues/536#issuecomment-343276205

>  Configuration process must enforce the value to be greater than delay_range + delay_min and show warning if it's less than delay_range + delay_min + 100. Default value is delay_range + delay_min + 1000 (enough ever fro cross atlantic connections and relatively slow hardware and software).

The last moment of time browser can send a request - `delay_range + delay_min`, and some time is required to deliver the request. `min_limit` adds extra 0.1s. Yes, this configuration is not very reliable: delivering via cross-Atlantic is slower, browsers can be slow, etc., while default limit (+1s) is reliable enough. But user may want to use less or more strict time lines and we should honour his choices.

ikoveshnikov:
@i-rinat Thank you for comments! The PR is already merged, but I'll fix this in my next PR.

PR #1087, opened 2018-10-26
Backport JS Challange fixes to 0.5 release

ikoveshnikov:
All review comments applied to original patchset was merged into current patchset.


krizhanovsky:
OK, but please don't forget to increment the patch level digit in `TFW_VERSION` with each backported patch.

PR #1097, opened 2018-11-12
Fix #1073: Add support MSI/MSI-X interrupts during NIC queues configuration

### LINKED ISSUE #1073: For Mellanox NIC the tfw_lib.sh script does not assign RX queues to CPUs
This place https://github.com/tempesta-tech/tempesta/blob/871200e5c64c1016dfe4891f365bc78d910cb97c/scripts/tfw_lib.sh#L59 does not work with the Melanox card: in `/proc/interrupts`, its queues are named in driver-like style (not like the logical interface), so the next cycle in script is completely skipped and only eight queues remain (despite the previous `ethtool -L` command) with eight IRQs and their `smp_affinity` is not changed (except for periodic changes from `irqbalance`). Although all created 128 queues can be found in `/sys/class/net/enp7s0 /queues/` but only 8 remain assigned to CPUs. The interrupt numbers for the script can be found in `/sys/class/net/enp7s0/device/msi_irqs/`, but only eight queues/IRQs exist in that directory and have been assigned to CPUs; the rest of created queues, apparently, remain without IRQ numbers.

aleksostapenko:


scripts/tempesta.sh
```diff
@@ -43,7 +43,8 @@ tdb_mod=tempesta_db
 tfw_mod=tempesta_fw
 declare -r LONG_OPTS="help,load,unload,start,stop,restart,reload"
 
-declare devs=$(ip addr show up | awk '/^[0-9]+/ { sub(/:/, "", $2); print $2}')
+declare devs=$(ip addr show up | grep -P '^[0-9]+' | grep -Pv '\bLOOPBACK\b' \
+	       | awk '{ sub(/:/, "", $2); print $2}')
```
krizhanovsky: A comment https://github.com/tempesta-tech/tempesta/issues/1073#issuecomment-427800276 and/or later one, explaining why do we exclude loopback from RPS setup, would be useful here.

scripts/tfw_lib.sh
```diff
@@ -36,27 +36,32 @@ calc()
...
+	# Wait for the interface reconfiguration.
+	opstate="$TFW_NETDEV_PATH/$dev/operstate"
+	while [ "$(cat $opstate)" = "down" ]; do
+		sleep 1
+	done
+
+	dev_irqs_path="/sys/class/net/$dev/device/msi_irqs"
+	irqs=($(grep $dev /proc/interrupts | sed -e 's/\s*\|:.*//g'))
+	if [ -z "$irqs" -a -d $dev_irqs_path ]; then
+		irqs=($(ls $dev_irqs_path))
```
krizhanovsky: The things are going complex, so please write a short (as above) comment why do we need the 2 sources to learn interrupts.

scripts/tfw_lib.sh
```diff
@@ -36,27 +36,32 @@ calc()
...
 	dev=$1
-	RXQ_MAX=$(ethtool -l $dev 2>/dev/null \
-		  | grep -m 1 RX | sed -e 's/RX\:\s*//')
+	RXQ_MAX=$2
 
-	echo "...distribute $dev queues"
+	echo "...set rx channels to $RXQ_MAX, please wait..."
+	# Set maximum number of available channels for better
+	# packets hashing.
+	ethtool -L $dev rx $RXQ_MAX >/dev/null 2>&1
```
krizhanovsky: I tried to do this with hands
```
root@9011:~# ethtool -L enp7s0 rx 128
rx unmodified, ignoring
no channel parameters changed, aborting
current values: tx 128 rx 8 other 0 combined 0
```
So we have the issue #1073 - we can not set the required number of channels and we shall not just silently ignore the error. Instead, please print a warning message so we'll be able to catch issues like #1073 quicker in future.

krizhanovsky: By the way why does the command fails?

scripts/tempesta.sh
aleksostapenko: Added.

scripts/tfw_lib.sh
aleksostapenko: Comment added.

scripts/tfw_lib.sh
aleksostapenko: Yes - agree, error message should be added.
Command fails because the count of queues had already been set to 128 before the command invocation - so, repeated attempt for the same value is rejected.

aleksostapenko: Added error message.

scripts/tfw_lib.sh
```diff
@@ -27,12 +27,55 @@ declare -r TFW_NAME=`basename $0` # program name (comm name in ps)
...
+declare -a IRQS_GLOB_LIST
 
 calc()
 {
 	echo "$1" | bc -iq | tail -1
 }
 
+# Assigned IRQs can be reassigned by irqbalance. To avoid this, assigned vectors
+# should be added to irqbalance config with '--banirq' option.
+irqbalance_ban_irqs()
```
ikoveshnikov: The third-party configuration file is modified here, so we need to add the good comment to document that the `BAN_IRQ` variable is automatically generated by the Tempesta. Also there is no need to modify the systemd service file and call `systemd daemon-reload`. We should only modify `/etc/default/irqbalance` in the following way: 
```
# Generated by Tempesta to restrict rebalancing for NIC irqs.... 
TFW_BAN_IRQS="--banirq ..." 

IRQBALANCE_ARGS="$TFW_BAN_IRQS"
```
And don't forget the quotes around `BAN_IRQS` value.

scripts/tempesta.sh
```diff
@@ -208,6 +208,15 @@ stop()
 	echo "...unload Tempesta modules"
 	unload_modules
 
+	systemctl status irqbalance.service >/dev/null
```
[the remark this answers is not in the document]
aleksostapenko: Done.

PR #1127, opened 2018-12-06
Fix bugs affecting load balancers

ikoveshnikov:
1. Fix incorrect bit operation, leading to incorrect scheduler otions
2. Fix resetting of APM data. `Minimal response time` is initialised as `0`, and can't be updated with real values here:
https://github.com/tempesta-tech/tempesta/blob/8bd88c970348bb3f1cdaa3c70f30a3a89ac6376a/tempesta_fw/apm.c#L319-L327
This lead to incorrect balancing in `ratio [dynamic|predict] minimal`

### DIFF
--- a/tempesta_fw/sock_srv.c
+++ b/tempesta_fw/sock_srv.c
@@ -1536,7 +1536,7 @@ tfw_cfg_sg_ratio_verify(TfwSrvGroup *sg)
 	int count = 0;
 
 	if (sg->flags & (TFW_SG_F_SCHED_RATIO_DYNAMIC
-			 || TFW_SG_F_SCHED_RATIO_PREDICT))
+			 | TFW_SG_F_SCHED_RATIO_PREDICT))
 	{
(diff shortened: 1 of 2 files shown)


tempesta_fw/sock_srv.c
```diff
@@ -1536,7 +1536,7 @@ tfw_cfg_sg_ratio_verify(TfwSrvGroup *sg)
 	int count = 0;
 
 	if (sg->flags & (TFW_SG_F_SCHED_RATIO_DYNAMIC
-			 || TFW_SG_F_SCHED_RATIO_PREDICT))
+			 | TFW_SG_F_SCHED_RATIO_PREDICT))
```
krizhanovsky: good catch!

ikoveshnikov: This is actually was spotted by @i-rinat once. Seems I've lost the attribution in commit message. It was at some branch

Issue #687, opened 2017-02-20
Improve the architecture that supports the correct order of HTTP responses


keshonok:
The current architecture extensively uses the locking of `seq_queue` in client connections and `fwd_queue` in server connections to enforce the correct order of responses to HTTP requests. The locks are held for a relatively long time, and due to multi-CPU nature of Tempesta's operation there's significant lock contention in high load situations.

As the current architecture and its behavioural patters are understood better, it should be improved with the goal of decreasing the lock contention and improving the overall performance.

krizhanovsky:
The background of the issue is pull request https://github.com/tempesta-tech/tempesta/pull/660 and comments https://github.com/tempesta-tech/tempesta/pull/660#discussion_r101899999 and https://github.com/tempesta-tech/tempesta/pull/660#discussion_r101902213 in particular.

ikoveshnikov:
Currently request processing routine looks like:
1. NET_RX_SOFTIRQ
2. parse request and detach from client connection;
3. make all required modifications to request before forwarding;
4. assign request into forward_queue on some server connection;
5. Forward queue lock is acquired;
6. forward all unsent requests on that server connection;
7. for each request to be forwarded: push server connection socket to workqueue with `SS_Send` command;
8. Forward queue lock is released;
9. NET_TX_SOFTIRQ
10. pop tasks from work queue and complete them: send already prepared buffers to sockets.

In this processing pipeline two queues are used: forward_queue and workqueue. First doesn't have distinct consumer and producer and protected by spinlocks, the last has distinct producer and consumer and implemented as lock-free. Probably we can revise the algorithm and get rid of locking;

1. NET_RX_SOFTIRQ
2. parse request and detach from client connection;
3. make all required modifications to request before forwarding;
4. assign request into forward_queue on some server connection. lock-free operation;
5. push server connection socket to workqueue with `SS_Send` command;
6. NET_TX_SOFTIRQ
7. pop tasks from work queue and complete them: get server connection from socket;
8. forward all unsent requests on that server connection;
7. for each request to be forwarded: send already prepared buffers to socket, don't pop requests from the forwarding_queue

Here producer to the forward_queue and consumer are split and no locks for forward_queue are required. Roughly the same algorithm for response processing should be implemented: push client socket into work queue after response is adjusted and ready to sent. 

[...]

PR #1129, opened 2018-12-07
Fix #1003: Store requests for resending if connection became inactive.

### LINKED ISSUE #1003: server_forward_retries directive is ignored if request is forwarded to inactive server connection
When a request is sent to backend in `tfw_http_req_fwd_send()`, an error may happen in `ss_send()`, e.g. socket is inactive. In this case the request is dropped and `500` error is returned to client: 
https://github.com/tempesta-tech/tempesta/blob/036a0f10f81e4af11069ef0d3337041f764f8cec/tempesta_fw/http.c#L1103-L1104

You can see warnings in the system log:
```
[ 6627.198664] [tempesta] Warning: request dropped: forwarding error, status 500: ::ffff:c0a8:7a01
[ 6641.007429] [tempesta] Warning: request dropped: forwarding error, status 500: ::ffff:c0a8:7a01
[ 6645.200972] [tempesta] Warning: request dropped: forwarding error, status 500: ::ffff:c0a8:7a01
[ 6668.633026] [tempesta] Warning: request dropped: forwarding error, status 500: ::ffff:c0a8:7a01
[ 6674.709124] [tempesta] Warning: request dropped: forwarding error, status 500: ::ffff:c0a8:7a01
[ 6676.332363] [tempesta] Warning: request dropped: forwarding error, status 500: ::ffff:c0a8:7a01

```

The `tfw_http_req_fwd_send()` function doesn't honour `server_forward_retries` directive and drops the request on very first `ss_send()` failure. Instead Tempesta must increment message retry count on failure and drop request only in `tfw_http_req_evict_retries()` function.
https://github.com/tempesta-tech/tempesta/blob/036a0f10f81e4af11069ef0d3337041f764f8cec/tempesta_fw/http.h#L499

aleksostapenko:


tempesta_fw/http.c
```diff
@@ -1281,7 +1296,7 @@ tfw_http_conn_resend(TfwSrvConn *srv_conn, bool first, struct list_head *eq)
 	{
 		if (tfw_http_req_evict(srv_conn, srv, req, eq))
 			continue;
-		if (!tfw_http_req_fwd_send(srv_conn, srv, req, eq))
+		if (tfw_http_req_fwd_send(srv_conn, srv, req, eq, true))
```
ikoveshnikov: By the way, why `tfw_http_conn_resend()` and `tfw_http_conn_fwd_unsent()` are so different? Seems like both should has the same role: send requests one-by-one from the queue until error happen. As for me `tfw_http_conn_resend()` should also be able to stop at the first error. It's also affected by the issue.

krizhanovsky: This place confuses me. Suppose there is a first retry to resend a request while we have say 5 retries in configuration. This try fails. `tfw_http_req_fwd_send()` because of `drain == true` adds the request to error queue `eq` and next we send an error message to a user. Why? For me it seems that we have the same scenario which the patch tries to fix. If `tfw_http_req_evict()` didn't "eat" the request then probably it has many tries...

tempesta_fw/http.c
```diff
@@ -1089,25 +1089,27 @@ tfw_http_req_evict(TfwSrvConn *srv_conn, TfwServer *srv, TfwHttpReq *req,
...
- * move it to the error queue @eq for sending an error response later.
- *
- * TODO: Perhaps, there's a small optimization. Ultimately, the thread
- * ends up in ss_send(). In some cases a connection is still active when
- * it's obtained, but not active by the time the thread is in ss_send().
- * In that case -EBADF is returned, and nothing destructive happens to
- * the request. So, perhaps, instead of sending an error in that case
- * these unlucky requests can be re-sent when the connection is restored.
+ * If forwarding of @req to server @srv_conn is not successful, requests
+ * should be leaved in @fwd_queue (to resend them to backend later) if
```
i-rinat: typo: `leaved` → `left`

tempesta_fw/http.c
```diff
@@ -1089,25 +1089,27 @@ tfw_http_req_evict(TfwSrvConn *srv_conn, TfwServer *srv, TfwHttpReq *req,
 	       || tfw_http_req_evict_retries(srv_conn, srv, req, eq);
 }
 
+#define CONN_ACTIVE(c)	((c) != -EBADF)
```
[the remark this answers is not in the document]

tempesta_fw/sock.c
```diff
@@ -475,14 +475,8 @@ ss_send(struct sock *sk, struct sk_buff **skb_head, int flags)
...
-	if (ss_wq_push(&sw, cpu)) {
-		TFW_DBG2("Cannot schedule socket %p for transmission"
-			 " (queue size %d)\n", sk,
-			 tfw_wq_size(&per_cpu(si_wq, cpu)));
-		sock_put(sk);
-		r = -EBUSY;
-		goto err;
-	}
+	while (ss_wq_push(&sw, cpu))
+		;
```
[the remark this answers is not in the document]
aleksostapenko: Added rescheduling of requests in case of `-EBUSY` error.

tempesta_fw/http.c

tempesta_fw/http.c
aleksostapenko: Added storing and rescheduling for requests re-sending procedure.

tempesta_fw/connection.h
```diff
@@ -284,6 +289,55 @@ tfw_srv_conn_hasnip(TfwSrvConn *srv_conn)
...
 
+/*
+ * Tell if connection is temporary inactive due to full work queue.
+ */
+static inline bool
+tfw_srv_conn_busy(TfwSrvConn *conn)
+{
+	if (test_bit(TFW_CONN_B_BUSY, &conn->flags))
+		return true;
+	rmb();
```
keshonok: What is the purpose of read memory barrier here?

tempesta_fw/connection.h
```diff
@@ -284,6 +289,55 @@ tfw_srv_conn_hasnip(TfwSrvConn *srv_conn)
...
+}
+
+/*
+ * Make connection temporary inactive due to full work queue.
+ */
+static inline void
+tfw_srv_conn_set_busy(TfwSrvConn *conn)
+{
+	if (!test_bit(TFW_CONN_B_BUSY, &conn->flags))
+		set_bit(TFW_CONN_B_BUSY, &conn->flags);
```
keshonok: Why test the bit first when the goal is to just set it? There's nothing here that depends on the bit NOT set before. Also, this sequence doesn't even provide this guarantee as it's not atomic.

tempesta_fw/connection.h
```diff
@@ -284,6 +289,55 @@ tfw_srv_conn_hasnip(TfwSrvConn *srv_conn)
...
+}
+
+/*
+ * Clear busy bit for server connection (if set).
+ */
+static inline void
+tfw_srv_conn_clear_busy(TfwSrvConn *conn)
+{
+	if (test_bit(TFW_CONN_B_BUSY, &conn->flags))
+		clear_bit(TFW_CONN_B_BUSY, &conn->flags);
```
keshonok: Why test the bit first when the goal is to just clear it? There's nothing here that depends on the bit being set before. Also, this sequence doesn't even provide this guarantee as it's not atomic.

tempesta_fw/connection.h
```diff
@@ -284,6 +289,55 @@ tfw_srv_conn_hasnip(TfwSrvConn *srv_conn)
...
+ * Set small delay for inactivity of busy connection to give time for
+ * unloading of the corresponding work queue.
+ */
+static inline void
+tfw_srv_clear_busy_delay(TfwSrvConn *conn, bool busy)
+{
+	if (busy)
+		WRITE_ONCE(conn->jbusytstamp, jiffies + msecs_to_jiffies(30));
+
+	smp_mb__before_atomic();
```
keshonok: It would be nice to have a comment that explains the logic of memory barriers in this set of functions.

aleksostapenko:
Addressing @ikoveshnikov comment:

>First, about issues in the PR. tfw_http_req_resched() is called, when the srv_conn->fwd_qlock is acquired. A request can be rescheduled to the same connection, so deadlock may happen.

The `TFW_CONN_B_BUSY` bit is intentionally set for connection at very beginning of rescheduling procedure on EBUSY (in `tfw_http_req_fwd_busy_resched`), so current connection is evicted from subsequent rescheduling and deadlock is avoided.

>Request processing become messy and infinite loops are possible.

>loop 3-4-3 is possible.

The possibility of infinite recursive rescheduling is intentionally excluded - via introduction of special forwarding state `FWD_STATE_RESCH`: if we already are in rescheduling state and the error is occurred, we just give up with this connection, leave it as is with all its requests, and move on to try next connection. Moreover, we do that on **any** error in `FWD_STATE_RESCH` state, so we do not spend much time with useless connections during busy rescheduling.
Besides, the number of reschedule attempts on EBUSY error is limited by some not big budget (now its value is 10, but it is subject to be discussed, of course), therefore, there cannot be recursive or any other infinite looping.

>In very unlucky situations all connections of the server group can be marked with TFW_CONN_B_BUSY flag and the whole service would be unavailable while all server connections are alive. This situation is very unlikely, but also possible.

Described situation can appear in two cases:
1. All connection of server group are from the same work queue (i.e. CPU) - in this case rescheduling is needed to try send requests through another work queues;
2. All work queues are busy - we have system overflow, and after limited number of reschedule attempts (as mentioned above) we just drop requests. 

>at least one request was sent to the server (msg_sent != 0). Nothing to do is required here, the next attempt to forward unsent requests will happened after a new request is scheduled to the connection or at response receiving.

This case is implemented in current PR.

>there is no requests forwarded (msg_sent == 0). Ideally is to repeat forwarding attempt after some time, if no new requests are assigned to the connection and no forwarding has happened. But this closely connected to tho #687. At this moment we can add a TODO and close server connection to recover. Nothing bad will happen if we close close server connection: no extra re-send will happen.

I agree that the drawback of solution in current PR is some additional complexity, introduced into existent code.
Closing server connection on EBUSY error - is possible workaround for the problem of hanged requests in `fwd_queue`, but this way has its own drawbacks: we need to close valid serviceable connection due to Tempesta internal EBUSY error and to start rather heavy failovering procedure instead of making some attempts to solve situation here and now via rescheduling requests through another work queues; the case with infinite `busy->failover->busy` looping is possible, so we need additional counter in `TfwSrvConn` and some kind of limit (maybe configurable) for this situation, dropping all the requests after the limit will be exhausted; besides, I cannot see in context of #687 the possible solution for the problem with requests hanged in `fwd_queue` - as I understand, that task is more about lock-free (or some kind of that) `fwd_queue`. I think the additional discussion with @krizhanovsky is needed, to make conclusive choice.

>We talked about rescheduling of requests that can't be forwarded, but server connection failovering process are already cares about rescheduling, there is no need to add extra rescheduling routines.
Same should be done for tfw_http_conn_fwd_repair().

Yes, I left busy rescheduling for failovering procedures, but I'm not sure about that, and it is the subject to be discussed. For `tfw_http_conn_shrink_fwdq_resched()` - maybe, rescheduling on EBUSY error is indeed redundant here; but from the other side, we are rescheduling requests initially - due to entire connection fail (connection error or connection closing), not due to EBUSY error, so if we faced EBUSY during this first rescheduling - we can decide to give the requests the "second chance" on the other, less loaded work queue. For `tfw_http_conn_repair()` and `tfw_http_conn_fwd_repair()` - motivation is almost the same: initially, we are re-sending requests due to connection failover, we are not rescheduling them due to EBUSY error, so in this case we also can decide to try on another connection if EBUSY appear.

tempesta_fw/connection.c
```diff
@@ -88,10 +88,15 @@ tfw_connection_release(TfwConn *conn)
...
- * only on an active socket.
+ * Send @msg through connection @conn. Code architecture decisions
+ * ensure that conn->sk remains valid for the life of @conn instance.
+ * The socket itself may have been closed, but not deleted. ss_send()
+ * makes sure that data is sent only on an active socket.
+ * Return value:
+ *   0		- @msg had been sent successfully;
+ *   -EBADF	- connection is broken;
+ *   -EBUSY	- transmission work queue is full;
+ *   -ENOMEM	- out-of-memory error occured.
```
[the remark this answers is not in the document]
aleksostapenko: In case of `-EBUSY` error we have the problem on current work queue and can try with another one. But in case `-ENOMEM` we have the problem with memory lacking, and with high probability we will not succeed on other work queues, so it seems that would be better to drop requests (with error responses or without them) - to avoid additional memory overload and, probably, to allow the system to recover free memory.

tempesta_fw/http.c
```diff
@@ -1089,58 +1108,130 @@ tfw_http_req_evict(TfwSrvConn *srv_conn, TfwServer *srv, TfwHttpReq *req,
 	       || tfw_http_req_evict_retries(srv_conn, srv, req, eq);
 }
 
+static int tfw_http_req_resched(TfwHttpReq *req, TfwServer *srv, FwdState *st);
```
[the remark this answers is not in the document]
aleksostapenko: Yes, but in this place we have cyclic dependency between functions.

tempesta_fw/work_queue.h
```diff
@@ -83,9 +83,17 @@ static inline long
...
-		return ticket;
+	long ticket;
+	int budget = 10;
+
+	/*
+	 * Set small threshold budget for push attempts in order
+	 * to pass through temporary queue overflow.
+	 */
+	while ((ticket = __tfw_wq_push(q, ptr))) {
+		if (--budget <= 0)
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/connection.h
aleksostapenko: Added comment for `tfw_srv_clear_busy_delay()`.

tempesta_fw/connection.h
aleksostapenko: @keshonok Thank you for review!
Due to CPU reordering some requests may be occasionally scheduled on busy connection in the moment of clearing busy bit. To prevent this situation memory barriers are used.
Added comment for `tfw_srv_clear_busy_delay()`.

tempesta_fw/connection.h
aleksostapenko: Yes, you are right. Forgot to remove that redundant check after code changes.
Corrected.

tempesta_fw/connection.h
aleksostapenko: It is expected that most of the time Tempesta would work not in overloaded state, i.e. - without `-EBUSY` appearance and without `TFW_CONN_B_BUSY` bit set for connections. So, with this code most of the time - only `test_bit()` call will be executed, which compiled into light read-only instruction:
```
test  imm8, reg/mem8
```
instead of `clear_bit()` call, which in turn compiled into significantly more heavy, locked read-modify-write instruction:
```
lock andb  imm8, mem8
```

tempesta_fw/http.c
```diff
@@ -1089,58 +1108,130 @@ tfw_http_req_evict(TfwSrvConn *srv_conn, TfwServer *srv, TfwHttpReq *req,
...
+	if (r == -EBADF
+	    || st->state == FWD_STATE_RESCH
+	    || (r == -EBUSY
+		&& srv_conn->msg_sent != NULL
+		&& st->state != FWD_STATE_RESEND))
+		return r;
+
+	if (r == -EBUSY
+	    && !tfw_http_req_fwd_busy_resched(srv_conn, srv, req, st))
+		return r;
```
krizhanovsky: I think this is the main problem of the patch set introducing the complexity and recursion.

I had a quick review of the rescheduling logic which we have before this change. I didn't dig deep, but I imagine following changes to make the code straightforward - maybe you can think in this direction and/or @ikoveshnikov proposal and figure out some more clean solution:

1. `tfw_http_req_fwd()` returns a status code - whether it could send a request or not, so `tfw_http_conn_fwd_unsent()` also returns a status code.
2. `tfw_http_req_resched()` also returns a status code by the result of `tfw_http_req_fwd()` call
3. move the end of `tfw_http_conn_shrink_fwdq_resched()` to a new function `good_func_name1()` which calls the request eviction and rescheduling in a loop, so we limit number of retries/reschedules.
4.  `tfw_http_req_cache_cb()` calls ``good_func_name2()` instead of `tfw_http_req_fwd()`. ``good_func_name2()` calls `tfw_http_req_fwd()` and `good_func_name1()` if the former fails.
5. `tfw_http_hm_srv_send()` just do the same as now.

aleksostapenko: Solution had been refactored to avoid recursion and states on stack.

tempesta_fw/work_queue.h
```diff
@@ -83,9 +83,17 @@ static inline long
...
+			return ticket;
```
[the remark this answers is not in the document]
aleksostapenko: Yeah, missed that. Moved budget check to `__tfw_wq_push()`.

tempesta_fw/http.c
```diff
@@ -1165,14 +1257,48 @@ tfw_http_conn_fwd_unsent(TfwSrvConn *srv_conn, struct list_head *eq)
...
+		 * while sending requests.
+		 */
+		if (ret == -EBUSY)
+			busy = true;
+		/*
+		 * If connection is broken or work queue is busy and connection
+		 * has request(s) forwarded but unanswered, we can leave all
+		 * requests in @fwd_queue: another attempt to send them will
+		 * be made when connection will be repaired or when response(s)
+		 * for unanswered request(s) will arrived from backend.
```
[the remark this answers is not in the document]

PR #1132, opened 2018-12-13
Fix various misspells

ikoveshnikov:
I've fixed some misspells in changed files in PR #1067. But that PR is too big and it's bad idea to mix everything up, so i've gone through other files, fixed errors there and created this PR.

### DIFF
--- a/tempesta_fw/cache.c
+++ b/tempesta_fw/cache.c
@@ -68,7 +68,7 @@ static const TfwStr tfw_cache_raw_headers_304[] = {
 /*
  * @trec	- Database record descriptor;
  * @key_len	- length of key (URI + Host header);
- * @status_len	- length of response satus line;
+ * @status_len	- length of response status line;
  * @hdr_num	- number of headers;
  * @hdr_len	- length of whole headers data;
  * @hdr_len_304	- length of headers data used to build 304 response;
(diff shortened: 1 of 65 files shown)


tempesta_fw/cache.c
```diff
@@ -345,7 +345,7 @@ tfw_cache_employ_req(TfwHttpReq *req)
 		req->cache_ctl.flags |= TFW_HTTP_CC_CFG_CACHE_BYPASS;
 		return false;
 	}
-	/* cache_fulfill - work as usual in cache mode. */
+	/* cache_fulfil - work as usual in cache mode. */
```
i-rinat: It's spelled `fulfill` everywhere else in the code, including constant names and even configuration parameter: `cache_fulfill`. As far as I understand, both `fulfil` and `fulfill` are accepted. We just need to use either one consistently.

tempesta_fw/t/unit/helpers.c
```diff
@@ -1,7 +1,7 @@
 /**
  *		Tempesta FW
  *
- * This file contains utils that help to test certain Tempesta FW modules.
+ * This file contains utills that help to test certain Tempesta FW modules.
```
i-rinat: doesn't look right

tempesta_fw/cache.c
ikoveshnikov: Yes, American/British English difference. This is the only mistakenly "fixed" spelling, reverted it. Thank you!

tempesta_fw/t/unit/helpers.c
ikoveshnikov: Fixed

PR #1133, opened 2018-12-13
Correctly delist request on response parsing errors

ikoveshnikov:
When a server sends invalid response, Tempesta sends an error
message to the client and destroys the request. But if the request
was the only request in server connection `fwd_queue`,  `msg_sent`
is not updated and contains invalid pointer. Later during server
connection repair this invalid pointer is used and a crash may happen.

Steps to reproduce the issue:
1. Configure exactly one server connection: `server 127.0.0.1:8080`
2. Send a request
3. Force server to send invalid request
4. Send another request, most likely that crash will happen

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -815,7 +815,7 @@ __tfw_http_conn_msg_sent_prev(TfwSrvConn *srv_conn)
  * queue state consistent.
  */
 static inline void
-__http_req_delist(TfwSrvConn *srv_conn, TfwHttpReq *req)
+tfw_http_req_delist(TfwSrvConn *srv_conn, TfwHttpReq *req)
 {
 	tfw_http_req_nip_delist(srv_conn, req);
 	list_del_init(&req->fwd_list);
@@ -843,7 +843,7 @@ tfw_http_req_err(TfwSrvConn *srv_conn, TfwHttpReq *req,
 		 const char *reason)
 {
 	if (srv_conn)
-		__http_req_delist(srv_conn, req);
+		tfw_http_req_delist(srv_conn, req);
 	__tfw_http_req_err(req, eq, status, reason);
 }
 
@@ -1018,7 +1018,7 @@ tfw_http_req_evict_dropped(TfwSrvConn *srv_conn, TfwHttpReq *req)
 		TFW_DBG2("%s: Eviction: req=[%p] client disconnected\n",
 			 __func__, req);
 		if (srv_conn)
-			__http_req_delist(srv_conn, req);
+			tfw_http_req_delist(srv_conn, req);
 		tfw_http_conn_msg_free((TfwHttpMsg *)req);
 		TFW_INC_STAT_BH(clnt.msgs_otherr);
 		return true;
@@ -1105,7 +1105,7 @@ tfw_http_req_fwd_send(TfwSrvConn *srv_conn, TfwServer *srv,
 		TFW_DBG2("%s: Forwarding error: conn=[%p] req=[%p]\n",
 			 __func__, srv_conn, req);
 		if (req->flags & TFW_HTTP_F_HMONITOR) {
-			__http_req_delist(srv_conn, req);
+			tfw_http_req_delist(srv_conn, req);
 			WARN_ON_ONCE(req->pair);
 			tfw_http_msg_free((TfwHttpMsg *)req);
 			TFW_WARN_ADDR("Unable to send health"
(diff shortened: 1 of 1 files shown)


tempesta_fw/http.c
```diff
@@ -3329,7 +3334,7 @@ tfw_http_resp_process(TfwConn *conn, const TfwFsmData *data)
 	 * we won't resend it.
 	 */
 	bad_req = hmresp->req;
-	tfw_http_req_delist((TfwSrvConn *)conn, bad_req);
+	tfw_http_req_delist(hmresp);
```
keshonok: I think the idea was that both `req` and `srv_conn` were known already so there's no need to get/derive them again. The called function is inlined and so the values are just used in place. Perhaps, GCC optimizes this just as needed, and there's no "calculation" of these values again.

keshonok: Also, is it possible that this situation can happen in other places as well? This particular function is only called once. The unlocked version is called in all other cases.

tempesta_fw/http.c
```diff
@@ -821,9 +821,14 @@ __http_req_delist(TfwSrvConn *srv_conn, TfwHttpReq *req)
...
 static inline void
-tfw_http_req_delist(TfwSrvConn *srv_conn, TfwHttpReq *req)
+tfw_http_req_delist(TfwHttpMsg *hmresp)
 {
+	TfwHttpReq *req = hmresp->req;
+	TfwSrvConn *srv_conn = (TfwSrvConn *)hmresp->conn;
+
 	spin_lock(&srv_conn->fwd_qlock);
+	if ((TfwMsg *)req == srv_conn->msg_sent)
+		srv_conn->msg_sent = NULL;
```
keshonok: Incidentally, this is exactly what's done in `tfw_http_popreq()`. If this change is here to stay, then this sequence in that function can be replaces with a call to `tfw_http_req_delist()`.

I think the question I have is whether this check should be here or in `_tfw_http_req_delist()`. If it's the latter, then are there cases where this check is unnecessary?

tempesta_fw/http.c
ikoveshnikov: @keshonok Thank you for review!

In "hot path" functions we normally try to reduce number of calculations, but this sometimes makes function readability worse. Take a look at this mostly obsolete (and resolved) example. Function takes two parameters: `int f(TfwHttpReq *req, TfwHttpResp *resp)`. But in the same time we know, that we can get `req` as `resp->req` or `resp` as `req->resp`. So why the two different pointers are passed here? Is it the same `req` as saved in `resp->req` or not? 

It isn't a hot path function, it's very and very unlikely situation instead. We think about of protected servers as reliable and good servers, so we actually expect server to send a valid message, not a garbage that can't be parsed or processed. So extra calculations (if happen) doesn't hurt much.

In the same time, the function is self-sufficient, it takes care of server connection forwarding queue, and keeps it to be consistent (see https://github.com/tempesta-tech/tempesta/pull/1133#discussion_r241723435 ). 

That is why I thought that extra calculations doesn't hurt and being more explicit helps to make understanding of the function role (readability) better. But taking into account review comments, it seems to be partly done, I should rename the function and add a good comment description to mirror function role.

ikoveshnikov:
Added function description, to make it more explicit and easy to understand.

tempesta_fw/http.c
```diff
@@ -2998,6 +3006,11 @@ tfw_http_popreq(TfwHttpMsg *hmresp)
 		srv_conn->msg_sent = NULL;
 	__http_req_delist(srv_conn, req);
 	tfw_http_conn_nip_adjust(srv_conn);
+
+	if (unlikely(!fwd_unsent)) {
+		spin_unlock(&srv_conn->fwd_qlock);
+		return;
+	}
```
krizhanovsky: `fwd_unsent` is used to execute the small code above only and exit the function. Moreover, the code above is `http_req_delist()` with zeroing and `tfw_http_conn_nip_adjust()` call. I propose to leave `http_req_delist()` as
```C
http_req_delist()
{
    if ((TfwMsg *)req == srv_conn->msg_sent)
        srv_conn->msg_sent = NULL;
    __http_req_delist(srv_conn, req);
    tfw_http_conn_nip_adjust(srv_conn);
```
call it here and define
```C
http_req_delist_locked()
{
    spin_lock(&srv_conn->fwd_qlock);
    http_req_delist();
    spin_unlock(&srv_conn->fwd_qlock);
}
```
and call it instead of `tfw_http_popreq( .., false)`

PR #1134, opened 2018-12-13
split skb as soon http_parser returned TFW_PASS

ikoveshnikov:
If ss_skb_split() fails, a part of the next http message will be added to the end of last parsed message. Forwarding such message will break response-request sequence. There is no sense to process the parsed message if split will fail: such message can't be forwarded and connection must be closed to recover.

Current master is also vulnerable for attack on response-request sequence attack:
- cilent sends a pack of pipilined requests, but the first request in the pack has a `Connection: close` header
- on current master Tempesta creates a subling message AND splits the skb only if `Connection: close` header is not set. So no split is performed.


tempesta_fw/http.c
```diff
@@ -2844,24 +2847,24 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
...
-			 * Pipelined requests: create a new sibling message.
-			 * @skb is replaced with pointer to a new SKB.
-			 */
+		/*
+		 * Pipelined requests: create a new sibling message.
+		 * If pipelined message can't be created, it still possible to
+		 * process current one. But @skb must be freed then, since it's
+		 * not owned by any message.
+		 */
+		if (!req_conn_close && skb) {
```
krizhanovsky: If `req_conn_close` and `skb != NULL` (i.e. there are pipelined messages), then the skb leaks - nobody frees it.

PR #1138, opened 2018-12-17
Move http parser to connection

ikoveshnikov:
There is no need to allocate a new instance of http parser for each incoming message. It can be allocated once per connection.

### DIFF
--- a/tempesta_fw/http_parser.h
+++ b/tempesta_fw/http_parser.h
@@ -1,0 +1,29 @@
+/**
+ *		Tempesta FW
+ *
+ * Copyright (C) 2018 Tempesta Technologies, Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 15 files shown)


tempesta_fw/http_parser.h
```diff
@@ -0,0 +1,110 @@
...
+ * Non-cacheable hop-by-hop headers in terms of RFC 7230.
+ *
+ * We don't store the headers in cache and create them from scratch if needed.
+ * Adding a header is faster then modify it, so this speeds up headers
+ * adjusting as well as saves cache storage.
+ *
+ * Headers unconditionally treated as hop-by-hop must be listed in
+ * tfw_http_init_parser_req()/tfw_http_init_parser_resp() functions and must be
+ * members of Special headers.
+ * group.
```
aleksostapenko: `group` looks like redundant word here.

PR #1140, opened 2018-12-21
Use standart bitops functions for message flags.

ikoveshnikov:
There was a discussion that a HTTP message flags defined as enum and a set of macroses doesn't look good, in the same time usage of standard bit operations looks cleaner and provide atomic possibilities.

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -1961,8 +1961,8 @@ tfw_http_set_hdr_date(TfwHttpMsg *hm)
 static int
 tfw_http_set_hdr_connection(TfwHttpMsg *hm, unsigned long conn_flg)
 {
-	BUILD_BUG_ON(BIT_WORD(__TFW_HTTP_MSG_M_CONN_MASK) != 0);
-	if (((hm->flags[0] & __TFW_HTTP_MSG_M_CONN_MASK) == conn_flg)
+	BUILD_BUG_ON(BIT_WORD(__TFW_HTTP_MSG_M_CONN) != 0);
+	if (((hm->flags[0] & __TFW_HTTP_MSG_M_CONN) == conn_flg)
 	    && (!TFW_STR_EMPTY(&hm->h_tbl->tbl[TFW_HTTP_HDR_CONNECTION]))
 	    && !test_bit(TFW_HTTP_B_CONN_EXTRA, hm->flags))
 		return 0;
@@ -1988,8 +1988,8 @@ tfw_http_set_hdr_keep_alive(TfwHttpMsg *hm, unsigned long conn_flg)
 {
 	int r;
 
-	BUILD_BUG_ON(BIT_WORD(__TFW_HTTP_MSG_M_CONN_MASK) != 0);
-	if ((hm->flags[0] & __TFW_HTTP_MSG_M_CONN_MASK) == conn_flg)
+	BUILD_BUG_ON(BIT_WORD(__TFW_HTTP_MSG_M_CONN) != 0);
+	if ((hm->flags[0] & __TFW_HTTP_MSG_M_CONN) == conn_flg)
 		return 0;
 
 	switch (conn_flg) {
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -358,7 +358,7 @@ tfw_http_prep_redirect(TfwHttpMsg *resp, unsigned short status, TfwStr *rmark,
 {
 	TfwHttpReq *req = resp->req;
 	size_t data_len;
-	int conn_flag = req->flags & __TFW_HTTP_MSG_M_CONN_MASK, ret = 0;
+	int ret = 0;
 	TfwMsgIter it;
 	static TfwStr rh_302 = {
 		.ptr = S_REDIR_302, .len = SLEN(S_REDIR_302) };
@@ -420,9 +420,9 @@ tfw_http_prep_redirect(TfwHttpMsg *resp, unsigned short status, TfwStr *rmark,
 					 TFW_HTTP_HDR_HOST, &host);
 
 	/* Set "Connection:" header field if needed. */
-	if (conn_flag == TFW_HTTP_F_CONN_CLOSE)
+	if (test_bit(TFW_HTTP_B_CONN_CLOSE, req->flags))
 		cookie_crlf = &crlf_close;
-	else if (conn_flag == TFW_HTTP_F_CONN_KA)
+	else if (test_bit(TFW_HTTP_B_CONN_KA, req->flags))
 		cookie_crlf = &crlf_keep;
 
 	/* Add variable part of data length to get the total */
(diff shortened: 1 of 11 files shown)


tempesta_fw/http.h
```diff
@@ -285,73 +285,53 @@ enum {
...
-#define TFW_HTTP_F_WHITELIST		(1U << TFW_HTTP_B_WHITELIST)
-#define TFW_HTTP_F_REQ_DROP		(1U << TFW_HTTP_B_REQ_DROP)
-
-#define TFW_HTTP_F_VOID_BODY		(1U << TFW_HTTP_B_VOID_BODY)
-#define TFW_HTTP_F_HDR_DATE		(1U << TFW_HTTP_B_HDR_DATE)
-#define TFW_HTTP_F_HDR_LMODIFIED	(1U << TFW_HTTP_B_HDR_LMODIFIED)
-#define TFW_HTTP_F_RESP_STALE		(1U << TFW_HTTP_B_RESP_STALE)
-#define TFW_HTTP_F_RESP_READY		(1U << TFW_HTTP_B_RESP_READY)
+#define __TFW_HTTP_MSG_M_CONN_MASK					\
+	(BIT(TFW_HTTP_CONN_CLOSE) | BIT(TFW_HTTP_CONN_KA))
```
keshonok: On one hand, these flags are mutually exclusive and they are handled and used in the code as mutually exclusive flags, i.e. if one is set, then do this, else do that.
On the other hand, there's an implicit understanding in the code that the `TFW_HTTP_CONN_CLOSE` flag takes precedence over the `TFW_HTTP_CONN_KA` flags.

I think a thorough comment is need to clearly state the rules of engagement with these two flags. Perhaps, in the enum where these flags are defined. Do they have to be maintained as mutually exclusive? Is it possible that both are set after a message is parsed? Does one takes precedence over the other?

tempesta_fw/http.c
```diff
@@ -1997,22 +2007,24 @@ tfw_http_set_hdr_connection(TfwHttpMsg *hm, int conn_flg)
...
  */
 static int
-tfw_http_set_hdr_keep_alive(TfwHttpMsg *hm, int conn_flg)
+tfw_http_set_hdr_keep_alive(TfwHttpMsg *hm, unsigned long conn_flg)
 {
 	int r;
 
-	if ((hm->flags & __TFW_HTTP_MSG_M_CONN_MASK) == conn_flg)
+	BUILD_BUG_ON(__TFW_HTTP_MSG_M_CONN_MASK > BITS_PER_LONG);
+	if ((hm->flags[0] & __TFW_HTTP_MSG_M_CONN_MASK) == conn_flg)
```
keshonok: This looks like a rather hacky way of doing this check. This code implicitly assumes that the connection's keep-alive/close flags are located somewhere in the first bits of the `hm->flags` bitmap (first 64 bits, in fact). That may be true, but it creates an unnecessary dependency on this implicit piece of knowledge. Again, there's the issue of whether these two flags are maintained as mutually exclusive or not, or whether these two flags  can be both set at the same time by the parser.

It appears that, according to changes in other parts of this patch, `conn_flag` is now guaranteed to only have one bit set, either `TFW_HTTP_CONN_KA` or `TFW_HTTP_CONN_CLOSE`. Then, perhaps, only a check for a specific flag is needed, like `test_bit(conn_flag, hm->flags)`.

tempesta_fw/http.h
```diff
@@ -285,73 +285,53 @@ enum {
...
-#define TFW_HTTP_F_HMONITOR		(1U << TFW_HTTP_B_HMONITOR)
-#define TFW_HTTP_F_WHITELIST		(1U << TFW_HTTP_B_WHITELIST)
-#define TFW_HTTP_F_REQ_DROP		(1U << TFW_HTTP_B_REQ_DROP)
-
-#define TFW_HTTP_F_VOID_BODY		(1U << TFW_HTTP_B_VOID_BODY)
-#define TFW_HTTP_F_HDR_DATE		(1U << TFW_HTTP_B_HDR_DATE)
-#define TFW_HTTP_F_HDR_LMODIFIED	(1U << TFW_HTTP_B_HDR_LMODIFIED)
-#define TFW_HTTP_F_RESP_STALE		(1U << TFW_HTTP_B_RESP_STALE)
-#define TFW_HTTP_F_RESP_READY		(1U << TFW_HTTP_B_RESP_READY)
+#define __TFW_HTTP_MSG_M_CONN_MASK					\
```
keshonok: The original idea was to have a special mark for these definitions. Flags would be `xxx_F_xxx`, bit numbers would be `xxx_B_xxx`, and masks would be `xxx_M_xxx`. Then, this definition should have been called `__TFW_HTTP_MSG_M_CONN` or maybe `__TFW_HTTP_MSG_M_CONN_FLAGS`. Alternatively, just lose this `_M_` and just have `__TFW_HTTP_MSG_CONN_MASK`.

tempesta_fw/http_parser.c
```diff
@@ -1004,8 +1004,8 @@ __FSM_STATE(RGen_HdrOtherV) {						\
 __FSM_STATE(RGen_BodyInit) {						\
 	TfwStr *tbl = msg->h_tbl->tbl;					\
 									\
-	TFW_DBG3("parse request body: flags=%#x content_length=%lu\n",	\
-		 msg->flags, msg->content_length);			\
+	TFW_DBG3("parse request body: flags=%#lx content_length=%lu\n",	\
+		 msg->flags[0], msg->content_length);			\
```
keshonok: Yeah, I think this is hacky, but perhaps justified in debug printouts.

tempesta_fw/http.h
ikoveshnikov: @keshonok many thanks for review!

Hm, yes, comment for `CONN_CLOSE` and `CONN_KA` is required. This flags actually has two different meanings in the same time and it always was in that way. 

`CONN_CLOSE` flag has two meanings:
- `Connection:` header contains `close` term;
- Connection is to be closed after this message. This flag is also set when there is no `Connection: close` term, i.e. for HTTP/1.0 connections,  or when error/attack has spotted and client connection is to be closed with a proper error response.

`CONN_KA` flag has only one meaning:
-   `Connection:` header contains `keep-alive` term.
Unlike to `CONN_CLOSE` there is no 'connection is in keep-alive mode'. It's implicitly so for HTTP/1.1 connections, so no special flag is required.

This two meanings are used for two aims:
- to init `SS_FLAG` when message is forwarded: to tell lower kernel levels to close the connection after transmission. We just need to know whether the `CONN_CLOSE` flag is set or not.
- to adjust `Connection:` header before forwarding. We would like to know about current content of the `Connection:` header, and these flags are used to take decision whether rewriting of the header is required or not.

Hard to say should we require mutual exclusivity for the flags or not. On the one hand there only two cases: close connection or keep it. No other variants. On the other hand, if the message originally had `Connection: keep-alive` header, but Tempesta decided to close the connection after it,  `keep-alive` token is still here and we have to modify the connection header. E.g.:
- just parsed request has `Connection: keep-alive` header;
- error happened while allocating sibling request message;
- we set `Conn_Stop` flag to the connection to stop further parsing operations and keep connection open to process all pending requests; 
- set `CONN_CLOSE` flag to the current request to close the connection once corresponding response is forwarded.

So I've updated the code in such a way, that mutual exclusivity is not required now. Precedence of `CONN_CLOSE` over `CONN_KA` flag is less strict but still effective condition.

tempesta_fw/http.c
keshonok: Uh, well, this `BUILD_BUG_ON()` states that all flags in the mask must be within the first 5 bits, assuming that `BITS_PER_LONG` is 64. Or, if there's just one flag and it's the 6-th bit. Was that the intention? It's non-obvious I would say.

Now that `conn_flag` is set to one specific bit/flag, Perhaps it's cleaner and clearer to just check for that bit/flag.

Issue #1043, opened 2018-07-24
Fast sticky session scheduler is broken


ikoveshnikov:
As sticky session scheduler was introduced (#593) it was claimed to be the fastest scheduler. When a new request arrived destination server was chosen by HTTP scheduler call and subsequent group scheduler call. HTTP scheduler used a lot of string compare functions and can dramatically slow down system in environments with super long HTTP rules lists. Sticky sessions scheduler was intended to replace that expensive string operations with cheap sticky cookie calculations. 

But recent #1017 patch broke this convention. VHost is assigned to a request, before sticky cookie is analysed. 'VHost assignment' operation is very close to old `http scheduler` operation: multiple string comparison operations may be involved. This means that Fast sticky sessions scheduler doesn't work any more.

'VHost assignment' operation is very important, since many configuration options have or will have vhost scope. To return Fast sticky scheduler back operational, target vhost must be saved into sticky cookie structure (`TfwStickyConn`).

Issue #899, opened 2018-02-01
[Cache] Unathorized clients should be blocked before accessing cache


ikoveshnikov:
When a new request is received it's forwarded to cache first:
https://github.com/tempesta-tech/tempesta/blob/195d85eaa609cc0c1a8b12f404221f7f6773f67d/tempesta_fw/http.c#L2373
and only after that sticky cookie is checked:
https://github.com/tempesta-tech/tempesta/blob/195d85eaa609cc0c1a8b12f404221f7f6773f67d/tempesta_fw/http.c#L1966-L1970 

That leads to unnecessary cache lookup for requests with violated or expired sticky cookie. Early blocking of unauthorized clients will help us to save resources under DDOS attacks.

PR #1141, opened 2018-12-21
Fixes on serving requests with errors

ikoveshnikov:
Requires #1140 and built upon it. Required for #1043 

Changes:
- If error has happened during request processing, don't close client connection immediately, send all pending requests first. Don't silently drop request to avoid breaking of the request-response sequence of client. #962 
- Replace `TFW_HTTP_SUSPECTED` flag with `TFW_HTTP_CONN_CLOSE`. Later is more generic and `Connection: ` is not set correctly with `TFW_HTTP_SUSPECTED` flag
- Check sticky session cookie before forward request to cache. Don't allow unauthorised client to stress cache and receive protected information. #899

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -1954,13 +1954,14 @@ tfw_http_set_hdr_date(TfwHttpMsg *hm)
 }
 
 /**
- * The client connection is to be closed after response to current request is
- * set. No need to drop the CONN_KA flag if set: see flags description.
- *
+ * Connection is to be closed after response for the request @req is forwarded
+ * to the client. Don't process new requests from the client and update
+ * message flags for proper "Connection: " header value.
  */
 static inline void
 tfw_http_req_set_conn_close(TfwHttpReq *req)
 {
+	TFW_CONN_TYPE(req->conn) |= Conn_Stop;
 	set_bit(TFW_HTTP_B_CONN_CLOSE, req->flags);
 }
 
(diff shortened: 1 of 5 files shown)


ikoveshnikov:
Tests must be patched to pass the tests. Build `361` on CI passes for this branch and tempesta-tech/tempesta-test#70

tempesta_fw/http.c
```diff
@@ -2794,25 +2838,94 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
...
+		 *
+		 * In the same time location may differ between requests, so the
+		 * sticky module can't fill it.
+		 */
+		if (!req->vhost)
+			req->vhost = tfw_http_tbl_vhost((TfwMsg *)req, &block);
+		if (req->vhost)
 			req->location = tfw_location_match(req->vhost,
 							   &req->uri_path);
+		else if (block) {
```
krizhanovsky: If HTTPtables didn't return a vhost, why they can't block the request, i.e. why the `else` appeared?

It'd be better to use curly braces for the `if` hading `else if` with curly braces.

tempesta_fw/http.c
```diff
@@ -2794,25 +2838,94 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
...
+			TFW_INC_STAT_BH(clnt.msgs_filtout);
+			tfw_http_req_parse_block(
+				req, 403,
```
krizhanovsky: You can leave `req, 403` on the same line with `tfw_http_req_parse_block(`. The same for the similar call on the above and below - there is no need to introduce additional line breaks

tempesta_fw/http.c
```diff
@@ -2374,85 +2403,135 @@ tfw_http_conn_error_log(TfwConn *conn, const char *msg)
...
-		tfw_connection_unlink_msg(req->conn);
-		spin_lock(&cli_conn->seq_qlock);
-		list_add_tail(&req->msg.seq_list, &cli_conn->seq_queue);
-		spin_unlock(&cli_conn->seq_qlock);
-		tfw_http_req_mark_error(req, status);
+	/*
+	 * Error was happened and request should be dropped or blocked,
+	 * but other modules (e.g. sticky cookie module) may have a response
+	 * prepared for this request. A new error response is to be generated
+	 * for the request, drop any previous request paired with the request.
```
krizhanovsky: -> "...any previous **response** paired with..."

tempesta_fw/http.c
```diff
@@ -2374,85 +2403,135 @@ tfw_http_conn_error_log(TfwConn *conn, const char *msg)
...
+			list_add_tail(&req->msg.seq_list, &cli_conn->seq_queue);
+			spin_unlock(&cli_conn->seq_qlock);
+		}
+		else {
+			WARN_ONCE(list_empty_careful(&req->msg.seq_list),
+				  "Request is not in in seq_queue\n");
+		}
+		/*
+		 * If !on_req_recv_event, then the request @req - some
+		 * random request from the seq_queue, not the last one.
```
krizhanovsky: I don't fully understand the statement: if we have only one request from a client, then the request will the only one and the last one. Why do we care here whether the request is last one or there are some other requests from the client?

tempesta_fw/http.c
```diff
@@ -2813,33 +2854,92 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
...
-					   " out");
+		tfw_http_req_parse_block(req, 503,
+			"request dropped: can't send JS challenge");
+		return TFW_BLOCK;
+
+	default:
+		TFW_INC_STAT_BH(clnt.msgs_otherr);
+		tfw_http_req_parse_block(req, 500,
+			"request dropped: internal error in Sticky "
+			"module");
```
krizhanovsky: No need for the line break, you can place "module" on previous line

PR #1143, opened 2019-01-01
Fix review comments for #1037

krizhanovsky:
Fix usage of uninitialized valiables, wrong usage of unsigned variables, some code mess.

@ikoveshnikov the build is failed in CI because of old kernel. Could you please deploy the new kernel version?


ikoveshnikov:
> the build is failed in CI because of old kernel. Could you please deploy the new kernel version?
Yep, just remembered about that. I'll update it today.

Issue #865, opened 2017-11-23
Unit tests for TfwStr


krizhanovsky:
Currently not all `str.[ch]` functions are covered by unit `t/unit/test_tfw_str.c`, e.g. `tfw_str_del_chunk()` and `tfw_str_add_compound()`. Need to fix it and write tests for the uncovered functions.

krizhanovsky:
Aggregate with #215:

TfwStr uses casts of `ptr` field to `char *` and `TfwStr []`, replacing it by

```
    union {
            char *data;
            TfwStr *chunks;
    }
```

will simplify the code and eliminate many casts.

Also it's good to replace current TfwStr->flags by bit field with 3 bytes for number of chunks and 1 byte for actual flags.

PR #1150, opened 2019-01-15
Replace void ptr by union

avbelov23:
Part 1 by #865 

Replace casts of ptr field to `char *` and `TfwStr []` by union

### DIFF
--- a/tempesta_fw/str.h
+++ b/tempesta_fw/str.h
@@ -209,20 +209,23 @@ size_t tfw_ultoa(unsigned long ai, char *buf, unsigned int len);
  * @flags	- 3 most significant bytes for number of chunks of compound
  * 		  string and the least significant byte for flags;
  */
-typedef struct {
-	void		*ptr;
+typedef struct tfwstr_t {
+	union {
+		char *data;
+		struct tfwstr_t *chunks;
+	};
 	struct sk_buff	*skb;
 	unsigned long	len;
 	unsigned char	eolen;
 	unsigned int	flags;
 } TfwStr;
 
-#define DEFINE_TFW_STR(name, val) TfwStr name = { (val), NULL,		\
+#define DEFINE_TFW_STR(name, val) TfwStr name = { .data = (val), NULL, \
 						  sizeof(val) - 1, 0 }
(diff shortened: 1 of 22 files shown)


tempesta_fw/str.h
```diff
@@ -209,20 +209,23 @@ size_t tfw_ultoa(unsigned long ai, char *buf, unsigned int len);
  * @flags	- 3 most significant bytes for number of chunks of compound
  * 		  string and the least significant byte for flags;
  */
-typedef struct {
-	void		*ptr;
+typedef struct tfwstr_t {
+	union {
+		char *data;
+		struct tfwstr_t *chunks;
+	} ptr;
```
ikoveshnikov: Totally agree with @i-rinat , need to make the union anonymous to address `data` and `chunks` directly.

avbelov23:
DONE

avbelov23:
DONE

avbelov23:
DONE

tempesta_fw/str.h
```diff
@@ -209,20 +209,24 @@ size_t tfw_ultoa(unsigned long ai, char *buf, unsigned int len);
  * @flags	- 3 most significant bytes for number of chunks of compound
  * 		  string and the least significant byte for flags;
  */
-typedef struct {
-	void		*ptr;
+typedef struct tfwstr_t {
+	union {
+		char *data;
+		struct tfwstr_t *chunks;
+		size_t num;
```
ikoveshnikov: Er, the field `num` is introduced to eliminate extra type conversions in recent changes by @i-rinat . But storing unsigned integer in `TfwStr->ptr` is a hack, the string is not supposed to be used until the pointer is set to some valid value. The hack is reasonable there and it's a kind of well-known hack. But I don't want to legalise it on architecture level: allowing arbitrary integer numbers can't be distinguished from the pointers, so any time we want to use the pointer, we need to check some flag in the `TfwStr` to be sure that the pointer is valid.  It's to huge penalty to support one single well known hack.

i-rinat: `num` field was removed.

PR #1162, opened 2019-01-28
fix #1147 : init frag variable before use.

### LINKED ISSUE #1147: Can't build 304 response error
The issue happened during CI tests on real server, but I'm able to reproduce it on my VM. Half of the tests for 'If-Modified-Since'  logic fails. The system log is populated  by the ` Warning: Can't build 304 response, key=50f0`. Seems like one of the `tfw_msg_write()` operations fails.

Steps to reproduce: run `cache.test_conditional.TestConditional.test_mod_since` test.

ikoveshnikov:
Original PR is #1149

### DIFF
--- a/tempesta_fw/msg.c
+++ b/tempesta_fw/msg.c
@@ -72,8 +72,9 @@ this_chunk:
 			skb_frag_t *frag = ss_skb_frag_next(&it->skb,
 							    it->skb_head,
 							    &it->frag);
-			if (unlikely(!frag)
-			    && ((c_size != f_room) || (c + 1 < end)))
+			if (WARN_ON_ONCE(!frag
+					 && ((c_size != f_room)
+					     || (c + 1 < end))))
 			{
 				return -E2BIG;
 			}
--- a/tempesta_fw/msg.c
+++ b/tempesta_fw/msg.c
@@ -36,7 +36,7 @@ tfw_msg_write(TfwMsgIter *it, const TfwStr *data)
 {
 	char *p;
 	const TfwStr *c, *end;
-	skb_frag_t *frag;
+	skb_frag_t *frag = &skb_shinfo(it->skb)->frags[it->frag];
 	unsigned int c_off = 0, f_size, c_size, f_room, n_copy;
 
 	BUG_ON(TFW_STR_DUP(data));
--- a/tempesta_fw/msg.c
+++ b/tempesta_fw/msg.c
@@ -34,20 +34,22 @@
 int
 tfw_msg_write(TfwMsgIter *it, const TfwStr *data)
 {
-	char *p;
 	const TfwStr *c, *end;
-	skb_frag_t *frag = &skb_shinfo(it->skb)->frags[it->frag];
-	unsigned int c_off = 0, f_size, c_size, f_room, n_copy;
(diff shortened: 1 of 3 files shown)


tempesta_fw/msg.c
```diff
@@ -34,20 +34,22 @@
...
 tfw_msg_write(TfwMsgIter *it, const TfwStr *data)
 {
-	char *p;
 	const TfwStr *c, *end;
-	skb_frag_t *frag;
-	unsigned int c_off = 0, f_size, c_size, f_room, n_copy;
 
 	BUG_ON(TFW_STR_DUP(data));
+	if (WARN_ON_ONCE(it->frag >= skb_shinfo(it->skb)->nr_frags))
+		return -E2BIG;
```
aleksostapenko: In case of `*skb == skb_head` in `ss_skb_frag_next()` we get here `skb_head == it->skb` and `0 == it->frag`; so, condition `it->frag >= skb_shinfo(it->skb)->nr_frags)` probably will be false here and we continue execute all the following `tfw_msg_write()` calls without warnings.

ikoveshnikov: The scenario you've described  may happen on subsequent `tfw_msg_write()` calls, which is very common case. So yes, the condition will false and we will continue to execute the function. But `*skb == skb_head` really means that we run out of the spare place in the skb fragments, and no more data can be pushed there. So if we continue to execute the function, we will walk through all the fragments once again and return `-E2BIG` error.

So there is no errors and the function can be called multiple times without any worries. The `tfw_msg_write()` function is used only for preallocated skbs, and data size is known before skb allocations, thus `data` should perfectly fit the `it`. 

The reason for `WARN_ON` is invalid `it` state, where is `it->frag` points to nonexistent fragment.

tempesta_fw/msg.c
```diff
@@ -34,20 +34,22 @@
...
 this_chunk:
-		if (!frag)
-			return -E2BIG;
 
 		c_size = c->len - c_off;
 		if (it->frag >= 0) {
-			frag = &skb_shinfo(it->skb)->frags[it->frag];
+			unsigned int f_size;
+			skb_frag_t *frag = &skb_shinfo(it->skb)->frags[it->frag];
+
```
aleksostapenko: It seems that there is a problem with using skb header when allocating more than one skb - in `tfw_http_msg_setup()` each skb is allocated with a place under the header:
https://github.com/tempesta-tech/tempesta/blob/5631d72b222aa62fdd74bc5fe56650d687a04042/tempesta_fw/ss_skb.h#L174
but in `tfw_msg_write()` only the first skb header in the list is used (`it->frag < 0`):
https://github.com/tempesta-tech/tempesta/blob/5631d72b222aa62fdd74bc5fe56650d687a04042/tempesta_fw/msg.c#L59
for subsequent skb `it-> frag` is always initialized by zero:
https://github.com/tempesta-tech/tempesta/blob/5631d72b222aa62fdd74bc5fe56650d687a04042/tempesta_fw/ss_skb.h#L132
so then we always fall into the `it->frag >= 0` branch.
Look like the current master version of `tfw_msg_write()` has the same issue too.

PR #1163, opened 2019-01-28
Fix the skb fragments extending.

aleksostapenko:
Patch fixes two problems:
1. In `__extend_pgfrags()` - in case of `n == 2 && si->nr_frags == MAX_SKB_FRAGS && from == MAX_SKB_FRAGS - 1`: a) resolve issue with incorrect fragments shift; b) avoid possible BUG_ON if `n_shift < 0`;
2. In `__split_pgfrag_add()` - in case of `i == MAX_SKB_FRAGS - 1`: resolve issue with incorrect determination of destination `frag` and, as a result, wrong `it->data` and `it->skb` output.

### DIFF
--- a/tempesta_fw/ss_skb.c
+++ b/tempesta_fw/ss_skb.c
@@ -283,11 +283,11 @@ __skb_insert_after(struct sk_buff *skb, struct sk_buff *nskb)
 static int
 __extend_pgfrags(struct sk_buff *skb_head, struct sk_buff *skb, int from, int n)
 {
-	int i, n_shift, n_excess = 0;
 	struct skb_shared_info *si = skb_shinfo(skb);
+	int i, n_shift, n_excess = 0, tail_frags = si->nr_frags - from;
 
 	BUG_ON((n <= 0) || (n > 2));
-	BUG_ON(from > si->nr_frags);
+	BUG_ON(tail_frags < 0);
 
 	/* No room for @n extra page fragments in the SKB. */
 	if (si->nr_frags + n > MAX_SKB_FRAGS) {
@@ -326,11 +326,11 @@ __extend_pgfrags(struct sk_buff *skb_head, struct sk_buff *skb, int from, int n)
 		}
 
 		/* No fragments to shift. */
-		if (from == si->nr_frags)
+		if (!tail_frags)
 			return 0;
 
 		/* Shift @n_excess number of page fragments to new SKB. */
-		for (i = n_excess - 1; i >= 0; --i) {
+		for (i = n_excess - 1; i >= max(n_excess - tail_frags, 0); --i) {
 			f = &si->frags[MAX_SKB_FRAGS - n + i];
 			skb_shinfo(nskb)->frags[i] = *f;
 			e_size += skb_frag_size(f);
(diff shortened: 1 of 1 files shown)


tempesta_fw/ss_skb.c
```diff
@@ -326,23 +326,26 @@ __extend_pgfrags(struct sk_buff *skb_head, struct sk_buff *skb, int from, int n)
...
 	}
-
-	/* Make room for @n page fragments in the SKB. */
-	n_shift = si->nr_frags - from - n_excess;
-	BUG_ON(n_shift < 0);
-	if (n_shift)
+	/*
+	 * Make room for @n page fragments in the SKB. Considering maximum @n
+	 * value must be not greater than 2, the minimum @n_shift value must
+	 * be not less than -1.
```
krizhanovsky: A very subtle place - I spent some time to check the assertion, but really if `tail_frags == 0` here, then `n_excess == 0` too, otherwise we'd return from the function above. Could you please add a comment or some more assertion (WARN_ON/BUG_ON) about linkage between `tail_frags` and `n_excess`.

aleksostapenko: Agree that this place indeed needs more explanation - I'll add detailed comment here.

aleksostapenko: >but really if tail_frags == 0 here, then n_excess == 0 too

Since:
```
tail_frags = si->nr_frags - from;
n_excess =  (si->nr_frags + n > MAX_SKB_FRAGS) ? (si->nr_frags + n - MAX_SKB_FRAGS) : 0;
```
so, if e.g. `si->nr_frags == MAX_SKB_FRAGS == 16`, `from == 16`, `n == 2`, we have `tail_frags == 0` but `n_excess == 2`. But in any case when `tail_frags == 0` we return from the function before - because we needn't any fragments shifting/moving in this case and all the remaining work will be done by the caller.
In general the linkage between `tail_frags` and `n_excess` can be described as following:
1. We must move `n_excess` fragments to next/new skb, except those fragments, which we are inserting ('n') - so we must move **last** `n_excess` fragments: not more than `tail_frags` (i.e. if `n_excess` > `tail_frags`, then the first fragments in next/new skb must be the fragments which are the on newly inserting `n` fragments), and not more than `n_excess` itself (i.e. if `n_excess` < `tail_frags`, then the **end** part of `tail_frags` must be moved to next/new skb and the **begin** part of `tail_frags` must be shifted inside current skb); in the last case `n_excess` - `tail_frags` will be negative, that's why the `max(n_excess - tail_frags, 0)` is used as first (in reverse cycle - as last) fragment which must be moved to next/new skb, and `si->nr_frags - n_excess + i` (which after `n_excess` substitution is equal to `MAX_SKB_FRAGS - n + i`) is used as index (in current skb) of fragment which must be moved.
2. We must shift `tail_frags` fragments inside current skb, except those fragments, which we must move to next/new skb, so this is `n_shift = tail_frags - n_excess` fragments; in case of too small `tail_frags` and/or too big `n`, the value of `n_shift` will be negative; but if maximum value on `n` is limited to 2 - then `n_shift` must be not less then -1.

tempesta_fw/ss_skb.c
```diff
@@ -326,23 +326,26 @@ __extend_pgfrags(struct sk_buff *skb_head, struct sk_buff *skb, int from, int n)
 		}
 
 		/* No fragments to shift. */
-		if (from == si->nr_frags)
+		if (!tail_frags)
 			return 0;
 
 		/* Shift @n_excess number of page fragments to new SKB. */
-		for (i = n_excess - 1; i >= 0; --i) {
+		for (i = n_excess - 1; i >= max(n_excess - tail_frags, 0); --i) {
```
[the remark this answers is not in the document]
krizhanovsky: OK, I see that we really need to move the frag by the index 1.

PR #1169, opened 2019-01-29
Update server APM stats right after response was received

ikoveshnikov:
APM update should happen earlier than cache actions: copying response to cache, cache logic is to become more complex and it adds some Tempesta's latancy delays to the APM stats. In the same time APM stats are used for scheduling and must provide information only about backend, not about Tempesta-backend chain.

Another reason for early updates - HM monitor responses are dropped before response can reach `tfw_http_resp_cache_cb()`

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -3163,6 +3163,16 @@ tfw_http_resp_cache_cb(TfwHttpMsg *msg)
 
 	TFW_DBG2("%s: req = %p, resp = %p\n", __func__, req, resp);
 	/*
+	 * Response will never be sent: client has disconnected already.
+	 * Don't adjust response, simply destroy it.
+	 */
+	if (unlikely(test_bit(TFW_HTTP_B_REQ_DROP, req->flags))) {
+		TFW_DBG2("%s: resp=[%p] dropped: client disconnected\n",
+			 __func__, resp);
+		tfw_http_resp_pair_free(req);
+		return;
+	}
+	/*
 	 * Typically we're at a node far from the node where @resp was
 	 * received, so we do an inter-node transfer. However, this is
 	 * the final place where the response will be stored. Upcoming
@@ -3171,8 +3181,8 @@ tfw_http_resp_cache_cb(TfwHttpMsg *msg)
 	 */
 	if (tfw_http_adjust_resp(resp)) {
 		tfw_http_conn_msg_free((TfwHttpMsg *)resp);
-		tfw_http_send_resp(req, 500, "response dropped:"
-				   " processing error");
+		tfw_http_send_resp(req, 500,
+				   "response dropped: processing error");
 		TFW_INC_STAT_BH(serv.msgs_otherr);
 		return;
 	}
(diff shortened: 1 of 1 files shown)


tempesta_fw/http.c
```diff
@@ -3162,6 +3162,16 @@ tfw_http_resp_cache_cb(TfwHttpMsg *msg)
...
+	/*
+	 * Response will never be sent: client has disconnected already.
+	 * Don't adjust response, simply destroy it.
+	 */
+	if (unlikely(test_bit(TFW_HTTP_B_REQ_DROP, req->flags))) {
+		TFW_DBG2("%s: resp=[%p] dropped: client disconnected\n",
+			 __func__, resp);
+		tfw_http_resp_pair_free(req);
+		return;
+	}
```
krizhanovsky: A client can be disconnected at any time: before we execute a caching process (also not so lightweight) or while the response is in forwarding queue, so it's not obvious why did you choose exactly this place. However, the place looks good - it makes sense to store a response in a cache regardless the client connection, so other clients can hit the cache entry. If this is what you had in mind, then just describe it in the comment above. Otherwise let's have a discussion on this in the chat.

ikoveshnikov: Yes, that was the exact reason. Should have commented it explicitly.

tempesta_fw/http.c
```diff
@@ -3162,6 +3162,16 @@ tfw_http_resp_cache_cb(TfwHttpMsg *msg)
 	TfwHttpReq *req = resp->req;
 
 	TFW_DBG2("%s: req = %p, resp = %p\n", __func__, req, resp);
+	/*
+	 * Response will never be sent: client has disconnected already.
+	 * Don't adjust response, simply destroy it.
+	 */
+	if (unlikely(test_bit(TFW_HTTP_B_REQ_DROP, req->flags))) {
+		TFW_DBG2("%s: resp=[%p] dropped: client disconnected\n",
```
aleksostapenko: Not only about this PR, but it looks like common issue: it seems that we have race condition if
1 thr.: `TFW_HTTP_B_REQ_DROP` set in `tfw_http_conn_cli_drop()`;
2 thr.: `TFW_HTTP_B_REQ_DROP` tested as 1 here (or in `tfw_http_req_evict_dropped()` or in `tfw_http_req_zap_error()`);
2 thr.: `tfw_http_conn_msg_free((TfwHttpMsg *)req)` processed;
1 thr.: the rest of `tfw_http_conn_cli_drop()` is executing (`list_del_init(&req->msg.seq_list)` etc).

tempesta_fw/http.c
```diff
@@ -3311,6 +3307,16 @@ tfw_http_resp_cache(TfwHttpMsg *hmresp)
...
+	/*
+	 * TODO: Currently APM holds the pure roundtrip time (RTT) from
+	 * the time a request is forwarded to the time a response to it
+	 * is received and parsed. Perhaps it makes sense to penalize
+	 * server connections which get broken too often. What would be
+	 * a fast and simple algorithm for that? Keep in mind, that the
+	 * value of RTT has an upper boundary in the APM.
+	 */
+	tfw_apm_update(((TfwServer *)resp->conn->peer)->apmref,
+		       resp->jrxtstamp, resp->jrxtstamp - req->jtxtstamp);
```
aleksostapenko: The `tfw_apm_update()` is also called from `tfw_http_hm_drop_resp()` (below).

Issue #309, opened 2015-10-11
HTTP/2


krizhanovsky:
The initial implementation of [HTTP/2](https://tools.ietf.org/html/rfc7540) was done in https://github.com/tempesta-tech/tempesta/pull/1176 . Need a review and further development of the code. I suggest to create a new pull request instead of developing #1176.

Consider [Cache Digests](http://mnot.github.io/I-D/h2-cache-digest/) as important for [Ideal HTTP Performance](https://www.mnot.net/blog/2016/04/22/ideal-http). 

There is [HTTP/2 test suite](https://github.com/http2/http2-test) which should be used to test the implementation.

Context of the issue is just a **robust and performance** architecture of HTTP/2 (with #687 in mind - we should not let similar issues to pass to the master and probably there will be cheap (quick) opportunities to fix some of #687 problems). All the small extensions, protocol features and bugs should be reported in **small** separate tasks.

An minimal requirement for the issue to be done is that a HTTP/2 capable browser must be able to load tempesta-tech.com via HTTP/2.

## QUIC foundations

Some of HTTP/2 mechanisms are moved to QUIC (#724): streams, frames, compression. So please make required TODO comments and necessary code adjustments for further extensions to QUIC. In particular at least these docs affect current HTTP/2 design:

- [QUIC](https://www.ietf.org/id/draft-ietf-quic-transport-18.txt) transport defines main primitives of QUIC.

- [QPACK](https://www.ietf.org/id/draft-ietf-quic-qpack-06.txt) for HPACK. Probably Avoiding Head-of-Line Blocking is the most influential feature of QPACK which affect synchronization design of HTTP/2 streams and maybe more connection-wide attributes.

- [HTTP/3](https://www.ietf.org/id/draft-ietf-quic-http-18.txt) for the whole new code. Chapters 4 and A.2 discuss similarities and differences in the framing layers.

## Notes

We should not support Huffman encoding, see https://github.com/tempesta-tech/tempesta/pull/1176#discussion_r257789810 . Huffman decoding is only required now. However, in #1125 we'll generate HTTP/2 requests and Huffman encoding [does make difference](https://www.mnot.net/blog/2018/11/27/header_compression) for HTTP headers. So I'd leave current Huffman encoding in the source code, but leave it unused until #1125.

[...]

PR #1176, opened 2019-02-02
HTTP/2 (reopened #694)

krizhanovsky:
Just need to review the code and understand which changes/extensions/fixes are required for #309. After that the branch can be closed and the code can be recommited to a new branch. The PR is just for documentation what is still for TODO.


tempesta_fw/http2/huffman.c
```diff
@@ -0,0 +1,1385 @@
...
+		GET_UWIDE(tail);     \
+	else			     \
+		GET_FIRST_BY_8(tail)
+
+#endif
+
+#endif
+
+unsigned int
+huffman_decode(const char *__restrict source, char *__restrict dst, uintptr_t n)
```
krizhanovsky: The algorithm looks good: `ht_decode` is relatively small (about 4*848 bytes, which is smaller that 4 pages of [huff_decode_table](https://github.com/nghttp2/nghttp2/blob/master/lib/nghttp2_hd_huffman_data.c#L96) and handles the most probable 1/2 of characters in single step. The algorithms is somewhat similar to http://fastcompression.blogspot.com/2015/07/huffman-revisited-part-3-depth-limited.html

However, I'd love to see the proof that the algorithm is really fast, so please write a benchmark test comparing it with [nghttp2_hd_huff_decode()](https://github.com/nghttp2/nghttp2/blob/master/lib/nghttp2_hd_huffman.c#L198).

A good readings about fast Huffman decoders:
https://www.ifi.uzh.ch/dam/jcr:ffffffff-82b7-d340-0000-000026dd8fb6/Prefix.pdf
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.93.9447&rep=rep1&type=pdf
http://fastcompression.blogspot.com/2015/07/huffman-revisited-part-1.html
http://fastcompression.blogspot.com/2015/07/huffman-revisited-part-2-decoder.html
http://fastcompression.blogspot.com/2015/07/huffman-revisited-part-3-depth-limited.html
http://fastcompression.blogspot.com/2015/10/huffman-revisited-part-4-multi-bytes.html
http://fastcompression.blogspot.com/2015/10/huffman-revisited-part-5-combining.html

However I don't see much sense in implementation of the multi-stream decoder since we need to feed the data to our parser and the data chunks aren't so big.

tempesta_fw/http2/hindex.c
```diff
@@ -0,0 +1,882 @@
...
+			vp = NULL;
+		}
+		static_table[i].name = np;
+		static_table[i].value = vp;
+		np = hpack_hash_add(ht, np, NULL);
+		if (vp) {
+			vp = hpack_hash_add(ht, vp, NULL);
+		}
+		Hash_Add(hp, static_table + i, static_table + i);
+		Hash_Add(hn, np, static_table + i);
```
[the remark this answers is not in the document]
aleksostapenko: Agree that approach with static hash table is overcomplicated. But with binary search tree - disturb the necessity of full string comparisons on each node. I'll think deeper about proposed structure. Also some kind of prefix tree may be a suitable variant here (will think in this direction too).

tempesta_fw/http2/hpack_decoder.c
```diff
@@ -0,0 +1,648 @@
...
+	);			    \
+	if (unlikely(* rc)) {	    \
+		goto Bug;	    \
+	}			    \
+	field->value = buffer->str; \
+} while (0)
+
+HTTP2Field *
+hpack_decode(HPack * __restrict hp,
+	     HTTP2Input * __restrict source,
```
aleksostapenko: It seems that 'source' must me passed here with already initialized (i.e. parsed) `TfwStr` (as underlying instance for `HTTP2Input`), but considering HTTP/2 logic - we must initially decode HTTP/2 message and only after that pass it to HTTP parser, where `TfwStr` instances will be initialized.

tempesta_fw/http2/hpack_decoder.c
```diff
@@ -0,0 +1,648 @@
...
+					if (state & HPack_Flags_Huffman_Name) {
+						DPRINTF("Decode %" PRIuPTR
+							" bytes of Huffman data...\n",
+							length);
+						unsigned int hrc;
+
+						buffer_close(source, m);
+						hrc =
+						    huffman_decode_fragments
+						    (source, buffer, length);
```
aleksostapenko: If I correctly understand, `'HTTP2Output *buffer'` is intended for storing the result of HPACK decoding process; regarding the integration with Tempesta FW code it seems that `HTTP2Input` and `HTTP2Output` buffers should be reworked in a way that one resulting buffer should remain (as descriptor of decoded output) and, possibly, incorporate with `TfwStr` descriptor.

tempesta_fw/http2/hpack_decoder.c
```diff
@@ -0,0 +1,648 @@
...
+							    Err_HTTP2_IntegerOveflow;
+						}
+#else
+						hrc =
+						    hpack_add_index(hp->dynamic,
+								    field,
+								    index,
+								    state,
+								    buffer);
+#endif
```
aleksostapenko: I cannot find the static or dynamic table searching during 'hpack_decode()' procedure - only adding indexes into dynamic table.

tempesta_fw/http2/buffers.c
```diff
@@ -0,0 +1,1332 @@
...
+{
+	p->str = str;
+	p->current = 0;
+	p->n = str->len;
+/* p->tail is initialized here only to avoid */
+/* partial writing into cache line, really */
+/* it does not used by buffer_get() call: */
+	p->tail = 0;
+	p->offset = 0;
+}
```
aleksostapenko: One more string descriptor (`HTTP2Input`) over the already existing one (`TfwStr`) seems too complicated; maybe it is worth to extend TfwStr and apply this logic for it or make separate descriptor.

tempesta_fw/http2/hindex.c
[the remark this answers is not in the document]
aleksostapenko: Agree that approach with static hash table is overcomplicated. But with binary search tree - disturb the necessity of full string comparisons on each node. I'll think deeper about proposed structure. Also some kind of prefix tree may be a suitable variant here (will think in this direction too).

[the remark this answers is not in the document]
krizhanovsky: I came to the idea exactly from patricia tree. The point is that in the example above we can, and probably, should treat the stings as long integers and compare longs instead of strings. We can set the most significant bit for the pointers and string lengths to guarantee that if we try to match 8-byte integer with say 3-byte string the set most significant bit in 4th byte will make the difference. Next idea is that we have 32-byte SIMD registers and, if we place the strings in the array with known offsets, we can load several string prefixes and match them concurrently. I have no read to implement concept, but there are several opportunities to couple fast search and efficient insert and eviction.

tempesta_fw/http2/hpack_decoder.c
krizhanovsky: Yes, agree. The implementation is done as a standalone library with some interfaces: the library does some processing on some data which was prepared by some other layer and pass the result to another layer - different logic is ran against different representations of the same data. But it's more efficient to make all the layers to know about each other and run all the processing logic at once on the same data chunk.

This is ideology for QUIC design and Tempesta FW is built on the same principles (we interbreed TCP, TLS, and HTTP) and this makes me think that we can natively integrate QUIC into our design.

tempesta_fw/http2/buffers.c
krizhanovsky: The interfaces are really complicated (seems in aiming to reach generic API for the library). Frankly, I didn't pay enough attention in review of this part of code because it's clear there shouldn't be such data structures.

ikoveshnikov:
Most of the changes from this PR was reworked and integrated into #1368, so the PR can be closed now.

Issue #1115, opened 2018-11-24
Temporal client accounting


krizhanovsky:
The task is from https://github.com/tempesta-tech/tempesta/issues/488#issuecomment-360432734:

From #100: the most urgent thing is to keep security accounting data for a client for some time after the last client connection is closed. This is very important to track client security limits properly for `Connection: closed` connections. See https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/client.c#L89

Since we have to evict client accounting data after 'some time', it has sense to store them in a TDB table. At this time TDB has issues with eviction - just let's leave with them untill they are fixed in Beta.

Next, the hash key for searching a client must be calculated by User-Agent plus IP address, otherwise, if no User-Agent, only IP address.

Currently, we use netowork IP address as the address of a client, however if a client work through a forward proxy, then a proxy can pass it's IP address by the first item in `X-Forwarded-For`, so if the header is present, then we shall reinsert the `TfwClient` in TDB with different key.

The reinsert operation must be implemented on TDB layer as a new routine `tdb_entry_reinsert()` accepting current and new keys. The function must call `tdb_htrie_insert()`, copying the data from the previous location, and a new empty `tdb_htrie_delete()` left as TODO for #515.

Traversal all the clients on shutdown must be done by a new TDB routine `tdb_walk()` which is just a walk on the radix tree plus `tdb_rec_next()` call. The routine must expect that there is no tree writers (the operation must be done after closing the listener socket and sycnrhonization barrier on all softirqs).

Issue #515, opened 2016-05-25
[Cache] background revalidation and eviction


krizhanovsky:
`tfw_cache_mgr` thread must traverse Web-cache entries

krizhanovsky:
It seems there is some race in the lock-free index or we actually hit the https://github.com/tempesta-tech/tempesta/issues/500 problem in scenario from #1435 : multiple parallel requests to large file
```
./wrk -d 3600 -c 16000 -t 8 -H 'connection: close' https://debian:443/research/web_acceleration_mechanics.pdf
```
combined with the Tempesta restart in the VM
```
# while :; do ./scripts/tempesta.sh --restart; sleep 30; done
```
sometimes produce warnings like
```
[ 1103.775556] [tdb] ERROR: out of free space
[ 1103.810415] [tdb] ERROR: out of free space
[ 1103.845177] [tdb] ERROR: Cannot allocate cache entry for key=0x37bed983985f3ea7
[ 1103.929897] [tdb] ERROR: out of free space
[ 1103.949002] [tdb] ERROR: Cannot allocate cache entry for key=0x37bed983985f3ea7
[ 1103.984315] [tdb] ERROR: out of free space
[ 1104.010543] [tdb] ERROR: Cannot allocate cache entry for key=0x37bed983985f3ea7
[ 1104.070816] [tdb] ERROR: out of free space
[ 1104.080997] [tdb] ERROR: Cannot allocate cache entry for key=0x37bed983985f3ea7
[ 1104.151540] [tdb] ERROR: out of free space
[ 1104.158845] [tdb] ERROR: Cannot allocate cache entry for key=0x37bed983985f3ea7
[ 1104.199489] [tdb] ERROR: out of free space
[ 1104.231891] [tdb] ERROR: Cannot allocate cache entry for key=0x37bed983985f3ea7
....
```

const-t:
I've made few roughly benchmarks HTTP2 with enabled caching.

`h2load -c700 -m100 --duration=30 -t2  https://debian`

## Tempesta

1kb response
```
finished in 30.14s, 337279.80 req/s, 393.06MB/s
requests: 10118394 total, 10188394 started, 10118394 done, 10118394 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 10118394 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 11.52GB (12364696856) total, 1.70GB (1821310920) headers (space savings 23.08%), 9.65GB (10361235456) data
[...]

PR #1178, opened 2019-02-07
Temporal client accounting

avbelov23:
Requires #1115

#1145

### DIFF
--- a/etc/tempesta_fw.conf
+++ b/etc/tempesta_fw.conf
@@ -1129,3 +1129,23 @@
 # Default:
 #   Health monitor is disabled.
 #
+
+# TAG: client_db
+#
+# Path to a client database file used as a storage for clients info.
+# The same as cache_db.
+#
+# Default:
+#   client_db /opt/tempesta/db/client.tdb;
+#
+
+# TAG: client_tbl_size
+#
+# Size of client drop table.
+#
+# Syntax:
+#   client_tbl_size SIZE
+#
+# Default:
+#   client_tbl_size 16777216;  # 16MB
+#
--- a/tempesta_db/core/htrie.c
+++ b/tempesta_db/core/htrie.c
@@ -806,6 +806,28 @@ tdb_htrie_lookup(TdbHdr *dbh, unsigned long key)
 	return TDB_PTR(dbh, o);
 }
 
+#define TDB_HTRIE_FOREACH_REC(dbh, b_tmp, b, r, body)			\
+	read_lock_bh(&(*b)->lock);					\
+	do {								\
+		r = TDB_HTRIE_BCKT_1ST_REC(*b);				\
+		do {							\
+			size_t rlen = sizeof(*r) +			\
+				      TDB_HTRIE_RBODYLEN(dbh, r);	\
+			rlen = TDB_HTRIE_RALIGN(rlen);			\
+			if ((char *)r + rlen - (char *)*b		\
+				> TDB_HTRIE_MINDREC			\
+			    && r != TDB_HTRIE_BCKT_1ST_REC(*b))		\
+				break;					\
+			body;						\
+			r = (TdbRec *)((char *)r + rlen);		\
+		} while ((char *)r + sizeof(*r) - (char *)*b		\
+			 <= TDB_HTRIE_MINDREC);				\
+		b_tmp = TDB_HTRIE_BUCKET_NEXT(dbh, *b);			\
+		if (b_tmp)						\
+			read_lock_bh(&b_tmp->lock);			\
+		read_unlock_bh(&(*b)->lock);				\
+		*b = b_tmp;						\
+	} while (*b)
 /**
  * Iterate over all records in collision chain with locked buckets.
  * Buckets are inspected according to following rules:
@@ -821,29 +843,11 @@ tdb_htrie_bscan_for_rec(TdbHdr *dbh, TdbBucket **b, unsigned long key)
 	TdbBucket *b_tmp;
 	TdbRec *r;
 
-	read_lock_bh(&(*b)->lock);
-
-	do {
-		r = TDB_HTRIE_BCKT_1ST_REC(*b);
-		do {
-			size_t rlen = sizeof(*r) + TDB_HTRIE_RBODYLEN(dbh, r);
-			rlen = TDB_HTRIE_RALIGN(rlen);
-			if ((char *)r + rlen - (char *)*b > TDB_HTRIE_MINDREC
-			    && r != TDB_HTRIE_BCKT_1ST_REC(*b))
-				break;
-			if (tdb_live_rec(dbh, r) && r->key == key)
-				/* Unlock the bucket by tdb_rec_put(). */
-				return r;
-			r = (TdbRec *)((char *)r + rlen);
-		} while ((char *)r + sizeof(*r) - (char *)*b
-			 <= TDB_HTRIE_MINDREC);
-
-		b_tmp = TDB_HTRIE_BUCKET_NEXT(dbh, *b);
-		if (b_tmp)
-			read_lock_bh(&b_tmp->lock);
-		read_unlock_bh(&(*b)->lock);
-		*b = b_tmp;
-	} while (*b);
+	TDB_HTRIE_FOREACH_REC(dbh, b_tmp, b, r, {
+		if (tdb_live_rec(dbh, r) && r->key == key)
+			/* Unlock the bucket by tdb_rec_put(). */
+			return r;
+	});
 
 	return NULL;
 }
@@ -930,4 +934,76 @@ void
 tdb_htrie_exit(TdbHdr *dbh)
 {
 	free_percpu(dbh->pcpu);
+}
+
+static int
+tdb_htrie_bucket_walk(TdbHdr *dbh, TdbBucket *b, int (*fn)(void *))
+{
+	TdbBucket *b_tmp;
+	TdbRec *r;
+
+	TDB_HTRIE_FOREACH_REC(dbh, b_tmp, &b, r, {
+		if (tdb_live_rec(dbh, r)) {
+			int res = fn(r->data);
+			if (unlikely(res)) {
+				read_unlock_bh(&b->lock);
+				return res;
+			}
+		}
+	});
+
+	return 0;
+}
+
+static int
+tdb_htrie_node_visit(TdbHdr *dbh, TdbHtrieNode *node, int (*fn)(void *))
+{
+	int bits;
+	int res;
+
+	for (bits = 0; bits < TDB_HTRIE_FANOUT; ++bits) {
+		unsigned long o;
+
+		BUG_ON(TDB_HTRIE_RESOLVED(bits));
+
+		o = node->shifts[bits];
+
+		if (likely(!o))
+			continue;
+
+		BUG_ON(TDB_DI2O(o & ~TDB_HTRIE_DBIT) < TDB_HDR_SZ(dbh) + sizeof(TdbExt)
+			   || TDB_DI2O(o & ~TDB_HTRIE_DBIT) > dbh->dbsz);
+
+		if (o & TDB_HTRIE_DBIT) {
+			TdbBucket *b;
+
+			/* We're at a data pointer - resolve it. */
+			o ^= TDB_HTRIE_DBIT;
+			BUG_ON(!o);
+
+			b = (TdbBucket *)TDB_PTR(dbh, TDB_DI2O(o));
+			res = tdb_htrie_bucket_walk(dbh, b, fn);
+			if (unlikely(res))
+				return res;
+		} else {
+			/*
+			 * The recursion depth being hard-limited.
+			 * The function has the deepest nesting 16.
+			 */
+			res = tdb_htrie_node_visit(dbh, TDB_PTR(dbh,
+						   TDB_II2O(o)), fn);
+			if (unlikely(res))
+				return res;
+		}
+	}
+
+	return 0;
+}
+
+int
+tdb_htrie_walk(TdbHdr *dbh, int (*fn)(void *))
+{
+	TdbHtrieNode *node = TDB_HTRIE_ROOT(dbh);
+
+	return tdb_htrie_node_visit(dbh, node, fn);
 }
--- a/tempesta_db/core/main.c
+++ b/tempesta_db/core/main.c
@@ -28,12 +28,15 @@
 #include "table.h"
 #include "tdb_if.h"
 
-#define TDB_VERSION	"0.1.16"
+#define TDB_VERSION	"0.1.17"
 
 MODULE_AUTHOR("Tempesta Technologies");
 MODULE_DESCRIPTION("Tempesta DB");
 MODULE_VERSION(TDB_VERSION);
 MODULE_LICENSE("GPL");
+
+/* Lock for atomic execution of lookup and create a record TDB */
+static DEFINE_SPINLOCK(get_alloc_lock);
 
 /**
  * Create TDB entry and copy @len contiguous bytes from @data to the entry.
@@ -52,6 +55,7 @@ EXPORT_SYMBOL(tdb_entry_create);
 
 /**
  * Create TDB entry to store @len bytes.
+ * TODO #515 function must holds a lock upon return.
  */
 TdbRec *
 tdb_entry_alloc(TDB *db, unsigned long key, size_t *len)
@@ -221,6 +225,63 @@ tdb_get_db(const char *path, int node)
 
 	return tdb_get(db);
 }
+
+/**
+ * Lookup and get a record if the record is found or create TDB entry to store
+ * @len bytes. If record exist then since we don't copy returned records,
+ * we have to lock the memory location where the record is placed and
+ * the user must call tdb_rec_put() when finish with the record.
+ *
+ * The caller must not call sleeping functions during work with the record.
+ * Typically there is only one large record per bucket, so the bucket lock
+ * is exactly the same as to lock the record. While there could be many
+ * small records in a bucket, so the caller should not perform long jobs
+ * with small records.
+ *
+ * @return pointer to record with acquired bucket lock if the record is
+ * found and create TDB entry without acquired locks otherwise.
+ *
+ * TODO #515 rework the function in lock-free way.
+ * TODO #515 TDB must be extended to support small records with constant memory
+ * address.
+ */
+TdbRec *
+tdb_rec_get_alloc(TDB *db, unsigned long key, size_t *len,
+		  bool (*predicate)(TdbRec *, void (*)(void *), void *),
+		  void (*init)(TdbRec *, void (*)(void *), void *),
+		  void (*cb)(void *), void *data, bool *is_new)
+{
+	TdbIter iter;
+	TdbRec *r;
+
+	spin_lock(&get_alloc_lock);
+
+	*is_new = false;
+	iter = tdb_rec_get(db, key);
+	while (!TDB_ITER_BAD(iter)) {
+		if ((*predicate)(iter.rec, cb, data)) {
+			spin_unlock(&get_alloc_lock);
+			return iter.rec;
+		}
+		tdb_rec_next(db, &iter);
+	}
+
+	*is_new = true;
+	r = tdb_entry_alloc(db, key, len);
+	init(r, cb, data);
+
+	spin_unlock(&get_alloc_lock);
+
+	return r;
+}
+EXPORT_SYMBOL(tdb_rec_get_alloc);
+
+int
+tdb_entry_walk(TDB *db, int (*fn)(void *))
+{
+	return tdb_htrie_walk(db->hdr, fn);
+}
+EXPORT_SYMBOL(tdb_entry_walk);
 
 /**
  * Open database file and @return its descriptor.
--- a/tempesta_fw/client.c
+++ b/tempesta_fw/client.c
@@ -4,7 +4,7 @@
  * Clients handling.
  *
  * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
- * Copyright (C) 2015-2018 Tempesta Technologies, Inc.
+ * Copyright (C) 2015-2019 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
@@ -28,29 +28,35 @@
 #include "connection.h"
 #include "log.h"
 #include "procfs.h"
+#include "tdb.h"
+#include "lib/str.h"
+#include "lib/common.h"
 
-#define CLI_HASH_BITS	17
-#define CLI_HASH_SZ	(1 << CLI_HASH_BITS)
+static struct {
+	unsigned int	db_size;
+	const char	*db_path;
+	unsigned int	expires_time;
+} client_cfg __read_mostly;
 
-typedef struct {
-	struct hlist_head	list;
-	spinlock_t		lock;
-} CliHashBucket;
-
-/*
- * TODO probably not the best container for the task and
- * HTrie should be used instead.
+/**
+ * Client tdb entry.
+ *
+ * @cli			- client descriptor
+ * @expires		- expiration time for the client descriptor after all
+ *			  connections are closed
+ * @lock		- lock for atomic change @expires and @users;
+ * @users		- reference counter.
+ * 			  Expiration state will begind, when the counter reaches
+ *			  zero
  */
-CliHashBucket cli_hash[CLI_HASH_SZ] = {
-	[0 ... (CLI_HASH_SZ - 1)] = {
-		HLIST_HEAD_INIT,
-	}
-};
+typedef struct {
+	TfwClient	cli;
+	time_t		expires;
+	spinlock_t	lock;
+	atomic_t	users;
+} TfwClientEntry;
 
-static struct kmem_cache *cli_cache;
-
-/* Total number of created clients. */
-static atomic64_t act_cli_n = ATOMIC64_INIT(0);
+static TDB *client_db;
 
 /**
  * Called when a client socket is closed.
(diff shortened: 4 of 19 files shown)


tempesta_fw/http.c
```diff
@@ -2930,6 +2934,37 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
...
+	 */
+	s_xff = req->h_tbl->tbl[TFW_HTTP_HDR_X_FORWARDED_FOR];
+	s_ip = tfw_str_next_str_val(&s_xff);
+	nchunks = 0;
+	TFW_STR_FOR_EACH_CHUNK(c, &s_ip, end) {
+		if (!(c->flags & TFW_STR_VALUE))
+			break;
+		nchunks++;
+	}
+	s_ip.nchunks = nchunks;
```
krizhanovsky: `tfw_http_req_process()` is too long and overloaded now, so it's better to move the str logic to a small helping function. I think the whole new code with the new local variables can and should be moved to the helper.

tempesta_fw/http.c
```diff
@@ -2930,6 +2934,37 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
...
+
+	if (TFW_STR_EMPTY(&s_ip) || tfw_addr_pton(&s_ip, &addr) != 0) {
+		ss_getpeername(conn->sk, &addr);
+	}
```
krizhanovsky: There is no sense to execute the code if we have no User-Agent since the client was already obtained in `tfw_sock_clnt_new()` for the same address.

BTW we do not use braces for single-line IF statements unless there is multi-line ELSE.

tempesta_fw/http.c
```diff
@@ -2930,6 +2934,37 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
...
+	tfw_http_msg_clnthdr_val(&req->h_tbl->tbl[TFW_HTTP_HDR_USER_AGENT],
+				 TFW_HTTP_HDR_USER_AGENT, &s_user_agent);
+
+	if (!TFW_STR_EMPTY(&s_ip) || !TFW_STR_EMPTY(&s_user_agent)) {
+		TfwClient *cli = tfw_client_obtain(addr, &s_user_agent, NULL);
+		TfwClient *conn_cli = (TfwClient *)conn->peer;
+		if (cli && conn_cli != cli) {
+			tfw_connection_unlink_from_peer(conn);
+			tfw_client_put(conn_cli);
+			tfw_connection_link_peer(conn, (TfwPeer *)cli);
```
krizhanovsky: A very subtle place. Here we're under the socket lock acquired by the TCP caller and asserted in `ss_tcp_data_ready()`, so we have no concurrent ingress requests. However, what if some other parallel softirq is processing the client and connection, e.g.on processing server response? Do we have such places which work with `TfwClient` on server responses? A comment about the operations safety is required. We had plenty of bugs on races with the connections linkage.

tempesta_fw/http_parser.c
```diff
@@ -4075,9 +4075,9 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len,
...
 	/* 'X-Forwarded-For:*OWS' is read, process field-value. */
-	TFW_HTTP_PARSE_SPECHDR_VAL(Req_HdrX_Forwarded_ForV, Req_I_XFF,
-				   msg, __req_parse_x_forwarded_for,
-				   TFW_HTTP_HDR_X_FORWARDED_FOR);
+	__TFW_HTTP_PARSE_SPECHDR_VAL(Req_HdrX_Forwarded_ForV, Req_I_XFF,
+				     msg, __req_parse_x_forwarded_for,
+				     TFW_HTTP_HDR_X_FORWARDED_FOR, 0);
 
 	/* 'User-Agent:*OWS' is read, process field-value. */
 	TFW_HTTP_PARSE_SPECHDR_VAL(Req_HdrUser_AgentV, Req_I_UserAgent,
```
krizhanovsky: This whole HTTP parser change must have a unit test in `t/unit/test_http_parser.c`, probably a new one, that we get the right IP address, which we pass to `tfw_client_obtain()`, from the header.

tempesta_fw/http_parser.c
```diff
@@ -3223,27 +3223,27 @@ __req_parse_x_forwarded_for(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
 		/* Eat OWS before the node ID. */
 		if (unlikely(IS_WS(c)))
-			__FSM_I_MOVE(Req_I_XFF);
+			__FSM_I_MOVE_fixup(Req_I_XFF, 1, 0);
 		/*
 		 * Eat IP address or host name.
 		 *
 		 * TODO: parse/validate IP addresses and textual IDs.
 		 * Currently we just validate separate characters, but the
 		 * whole value may be invalid (e.g. "---[_..[[").
```
krizhanovsky: The comment looks scary in context of #1115: we definitely do not want to identify client by garbage like `---[_..[[`. I'm wondering why don't we have the string in `suspicious_x_forwarded_for` test in `t/unit/test_http_parser.c`. So please add the string to the test. Next, we need a verification of IP address if we use it (previously we just forwarded it to a backend w/o any processing), i.e. need to implement the TODO.

tempesta_db/core/htrie.c
```diff
@@ -931,3 +931,90 @@ tdb_htrie_exit(TdbHdr *dbh)
 {
 	free_percpu(dbh->pcpu);
 }
+
+static int
+tdb_bucket_walk(TdbHdr *dbh, TdbBucket *b, int (*fn)(void *))
```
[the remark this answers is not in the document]
avbelov23: No, these functions have different return values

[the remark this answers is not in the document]
krizhanovsky: Does this really makes the unification impossible?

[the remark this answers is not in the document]
avbelov23: Рossible through macros

tempesta_fw/http_parser.c
avbelov23: Now we check the IP address in https://github.com/tempesta-tech/tempesta/pull/1178/files/18cc11ef8c963acf47447f8ffac211125e2f6d6e#diff-0d15717c2e20453f4271befe0024c2d4R2951

tempesta_fw/http.c
ikoveshnikov: I against the change. We mustn't update the `TfwCliConn->peer` on processing a random response. I suppose that a new member `TfwHttpReq->peer` is required:
- The peer in `TfwCliConn->peer` - is hop-by-hop peer, which can be an intermediate proxy.
- The peer in `TfwHttpReq->peer` - is end-to-end peer, which is (mostly but not always) browser.

Take a look at `http_resp_code_block` Frang directive. The directive adds a _server connection_ hook, and it's targeted to block clients that e.g. tries to bruteforce passwords. It was discussed in the chat some day, that we don't want to block the whole intermediate proxy, only some clients behind it. So we'll need to some how find an end-to-end peer, having only `TfwHttpReq` in sight.

With the changes proposed, `TfwHttpReq->conn->peer` will return a random end-to-end client and ban it on.

tempesta_fw/client.c
```diff
@@ -28,121 +28,117 @@
...
 }
 EXPORT_SYMBOL(tfw_client_put);
 
+static bool
+addr_eq(TdbRec *rec, void *data)
+{
+	TfwClientEntry *ent = (TfwClientEntry *)rec->data;
+	TfwClient *cli = &ent->cli;
+
+	return !memcmp_fast(&cli->addr, data, sizeof(cli->addr));
```
ikoveshnikov: The function is used to find the `TfwClient` structure matching to the client's address. But all fields of `TfwAddr` are matched, including port. This will allocate a new `TfwClient` structure for connection with the same ip, but different source port. It's incorrect and insecure behaviour. Bots will just open a new connection, but we will treat it as a new client.

Previously, only IP address was used to match `TfwClient` structure for the client.

tempesta_fw/client.c
```diff
@@ -28,121 +28,117 @@
...
-		spin_unlock(cli->hb_lock);
-		return;
-	}
-	BUG_ON(!list_empty(&cli->conn_list));
-
-	hlist_del(&cli->hentry);
-
-	spin_unlock(cli->hb_lock);
+	ent = (TfwClientEntry *)cli;
+	ent->expires = tfw_current_timestamp() + client_cfg.lifetime;
```
ikoveshnikov: Need to hold a tdb lock here. `ent->expires` may be modified in both `tfw_client_put()` and `tfw_client_obtain()` simultaneously.

etc/tempesta_fw.conf
```diff
@@ -1129,3 +1129,39 @@
...
+#
+# Syntax:
+#   client_lifetime NUM;
+#
+# Example:
+#   client_lifetime 3600;
+#
+# Default:
+#   client_lifetime 0;
+#
```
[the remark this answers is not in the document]
ikoveshnikov: We don't have any dependencies between this parameter and other *lifetime parameters.  The directive say `Life time of client accounting data after last client connection was closed`. It's closer to garbage collecting timeout: how many time to wait before _unused_ TfwClient entry will be destroyed. 

But dependencies between `session_lifetime` and frang limits truly exist.

> Probably it makes sense to generalize session_lifetime to client records lifetime and our sessions?
Sounds good. There was a discussion one day, that we would like to block certain clients, not proxies that they use. Probably frang accounting data should be also stored in `TfwHttpSess`?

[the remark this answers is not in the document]
krizhanovsky: It could make sense in context of discussion https://github.com/tempesta-tech/tempesta/pull/1178/files#r255660386 . Previously we obtained `TfwClient` solely by IP address, i.e. we essentially limited TCP connections. The granularity didn't satisfy us and we came to obtaining clients based on HTTP headers, so now we have sticky cookie, User-Agent, X-Forwaded-For whatever and we should limit clients identified by their HTTP data.

And we can leave all TCP connection limiting to nftables layer while Tempesta cares solely on HTTP limiting - clear and simple.

tempesta_fw/http_parser.c
krizhanovsky: Ok, probably we can leave the parser code as is for now. But 2 things are required:

1. a test, probably functional, that we do really block a client with malformed `X-Forwarded-For`

2. block the client if we can not translate its `X-Forwarded-For` value to an IP address instead of just silently ignore the input

tempesta_db/core/htrie.c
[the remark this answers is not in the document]
i-rinat: Code hidden behind macroses make me sad. But that's a personal preference. Aside from that, there are potential performance implications of such unification. Instead of direct calls, they become indirect through a function pointer.

I haven't tested performance impact of indirect vs. direct calls, so that's not clear. But here are some results from others: https://gist.github.com/rianhunter/0be8dc116b120ad5fdd4. Be sure to read the comments too, they are important.

tempesta_db/core/main.c
```diff
@@ -222,6 +224,41 @@ tdb_get_db(const char *path, int node)
...
+	while (!TDB_ITER_BAD(iter)) {
+		if ((*predicate)(iter.rec, data)) {
+			spin_unlock(&get_alloc_lock);
+			return iter.rec;
+		}
+		tdb_rec_next(db, &iter);
+	}
+
+	*is_new = true;
+	r = tdb_entry_alloc(db, key, len);
```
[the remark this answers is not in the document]
i-rinat: I see that `tdb_rec_get()` returns with a lock to the bucket held. But couldn't find in the code if `tdb_entry_alloc()` holds a lock upon return. I believe it is not. So if we return from the function if `(*predicate)()` produces `true`, lock is held. But if we create a new record, it is not.

etc/tempesta_fw.conf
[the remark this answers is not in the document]
avbelov23: There is nothing about sticky cookies in #1115

tempesta_db/core/htrie.c
```diff
@@ -931,3 +931,90 @@ tdb_htrie_exit(TdbHdr *dbh)
...
+				tdb_node_visit(dbh, TDB_PTR(dbh, TDB_II2O(o)),
+					       fn);
+		}
+	}
+
+	return 0;
+}
+
+int
+tdb_walk(TdbHdr *dbh, int (*fn)(void *))
```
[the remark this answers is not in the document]
avbelov23: We usually use foreach for macros when the body of the loop goes beyond the foreach.

tempesta_fw/http.c
```diff
@@ -2930,6 +2934,37 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
...
+
+	if (TFW_STR_EMPTY(&s_ip) || tfw_addr_pton(&s_ip, &addr) != 0) {
+		ss_getpeername(conn->sk, &addr);
+	}
+
+	tfw_http_msg_clnthdr_val(&req->h_tbl->tbl[TFW_HTTP_HDR_USER_AGENT],
+				 TFW_HTTP_HDR_USER_AGENT, &s_user_agent);
+
+	if (!TFW_STR_EMPTY(&s_ip) || !TFW_STR_EMPTY(&s_user_agent)) {
+		TfwClient *cli = tfw_client_obtain(addr, &s_user_agent, NULL);
```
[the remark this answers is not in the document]
avbelov23: Yes. I found it myself

tempesta_fw/client.c
```diff
@@ -28,121 +28,136 @@
...
 
-	hb = &cli_hash[hash_min(key, CLI_HASH_BITS)];
-
-	spin_lock(&hb->lock);
+	if (cli_addr)
+		key ^= hash_calc((const char *)&cli_addr->sin6_addr,
+				 sizeof(cli_addr->sin6_addr));
+	if (user_agent) {
+		char buf[256];
+		size_t len = tfw_str_to_cstr(user_agent, buf, sizeof(buf));
```
ikoveshnikov: You spin over TfwStr chunks twice: first, copying string to buffer, second time - during hash calculation. This is not very good. Copy operation should be avoided. See https://github.com/tempesta-tech/tempesta/blob/5adf46dd9b94c9fa67bdd0233e3b836bc4931ad9/tempesta_fw/hash.c#L33 I've wrote about this function.

avbelov23: User-Agent can be very large. It will slowly work

krizhanovsky: The comment isn't addressed. It makes sense to limit how many bytes we use for hashing, but it'd appreciated if you extend `tfw_hash_str()` with a length of data to process from `str`. The copying isn't nice at all.

tempesta_fw/client.c
```diff
@@ -28,121 +28,136 @@
...
+
 	TFW_DBG2("put client %p, conn_users=%d\n",
 		 cli, atomic_read(&cli->conn_users));
 
 	if (!atomic_dec_and_test(&cli->conn_users))
 		return;
 
-	spin_lock(cli->hb_lock);
+	rec = (TdbRec *)((void *)cli - offsetof(TdbRec, data));
+	tdb_rec_get_lock(rec);
```
krizhanovsky: `tfw_client_put()` releases the bucket read lock acquired in `tfw_client_obtain()`, so this is a nested and wrong lock.

tempesta_db/core/main.c
```diff
@@ -222,6 +225,72 @@ tdb_get_db(const char *path, int node)
...
+ * The caller must not call sleeping functions during work with the record.
+ * Typically there is only one large record per bucket, so the bucket lock
+ * is exactly the same as to lock the record. While there could be many
+ * small records in a bucket, so the caller should not perform long jobs
+ * with small records.
+ *
+ * @return pointer to record with acquired bucket lock if the record is
+ * found and create TDB entry without acquired locks otherwise.
+ *
+ * TODO #515 rework the function is lock-free way
```
krizhanovsky: Should be "**in** lock-free" and use a dot at the end of sentences. Please be more accurate in comments.

tempesta_fw/client.c
krizhanovsky: Why?

tempesta_fw/http_parser.c
krizhanovsky: There are still no unit tests neither for the new TDB routines nor for the parser changes.

tempesta_fw/http.c
```diff
@@ -2930,6 +2959,26 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
...
+		tfw_http_msg_clnthdr_val(ua, TFW_HTTP_HDR_USER_AGENT,
+					 &s_user_agent);
+		cli = tfw_client_obtain(conn_cli->addr, &addr, &s_user_agent,
+					NULL);
+		if (cli && cli != conn_cli)
+			req->peer = cli;
+		else
+			tfw_client_put(cli);
+	}
+
```
krizhanovsky: Why the whole this block can't be moved to a new function like `tfw_http_req_client_link()` and here we can call only
```C
if ((r = tfw_http_req_cient_link(req)))
    return r;
```

tempesta_fw/http.c
```diff
@@ -2930,6 +2959,26 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
...
+		conn_cli = (TfwClient *)conn->peer;
+		ua = &req->h_tbl->tbl[TFW_HTTP_HDR_USER_AGENT];
+		tfw_http_msg_clnthdr_val(ua, TFW_HTTP_HDR_USER_AGENT,
+					 &s_user_agent);
+		cli = tfw_client_obtain(conn_cli->addr, &addr, &s_user_agent,
+					NULL);
+		if (cli && cli != conn_cli)
+			req->peer = cli;
+		else
+			tfw_client_put(cli);
```
krizhanovsky: Will be a crash on `cli == NULL`

tempesta_fw/client.c
krizhanovsky: We're in big trouble actually with the solution and whole #1115: TDB is simply wasn't designed to keep directly referenced data. The current case for filter or cache tables is to lookup an entry, do some quick operations with it and forget about it. Clients handling is very different: we keep a pointer to stored TDB entry in many places as `TfwClient *peer`, i.e. once created, a client must have the same memory address.

Currenty TDB uses the read lock to protect a referenced entry from being reallocated on HTrie index split (node bursting), so while we use `TfwClient`, i.e. there is a `peer` pointer, the bucket can not be split. Having that we start from only 16 items in root node, we have plenty of splits first time and deadlocks will occur.

A possible solution could be dereference HTrie index each time on `peer` dereferencing, but it's too expensive. TDB must be extended to support small records with constant memory address. Actually, large records are always placed in separate buckets and only small records are compacted into one bucket. Let's leave making TDB nice for #515 and now do a very simple thing: make `TfwClient` data structure large enough to not to be processed in small records TDB work flow. Please don't forget to write `TODO #515` as a comment for the `TfwClient` data structure enhancement.

tempesta_fw/client.c
```diff
@@ -28,119 +28,158 @@
...
+	}
 
-	TFW_DBG2("client %p, conn_users=%d\n",
-		 cli, atomic_read(&cli->conn_users));
+	TFW_DBG2("client %p, conn_users=%d\n", cli, conn_users);
+
+	if (!is_new)
+		tdb_rec_put(rec);
 
 	spin_unlock(&hb->lock);
```
krizhanovsky: Why do we put the record? If there is 2 connections in the system referring to the same client, then they both must keep read locks for the TDB bucket containing the client record.

What's the paired lock acquisition for `spin_unlock(&hb->lock);`?

etc/tempesta_fw.conf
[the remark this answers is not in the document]
krizhanovsky: Please just add the lifetime descrption to the same Wiki page

tempesta_db/core/main.c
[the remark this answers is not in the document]
krizhanovsky: The comment seems addressed in wrong way

tempesta_fw/client.c
```diff
@@ -28,119 +28,158 @@
...
 /*
  * TODO probably not the best container for the task and
  * HTrie should be used instead.
  */
-CliHashBucket cli_hash[CLI_HASH_SZ] = {
-	[0 ... (CLI_HASH_SZ - 1)] = {
-		HLIST_HEAD_INIT,
-	}
-};
+CliHashBucket cli_hash[CLI_HASH_SZ];
```
krizhanovsky: Why there is `cli_hash` again?

tempesta_fw/client.c
avbelov23: The record doesn't change its location in TDB, since it is more than TDB_HTRIE_MINDREC, and we need to unlock the bucket with the client as soon as possible.

etc/tempesta_fw.conf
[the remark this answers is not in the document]
avbelov23: Decided that it still depends from the time frame http_resp_code_block

tempesta_fw/client.c
```diff
@@ -28,122 +28,161 @@
...
 
 typedef struct {
-	struct hlist_head	list;
-	spinlock_t		lock;
-} CliHashBucket;
+	TfwClient	cli;
+	time_t		expires;
+	spinlock_t	lock;
+	atomic_t	users;
+} TfwClientEntry;
```
aleksostapenko: Explanatory comments for structure fields are appreciated, especially - for `lock` field (which protect against `tfw_client_obtain/tfw_client_put` race condition), since we have now rather complicated synchronization case in `tfw_client_obtain()` -> `tdb_rec_get_alloc()` with three separate nested locks with different protected resources.
In this context - it would also be nice to add comment for global `get_alloc_lock` (in tempesta_db/core/main.c), which provides now atomic get/alloc TDB operation and protection against races during instances' constructor execution.

tempesta_fw/http.c
```diff
@@ -2811,6 +2815,58 @@ tfw_http_chop_skb(TfwHttpMsg *hm, struct sk_buff *skb,
...
+
+		conn_cli = (TfwClient *)conn->peer;
+		ua = &req->h_tbl->tbl[TFW_HTTP_HDR_USER_AGENT];
+		tfw_http_msg_clnthdr_val(ua, TFW_HTTP_HDR_USER_AGENT,
+					 &s_user_agent);
+		cli = tfw_client_obtain(conn_cli->addr, &addr, &s_user_agent,
+					NULL);
+		if (cli) {
+			if (cli != conn_cli)
+				req->peer = cli;
```
i-rinat: `req->peer` most probably was obtained by calling `tfw_client_obtain()`. Since this assignment loses `req->peer`, shouldn't it release it by calling `tfw_client_put()`?

i-rinat: Confused with `conn->peer` which is set for every connection. Unlike that, `req->peer` is only set here, so there is no need to release previous value.

tempesta_fw/main.c
```diff
@@ -126,10 +126,8 @@ tfw_cleanup(struct list_head *mod_list)
 
 	tfw_cfg_cleanup(mod_list);
 
-	if (!tfw_runstate_is_reconfig()) {
+	if (!tfw_runstate_is_reconfig())
 		tfw_sg_wait_release();
-		tfw_cli_wait_release();
```
ikoveshnikov: As long I remember, this was introduces to be sure that all client related callbacks has finished and all client connections was destroyed. It's required since client connections destruction is asynchronous operations. What guarantees it now?

aleksostapenko: The reasons for this change are explained in https://github.com/tempesta-tech/tempesta/pull/1145#discussion_r249758558.

ikoveshnikov: Ok, thanks.

tempesta_fw/vhost.c
```diff
@@ -1603,6 +1604,8 @@ tfw_cfgop_frang_rsp_code_block(TfwCfgSpec *cs, TfwCfgEntry *ce, FrangCfg *conf)
 	    || frang_parse_ushort(ce->vals[ce->val_n - 1], &cb->tf))
 		return -EINVAL;
 
+	tfw_client_set_lifetime(cb->tf);
```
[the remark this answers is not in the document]
avbelov23: This will lead to duplication of the collection timeout code, as there are also `http_resp_code_block` in `locations`.

tempesta_fw/client.c
```diff
@@ -28,122 +28,173 @@
...
+	key = hash_calc((const char *)&addr.sin6_addr,
+			sizeof(addr.sin6_addr));
 
-	TFW_DBG2("client %p, conn_users=%d\n",
-		 cli, atomic_read(&cli->conn_users));
+	if (cli_addr)
+		key ^= hash_calc((const char *)&cli_addr->sin6_addr,
+				 sizeof(cli_addr->sin6_addr));
+	if (user_agent)
+		key ^= tfw_hash_str_len(user_agent, 256);
```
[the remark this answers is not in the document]
avbelov23: Changing only the User-Agent will change the hash and will create entries in different baskets. Another thing is that a new attack vector is possible through the clogging of the client base.

tempesta_fw/client.c
```diff
@@ -28,122 +28,222 @@
...
+	if (memcmp_fast(&cli->addr.sin6_addr, &ctx->addr->sin6_addr,
+			sizeof(cli->addr.sin6_addr)))
+	{
+		return false;
+	}
 
-	spin_lock(&hb->lock);
+	if ((!ctx->cli_addr && memcmp_fast(&ent->cli_addr.sin6_addr, &any_addr,
+					   sizeof(ent->cli_addr.sin6_addr))) ||
+	    (ctx->cli_addr && memcmp_fast(&ent->cli_addr.sin6_addr,
```
ikoveshnikov: There are already `if (cli_addr)` and `if (user_agent)` conditions in `tfw_client_obtain()`. And you can init `ctx->cli_addr` as `any_addr` and `ctx->user_agent` as empty string there to make conditions in this function simpler. But it's just a matter of taste.

tempesta_fw/client.c
```diff
@@ -28,122 +28,173 @@
...
+	int users;
+
+	if (!memcmp_fast(&cli->addr.sin6_addr, &addr->sin6_addr,
+			 sizeof(cli->addr.sin6_addr)))
+		return false;
+
+	spin_lock(&ent->lock);
+
+	if (curr_time > ent->expires) {
+		bzero_fast(&cli->class_prvt, sizeof(cli->class_prvt));
```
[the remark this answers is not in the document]
ikoveshnikov: I was thinking about the problem more and it seems it doesn't really exist. Client entry wont be expired untill connection is open and no overwrite will happen.

tempesta_fw/http.h
```diff
@@ -367,6 +368,9 @@ typedef struct {
  * @vhost	- virtual host for the request;
  * @location	- URI location;
  * @sess	- HTTP session descriptor, required for scheduling;
+ * @peer	- end-to-end peer. That the peer is not set if
```
ikoveshnikov: `That` is extra here.

tempesta_fw/client.c
```diff
@@ -28,122 +28,222 @@
...
+ *			  zero
  */
-CliHashBucket cli_hash[CLI_HASH_SZ] = {
-	[0 ... (CLI_HASH_SZ - 1)] = {
-		HLIST_HEAD_INIT,
-	}
-};
+typedef struct {
+	TfwClient	cli;
+	TfwAddr		cli_addr;
```
ikoveshnikov: `cli_addr` can be misleading, since `cli->addr` exists. The field describes `X-Forwarded-For` address, should we name it `xff_addr` instead?

tempesta_fw/client.c
```diff
@@ -28,122 +28,222 @@
...
+#define UA_CMP_LEN	256
 
-typedef struct {
-	struct hlist_head	list;
-	spinlock_t		lock;
-} CliHashBucket;
+static struct {
+	unsigned int	db_size;
+	const char	*db_path;
+	unsigned int	lifetime;
```
avbelov23: Maybe I will also rename `life` to` expires` and `tfw_client_set_lifetime` to` tfw_client_set_expires`?

Issue #498, opened 2016-05-23
HTTP message buffering and streaming


krizhanovsky:
Probably good to be done with #1902 

## General requirements

Tempesta must support two modes of operation: HTTP messages buffering as now and streaming. In current mode of operation all HTTP messages are buffered, i.e. we deliver a proxied request or response only when we fully receive it. As opposite each received `skb` must be forwarded to client or server immediately.

Full buffering and streaming are two edge modes and intermediate mode must be supported as well - partial message buffering based on TCP receive buffer. 

HTTP headers must always be buffered - we need the header to decide what we should do with the message and how to forward/cache/whatever to do with it.

## Configuration

The behavior must be controlled by new configuration options. Since Linux doesn't support per-socket `sysctl`s available for a user, we have to introduce our own memory limits for  server and client sockets separately.

1. `client_mem <soft_limit> <hard_limit>` - controls haw many memory is used to store unanswered client requests and requests with linked responses which can not be forwarded to a client. If `soft_limit` is zero, then streaming mode is used, i.e. each received `skb` is immediately forwarded. Otherwise the message is buffered, but not more than for `soft_bytes` bytes. `hard_limit` is `2 * soft_limit` by default, see description of  in security operations section.

2. `client_msg_buffering N` - controls message buffering. Buffer only first N bytes of requests when forwarding requests to backend servers. Longer messages are forwarded part-by-part and never fully assembled in Tempesta. If request headers are longer than N bytes, then they are still buffered, since full set of headers is required to correctly serve and forward the request. The limit is applied per message and must not overcome per-connecton client_conn_bufferng limit.

3. `server_msg_buffering N` - same as `client_msg_buffering` but for the server connections.

Previous attempts to implement the issue use `client_rmem` very similar to `tcp_rmem` sysctl, however the current `client_mem` is very different because it accounts linked server responses.

## TCP interactions

All ingress `skb`s are immediately evicted and ACKed from the TCP receive queue, so actually we don't use TCP receive buffer for now. With the new enhancement we must account all HTTP data kept in Tempesta memory as residing in the TCP receive buffer of the socket which we received the data on, so TCP will send lower TCP receive windows. See `ss_rcv_space_adjust()` and `tfw_cli_rmem_{reserve,release}()` in #1183.

`client_mem <soft_limit> <hard_limit>` determines TCP receive window, but it also account responses. A simplified example for `client_mem 10 20`:
```
0. initial rwnd=10
[...]

PR #1183, opened 2019-02-12
A new iteration of #498: message streaming mode. NOT done and work in process

vankoven:
The task is not completed yet and the work is still in progress. It's here to make a very high level review to estimate the pros and cons of the used approach.

The code compiles and it even works in _some_ scenarios, but it doesn't look good for now. Last week I've tried to fix issues I had in streaming, especially in scenarios when one (or both) of the streaming peers suddenly disconnects during streaming, but I completely ruined the streaming process and the code became over sophisticated, unreliable and tending to crash. So I tried to revert to very beginning and move forward commit-by-commit trying to simplify the code and make it more straight forward, but I didn't manage to finish it yet and some pieces of code are too messy and unreliable.

Here the issues that appeared during implementation.

TCP window steering
-------------------

TCP receive window steering. There are only two situations, when the window steering is required. In first, client feeds us tons of pipelined requests. Not very big problem and can be resolved via Frang limits. Frang will block too fast clients, normally clients are slower than the servers. Probably closing client connections with [429 status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) is acceptable. The second case is message streaming. We don't want to allow client to eat all the memory by a sing request (e.g. POST request with Blue-Ray image encapsulated).

Note, that the window steering can't slow down clients that doesn't use request pipelining. It sends request - window gets lower, it receives the response - window is maximum agains. 

The task is to set N as buffer size for both unparsed skbs in receive queue and parsed requests in `TfwCliConn->seq_queue`. If requests strored in `seq_queue` are already filled the buffer, then advertize a zero window to the client. If a response is forwarded to the client, then the corresponding request must be destroyed and the advertised window should grow on.

To control window size, we need to force the kernel to take into account the space occupied by the requests in the `seq_queue`. If we look at tcp bufers we will see two buffers: "network" buffer advertised as receive window and "application" buffer used to isolate application from network. Currently skbs are orphaned from the tcp socket right after receive, so the TCP stack sees that there is no allocated memory from the socket and advertises the maximum window size. My idea was to register size of parsed HTTP message in the `sk->sk_rmem_alloc` to make the Kernel aware of the used memory and avoid manual window size calculations. The trick worked. While client was filling the "application" buffer, he was advertised a maximum window size, but while "network" buffer was used the window got smaller, up to zero size. Another great result was possibility to set different buffer limits for client and server connections. 

The code worked in all scenarios. Except one. Closing connections in very concurrent environment. Dropped requests might be deregistered twice from the `sk->sk_rmem_alloc`, so it wasn't empty on socket destruction, so kernel might to generate a pretty long traces on `wrk` stops. The issue has a sporadic character.

Streaming tricky situations
---------------------------

Streamed messages can be buffered.

This happen, when the receiver is not ready to receive a stream. E.g. client requested two pipelined uris: first is very slow to complete, second is fast, but it can be a full BD image. It's not possible to stream the BD image until the first request is responsed. We can't put server connection on hold. Client may intentionally slow down receiving, so the server connection would get stack.  That is why window steering can't be applied to server connections.

Some parts of a streamed messages must be buffered for some time

Some high level http processing can't work with partial data. E.g. trailer headers can't be processed by parts, we have to buffer them fully before process. Although RFC allow to ignore trailer headers in stream mode, we still want to validate the headers and modify them if `req_hdr_set` directive is set.

Response can appear at any time.

Http is request-response protocol. Response is sent only after request was received. Well, forget about that during the streaming. Imagine, that a part of the request was streamed to backend and a new part is being parsed. There can be already a response received from the backend. Scenario 1: try to stream GET request with a long body to Nginx and make a pause, Nginx will respond immediately without waiting for the message end. Scenario 2: stream a part of request and make a syntax error, backend will respond with 4xx code immediately (Tempesta has the same behaviour).

Target connections to stream a message can dissappear at any time.

Imagine a request is streamed, a part of it was already received by the backend. It already sent a response. Now error happens and backend connection is closed. We still need to receive the rest of the stream, but there is no need to send it anywhere since response is received. Just drop it part by part.

Client disconnected when streamed message wasn't received in full

If the partly received request wasn't sent to backend, simply evict it. It it was, then there is no recovery possibilities, only to close server connection. 

(The list is not finished ind to be continued shortly.)


krizhanovsky:
Previous version of the pull request with comments - https://github.com/tempesta-tech/tempesta/pull/1012

tempesta_fw/http_msg.c
```diff
@@ -967,3 +973,57 @@ __tfw_http_msg_alloc(int type, bool full)
 	return hm;
 }
 
+TfwHttpMsg *
+tfw_http_msg_alloc_stream_part(TfwHttpMsg *hm_head)
+{
+	TfwHttpStream *stream;
```
krizhanovsky: I supposed that if we receive a part of streamed HTTP message, then there is current `TfwHttpMsg` which we can attach the new data, use it for the parser and so on. Why do we need `TfwHttpStream` and even allocate the new descriptor?

tempesta_fw/http.h
```diff
@@ -435,6 +486,37 @@ struct tfw_http_resp_t {
...
+ * @rcv_conn		- connection receiving the stream;
+ * @off			- offset from the buffer part end;
+ * @proc_off		- offset of unprocessed data in @body
+ *
+ * It's not always possible to stream the message as soon as it arrives,
+ * receiver may not be ready for this particular message, e.g. if there is one
+ * more streamed and not fully received message in its queue. So the streamed
+ * message is actually get buffered until the receiver is ready. There are also
+ * some cases when a more of the message must be buffered before is can be
+ * processed, e.g. trailer header processing.
```
krizhanovsky: If there are 2 streamed messages, then there are should be 2 incomplete `TfwHttpMsg` descriptors - why not to use them? We even can use parts of `TfwMsg->skb_head` list, possibly with some skb pointers in the list, to handle a message headers and only current stream buffer...

tempesta_fw/sock.c
```diff
@@ -778,6 +779,72 @@ ss_tcp_process_skb(struct sock *sk, struct sk_buff *skb, int *processed)
...
+			/* Make the window clamp follow along.  */
+			tp->window_clamp = rcvwin;
+		}
+	}
+	tp->rcvq_space.space = copied;
+
+new_measure:
+	tp->rcvq_space.seq = tp->copied_seq;
+	tp->rcvq_space.time = tp->tcp_mstamp;
+}
```
krizhanovsky: OK, I love what's going on here - now it seems like the right way to solve the issue!

PR #1193, opened 2019-02-20
Fix race condition in client connection drop

### LINKED ISSUE #1171: Race condition in client connection drop
Thanks to @aleksostapenko the issue was highlighted in the https://github.com/tempesta-tech/tempesta/pull/1169#discussion_r252144274 and applies to this function: https://github.com/tempesta-tech/tempesta/blob/00e82ddac0f06ad1a730a9fc03ea457bd9d0cc4c/tempesta_fw/http.c#L1852-L1860

Sample scenario:
1 thr.: `TFW_HTTP_B_REQ_DROP` set in `tfw_http_conn_cli_drop()`;
2 thr.: `TFW_HTTP_B_REQ_DROP` tested as 1 here (or in `tfw_http_req_evict_dropped()` or in `tfw_http_req_zap_error()`);
2 thr.: `tfw_http_conn_msg_free((TfwHttpMsg *)req)` processed;
1 thr.: the rest of `tfw_http_conn_cli_drop()` is executing (`list_del_init(&req->msg.seq_list)` etc).

ikoveshnikov:
fix #1171

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -2695,13 +2695,6 @@ tfw_http_cli_error_resp_and_log(TfwHttpReq *req, int status, const char *msg,
 			list_add_tail(&req->msg.seq_list, &cli_conn->seq_queue);
 			spin_unlock(&cli_conn->seq_qlock);
 		}
-		else {
-			spin_lock(&cli_conn->seq_qlock);
-			WARN_ONCE(list_empty(&req->msg.seq_list)
-				  && !test_bit(TFW_HTTP_B_REQ_DROP, req->flags),
-				  "Request is not in in seq_queue\n");
-			spin_unlock(&cli_conn->seq_qlock);
-		}
 		/*
 		 * If !on_req_recv_event, then the request @req may be some
 		 * random request from the seq_queue, not the last one.
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -2021,15 +2021,18 @@ tfw_http_conn_cli_drop(TfwCliConn *cli_conn)
 	 */
 	spin_lock(&cli_conn->seq_qlock);
 	list_for_each_entry_safe(req, tmp, seq_queue, msg.seq_list) {
-		set_bit(TFW_HTTP_B_REQ_DROP, req->flags);
-		list_del_init(&req->msg.seq_list);
 		/*
-		 * Response is processed and removed from fwd_queue, need to be
-		 * destroyed.
+		 * Request must be destroyed if the response is fully processed
+		 * and removed from fwd_queue. If the request is still in use
+		 * immediately after REQ_DROP flag is set, the request-response
+		 * pair can be destroyed in other thread.
 		 */
-		if (req->resp
-		    && test_bit(TFW_HTTP_B_RESP_READY, req->resp->flags))
-		{
+		bool unused = req->resp
+			&& test_bit(TFW_HTTP_B_RESP_READY, req->resp->flags);
+		list_del_init(&req->msg.seq_list);
+		smp_mb__before_atomic();
+		set_bit(TFW_HTTP_B_REQ_DROP, req->flags);
+		if (unused) {
 			tfw_http_resp_pair_free(req);
 			TFW_INC_STAT_BH(serv.msgs_otherr);
 		}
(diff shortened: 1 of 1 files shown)


tempesta_fw/http.c
```diff
@@ -2021,15 +2021,16 @@ tfw_http_conn_cli_drop(TfwCliConn *cli_conn)
...
+		 * Request must be destroyed if the response is fully processed
+		 * and removed from fwd_queue.
 		 */
-		if (req->resp
-		    && test_bit(TFW_HTTP_B_RESP_READY, req->resp->flags))
-		{
+		bool unused = req->resp
+			&& test_bit(TFW_HTTP_B_RESP_READY, req->resp->flags);
+		set_bit(TFW_HTTP_B_REQ_DROP, req->flags);
+		smp_mb__after_atomic();
```
krizhanovsky: Why the request can not be freed right here and on the next line `req->msg` dereference lead to crash?

aleksostapenko: Agree with @krizhanovsky: race is still possible here. In my opinion, the simple solution (but maybe not the best) is to lock `seq_qlock` on the other side (`tfw_http_resp_cache_cb()` and others) - during `TFW_HTTP_B_REQ_DROP` checking and request removal.
Also, in this context - it seems that we have the similar problem [here](https://github.com/tempesta-tech/tempesta/blob/8e828719bd03c38aa4d2f3b475753bcdf6429302/tempesta_fw/http.c#L1969): verifying `list_empty_careful(&req->msg.seq_list)` before the lock for the queue is taken.

aleksostapenko: Agree with @krizhanovsky: race is still possible here. In my opinion, the simple solution (but maybe not the best) is to lock `seq_qlock` on the other side (`tfw_http_resp_cache_cb()` and others) - during `TFW_HTTP_B_REQ_DROP` checking and request removal.
Also, in this context - it seems that we have the similar problem [here](https://github.com/tempesta-tech/tempesta/blob/8e828719bd03c38aa4d2f3b475753bcdf6429302/tempesta_fw/http.c#L1969): verifying `list_empty_careful(&req->msg.seq_list)` before the lock for the queue is taken.

aleksostapenko: Agree with @krizhanovsky: race is still possible here. In my opinion, the simple solution (but maybe not the best) is to lock `seq_qlock` on the other side (`tfw_http_resp_cache_cb()` and others) - during `TFW_HTTP_B_REQ_DROP` checking and request removal.
Also, in this context - it seems that we have the similar problem [here](https://github.com/tempesta-tech/tempesta/blob/8e828719bd03c38aa4d2f3b475753bcdf6429302/tempesta_fw/http.c#L1969): verifying `list_empty_careful(&req->msg.seq_list)` before the lock for the queue is taken.

ikoveshnikov: There is no need to lock seq_queue when `TFW_HTTP_B_REQ_DROP` is checked all over the  the code. 
If `TFW_HTTP_B_REQ_DROP` flag is set, the request is already removed from the seq_queue and can be accessed only by `resp->req` link. The `RESP_READY` flag for response is set under seq_queue lock, so it's guaranteed that the same request is not accessed by both `tfw_http_conn_cli_drop()` and `tfw_http_resp_fwd()`.

In the same time if the response is not ready, the request is not destroyed in `tfw_http_conn_cli_drop()`. When other thread tests the `DROP` flag, it destroys the request. It's safe since it's just performance optimisation. Without the check the response-request pair will be passed into `tfw_http_resp_fwd()` and will be destroyed there.

The only race I have found was in `resp_and_log()` and I've fixed it.

krizhanovsky: OK, makes sense. If there is logical dependency between `req->msg.seq_list` (a request is in the list) and the flag, then don't we need `smp_mb__before_atomic()` between the list deletion and setting the flag? BTW why do we need the `smp_mb__after_atomic()`?

aleksostapenko: Yeah, this solution is better then additional locks.
Now the sequence
```
set_bit(TFW_HTTP_B_REQ_DROP, req->flags);
```
should not be reordered to avoid races, thus `smp_mb__before_atomic()` is desirable between them.

tempesta_fw/http.c
```diff
@@ -2693,7 +2694,8 @@ tfw_http_cli_error_resp_and_log(TfwHttpReq *req, int status, const char *msg,
 			spin_unlock(&cli_conn->seq_qlock);
 		}
 		else {
-			WARN_ONCE(list_empty_careful(&req->msg.seq_list),
+			WARN_ONCE(list_empty_careful(&req->msg.seq_list)
+				  && !test_bit(TFW_HTTP_B_REQ_DROP, req->flags),
 				  "Request is not in in seq_queue\n");
```
aleksostapenko: Looks like this check can has false positives in any way: it can be called during the other thread is in `tfw_http_conn_cli_drop()` -  just between the sequence mentioned in [comment above](https://github.com/tempesta-tech/tempesta/pull/1193/files#r260360504).

tempesta_fw/http.c
ikoveshnikov: Fixed memory barriers.

tempesta_fw/http.c
ikoveshnikov: Yes, the only way to avoid false positives here is locking the seq_queue. Lock was added.

Issue #1187, opened 2019-02-13
TLS crash during establishing of the secure connection


ikoveshnikov:
Just caught the bug on current master of Tempesta 5adf46dd9b94c9fa67bdd0233e3b836bc4931ad9 and the kernel tempesta-tech/linux-4.14.32-tfw@e90410720527a639ad6f50f0193e1ff00e7f0076.

First issue happen during build, same issue reproduced on CI:
```
...
  CC [M]  /home/user/tempesta/tempesta/tempesta_fw/t/unit/test_wq.o
  CC [M]  /home/user/tempesta/tempesta/tempesta_fw/t/unit/test_sched_hash.o
  CC [M]  /home/user/tempesta/tempesta/tempesta_fw/t/unit/test_tfw_str.o
  CC [M]  /home/user/tempesta/tempesta/tempesta_fw/t/unit/test_mem_fast.o
  CC [M]  /home/user/tempesta/tempesta/tempesta_fw/t/unit/test_hash.o
  CC [M]  /home/user/tempesta/tempesta/tempesta_fw/t/unit/test_sched_ratio.o
  LD [M]  /home/user/tempesta/tempesta/tempesta_fw/t/unit/tfw_test.o
  Building modules, stage 2.
  MODPOST 7 modules
WARNING: modpost: Found 1 section mismatch(es).
To see full details build your kernel with:
'make CONFIG_DEBUG_SECTION_MISMATCH=y'
  CC      /home/user/tempesta/tempesta/lib/tempesta_lib.mod.o
  CC      /home/user/tempesta/tempesta/tempesta_db/core/tempesta_db.mod.o
  CC      /home/user/tempesta/tempesta/tempesta_fw/t/tfw_bomber.mod.o
  CC      /home/user/tempesta/tempesta/tempesta_fw/t/tfw_fuzzer.mod.o
...
```

Then I've started the Tempesta with configuration as follows:
```
server 127.0.0.1:8080;

listen 443  proto=https;

tls_certificate /home/user/tls/tfw-root.crt;
tls_certificate_key /home/user/tls/tfw-root.key;
```

And used curl to access the site (uri `/hello` is text file length of 20 bytes):
```
% curl -v https://192.168.122.12/hello                                       :(
* Expire in 0 ms for 6 (transfer 0x55b06a492e80)
*   Trying 192.168.122.12...
* TCP_NODELAY set
* Expire in 200 ms for 4 (transfer 0x55b06a492e80)
[...]

PR #1197, opened 2019-02-25
tls: stop changing io->rlen at buffer end

i-rinat:
`io->rlen` is used to track how many bytes were processed by a particular FSM state. It should not be touched when we exit the state machine in `TTLS_HS_FSM_FINISH()`.

Having `io->rlen` increased in `TTLS_HS_FSM_FINISH()` causes #1187. Here is why.

Let's assume we are entering `T_FSM_STATE(TTLS_CH_HS_SESS)` and there are some unprocessed data bytes left. Say, it's `10`. We copy ten bytes of session ID into the `tls->sess.id`, then increase `io->rlen` by `10`, making it `10`. As `10` bytes are not the full session ID, we call `T_FSM_EXIT()`, which jumps out of state machine, causing `io->rlen` to be increased by size of data processed so far, which may be about 50 bytes. `io->rlen` becomes `60`. Next time we enter state machine, statement `BUG_ON(io->rlen >= tls->sess.id_len);` triggers.

(related: #1187)

### DIFF
--- a/tls/tls_internal.h
+++ b/tls/tls_internal.h
@@ -360,13 +360,14 @@ enum {
 #define TTLS_HS_FSM_FINISH()						\
 	T_FSM_FINISH(r, tls->state);					\
 	*read += p - buf;						\
-	io->rlen += p - buf;
+	io->rlen += p - state_p;
 
 /* Move to @st if we have some bytes to process. */
 #define TTLS_HS_FSM_MOVE(st)						\
 do {									\
 	WARN_ON_ONCE(p - buf > len);					\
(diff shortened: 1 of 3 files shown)


i-rinat:
> It's better than the current fix because of smaller FSM code which is easer to work with and the code becomes faster.

I've just removed explicit `io->rlen` updates. But surprisingly that made code 16 bytes larger. I did not investigate reasons in depth, but can speculate that added `state_p = p` assignment now appearing in every state transition macro instantiation may have introduced that difference.

Issue #1128, opened 2018-12-07
[Parser] Incorrect move in 'End of Term` state right after TRY_STR macro was used


ikoveshnikov:
All over the code TRY_STR macro matches only prefix of the term and moves parser to End of Term state. If next character is allowed token symbol, token delimiter is ignored.

It's more easy to show the error on example, let it be [__req_parse_accept()](https://github.com/tempesta-tech/tempesta/blob/cac94130f4a47060221dc34e24c6d42750b21db4/tempesta_fw/http_parser.c#L1936).

At some stage we match current token `text/htmlbla-bla-bla` with known token pattern. It matches, we set the flag and move to EndOfTerm (`I_EoT`) state:
https://github.com/tempesta-tech/tempesta/blob/cac94130f4a47060221dc34e24c6d42750b21db4/tempesta_fw/http_parser.c#L1935-L1938

At `I_EoT` state the next character is checked, it's token character, so we move to `Req_I_Accept` state back again. Now we try to match `bla-bla-bla` with known patterns and move to state `Req_I_AcceptOther` and continue parsing.

The original expression `text/htmlbla-bla-bla` shouldn't be parsed as known pattern at all. But it was parsed as combination of two patterns `text/html` and `bla-bla-bla`, which is not valid parsing.

The same error happens whenever `TRY_STR` macro is used. E.g. the patch as follows _must_ fail the unit test `TEST(http_parser, fills_hdr_tbl_for_req)`, but it passes.
```diff
--- a/tempesta_fw/t/unit/test_http_parser.c
+++ b/tempesta_fw/t/unit/test_http_parser.c
@@ -771,7 +771,7 @@ TEST(http_parser, fills_hdr_tbl_for_req)
        const char *s_xch = "X-Custom-Hdr: custom header values";
        const char *s_dummy9 = "Dummy9: 9";
        const char *s_dummy4 = "Dummy4: 4";
-       const char *s_cc  = "Cache-Control: max-age=1, no-store, min-fresh=30";
+       const char *s_cc  = "Cache-Control: max-age=1, no-storer, min-fresh=30";
        const char *s_te  = "compress, gzip, chunked";
        /* Trailing spaces are stored within header strings. */
        const char *s_pragma =  "Pragma: no-cache, fooo ";
@@ -796,7 +796,7 @@ TEST(http_parser, fills_hdr_tbl_for_req)
                "Dummy7: 7\r\n"
[...]

PR #1201, opened 2019-03-04
Fix #1131: use GCC jumps to labels stored as pointers.

### LINKED ISSUE #1131: HTTP parser: replace switch by direct jumps by lables address
The main our FSM for HTTP parser as well as generic `lib/fsm.h` and frang state machines still use `switch` statements for very initial entry to a FSM - all consequent jumps are performed directly by jumps. The HTTP FSM is pretty large, so even having compiler optimizations for large `switch` such as binary search and hashing, the lookup can be costly and there is no sense to spend CPU on them.

[GCC allows](https://gcc.gnu.org/onlinedocs/gcc/Labels-as-Values.html) to directly get addresses of labels, so we can use them in `goto` statements with a variable containing a label address. Also this reading can be good and useful [JIT Compilation Without (Much) Assembler in GNU C](http://dginasa.blogspot.com/2012/05/jit-compilation-without-much-assembler.html).

All the FSMs must be reworked to use the direct immediate jumps. With the change we also can remove doubling of `case` labels and labels used for `goto` statements leaving only the last ones.

krizhanovsky:
The FSM becomes complex, so lib/fsm.h left as is (all the instances of the FSM definition don't use so many labels after all).

Fix objtool warnings about sibling call in cfg.c (also due to usage of the same GCC extension).

Remove HTTP normalization zero logic - #2 defines the logic in different way.

Fix the tests mess with not-passed DEBUG define and redifinition hell in test_http_sticky.c.

Move Tempesta FW version to 0.7.

I tested the patch on my 4 core VM with Apache HTTP as a backend serving 3 byte index.html and wrk running from the host as:
```
./wrk -t 4 -c 128 -d 60s --header 'Connection: keep-alive' --header 'Upgrade-Insecure-Requests: 1' --header 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36' --header 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8' --header 'Accept-Encoding: gzip, deflate, sdch' --header 'Accept-Language: en-US,en;q=0.8,ru;q=0.6' --header 'Cookie: a=sdfasd; sdf=3242u389erfhhs; djcnjhe=sdfsdafsdjfb324te1267dd; sdaf=mo2u8943478t67437461746rfdgfcdc; ityu=9u489573484duifhd; GTYFT=nsdjhcbyq3te76ewgfcZ; uityut=23Y746756247856425784657; GA=URHUFVHHVSDNFDHGYSDGF; a=%45345%dfdfg%4656%4534sdfjhsdb.sdfsg.sdfgsf' http://192.168.100.4:80/
```
Tempesta config:
```
listen 192.168.100.4:80;
srv_group default {
	server 127.0.0.1:8080; # apache
}
vhost default {
	proxy_pass default;
}
cache 1;
cache_fulfill * *;
http_chain {
	-> default;
}
```

Before the patch:
```
     7.64%  [tempesta_fw]  [k] tfw_http_parse_req
     2.79%  [e1000]        [k] e1000_xmit_frame
     2.32%  [tempesta_fw]  [k] __tfw_strspn_simd
     2.31%  [tempesta_fw]  [k] __tfw_http_msg_add_str_data
     1.60%  [tempesta_fw]  [k] __new_pgfrag
     1.58%  [kernel]       [k] skb_release_data
     1.55%  [tempesta_fw]  [k] __str_grow_tree
     1.41%  [kernel]       [k] __inet_lookup_established
     1.35%  [tempesta_fw]  [k] tfw_cache_do_action
     1.35%  [tempesta_fw]  [k] __tfw_strcmpspn
```

After the patch:
```
     5.78%  [tempesta_fw]  [k] tfw_http_parse_req
     2.90%  [e1000]        [k] e1000_xmit_frame
     2.29%  [tempesta_fw]  [k] __tfw_http_msg_add_str_data
     2.14%  [tempesta_fw]  [k] __tfw_strspn_simd
     2.06%  [kernel]       [k] swiotlb_tbl_unmap_single
     1.53%  [kernel]       [k] skb_release_data
     1.43%  [tempesta_fw]  [k] tfw_cache_do_action
     1.41%  [tempesta_fw]  [k] __str_grow_tree
     1.40%  [kernel]       [k] __inet_lookup_established
     1.29%  [kernel]       [k] queued_spin_lock_slowpath
```


tempesta_fw/http_parser.c
```diff
@@ -3825,7 +3239,7 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len,
 				   && TFW_LC(*(p + 6)) == 't'
 				   && *(p + 13) == ':'))
 			{
-				parser->_i_st = Req_HdrAcceptV;
+				parser->_i_st = &&Req_HdrAcceptV;
```
ikoveshnikov: The half of the `parse_req|resp` functions is parsing known headers char-by-char or bulk parsing like on these lines. Can we move that parsing to separate function like `parser_get_header()` which could collect the header.
PROs: 
- the whole function will be much shorter
- Less jumps in one function. Not really important since the patch implements direct jumps.
- The most important: the function can be generated ac compile time, from very simple description: header name and target label.
It's a general proposal, not for this patchset.

tempesta_fw/t/unit/test_http_sticky.c
```diff
@@ -35,9 +35,13 @@
 #define __init
 #endif
 
-#if DBG_SS == 0 || DBG_TLS == 0
+/*
+ * TODO #74: the test is complete mess - it includes half of Tempesta spurce
+ * code and I gave up to fix multiple definition conflicts for debug mode.
+ */
 #undef DEBUG
```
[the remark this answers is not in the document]
krizhanovsky: I bothered with the DEBUG macro exactly to be able to debug failed unit tests, so the switch is useful for the unit tests. Sticky cookie unit test is a big headache to be replaced with a functional tests - this is the only reason why I just undefined DEBUG for it.

tempesta_fw/http.h
```diff
@@ -229,6 +229,8 @@ enum {
 	TFW_HTTP_B_CONN_EXTRA,
 	/* Chunked transfer encoding. */
 	TFW_HTTP_B_CHUNKED,
+	/* The message body is limited by the connection closing. */
+	TFW_HTTP_B_UNLIMITED,
```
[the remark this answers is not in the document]
krizhanovsky: Yeah, thank you for noticing! With the parser changes we can not check parser state, so in `tfw_http_parse_terminate()` we can not check `Resp_BodyUnlim*` states and the flag was introduced to treat HTTP responses without neither chunked encoding nor Content-Length. Fixed in `tfw_http_parse_terminate()`.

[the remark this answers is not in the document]
krizhanovsky: BTW I didn't find an appropriate test, so created https://github.com/tempesta-tech/tempesta-test/issues/94 - could you please check that we actually don't have the test?

tempesta_fw/http_parser.c
krizhanovsky: Yes, definitely. In #309 we at least need to split the parser functions into headers and body parts. I saw that even with `barrier()` (the improvement is coming soon in the branch) GCC generates poor code layout, so smaller parsing functions are better.

The only confusing thing is how to call them? It seems we need additional branching for this, like
```
if (state is header parsing)
    parse_headers();
else if (state is body parsing)
    parse_body();
else
    some_error();
```

We already do the FSM split by our internal (`i`) states, e.g. to process certain header values.

How we can efficiently move the curacter-by-caracter logic out of the main logic?

tempesta_fw/http_parser.c
```diff
@@ -4717,21 +4108,90 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
+	int r = CSTR_NEQ;
+	TfwHttpResp *resp = (TfwHttpResp *)msg;
 	TfwHttpParser *parser = &msg->conn->parser;
 
-	ret = __parse_http_date(msg, data, len);
-	if (ret < CSTR_POSTPONE) {  /* (ret < 0) && (ret != POSTPONE) */
+	if (!test_bit(TFW_HTTP_B_HDR_LMODIFIED, resp->flags))
+		r = __parse_http_date(msg, data, len);
+
+	if (r < CSTR_POSTPONE) {  /* (ret < 0) && (ret != POSTPONE) */
```
i-rinat: By the way, GCC is smart enough to optimize `ret < 0 && ret != -1` to `ret < -1`.

tempesta_fw/t/unit/test_http_sticky.c
```diff
@@ -35,9 +35,13 @@
 #define __init
 #endif
 
-#if DBG_SS == 0 || DBG_TLS == 0
+/*
+ * TODO #74: the test is complete mess - it includes half of Tempesta spurce
```
i-rinat:
```suggestion
 * TODO #74: the test is complete mess - it includes half of Tempesta source
```

tempesta_fw/http_parser.c
```diff
@@ -381,6 +383,9 @@ __FSM_STATE(st) {							\
 	__FSM_MOVE_nofixup(Req_MUSpace);				\
 }
 
+/* 4-byte (Integer) access to a string Pointer. */
+#define PI(p)	(*(unsigned int *)(p))
```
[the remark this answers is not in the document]
krizhanovsky: It's also not to nice to see many type conversions `*(unsigned int *)`. Maybe it'd have sense to rename the macro if we find it not clear enough.

tempesta_fw/http_parser.c
```diff
@@ -112,32 +98,44 @@ do {									\
...
 } while (0)
 
-#define __FSM_START(s)							\
+#define __FSM_START(s, st0)						\
 fsm_reenter: __attribute__((unused))					\
-	TFW_DBG3("enter FSM at state %d\n", s);				\
-switch (s)
+	T_DBG3("enter FSM at state %pK\n", s);				\
+if (!s)									\
+	s = &&st0;							\
```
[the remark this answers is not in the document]
krizhanovsky: Yes and thank you for the good point! Actually we need the auxiliary state variable in the parser **only** when we exit the FSM with `POSTPONE`, i.e. we're going to enter the FSM from the same, not initial, state with new data. We save the state in MOVE statements if we're going to exit to wait for new data and this was contradictional the the initial attempt to remove the auxiliary state handling.

Fixed this and also have a look on changes replacing `__FSM_I_MOVE_n(..., 0)` with `__FSM_I_JMP()` and the last one also can be replaced by just `/* Fall through. */` if the state is the next one - we don't need to save the parser state. @ikoveshnikov @aleksostapenko @avb please also keep this in mind.

tempesta_fw/http_parser.c
```diff
@@ -259,16 +257,20 @@ do {									\
...
 #define __FSM_I_MATCH_MOVE_finish(alphabet, to, finish)			\
 do {									\
 	__fsm_n = __data_remain(p);					\
 	__fsm_sz = tfw_match_##alphabet(p, __fsm_n);			\
 	if (unlikely(__fsm_sz == __fsm_n)) {				\
 		__msg_hdr_chunk_fixup(data, len);			\
-		parser->_i_st = to;					\
-		__fsm_const_state = to;					\
+		parser->_i_st = &&to;					\
+		__fsm_const_state = &&to;				\
```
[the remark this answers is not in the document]
krizhanovsky: It seems we don't need the variable at all - we can just write `parser->state` if we're about to exit from the parsing function.

tempesta_fw/http_parser.c
```diff
@@ -18,47 +18,33 @@
  * this program; if not, write to the Free Software Foundation, Inc., 59
  * Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  */
+#pragma GCC optimize("O3", "unroll-loops", "inline", "no-strict-aliasing")
+#pragma GCC target("mmx", "sse4.2", "avx2")
```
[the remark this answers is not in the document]
krizhanovsky: Done

tempesta_fw/http_parser.c
```diff
@@ -3433,115 +2928,113 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len,
 		/* fall through */
 	}
 
+/*
+ * Fot most (or at least most frequent) methods @step_inc should be
+ * optimized out. The macro is used to reduce the FSM size, so there is not
+ * sense to use it specific versions for the few states, e.g. for 'GET '.
+ */
+#define __MATCH_METH(meth, step_inc)					\
```
avbelov23: I think you should not save two letters and use `METHOD`

tempesta_fw/http_parser.c
```diff
@@ -3433,115 +2928,113 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len,
...
-			switch (*(unsigned int *)p) {
-			/* Most expected methods: GET, HEAD, POST. */
-			case TFW_CHAR4_INT('G', 'E', 'T', ' '):
+		if (likely(__data_available(p, 9))) {
+			/*
+			 * Move most frequent methods forward and do not use
+			 * switch to make compiler not to merge it with the
+			 * switch at the below.
+			 *
+			 * Usually we have enough data (smalest HTTP/1.1 reqeust
```
avbelov23: smallest and request

tempesta_fw/http_parser.c
```diff
@@ -5107,58 +4641,54 @@ tfw_http_parse_resp(void *resp_data, unsigned char *data, size_t len,
 	}
 
 	/* 'Age:*OWS' is read, process field-value. */
-	TFW_HTTP_PARSE_RAWHDR_VAL(Resp_HdrAgeV, Resp_I_Age, resp,
-				  __resp_parse_age);
+	TFW_HTTP_PARSE_RAWHDR_VAL(Resp_HdrAgeV, resp, __resp_parse_age);
 
 	/* 'Cache-Control:*OWS' is read, process field-value. */
-	TFW_HTTP_PARSE_RAWHDR_VAL(Resp_HdrCache_ControlV, Resp_I_CC, resp,
+	TFW_HTTP_PARSE_RAWHDR_VAL(Resp_HdrCache_CtrlV, resp,
```
i-rinat: All other state names look like the header names themselves. `Resp_HdrCache_CtrlV` is now different.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -755,6 +755,140 @@ TEST(http_parser, alphabets)
...
 }
 
+/**
+ * Test for case (in)sensitive matching of letters and special characters.
+ */
+TEST(http_parser, casesense)
+{
+	FOR_REQ("GET / HTTP/1.1\r\n"
+		"hOST: test\r\n"
+		/* We don't match open and closing quotes. */
```
i-rinat: This is a copied comment, but unlike the place it was copied from, the line below has no `"` inside. I'd propose to delete the comment.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -755,6 +755,140 @@ TEST(http_parser, alphabets)
...
+						      | TFW_HTTP_CC_NO_STORE));
+	}
+
+	EXPECT_BLOCK_RESP("GET / HTTP/1.1\r\n"
+			  "Date: Jana, 23 May 2005 22:38:34 GMT\r\n"
+			  "Content-Length: 0\r\n"
+			  "\r\n");
+
+	/*
+	 * If we have Chunked-Encoding, then we must have 'chunked' among
```
i-rinat: "Transfer-Encoding"?

avbelov23:
Looks good to me

tempesta_fw/http_parser.c
krizhanovsky: We have 80-characters line limits, so there is no sense to use long names. Any way you understood that `METH` is `METHOD` ;)

keshonok: Perhaps he meant that `METH` is a very popular shortcut for a certain substance. `MTOD` could be another clear shortcut for this without the undesireable connotation.

tempesta_fw/http_parser.c
```diff
@@ -1747,25 +1729,32 @@ __parse_transfer_encoding(TfwHttpMsg *hm, unsigned char *data, size_t len)
 	__FSM_STATE(I_TransEncod) {
 		if (TFW_CONN_TYPE(hm->conn) & Conn_Srv) {
 			unsigned int status = ((TfwHttpResp *)hm)->status;
-			if ((status - 100U < 100U) || (status == 204))
+			if (status - 100U < 100U || status == 204)
```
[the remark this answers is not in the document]
krizhanovsky: Oh, didn't know. But I think there is no sense to make a change to remove the optimization.

tempesta_fw/http_parser.c
```diff
@@ -3714,20 +3211,19 @@ tfw_http_parse_req(void *req_data, unsigned char *data, size_t len,
...
 		if (__fsm_n == CSTR_POSTPONE) {
 			p += __fsm_sz;
+			parser->state = &&Req_UriMark;
 			__FSM_EXIT(TFW_POSTPONE);
 		}
 		if (__fsm_n < 0)
 			TFW_PARSER_BLOCK(Req_UriMark);
 
-		parser->_i_st = I_0;
+		parser->_i_st = I_ST(I_0);
```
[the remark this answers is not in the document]
krizhanovsky: We can use the whole body of `__FSM_START_ALT()` in `__parse_http_date()`, but it'd be ugly. `I_ST(I_0)` is also used in `tfw_http_parse_req()`. I thought about the version of the code, but in case of
```C
 if (parser->_i_st == I_ST(I_EoL))
    goto I_EoL;
__FSM_START(parser->_i_st);
```
we won't be able to rping debug message about current state.

tempesta_fw/http_parser.c
krizhanovsky: Yes, but it's to long and we don't use the state names for atimatic code generation.

tempesta_fw/http_parser.c
```diff
@@ -1791,512 +1780,46 @@ __parse_transfer_encoding(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+		__FSM_I_JMP(Req_I_AcceptOther);
+	}
+
+	__FSM_STATE(Req_I_AcceptHtml) {
+		if (IS_WS(c) || c == ',' || c == ';' || IS_CRLF(c)) {
+			__set_bit(TFW_HTTP_B_ACCEPT_HTML, req->flags);
+			__FSM_I_JMP(I_EoT);
+		}
+		__FSM_I_MOVE(Req_I_AcceptOther);
 	}
```
aleksostapenko: It seems that the problem of #1128 is still here: we match `text/htmlbla-bla-bla` in `Req_I_Accept`, pass to `Req_I_AcceptHtml`, then go to the end of block and after `__FSM_I_MOVE(Req_I_AcceptOther)` go to the `Req_I_AcceptOther`, where we try to match `la-bla-bla`.

tempesta_fw/http_parser.c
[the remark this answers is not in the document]
i-rinat: > there is no sense to make a change to remove the optimization

Yeah.

tempesta_fw/http_parser.c
i-rinat: ok

tempesta_fw/http_parser.c
krizhanovsky: Why's that? We have test case
```C
        FOR_REQ("GET / HTTP/1.1\r\n"
                "Accept: text/htmlK, foo\r\n"                                          
                "Connection: keep-aliveA\r\n"
                "Cache-Control: no-transform\", only-if-cachedd\r\n"                   
                "\r\n");                                                               
        {       
                EXPECT_FALSE(test_bit(TFW_HTTP_B_ACCEPT_HTML, req->flags));            
                EXPECT_FALSE(test_bit(TFW_HTTP_B_CONN_KA, req->flags));
                EXPECT_FALSE(req->cache_ctl.flags & (TFW_HTTP_CC_NO_TRANSFORM          
                                                     | TFW_HTTP_CC_OIFCACHED
                                                     | TFW_HTTP_CC_NO_STORE));         
        
        }
```
What's the problem with falling to `Req_I_AcceptOther` if we do not set `TFW_HTTP_B_ACCEPT_HTML`?

tempesta_fw/http_parser.c
```diff
@@ -1791,512 +1769,46 @@ __parse_transfer_encoding(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
-	Resp_BodyUnlimRead,
-
-	Resp_StatesNum
-};
-
-/*
- * Helping (interior) FSM states
- * for processing specific parts of an HTTP request.
+/**
+ * Accept header parser, RFC 7321 5.3.2.
```
aleksostapenko: `7321` -> `7231`

tempesta_fw/http_parser.c
aleksostapenko: It seems that I somewhat incorrect understand the issue: in fact it turns out, that we want to pass such messages, but the appropriate flags just won't be set for them.

PR #1206, opened 2019-03-07
fix encoding to chunked: chunk size is hex digit, not decimal

ikoveshnikov:


tempesta_fw/str.c
```diff
@@ -79,24 +79,35 @@ __dig_num(unsigned long a)
...
-	if (unlikely(n + 1 > len))
+	if (unlikely((n + 1 > len) || (base != 10 && base != 16)))
 		return 0;
 
 	do {
 		/* Compiled to only one MUL instruction with -O2. */
-		*p-- = "0123456789"[ai % 10];
-		ai /= 10;
+		*p-- = "0123456789abcdef"[ai % base];
+		ai /= base;
```
i-rinat: This makes compiler to use `div` instruction, which may slow down the code. (Division by a constant is converted to multiplication since that's faster to compute).

I think it's better to make a separate version of the function for 16-base case.

ikoveshnikov: Fixed

krizhanovsky:
Please backport the fix to release-0.6

PR #1216, opened 2019-03-20
always expand debug macros into expressions

i-rinat:
**Upd.** I was wrong, see comments below for details.

In release builds, debug macros like `T_DBG()` are expanded to nothing. This allows to remove debug code from resulting build completely, but makes code counter-intuitive and bug-prone. Such macros are expected to behave like statements, but they are not. For example, code:
```c
int
ttls_md_update(TlsMdCtx *ctx, const unsigned char *input, size_t ilen)
{
        int r;

        BUG_ON(!ctx || !ctx->md_info);

        r = crypto_shash_update(&ctx->md_ctx, input, ilen);
        if (r)
                T_DBG("cannot update hash ctx, %d\n", r);

        return r;
}
```
expands to
```c
int
ttls_md_update(TlsMdCtx *ctx, const unsigned char *input, size_t ilen)
{
        int r;

        BUG_ON(!ctx || !ctx->md_info);

        r = crypto_shash_update(&ctx->md_ctx, input, ilen);
        if (r)
                

        return r;
}
```
Note that `return r` is conditional, and is called only when `r` is not `0`. Function returning value now doesn't return anything. It just happens to work as expected since `rax` register is filled with correct return code.

Rather than fixing all places where such macros are used, it's easier to change macros themselves to expand to empty expressions. This also prevents similar issues in the future.

### DIFF
--- a/ktest/linux/threads.h
+++ b/ktest/linux/threads.h
@@ -28,8 +28,8 @@
  * In user space all threads have different identifiers,
  * so there is no problems with preemption.
  */
-#define local_bh_disable()
-#define local_bh_enable()
+#define local_bh_disable()	do { } while (0)
+#define local_bh_enable()	do { } while (0)
 
 static size_t __thr_max = 0;
 static size_t __thread __thr_id;
--- a/lib/log.h
+++ b/lib/log.h
@@ -67,13 +67,13 @@ enum {
 #if defined(DEBUG) && (DEBUG >= 1)
 #define T_DBG(...) 	__T_DBG1(__VA_ARGS__)
 #else
-#define T_DBG(...)
+#define T_DBG(...)	do { } while (0)
 #endif
 
 #if defined(DEBUG) && (DEBUG >= 2)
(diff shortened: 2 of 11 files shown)


ikoveshnikov:
> I believe the statement with an empty macro expands to `;`

Agreed. `-E` gcc option gives the preprocessored version:
```C
. . .

# 5 "t.c"
int main(int argc, char *argv[])
{
 printf("argc = %d\n", argc);
 if (argc > 1)
  ;
 printf("not under if\n");
 return 0;
}
```

More over, if the `return` statement is moved into `if` clause, then gcc with all the compilation flags we use will throw a compilation error:
```sh
error: control reaches end of non-void function [-Werror=return-type]
```

So it seems that the patch doesn't really change anything. @i-rinat Maybe there was some different type of error? Where exactly the issue had happen?

i-rinat:
I was wrong, there are no issues with macros. There probably can be some excuses to introduce changes from this PR, but they are not required.

Issue #1181, opened 2019-02-08
Test coverage & fuzzing


krizhanovsky:
# Functional tests coverage

Need to deploy a [gcov](https://01.org/linuxgraphics/gfx-docs/drm/dev-tools/gcov.html) in https://github.com/tempesta-tech/tempesta-test . A test coverage report must be generated (preferably for each CI build or at least by a separate test run). `gcov(1)` show percentage coverage, so the output can be used directly in the report. Maybe it makes sense to use `lcov(1)` to get HTML output for all the files at once.

The task is crucial since high (as close to 100% as possible) statement coverage is the minimal requirement for GA (high branch and predicate coverages are also wished).

# Fuzzing

Probably Kcov with AFL should be used for fuzzing, see [A gentle introduction to Linux Kernel fuzzing](https://blog.cloudflare.com/a-gentle-introduction-to-linux-kernel-fuzzing/).

This must be done at least for TLS handshakes and the HTTP parser - the most security sensitive pieces of code. Different functions should be called by the fuzzer to reveal issues like #1683 , i.e. there should be unit tests using fuzzing.

Open source tools to consider:

- https://github.com/OpenRCE/sulley
- https://github.com/google/AFL
- https://llvm.org/docs/LibFuzzer.html
- https://github.com/google/fuzztest

# References
- [`linux/kernel/gcov`](https://github.com/tempesta-tech/linux-4.14.32-tfw/tree/master/kernel/gcov). 
- [Writing Better Function Tests with GCOV](https://s3.amazonaws.com/connect.linaro.org/yvr18/presentations/yvr18-224.pdf)

krizhanovsky:
With this test design decide whether we need TfwBomber and close https://github.com/tempesta-tech/tempesta/issues/375 if there is no application for our own benchmarking tool (it seems it makes sense for fuzzing only if we don't find a fast enough infrastructure to run the fuzzer).

krizhanovsky:
We need to know our test coverage to release Beta: 80% is absolute minimum, 95% is the desired value.

krizhanovsky:
Let's implement test coverage for now and reschedule fuzzing for later milestone

krizhanovsky:
While Augment [didn't show](https://github.com/tempesta-tech/tempesta/issues/1528#issuecomment-4229801204) any good performance, Anthropic Mythos [looks promising](https://red.anthropic.com/2026/mythos-preview/)

RomanBelozerov:
[...]

PR #1219, opened 2019-03-23
#1181: GCOV compilation option

krizhanovsky:
The option is useful for profile guided optimization (PGO). I tried to build
http_parser.c with "profile-use=/root/tempesta/http_parser.c.gcov" in #pragma
GCC optimize. http_parser.c.gcov is generated by

$ cd ~/tempesta
$ ./scripts/gcov_gather.sh /tmp/gcov.tgz
$ cd /tmp && tar -zxf gcov.tgz && cd -
$ gcov -o /tmp/sys/kernel/debug/gcov/root/tempesta/tempesta_fw http_parser.c

perf(1) also can be used for the profile generation, see -fauto-profile GCC
command line option.

PGO generates slightly better code layout, but the profile counts total
samples and doesn't record sequence of called code. E.g. for a normal HTTP
request the parser spends more time in URI processing than in method
processing, so PGO moves URI before method processing which is completely
wrong.

The second drawback of PGO is that each code update requires regeneration of
the profile. Moreover, in general it's tricky to correctly choose the right
workload for the profile, so it looks more like workload programming than the
parser programming.

All in all, I leave the GCOV compilation option only for the test coverage and
we need to use more flexible way to generate proper code layout.

### DIFF
--- a/scripts/gcov_gather.sh
+++ b/scripts/gcov_gather.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+#
+# Gcov gather script for Tempesta FW data.
+# Mainly bothered from linux/Documentation/dev-tools/gcov.rst
+#
+# Copyright (C) 2019 Tempesta Technologies, Inc.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License,
+# or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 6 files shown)


scripts/gcov_gather.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+#
+# Gcov gather script for Tempesta FW data.
+# Mainly bothered from linux/Documentation/dev-tools/gcov.rst
```
i-rinat: borrowed?

PR #1226, opened 2019-03-26
bmi2 check for preload avx2 cripto

avbelov23:
For avx2 crypto modules, the bmi2 instruction set is required.


i-rinat:
(obsoleted by #1227)

Issue #1224, opened 2019-03-26
tempesta_tls.ko: Unknown symbol in module


Theoooooo:
There is the steps to reproduce the problem :

- Installed Debian 9.5.0 and configure an access to pull the required packages for the installation
- Build the kernel patch (4.14.32) from source and install it with these instructions : https://github.com/tempesta-tech/tempesta/wiki/Install-from-Sources
-  Build and install the Tempesta source with the same link.

When i start the service or load the modules with the tempesta.sh, i got the following errors :

----@Tempesta:~/linux-4.14.32-tfw/tempesta# scripts/tempesta.sh --load
Loading Tempesta kernel modules...
Loading module tempesta_tls
insmod: ERROR: could not insert module /root/linux-4.14.32-tfw/tempesta/tls/tempesta_tls.ko: Unknown symbol in module
ERROR: cannot load tempesta TLS module

I got no error during the compilation/installation from both the kernel and tempesta.
I still have the entire session logs if you need the full logs during the compilation/installation of the kernel patch and tempesta.

ikoveshnikov:
Hi Théo,

I just sent an email to you. Could you provide some more information on the issue:
- are you building Tempesta from current master?
- what is the missed symbol? Could you provide sytem log after `scripts/tempesta.sh --load` command was called?

ikoveshnikov:
related to #1218

Theoooooo:
Thanks for the answer, there is the informations you need :

- I build the sources from the "git clone https://github.com/tempesta-tech/tempesta"
- After installer the kernel patch, i rebooted with the patch version properly (i think)
- There is the error return by the script with the "dmesg" command :
`[ 2237.062950] tempesta_lib: loading out-of-tree module taints kernel.
[ 2237.316588] [tempesta tls] ERROR: Cannot preload sha512-avx2 module, please check /proc/crypto.
`

avbelov23:
@Theoooooo, do you try under virtualbox under windows?

Theoooooo:
@avbelov23 I can't try it to VirtualBox i don't have the necessary right to test this on my computer. I install it in a VM under VMware ESXi 6.0.0

avbelov23:
@Theoooooo, please show `cat / proc / cpuinfo` output on vm

ikoveshnikov:
[...]

PR #1227, opened 2019-03-26
Request non-specific cryptographic modules

avbelov23:
Requires #1224 

We request non-specific cryptographic modules, since on some virtual machines there is no bmi2 instruction set needed for avx2 cryptographic modules

### DIFF
--- a/tls/crypto.c
+++ b/tls/crypto.c
@@ -161,29 +161,29 @@ typedef struct {
 static TlsMdInfo ttls_sha224_info = {
 	.type		= TTLS_MD_SHA224,
 	.name		= "SHA224",
-	.alg_name	= "sha224-avx2",
-	.hmac_name	= "hmac(sha224-avx2)",
+	.alg_name	= "sha224",
+	.hmac_name	= "hmac(sha224)",
 };
 
 static TlsMdInfo ttls_sha256_info = {
 	.type		= TTLS_MD_SHA256,
 	.name		= "SHA256",
-	.alg_name	= "sha256-avx2",
-	.hmac_name	= "hmac(sha256-avx2)",
+	.alg_name	= "sha256",
+	.hmac_name	= "hmac(sha256)",
 };
 
 static TlsMdInfo ttls_sha384_info = {
 	.type		= TTLS_MD_SHA384,
 	.name		= "SHA384",
-	.alg_name	= "sha384-avx2",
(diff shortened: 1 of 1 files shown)


tls/crypto.c
```diff
@@ -161,29 +161,29 @@ typedef struct {
 static TlsMdInfo ttls_sha224_info = {
 	.type		= TTLS_MD_SHA224,
 	.name		= "SHA224",
-	.alg_name	= "sha224-avx2",
-	.hmac_name	= "hmac(sha224-avx2)",
+	.alg_name	= "sha224",
```
ikoveshnikov: @i-rinat has explained in chat about priorities in Linux crypto, why this solution works and why it's best option. Please add the comment before `ttls_sha224_info`

tls/crypto.c
```diff
@@ -442,6 +457,25 @@ ttls_crypto_modinit(void)
...
+	/*
+	 * Below, so that when the "shaN" request is loaded, the driver with the
+	 * highest priority is loaded, even if some one with a lower priority is
+	 * already loaded, you must try to load all cripto kernel modules.
+	 */
+	for (inst_set = inst_sets; *inst_set; ++inst_set) {
+		for (h = ttls_hashes; h->info; ++h) {
+			size_t len = snprintf(NULL, 0, "crypto-%s%s",
+					      h->info->alg_name, *inst_set);
+			char *module = kmalloc(len + 1, GFP_KERNEL);
```
i-rinat: You don't need to allocate memory here. `request_module()` is using `vsnprintf()` inside, so you can just:
```c
request_module("crypto-%s%s", h->info->alg_name, *inst_set);
```

tls/crypto.c
```diff
@@ -442,6 +457,25 @@ ttls_crypto_modinit(void)
 		}
 	}
 
+	/*
+	 * Below, so that when the "shaN" request is loaded, the driver with the
+	 * highest priority is loaded, even if some one with a lower priority is
+	 * already loaded, you must try to load all cripto kernel modules.
```
i-rinat: I'd change that to something like:
"Each hash algorithm may have multiple implementations, optimized for different instruction set extensions (such as AVX, AVX2, and so on). As we want the fastest available, we try to load all the implementations. Crypto API will then choose one with the highest priority."

Issue #1196, opened 2019-02-23
HTTP/2 streams prioritization


krizhanovsky:
According to [RFC 7540 5.3](https://tools.ietf.org/html/rfc7540#section-5.3) we should schedule or service from the cache requests according to their priority and dependency. Reprioritization also must be implemented.

Depends on #309.

aleksostapenko:
Additional discussion regarding possible dependency/priority tree structure is in https://github.com/tempesta-tech/tempesta/pull/1233#issue-265503627.

krizhanovsky:
## TCP send queue and frames prioritization

At the moment Tempesta FW pushes all the HTTP/2 frames to the TCP send queue, just like it does for HTTP/1 pipelining, so a higher priority frame can be pushed to the queue long after many low priority frames in the head of the queue. During prioritization we need to reorder the queue. However, the queue reordering will introduce unnecessary overhead.

### Generic considerations

1. Prioritization should be done in `tcp_write_xmit()` since streams priorities may change while we're processing the TCP send queue
2. Due to skb splitting no any skb can keep data from different streams (HTTP responses), so streams prioritization is equal to skbs prioritization
3. We do our best to minimize the kernel changes, however with QUIC from #724, which we're also going to upstream, streams handling should go to the kernel. With this task we should keep as much logic as possible in Tempesta code base and patch the kernel only in a minimal way, but it'd be good to make the streams foundation in the kernel for QUIC streams.

### Proposal

1. `ss_do_send()` should call a new function instead of `skb_entail()` operating with the TCP write queue. The function should add an skb to a tree (or other data structure depending on the scheduling algorithm) of lists. Each list correponds to a stream (HTTP response).
2. `ss_do_send()` also should call a new function instead of `tcp_push()`, which does seems unnecessary for us tasks (e.g. URG data is never used with HTTP). The function is probably just a wrapper for `__tcp_push_pending_frames()`.
3. `tcp_write_xmit()` should call for HTTP/2 sockets a Tempesta FW prioritization callback instead of `tcp_send_head()`. THe callback will run the scheduling algorithm and return an skb, which it decides to send now. @EvgeniiMekhanik does this callback can be used also in #1394 as the required second callback?

krizhanovsky:
[Comment](https://github.com/tempesta-tech/tempesta/issues/1394#issuecomment-1522967817) from #1394 :

Doesn't block connection and continue to forward Headers frames while flow control limit is exhausted
[...]

PR #1233, opened 2019-03-28
HTTP/2 Framing layer implementation (#309).

aleksostapenko:
#### Working HTTP/2 functionality in PR
1. PING and SETTINGS frames exchange: receiving, parsing and response with ACK; tested via `h2i` utility (https://github.com/golang/net/tree/master/http2/h2i);
2. Establishing HTTP/2 connection; tested via `nghttp` utility - the part of `nghttp2`/`nghttp2-client` packages (https://nghttp2.org):
   1. Client's connection preface (starting magic sequence and SETTINGS frame) receiving and parsing;
   2. Response with server side preface (initial SETTINGS frame);
   3. Send SETTINGS ACK frame (in response to client's preface SETTINGS frame);
   4. Receiving and parsing client's SETTINGS ACK frame (in response to server side preface SETTINGS fram
   5. Receiving and parsing client's PRIORITY frames;
   6. Receiving and parsing client's HEADERS frame;
   7. Applying client's HEADERS as initial frame for new stream creation.

Command:
```
nghttp -v https://<name_of_host_with_started_tempesta_fw>
```
Note: to test HTTP/2 connection establishing via `nghttp` (described above) - it is necessary to replace the implementation of `APP_FRAME()` macro with simple `false` value, since upper level for processing of application frames (HEADERS, DATA and CONTINUATION) is not implemented yet. 

#### HTTP/2 Stream Scheduler
In context of #309 only base implementation of priority/dependency scheduler is planned (the rest must be implemented in #1196). So this PR contains only initial form of scheduler: at stage of #309 all streams are assumed to have equal weights and depends on the one root logical stream (with ID equal 0); thus, streams are processed in order of their frames/request/response come from TCP connection, and in this case we need just a simple streams' storage with fast search by ID. In current PR red-black tree based storage is used; this structure has been chosen - compared to hash table - due to lack of per-connection memory overhead, since for hash table some amount of memory (maybe large enough) must be allocated initially - at the connection's creation moment, and besides, this could be a ddos vector.

Considering future priority/dependency scheduler implementation in context of #1196 - an appropriate scheme for the scheduler, in my opinion, would be a tree-like structure (let's call it a dependency tree), built with the help of, "parent" and "child" pointers (embedded into Stream structure) for vertical links and "prev" and "next" pointers - for the horizontal links between neighbor streams of the same level. A logical stream with identifier 0 acts as a root of this tree. In addition to the specified links, each stream has a pointer to a special priority queue (based on the binary heap), in which only its child single-level streams with jobs to be processed (frames/requests/responses) are placed.

Thus, the red-black tree, mentioned above, in proposed scheme acts as a basic storage structure for a quick search of streams; on top of it - the dependency tree acts as a structure to track current dependencies between streams and to calculate effective priorities for each stream in the tree — based on their weights (in case of re-prioritization etc.); also dependency tree provides up/down traversals between dependency levels; and the priority queues, in turn, track the next stream to be processed for each level.


tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,950 @@
...
+
+#define VERIFY_FRAME_SIZE(ctx)						\
+do {									\
+	if ((ctx)->hdr.length < 0)					\
+		return tfw_http2_conn_terminate(ctx,			\
+						FRAME_ECODE_SIZE_ERROR); \
+} while (0)
+
+static int
+tfw_htt2_stream_verify(TfwHttp2Ctx *ctx)
```
i-rinat: Looks like a typo:
```suggestion
tfw_http2_stream_verify(TfwHttp2Ctx *ctx)
```

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,950 @@
...
+	 * TODO: apply settings entry.
+	 */
+	return T_OK;
+}
+
+static int
+tfw_http2_settings_process(TfwHttp2Ctx *ctx)
+{
+	TfwFrameHdr *hdr = &ctx->hdr;
+	int id  = ntohl(*(int *)&ctx->rbuf[0]);
```
i-rinat:
```suggestion
	int id  = ntohs(*(unsigned short *)&ctx->rbuf[0]);
```

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,950 @@
...
+
+#define FRAME_FSM_READ_LAMBDA(to_read, lambda)				\
+do {									\
+	WARN_ON_ONCE(ctx->rlen >= (to_read));				\
+	n = min_t(int, (to_read) - ctx->rlen, buf + len - p);		\
+	lambda;								\
+	p += n;								\
+	ctx->rlen += n;							\
+	if (unlikely(ctx->rlen < (to_read)))				\
+		T_FSM_EXIT();						\
```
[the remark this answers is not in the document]
aleksostapenko: In this case we'll exit only if amount of already read bytes is less then needed amount, so we must continue reading in current state until `ctx->rlen == to_read`.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,950 @@
...
+	TfwStr data = {
+		.chunks = (TfwStr []){
+			{},
+			{ .data = ctx->rbuf, .len = ctx->rlen }
+		},
+		.len = ctx->rlen,
+		.nchunks = 2
+	};
+	TfwFrameHdr hdr = {
+		.length = FRAME_SRVC2_SIZE,
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/http_frame.c
aleksostapenko: Corrected.

tempesta_fw/http_frame.c
aleksostapenko: Corrected.

tempesta_fw/http_frame.c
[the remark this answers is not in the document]
i-rinat: Let's assume we received data chunk that has initial magic string, then SETTINGS frame header and part of the first setting in the SETTING frame.
We'll start at `HTTP2_RECV_CLI_START_SEQ`, read the magic string, then move to `HTTP2_RECV_FIRST_SETTINGS`, parse frame header, then move to `HTTP2_RECV_FRAME_SETTINGS`, where `FRAME_FSM_READ_SRVC()` will invoke `FRAME_FSM_READ_LAMBDA()` where we read part of a setting when data chunk ends. At that point we'll get here, call `T_FSM_EXIT()`, which jumps to `__fsm_done` defined in `T_FSM_FINISH()`, where `ctx->state` will be set to `__fsm_const_state`. Most probably it will be `0`, and not `HTTP2_RECV_FRAME_SETTINGS`, as expected. So when the next data chunk arrives, we'll start reading the magic string again.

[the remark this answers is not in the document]
aleksostapenko: While we are jumping from state to state - `__fsm_const_state` is updated in `T_FSM_STATE`, so it should have the actual value in `T_FSM_FINISH`.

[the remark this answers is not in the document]
i-rinat: Ah, I see now. Yes, that's OK.

tempesta_fw/tls.c
```diff
@@ -666,6 +703,84 @@ tfw_tls_cfg_require(void)
...
+	protos = kzalloc(TTLS_ALPN_PROTOS * sizeof(ttls_alpn_proto), GFP_ATOMIC);
+	if (unlikely(!protos))
+		return -ENOMEM;
+
+	protos[order].name = TTLS_ALPN_HTTP1;
+	protos[order].len = sizeof(TTLS_ALPN_HTTP1) - 1;
+	protos[order].id = TTLS_ALPN_ID_HTTP1;
+	protos[!order].name = TTLS_ALPN_HTTP2;
+	protos[!order].len = sizeof(TTLS_ALPN_HTTP2) - 1;
+	protos[!order].id = TTLS_ALPN_ID_HTTP2;
```
i-rinat: Looks like HTTP/2 is unconditionally enabled even if `listen` directive has `https:http1` and doesn't mention http2 at all. Is this intended behavior?

tls/tls_srv.c
```diff
@@ -308,14 +308,13 @@ ttls_parse_session_ticket_ext(TlsCtx *tls, unsigned char *buf, size_t len)
...
+	size_t list_len, cur_len;
 	const unsigned char *theirs, *start, *end;
-	const char **ours;
+	const ttls_alpn_proto *our;
 
-	/* If ALPN not configured, just ignore the extension */
-	/* TODO #309 #834 confugure on Tempesta side: "h2, http/1.1" */
-	if (!tls->conf->alpn_list)
-		return 0;
+	/* If TLS processig is enabled, ALPN must be configured. */
```
i-rinat:
```suggestion
	/* If TLS processing is enabled, ALPN must be configured. */
```

tempesta_fw/connection.h
```diff
@@ -209,9 +210,11 @@ enum {
 typedef struct {
 	TfwCliConn	cli_conn;
 	TlsCtx		tls;
+	TfwHttp2Ctx	*h2;
```
ikoveshnikov: The structure name becomes misleading, now it's more than TLS connection, it also stores the HTTP2 context. And a new issue is introduced: HTTP2 context is not allocated for non-TLS HTTP2 connections (which are allowed by standard).
In the same time HTTP2 connections will have two contexts in the same time: HTTP2 and HTTP1. HTTP1 context is stored as `TfwConn->TfwHttpParser`. Probably we should introduce a union in `TfwConn` class to store 'protocol context' whatever http2 or http1. Seems like we never want both. Even is the HTTP1 connection is upgraded to HTTP2 we can easily re-init the field.

tempesta_fw/http.c
```diff
@@ -3884,8 +3885,8 @@ tfw_http_resp_process(TfwConn *conn, const TfwFsmData *data)
 /**
  * @return status (application logic decision) of the message processing.
  */
-static int
-__tfw_http_msg_process(void *conn, TfwFsmData *data)
+int
+tfw_http_msg_process_generic(void *conn, TfwFsmData *data)
```
ikoveshnikov: Bad naming. It's not generic. It's HTTP1 processing.

tempesta_fw/http.c
```diff
@@ -3930,7 +3931,9 @@ tfw_http_msg_process(void *conn, TfwFsmData *data)
 		if (likely(r == T_OK || r == T_POSTPONE)) {
 			data->skb->next = data->skb->prev = NULL;
 			data->trail = !next ? trail : 0;
-			r = __tfw_http_msg_process(conn, data);
+			r = TFW_CONN_TLS((TfwConn *)conn) && tfw_http2_context(conn)
```
ikoveshnikov: Seems like `TFW_CONN_TLS((TfwConn *)conn)` is extra here. HTTP2 recommends using HTTP2 over TLS (`h2`) but it still allows HTTP2 over TCP (`h2c`). Personally I don't understand why the unencrypted connections are allowed by standard, but we have to follow them.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1017 @@
+/**
+ *		Tempesta FW
+ *
+ * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
+ * Copyright (C) 2015-2019 Tempesta Technologies, Inc.
```
ikoveshnikov: Bad copy-paste. This file wasn't existed before, so only `2019` here.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1017 @@
...
+#undef DEBUG
+#endif
+#include "lib/fsm.h"
+#include "lib/str.h"
+#include "procfs.h"
+#include "http.h"
+#include "http_frame.h"
+
+
+#define FRAME_CLI_MAGIC		"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n"
```
ikoveshnikov: Why `CLI`? Just curious. I would prefer names closer to the standard to make searching easier, e.g. `PREFACE`, but it's just a matter of taste and nothing more.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1017 @@
...
+	HTTP2_WINDOW_UPDATE,
+	HTTP2_CONTINUATION
+} TfwFrameType;
+
+/*
+ * HTTP/2 error codes (RFC 7540 section 7). Used in RST_STREAM
+ * and GOAWAY frames to report the reasons of the stream or
+ * connection error.
+ */
+#define FRAME_ECODE_PROTO	0x1
```
ikoveshnikov: Section 7 contains 14 error codes. I understand that only two are needed at this moment, but it seems that we will definitely will need all of them, let's introduce enum like `TfwFrameType` to store all the possible errors.

tempesta_fw/http_frame.h
```diff
@@ -0,0 +1,116 @@
+/**
+ *		Tempesta FW
+ *
+ * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
+ * Copyright (C) 2015-2019 Tempesta Technologies, Inc.
```
ikoveshnikov: Same here. Bad copy-paste.

tempesta_fw/http_frame.h
```diff
@@ -0,0 +1,116 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 59
+ * Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ */
+#ifndef __HTTP_FRAME__
```
ikoveshnikov: `__HTTP2_FRAME__`?

tempesta_fw/http_frame.h
```diff
@@ -0,0 +1,116 @@
...
+	HTTP2_RECV_FRAME_SETTINGS,
+	HTTP2_RECV_FRAME_GOAWAY,
+	HTTP2_RECV_FRAME_PADDED,
+	HTTP2_RECV_HEADER_PRI,
+	HTTP2_IGNORE_FRAME_DATA,
+	__HTTP2_RECV_FRAME_APP,
+	HTTP2_RECV_HEADER = __HTTP2_RECV_FRAME_APP,
+	HTTP2_RECV_CONT,
+	HTTP2_RECV_DATA
+} TfwFrameState;
```
ikoveshnikov: Do we really need to expose this states to other modules? Why it's not in `http2_frame.c`?

tempesta_fw/http_frame.h
```diff
@@ -0,0 +1,116 @@
...
+ * 4.1). Reserved bit is not present here since it has no any semantic
+ * value for now and should be always ignored.
+ *
+ * @length	- the frame's payload length;
+ * @stream_id	- id of current stream (which frame is processed);
+ * @type	- the type of frame being processed;
+ * @flags	- frame's flags;
+ */
+typedef struct {
+	int		length;
```
ikoveshnikov: Never going to be below 0, thus should be unsigned.

i-rinat: > Never going to be below 0

Currently, code in `tfw_http2_recv_priority()` may temporarily make `length` negative. If, for example, client sends a wrong length in PRIORITY frame.

tempesta_fw/http_frame.h
aleksostapenko: Yeah, missed that. Thanks!

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1017 @@
...
+		FRAME_FSM_READ_SRVC(ctx->to_read);
+
+		if (tfw_http2_frame_type_process(ctx))
+			FRAME_FSM_EXIT(T_DROP);
+
+		FRAME_FSM_EXIT(T_OK);
+	}
+
+	T_FSM_STATE(HTTP2_RECV_FRAME_SETTINGS) {
+		BUG_ON(ctx->to_read > FRAME_HEADER_SIZE);
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1017 @@
...
+			goto out;
+		}
+	}
+
+	/*
+	 * Before transferring the skb with app frame for further processing,
+	 * certain service data should be separated from it (placed at the
+	 * frame's beginning): frame header, optional pad length and optional
+	 * priority data (the latter is for HEADERS frames only). Besides,
+	 * DATA and HEADERS frames can containing some padding in the frame's
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1017 @@
...
+	 * tail, but we don't need to worry about that here since such padding
+	 * is processed as service data, separately from app frame, and it
+	 * will be just splitted into separate skb (above).
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1017 @@
...
+	}
+	}
+
+	FRAME_FSM_FINISH();
+
+	return r;
+}
+
+/*
+ * Initialization of HTTP/2 framing context. Due to passing frames to
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1017 @@
...
+static inline void
+tfw_http2_context_init(TfwHttp2Ctx *ctx, bool postponed)
+{
+	if (!APP_FRAME(ctx) || (!postponed && !ctx->padlen)) {
+		bzero_fast(ctx->__off,
+			   sizeof(*ctx) - offsetof(TfwHttp2Ctx, __off));
+		return;
+	}
+	if (!postponed && ctx->padlen) {
+		ctx->state = HTTP2_IGNORE_FRAME_DATA;
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1017 @@
...
+#include "lib/fsm.h"
+#include "lib/str.h"
+#include "procfs.h"
+#include "http.h"
+#include "http_frame.h"
+
+
+#define FRAME_CLI_MAGIC		"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n"
+#define FRAME_CLI_MAGIC_LEN	24
+#define FRAME_SRVC1_SIZE	4
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1017 @@
...
+#define FRAME_PRIORITY_SIZE	5
+#define FRAME_STNGS_ENTRY_SIZE	6
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1017 @@
...
+		break;
+
+	case HTTP2_HEADERS:
+		if (hdr->flags & HTTP2_F_PRIORITY)
+			return tfw_http2_recv_priority(ctx);
+		ctx->state = HTTP2_RECV_HEADER;
+		break;
+
+	default:
+		/* Only DATA and HEADERS frames cab be padded. */
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tls/tls_srv.c
aleksostapenko: Corrected.

tempesta_fw/http_frame.c
aleksostapenko: Corrected.

tempesta_fw/http_frame.c
aleksostapenko: Done.

tempesta_fw/http_frame.h
aleksostapenko: Corrected.

tempesta_fw/tls.c
aleksostapenko: Yes, currently `http1` is just a shortcut for `http1,http2` and `http2` is a shortcut for `http2,http1`. Thus, we always enable http2 handling, and since our protocols order takes precedence, we can define what protocol will be used when client sends us in any order both `http/1.1` and `h2` in ALPN.
But in any case, this point (as well as the names for HTTP/1.1 and HTTP/2 protocols in configuration) remains the subject TBD.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1017 @@
...
+	 * TODO: send appropriately filled GOAWAY frame, set
+	 * Conn_Stop flag for connection, and close it (possibly
+	 * via SS_F_CONN_CLOSE flag).
+	 */
+	return 0;
+}
+
+#define VERIFY_FRAME_SIZE(ctx)						\
+do {									\
+	if ((ctx)->hdr.length <= 0)					\
```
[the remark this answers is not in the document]
aleksostapenko: Added handling for zero length payload.

tempesta_fw/http_frame.c
aleksostapenko: `CLI` - since only the client sends this magic sequence in its preface during HTTP/2 connection establishing (https://tools.ietf.org/html/rfc7540#section-3.5). But I agree that `PREFACE` will also not be superfluous here.
Done.

aleksostapenko: Done.

tempesta_fw/connection.h
aleksostapenko: >HTTP2 context is not allocated for non-TLS HTTP2 connections (which are allowed by standard)

In result of discussion in #309 and in chat - it has been decided not to support non-TLS HTTP/2 in TempestaFW.
>Probably we should introduce a union in TfwConn class to store 'protocol context' whatever http2 or http1.

Yes, agree with that. The similar approach is proposed in https://github.com/tempesta-tech/tempesta/issues/309#issuecomment-467447205 (in `Stream Priority/Dependencies`). These changes is part of future PRs in context of #309 (`"Request HTTP/2 -> HTTP/1.1"` or more sooner `"Disable @seq_queue for HTTP/2"`). During implementation of Framing, Scheduler and Stream States - I temporary left that part as is - for testing/debugging purposes.

tempesta_fw/http.c
aleksostapenko: It has been decided not to support non-TLS HTTP/2 in TempestaFW: standard allows HTTP/2 over plain TCP, but does not force to use it.

Makefile
```diff
@@ -35,8 +35,10 @@ DBG_HTTP_PARSER ?= 0
 DBG_SS ?= 0
 DBG_TLS ?= 0
 DBG_APM ?= 0
+DBG_HTTP_FRAME ?= 0
```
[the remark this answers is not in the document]
aleksostapenko: For now I cannot do such separation: these sets do not intersect yet; also both HTTP/1.1 and HTTP/2 protocols  belong to the same HTTP domain, so there is no contradiction here. But in the nearest future (e.g. during HTTP parser implementation changing in context of HTTP/2 processing) mentioned intersections may appear, and in this case the separation for `HTTP1_*` and `HTTP2_*`  may be really needed and will be made for preprocessor variables (and possibly for filenames too, e.g. `http_frame.c/h`, `http_stream.c/h`).

tempesta_fw/http_frame.h
aleksostapenko: The file has name `http_frame.h`, so this will be irrelevant (also see https://github.com/tempesta-tech/tempesta/pull/1233#discussion_r279336557).

tempesta_fw/http.c
aleksostapenko: In nearest future PRs (in context of #309) this function is planned for usage for HTTP/2 processing too, so I ran a little forward here. Most part of `tfw_http_msg_process_generic()` is general processing, whatever HTTP/1.1 or HTTP/2 protocols (except the parsing stage only), so it is worth to try adopt it for both protocols. In any case if this idea will fail, and the separate procedures will be needed - the name will be reverted accordingly.

tempesta_fw/http_stream.c
```diff
@@ -0,0 +1,408 @@
...
+			new = &parent->rb_left;
+		} else if (id > stream->id) {
+			new = &parent->rb_right;
+		} else {
+			WARN_ON_ONCE(1);
+			return NULL;
+		}
+	}
+
+	new_stream = kzalloc(sizeof(TfwStream), GFP_ATOMIC);
```
krizhanovsky: The same allocation problem: basically we usually don't use pattern `kzalloc(sizeof(....), ...)` in any more or less performance crucial code - if we have fixed size allocations, then we use some kind of SLAB allocator.

tempesta_fw/http_stream.c
```diff
@@ -0,0 +1,408 @@
...
+	if (unlikely(!new_stream))
+		return NULL;
+
+	tfw_http2_init_stream(new_stream, id, weight);
+
+	rb_link_node(&new_stream->node, parent, new);
+	rb_insert_color(&new_stream->node, &sched->streams);
```
krizhanovsky: We'll need priority index plus to current index for stream ID (as you said in the PR description), so we need two indexes. However, binary heap is technically an array (continuous memory location which usually gives us good spactial locality in run time - this is why I wanted binary heap in #309), and, since you already use dynamic memory allocator, all your `TfwStream` instances reside somewhere in memory and it's not wise to use additional array just for heap index. It's better to introduce additional `node` entries in `TfwStream` to build the second index.

If you fix the `kzalloc()` above, then SLAB/SLOB/SLUB already cares about spacial locality. The question is: do we expect to see more frequently frames from the same TCP connection streams or from streams as they first appeared (to fit the internal SLAB allocation pattern)? I think the first one, but I won't bet that there is so strong spacial locality in access pattern to streams from the same heap/rb-tree. Moreover, if we use binary heap, as I proposed, on top of an array, we'll have to manager erased items, i.e. move stored items and this is basically slower than SLAB's bitmasks.

All it all, I think RB tree is a good enough decision here since there is no confidence in access pattern. I write the comment for #1196 and will leave link there to the comment.

krizhanovsky: We'll need priority index plus to current index for stream ID (as you said in the PR description), so we need two indexes. However, binary heap is technically an array (continuous memory location which usually gives us good spactial locality in run time - this is why I wanted binary heap in #309), and, since you already use dynamic memory allocator, all your `TfwStream` instances reside somewhere in memory and it's not wise to use additional array just for heap index. It's better to introduce additional `node` entries in `TfwStream` to build the second index.

If you fix the `kzalloc()` above, then SLAB/SLOB/SLUB already cares about spacial locality. The question is: do we expect to see more frequently frames from the same TCP connection streams or from streams as they first appeared (to fit the internal SLAB allocation pattern)? I think the first one, but I won't bet that there is so strong spacial locality in access pattern to streams from the same heap/rb-tree. Moreover, if we use binary heap, as I proposed, on top of an array, we'll have to manager erased items, i.e. move stored items and this is basically slower than SLAB's bitmasks.

All it all, I think RB tree is a good enough decision here since there is no confidence in access pattern. I write the comment for #1196 and will leave link there to the comment.

tempesta_fw/tls.c
```diff
@@ -59,6 +60,32 @@ tfw_tls_chop_skb_rec(TlsCtx *tls, struct sk_buff *skb, TfwFsmData *data)
...
+	TlsCtx *tls = &conn->tls;
+
+	if (likely(!tls->alpn_chosen
+		   || tls->alpn_chosen->id != TTLS_ALPN_ID_HTTP2
+		   || conn->h2))
+	{
+		return 0;
+	}
+
+	h2 = kzalloc(sizeof(TfwHttp2Ctx), GFP_ATOMIC);
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/tls.c
aleksostapenko: OK. Corrected.

tempesta_fw/http_frame.c
```diff
@@ -0,0 +1,1411 @@
...
+			 * (see RFC 7540 section 5.1.2 for details).
+			 */
+			if (ctx->lsettings.max_streams <= ctx->streams_num) {
+				T_DBG("Max streams number exceeded: %lu\n",
+				      ctx->streams_num);
+				SET_TO_READ_VERIFY(ctx, HTTP2_IGNORE_FRAME_DATA);
+				return tfw_http2_stream_terminate(ctx,
+								  hdr->stream_id,
+								  NULL,
+								  HTTP2_ECODE_REFUSED);
```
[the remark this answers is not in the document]
aleksostapenko: Corrected.

tempesta_fw/http_stream.c
aleksostapenko: Yeah - missed that. Thanks!

aleksostapenko: Corrected.

tempesta_fw/tls.c
```diff
@@ -666,6 +706,84 @@ tfw_tls_cfg_require(void)
...
+int
+tfw_tls_cfg_alpn_protos(const char *cfg_str)
+{
+	ttls_alpn_proto *protos;
+	size_t len = strlen(cfg_str);
+	int order = 0;
+
+#define CONF_HTTPS	"https"
+#define CONF_HTTP1	"http1"
+#define CONF_HTTP2	"http2"
```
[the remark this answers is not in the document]
aleksostapenko: Done.

tempesta_fw/http.c
```diff
@@ -3930,7 +3931,9 @@ tfw_http_msg_process(void *conn, TfwFsmData *data)
...
+				? tfw_http2_frame_process(conn, data)
+				: tfw_http_msg_process_generic(conn, data);
```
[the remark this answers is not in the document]
aleksostapenko: >Check H2 state in tfw_http_req_process()?
Yeah - this must be a frame type and, maybe, a stream state (since we have to track request's start/end, and this exactly matches with stream's states, as mentioned in penultimate paragraph in https://tools.ietf.org/html/rfc7540#section-8.1).
But currently I'm not very optimistic about the parser division into separate parts: indeed, we have rather effective implementation with single-jump cost to enter any state; and with divided parser we will get costs of additional function calls (if not inlined) and necessity for additional branching to call separate parser parts (with associated costs). But in general, I am still thinking about this point.

tempesta_fw/http_stream.c
aleksostapenko: OK. I'll think yet further about building a dependency/priority tree.

PR #1238, opened 2019-03-29
Add changelog

ikoveshnikov:
Writing a good changelog is very time consuming and manual operation, which is not compatible with CD. When a new feature is implemented or issue is fixed, author should apply changes to `CHANGELOG.md`. If it would be so, then chandgelog generation for binary packages can be created automatically in CD.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,82 @@
+0.7.0 (Unreleased)
+
+Enhancements:
+
+  * Improve parser performance by GCC jumps to labels feature.
+
+Fixes:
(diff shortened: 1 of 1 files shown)


ikoveshnikov:
>  I'm not sure whthere it makes sense to log each closed issue to the file because of quick pollution by minor changes probably not so interested to users

Agreed. I was thinking about that and came to conclusion, that the change log is list of major changes that can significantly affect user experience: crucial  issue fixes, new features, incompatible changes. So the changelog only contains major changes.

PR #1241, opened 2019-04-03
tls: avoid updating the checksum with contents of "Finished" message

i-rinat:
Sometimes, data come in pieces, and the Finished message may become split into two or more parts. In that case `ttls_recv()` gets `T_POSTPONE` from `ttls_handshake_step()` and feeds the received bytes from the Finished message the the (handshake_messages) checksum, which is not the desired behavior. The patch avoids that by making state `TTLS_CLIENT_FINISHED` non-checksummable.

As `ttls_handshake_server_step()` returns `T_OK` after each part successfully parsed, all data before Finished message will always be checksummed as required.


i-rinat:
> if there are no ideas how to make the code better, then at least a good comment explaining each state check is required for ttls_hs_checksumable().

I found no sensible way to change the current checksumming scheme, so I just added a comment with explanation. More to that, checksumming comment in `ttls_parse_finished()` was also extended, as I misunderstood the reason of including of client's Finished into the sum the first time I read the code.

PR #1249, opened 2019-05-06
Rewrite HTTP strings processing in assembler

krizhanovsky:
Rewrite HTTP strings processing in assembler:
1. remove tls/dummy_headers
2. replace dynaic table initialization with static one
3. shrink tables size
4. small assembly optimizations
    
Introduce test_debug_relax() to avoid debug message drop on unit tests with debug.


tempesta_fw/str_avx2.S
```diff
@@ -0,0 +1,2296 @@
...
+	subq	$88, %rsp
+	movq	__A, %r9
+	movq	__D, %r8
+
+	/* Process 8 byte prefix of suffix of the string. */
+.str2low_prefix_suffix:
+	cmpq	$8, %rdx
+	ja	.str2low_test_len128
+	movq	.str2low_switch(,%rdx,8), %rax
+	jmpq	*%rax
```
i-rinat: These indirect jumps make objtool generate a warning: "warning: objtool: __tfw_strtolower_avx2()+0x46: indirect jump found in RETPOLINE build". To suppress it we need either annotate these jumps like that:
```diff
diff --git a/tempesta_fw/str_avx2.S b/tempesta_fw/str_avx2.S
index 5f890fe6..df3e2e2f 100644
--- a/tempesta_fw/str_avx2.S
+++ b/tempesta_fw/str_avx2.S
@@ -24,6 +24,7 @@
 #include <linux/linkage.h>
 #include <asm/alternative-asm.h>
 #include <asm/export.h>
+#include <asm/nospec-branch.h>
 
 #define CASE	0x2020202020202020
 
@@ -476,6 +477,7 @@ ENTRY(__tfw_strtolower_avx2)
 	cmpq	$8, %rdx
 	ja	.str2low_test_len128
 	movq	.str2low_switch(,%rdx,8), %rax
+	ANNOTATE_RETPOLINE_SAFE
 	jmpq	*%rax
 .section .rodata
 .align	8
@@ -653,6 +655,7 @@ ENTRY(__tfw_stricmp_avx2)
 
 	/* Process short strings below 8 bytes in length. */
 	movq	.stricmp_switch(,%rdx,8), %rax
+	ANNOTATE_RETPOLINE_SAFE
 	jmpq	*%rax
 .section .rodata
 .align 8
@@ -1009,6 +1012,7 @@ ENTRY(__tfw_stricmp_avx2_2lc)
 	ja	.sic2lc_short
 
 	movq	.sic2lc_switch(,%rdx,8), %rax
+	ANNOTATE_RETPOLINE_SAFE
 	jmpq	*%rax
 .section	.rodata
 .align 8
```
or enable Spectre mitigation:
```diff
diff --git a/tempesta_fw/str_avx2.S b/tempesta_fw/str_avx2.S
index 5f890fe6..d5d2334c 100644
--- a/tempesta_fw/str_avx2.S
+++ b/tempesta_fw/str_avx2.S
@@ -24,6 +24,7 @@
 #include <linux/linkage.h>
 #include <asm/alternative-asm.h>
 #include <asm/export.h>
+#include <asm/nospec-branch.h>
 
 #define CASE	0x2020202020202020
 
@@ -476,7 +477,7 @@ ENTRY(__tfw_strtolower_avx2)
 	cmpq	$8, %rdx
 	ja	.str2low_test_len128
 	movq	.str2low_switch(,%rdx,8), %rax
-	jmpq	*%rax
+	JMP_NOSPEC	%rax
 .section .rodata
 .align	8
 .str2low_switch:
@@ -653,7 +654,7 @@ ENTRY(__tfw_stricmp_avx2)
 
 	/* Process short strings below 8 bytes in length. */
 	movq	.stricmp_switch(,%rdx,8), %rax
-	jmpq	*%rax
+	JMP_NOSPEC	%rax
 .section .rodata
 .align 8
 .stricmp_switch:
@@ -1009,7 +1010,7 @@ ENTRY(__tfw_stricmp_avx2_2lc)
 	ja	.sic2lc_short
 
 	movq	.sic2lc_switch(,%rdx,8), %rax
-	jmpq	*%rax
+	JMP_NOSPEC	%rax
 .section	.rodata
 .align 8
 .sic2lc_switch:
```

tempesta_fw/str_avx2.S
```diff
@@ -0,0 +1,2296 @@
...
+	vmovdqa	__D, %xmm0
+	vpsubb	__C(%rip), %xmm1, %xmm2
+	vpcmpgtb %xmm2, %xmm0, %xmm0
+	vpand	__CASE, %xmm0, %xmm0
+	vpor	%xmm1, %xmm0, %xmm0
+	vmovups	%xmm0, (%rdi,%r11)
+	movq	%rbx, %r11
+	jmp	.str2low_test_len8
+
+.str2low_tolower8:
```
i-rinat: Use of `volatile` in the C source code resulted in obviously redundant memory stores and loads. As far as I understand, code below is equivalent to:
```asm
        movq    (%rsi,%r11), %rax
        movq    %r9, %mm2
        movq    %r8, %mm1
        movq    %rax, %mm0
        movq    %rax, %r12
        psubb   %mm2, %mm0
        pcmpgtb %mm0, %mm1
        movq    %mm1, %rax
        andq    %rcx, %rax
        orq     %r12, %rax
        movq    %rax, (%rdi,%r11)
        movslq  %ebx, %rax
        jmp     .str2low_small_len
```

tempesta_fw/str_avx2.S
```diff
@@ -0,0 +1,2296 @@
...
+	andq	%rcx, %rax
+	orq	%r12, %rax
+	movq	%rax, (%rdi,%r11)
+	movslq	%ebx, %rax
+	jmp	.str2low_small_len
+
+ENDPROC(__tfw_strtolower_avx2)
+
+/**
+ * Case insensitive comprison of the strings %RDI and %RSI of length %RDX.
```
i-rinat:
```suggestion
 * Case insensitive comparison of the strings %RDI and %RSI of length %RDX.
```

tempesta_fw/str_avx2.S
```diff
@@ -0,0 +1,2296 @@
...
+	jmp	.stricmp_do_len3
+.stricmp_len4:
+	xorl	%eax, %eax
+	jmp	.stricmp_do_len4
+.stricmp_len5:
+	xorl	%edx, %edx
+	jmp	.stricmp_do_len5
+.stricmp_len6:
+	xorl	%edx, %edx
+	jmp	.stricmp_do_len6
```
i-rinat: I wonder, what's happening here? Why are `xorl %edx,%edx` are all around? It can be zeroed just before `jmpq *%rax`. That will remove at least one jump from cases 1 to 7. (I know it's machine-generated. I'd like to understand why compiler did that.)

tempesta_fw/str_avx2.S
```diff
@@ -0,0 +1,2296 @@
...
+	vmovdqa	__CASE, %ymm0
+	vpxor	%ymm6, %ymm1, %ymm6
+	vmovdqa	__D, %ymm5
+	vpor	%ymm1, %ymm0, %ymm1
+	vpsubb	%ymm7, %ymm1, %ymm1
+	vpcmpgtb %ymm1, %ymm5, %ymm2
+	vpcmpeqb %ymm0, %ymm6, %ymm1
+	vpand	%ymm1, %ymm0, %ymm1
+	vpand	%ymm2, %ymm1, %ymm1
+	vpxor	%xmm2, %xmm2, %xmm2
```
i-rinat: This looks out of place. Other instructions are working with `ymm` registers, but this one only clears `xmm2`, lower part of `ymm2`. Why?

tempesta_fw/str_avx2.S
i-rinat: Nevermind. I now see that registers nullified are different. Sometimes it's `%rax`, not `%rdx`, because compare code alternates between them for some reason.

tempesta_fw/str_avx2.S
```diff
@@ -0,0 +1,2296 @@
...
+	vpmovmskb %ymm0, %eax
+	cmpl	$~0, %eax
+	jne	.sic2lc_nomatch
+	movl	%ecx, %eax
+	jmp	.sic2lc_tail
+
+ENDPROC(__tfw_stricmp_avx2_2lc)
+
+/**
+ * Match input string %RDI with legth %RSI against custom allowed alphabet
```
i-rinat:
```suggestion
 * Match input string %RDI with length %RSI against custom allowed alphabet
```

tempesta_fw/str.c
```diff
@@ -897,4 +1245,25 @@ tfw_str_dprint(const TfwStr *str, const char *msg)
 				(int)c->len, c->data);
 	}
 }
+
+#define D(n)	(unsigned int)v[n]
```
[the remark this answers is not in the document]
krizhanovsky: Seems not:
```C
$ cat t.c 
#include <stdio.h>

int
main(int argc, char *argv[])
{
	unsigned char c = 0xff;
	char s = 0xff;
	return printf("%x %x\n", (unsigned int)c, (unsigned int)s);
}
$ gcc t.c && ./a.out 
ff ffffffff
```

tempesta_fw/str_avx2.S
i-rinat: Oh, I forgot that VEX-encoded instructions operating on `xmm` reset upper half of `ymm` register, so this effectively zeroes the whole `ymm2`.

tempesta_fw/str.c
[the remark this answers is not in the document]
i-rinat: At the time of writing I was sure that `v` is `char *`, not `unsigned char *`. Now that I see it's unsigned, I also see that my comment is wrong. It's fine to use `(unsigned int)` here.

tempesta_fw/str_avx2.S
krizhanovsky: We check jump array bounds against a constant, encoded in an instruction opcode, so there is no cache miss on boundary check [required for Spectre](https://spectreattack.com/spectre.pdf). `JPM_NOSPEC` gives about 15% performance impact in comparison with `ANNOTATE_RETPOLINE_SAFE`: 101ms vs 85ms in the `tfw_str` microbenchmark.

PR #1251, opened 2019-05-06
tls: check union field accesses in TlsHandshake

i-rinat:
~~During early stages of handshake process code calculates both sha384 and sha256 sums. Context for the latter is stored where `ecdh_ctx` is located. It turned out that there are access clashes in the code, and `ecdh_ctx` may be changed while sha256 sum is still in the process of calculation.~~

~~The patchset has a patch for fixing the clash,~~ and another patch, that adds a way to determine such clashes at runtime in debug builds.

**Upd.** During PR discussion it was found that although there were accesses to `edch_ctx` while its storage is used for sha256 context, there is no possibility of sha256 context corruption. So it's safe, and this PR doesn't fix anything, but only adds a framework of union field access tracking.

### DIFF
--- a/tls/tls_internal.h
+++ b/tls/tls_internal.h
@@ -135,6 +135,7 @@ typedef struct tls_handshake_t {
 	ttls_sig_hash_set_t		hash_algs;
 	int				sni_authmode;
 
+	unsigned char			ecdh_point_format;
 	unsigned char			point_form		: 1,
 					extended_ms		: 1,
 					new_session_ticket	: 1,
--- a/tls/tls_internal.h
+++ b/tls/tls_internal.h
@@ -143,6 +143,14 @@ typedef struct tls_handshake_t {
 					cli_exts		: 1,
 					curves_ext		: 1;
 
+#ifdef DEBUG
+	union {
+		unsigned char dhm_ctx_tag : 2;
+		unsigned char ecdh_ctx_tag : 2;
+		unsigned char tmp_sha256_tag : 2;
+	};
+#endif /* DEBUG */
+
 	size_t				pmslen;
 	ttls_key_cert			*key_cert;
 	ttls_key_cert			*sni_key_cert;
@@ -154,10 +162,14 @@ typedef struct tls_handshake_t {
 	int  (*tls_prf)(const unsigned char *, size_t, const char *, size_t,
 			const unsigned char *, size_t, unsigned char *, size_t);
 
+	union {
 #if defined(TTLS_DHM_C)
-	ttls_dhm_context		dhm_ctx;
+		ttls_dhm_context	dhm_ctx;
 #endif
-	ttls_ecdh_context		ecdh_ctx;
+		ttls_ecdh_context	ecdh_ctx;
+		ttls_sha256_context	tmp_sha256;
+	};
+
 	union {
 		struct shash_desc	desc; /* common for both the contexts */
 		ttls_sha256_context	fin_sha256;
@@ -173,6 +185,55 @@ typedef struct tls_handshake_t {
 		unsigned char		tmp[TTLS_HS_RBUF_SZ];
 	};
 } TlsHandshake;
+
+#if defined(DEBUG) && (DEBUG >= 1)
+
+/*
+ * Debug helper macro that checks if it's OK to use a particular union field.
+ * Current valid union field is tracked in a separate variable — tag.
+ *
+ * Initially, the tag value is expected to be zero. That means it's not decided
+ * yet what union field is used. First access to a protected union field with
+ * help of the macro, sets a tag to some non-zero value. Then access to any
+ * other union field fires a BUG_ON.
+ */
+#define TTLS_HS_generic_field(hs, field, tag)				\
+(*({									\
+	T_DBG3("%s:%d accessing hs->%s, tag: %d → %d\n",		\
+	       __func__, __LINE__, #field, (hs)->field##_tag, tag);	\
+	BUG_ON((hs)->field##_tag != 0 && (hs)->field##_tag != tag);	\
+	(hs)->field##_tag = tag;					\
+	&(hs)->field;							\
+}))
+
+/*
+ * Similar to TTLS_HS_generic_field, but without tag checking.
+ * Always resets the tag.
+ */
+#define TTLS_HS_generic_field_reset_tag(hs, field, tag)			\
+(*({									\
+	T_DBG3("%s:%d accessing hs->%s, tag %d → %d (override)\n",	\
+	       __func__, __LINE__, #field, (hs)->field##_tag, tag);	\
+	(hs)->field##_tag = tag;					\
+	&(hs)->field;							\
+}))
+
+#else
+
+#define TTLS_HS_generic_field(hs, fieldname, tag) ((hs)->fieldname)
+#define TTLS_HS_generic_field_reset_tag(hs, fieldname, tag) ((hs)->fieldname)
+
+#endif /* defined(DEBUG) && (DEBUG >= 1) */
+
+#define TTLS_HS_dhm_ctx(hs)	TTLS_HS_generic_field(hs, dhm_ctx, 1)
+#define TTLS_HS_ecdh_ctx(hs)	TTLS_HS_generic_field(hs, ecdh_ctx, 2)
+#define TTLS_HS_tmp_sha256(hs)	TTLS_HS_generic_field(hs, tmp_sha256, 3)
+
+#define TTLS_HS_dhm_ctx_reset_tag(hs) 					\
+	TTLS_HS_generic_field_reset_tag(hs, dhm_ctx, 1)
+
+#define TTLS_HS_ecdh_ctx_reset_tag(hs) 					\
+	TTLS_HS_generic_field_reset_tag(hs, ecdh_ctx, 2)
 
 /*
  * List of certificate + private key pairs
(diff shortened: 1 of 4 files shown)


tls/tls_internal.h
```diff
@@ -135,13 +135,22 @@ typedef struct tls_handshake_t {
 	ttls_sig_hash_set_t		hash_algs;
 	int				sni_authmode;
 
+	unsigned char			ecdh_point_format;
 	unsigned char			point_form		: 1,
```
krizhanovsky: This is probably good to debug, but now in production code we have point format here, in handshake descriptor, and in ECDH context - this is confusing and it simply has no sense for production code. It'd be better to make some nicer solution or at least move the field under `ifdef DEBUG`.

tls/tls_internal.h
```diff
@@ -173,6 +186,55 @@ typedef struct tls_handshake_t {
...
+#define TTLS_HS_dhm_ctx(hs)	TTLS_HS_generic_field(hs, dhm_ctx, 1)
+#define TTLS_HS_ecdh_ctx(hs)	TTLS_HS_generic_field(hs, ecdh_ctx, 2)
+#define TTLS_HS_tmp_sha256(hs)	TTLS_HS_generic_field(hs, tmp_sha256, 3)
+
+#define TTLS_HS_dhm_ctx_reset_tag(hs) 					\
+	TTLS_HS_generic_field_reset_tag(hs, dhm_ctx, 1)
+
+#define TTLS_HS_ecdh_ctx_reset_tag(hs) 					\
+	TTLS_HS_generic_field_reset_tag(hs, ecdh_ctx, 2)
+
```
krizhanovsky: Clever enough and I appreciate the debugging code. Do you think that we'll need the code later? I thought that this is single-time problem, which is unlikely to appear once to be fixed... I like the code, but I'm not sure whether it's useful. If you believe that it's good to have the debugging option, then just fix the issue with double ECDH point format.

tls/tls_srv.c
```diff
@@ -1287,8 +1287,19 @@ ttls_write_server_key_exchange(TlsCtx *tls, struct sg_table *sgt,
...
+		TTLS_HS_ecdh_ctx(tls->hs).point_format =
+			tls->hs->ecdh_point_format;
 		ttls_get_ecdh_params_from_cert(tls);
+	} else if (ttls_ciphersuite_uses_ecdhe(ci)) {
+		ttls_ecdh_init(&TTLS_HS_ecdh_ctx_reset_tag(tls->hs));
+		TTLS_HS_ecdh_ctx(tls->hs).point_format =
+			tls->hs->ecdh_point_format;
+	} else if (ttls_ciphersuite_uses_dhe(ci)) {
+		ttls_dhm_init(&TTLS_HS_dhm_ctx_reset_tag(tls->hs));
+	}
```
krizhanovsky: This is the place which confuses me: the debugging mess which works in production and if you need to understand the code, you still have no understanding that the `point_format` games are just for debugging.

tls/ttls.c
```diff
@@ -371,6 +370,7 @@ ttls_update_checksum(TlsCtx *tls, const unsigned char *buf, size_t len)
 	 * hs->ecdh_ctx to store SHA256 checksum data.
 	 */
 	if (unlikely(IS_ERR_OR_NULL(ci))) {
+		ttls_sha256_context *sha256 = &TTLS_HS_tmp_sha256(hs);
```
krizhanovsky: Yeah, this looks much better than previous type cast.

tls/tls_internal.h
i-rinat: It wasn't for debugging. This new `ecdh_point_format` field is to avoid (possible) data corruption.

When we are receiving handshake data, we actively use temporary sha256 context which lives where `ecdh_ctx` is located. But also during handshake processing we use `ecdh_ctx` in `tls->hs->ecdh_ctx.point_format = p[0];`. That way we may accidentally change some byte in sha256 context while it's still used.

I've just checked the numbers again, and it looks like my hypothesis about field access was wrong all the time. Since `offsetof(ttls_ecdh_context, point_format)` is `376`, but `sizeof(ttls_sha256_context)` is only `120`, stores to `ecdh_ctx.point_format` will never touch sha256 context.

So, it turned out the PR doesn't fix any bug. But I'd like to keep it, as it helps to reduce anxiety. At the current moment `ecdh_ctx.point_format` field is at the safe place. However if somebody later tries to reorder `ttls_ecdh_context` struct, and moves `point_format` out of the safe zone, a bug will appear. To avoid that, we need to store `point_format` somewhere before we stop using `ecdh_ctx` memory for anything other than `ecdh_ctx`.

tls/tls_internal.h
i-rinat: I'd like to have those macros in the code. They add a little noise to the code, but help to test whenever there are union field collisions. In release build macros are expanded to direct field accesses, so they are not slowing hot paths.

I like the macros because I came up with the idea. So I need an opinion other than mine to determine if they are useful. :-) I guess they may be useful if we'll need to pack fields into unions.

tls/tls_srv.c
i-rinat: > are just for debugging.

I merged storage spaces of `ecdh_ctx` and `dhm_ctx` into the same union, as they are not used simultaneously. And since they are not separate anymore, we can't just initialize both `ecdh_ctx` and `dhm_ctx` unconditionally. This function is where we decide what we use, and here we are initializing either `dhm_ctx` or `ecdh_ctx`, depending on what is used for the handshake.

Initializations are clearing both `ecdh_ctx` and `dhm_ctx`. If `point_format` is not stored anywhere else, we'll lose it.

One of the solutions will be something like this:
```c
{
    unsigned char pf = tls->hs->ecdh_point_format;
    ttls_ecdh_init(&tls->hs->ecdh_ctx);
    tls->hs->ecdh_point_format = pf;
}
```
But as said in comments above, I'd like to avoid storing anything in `ecdh_ctx` while its memory is used for something else (sha256).

tls/tls_internal.h
krizhanovsky: The data structures reusage is a hack. I had made several tries to reuse different data structures before I came to ECDH descriptor and I was still wrong. I chose ECDH descriptor because it didn't require additional changes crucial for the code design and clarity (of course any memory reusage is doubtful in terms of code clarity, but it's still a common practice, so that's fine).

I propose to look for other data structures which can be used to place the second checksum. On in this case I'd prefer just to make a comment about special requirements to the data structure members placement.

> But I'd like to keep it, as it helps to reduce anxiety.

I love the macros which make the pointers cast nicer and debugging code which helps the type conversions tracking, but I still do not like the double ecdh point format games.

> Oh, and there is a gap in the struct around there, so adding another unsigned char doesn't increase structure size.

That's fine, but it's still very confusing to seem the same ecdh point format in 2 places.

tls/tls_internal.h
krizhanovsky: There is old time agreement about Tempesta code is that we do our best to not to pollute the production code by code which is helpful for testing. We still have a lot of EXPORT statements just to satisfy out test module - that's not good, but also not so harmful. Debugging code is also a pollution, but we can't live without it.

However, with the ECDH point format discussion above, I though about a TLS unit test - why not to create an instance of necessary data structures in the unit test and just verify that all the members (at least on boudns of data structures) are safe? If somebody later change the members order - the unit test fails there there will be good comments what's wrong. What do you think?

tls/tls_srv.c
krizhanovsky: Oh, I see, I missed the union for dhm and ecdh.

> I'd like to avoid storing anything in ecdh_ctx while its memory is used for something else (sha256)

Why? If we have a unit test verifying the data structures layouts and sizes to guarantee that they don't overwrite crucial members, will we be good?

BTW https://www.youtube.com/watch?v=GgTwx1399Dc is good to watch - keeping the data structures small is crucial. But we also should not sacrifice the code design clarity.

i-rinat: > Why?

That simplifies reasoning about code. With access restrictions no clashing may occur, so we can stop worrying about that and focus on other issues. Checking macros cause immediate crash if access was wrong. So that not only helps to detect existing field access bugs, and also to prevent similar bugs in the future.

> If we have a unit test verifying the data structures layouts and sizes to guarantee that they don't overwrite crucial members, will we be good?

Yeah, kind of. But first we need to find such members, and that's hard to do simply by looking at the code.

As this discussion may continue indefinitely, I propose to extract some changes (merging storage spaces of `ecdh_ctx` and `dhm_ctx`, more precise initialization and freeing of that members, adding explicit `tmp_sha256` field to avoid type casting in the code) into https://github.com/tempesta-tech/tempesta/pull/1254, and close this PR (#1251).

i-rinat:
Extracted the non-disputed parts of this PR into https://github.com/tempesta-tech/tempesta/pull/1254.

tls/tls_internal.h
i-rinat: > a TLS unit test - why not to create an instance of necessary data structures in the unit test and just verify that all the members (at least on boudns of data structures) are safe?

Yeah, we should definitely check that. I've added a `BUILD_BUG_ON` statement testing that (in #1254).

tls/tls_internal.h
i-rinat: > very confusing to seem the same ecdh point format in 2 places.

Removed that field, reverted to the version without separate field. (See 1254).

PR #1254, opened 2019-05-13
tls: merge dhm_ctx, ecdh_ctx, and tmp_sha256 fields into a union

i-rinat:
As ecdh_ctx and dhm_ctx share same storage space, their constructors and destructors are now called in a more precise way.

### DIFF
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -355,7 +355,6 @@ ttls_update_checksum(TlsCtx *tls, const unsigned char *buf, size_t len)
 	int r;
 	TlsHandshake *hs = tls->hs;
 	const TlsCiphersuite *ci = tls->xfrm.ciphersuite_info;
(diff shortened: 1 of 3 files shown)


tls/ttls.c
```diff
@@ -1262,10 +1262,16 @@ ttls_handshake_free(TlsHandshake *hs)
 
 	crypto_free_shash(hs->desc.tfm);
 
+	if (!IS_ERR_OR_NULL(ci)) {
+		if (ttls_ciphersuite_uses_ecdh(ci) ||
+		    ttls_ciphersuite_uses_ecdhe(ci))
+			ttls_ecdh_free(&hs->ecdh_ctx);
```
ikoveshnikov: Usually we use curve brackets if the `if` statement takes more than one line.

i-rinat: Added.

PR #1255, opened 2019-05-14
tls: merge dhm_ctx, ecdh_ctx, and tmp_sha256 fields into a union

i-rinat:
As ecdh_ctx and dhm_ctx share same storage space, their constructors and
destructors are now called in a more precise way.

(Backporting https://github.com/tempesta-tech/tempesta/pull/1254, because it's not solely an optimization, but also a bugfix. Crashes were possible if client were to disconnect during initial stages of handshake processing. In that case `ttls_ecdh_free()` could be called on incorrect data.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+0.6.3 (2019-05-17)
+
+  * Fix crashes on TLS handshakes utilizing SHA384.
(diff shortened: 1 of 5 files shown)


i-rinat:
Changed version to 0.6.3, added changelog file entry.

i-rinat:
Added `0.6.3` tag.

PR #1259, opened 2019-05-27
tls: propagate sequence numbers through the touched parts of tcp write queue

i-rinat:
Currently, we encrypt data in TLS connections in `tfw_tls_encrypt()`, and during that operation, data chunks are enlarged to fit auxiliary data. As TCP sequence numbers are calculated earlier, we adjust the numbers on the go. To reduce TLS overhead, multiple skb's are aggregated into a single record if possible.

The patchset corrects sequence number calculation in the case where skb's are actually aggregated. Also, couple of side issues are addressed too: a NULL dereference and spurious WARN_ON assertion triggering.

(fixes #1177)

PR #1260, opened 2019-05-27
tls: propagate sequence numbers through the touched parts of tcp write queue (release-0.6)

i-rinat:
Currently, we encrypt data in TLS connections in `tfw_tls_encrypt()`, and during that operation, data chunks are enlarged to fit auxiliary data. As TCP sequence numbers are calculated earlier, we adjust the numbers on the go. To reduce TLS overhead, multiple skb's are aggregated into a single record if possible.

The patchset corrects sequence number calculation in the case where skb's are actually aggregated. Also, couple of side issues are addressed too: a NULL dereference and spurious WARN_ON assertion triggering.

(backport of #1259)


i-rinat:
Synchronized changes with #1259.

Issue #1217, opened 2019-03-21
TLS: Memory corruption


avbelov23:
Tempesta config:
```
server 127.0.0.1:80; #nginx
listen 443 proto=https;
tls_certificate /home/avb/Projects/tempesta/tfw-root.crt;
tls_certificate_key /home/avb/Projects/tempesta/tfw-root.key;
```

Log sh:
```
$ cat /var/www/html/index.html 
<!DOCTYPE html>
<html>
<body>
Test
</body>
</html>
$ curl -k https://localhost
<!DOCTYPE html>
<html>
<body>
Test
</body>
</html>
$ cat /var/www/html/index.html 
�\>�(���e�����m�#����O̬�����r��(���t�����Zl���7�>+ɩ�����4�k�!�/$QS��N���
$ echo 3 | sudo tee /proc/sys/vm/drop_caches
3
$ cat /var/www/html/index.html 
<!DOCTYPE html>
<html>
<body>
Test
</body>
</html>
```

avbelov23:
Blame `sendfile        on;`

avbelov23:
It turns out that thanks to the in-place implementation of TLS, we encrypt data pages without allocating additional memory and copying, but because sendfile links the socket with the file cache, we overwrite the file cache.

If several clients request the same file, we will overwrite it in different places at once and each client will receive a unique file.

Then the problem can be reproduced not only with the help of sendfile. If the file is given from the cache, then the result will be the same. When sending from our cache, we reuse the pages of memory from the cache.

krizhanovsky:
@avb good catch - this the fundamental problem of out TLS encryption!

The fix should be straightforward enough - we need to copy skb data before encryption if an skb is formed from splice() (sendfile() for file cache or splice() for user-space buffer, but the core is the same) or our web cache. We do something similar in `ss_send()`. It's easy to do for our cache, but I'm not sure whether we know the nature of skb data in `tcp_write_xmit()` - probably some kernel patching is required. However, maybe `SKBTX_SHARED_FRAG` flag will help us, see `do_tcp_sendpages()`.

ikoveshnikov:
Also note, that only payload part of the response served from cache reuses pages. The headers are created from scratch for every new response. Same can happen to `sendfile()` responses. But in the same time backend server may have its own cache and use zero-copy TCP transmissions. So it's can be very tricky.

Probably the copying of skb data should be configurable option:
[...]

PR #1264, opened 2019-05-29
tls: encrypting in new sk_buff pages, if `sendfile()` is used or response from cache

avbelov23:
#1217

Transfer the `tx_flags` to the new `skb_buff` and encrypt in the new `sk_buff` pages if SKBTX_SHARED_FRAG is set or response from cache when send response via https.

### DIFF
--- a/tempesta_fw/cache.c
+++ b/tempesta_fw/cache.c
@@ -1539,6 +1539,13 @@ tfw_cache_build_resp(TfwHttpReq *req, TfwCacheEntry *ce)
 		goto free;
 
 	/*
+	 * Mark skb as SKBTX_SHARED_FRAG so that when sending it to the client
+	 * via https, encryption routines do not change data in-place.
+	 */
+	if (TFW_CONN_TLS(req->conn))
+		skb_shinfo(it.skb)->tx_flags |= SKBTX_SHARED_FRAG;
+
+	/*
 	 * Allocate HTTP headers table of proper size.
 	 * There were no other allocations since the table is allocated,
 	 * so realloc() just grows the table and returns the same pointer.
--- a/tempesta_fw/tls.c
+++ b/tempesta_fw/tls.c
@@ -230,7 +230,7 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
 #define AUTO_SEGS_N	8
 
 	int r = -ENOMEM;
-	unsigned int head_sz, tag_sz, len, frags, t_sz;
+	unsigned int head_sz, tag_sz, len, frags, t_sz, out_frags;
 	unsigned char type;
 	struct sk_buff *next = skb, *skb_tail = skb;
 	struct tcp_skb_cb *tcb = TCP_SKB_CB(skb);
@@ -239,8 +239,12 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
 	TlsXfrm *xfrm;
 	struct sg_table sgt = {
 		.nents = skb_shinfo(skb)->nr_frags + !!skb_headlen(skb),
+	}, out_sgt = {
+		.nents = skb_shinfo(skb)->nr_frags + !!skb_headlen(skb),
 	};
-	struct scatterlist sg[AUTO_SEGS_N];
+	struct scatterlist sg[AUTO_SEGS_N], out_sg[AUTO_SEGS_N];
+	struct page **pages = NULL, **pages_end, **p;
+	struct page *auto_pages[AUTO_SEGS_N];
 
 	if (unlikely(sk->sk_user_data == NULL))
 		return -EINVAL;
@@ -297,6 +301,7 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
 
 		len += next->len;
 		sgt.nents += skb_shinfo(next)->nr_frags + !!skb_headlen(next);
+		out_sgt.nents += skb_shinfo(next)->nr_frags + !!skb_headlen(next);
 		skb_tail = next;
 	}
 
@@ -317,6 +322,7 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
 		if (r < 0)
 			goto out;
 		sgt.nents += r;
+		out_sgt.nents += r;
 
 		r = ss_skb_expand_head_tail(skb_tail->next, skb_tail, 0,
 					    tag_sz);
@@ -324,6 +330,7 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
 			goto out;
 	}
 	sgt.nents += r;
+	out_sgt.nents += r;
 
 	/*
 	 * The last skb in our list will bring TLS tag - add it to end_seqno.
@@ -379,17 +386,29 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
 
 	if (likely(sgt.nents <= AUTO_SEGS_N)) {
 		sgt.sgl = sg;
+		out_sgt.sgl = out_sg;
+		pages = auto_pages;
 	} else {
-		sgt.sgl = kmalloc(sizeof(struct scatterlist) * sgt.nents,
-				  GFP_ATOMIC);
+		char *ptr = kmalloc(sizeof(struct scatterlist) * sgt.nents +
+			            sizeof(struct scatterlist) * out_sgt.nents +
+			            sizeof(struct page *) * out_sgt.nents,
+				    GFP_ATOMIC);
+		sgt.sgl = ptr;
 		if (!sgt.sgl) {
-			T_WARN("cannot alloc TLS encryption scatter list.\n");
+			T_WARN("cannot alloc memory for TLS encryption.\n");
 			return -ENOMEM;
 		}
+
+		ptr += sizeof(struct scatterlist) * sgt.nents;
+		out_sgt.sgl = ptr;
+
+		ptr += sizeof(struct scatterlist) * out_sgt.nents;
+		pages = pages_end = ptr;
 	}
 	sg_init_table(sgt.sgl, sgt.nents);
+	sg_init_table(out_sgt.sgl, out_sgt.nents);
 
-	for (next = skb, frags = 0; ; ) {
+	for (next = skb, frags = 0, out_frags = 0; ; ) {
 		/*
 		 * skb data and tails are already adjusted above,
 		 * so use zero offset and skb->len.
@@ -405,6 +424,14 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
 		if (r <= 0)
 			goto out;
 		frags += r;
+
+		r = ss_skb_to_sgvec_with_new_pages(next,
+		                                   out_sgt.sgl + out_frags,
+		                                   &pages_end);
+		if (r <= 0)
+			goto out;
+		out_frags += r;
+
 		tempesta_tls_skb_clear(next);
 		if (next == skb_tail)
 			break;
@@ -412,6 +439,7 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
 			break;
 		next = tcp_write_queue_next(sk, next);
 		sg_unmark_end(&sgt.sgl[frags - 1]);
+		sg_unmark_end(&out_sgt.sgl[out_frags - 1]);
 	}
 	WARN_ON_ONCE(sgt.nents != frags);
 
@@ -420,10 +448,13 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
 	/* Set IO context under the lock before encryption. */
 	io->msglen = len - TLS_HEADER_SIZE;
 	io->msgtype = type;
-	if (!(r = ttls_encrypt(tls, &sgt)))
+	if (!(r = ttls_encrypt(tls, &sgt, &out_sgt)))
 		ttls_aad2hdriv(xfrm, skb->data);
 
 	spin_unlock(&tls->lock);
+
+	for (p = pages; p < pages_end; ++p)
+		put_page(*p);
 
 out:
 	if (unlikely(sgt.nents > AUTO_SEGS_N))
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -770,7 +770,7 @@ ttls_aead_req_free(struct crypto_aead *tfm, struct aead_request *req)
  * different sessions, always have different ciphertexts.
  */
 int
-ttls_encrypt(TlsCtx *tls, struct sg_table *sgt)
+ttls_encrypt(TlsCtx *tls, struct sg_table *sgt, struct sg_table *out_sgt)
 {
 	int r, elen;
 	TlsXfrm *xfrm = &tls->xfrm;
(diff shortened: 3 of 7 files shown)


tempesta_fw/tls.c
```diff
@@ -569,6 +570,43 @@ tfw_tls_conn_drop(TfwConn *c)
...
+		unsigned int frag_size = skb_frag_size(f);
+
+		p = skb_frag_page(f);
+		order = get_order(frag_size);
+		p2 = alloc_pages(__GFP_MEMALLOC, order);
+		if (!p2)
+			goto err;
+		vaddr = kmap_atomic(p);
+		vaddr2 = kmap_atomic(p2);
+		memcpy(vaddr2, vaddr + f->page_offset, frag_size);
```
i-rinat: Currently, encryption overwrites data in-place. But we may create another scatter-gather list for destination, and avoid copying data here. I'm talking about `aead_request_set_crypt(req, sgt->sgl, sgt->sgl, elen, xfrm->iv_enc);` in `ttls_encrypt()`. As you can see, currently third argument which is destination, is the same as second (source).

tempesta_fw/tls.c
```diff
@@ -577,10 +615,23 @@ tfw_tls_conn_drop(TfwConn *c)
...
 tfw_tls_conn_send(TfwConn *c, TfwMsg *msg)
 {
+	struct sk_buff *skb, *skb_head = NULL;
 	int r;
 	TlsCtx *tls = tfw_tls_context(c);
 	TlsIOCtx *io = &tls->io_out;
 
+	while ((skb = ss_skb_dequeue(&msg->skb_head))) {
+		if (skb_shinfo(skb)->tx_flags & SKBTX_SHARED_FRAG) {
+			r = tfw_tls_skb_copy_pages(skb);
```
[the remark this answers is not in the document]
avbelov23: This function (`tfw_tls_conn_send()`) calls only when sent via https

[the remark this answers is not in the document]
i-rinat: Yeah, indeed. I forgot that we get here only on TLS connections.

tempesta_fw/tls.c
```diff
@@ -569,6 +570,43 @@ tfw_tls_conn_drop(TfwConn *c)
 	tfw_conn_hook_call(TFW_FSM_HTTP, c, conn_drop);
 }
 
+int tfw_tls_skb_copy_pages(struct sk_buff *skb)
+{
+	int i;
+	struct sk_buff *frag_iter;
+
```
krizhanovsky: Need a comment why don't we need to copy skb head.

tempesta_fw/tls.c
```diff
@@ -569,6 +570,43 @@ tfw_tls_conn_drop(TfwConn *c)
...
+	for (i = 0; i < skb_shinfo(skb)->nr_frags; i++) {
+		skb_frag_t *f = &skb_shinfo(skb)->frags[i];
+		struct page *p, *p2;
+		int order;
+		u8 *vaddr, *vaddr2;
+		unsigned int frag_size = skb_frag_size(f);
+
+		p = skb_frag_page(f);
+		order = get_order(frag_size);
+		p2 = alloc_pages(__GFP_MEMALLOC, order);
```
krizhanovsky: We're in softirq, so `GFP_ATOMIC` must be used instead

tempesta_fw/tls.c
krizhanovsky: `memcpy()` sucks, we can and should use `memcpy_fast()`

tempesta_fw/tls.c
```diff
@@ -569,6 +570,43 @@ tfw_tls_conn_drop(TfwConn *c)
...
+		if (!p2)
+			goto err;
+		vaddr = kmap_atomic(p);
+		vaddr2 = kmap_atomic(p2);
+		memcpy(vaddr2, vaddr + f->page_offset, frag_size);
+		kunmap_atomic(vaddr2);
+		kunmap_atomic(vaddr);
```
krizhanovsky: We're in softirq, so there is no preemtion or page faults, so it's safe just to use `page_address()`.

tempesta_fw/tls.c
krizhanovsky: Yes, this is definitely the better approach. If TCP/IP stack clears `SKBTX_SHARED_FRAG`, then we can use `tempesta_tls_skb_settype()` to pass the knowledge about the skb nature through the stack. Also probably `ttls_encrypt()` should accept an argument controlling how to perform the encryption - with or without copying.

tempesta_fw/tls.c
```diff
@@ -379,23 +388,87 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
...
 		 * skb data and tails are already adjusted above,
 		 * so use zero offset and skb->len.
 		 */
 		r = skb_to_sgvec(next, sgt.sgl + frags, 0, next->len);
 
+		if (skb_shinfo(next)->tx_flags & SKBTX_SHARED_FRAG ||
+		    tls->resp_from_cache)
+		{
+			sg_set_buf(out_sgt.sgl + out_frags, next->data,
+			           skb_headlen(next));
```
i-rinat: `sg_set_buf()` call can make a zero-sized entry if `skb_headlen(next)` is empty, which may happen if `next` has data only in fragments. I'd make this call conditional.

tempesta_fw/tls.c
```diff
@@ -379,23 +388,87 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
...
+					if (!p) {
+						r = -ENOMEM;
+						goto out;
+					}
+					remain = (1 << order) * PAGE_SIZE;
+					offset = 0;
+				} else {
+					skb_frag_t *prev_f;
+
+					prev_f = &skb_shinfo(next)->frags[i - 1];
```
i-rinat: Say, we processed one `skb`, and after the last fragment of that `skb` there is remaining room in an allocated page. Then we pick another skb (`next = tcp_write_queue_next(sk, next);`), and start the cycle again. This time `remain` may be large enough so we get to this line with `i` being `0`. With subsequent out-of-bounds access.

tempesta_fw/tls.c
```diff
@@ -379,23 +388,87 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
...
+	remain = offset = 0;
+	for (next = skb, frags = 0, out_frags = 0, elt = 0; ; ) {
 		/*
 		 * skb data and tails are already adjusted above,
 		 * so use zero offset and skb->len.
 		 */
 		r = skb_to_sgvec(next, sgt.sgl + frags, 0, next->len);
 
+		if (skb_shinfo(next)->tx_flags & SKBTX_SHARED_FRAG ||
+		    tls->resp_from_cache)
```
i-rinat: Just a thought. All skb's with shared should have fragments point to pages which are also held elsewhere. So can we determine which fragments contain shared data by checking `page_ref_count(skb_frag_page(f))`?

tls/ttls.c
```diff
@@ -1760,7 +1766,7 @@ ttls_write_finished(TlsCtx *tls, struct sg_table *sgt, unsigned char **in_buf)
 
 	sg_init_table(&sg, 1);
 	sg_set_buf(&sg, p, TLS_HEADER_SIZE + TTLS_HS_FINISHED_BODY_LEN);
-	if ((r = ttls_encrypt(tls, &enc_sgt)))
+	if ((r = ttls_encrypt(tls, &enc_sgt, NULL)))
```
i-rinat: If you use `&enc_sgt` for both second and third parameter here, `if (out_sgt)` checks from `ttls_encrypt()` could be removed.

tempesta_fw/tls.c
i-rinat: Also, how about setting `SKBTX_SHARED_FRAG` flag in `tfw_cache_build_resp_body()` where generated skb's with data from the cache? If that works, it may make passing `resp_from_cache` flag through the call chain obsolete.

tls/ttls.c
```diff
@@ -791,10 +791,16 @@ ttls_encrypt(TlsCtx *tls, struct sg_table *sgt)
 	T_DBG3_BUF("IV used", xfrm->iv_enc, xfrm->ivlen);
 
 	elen = ttls_msg2crypt_len(io, xfrm);
-	ttls_make_aad(tls, io, sg_virt(sgt->sgl));
+	if (out_sgt)
```
i-rinat: Let's never pass `NULL` as `out_sgt` here, that will simplify code a bit.

tempesta_fw/tls.c
ikoveshnikov: Agree here. The whole `resp_from_cache` thing looks like messy workaround. Anyway, the copy should happen not for the whole message, but only for the body. If setting `SKBTX_SHARED_FRAG` flag is not applicable for the skb (some very specific processing insed the kernel), we can copy data in `cache.c` and use in-place crypto operations later.

tempesta_fw/tls.c
```diff
@@ -577,10 +615,23 @@ tfw_tls_conn_drop(TfwConn *c)
 static int
 tfw_tls_conn_send(TfwConn *c, TfwMsg *msg)
 {
+	struct sk_buff *skb, *skb_head = NULL;
 	int r;
 	TlsCtx *tls = tfw_tls_context(c);
 	TlsIOCtx *io = &tls->io_out;
 
+	while ((skb = ss_skb_dequeue(&msg->skb_head))) {
+		if (skb_shinfo(skb)->tx_flags & SKBTX_SHARED_FRAG) {
```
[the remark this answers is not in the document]
i-rinat: As far as I can tell, `SKBTX_DEV_ZEROCOPY` is set when MSG_ZEROCOPY is used:
```c
static inline struct ubuf_info *skb_zcopy(struct sk_buff *skb)
{
        bool is_zcopy = skb && skb_shinfo(skb)->tx_flags & SKBTX_DEV_ZEROCOPY;
        
        return is_zcopy ? skb_uarg(skb) : NULL;
}
```
```c
#define SKBTX_ZEROCOPY_FRAG     (SKBTX_DEV_ZEROCOPY | SKBTX_SHARED_FRAG)
```
So, maybe we should check for `SKBTX_ZEROCOPY_FRAG` rather than only for `SKBTX_SHARED_FRAG`?

tempesta_fw/tls.c
avbelov23: @ikoveshnikov, how do I know where the body is the message, and where are the headers in the message? In my opinion, if there is a flag SKBTX_SHARED_FRAG you need to copy.

avbelov23: All answers in Rinat's comments.

i-rinat: But why do you care where the headers or the body are? Isn't it enough to mark body data skb's coming from the cache as shared?

avbelov23: >So can we determine which fragments contain shared data by checking page_ref_count(skb_frag_page(f))?

I first thought that the answer to the question is yes. But the answer is no

avbelov23: > we can copy cached response body
This is will copy, but we wish w/o copy.

avbelov23:
DONE

tempesta_fw/cache.c
```diff
@@ -1579,6 +1579,9 @@ tfw_cache_build_resp(TfwHttpReq *req, TfwCacheEntry *ce)
 	if (tfw_cache_build_resp_body(db, resp, trec, &it, p))
 		goto err;
 
+	if (TFW_CONN_TLS(req->conn))
```
ikoveshnikov: This should happen right after the response object is allocated in `tfw_cache_build_resp()` function, i.e. after `tfw_http_msg_setup()`.  Please add a comment here, why the flag is added and which problem it solves.

And we have problem here, only very  first skb is marked with the flag, all the following skbs in the message won't have the flag. Bit in the same time huge responses may require more than one skb. Seems like we should care of this case in `tfw_msg_iter_append_skb()` function.

avbelov23: DONE

tempesta_fw/tls.c
```diff
@@ -379,23 +388,91 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
...
-	for (next = skb, frags = 0; ; ) {
+	for (next = skb, frags = 0, out_frags = 0, elt = 0; ; ) {
+		unsigned int head_data_len = skb_headlen(next);
 		/*
 		 * skb data and tails are already adjusted above,
 		 * so use zero offset and skb->len.
 		 */
 		r = skb_to_sgvec(next, sgt.sgl + frags, 0, next->len);
 
+		if (skb_shinfo(next)->tx_flags & SKBTX_ZEROCOPY_FRAG) {
```
i-rinat: Now I think I was wrong for suggesting testing for `SKBTX_ZEROCOPY_FRAG`.

It's still right in a sense that data is shared with something, but both `SKBTX_SHARED_FRAG` and `SKBTX_DEV_ZEROCOPY` bits set means fragments are **user pages**, and they need to be mapped before use. We should revert to `SKBTX_SHARED_FRAG` here, otherwise we may fail badly on any code that uses `MSG_ZEROCOPY`.

There is an example in `skb_copy_ubufs()` of how `kmap_atomic()`/`kunmap_atomic()` are used to get to the data. The function also uses newly-allocated pages efficiently, to the last byte. Maybe we should also consider using similar technique to save memory.

tempesta_fw/cache.c
```diff
@@ -1538,6 +1538,14 @@ tfw_cache_build_resp(TfwHttpReq *req, TfwCacheEntry *ce)
 	if (tfw_http_msg_setup((TfwHttpMsg *)resp, &it, ce->hdr_len + 2))
 		goto free;
 
+	/*
+	 * Mark skb as SKBTX_SHARED_FRAG so that when sending it to the client
+	 * via https, encrypt to a separate buffer and not corrupt the data in
+	 * the cache
+	 */
+	if (TFW_CONN_TLS(req->conn))
```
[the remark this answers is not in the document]
i-rinat: We didn't see any issues with above mentioned. Anyway, it's not high-priority, so may wait.

tempesta_fw/tls.c
```diff
@@ -379,23 +388,92 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
...
+			T_WARN("cannot alloc TLS encryption scatter list.\n");
+			r = -ENOMEM;
+			goto err_out_sgt_alloc;
+		}
+		pages = kmalloc(sizeof(struct page *) * out_sgt.nents,
+				GFP_ATOMIC);
+		if (!pages) {
+			T_WARN("cannot alloc memory.\n");
+			r = -ENOMEM;
+			goto err_pages_alloc;
```
krizhanovsky: Dynamic allocation is relatively heavy operation and I see no reason to call `kmalloc()` 3 times - you free all the allocations at the same time.

tempesta_fw/tls.c
```diff
@@ -379,23 +388,92 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
...
+			if (out_frags > 0)
+				sg_mark_end(&out_sgt.sgl[out_frags - 1]);
+			skb_shinfo(next)->tx_flags &= ~SKBTX_SHARED_FRAG;
+		} else {
+			r = skb_to_sgvec(next, out_sgt.sgl + out_frags, 0,
+				         next->len);
+			if (r <= 0)
+				goto out;
+			out_frags += r;
+		}
```
krizhanovsky: `tfw_tls_encrypt()` is already overloaded function and the code just prepares memory for encrypted data, so it could make sense to move it to separate function in `ss_skb`. However, I'm not sure how many local variables are used - if the function must accept more than standard 6 arguments, them let's leave with the current code.

tempesta_fw/tls.c
```diff
@@ -420,15 +499,25 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
...
 	io->msgtype = type;
-	if (!(r = ttls_encrypt(tls, &sgt)))
+	if (!(r = ttls_encrypt(tls, &sgt, &out_sgt)))
 		ttls_aad2hdriv(xfrm, skb->data);
 
 	spin_unlock(&tls->lock);
 
+	for (i = 0; i < elt; ++i) {
+		put_page(pages[i]);
+	}
```
krizhanovsky: Do not need braces for single line code blocks

PR #1266, opened 2019-06-05
vhost selection by sticky sessions

i-rinat:
The patchset adds `vhost` caching in HTTP sessions. We now save `vhost` after the first request in a session is processed, and then reuse it for other requests in the session, thus avoiding the selection procedure which may be quite costly.

(fixes https://github.com/tempesta-tech/tempesta/issues/1043)

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -3171,9 +3171,6 @@ next_msg:
 	 * client may use just a few cookie variants. Even if different
 	 * sticky cookies are used for each vhost, the sticky module should
 	 * be faster than tfw_http_tbl_vhost().
-	 *
-	 * TODO: #1043 sticky cookie module must assign correct vhost when it
-	 * returns TFW_HTTP_SESS_SUCCESS or TFW_HTTP_SESS_REDIRECT_NEED;
 	 */
 	switch (tfw_http_sess_obtain(req))
 	{
@@ -3216,8 +3213,10 @@ next_msg:
 	 * In the same time location may differ between requests, so the
 	 * sticky module can't fill it.
 	 */
-	if (!req->vhost)
+	if (!req->vhost) {
 		req->vhost = tfw_http_tbl_vhost((TfwMsg *)req, &block);
+		tfw_http_sess_pin_vhost(req->sess, req->vhost);
+	}
 	if (req->vhost)
 		req->location = tfw_location_match(req->vhost,
 						   &req->uri_path);
--- a/tempesta_fw/http_sess.c
+++ b/tempesta_fw/http_sess.c
@@ -817,15 +817,15 @@ tfw_http_sess_resp_process(TfwHttpResp *resp)
  * current configuration.
  */
 static void
-tfw_http_sess_unpin_srv(TfwStickyConn *st_conn)
+tfw_http_sess_unpin_srv(TfwHttpSess *sess)
 {
 	TfwServer *srv;
 
-	if (!st_conn->srv_conn)
+	if (!sess->srv_conn)
 		return;
 
-	srv = (TfwServer *)st_conn->srv_conn->peer;
-	st_conn->srv_conn = NULL;
+	srv = (TfwServer *)sess->srv_conn->peer;
+	sess->srv_conn = NULL;
 	tfw_server_unpin_sess(srv);
 }
 
@@ -837,7 +837,7 @@ tfw_http_sess_put(TfwHttpSess *sess)
 		 * Use counter reached 0, so session already expired and evicted
 		 * from the hash table.
 		 */
-		tfw_http_sess_unpin_srv(&sess->st_conn);
+		tfw_http_sess_unpin_srv(sess);
 		kmem_cache_free(sess_cache, sess);
 	}
 }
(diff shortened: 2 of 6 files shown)


tempesta_fw/http.c
```diff
@@ -3353,6 +3353,20 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
 		TFW_INC_STAT_BH(clnt.msgs_otherr);
 	}
 
+	if (req->sess) {
```
ikoveshnikov: `req` must not be accessed after `tfw_http_send_resp()|tfw_cache_process()`. Request can be sent out there or even destroyed.

Saving `vhost` value should take place right after vhost is determined for the first time (same function, line 3219):
```diff
-	if (!req->vhost)
+	if (!req->vhost) {
		req->vhost = tfw_http_tbl_vhost((TfwMsg *)req, &block);
+		tfw_http_sess_pin_vhost(req);
+	}
	if (req->vhost)
		req->location = tfw_location_match(req->vhost,
						   &req->uri_path);
```

And I would like to move session-specific operations to `http_sess.c_

i-rinat: Moved code to `http_sess.c`. Since `tfw_http_sess_pin_vhost()` is now part of API for sessions and can be used by some other code, it now also checks whenever `sess->vhost` was non-NULL to track references to `vhost` more accurately. Also removed outdated comments that I forgot to remove in the initial submission.

i-rinat:
Changed code to address review comments. However there is still functional tests failure — investigating that.

i-rinat:
There was a bug in the previous version of the patchset: reference to a vhost was stored in a session, which through indirection held references to server groups. After configuration reload that references were keeping server group structures in zombie state — not dead, but also not functional, since schedulers were removed from them. That caused `502` responses emitted until client closed their connection.

The issue was resolved by introducing a flag into TfwVhost. It's set during reconfiguration (in `tfw_cfgop_vhosts_list_free()`) and checked when we are about to reuse cached vhost in `tfw_http_sess_obtain()`.

i-rinat:
> thus `vhost->main_sg` becomes invalid

@ikoveshnikov,
As far as I understand the code, `TfwVhost` holds a reference to `main_sg`, indirectly, through instances of `TfwLocation`. So if `TfwVhost` instance is alive, so are dependent instances of `TfwSrvGroup`. That's why I think this is not an issue.

What may happen is zombie `TfwSrvGroup`, which holds memory, but has `->sched_data` cleared. It's not functional, but still safe to access, so no crashes, but misbehavior instead. I believe that issue was addressed in recent changes: by checking if `TfwVhost` instance was abandoned (`TFW_VHOST_B_INVALID` bit in flags).

tempesta_fw/http_sess.c
```diff
@@ -957,8 +960,37 @@ tfw_http_sess_obtain(TfwHttpReq *req)
...
+				 * configuration anymore and therefore is not
+				 * useful to us at all. We can only release it
+				 * to free associated resources.
+				 */
+				TfwVhost *cached_vhost = sess->vhost;
+				read_unlock(&sess->lock);
+				write_lock(&sess->lock);
+				sess->vhost = NULL;
+				write_unlock(&sess->lock);
+				tfw_vhost_put(cached_vhost);
```
ikoveshnikov: Seems like suspicious optimisation. Same HTTP session can appear in multiple requests that can be processed on multiple CPUs at the same time. Imagine the situation: two threads are taking the read lock and find out that the vhost is removed; then both decides to drop the reference; since the reference is cached during read lock the reference would be dropped twice, causing use-after-free errors. I suggest following change:
```diff
-				TfwVhost *cached_vhost = sess->vhost;
				read_unlock(&sess->lock);
				write_lock(&sess->lock);
-				sess->vhost = NULL;
+				if (sess->vhost) {
+					tfw_vhost_put(sess->vhost);
+					sess->vhost = NULL;
+				}
				write_unlock(&sess->lock);
-				tfw_vhost_put(cached_vhost);
```

tempesta_fw/http_sess.c
```diff
@@ -957,8 +960,37 @@ tfw_http_sess_obtain(TfwHttpReq *req)
...
+				goto found;
+			}
+
+			tfw_vhost_get(sess->vhost);
+			req->vhost = sess->vhost;
+			read_unlock(&sess->lock);
 			goto found;
```
ikoveshnikov: Something seemed strange to me here, I was thinking about possible configuration updates like scheduler changes. vhost name changes and other changes that can spoil moving session from old vhost instance to a new one. I took me some time to recheck this, but everything looks good here.

tempesta_fw/vhost.h
```diff
@@ -128,6 +128,10 @@ enum {
 	TFW_D_CACHE_PURGE_INVALIDATE,
 };
 
+enum {
+	TFW_VHOST_B_INVALID = 0,
```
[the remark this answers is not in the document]
i-rinat: Changed to `TFW_VHOST_B_REMOVED`. It does look better, indeed.

tempesta_fw/http_sess.c
i-rinat: Changed code according to your suggestion. Thanks!

PR #1268, opened 2019-06-06
cleanup tfw_addr_ifmatch() tfw_addr_eq()

### LINKED ISSUE #774: [cfg] Make sure Tempesta doesn't specify itself as a backend
A backend server (an IP address and port number) specified in Tempesta configuration may in fact turn out to be the same as an address where Tempesta listens to incoming requests from clients. The mistake is very easy to make. There should be a protection against that.

The problem was recognized before and seemingly fixed in 83c987fb. However, the fix has issues of its own.
* A listening address and port and  backend address and port may be specified in different forms. One may be specified as an IPv4 address, while the other may be specified as an IPv6 address. The function `tfw_addr_ifmatch()` that compares the addresses doesn't seem to handle this case. See issue #775.
* The check to see if the addresses of backend servers and Tempesta listeners match is performed after the configuration is processed by each module. However, the implementation details of configuration processing are that servers specified outside of any group are grouped together in the group named `default`, and that is done after complete configuration has been processed. So at the time the check is performed these servers are not in the configuration yet, so they are missed by the check.

The listener's module is called and initialized before the module that handles backends. Yet the check is implemented in the listener's module. Perhaps it makes more sense to move it to the module for backends.

kevgs:
Remove misleading TODO: v6 case is alredy the same as v4 one.

Replace direct sockaddr_in::sin6_port access with accessor
function tfw_addr_port().

### DIFF
--- a/tempesta_fw/addr.c
+++ b/tempesta_fw/addr.c
@@ -406,8 +406,6 @@ tfw_addr_ifmatch(const TfwAddr *server, const TfwAddr *listener)
 			/* backend = [::1] */
 			return 1;
 		}
-
-		/* TODO: same as in v4 case */
 	}
 
 	return tfw_addr_eq(server, listener);


tempesta_fw/addr.c
```diff
@@ -382,9 +382,6 @@ tfw_addr_ifmatch(const TfwAddr *server, const TfwAddr *listener)
 		return 0;
 	}
 
-	if (tfw_addr_port(listener) != tfw_addr_port(server))
```
i-rinat: This comparison is needed for the case where the code below returns `1`. If listening address is `0.0.0.0` and server address is in loopback subnet, we need to be sure that ports match. And only then it's a closed loop. Otherwise it's fine: we may serve at 0.0.0.0:80 while having Nginx at 127.0.0.1:81 as a backend.

kevgs: You're right. Thanks!

i-rinat:
I still don't feel comfortable with the statement that this PR close #774, since it doesn't. The issue was solved already, and this patch is a cleanup, not a fix.

kevgs:
Right, this is just cleanup. Commit message is fixed.

i-rinat:
I've removed from the patch calls to `tfw_addr_port()`, as we are already in the file implementing `tfw_addr_*`, next condition expression explicitly uses IPv6 parts of the address, so there is no sense in hiding implementation inside the implementation.

PR #1284, opened 2019-06-30
guard against NULL deref in tfw_client_for_each

i-rinat:
May happen after initialization failure. To reproduce, add `return -EINVAL` to `tfw_filter_start()` before `tdb_open()` call.

Upd. During offline discussion it was decided to add guards to `tfw_mods_start()` and `tfw_mods_end()` too.

### DIFF
--- a/tempesta_fw/client.c
+++ b/tempesta_fw/client.c
@@ -253,6 +253,8 @@ EXPORT_SYMBOL(tfw_client_obtain);
 int
 tfw_client_for_each(int (*fn)(void *))
 {
+	if (!client_db)
+		return 0;
 	return tdb_entry_walk(client_db, fn);
 }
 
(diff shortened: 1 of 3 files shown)


i-rinat:
> Isn't it better to fix `tfw_init()`?

Could you please provide some details? I don't think I understand.

As far as I can see, `tfw_init()` is about registering modules, not really about calling their "start" or "stop" handlers. The issue happens when we start Tempesta, but fail during the process, in `tfw_client_start()`. Then we call "stop" handler of every module. So that at least requires that every module's "stop" handler must check that "start" was actually called and succeeded. They do so, but in `tfw_sock_clnt_stop()` function from another module is called: `tfw_client_for_each()`. It doesn't check if initialization of module "client" was successful, as modules care about themselves only. That leads to a NULL dereference.

So the only solution I can think of, is to add a NULL check to the `tfw_client_for_each()`.

PR #1300, opened 2019-07-09
tls: fix decryption of a record spanning over multiple skbs

### LINKED ISSUE #1299: POST request with a body crashes TLS
(current version: e148881689fc72e300bd7f6d97b3e5b2f70ed75a)

POST requests with a moderate size body cause kernel panic. Here is the script to generate such request:
```python
import socket, ssl
# -------------------------------
target = ('192.168.122.152', 443)
payload_length = 9001
# -------------------------------
context = ssl.SSLContext(ssl.PROTOCOL_TLS)
context.load_default_certs()
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
ssl_sock = context.wrap_socket(s)
ssl_sock.connect(target)

req = ('POST / HTTP/1.1\r\n' +
      'Host: localhost\r\n' +
      'Content-Length: ' + str(payload_length) + '\r\n' +
      'Connection: keep-alive\r\n' +
      '\r\n' +
      ('x' * payload_length))

print('req length =     {}'.format(len(req)))
print('payload length = {}'.format(payload_length))

r = ssl_sock.write(req)
print(r)

while True:
    data = ssl_sock.read(100000)
    if not data:
        break
    print(data)
```

I have Tempesta running in VM, listening at 192.168.122.152:443. Script is called from the host. VM uses virtio networking. The crash does not appear when script is called from inside VM.

Here is the panic from the log:
```
[ 1903.751665] BUG: unable to handle kernel NULL pointer dereference at 0000000000000008
[ 1903.752604] IP: scatterwalk_copychunks+0x10a/0x180
[ 1903.752604] PGD 0 P4D 0 
[ 1903.752604] Oops: 0000 [#1] SMP PTI
[ 1903.752604] Modules linked in: tempesta_fw(O) tempesta_db(O) tempesta_tls(O) tempesta_lib(O) sha256_mb mcryptd sha256_ssse3 sha512_ssse3 sha512_generic crc32_pclmul crc32c_intel ghash_clmulni_intel pcbc aesni_intel crypto_simd aes_x86_64 cryptd glue_helper virtio_balloon virtio_console virtio_net virtio_pci virtio_ring virtio [last unloaded: tempesta_lib]
[ 1903.752604] CPU: 2 PID: 0 Comm: swapper/2 Tainted: G        W  O    4.14.32-tfw #0.0.0
[ 1903.752604] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS 1.12.0-1 04/01/2014
[ 1903.752604] task: ffffa2eedab3aac0 task.stack: ffffbd1b40688000
[ 1903.752604] RIP: 0010:scatterwalk_copychunks+0x10a/0x180
[ 1903.752604] RSP: 0018:ffffa2eeffd038b8 EFLAGS: 00010246
[ 1903.752604] RAX: 0000000000000000 RBX: 0000000000000750 RCX: 0000000000007696
[ 1903.752604] RDX: 0000000000000696 RSI: ffffa2eeb8e7f696 RDI: ffffa2eed89bb098
[ 1903.752604] RBP: ffffa2eeffd038f0 R08: 0000000000007696 R09: fffffffffffffffa
[ 1903.752604] R10: ffffa2eeffefab50 R11: 0000000000000002 R12: ffffa2eeb8e69c48
[ 1903.752604] R13: 0000000000000000 R14: ffffa2eedab3aac0 R15: ffffa2eedab3aac0
[ 1903.752604] FS:  0000000000000000(0000) GS:ffffa2eeffd00000(0000) knlGS:0000000000000000
[ 1903.752604] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[ 1903.752604] CR2: 0000000000000008 CR3: 000000005520a001 CR4: 00000000003606e0
[ 1903.752604] DR0: 0000000000000000 DR1: 0000000000000000 DR2: 0000000000000000
[ 1903.752604] DR3: 0000000000000000 DR6: 00000000fffe0ff0 DR7: 0000000000000400
[ 1903.752604] Call Trace:
[ 1903.752604]  <IRQ>
[ 1903.752604]  scatterwalk_map_and_copy+0x39/0x50
[ 1903.752604]  ? kmalloc_order+0x4b/0x60
[ 1903.752604]  ? kmalloc_order_trace+0x18/0xa0
[ 1903.752604]  gcmaes_decrypt.constprop.13+0x83/0x270 [aesni_intel]
[ 1903.752604]  ? generic_gcmaes_decrypt+0x4b/0x60 [aesni_intel]
[ 1903.752604]  ? ttls_decrypt+0x2cb/0x520 [tempesta_tls]
[ 1903.752604]  ? helper_rfc4106_decrypt+0xb0/0xb0 [aesni_intel]
[ 1903.752604]  ? ttls_decrypt+0x2cb/0x520 [tempesta_tls]
[ 1903.752604]  ? ttls_recv+0xe5/0x790 [tempesta_tls]
[ 1903.752604]  ? ttls_decrypt+0x520/0x520 [tempesta_tls]
[ 1903.752604]  ? ss_skb_process+0x164/0x1c0 [tempesta_fw]
[ 1903.752604]  ? tfw_tls_msg_process+0xd3/0x310 [tempesta_fw]
[ 1903.752604]  ? tfw_tls_msg_process+0x5a/0x310 [tempesta_fw]
[ 1903.752604]  ? __gfsm_fsm_exec+0x51/0x90 [tempesta_fw]
[ 1903.752604]  ? ip_output+0x59/0xc0
[ 1903.752604]  ? tfw_connection_recv+0x39/0x40 [tempesta_fw]
[ 1903.752604]  ? tfw_connection_send+0x20/0x20 [tempesta_fw]
[ 1903.752604]  ? ss_tcp_process_data+0x1c0/0x410 [tempesta_fw]
[ 1903.752604]  ? ss_tcp_data_ready+0x3e/0x80 [tempesta_fw]
[ 1903.752604]  ? tcp_rcv_established+0x4a6/0x540
[ 1903.752604]  ? tcp_v4_do_rcv+0x124/0x1c0
[ 1903.752604]  ? tcp_v4_rcv+0x8c1/0x9b0
[ 1903.752604]  ? ip_local_deliver_finish+0x6b/0x190
[ 1903.752604]  ? ip_local_deliver+0x56/0xc0
[ 1903.752604]  ? ip_rcv_finish+0x340/0x340
[ 1903.752604]  ? ip_rcv+0x24f/0x370
[ 1903.752604]  ? inet_del_offload+0x30/0x30
[ 1903.752604]  ? __netif_receive_skb_core+0x318/0xb00
[ 1903.752604]  ? tcp_write_xmit+0x29c/0x1190
[ 1903.752604]  ? virtqueue_napi_complete+0x34/0x60 [virtio_net]
[ 1903.752604]  ? process_backlog+0x92/0x120
[ 1903.752604]  ? process_backlog+0x92/0x120
[ 1903.752604]  ? net_rx_action+0x23d/0x380
[ 1903.752604]  ? __do_softirq+0x104/0x297
[ 1903.752604]  ? irq_exit+0xa3/0xb0
[ 1903.752604]  ? do_IRQ+0x45/0xc0
[ 1903.752604]  ? common_interrupt+0x7a/0x7a
[ 1903.752604]  </IRQ>
[ 1903.752604]  ? __sched_text_end+0x6/0x6
[ 1903.752604]  ? default_idle+0x12/0xf0
[ 1903.752604]  ? do_idle+0x156/0x1b0
[ 1903.752604]  ? cpu_startup_entry+0x14/0x20
[ 1903.752604]  ? secondary_startup_64+0xa5/0xb0
[ 1903.752604] Code: 4d 08 74 66 48 8b 45 00 49 01 d4 48 29 d3 44 8b 40 0c 44 03 40 08 44 39 c1 0f 82 32 ff ff ff 48 89 c7 e8 5a 8d 04 00 48 89 45 00 <8b> 48 08 89 4d 08 44 8b 40 0c 44 03 40 08 e9 13 ff ff ff 48 8b 
[ 1903.752604] RIP: scatterwalk_copychunks+0x10a/0x180 RSP: ffffa2eeffd038b8
[ 1903.752604] CR2: 0000000000000008
[ 1903.752604] ---[ end trace a43ddfe0310e5787 ]---
[ 1903.752604] Kernel panic - not syncing: Fatal exception in interrupt
[ 1903.752604] Kernel Offset: 0x1a00000 from 0xffffffff81000000 (relocation range: 0xffffffff80000000-0xffffffffbfffffff)
[ 1903.752604] ---[ end Kernel panic - not syncing: Fatal exception in interrup
```

i-rinat:
There was a couple of issues with parsing medium-sized requests.

First was about missing `sg_unmark_end()` after `skb_to_sgvec()`. Latter calls `sg_mark_end()`, so since we want to pack multiple skb's into a single scatter-list, we must undo the changes.

Second was about cutting TLS-specific data from skb's. We cut them, but only after a whole request is parsed. If a request spans over multiple skb's, information about data offset is lost, so we ended up with passing TLS auxiliary data to a backend. We need to cut data from the head while offset is still known.

Fixes #1299.

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -3103,6 +3103,8 @@ next_msg:
 		tfw_http_req_parse_drop(req, 400, "failed to parse request");
 		return TFW_BLOCK;
 	case TFW_POSTPONE:
+		if (tfw_http_chop_skb((TfwHttpMsg *)req, skb, off, curr_skb_trail))
+			return TFW_BLOCK;
 		r = tfw_gfsm_move(&conn->state, TFW_HTTP_FSM_REQ_CHUNK,
 				  &data_up);
 		TFW_DBG3("TFW_HTTP_FSM_REQ_CHUNK return code %d\n", r);
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -164,6 +164,7 @@ ttls_crypto_req_sglist(TlsCtx *tls, struct crypto_aead *tfm, unsigned int len,
 			n = skb_to_sgvec(skb, sg_i, off, to_read);
 			if (n <= 0)
 				goto err;
+			sg_unmark_end(sg_i + n - 1);
 			T_DBG3_SL("build req sglist", sg_i, n, 0, (size_t)len);
 			len -= to_read;
 			sg_i += n;


tempesta_fw/http.c
```diff
@@ -3103,6 +3103,9 @@ tfw_http_req_process(TfwConn *conn, const TfwFsmData *data)
 		tfw_http_req_parse_drop(req, 400, "failed to parse request");
 		return TFW_BLOCK;
 	case TFW_POSTPONE:
+		if (off > 0)
+			if (tfw_http_chop_skb((TfwHttpMsg *)req, skb, off, 0))
+				return TFW_BLOCK;
```
aleksostapenko: Good catch!

aleksostapenko: Shouldn't we try to chop `curr_skb_trail` here too?

ikoveshnikov: Agree, last argument must be `curr_skb_trail`. Also the condition `if (off > 0)` is not really required,  `tfw_http_chop_skb()` perfectly handles the case, when both `off` and `trail` are `0`.

tls/ttls.c
```diff
@@ -164,6 +164,7 @@ ttls_crypto_req_sglist(TlsCtx *tls, struct crypto_aead *tfm, unsigned int len,
 			n = skb_to_sgvec(skb, sg_i, off, to_read);
 			if (n <= 0)
 				goto err;
+			sg_unmark_end(sg_i + n - 1);
```
krizhanovsky: The statement doesn't seem correct. If n is the same as previous and we're going to exit on current iteration from the loop, then we lose just set mark end by `skb_to_sgvec()`. I think `sg_unmark_end()` must be just before `skb_to_sgvec()` call to unset previously set end mark (either from `sg_init_table()` or previous iteration) and let `skb_to_sgvec()` define the last (current) end mark.

i-rinat: I believe this is not true, and the proposed version is fine. As there is an explicit call of  `sg_mark_end(*sg + *sgn - 1)` just outside the loop, we will never fail to put the end mark.

Although it's possible to remove that `sg_mark_end()` and to call `sg_unmark_end()` only when it's absolutely necessary, it will make code look wrong at the first glance as there will be no explicit `sg_mark_end()`. Then, reader will realize that it's `skb_to_sgvec()` that sets the mark. And it will gain nothing, as the saved load-store in (inlined) `sg_unmark_end()` will be eaten by the tracking logic — one must avoid calling `sg_unmark_end()` on the first iteration. I'd like propose to leave the calls as they are now.

Still, if you prefer to have `skb_to_sgvec()` setting the last mark, I'll change the code to do that.

tempesta_fw/http.c
i-rinat: Thanks.
Changed the code to use `curr_skb_trail` and to call `tfw_http_chop_skb()` unconditionally.

PR #1306, opened 2019-07-14
Verify extension length before accessing the data

krizhanovsky:


tls/tls_srv.c
```diff
@@ -62,7 +68,9 @@ ttls_parse_servername_ext(TlsCtx *tls, const unsigned char *buf, size_t len)
 	while (servername_list_size > 0) {
 		hostname_len = ((p[1] << 8) | p[2]);
 		if (hostname_len + 3 > servername_list_size) {
-			T_DBG("ClientHello: bad hostname size\n");
+			T_DBG("ClientHello: bad hostname size"
+			      " (%lu, expected not more %lu)\n",
+			      hostname_len, servername_list_size + 3);
```
i-rinat:
```suggestion
			      hostname_len, max_t(size_t, servername_list_size, 3) - 3);
```
we can't just subtract 3 because the variables are unsigned.

tls/tls_srv.c
```diff
@@ -325,7 +345,7 @@ ttls_parse_alpn_ext(TlsCtx *tls, const unsigned char *buf, size_t len)
 	 */
```
i-rinat: I think `ttls_parse_supported_point_formats()` needs to be updated too. We dereference `buf` there as `buf[0]` without checking available length first. Also, there is a similar `buf[0]` access in `ssl_parse_supported_point_formats_ext()`.

PR #1313, opened 2019-07-19
frang: ignore responses if request was originated by HealthMonitor

### LINKED ISSUE #1312: Frang: null pointer dereference, if HM enabled
When Health Monitor is enabled, it generates requests on Tempesta side and forwards them to backend. When a corresponding response is received, it's validated across Frang limits. Frang tries to access data behind `resp->req->conn` pointer, but the pointer is `NULL` for locally generated requests and responses. Thus a null dereference happens.

The issue exists for a long time, but was spotted on the CI only recently on multiple branches.

ikoveshnikov:
Health Monitor requests are generated by Tempesta itself thus
`req->conn` is `NULL` for them.
Fix #1312


tempesta_fw/http_limits.c
```diff
@@ -1048,7 +1069,12 @@ frang_resp_fwd_process(TfwHttpResp *resp)
 	TfwHttpReq *req = resp->req;
 	__FRANG_CFG_VAR(conf, http_resp_code_block);
 
-	if (!conf)
+	/*
+	 * Request originated by Health Monitor are generated by Tempesta,
```
avbelov23: `The request` and `is generated`

tempesta_fw/http_limits.c
```diff
@@ -1048,7 +1069,12 @@ frang_resp_fwd_process(TfwHttpResp *resp)
...
+	 * there is no reason to limit them. Such requests has no connection
```
avbelov23: `Such requests don't have a connection`?

Issue #1314, opened 2019-07-19
BUG at stopping tempesta if port is busy


avbelov23:
```
[  102.495511] kernel BUG at /home/avb/Projects/tempesta/tempesta_fw/sock_clnt.c:381!
[  102.496675] invalid opcode: 0000 [#1] SMP PTI
[  102.497353] Modules linked in: tempesta_fw(O) tempesta_db(O) mcryptd sha256_ssse3 sha512_ssse3 sha512_generic tempesta_tls(O) tempesta_lib(O) crc32_pclmul crc32c_intel ghash_clmulni_intel pcbc aesni_intel aes_x86_64 crypto_simd glue_helper cryptd vboxvideo(O) vboxguest(O) ttm
[  102.501353] CPU: 0 PID: 2787 Comm: sysctl Tainted: G           O    4.14.32+ #1
[  102.502640] Hardware name: innotek GmbH VirtualBox/VirtualBox, BIOS VirtualBox 12/01/2006
[  102.504098] task: ffff892dd4999b80 task.stack: ffff9e1280830000
[  102.505028] RIP: 0010:tfw_cfgop_cleanup_sock_clnt+0x4b/0x70 [tempesta_fw]
[  102.506165] RSP: 0000:ffff9e1280833d48 EFLAGS: 00010282
[  102.507025] RAX: ffff892dc6a91200 RBX: ffffffffc03c80b0 RCX: ffffffffc03c80d0
[  102.508088] RDX: ffffffffc03953c0 RSI: 0000000000000000 RDI: ffff892dc6a911e0
[  102.509258] RBP: 0000000000000000 R08: 0000000000000000 R09: 0000000000000000
[  102.510461] R10: 000000000061ec00 R11: ffff892ddfc20a20 R12: ffff9e1280833e20
[  102.511490] R13: ffffffffc03c7d60 R14: ffffffffc03c7fc0 R15: 00000000ffffff9e
[  102.512652] FS:  00007f10deef8880(0000) GS:ffff892ddfc00000(0000) knlGS:0000000000000000
[  102.513873] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[  102.514780] CR2: 00007f3e94f1b018 CR3: 00000000a2d66002 CR4: 00000000000606f0
[  102.515824] Call Trace:
[  102.516164]  spec_cleanup+0x54/0x70 [tempesta_fw]
[...]

PR #1315, opened 2019-07-19
Release listen socket if error was

avbelov23:
https://github.com/tempesta-tech/tempesta/issues/1314

### DIFF
--- a/tempesta_fw/sock_clnt.c
+++ b/tempesta_fw/sock_clnt.c
@@ -425,7 +425,7 @@ tfw_listen_sock_start(TfwListenSock *ls)
 	r = ss_bind(sk, addr);
 	if (r) {
 		TFW_ERR_ADDR("can't bind to", addr, TFW_WITH_PORT);
-		return r;
+		goto err;
 	}
 
 	TFW_DBG("start listening on socket: sk=%p\n", sk);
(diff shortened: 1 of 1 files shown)


ikoveshnikov:
> Some modules do clean up if startup fails (tfw_filter_start), some do not (tfw_sock_srv_start).

Yeah, thats a real mess. The `TfwMod` description must contain description of the of the module contract: when the cleanup functions are called and when - not. Since now only properly started modules are stopped, some doesn't invoke cleanup and the #1314 issue is actually deeper than it seemed.

> But I'd prefer not doing that — to keep the clear semantics: one needs to stop only that what was properly started. 

I don't see there real difference here. Anyway, clean up on the failed start is mostly the same as the `stop()` function.

i-rinat:
> I don't see there real difference here. Anyway, clean up on the failed start is mostly the same as the stop() function.

Yeah, if there were a significant differences I definitely would've described them. Instead, it's just a personal preference. I like it when APIs move complexity out of my way, thus reducing a number of things I have to worry about. 

> The TfwMod description must contain description of the of the module contract: when the cleanup functions are called and when - not.

I insist on not doing that. We should select one way and change remaining code to follow that. There is no benefit in having that flexibility, as it's an internal API, and we have the freedom to do as we please there.

ikoveshnikov:
@avbelov23 Have you looked to other modules? do we have the same issue anywhere else?

avbelov23:
What is the issue?

>Some modules do clean up if startup fails (tfw_filter_start), some do not (tfw_sock_srv_start).
Is this issue?

ikoveshnikov:
Yes, that is.

avbelov23:
I propose to fix the issues in other modules with separate PR

avbelov23:
DONE

avbelov23:
There is no issue in `sock_srv`. The release of resources occurs during the removal of servers from groups

PR #1316, opened 2019-07-19
tls: release spinlock before invoking gfsm move

i-rinat:
Invoking the GFSM may cause other code to execute. If that code tries to send an immediate response, for example, on request parsing error, it will try to acquire `tls->lock` while it's held in `tfw_tls_msg_process()`.

In fact that the way to reproduce the bug: send a malformed HTTPS request. Each such requests causes a deadlock. `N` such requests deadlock whole machine given it has `N` cores.


i-rinat:
> It's better to acquire it below next_msg, remove out_err (there is only one goto there), unlock the lock inside the IF statement, and remove the spin_unlock() at the end of the function.

Changed the patch according to the suggestions.

Issue #1283, opened 2019-06-29
Decryption failure on 16380-byte messages


krizhanovsky:
Use following patch for `test_tls_integrity.py` (introduced in https://github.com/tempesta-tech/tempesta-test/pull/103/commits/519f4d8144e223c3b64927f232d26a0e7b4b8604) just to run single test (all previous tests pass):
```diff
--- a/tls/test_tls_integrity.py
+++ b/tls/test_tls_integrity.py
@@ -148,12 +148,12 @@ class Cache(TlsIntegrityTester):
 
     def test_various_req_resp_sizes(self):
         self.start_all()
-        self.common_check(1, 1)
-        self.common_check(19, 19)
-        self.common_check(567, 567)
-        self.common_check(1755, 1755)
-        self.common_check(4096, 4096)
+        #self.common_check(1, 1)
+        #self.common_check(19, 19)
+        #self.common_check(567, 567)
+        #self.common_check(1755, 1755)
+        #self.common_check(4096, 4096)
         self.common_check(16380, 16380)
-        self.common_check(65536, 65536)
-        self.common_check(1000000, 1000000)
+        #self.common_check(65536, 65536)
+        #self.common_check(1000000, 1000000)
```

The test fails with
```
======================================================================
FAIL: test_various_req_resp_sizes (tls.test_tls_integrity.Cache)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/tempesta-test/tls/test_tls_integrity.py", line 156, in test_various_req_resp_sizes
    self.common_check(16380, 16380)
  File "/root/tempesta-test/tls/test_tls_integrity.py", line 66, in common_check
    " (len=%d)" % (req_len, resp_len))
AssertionError: Cannot process request (len=16380) or response (len=16380)

----------------------------------------------------------------------
Ran 1 test in 6.333s

FAILED (failures=1)
```

I built Tempesta FW with `DEBUG=2 DBG_TLS=1 DBG_HTTP_PARSER=1` (enough verbosity for the error message and not to much to hit any timeouts). The log shows following:
```
[18598.582825] [tempesta tls]   handshake message: msglen=233 type=1 hslen=229 read=9
[18598.584577] [tempesta tls]   server state: 0
[18598.585747] [tempesta tls]   found supported point formats extension
[18598.587263] [tempesta tls]   ClientHello: point format selected: 0
[18598.588716] [tempesta tls]   found supported elliptic curves extension
[...]

PR #1322, opened 2019-07-23
tls: ciphertexts larger than 16384 bytes

i-rinat:
TLS records have plaintext size limitation, they cannot be larger than 2^14 bytes. However, ciphertext may be as large as 2^14 + 2048 bytes. But the plaintext obtained from that ciphertext should still be 2^14 bytes or smaller. We need to respect that limits precisely.

The patchset consists of two patches. One changes the limits, and other fixes a defect in the parsed bytes calculation.

Related to #1283, but doesn't fix it entirely yet.

### DIFF
--- a/tempesta_fw/ss_skb.c
+++ b/tempesta_fw/ss_skb.c
@@ -1067,6 +1067,7 @@ ss_skb_process(struct sk_buff *skb, unsigned int off, unsigned int trail,
 			n = frag_size - off;
 			if (trail > frag_len)
 				n -= trail - frag_len;
+			_processed = 0;
 			r = actor(objdata, frag_addr + off, n, &_processed);
 			*processed += _processed;
 			if (r != SS_POSTPONE || trail >= frag_len)
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -46,6 +46,19 @@ static DEFINE_PER_CPU(struct aead_request *, g_req) ____cacheline_aligned;
 static struct kmem_cache *ttls_hs_cache = NULL;
 static ttls_send_cb_t *ttls_send_cb;
 
+static inline size_t
+ttls_max_ciphertext_len(const TlsXfrm *xfrm)
+{
+	/*
+	 * Although RFC 5246 6.2.3 allows ciphertexts to be as large as
+	 * (2^14 + 2048) bytes, actual limits are specific to particular
+	 * cipher suites. We are supporting only AEAD ciphers (GCM and CCM).
+	 * Their transforms increase data size by a constant amount of bytes.
+	 * To be specific, those are explicit part of IV and a tag.
+	 */
+	return TLS_MAX_PAYLOAD_SIZE + xfrm->minlen;
+}
+
 static inline unsigned short
 ttls_msg2crypt_len(const TlsIOCtx *io, const TlsXfrm *xfrm)
 {
@@ -906,6 +919,11 @@ ttls_decrypt(TlsCtx *tls, unsigned char *buf)
 	int r;
 	TlsIOCtx *io = &tls->io_in;
 
+	if (io->msglen > ttls_max_ciphertext_len(&tls->xfrm)) {
+		T_DBG("bad message length %u\n", io->msglen);
+		return T_DROP;
+	}
+
 	if ((r = __ttls_decrypt(tls, buf))) {
 		/* Error out (and send alert) on invalid records */
 		if (r == TTLS_ERR_INVALID_MAC)
@@ -925,10 +943,6 @@ ttls_decrypt(TlsCtx *tls, unsigned char *buf)
 		return T_DROP;
 	} else {
 		tls->nb_zero = 0;
-	}
-	if (io->msglen > TLS_MAX_PAYLOAD_SIZE) {
-		T_DBG("bad message length %u\n", io->msglen);
-		return T_DROP;
 	}
 
 	if (io->msgtype == TTLS_MSG_ALERT)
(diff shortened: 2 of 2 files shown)


tls/ttls.c
```diff
@@ -46,6 +46,12 @@ static DEFINE_PER_CPU(struct aead_request *, g_req) ____cacheline_aligned;
 static struct kmem_cache *ttls_hs_cache = NULL;
 static ttls_send_cb_t *ttls_send_cb;
 
+static inline size_t
+ttls_max_ciphertext_len(const TlsXfrm *xfrm)
+{
+	return TLS_MAX_PAYLOAD_SIZE + xfrm->minlen;
```
krizhanovsky: This is crucial limit - please reference appropriate RFC in a comment. `minlen` is typically very small as being initialized in `ttls_derive_keys()` and it seems never reaches 2048 as described in the PR comment.

tls/ttls.c
```diff
@@ -1049,11 +1056,6 @@ ttls_hdr_check(TlsCtx *tls)
 		T_DBG("minor version mismatch %d\n", tls->minor);
 		return T_DROP;
 	}
-	/* Check length against the size of our buffer */
-	if (unlikely(io->msglen > TTLS_PAYLOAD_LEN)) {
-		T_DBG("too big message length: %u\n", io->msglen);
-		return T_DROP;
-	}
```
krizhanovsky: We first read `msglen` in `ttls_parse_record_hdr()` and we must check the record length as early as possible and it's now enough to check the length in `ttls_decrypt()`, e.g. it seems now we go to handshake FSM w/o the length checking.

tls/ttls.c
```diff
@@ -1064,17 +1066,14 @@ ttls_hdr_check(TlsCtx *tls)
...
 		}
 	} else {
-		/*
-		 * TLS encrypted messages can have up to 256 bytes of padding.
-		 */
-		if (io->msglen < tls->xfrm.minlen
-		    || io->msglen
-		       > tls->xfrm.minlen + TLS_MAX_PAYLOAD_SIZE + 256)
+		if (io->msglen < tls->xfrm.minlen ||
+		    io->msglen > ttls_max_ciphertext_len(&tls->xfrm))
```
krizhanovsky: `256` inherited from mbedTLS - is it wrong? If not, then we need to describe this in comment for `ttls_max_ciphertext_len()`.

tempesta_fw/ss_skb.c
```diff
@@ -1067,6 +1067,7 @@ ss_skb_process(struct sk_buff *skb, unsigned int off, unsigned int trail,
 			n = frag_size - off;
 			if (trail > frag_len)
 				n -= trail - frag_len;
+			_processed = 0;
 			r = actor(objdata, frag_addr + off, n, &_processed);
 			*processed += _processed;
```
krizhanovsky: Good catch - `*processed += _processed` really seems very wrong. But `ttls_recv()` and probably other actors use `_processed` as a counter of currently read record, so it seems wrong to zeroize it.

tls/ttls.c
i-rinat: > is it wrong

It's not wrong, it's just too forgiving. We are not supporting CBC modes, only AEAD ciphers, and those are not using padding. So there is no need to keep that additional 256 bytes in the size limit. I've added a comment to the ttls_max_ciphertext_len().

tempesta_fw/ss_skb.c
i-rinat: Yeah, that part is tricky. I'll look further into it.

tls/ttls.c
i-rinat: I've added a comment with description to the patch.

In short, there are limits on plaintext size (2^14 bytes) and on ciphertext size (2^14 + 2048). But since we use ciphers that add a constant overhead to each record, there is no way 2^14 + 2048 bytes ciphertext after deciphering will fit into 2^14 byte limit of a plaintext. So the limit is 2^14 plus the cipher overhead, which is `xfrm->minlen`.

tls/ttls.c
i-rinat: If you open the file context below, you'll see that there are no calls to `ttls_decrypt()` there. Instead, `io->msgtype` is almost immediately checked again. But with the knowledge of current state, with encryption or not, the checks are more precise. I think this particular check is redundant, as it doesn't prevent any costly calculations anyway.

tempesta_fw/ss_skb.c
i-rinat: Throughout the code `ss_skb_process()` is called with various "actor" callbacks. Some of them increment `*parsed`, some do not. Specifically:
* ttls_recv — increments;
* tfw_http_parse_req — resets to `0` immediately, and then writes a new value;
* tfw_http_parse_resp — resets to `0`, then writes a new value;
* tfw_h2_frame_recv — increments;
* tfw_bmb_print_msg — writes a new value that discards any previous value.

I haven't found any dependency on current values of `*parsed` in functions. There are some uses, but those save values to figure out relative changes of the counter after child function calls. From what I found I think it's safe to reset `_processed` before calling `actor()`. Effectively, http parsing routines do reset the counter. Only TLS-related do not.

krizhanovsky: Yes, agree. Please also note that `ttls_parse_client_hello()` doesn't even update `read` - looks like a bug. Please also check other handshake functions against `read` usage.

i-rinat: > `ttls_parse_client_hello()` doesn't even update read

There is an update hidden in `TTLS_HS_FSM_FINISH()`: `*read += p - buf;`.

i-rinat:
(rebased to the current master branch state)

Issue #1308, opened 2019-07-15
TempestaTLS doesn't send alerts


krizhanovsky:
Just run new test
```
# ./run_tests.py -n tls.test_tls_handshake.TlsHandshakeTest.test_long_ext
```
from https://github.com/tempesta-tech/tempesta-test/pull/103/commits/c16b8ebef06689bed491be910b34b9f9a0824914 which sends bad TLS extension and now correctly checks that TLS layer throws on exception on receiving alert.

Tempesta with `DEBUG=3 DBG_TLS=1 DBG_SS=1` build options shows that it does send alert:
```
[56216.369744] [tempesta tls]   ClientHello: bad extension size 1032 (ext_len=1139)
[56216.372024] [tempesta tls]   send alert level=2 message=50
[56216.373921] [tempesta tls]   write record: type=21 len=2 hslen=2 sgt=0000000000000000/0 sg_i=0
[56216.376363] [tempesta tls]       use record hdr: type=21 ver=3:3 hdr_len=2
[56216.378368] [tempesta fw]   TLS 7 bytes +0 segments (2 bytes, last msgtype 0x15) are to be sent on conn=ffff91e43ffba040/sk_write_xmit=ffffffffc03e13c0 ready=0
[56216.382500] [tempesta fw]       [1]: ss_send: sk=ffff91e4444b20c0 (cpu=1) state=Established
[56216.384904] [tempesta tls]   handshake error: -30976
[56216.386621] [tempesta fw] Warning: Unrecognized TLS receive return code -30976, drop packet
```
But tcpdump shows only zero-length FIN and ACK segments from Tempesta in response to the ClientHello with malicious SNI.
```
00:24:03.661736 IP 127.0.0.1.49336 > 127.0.0.1.443: Flags [P.], seq 1:1234, ack 1, win 342, options [nop,nop,TS val 3681251550 ecr 3681251309], length 1233
	0x0000:  4500 0505 c652 4000 4006 719e 7f00 0001  E....R@.@.q.....
	0x0010:  7f00 0001 c0b8 01bb 7228 8b0c eb05 ca0e  ........r(......
	0x0020:  8018 0156 02fa 0000 0101 080a db6b 70de  ...V.........kp.
	0x0030:  db6b 6fed 1603 0304 cc01 0004 c803 0322  .ko............"
	0x0040:  2222 2211 1111 1111 1111 1111 1111 1111  """.............
[...]

Issue #861, opened 2017-11-12
TCP: reset blocked connections


krizhanovsky:
# Scope

There is old TODO comment https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/sock.c#L495 which is still not implemented. However discussion https://github.com/tempesta-tech/tempesta/pull/855#discussion_r148932522 shows clear use case for TCP RST to close connections which faced TFW_BLOCK at some point.

Discussion https://github.com/tempesta-tech/tempesta/pull/1200/files#r289656478 revealed the problem with closing connection on blocked IP - we can not do normal close with blocked IP because we can not receive client FIN-ACK and the connection hangs. We should call `tfw_filter_block_ip()` only with connection closing, so please a new API function or adjust `tfw_filter_block_ip()` to call connection resetting. All ``tfw_filter_block_ip()` calls must be revised, including the http_limits.

There is relevant discussion https://github.com/tempesta-tech/tempesta/pull/1323/files#r309707851 . `ss_tcp_process_data()` -> `ss_tcp_process_skb()` returns true/false on all kind of problems: e.g. inability to unroll an skb and a security event, like password cracking. However, the problems are very different - we should send pending data in first case and send RST (subject for #861) in the second. So need to amend `ss_tcp_process_skb()` to return SS_BAD in generic errors and SS_DROP on security events and call `ss_close()` with the right argument from `ss_tcp_data_ready()`.

See TLS related comment https://github.com/tempesta-tech/tempesta/pull/1465#issuecomment-736581567 and reference TLS half-closed connections in https://security.stackexchange.com/questions/82028/ssl-tls-is-a-server-always-required-to-respond-to-a-close-notify and https://tools.ietf.org/html/rfc8446#section-6.1 . `tfw_tls_encrypt()` also needs to reset a TCP connection if the TLS layer can not encrypt data - we can not send FIN in this case because we truncate the application data.

The foundation for the restting logic can be `tcp_v4_send_reset()` and `tcp_v6_send_reset()` linux routines, except that this logic must not reverse dst and src addresses.

# Testing

https://github.com/tempesta-tech/tempesta-test/issues/286

PR #1323, opened 2019-07-24
Close the connection after sending tls alerts in the queue

avbelov23:
#1308

- Close the socket after sending all pending data on TCP_CLOSE_WAIT and internal errors.

- Close the connection after sending fatal tls alerts and close_notify alert in the queue

- Not encryption for TTLS_SERVER_FINISHED, because encryption is done in `ttls_write_finished()`.

- Sending close_notify alert on close_notify alert.

### DIFF
--- a/tempesta_fw/sock.c
+++ b/tempesta_fw/sock.c
@@ -636,13 +636,13 @@ adjudge_to_death:
  *
  * This is unintentional connection closing, usually due to some data errors.
  * This is not socket error, but still must lead to connection failovering
- * for server sockets. So connection_error callback is called here.
+ * for server sockets.
  */
 static void
 ss_linkerror(struct sock *sk)
 {
 	ss_do_close(sk);
-	SS_CALL_GUARD_EXIT(connection_error, sk);
+	SS_CALL_GUARD_EXIT(connection_drop, sk);
 	sock_put(sk);	/* paired with ss_do_close() */
 }
 
@@ -866,17 +866,27 @@ ss_tcp_data_ready(struct sock *sk)
 		if (ss_tcp_process_data(sk) &&
 		    !(SS_CONN_TYPE(sk) & Conn_Stop)) {
 			/*
-			 * Drop connection in case of internal errors,
+			 * Close connection in case of internal errors,
 			 * banned packets, or FIN in the received packet,
 			 * and only if it's not on hold until explicitly
 			 * closed.
 			 *
-			 * ss_linkerror() is responsible for calling
+			 * ss_close() is responsible for calling
 			 * application layer connection closing callback.
 			 * The callback will free all SKBs linked with
 			 * the message that is currently being processed.
+			 *
+			 * On receiving FIN we have to reply ACK and move to
+			 * TCP_CLOSE WAIT state where we send all required data
+			 * to the peer and only after than send our FIN.
+			 *
+			 * TODO #861. ss_tcp_process_data() returns true/false
+			 * on all kind of problems: e.g. inability to unroll an
+			 * skb and a security event. However, the problems are
+			 * very different - we should send pending data in first
+			 * case (SS_BAD) and send RST in the second (SS_DROP).
 			 */
-			ss_linkerror(sk);
+			ss_close(sk, SS_F_SYNC);
 		}
 	}
 	else {
@@ -990,7 +1000,12 @@ ss_tcp_state_change(struct sock *sk)
 		if (!skb_queue_empty(&sk->sk_receive_queue))
 			ss_tcp_process_data(sk);
 		T_DBG2("[%d]: Peer connection closing\n", smp_processor_id());
-		ss_linkerror(sk);
+		/*
+		 * On receiving FIN we have to reply ACK and more to
+		 * TCP_CLOSE WAIT state where we send all required data
+		 * to the peer and only after than send our FIN.
+		 */
+		ss_close(sk, SS_F_SYNC);
 	}
 	else if (sk->sk_state == TCP_CLOSE) {
 		/*
@@ -1362,18 +1377,30 @@ ss_tx_action(void)
 				bh_unlock_sock(sk);
 				break;
 			}
-			__sk_close_locked(sk); /* paired with bh_lock_sock() */
+			/* paired with bh_lock_sock() */
+			__sk_close_locked(sk);
 			break;
 		case SS_CLOSE:
+			/*
+			 * We were asked to close the socket. If current state
+			 * is either ESTABLISHED or SYN_SENT, we initiate the
+			 * closure process. If for some reason other side sent
+			 * us FIN, we are at CLOSE_WAIT, so the socket closure
+			 * is in progress, but still need to cleanup on our
+			 * side. If we get here while the socket in any other
+			 * state, resources were either already freed or were
+			 * never allocated.
+			 */
 			if (!((1 << sk->sk_state)
-			      & (TCPF_ESTABLISHED | TCPF_SYN_SENT)))
+			      & (TCPF_ESTABLISHED | TCPF_SYN_SENT | TCPF_CLOSE_WAIT)))
 			{
 				T_DBG2("[%d]: %s: Socket inactive: sk %p\n",
 				       smp_processor_id(), __func__, sk);
 				bh_unlock_sock(sk);
 				break;
 			}
-			__sk_close_locked(sk); /* paired with bh_lock_sock() */
+			/* paired with bh_lock_sock() */
+			__sk_close_locked(sk);
 			break;
 		default:
 			BUG();
--- a/tempesta_fw/sock.c
+++ b/tempesta_fw/sock.c
@@ -2,7 +2,7 @@
  *		Synchronous Socket API.
  *
  * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
- * Copyright (C) 2015-2018 Tempesta Technologies, Inc.
+ * Copyright (C) 2015-2019 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
@@ -876,9 +876,8 @@ ss_tcp_data_ready(struct sock *sk)
 			 * The callback will free all SKBs linked with
 			 * the message that is currently being processed.
 			 *
-			 * On receiving FIN we have to reply ACK and move to
-			 * TCP_CLOSE WAIT state where we send all required data
-			 * to the peer and only after than send our FIN.
+			 * Closing a socket should go through the queue and
+			 * should be done after all pending data has been sent.
 			 *
 			 * TODO #861. ss_tcp_process_data() returns true/false
 			 * on all kind of problems: e.g. inability to unroll an
@@ -1001,9 +1000,8 @@ ss_tcp_state_change(struct sock *sk)
 			ss_tcp_process_data(sk);
 		T_DBG2("[%d]: Peer connection closing\n", smp_processor_id());
 		/*
-		 * On receiving FIN we have to reply ACK and more to
-		 * TCP_CLOSE WAIT state where we send all required data
-		 * to the peer and only after than send our FIN.
+		 * Closing a socket should go through the queue and should be
+		 * done after all pending data has been sent.
 		 */
 		ss_close(sk, SS_F_SYNC);
 	}
@@ -1384,8 +1382,8 @@ ss_tx_action(void)
 			/*
 			 * We were asked to close the socket. If current state
 			 * is either ESTABLISHED or SYN_SENT, we initiate the
-			 * closure process. If for some reason other side sent
-			 * us FIN, we are at CLOSE_WAIT, so the socket closure
+			 * closing process. If for some reason other side sent
+			 * us FIN, we are at CLOSE_WAIT, so the socket closing
 			 * is in progress, but still need to cleanup on our
 			 * side. If we get here while the socket in any other
 			 * state, resources were either already freed or were
--- a/tempesta_fw/sock_clnt.c
+++ b/tempesta_fw/sock_clnt.c
@@ -222,17 +222,18 @@ err_client:
 	return r;
 }
 
-/**
- * Do the same stuff for intentional client connection closing and due to some
- * error on TCP socket or application layers.
+/*
+ * The hook is executed when a client connection is closed by either
+ * side of the connection or client connection is terminated due to
+ * an error of any kind.
  */
 static void
-tfw_sock_clnt_do_drop(struct sock *sk, const char *msg)
+tfw_sock_clnt_drop(struct sock *sk)
 {
 	TfwConn *conn = sk->sk_user_data;
 
-	T_DBG3("%s: close client socket: sk=%p, conn=%p, client=%p\n",
-	       msg, sk, conn, conn->peer);
+	T_DBG3("connection lost: close client socket: sk=%p, conn=%p, "
+	       "client=%p\n", sk, conn, conn->peer);
 	/*
 	 * Withdraw from socket activity. Connection is now closed,
 	 * and Tempesta is not called anymore on events in the socket.
@@ -251,30 +252,9 @@ tfw_sock_clnt_do_drop(struct sock *sk, const char *msg)
 	tfw_connection_put(conn);
 }
 
-/*
- * The hook is executed when a client connection is closed by either
- * side of the connection.
- */
-static void
-tfw_sock_clnt_drop(struct sock *sk)
-{
-	tfw_sock_clnt_do_drop(sk, "connection lost");
-}
-
-/*
- * The hook is executed when a client connection is terminated due to
- * an error of any kind.
- */
-static void
-tfw_sock_clnt_error(struct sock *sk)
-{
-	tfw_sock_clnt_do_drop(sk, "connection error");
-}
-
 static const SsHooks tfw_sock_clnt_ss_hooks = {
 	.connection_new		= tfw_sock_clnt_new,
 	.connection_drop	= tfw_sock_clnt_drop,
-	.connection_error	= tfw_sock_clnt_error,
 	.connection_recv	= tfw_connection_recv,
 };
 
--- a/tempesta_fw/sync_socket.h
+++ b/tempesta_fw/sync_socket.h
@@ -59,21 +59,16 @@ typedef struct ss_hooks {
 	int (*connection_new)(struct sock *sk);
 
 	/*
-	 * Drop TCP connection associated with the socket.
-	 * The callback is called on intentional socket closing when the socket
-	 * is already closed (i.e. there could not be ingress data on it) and we
-	 * can safely do some clenup stuff. We need the callback sine socket
-	 * closing always has chance to run asynchronously on other CPU and a
-	 * caller doesn't know the it completes.
+	 * Intentional socket closing when the socket is already closed (i.e. there
+	 * could not be ingress data on it) and we can safely do some clenup stuff
+	 * or error on TCP connection (on Linux TCP socket layer) associated with
+	 * the socket or at application (data processing) layer, i.e. unintentional
+	 * connection closing.
+	 * We need the callback since socket closing always has a chance to run
+	 * asynchronously on another CPU and a caller doesn't know when it
+	 * completes.
 	 */
 	void (*connection_drop)(struct sock *sk);
-
-	/*
-	 * Error on TCP connection (on Linux TCP socket layer) associated
-	 * with the socket or at application (data processing) layer,
-	 * i.e. unintentional connection closing.
-	 */
-	void (*connection_error)(struct sock *sk);
 
 	/* Process data received on the socket. */
 	int (*connection_recv)(void *conn, struct sk_buff *skb,
--- a/tempesta_fw/tls.c
+++ b/tempesta_fw/tls.c
@@ -530,9 +530,10 @@ tfw_tls_send(TlsCtx *tls, struct sg_table *sgt, bool close)
 			       skb, skb->len);
 		}
 	}
-
-	if (close)
+	if (close) {
 		flags |= SS_F_CONN_CLOSE;
+		TFW_CONN_TYPE(&conn->cli_conn) |= Conn_Stop;
+	}
 	if (ttls_xfrm_ready(tls))
 		flags |= SS_F_ENCRYPT;
 
--- a/tempesta_fw/tls.c
+++ b/tempesta_fw/tls.c
@@ -534,8 +534,8 @@ tfw_tls_send(TlsCtx *tls, struct sg_table *sgt, bool close)
 		flags |= SS_F_CONN_CLOSE;
 		TFW_CONN_TYPE(&conn->cli_conn) |= Conn_Stop;
 	}
-	if (ttls_xfrm_ready(tls))
-		flags |= SS_F_ENCRYPT;
+	if (ttls_xfrm_need_encript(tls))
+		flags |= SS_SKB_TYPE2F(io->msgtype) | SS_F_ENCRYPT;
 
 	return ss_send(conn->cli_conn.sk, &io->skb_list, flags);
 }
--- a/tempesta_fw/tls.c
+++ b/tempesta_fw/tls.c
@@ -534,7 +534,7 @@ tfw_tls_send(TlsCtx *tls, struct sg_table *sgt, bool close)
 		flags |= SS_F_CONN_CLOSE;
 		TFW_CONN_TYPE(&conn->cli_conn) |= Conn_Stop;
 	}
-	if (ttls_xfrm_need_encript(tls))
+	if (ttls_xfrm_need_encrypt(tls))
 		flags |= SS_SKB_TYPE2F(io->msgtype) | SS_F_ENCRYPT;
 
 	return ss_send(conn->cli_conn.sk, &io->skb_list, flags);
--- a/tls/tls_srv.c
+++ b/tls/tls_srv.c
@@ -2264,6 +2264,8 @@ ttls_handshake_finished(TlsCtx *tls)
 		if ((r = ttls_write_finished(tls, &sgt, &p)))
 			return r;
 		CHECK_STATE(TLS_HEADER_SIZE + TTLS_HS_FINISHED_BODY_LEN);
+		sg_mark_end(&sgt.sgl[sgt.nents - 1]);
+		r = __ttls_send_record(tls, &sgt, false);
 		/*
 		 * In case of session resuming, invert the client and server
 		 * ChangeCipherSpec messages order.
@@ -2272,7 +2274,7 @@ ttls_handshake_finished(TlsCtx *tls)
 			     ? TTLS_CLIENT_CHANGE_CIPHER_SPEC
 			     : TTLS_HANDSHAKE_WRAPUP;
 		sg_mark_end(&sgt.sgl[sgt.nents - 1]);
-		return __ttls_send_record(tls, &sgt, false);
+		return r;
 	}
 	}
 	T_FSM_FINISH(r, tls->state);
--- a/tls/tls_srv.c
+++ b/tls/tls_srv.c
@@ -2273,7 +2273,6 @@ ttls_handshake_finished(TlsCtx *tls)
 		tls->state = tls->hs->resume
 			     ? TTLS_CLIENT_CHANGE_CIPHER_SPEC
 			     : TTLS_HANDSHAKE_WRAPUP;
-		sg_mark_end(&sgt.sgl[sgt.nents - 1]);
 		return r;
 	}
 	}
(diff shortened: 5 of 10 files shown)


tempesta_fw/sock.c
```diff
@@ -990,7 +994,12 @@ ss_tcp_state_change(struct sock *sk)
 		if (!skb_queue_empty(&sk->sk_receive_queue))
 			ss_tcp_process_data(sk);
 		TFW_DBG2("[%d]: Peer connection closing\n", smp_processor_id());
-		ss_linkerror(sk);
+		/*
+		 * Use ss_close() to close the connection after
+		 * sending responses in the queue, for example,
+		 * tls alerts
+		 */
+		ss_close(sk, SS_F_SYNC);
```
ikoveshnikov: Let's see, what is really happen here. There could be a few cases:

- Backend server has sent us a TCP frame with `FIN` flag set. It means that it has replied everything it wanted, there is no any good to send it more requests: it'll never reply to them. So it's better for us to immediately close the connection and avoid sending more requests to it.

- Client has sent us `FIN` in last request inside http connection. I don't remember such situation in the RFC, but it seems  that here is no need to process requests in the receive queue at all: any way it is to be closed before response from backend server will be received. In this case we should either close client immediately without receive_queue processing, or send all the responses and close the connection after last is sent.

- Client has sent us `FIN` inside HTTPS connection and we have alerts to send - yes, it seams reasonable to perform a scheduled connection close by `ss_close()` call.

I would say that replacing `ss_linkerror()` to `ss_close()` is mostly performance question here.

tempesta_fw/sock.c
```diff
@@ -1365,14 +1374,6 @@ ss_tx_action(void)
 			__sk_close_locked(sk); /* paired with bh_lock_sock() */
 			break;
 		case SS_CLOSE:
-			if (!((1 << sk->sk_state)
-			      & (TCPF_ESTABLISHED | TCPF_SYN_SENT)))
```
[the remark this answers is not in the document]
avbelov23: How is this possible?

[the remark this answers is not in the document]
avbelov23: Removed what to call `ss_close()` from TCP_CLOSE_WAIT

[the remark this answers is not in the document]
avbelov23: On the tcp state machine, we can not get into `TCP_CLOSE_WAIT` bypassing `TCP_ESTABLISHED` and we may call `ss_close()` into `TCP_CLOSE_WAIT` and must bypass this condition

tempesta_fw/sock.c
avbelov23: @krizhanovsky

tempesta_fw/sock.c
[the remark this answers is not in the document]
keshonok: In my understanding, this check  is about remembering that there's always two sides of a connection: local and remote. When a postponed/scheduled action is performed, there's always a chance that the connection's state had changed due to an action on behalf of the other side (such as FIN, for example). In this case, a state change may have occurred between a request to intentionally close a connection from within Tempesta and the time the request is actually processed. That basically means that the requested closing action has been done already, just by the request from the other side. Thus if the connection is in `TCP_CLOSE_WAIT` state, then the connection is being closed already  and there's no need to do anything here in this place.

tempesta_fw/sock.c
keshonok: Is there a clear comment why `ss_linkerror()` is no longer adequate in `ss_tx_action()` and `ss_close()` is preferred there? I am not sure I understand the reason for the change. Thanks.

avbelov23: This is `ss_tcp_state_change()`, not `ss_tx_action()`.
We do not want to close the connection immediately, but only after sending the responses in the queue. For example, TLS alerts. See comment above.

keshonok: I am sorry that I am not familiar with the TLS part.

I am not sure which connection and which side of the connection is in question here. This particular place is where TCP state change occurs when the other side has sent `FIN`  - initiated the closing of the connection. After that, the other side does not care of receiving any data. Perhaps, that's not the case for a TLS connection, and the other side, the one that has sent `FIN`, expects to receive something after that? That would be strange in terms of a TCP connection, in my opinion.  Perhaps, a better comment is necessary.

Now, the receive queue is processed here because the packet with `FIN` might not be the only one in the queue but be the last one. Data in those preceding packets, including the packet with `FIN`, need to be processed by Tempesta. This place is called before `ss_tcp_data_ready()`, and so we just process the receive queue and close the connection directly, thus making it so `ss_tcp_data_ready()` is not called at all.

If  we're talking about doing something (like sending data to the other side of a _live_ connection) _before_ sending FIN ourselves (before _initiating_ the closing of a connection), then perhaps the place in `ss_tcp_state_change()` is not the right place for that. This place is about handling a connection closing initiated by the other side.

It's very similar in the other function `ss_tcp_data_ready()`, the logic is very much the same. There it's more about internal Tempesta errors though. Note that nearly all errors are handled within Tempesta itself by performing some actions and initiating a connection closing (usually, a scheduled closing). It's a rare case an error is returned from Tempesta to the socket layer to `ss_tcp_data_ready()`. Those are very specific cases of "hard" errors where that is justified, and so `ss_linkerror()` is the right choice there. If an error is returned from TLS layer to the socket layer in a case that needs additional handling while the socket is alive, then, perhaps, that is something that should not happen, by the overall design of Tempesta.

(_The comment about FIN in `ss_tcp_data_ready()` there may no longer be true now, need to check that. The socket should be dead after closing in `ss_tcp_state_change()`, and so `ss_tcp_data_ready() is not called. Also, all skb are flushed from the queue in the process anyway._)

The difference  between `ss_linkerror()` and `ss_close()` is two-fold.

- Different Tempesta callback is called, `connection_error()` versus `connection_drop()`
- One is direct closing, and the other is a scheduled, postponed closing.

In both cases a callback into Tempesta is called _after_ the connection is closed. In that regard, the connection _is_ closed "immediately", and not after something else is done in a callback. The only very subtle difference is the "postpone" part, where the closing action is performed in a specific SoftIRQ (different from RX/TX SoftIRQ), and therefore there _may_ be a "cycle" of Tempesta work before that specific SoftIRQ kicks in. But that's not guaranteed - this  scheduling of the closing action may be the last work for RX/TX SoftIRQ, and then the specific  SoftIRQ would kick in immediately.

All in all, I believe this issue would greatly benefit from a clear description of the scenario (or several scenarios) in mind, and how the changes help in making those scenarios work correctly versus the existing code. Perhaps, describe that in the issue on Github. Right now it's difficult to see how these changes fix the issue.

krizhanovsky: > Backend server has sent us a TCP frame with FIN flag set. It means that it has replied everything it wanted, there is no any good to send it more requests: it'll never reply to them. So it's better for us to immediately close the connection and avoid sending more requests to it.

Good point: we should take advantage on knowledge that at TCP layer we know about application layer and reduce amount of work to do. It could make sense to leave current `ss_linkerror()` for server connections, introducing additional knowledge of the socket type on SS layer (now we don't know in sock.c which sockets are server and which are client).

However, with that we've been living happily with the code until TLS in mind, this proposal seems error prone. We'll have full TLS proxying (so, we'll have to send alerts to backends as well) and also there will be websockets with full duplex connection and, the worse, we don't plan to understand websockets protocol. So at this point we can make the optimization for HTTP server connections only. Maybe we'll see in future a case with massive FINs from HTTP backends, but I don't want to do the optimization now.

But I believe the opportunity must be commented.

> Client has sent us FIN in last request inside http connection. I don't remember such situation in the RFC, but it seems that here is no need to process requests in the receive queue at all: any way it is to be closed before response from backend server will be received.

This is allowed and adequate for a client to send a request with `Connection: close` and add FIN to the same TCP segment. It seems we simply have no test cases for the scenario, but I believe #1308 also affect the scenario.

`ss_linkerror()`, called on receiving FIN from a peer, calls `ss_do_close()` with immediate FIN transmission to the peer regardless some application data in our work queue which is still was not added to TCP send queue. This breaks TCP FSM: on receiving FIN we have to reply ACK and more to CLOSE WAIT state where we send all required data to the peer and only after than send our FIN. An application layer, as we see with TLS now and probably see with client request with FIN and `Connection: close`, may be broken by our current behaviour.

tempesta_fw/sock.c
```diff
@@ -1366,7 +1375,8 @@ ss_tx_action(void)
 			break;
 		case SS_CLOSE:
 			if (!((1 << sk->sk_state)
-			      & (TCPF_ESTABLISHED | TCPF_SYN_SENT)))
+			      & (TCPF_ESTABLISHED | TCPF_SYN_SENT |
+			         TCPF_CLOSE_WAIT)))
```
krizhanovsky: Linux `tcp_fin()` sets `tcp_set_state(sk, TCP_CLOSE_WAIT)` when we receive FIN from a peer, next the function calls `sk->sk_state_change(sk)`, i.e. our `ss_tcp_state_change()` where we call `ss_close()` now, so our closing work is queued into work queue. `ss_tx_action()` sends all pending data and reaches the closing work. Now the socket in `TCPF_CLOSE_WAIT` state and we do call `__sk_close_locked()`. The change looks correct. We didn't need it before because with previous `ss_linkerror()` we immediately moved from `TCPF_CLOSE_WAIT` to next states like TCP_CLOSE.

tempesta_fw/sock.c
```diff
@@ -866,17 +866,21 @@ ss_tcp_data_ready(struct sock *sk)
...
 			 * application layer connection closing callback.
 			 * The callback will free all SKBs linked with
 			 * the message that is currently being processed.
+			 *
+			 * Use ss_close() to close the connection after
+			 * sending responses in the queue, for example,
+			 * tls alerts
 			 */
-			ss_linkerror(sk);
+			ss_close(sk, SS_F_SYNC);
```
krizhanovsky: I reviewed the code around once more and it seems that we should do more complex changes. `ss_tcp_process_data() -> `ss_tcp_process_skb()` returns true/false on all kind of problems: e.g. inability to unroll an skb and a security event, like password cracking. However, the problems are very different - we should send pending data in first case and send RST (subject for #861) in the second. However, it's proposed to extend `ss_close()` with an argument for RST and this change just makes step further. In #861 we'll need to amend `ss_tcp_process_skb()` to return SS_BAD in generic errors and SS_DROP on security events. So I'm still fine with the change and will just add the requirement to #861.

Please add TODO #861 with the discussion summary to the comment and put coma at the end of comment.

tempesta_fw/sock.c
keshonok: Yes, it does make sense when it's together with the change in `ss_tcp_state_change()`. Gotcha, and thanks for a clear explanation. This is a good piece of explanation to put somewhere in a comment or, perhaps, just refer to this one in the PR.

krizhanovsky:
Just saw kernel oops on CI http://93.115.28.191:4010/#/builders/5/builds/810/steps/21/logs/stdio:
```
[  479.832011] ------------[ cut here ]------------
[  479.834442] kernel BUG at /root/tempesta/tempesta/tempesta_fw/http.c:3058!
[  479.837687] invalid opcode: 0000 [#1] SMP PTI
[  479.840053] Modules linked in: tempesta_fw(O) tempesta_db(O) tempesta_tls(O) tempesta_lib(O) sha256_ssse3 sha512_ssse3 sha512_generic ccm fuse ata_generic intel_rapl sb_edac crct10dif_pclmul crc32_pclmul joydev ghash_clmulni_intel ata_piix cirrus ttm libata drm_kms_helper xen_netfront intel_rapl_perf psmouse pcspkr drm scsi_mod i2c_piix4 floppy button ip_tables x_tables autofs4 ext4 crc16 mbcache jbd2 crc32c_generic fscrypto ecb crc32c_intel aesni_intel evdev xen_blkfront aes_x86_64 crypto_simd cryptd glue_helper serio_raw [last unloaded: tempesta_lib]
[  479.860163] CPU: 1 PID: 0 Comm: swapper/1 Tainted: G           O    4.14.0-tempesta-amd64 #1 Debian 4.14.32-tfw6-1
[  479.864413] Hardware name: Xen HVM domU, BIOS 4.8.5 01/11/2019
[  479.867288] task: ffff9799007b9e00 task.stack: ffffb6ff80698000
[  479.870208] RIP: 0010:tfw_http_msg_process_generic+0x968/0xd00 [tempesta_fw]
[  479.873531] RSP: 0018:ffff97990f103970 EFLAGS: 00010216
[  479.876275] RAX: 0000000000000007 RBX: ffff9799019215f0 RCX: 0000000000000010
[  479.879614] RDX: ffff979904029000 RSI: ffff97990f103b50 RDI: ffff979904029600
[  479.882933] RBP: ffff9798b1682020 R08: 0000000000000001 R09: 0000000000000000
[  479.886277] R10: 0000000000000000 R11: ffff9798b16b6140 R12: ffff97990f103b50
[  479.889603] R13: 0000000000000000 R14: 0000000000000000 R15: ffff979904029600
[  479.892912] FS:  0000000000000000(0000) GS:ffff97990f100000(0000) knlGS:0000000000000000
[  479.896496] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[  479.899417] CR2: 00007ff8c31215d8 CR3: 000000005c20a005 CR4: 00000000001606e0
[  479.902747] Call Trace:
[  479.904583]  <IRQ>
[  479.906282]  ? generic_gcmaes_decrypt+0x5f/0x80 [aesni_intel]
[  479.909139]  ? ttls_decrypt+0x297/0x580 [tempesta_tls]
[  479.911828]  tfw_http_msg_process+0x9e/0xe0 [tempesta_fw]
[  479.914610]  ? pg_skb_alloc+0x253/0x470
[  479.916842]  __gfsm_fsm_exec+0x56/0x90 [tempesta_fw]
[  479.919446]  ? skb_split+0x1ff/0x2e0
[  479.921594]  tfw_gfsm_move+0x132/0x160 [tempesta_fw]
[  479.924197]  tfw_tls_msg_process+0x1de/0x360 [tempesta_fw]
[  479.926930]  __gfsm_fsm_exec+0x56/0x90 [tempesta_fw]
[  479.929535]  tfw_connection_recv+0x4e/0x70 [tempesta_fw]
[  479.932232]  ? tfw_connection_send+0x30/0x30 [tempesta_fw]
[  479.935009]  ss_tcp_process_data+0x1db/0x440 [tempesta_fw]
[  479.937729]  ss_tcp_data_ready+0x43/0x90 [tempesta_fw]
[  479.940413]  tcp_rcv_established+0x4d2/0x570
[  479.942795]  tcp_v4_do_rcv+0x129/0x1d0
[  479.944992]  tcp_v4_rcv+0x947/0xa50
[  479.947092]  ip_local_deliver_finish+0x9a/0x1c0
[  479.949529]  ip_local_deliver+0x6b/0xe0
[  479.951714]  ? tcp_v4_early_demux+0x112/0x150
[  479.954011]  ? ip_rcv_finish+0x17a/0x400
[  479.956197]  ip_rcv+0x289/0x3c0
[  479.958134]  ? inet_del_offload+0x40/0x40
[  479.960302]  __netif_receive_skb_core+0x84f/0xb30
[  479.962713]  ? process_backlog+0xa3/0x160
[  479.964849]  process_backlog+0xa3/0x160
[  479.966961]  net_rx_action+0x28e/0x3f0
[  479.969026]  __do_softirq+0x10f/0x2a8
[  479.971078]  irq_exit+0xae/0xb0
[  479.972953]  xen_evtchn_do_upcall+0x2c/0x40
[  479.975072]  xen_hvm_callback_vector+0x7d/0x90
[  479.977231]  </IRQ>
[  479.978667] RIP: 0010:native_safe_halt+0x2/0x10
[  479.980787] RSP: 0018:ffffb6ff8069beb8 EFLAGS: 00000246 ORIG_RAX: ffffffffffffff0c
[  479.983908] RAX: ffffffffbc897e60 RBX: ffff9799007b9e00 RCX: 0000000000000000
[  479.986907] RDX: 0000000000000000 RSI: 0000000000000000 RDI: 0000000000000000
[  479.989835] RBP: 0000000000000001 R08: 00000000593296c0 R09: ffff979901a3ef00
[  479.992813] R10: 0000000000000000 R11: 0000013cfadf1cfb R12: ffff9799007b9e00
[  479.995780] R13: ffff9799007b9e00 R14: 0000000000000000 R15: 0000000000000000
[  479.998696]  ? __sched_text_end+0x3/0x3
[  480.000585]  default_idle+0x1a/0xf0
[  480.002339]  do_idle+0x16e/0x1f0
[  480.004029]  cpu_startup_entry+0x6f/0x80
[  480.005937]  start_secondary+0x1a9/0x200
[  480.007837]  secondary_startup_64+0xa5/0xb0
[  480.009798] Code: 00 00 49 8b 95 80 00 00 00 48 8d b5 c8 00 00 00 e8 6e 86 fe ff 84 c0 0f 84 d4 fd ff ff f0 41 80 a5 a9 00 00 00 fd e9 c6 fd ff ff <0f> 0b 83 f8 03 75 32 41 b8 01 00 00 00 b9 01 00 00 00 48 c7 c2 
[  480.016945] RIP: tfw_http_msg_process_generic+0x968/0xd00 [tempesta_fw] RSP: ffff97990f103970
[  480.020320] ---[ end trace 8c443069c4c3736c ]---
[  480.022479] Kernel panic - not syncing: Fatal exception in interrupt
[  480.025191] Kernel Offset: 0x3b200000 from 0xffffffff81000000 (relocation range: 0xffffffff80000000-0xffffffffbfffffff)
```

ikoveshnikov:
> Just saw kernel oops

It seems like it doesn't related to this PR, I've got the same on master branch, https://github.com/tempesta-tech/tempesta/issues/1283#issuecomment-518254715

tempesta_fw/sock.c
```diff
@@ -1323,13 +1340,15 @@ ss_getpeername(struct sock *sk, TfwAddr *addr)
 }
 EXPORT_SYMBOL(ss_getpeername);
 
-#define __sk_close_locked(sk)					\
-do {								\
-	ss_do_close(sk);					\
-	bh_unlock_sock(sk);					\
-	SS_CALL_GUARD_EXIT(connection_drop, sk);		\
```
ikoveshnikov: What is the reason to drop call of  `SS_CALL_GUARD_EXIT`?

avbelov23: Yes. Did not have time to push

tempesta_fw/sock.c
```diff
@@ -1362,18 +1377,34 @@ ss_tx_action(void)
...
+			 * we receive FIN from a peer, next the function calls
+			 * sk->sk_state_change(sk), i.e. our
+			 * ss_tcp_state_change() where we call ss_close() now,
+			 * so our closing work is queued into work queue.
+			 * ss_tx_action() sends all pending data and reaches the
+			 * closing work. Now the socket in TCPF_CLOSE_WAIT state
+			 * and we do call __sk_close_locked(). We didn't need it
+			 * before because with previous ss_linkerror() we
+			 * immediately moved from TCPF_CLOSE_WAIT to next states
+			 * like TCP_CLOSE.
```
i-rinat: I'd like this to be a description of a code rather than a description of a change. The line "We didn't need it before because with previous ss_linkerror()" talks about what was code before this particular PR. It will be extremely confusing later.

It would be nice to have this a comment about why those particular ESTABLISHED, SYN_SENT, and CLOSE_WAIT are the states where we want to call `__sk_close_locked()`. Something like:

>We were asked to close the socket. If current state is either ESTABLISHED or SYN_SENT, we initiate the closure process. If for some reason other side sent us FIN, we are at CLOSE_WAIT, so the socket closure is in progress, but still need to cleanup on our side. If we get here while the socket in any other state, resources were either already freed or were never allocated.

I'm still not sure. Also, why SYN_RECV is not in the selected states?

tls/ttls.c
```diff
@@ -1334,10 +1334,37 @@ ttls_send_alert(TlsCtx *tls, unsigned char lvl, unsigned char msg)
 	io->alert[0] = lvl;
 	io->alert[1] = msg;
 
-	if (msg == TTLS_ALERT_MSG_CLOSE_NOTIFY)
+	switch (msg) {
+	case TTLS_ALERT_MSG_CLOSE_NOTIFY:
```
i-rinat: Do we really need to enumerate all alerts here? If alert is fatal, we close. If it's a warning, but close_notify, we also close. So I think it's enough to add `|| lvl == TTLS_ALERT_LEVEL_FATAL` to the `if`.

There is only one place in the code where we issue a warning-level alert that is not close_notify. It's `TTLS_ALERT_MSG_NO_RENEGOTIATION`.  I think we need a fatal level there since we treat that's as an error, and refuse to proceed by returning `TTLS_ERR_UNEXPECTED_MESSAGE`.

tempesta_fw/tls.c
```diff
@@ -476,7 +476,7 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
  * initiates a record transmission, e.g. alert or a handshake message.
  */
 static int
-tfw_tls_send(TlsCtx *tls, struct sg_table *sgt, bool close)
+tfw_tls_send(TlsCtx *tls, struct sg_table *sgt, bool close, bool was_encrypt)
```
i-rinat: "was encrypt" sounds wrong. Please at least make it "was encrypted".

tempesta_fw/tls.c
```diff
@@ -530,11 +530,12 @@ tfw_tls_send(TlsCtx *tls, struct sg_table *sgt, bool close)
...
 	}
-
-	if (close)
+	if (close) {
 		flags |= SS_F_CONN_CLOSE;
-	if (ttls_xfrm_ready(tls))
-		flags |= SS_F_ENCRYPT;
+		TFW_CONN_TYPE(&conn->cli_conn) |= Conn_Stop;
+	}
+	if (ttls_xfrm_ready(tls) && !was_encrypt)
```
i-rinat: Having both `ttls_xfrm_ready()` and `was_encrypt` looks redundant and wrong.

We already know which records we have to encrypt and which not. In the `T_FSM_STATE(TTLS_SERVER_FINISHED)` we change `tls->state` before `__ttls_send_record()` is called, so here `tls->state` is a bit off. But order can be changed. Here is what I'm talking about:
```patch
diff --git a/tls/tls_srv.c b/tls/tls_srv.c
index 58699482..b9e2cb7f 100644
--- a/tls/tls_srv.c
+++ b/tls/tls_srv.c
@@ -2262,19 +2262,20 @@ ttls_handshake_finished(TlsCtx *tls)
 	}
 	T_FSM_STATE(TTLS_SERVER_FINISHED) {
 		if ((r = ttls_write_finished(tls, &sgt, &p)))
 			return r;
 		CHECK_STATE(TLS_HEADER_SIZE + TTLS_HS_FINISHED_BODY_LEN);
+		sg_mark_end(&sgt.sgl[sgt.nents - 1]);
+		r = __ttls_send_record(tls, &sgt, false, false);
 		/*
 		 * In case of session resuming, invert the client and server
 		 * ChangeCipherSpec messages order.
 		 */
 		tls->state = tls->hs->resume
 			     ? TTLS_CLIENT_CHANGE_CIPHER_SPEC
 			     : TTLS_HANDSHAKE_WRAPUP;
-		sg_mark_end(&sgt.sgl[sgt.nents - 1]);
-		return __ttls_send_record(tls, &sgt, false, true);
+		return r;
 	}
 	}
 	T_FSM_FINISH(r, tls->state);
 
 	/* If we exit here, then something went wrong. */
diff --git a/tls/ttls.c b/tls/ttls.c
index 5041ce56..1b0c6e89 100644
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -242,11 +242,12 @@ EXPORT_SYMBOL(ttls_register_bio);
  */
 bool
 ttls_xfrm_ready(TlsCtx *tls)
 {
 	return tls->state >= TTLS_CLIENT_FINISHED
-	       && tls->state != TTLS_SERVER_CHANGE_CIPHER_SPEC;
+	       && tls->state != TTLS_SERVER_CHANGE_CIPHER_SPEC
+	       && tls->state != TTLS_SERVER_FINISHED;
 }
 EXPORT_SYMBOL(ttls_xfrm_ready);
 
 #if defined(TTLS_CLI_C)
 static int
```

I think, it should work. And then `was_encrypt` can be removed altogether.

tempesta_fw/sock_clnt.c
```diff
@@ -253,28 +253,18 @@ tfw_sock_clnt_do_drop(struct sock *sk, const char *msg)
...
 }
 
-/*
- * The hook is executed when a client connection is terminated due to
- * an error of any kind.
- */
-static void
-tfw_sock_clnt_error(struct sock *sk)
-{
-	tfw_sock_clnt_do_drop(sk, "connection error");
```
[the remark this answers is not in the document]
i-rinat: From the current PR state I can say that this was addressed.

tempesta_fw/tls.c
avbelov23: It will fall here
https://github.com/tempesta-tech/tempesta/blob/master/tls/ttls.c#L796

i-rinat: Well, maybe. But anyway, `tls->state` should be enough to figure out if a record needs to be encrypted. Perhaps we'll need a new predicate for that.

tempesta_fw/sync_socket.h
```diff
@@ -59,21 +59,16 @@ typedef struct ss_hooks {
...
-	 * can safely do some clenup stuff. We need the callback sine socket
-	 * closing always has chance to run asynchronously on other CPU and a
-	 * caller doesn't know the it completes.
-	 */
-	void (*connection_drop)(struct sock *sk);
-
-	/*
-	 * Error on TCP connection (on Linux TCP socket layer) associated
-	 * with the socket or at application (data processing) layer,
+	 * Drop TCP connection associated with the socket (Intentional socket
```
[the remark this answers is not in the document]
i-rinat: Looks like this was addressed.

tls/ttls.c
i-rinat: I can see changes in the patches that are addressing this comment.

tempesta_fw/tls.c
i-rinat: Addressed.

tempesta_fw/sync_socket.h
```diff
@@ -59,21 +59,16 @@ typedef struct ss_hooks {
...
+	 * closing when the socket is already closed (i.e. there could not be
+	 * ingress data on it) and we can safely do some clenup stuff. We need
+	 * the callback sine socket closing always has chance to run
+	 * asynchronously on other CPU and a caller doesn't know the it
```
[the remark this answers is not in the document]
i-rinat: addressed too

tempesta_fw/sock.c
avbelov23: SYNC_RECV is not needed, because we select the sockets where we are the server from the kernel upon reaching ESTABLISHED (`ss_tcp_state_change ()`), and we control the sockets where we are the client from the beginning of the life cycle

i-rinat: > SYNC_RECV is not needed

Oh, I see. Thanks for the explanation.

tls/ttls.c
```diff
@@ -248,6 +248,19 @@ ttls_xfrm_ready(TlsCtx *tls)
...
+/*
+ * There are states in which encryption is not used or performed in advance, as
+ * with TTLS_SERVER_FINISHED
+ */
+bool
+ttls_xfrm_need_encript(TlsCtx *tls)
+{
+	return tls->state >= TTLS_CLIENT_FINISHED
+	       && tls->state != TTLS_SERVER_CHANGE_CIPHER_SPEC
+	       && tls->state != TTLS_SERVER_FINISHED;
```
krizhanovsky: The code would be easier to understand with:
```C
return ttls_xfrm_ready(tls) && tls->state != TTLS_SERVER_FINISHED;
```

tls/tls_srv.c
```diff
@@ -2272,7 +2274,7 @@ ttls_handshake_finished(TlsCtx *tls)
 			     ? TTLS_CLIENT_CHANGE_CIPHER_SPEC
 			     : TTLS_HANDSHAKE_WRAPUP;
 		sg_mark_end(&sgt.sgl[sgt.nents - 1]);
```
krizhanovsky: Forgot to remove the line

tls/ttls.c
```diff
@@ -2218,6 +2231,11 @@ ttls_recv(void *tls_data, unsigned char *buf, size_t len, unsigned int *read)
 		if (unlikely(!ttls_xfrm_ready(tls))) {
 			if (!(r = ttls_handle_alert(io)))
 				goto skip_record;
+			if (io->alert[0] == TTLS_ALERT_LEVEL_WARNING
+			    && io->alert[1] == TTLS_ALERT_MSG_CLOSE_NOTIFY)
+			{
+				ttls_close_notify(tls);
+			}
```
krizhanovsky: `ttls_recv()` is overloaded and you copy & paste the same code, so please more the code to `ttls_handle_alert()` and amend the function to accept `tls` instead of `io`.

PR #1327, opened 2019-07-30
Reduce Tempesta logging on start/stop

krizhanovsky:
Tempesta produces too many log records on start & stop operations. This not only is boding, but also makes testing hard since netratelimiting message appears only on next log records, i.e. there is a possibility to loose a log record and learn it only on next start/stop.


avbelov23:
LGTM

PR #1330, opened 2019-08-06
remove "off" and "trail" fields from TfwFsmData

i-rinat:
TLS layer feeds `tfw_http_msg_process()` with multiple skb's at a time. As the smallest possible skb can be as small as one byte, the tail where TLS record tag is located, can span over two or more skb's. Previously we assumed that never happens, but it does depending on network conditions.

It's hard to deal with heading and trailing forbidden parts of an skb or an skb chain. So during the discussion of the PR it was decided to eliminate the notion of heading and trailing parts, so that the consumers of the data could work with skb data without tracking those parts. TLS-related code is now cuts off header and the trailing part that contains TLS tag.

This PR handles the case from #1283 with "kernel BUG at /root/tempesta/tempesta/tempesta_fw/http.c:3058!".

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -3172,24 +3172,6 @@ tfw_http_hm_drop_resp(TfwHttpResp *resp)
 	tfw_http_msg_free((TfwHttpMsg *)req);
 }
 
-/**
- * In general case the @skb, even if it has TLS record header @off and tag
- * @tail, it can be at any place of resulting HTTP message. Now we have to
- * chop the TLS extra data. It'd be good to preserve the room for the message
- * adjusting (e.g. adding a new header), but typically we need a space only in
- * skbs carrying CRLF and just before the CRLF pointer. So TLS rooms can be
- * reused only if they're just before CRLF (very rarely). Thus we just chop
- * the data.
- *
- * TODO #1103: this should be revised for the new placement strategies.
- */
-static int
-tfw_http_chop_skb(TfwHttpMsg *hm, struct sk_buff *skb,
-		  unsigned int off, unsigned int trail)
-{
-	return ss_skb_chop_head_tail(hm->msg.skb_head, skb, off, trail);
-}
-
 static TfwStr
 tfw_http_get_ip_from_xff(TfwHttpReq *req)
 {
@@ -3339,8 +3321,7 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, const TfwFsmData *data)
 {
 	bool block;
 	ss_skb_actor_t *actor;
-	unsigned int parsed, curr_skb_trail;
-	unsigned int off = data->off, trail = data->trail;
+	unsigned int parsed;
 	struct sk_buff *skb = data->skb;
 	TfwHttpReq *req;
 	TfwHttpMsg *hmsib;
@@ -3349,11 +3330,9 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, const TfwFsmData *data)
 	int r = TFW_BLOCK;
 
 	BUG_ON(!stream->msg);
-	BUG_ON(off >= skb->len);
-	BUG_ON(trail >= skb->len);
 
 	T_DBG2("Received %u client data bytes on conn=%p msg=%p\n",
-	       skb->len - off, conn, stream->msg);
+	       skb->len, conn, stream->msg);
 
 	/*
 	 * Process pipelined requests in a loop
@@ -3367,10 +3346,8 @@ next_msg:
 	parser = &stream->parser;
 	actor = TFW_MSG_H2(req) ? tfw_h2_parse_req : tfw_http_parse_req;
 
-	r = ss_skb_process(skb, off, trail, actor, req, &req->chunk_cnt,
-			   &parsed);
+	r = ss_skb_process(skb, actor, req, &req->chunk_cnt, &parsed);
 	req->msg.len += parsed;
-	curr_skb_trail = off + parsed + trail < skb->len ? 0 : trail;
 	TFW_ADD_STAT_BH(parsed, clnt.rx_bytes);
 
 	T_DBG2("Request parsed: len=%u next=%pK parsed=%d msg_len=%lu"
@@ -3383,8 +3360,6 @@ next_msg:
 	 * feed the new data version to FSMs registered on our states.
 	 */
 	data_up.skb = skb;
-	data_up.off = off;
-	data_up.trail = curr_skb_trail;
 	data_up.req = (TfwMsg *)req;
 	data_up.resp = NULL;
 
@@ -3397,8 +3372,16 @@ next_msg:
 		tfw_http_req_parse_drop(req, 400, "failed to parse request");
 		return TFW_BLOCK;
 	case TFW_POSTPONE:
-		if (tfw_http_chop_skb((TfwHttpMsg *)req, skb, off, curr_skb_trail))
+		if (WARN_ON_ONCE(parsed != data_up.skb->len)) {
+			/*
+			 * The parser should only return TFW_POSTPONE if it ate
+			 * all available data, but that weren't enough.
+			 */
+			TFW_INC_STAT_BH(clnt.msgs_otherr);
+			tfw_http_req_parse_block(req, 500,
+				"Request parsing inconsistency");
 			return TFW_BLOCK;
+		}
 		r = tfw_gfsm_move(&conn->state, TFW_HTTP_FSM_REQ_CHUNK,
 				  &data_up);
 		T_DBG3("TFW_HTTP_FSM_REQ_CHUNK return code %d\n", r);
@@ -3427,13 +3410,6 @@ next_msg:
 	}
 
 	/*
-	 * Chop TLS overhead for current skb after ss_skb_process() which
-	 * works with offset and trail on its own and before we split the skb.
-	 */
-	if (tfw_http_chop_skb((TfwHttpMsg *)req, skb, off, curr_skb_trail))
-		return TFW_BLOCK;
-
-	/*
 	 * The message is fully parsed, the rest of the data in the
 	 * stream may represent another request or its part.
 	 * If skb splitting has failed, the request can't be forwarded
@@ -3449,7 +3425,6 @@ next_msg:
 						 "Can't split pipelined requests");
 			return TFW_BLOCK;
 		}
-		off = 0;
 	} else {
 		skb = NULL;
 	}
@@ -3807,7 +3782,6 @@ tfw_http_resp_cache(TfwHttpMsg *hmresp)
 	 * cache shouldn't be accounted.
 	 */
 	data.skb = NULL;
-	data.off = data.trail = 0;
 	data.req = (TfwMsg *)req;
 	data.resp = (TfwMsg *)hmresp;
 	tfw_gfsm_move(&hmresp->conn->state, TFW_HTTP_FSM_RESP_MSG_FWD, &data);
@@ -3866,8 +3840,6 @@ tfw_http_resp_terminate(TfwHttpMsg *hm)
 	 */
 	data.skb = ss_skb_peek_tail(&hm->msg.skb_head);
 	BUG_ON(!data.skb);
-	data.off = data.skb->len;
-	data.trail = 0;
 	data.req = NULL;
 	data.resp = (TfwMsg *)hm;
 
@@ -3885,7 +3857,6 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, const TfwFsmData *data)
 {
 	int r = TFW_BLOCK;
 	unsigned int chunks_unused, parsed;
-	unsigned int off = data->off;
 	struct sk_buff *skb = data->skb;
 	TfwHttpReq *bad_req;
 	TfwHttpMsg *hmresp, *hmsib;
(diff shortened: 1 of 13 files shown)


tempesta_fw/http.c
```diff
@@ -3943,10 +3943,25 @@ tfw_http_msg_process(void *conn, TfwFsmData *data)
...
+	if (data->skb->prev) {
+		last_skb = data->skb->prev;
 		data->skb->prev->next = NULL;
+	} else {
+		last_skb = data->skb;
+		while (last_skb->next != NULL)
+			last_skb = last_skb->next;
+	}
+
+	while (last_skb->len <= trail) {
```
krizhanovsky: OK, I'm good with the loop - the condition in improbable (please mark it as `unlikely` BTW) and the loop is short. We *assume* so. But I don't like the first loop: firstly, it looks like a workaround for a 'bad' API - if we need last item in the list, then we always have to have `skb->prev`; secondly, it's unclear how long the loop could be. And finally I see no reason to traverse the list 3 time while in the first and second loop we have the same termination condition.

Recently a loop was introduced for cache-constructed skbs - it's not good as well, having in mind that skb is in most 64KB and current limit for cached response is IIRC 1GB. However, generally the behaviour is wrong and to be fixed in #498. However, introducing more and more loops on skb list is a bad practice.

i-rinat: I've removed the need for loops by passing everywhere a usable data length instead of a tail size. That way we can track whenever we are out of data, so the trailing skb's in a chain can be freed (see `tfw_http_msg_process()`).

i-rinat:
I've changed the code to pass everywhere a usable data length instead of a tail size. That way we can track whenever we are out of data, so the trailing skb's in a chain can be freed. See changes in `tfw_http_msg_process()` for details.

tempesta_fw/http.c
```diff
@@ -4141,25 +4148,33 @@ tfw_http_msg_process(void *conn, TfwFsmData *data)
...
-			kfree(data->skb);
+			kfree_skb(data->skb);
 		}
 	}
 
+	while (data->skb) {
+		next = data->skb->next;
+		kfree_skb(data->skb);
+		data->skb = next;
+	}
```
krizhanovsky: I was puzzled with the addition for some time, so it'd be reader friendly to have a comment here. We need the code for `!data_left && data->skb` and we reach the condition only for unusable for HTTP data, TLS tag in particular. I was puzzled because the `for` loop is complex enough and from quick reading it looks like the upper layer must set `data->skb = NULL` or isn't in charge to free skb, but it's not true...

tempesta_fw/http.c
```diff
@@ -4141,25 +4148,33 @@ tfw_http_msg_process(void *conn, TfwFsmData *data)
...
-			data->trail = !next ? trail : 0;
+			data->len = min_t(unsigned int, data_left,
+					  data->skb->len - data->off);
+			data_left -= data->len;
 			r = TFW_CONN_H2(conn)
 				? tfw_h2_frame_process(conn, data)
 				: tfw_http_msg_process_generic(conn, stream, data);
 		} else {
-			kfree(data->skb);
+			kfree_skb(data->skb);
```
ikoveshnikov: Ar this stage `skb` is orphaned from queues and used only inside Templesta, in all surrounding functions we use `__kfree_skb()` without reference checking and tracing events. should it be `__kfree_skb()` also?

tempesta_fw/tls.c
```diff
@@ -133,6 +135,20 @@ tfw_tls_msg_process(void *conn, TfwFsmData *data)
...
+		 * an skb arrived, so there is actually a saved msglen field
+		 * value which we can use.
+		 */
+		data_up.len = msglen - ttls_xfrm_taglen(&tls->xfrm) -
+			      ttls_expiv_len(&tls->xfrm);
+	} else {
+		WARN_ON(tls->io_in.msglen != 0);
+		data_up.len = parsed - ttls_xfrm_taglen(&tls->xfrm) -
+			      ttls_payload_off(&tls->xfrm);
+	}
```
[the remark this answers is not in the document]
i-rinat: The parsing code eventually called by `ttls_recv()` resets the `tls->io_in` by zeroing everything past `__initoff` after the whole record is parsed. So there are two cases.
First: TLS record was spanning over multiple skbs. We just finished processing a last skb with the record. At this moment `tls->io_in` was already reset, but it wasn't cleared before we called `ss_skb_process()`, so what we saved to `msglen` was the current record length.
Second: TLS record was entirely in the skb. Then `msglen` will be `0` since previous record was processed and `tls->io_in` was reset previously. Then we processed the current record, then `tls->io_in` was reset again. So there is no way to get record length from `tls->io_in`. But fortunately, we have `parsed`, which we use to infer the current record length.

This is a shady part in the patch. It's far from obvious. The possible solution was to change `tls->io_in` processing to clear it before parsing, not after. But that was going to complicate the patch more, so I stopped there.

tempesta_fw/http_frame.c
```diff
@@ -1655,27 +1655,27 @@ tfw_h2_frame_process(void *c, TfwFsmData *data)
...
-
-	curr_tail = off + parsed + tail < skb->len ? 0 : tail;
-	if (r >= T_POSTPONE && ss_skb_chop_head_tail(NULL, skb, off, curr_tail))
+	if (r >= T_POSTPONE &&
+	    ss_skb_chop_head_tail(NULL, skb, off, skb->len - off - data_left))
 	{
 		r = T_DROP;
 		goto out;
 	}
+	data_left -= parsed;
```
aleksostapenko: It seems that we also should set `data_up.len` to `parsed` value below - before `tfw_http_msg_process_generic()` call.

tempesta_fw/http.c
```diff
@@ -3383,8 +3384,8 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, const TfwFsmData *data)
 	 * feed the new data version to FSMs registered on our states.
 	 */
 	data_up.skb = skb;
-	data_up.off = off;
-	data_up.trail = curr_skb_trail;
+	data_up.off = 0;
```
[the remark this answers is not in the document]
aleksostapenko: Yes, agree with @ikoveshnikov on both points.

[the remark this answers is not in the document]
i-rinat: That's confusing.
In the code below there are calls of `tfw_http_chop_skb` before calls of `tfw_gfsm_move()`. The one in the `TFW_POSTPONE` case is what I recently added, but the one in the `TFW_PASS` (falls through the switch-case) was there before. So `TfwFsmData` instance was passed with `off` set to possibly non-zero value. But at the same time, `tfw_http_chop_skb()` ensured that that `off` bytes were cut off.

So... what `data_up.off` should actually be? Is it `0` or `off`?

Similarly, `data_up.trail = curr_skb_trail;` confuses me. As `curr_skb_trail` can be either `trail` (for the last message in skb) or `0`, `data_up` got even the trailing data which it shouldn't be getting. But it all worked somehow. I wonder how.

I was hoping to avoid digging further into the overall data processing. But it looks inevitable now. :confused:

tempesta_fw/http_frame.c
i-rinat: Thanks for pointing this out. I'll dig deeper into the `data_up` processing.

tempesta_fw/http.c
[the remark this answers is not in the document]
aleksostapenko: >But it all worked somehow. I wonder how.

It seems, that was a hidden bug. Perhaps it is worth to move `tfw_http_chop_skb()` before `data_up` setting (like in `tfw_tls_msg_process()` or in `tfw_h2_frame_process()`) - to deal with offsets and trails from the previous layer at first, and after that - prepare `data_up` for further transferring upstairs. In this variant `data_up.off` really should be 0 here, and `data_up.len` should be set to `parsed`.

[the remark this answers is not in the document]
ikoveshnikov: >  Perhaps it is worth to move tfw_http_chop_skb() before data_up setting 

Sounds like a good idea. All that offsets and lengths create an easy way to get lost and garbage data.

> But it all worked somehow. I wonder how.

Actually it's OK for now to have invalid `off` and `len` members values in `data_up`. Since there is no code that uses them (: The only FSM called after HTTP is Frang (http_limits), it doesn't work with that members. Even if there is issues, we don't see them.

The issue with incorrect trail in `tfw_http_chop_skb()` can be also hidden by `ss_skb_split()` which is called right after the `tfw_http_chop_skb()`. Not sure this really happens, but it also cat chop extra data after request end...

tempesta_fw/tls.c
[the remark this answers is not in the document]
krizhanovsky: OK, I see.

This code takes control only for application data records and it'd be good to have appropriate `WARN_ON()` by the end of the `switch` statement.

I'm still not sure about the `else` branch: if we have 2 chunks of data - e.g. 4 bytes of the header and the rest of the record - `msglen` on receiving the second chunk is still 0 since we didn't read the header fully, but `parsed` contains only length of the second chunk and the first 4 bytes seem just lost in the computation.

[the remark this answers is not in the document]
i-rinat: > if we have 2 chunks of data - e.g. 4 bytes of the header and the rest of the record

You are right, that's the bug. Thanks!

tempesta_fw/http.c
```diff
@@ -3367,24 +3368,39 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, const TfwFsmData *data)
...
-	curr_skb_trail = off + parsed + trail < skb->len ? 0 : trail;
+	data_left -= parsed;
 	TFW_ADD_STAT_BH(parsed, clnt.rx_bytes);
 
 	T_DBG2("Request parsed: len=%u next=%pK parsed=%d msg_len=%lu"
 	       " ver=%d res=%d\n",
 		 skb->len, skb->next, parsed, req->msg.len, req->version, r);
 
+	switch (r) {
+	case TFW_POSTPONE:
```
krizhanovsky: Why do we chop the message on receiving *each* chunk? I believe we should chop it only once.

Besides the logic issue, the construction is quite ugly - why not to use `if (r == TFW_POTPONE || r == TFW_PASS)`?

tempesta_fw/http.c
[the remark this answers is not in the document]
krizhanovsky: Yes, there is some mess with message (HTTP requests and TLS application data records) chopping and the code redesign is required.

`tfw_http_chop_skb()` and `tfw_tls_chop_skb_rec()` seems do very close things, but do them on different layers. Probably, we need only one of them and/or can unify the functions. I'd suppose that chopping must be called by TLS layer only since it has knowledge about TLS records. `tfw_http_chop_skb()` just uses the knowledge passed from TLS layer.

Initially `TfwFsmData` `off` and `trail` (now `len`) were introduced to make HTTP and TLS layers independent: TLS knows about record header and trailer and calls HTTP FSM with explicit description which data must be ignored. If we chop data, then why do we need `off` and `len`? The things are even more simple with that we assemble full TLS record and know its header and trailer before decryption and processing on HTTP level. Probably, with proper chopping, we can simplify `TfwFsmData` structure.

tempesta_fw/tls.c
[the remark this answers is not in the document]
i-rinat: I've moved `msglen` field before `__initoff` so it survives "the purge", and can be used without tricks.

tempesta_fw/http.c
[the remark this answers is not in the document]
i-rinat: > Perhaps it is worth to move tfw_http_chop_skb() before data_up setting

Moved it to the earlier lines.

tempesta_fw/http.c
i-rinat: Looks safe. Changed to `__kfree_skb()`.

tempesta_fw/http_frame.c
i-rinat: Added.

tempesta_fw/http.c
i-rinat: > Why do we chop the message on receiving each chunk?

Legacy. :smiley: 
As you mentioned earlier, we should probably completely remove the need to chop skbs here by preparing them earlier. But at the moment there are offsets, and we address them this way.

> why not to use if

Changed to `if`. I was trying to mimic the switch-case below for easier comparison.

i-rinat:
There was an off-site discussion that made us to decide that we need to remove `off` and `trail` fields from `TfwFsmData` structure. It will only hold `skb` and whole skb data are now the usable data. Headers and trailers that... happen, are removed as soon as possible.

tempesta_fw/http.c
i-rinat: There were changes that removed the need to handle heading and trailing parts here, so this discussion is no longer relevant.

tempesta_fw/http_frame.c
i-rinat: There were changes that removed the need to handle heading and trailing parts of messages. It's no longer required to track amount data available in the instances of `TfwFsmData`, so the code mentioned in this discussion thread was removed.

tempesta_fw/http.c
i-rinat: This code was extracted into #1349 and merged.

tempesta_fw/http.c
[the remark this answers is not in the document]
i-rinat: There were changes that removed `off` and `trail`. They are no longer cut here, and instead are cut in the TLS layer.

Issue #1064, opened 2018-09-09
TLS: further performance improvements and cleanups


krizhanovsky:
Changed for #614 grow significantly, so following tasks are move from #614 scope:

- [ ] bignum, ECDSA, RSA assembly implementations and optimizations. Cleanup list of supported curves and algorithms (see and update `tls/test_tls_cert.py` if necessary), also update https://github.com/tempesta-tech/tempesta/wiki/Tempesta-TLS
- [x] remove `dummy_headers` and replace GCC SIMD intrinsics with assembly code
- [ ] adjust code according to RFC 8422 which obsoletes RFC 4492
- [ ] remove ifdef (w/ or w/o content) for `TTLS_PK_PARSE_EC_EXTENDED`, SECG SEC 1.
- [ ] need to measure current performance of TLS handshakes with web cache and pure LB/proxy modes, update #391 and related issues
- [ ] test and probably replace standard `memcpy()` calls from `skb_copy_*()` calls by `fast_memcpy()`. Probably other standard functions using `memcpy()`, `memset()` or `memcmp()` can be accelerated.
- [ ] Security sensitive branching must be analyzed and protected against Meltdown/Spectre vulnerabilities.

Not part of the issue, but Linux crypto API also has performance issues, so our effort to get zero-copy TLS starves on the underlying API. E.g. `gcmaes_encrypt()` in most cases goes through `kmalloc()` path with **2(!) data copies**.

### Testing

The new crypto routines must be unit tested (see `mbedtls/crypto/tests/suites/`).

krizhanovsky:
Current profile under

```# ./src/thc-ssl-dos --accept -c ECDHE-ECDSA-AES128-GCM-SHA256 -l 1000 127.0.0.1 443```

with configuration 

```
listen 443 proto=https;
server 127.0.0.1:8080;

cache 1;
cache_fulfill * *;

tls_certificate /root/tempesta/etc/tfw-root.crt;
tls_certificate_key /root/tempesta/etc/tfw-root.key;
# wrk sends IP address in SNI, so we test the option here.
tls_match_any_server_name;
```

```
    12.12%  [kernel]        [k] memset_erms
    10.69%  [tempesta_tls]  [k] ecp_mod_p256
     7.10%  [kernel]        [k] __kmalloc
     5.58%  [kernel]        [k] kfree
     5.08%  [kernel]        [k] memcpy_erms
     4.99%  [tempesta_tls]  [k] mpi_mul_hlp
[...]

PR #1331, opened 2019-08-08
Fix possible memory corruption.

krizhanovsky:
TlsHandshake->tmp_sha256 resides in the same union with ecdh_ctx, so
ttls_write_server_key_exchange() -> ttls_ecp_group_load() may find
dirty memory in tls->hs->ecdh_ctx.grp and ttls_ecp_group_free() calls
kfree on bad pointers.

Fix for current 0.7 will be done in context of #1064.


krizhanovsky:
`ttls_ecdh_init()` initializes the context, so we're good and there is no need for the patch.

PR #1339, opened 2019-08-14
make tfw_http_msg_setup() accept skb flags

i-rinat:
When serving responses from the cache, we need to mark generated skbs with the `SKBTX_SHARED_FRAG` flag to avoid cache pages corruption. Rather than traversing the skb chain, let's set skb flags right after skb allocation.

This patch removes the loop mentioned in https://github.com/tempesta-tech/tempesta/pull/1330#discussion_r311292965.


avbelov23:
LGTM

PR #1342, opened 2019-08-21
gfsm: fix possible value overlaps in shift calculations

i-rinat:
There was a possible overlapping in the fsm_hooks subarray index calculations.
As states are in the [0, 31] closed interval, for each priority increase we need to add 32, which is `TFW_GFSM_STATE_N`. Smaller values can cause expected behavior, and the previous value was 16.


i-rinat:
> we didn't caught the issue before, since all the hooks was added with highest priority (which is 0), so calculations results the same before and after patch.

Indeed so. And it's not likely we'll ever need other priorities. But it won't hurt to have a correct multiplier anyway.

Issue #1344, opened 2019-08-21
Inaccurate Content-Type header parsing


ikoveshnikov:
## Scope

The `Content-Type` header is not parsed correctly, part of the header is missing in it's `TfwStr`. To reproduce send a request with the header and try to print the header `req->h_tbl->tbl[TFW_HTTP_HDR_CONTENT_TYPE]`. Part of the header is missed, e.g. `Content-Type: text/plain` turns into `Content-Type: /plain`  

## Testing

The tests was added in `ik-ct-parse-err` branch, very trivial but the issue is caught, see #1343 

Output (part of it) on that branch:
```sh
...
[  997.919936] tfw_test: === request: [POST / HTTP/1.1
               Host: localhost.localdomain
               Content-Type: text/plain; name="value"; 
               Content-Length: 0
               Keep-Alive: timeout=98765
               
               ]
[  997.922889] tfw_test: /home/user/qtc/release/tempesta/tempesta_fw/t/unit/test_http_parser.c:2603: error: 
[  997.924114] tfw_test: FAIL:
[  997.924452] tfw_test:   test__http_parser__content_type_line_parser()
[  997.925239] tfw_test:   EXPECT_EQ((true), (tfw_str_eq_cstr(&req->h_tbl->tbl[TFW_HTTP_HDR_CONTENT_TYPE], "Content-Type: text/plain; name=\"value\"; ", strlen("Content-Type: text/plain; name=\"value\"; "), 0))) => (0x1, 0x0)
[  997.927579] tfw_test: /home/user/qtc/release/tempesta/tempesta_fw/t/unit/test_http_parser.c:2603: error: 
....
[  998.599495] tfw_test: === request: [POST / HTTP/1.1
               Host: localhost.localdomain
               Content-Type: text/plain; name="val\"ue" 
               Content-Length: 0
               Keep-Alive: timeout=98765
               
               ]
[  998.602289] tfw_test: /home/user/qtc/release/tempesta/tempesta_fw/t/unit/test_http_parser.c:2612: error: 
[  998.603577] tfw_test: FAIL:
[  998.603907] tfw_test:   test__http_parser__content_type_line_parser()
[...]

PR #1343, opened 2019-08-21
tests: http parser: catch content-type parsing errors

ikoveshnikov:
The unit tests will fail on this branch. Don't merge until the issue is fixed.

PR #1345, opened 2019-08-22
Saving not multipart content type

avbelov23:
#1344 

If there is one token in the Content-type value, then it is not saved.
Also, part of the value is not saved until the slash.


tempesta_fw/http_parser.c
```diff
@@ -1657,10 +1657,9 @@ __req_parse_content_type(TfwHttpMsg *hm, unsigned char *data, size_t len)
 	__FSM_STATE(I_ContTypeOtherType) {
 		__FSM_I_MATCH_MOVE_fixup(token, I_ContTypeOtherType, 0);
 		if (IS_CRLF(*(p + __fsm_sz))) {
-			p += __fsm_sz;
-			goto finalize;
+			__FSM_I_MOVE_fixup(I_EoL, __fsm_sz, 0);
```
i-rinat: Looks like we have no unit-test covering this branch yet (and #1343 doesn't seem to add any).

avbelov23: DONE

PR #1348, opened 2019-09-05
Some tls fixes

### LINKED ISSUE #1324: Multiple crashes on chunked TLS traffic
Current master as of 1b4f36d2e86d4b8fe8804baad3357570ea80927b. I also tested chunks 2, 3, 4, 5, 6, 7, 8, 11 and 64 bytes - the similar crashes.

### 1Byte chunks

```
# ./run_tests.py -n tls.test_tls_handshake.TlsHandshakeTest.test_1byte_transfer
```
produces crash:
```
[  313.296256] kernel BUG at /root/tempesta/tls/tls_srv.c:992!
[  313.297290] invalid opcode: 0000 [#1] SMP PTI
[  313.298097] Modules linked in: tempesta_fw(O) tempesta_db(O) tempesta_tls(O) tempesta_lib(O) binfmt_misc bochs_drm ttm drm_kms_helper crct10dif_pclmul ppdev drm parport_pc fb_sys_fops parport sg syscopyarea sysfillrect sysimgblt button serio_raw pcspkr ip_tables x_tables ext4 crc16 mbcache jbd2 fscrypto sr_mod cdrom sd_mod ata_generic ata_piix libata psmouse i2c_piix4 e1000 scsi_mod
[  313.303814] CPU: 0 PID: 978 Comm: python2 Tainted: G           O    4.14.32-kdump+ #142
[  313.306301] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.12.0-2.fc30 04/01/2014
[  313.308115] task: ffff9a8eaf565040 task.stack: ffff9a8ebff28000
[  313.309293] RIP: 0010:ttls_handshake_server_step+0x185f/0x1cb0 [tempesta_tls]
[  313.310728] RSP: 0018:ffff9a8effc03980 EFLAGS: 00010297
[  313.311781] RAX: 000000000000000c RBX: ffff9a8ebffb9ab8 RCX: ffff9a8ebffbb000
[  313.313344] RDX: 0000000000000001 RSI: 0000000000000000 RDI: 00000000000000a9
[  313.314915] RBP: ffff9a8ec01cb2e0 R08: ffff9a8ec01cb2e1 R09: 00000000000000a9
[  313.316603] R10: 0000000000000016 R11: 00000000ff95d800 R12: ffff9a8effc03ad4
[  313.318131] R13: 0000000000000000 R14: ffff9a8ebffbb000 R15: 0000000000000001
[  313.319574] FS:  00007f5e7385c700(0000) GS:ffff9a8effc00000(0000) knlGS:0000000000000000
[  313.320961] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[  313.321955] CR2: 00007f5e736ab7e8 CR3: 0000000059b5a004 CR4: 00000000003606f0
[  313.323374] Call Trace:
[  313.323812]  <IRQ>
[  313.324211]  ? crypto_acomp_report+0xa0/0xa0
[  313.324986]  ? netif_rx_internal+0x125/0x130
[  313.325734]  ? loopback_xmit+0xac/0xf0
[  313.326416]  ttls_recv+0x2d3/0x790 [tempesta_tls]
[  313.327245]  ? ttls_decrypt+0x580/0x580 [tempesta_tls]
[  313.328186]  ss_skb_process+0x187/0x1f0 [tempesta_fw]
[  313.329072]  tfw_tls_msg_process+0xe3/0x350 [tempesta_fw]
[  313.330144]  ? tfw_tls_msg_process+0x84/0x350 [tempesta_fw]
[  313.331365]  __gfsm_fsm_exec+0x51/0x90 [tempesta_fw]
[  313.332277]  tfw_connection_recv+0x49/0x70 [tempesta_fw]
[  313.334328]  ? tfw_connection_send+0x20/0x20 [tempesta_fw]
[  313.337178]  ss_tcp_process_data+0x1d6/0x440 [tempesta_fw]
[  313.339563]  ss_tcp_data_ready+0x3e/0x80 [tempesta_fw]
[  313.341552]  tcp_rcv_established+0x4cd/0x570
[  313.343330]  tcp_v4_do_rcv+0x124/0x1c0
[  313.344916]  tcp_v4_rcv+0x93f/0xa40
[  313.346570]  ip_local_deliver_finish+0x95/0x1c0
[  313.348363]  ip_local_deliver+0x66/0xe0
[  313.349971]  ? nf_hook_slow+0x3e/0xc0
[  313.351562]  ip_rcv+0x284/0x3b0
[  313.352850]  ? inet_del_offload+0x30/0x30
[  313.354252]  __netif_receive_skb_core+0x84a/0xb30
[  313.355903]  ? enqueue_task_fair+0x92/0x5d0
[  313.357493]  ? process_backlog+0x92/0x120
```

### 10Byte chunks
```
# ./run_tests.py -n tls.test_tls_handshake.TlsHandshakeTest.test_10byte_transfer
```
crashes Tempesta with:
```
[   51.976508] kernel BUG at /root/tempesta/tls/ttls.c:139!
[   51.977521] invalid opcode: 0000 [#1] SMP PTI
[   51.978382] Modules linked in: tempesta_fw(O) tempesta_db(O) tempesta_tls(O) tempesta_lib(O) binfmt_misc bochs_drm ttm drm_kms_helper drm crct10dif_pclmul sg fb_sys_fops syscopyarea sysfillrect sysimgblt ppdev button serio_raw pcspkr parport_pc parport ip_tables x_tables ext4 crc16 mbcache jbd2 fscrypto sr_mod cdrom sd_mod ata_generic ata_piix libata psmouse scsi_mod e1000 i2c_piix4
[   51.984444] CPU: 0 PID: 898 Comm: python2 Tainted: G           O    4.14.32-kdump+ #142
[   51.985872] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.12.0-2.fc30 04/01/2014
[   51.987420] task: ffff9ca3c1f06140 task.stack: ffff9ca3af45c000
[   51.988566] RIP: 0010:ttls_decrypt+0x531/0x580 [tempesta_tls]
[   51.989619] RSP: 0018:ffff9ca3ffc039e0 EFLAGS: 00010206
[   51.990565] RAX: 00000000000001f4 RBX: ffff9ca3dc4f6370 RCX: ffff9ca3dc4f63ad
[   51.991913] RDX: 000000000000000d RSI: 0100000000000000 RDI: ffff9ca3dc4f66df
[   51.993637] RBP: 0000000000000244 R08: ffff9ca3dc4f63a5 R09: ffff9ca3af465cc0
[   51.995165] R10: 0000000000000000 R11: 0000000000000000 R12: 0000000000000010
[   51.996661] R13: 0000000000000008 R14: ffff9ca3d6186a00 R15: ffff9ca3ca2f7b00
[   51.998280] FS:  00007f6c070f3700(0000) GS:ffff9ca3ffc00000(0000) knlGS:0000000000000000
[   52.000193] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[   52.001333] CR2: 00007f6c00cacd70 CR3: 0000000042076002 CR4: 00000000003606f0
[   52.003575] Call Trace:
[   52.004402]  <IRQ>
[   52.005025]  ? loopback_xmit+0xac/0xf0
[   52.006076]  ttls_recv+0xe5/0x790 [tempesta_tls]
[   52.007104]  ? ttls_decrypt+0x580/0x580 [tempesta_tls]
[   52.008395]  ss_skb_process+0x187/0x1f0 [tempesta_fw]
[   52.009591]  ? lock_timer_base+0x6f/0x90
[   52.010479]  tfw_tls_msg_process+0xe3/0x350 [tempesta_fw]
[   52.011652]  ? tfw_tls_msg_process+0x84/0x350 [tempesta_fw]
[   52.012930]  __gfsm_fsm_exec+0x51/0x90 [tempesta_fw]
[   52.013866]  tfw_connection_recv+0x49/0x70 [tempesta_fw]
[   52.015948]  ? tfw_connection_send+0x20/0x20 [tempesta_fw]
[   52.017960]  ss_tcp_process_data+0x1d6/0x440 [tempesta_fw]
[   52.020029]  ss_tcp_data_ready+0x3e/0x80 [tempesta_fw]
[   52.022307]  tcp_rcv_established+0x4cd/0x570
[   52.023979]  tcp_v4_do_rcv+0x124/0x1c0
[   52.025521]  tcp_v4_rcv+0x93f/0xa40
[   52.026861]  ip_local_deliver_finish+0x95/0x1c0
[   52.028762]  ip_local_deliver+0x66/0xe0
[   52.030832]  ? nf_hook_slow+0x3e/0xc0
[   52.032694]  ip_rcv+0x284/0x3b0
[   52.035990]  ? inet_del_offload+0x30/0x30
[   52.038633]  __netif_receive_skb_core+0x84a/0xb30
[   52.040849]  ? enqueue_task_fair+0x92/0x5d0
[   52.042540]  ? process_backlog+0x92/0x120
```

### 9Byte chunks
Doesn't crash the system, but still doesn't work properly:
```
# ./run_tests.py -n tls.test_tls_handshake.TlsHandshakeTest.test_9byte_transfer
...
Running deproxy
Running deproxy server manager
	Deproxy: Server: Start on 127.0.0.1:8000.
	Starting TempestaFW on localhost
ERROR
	Teardown
	Stopping TempestaFW on localhost
	Deproxy: Server: Stop on 127.0.0.1:8000.
Stopping deproxy
Finished deproxy manager

======================================================================
ERROR: test_9byte_transfer (tls.test_tls_handshake.TlsHandshakeTest)
  File "/root/tempesta-test/tls/test_tls_handshake.py", line 75, in test_9byte_transfer
    res = TlsHandshake(addr='127.0.0.1', port=443, chunk=9).do_12()
  File "/root/tempesta-test/tls/handshake.py", line 348, in do_12
    return self._do_12_req(fuzzer)
  File "/root/tempesta-test/tls/handshake.py", line 329, in _do_12_req
    resp = self.send_recv(tls.TLSPlaintext(data=req))
  File "/root/tempesta-test/tls/handshake.py", line 174, in send_recv
    raise tls.TLSProtocolError(sock_except, pkt, resp)
TLSProtocolError: [Errno 32] Broken pipe

Ran 1 test in 2.772s

FAILED (errors=1)
```
and there is warning in the dmesg:
```
Warning: Unrecognized TLS receive return code -32384, drop packet
```

avbelov23:
fix #1324
- add custom macro of jump into state for handshake with initialization of work in a new state otherwise the crash on `BUG_ON(io->rlen > ext_sz)` in `TTLS_CH_HS_EX` state.
-  In function `ttls_crypto_req_sglist()` `skb->len` is only the length of the first `skb_buff` in `io->skb_list'. `len` is the length of all the data that needs to be decrypted, as in `io->skb_list', as in `buf`.
- if after parsing headers not only handshake header was read, we compute handshake checksum for the message body and handshake header in one shot later

The patch (https://github.com/tempesta-tech/tempesta/pull/1330) is also needed

### DIFF
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -2205,7 +2205,9 @@ next_record:
 		if (io->msgtype == TTLS_MSG_HANDSHAKE
 		    && ttls_hs_checksumable(tls))
 		{
-			if (likely(*read - parsed >= TTLS_HS_HDR_LEN)) {
+			if (likely(*read - parsed >= TTLS_HS_HDR_LEN &&
+			           len > *read - parsed))
+			{
 				/*
 				 * Compute handshake checksum for the message
 				 * body and handshake header in one shot.
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -892,8 +892,12 @@ __ttls_decrypt(TlsCtx *tls, unsigned char *buf)
 	memcpy_fast(xfrm->iv_dec + xfrm->fixed_ivlen, io->iv, sizeof(io->iv));
 	req = ttls_crypto_req_sglist(tls, tfm, dec_msglen + taglen, buf,
 				     &sg, &sgn);
-	if (!req || WARN_ON_ONCE(sgn < 2))
+	if (!req)
 		return TTLS_ERR_INTERNAL_ERROR;
+	if (WARN_ON_ONCE(sgn < 2)) {
+		r = TTLS_ERR_INTERNAL_ERROR;
+		goto out;
+	}
 	ttls_make_aad(tls, io, aad_buf);
 	sg_set_buf(sg, aad_buf, TLS_AAD_SPACE_SIZE);
 
@@ -921,6 +925,7 @@ __ttls_decrypt(TlsCtx *tls, unsigned char *buf)
 	if (unlikely(++io->ctr > (~0UL >> 1)))
 		T_WARN("incoming message counter would wrap\n");
 
+out:
 	kfree(req);
 
 	return r;
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -141,6 +141,9 @@ ttls_crypto_req_sglist(TlsCtx *tls, struct crypto_aead *tfm, unsigned int len,
 	unsigned int sz, aead_sz, to_read, off;
 	int n;
 
+	WARN_ON_ONCE(len == 0); /* nothing to decrypt */
+	WARN_ON_ONCE(!buf && !skb);
+
 	sz = aead_sz = sizeof(*req) + crypto_aead_reqsize(tfm);
 	if (buf) {
 		off = 0;
@@ -149,7 +152,6 @@ ttls_crypto_req_sglist(TlsCtx *tls, struct crypto_aead *tfm, unsigned int len,
 		off = io->off;
 		n = *sgn + io->chunks;
 	}
-	BUG_ON(!buf && (!skb || skb->len <= off)); /* nothing to decrypt */
 	sz += n * sizeof(**sg);
 
 	/* Don't use g_req for better spacial locality. */
--- a/tls/tls_internal.h
+++ b/tls/tls_internal.h
@@ -391,6 +391,14 @@ do {									\
 	T_FSM_MOVE(st, {});						\
 } while (0)
 
+/* Unconditional jump to state @st w/o additional logic and/or eating data. */
+#define TTLS_HS_FSM_JMP(st)						\
+do {									\
+	io->rlen = 0;							\
+	state_p = p;							\
+	T_FSM_JMP(st);							\
+} while (0)
+
 /*
  * Size of temporary storage in TlsCtx->hs->tmp.
  * The temporary buffer is used to store 2 types of temporary data: utility
--- a/tls/tls_srv.c
+++ b/tls/tls_srv.c
@@ -953,7 +953,7 @@ ttls_parse_client_hello(TlsCtx *tls, unsigned char *buf, size_t len,
 		if (n)
 			TTLS_HS_FSM_MOVE(TTLS_CH_HS_EX);
 		else
-			T_FSM_JMP(TTLS_CH_HS_EX);
+			TTLS_HS_FSM_JMP(TTLS_CH_HS_EX);
 	}
 
 	/* Parse an extension. */


tls/ttls.c
```diff
@@ -2203,8 +2204,14 @@ ttls_recv(void *tls_data, unsigned char *buf, size_t len, unsigned int *read)
 	}
 	WARN_ON_ONCE(!io->msglen);
 	delta = *read - parsed;
-	if (delta == len)
+	if (delta == len) {
+		if (io->msgtype == TTLS_MSG_HANDSHAKE
+		    && ttls_hs_checksumable(tls))
+		{
+			ttls_update_checksum(tls, buf + delta - hh_len, hh_len);
+		}
```
krizhanovsky: The fix is correct, but current code with checksumming and chunks is tricky and the fix makes it even more complicated. Here we care about handshake messages only, unencrypted since encryption takes place on CLIENT_FINISHED only and this state is not checksumable. So `ttls_parse_record_hdr()` returns `T_OK` only when it fully reads both the TLS and handshake headers and each of them can be split into several chunks.

If in current call we've read a chunk containing only tail of handshake header chunk, them we do calculate checksum above. We also go here with zero `hh_len`, i.e. do not do any checksumming actually. This is unclear and adds extra call without any actual work.

If current chunk contains the full handshake header, then we're going to compute it's checksum later, when we read more data after the header. But, yes, if we have no more data, then we do not compute checksum and we need the fix.

So to make the fix more clear I'd move the checksumming with `*read - parsed == len` condition into the double if above. And extend the comment with words that we go here if we read a chunk fully containing handshake header and probably we reached end of data, so we need to compute the checksum now.

tls/ttls.c
```diff
@@ -2203,15 +2204,19 @@ ttls_recv(void *tls_data, unsigned char *buf, size_t len, unsigned int *read)
 		if (io->msgtype == TTLS_MSG_HANDSHAKE
 		    && ttls_hs_checksumable(tls))
 		{
-			if (likely(*read - parsed >= TTLS_HS_HDR_LEN)) {
+			if (likely(*read - parsed != len)) {
```
i-rinat: Having `*read - parsed != len` doesn't mean `ttls_parse_record_hdr()` ate whole header and there is still data in the buffer. It means the function ate at least last bytes of a handshake header and there is still payload available.

So, if we get handshake in two pieces, one of one byte and other of everything else, we'll have `*read - parsed != len`, where `*read - parsed` not equal to `TTLS_HS_HDR_LEN`. That will break the checksum calculation trick.

tls/ttls.c
```diff
@@ -149,7 +151,6 @@ ttls_crypto_req_sglist(TlsCtx *tls, struct crypto_aead *tfm, unsigned int len,
 		off = io->off;
 		n = *sgn + io->chunks;
 	}
-	BUG_ON(!buf && (!skb || skb->len <= off)); /* nothing to decrypt */
```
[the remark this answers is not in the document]
i-rinat: > When do we have skb->len == 0?

This BUG_ON was triggering on `skb->len` of 10 bytes while `off` was 13. I don't think the issue was with zero-length skbs. This is the place before skb chain trimming, so it's expected to have `skb->len` less than `off`. The code below handles that case explicitly, so I think the condition was more strict than it needed to be. We still need to ensure that either `buf` or `skb` is non-NULL.

> Probably, this is job for #1330

I'm inclined to keep trimming-related fixes outside this PR, and believe that #1330 (or a separate PR) is a better place for that changes.

tls/ttls.c
avbelov23: No.  `ttls_parse_record_hdr()` returns `T_POSTPONE` if it has not eaten the whole header (`TLS_HEADER_SIZE + TTLS_HS_HDR_LEN`) and does not reach here

i-rinat: Wait a minute, I'm confused.
Say, a single-byte packet arrives. Then `len` is 1, we call `ttls_parse_record_hdr()`, it returns `T_POSTPONE`, we return that to the called. Then, a large packet arrives. To simplify things, let's assume it contains everything else in the TLS record. We pass that to `ttls_parse_record_hdr()`. It increases `*read` by some number, but that number is not `len`, as there is a payload after a header. We go into the "then" branch, set `hh_len` to `TTLS_HS_HDR_LEN`, which subsequently will trigger the checksumming trick. We'll calculate checksum not from `buf`, but from `buf - hh_len`. But we don't have that first single byte in the continuous buffer. It was in the previous skb, not in the current one.

Is that right? Or am I missing something here?

avbelov23: I understand what you mean. You're right. The trick with `hh_len` is the problem

i-rinat: OK.

Also, that `TTLS_HS_HDR_LEN` looks suspicious. I mean, there is the record header, which is `TLS_HEADER_SIZE` bytes long. And then there is handshake header, that takes additional `TTLS_HS_HDR_LEN` bytes. So shouldn't there be sum of those two, that is 9 bytes instead of 5?

avbelov23: Not. We are trying to fix the case when 9 bytes just arrive and they are discarded as we assign 4 (https://github.com/tempesta-tech/tempesta/blob/master/tls/ttls.c#L2211) to the variable and hope to update the checksum later, but exit here (https://github.com/tempesta-tech/tempesta/blob/master/tls/ttls.c#L2221)
But unfortunately the problem is not only in this, but in general in the presence of a variable `hh_len`

i-rinat: Yep, first 9 bytes are the headers. There is no sense in postponing checksum calculation if there will be no checksum calculation on the buffer later. But if `ttls_parse_record_hdr()` returns `9`, and it's a handshake record, and there is more than 9 bytes in the buffer, then there will be further processing, and later we'll calculate checksum. Otherwise there will be no occasion, so we need to update checksum immediately. How about that?

i-rinat: There was an off-site discussion, so I'll put here the bits that were figured out.

We do not checksum TLS header, only handshake header, so we need to check for 4 bytes, which is `TTLS_HS_HDR_LEN`. If we have 4 bytes parsed and there are more data in the buffer, we still can checksum both the header and the payload in a single call. The condition (pre-patch) looks fine. Although it's questionable what will happen if we parsed 4 bytes and there are no bytes left in the buffer. Or what will happen if buffer starts with remaining bits of a TLS header, and the like.

tls/ttls.c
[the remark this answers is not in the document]
ikoveshnikov: There is no need to handle `!buf && !skb` case, no null derefence will happen. The `skb` is used inside the `for` loop, where the condition `skb != NULL` is evaluated before entering the loop. So it's OK. Condition `skb->len <= off` is really buggy here, so I;m generally OK with the change.

But one thing is bothering me. If `off` is outside the skb list (e.g. total size of skbs is 10, while the offset is 13), or if `!buf && !skb = true`, the function will return the `req` to the caller - `__ttls_decrypt()`. In this case it's very possible that `sgn` will be set to `1`. Condition `if (!req || WARN_ON_ONCE(sgn < 2))` in `__ttls_decrypt()` will fail and a error will be generated. But the `req` is not destroyed and memory leak happens. We need to properly release `req` on errors. The leakage is not possible on master since the `BUG_ON` statement is more strict that it's required.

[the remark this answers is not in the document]
ikoveshnikov: > 	WARN_ON_ONCE(len == 0); /* nothing to decrypt */

Seems it never happen. The `ttls_crypto_req_sglist()` is called only from `__ttls_decrypt()`. According to the later: https://github.com/tempesta-tech/tempesta/blob/0483d05ac48d6d55090913ddf874181a13dd9c19/tls/ttls.c#L880-L892

So we get `len = dec_msglen + taglen = io->msglen - expiv_len - taglen + taglen =  io->msglen - expiv_len`. But the condition above guarantee that the `io->msglen >= expiv_len + taglen`, and `taglen` is always positive non-null value. Thus `len >= taglen` and always positive, and `WARN_ON_ONCE(len == 0)`  never happens.

tls/ttls.c
```diff
@@ -141,6 +141,9 @@ ttls_crypto_req_sglist(TlsCtx *tls, struct crypto_aead *tfm, unsigned int len,
 	unsigned int sz, aead_sz, to_read, off;
 	int n;
 
+	WARN_ON_ONCE(len == 0); /* nothing to decrypt */
+	BUG_ON(!buf && !skb);
```
i-rinat: As @ikoveshnikov said, it shouldn't cause a crash, so WARN_ON or WARN_ON_ONCE should be enough. I still like to keep that as a warning, rather than removing completely, as this condition means something went wrong at the caller.

tls/ttls.c
```diff
@@ -2198,12 +2200,17 @@ ttls_recv(void *tls_data, unsigned char *buf, size_t len, unsigned int *read)
...
+		unsigned char old_hdr_cpsz = io->hdr_cpsz;
 		if ((r = ttls_parse_record_hdr(tls, buf, len, read)))
 			return r;
 		if (io->msgtype == TTLS_MSG_HANDSHAKE
 		    && ttls_hs_checksumable(tls))
 		{
-			if (likely(*read - parsed >= TTLS_HS_HDR_LEN)) {
+			if (likely(old_hdr_cpsz <= TLS_HEADER_SIZE &&
+			           old_hdr_cpsz + *read - parsed >
+			                   TLS_HEADER_SIZE + TTLS_HS_HDR_LEN
```
i-rinat: I believe the second part is always false, and that effectively turns the checksumming trick off. Still, it should work that way, but we'll play safely then.

Instead, I'd like to propose this way:
```patch
+                       if (likely(*read - parsed >= TTLS_HS_HDR_LEN &&
+                                  len > *read - parsed))
```
We keep the first part of the condition, as it's true when we've just finished header parsing, and at least handshake header is in the continuous buffer. The second part is the addition. We need to have at least one more byte in the continuous buffer to postpone checksum calculation. Otherwise we exit in the `if (delta == len)` below, and never factor the handshake header into a checksum.

PR #1351, opened 2019-09-19
parsing accept header according to rfc

### LINKED ISSUE #1203: Inaccurate Accept parsing
From #1050: while we don't know all the possible content types we may only check syntax of the header and the RFC [requires](https://tools.ietf.org/html/rfc7231#section-5.3.2) `/` in the type, so `Req_I_AcceptOther` must be fixed. `invalid/foo`, `invalid/*` - are allowed, `invalid/` or `invalid` - aren't.

Also please update the HTTP parser unit test by the Date cases and select appropriate checkboxes in #1059 corresponindinf to date headers. No need to port the fix to 0.6.

avbelov23:
Fix #1203

RFC: https://tools.ietf.org/html/rfc7231#section-5.3.2

The MIME-type must have a type and subtype.
In addition, we validate the parameter q. We do not validate all other parameters.

Also added and fixed tests. In previous tests, there were incorrect values. Two slashes are clearly not the correct value, like the lack of a slash for the Accept.

### DIFF
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -1812,11 +1812,54 @@ __req_parse_accept(TfwHttpReq *req, unsigned char *data, size_t len)
 
 	__FSM_START(parser->_i_st);
 
+	__FSM_STATE(Req_I_WSAccept) {
+		if (IS_WS(c))
+			__FSM_I_MOVE(Req_I_WSAccept);
+		/* Fall through. */
+	}
+
 	__FSM_STATE(Req_I_Accept) {
-		TRY_STR("text/html", Req_I_Accept, Req_I_AcceptHtml);
-		TRY_STR("*/*", Req_I_Accept, Req_I_AcceptHtml);
+		TRY_STR("text", Req_I_Accept, Req_I_AfterText);
+		/*
+		 * TRY_STR() compares the string with the substring at the
+		 * beginning of the chunk sequence, but @c is the first
+		 * non-matching character with the string of the previous
+		 * TRY_STR(). If we will use @c to compare with "*", then we will
+		 * catch matches not only with "*", but also with "t*", "te*",
+		 * "tex*".
+		 */
+		TRY_STR("*", Req_I_Accept, Req_I_AfterStar);
 		TRY_STR_INIT();
-		__FSM_I_JMP(Req_I_AcceptOther);
+		if (IS_TOKEN(c))
+			__FSM_I_JMP(Req_I_Type);
+		return CSTR_NEQ;
+	}
+
+	__FSM_STATE(Req_I_AfterText) {
+		if (c == '/')
+			__FSM_I_MOVE(Req_I_AfterTextSlash);
+
+		__FSM_I_MOVE(Req_I_Type);
+	}
+
+	__FSM_STATE(Req_I_AfterTextSlash) {
+		TRY_STR("html", Req_I_AfterTextSlash, Req_I_AcceptHtml);
+		TRY_STR_INIT();
+		__FSM_I_JMP(Req_I_Subtype);
+	}
+
+	__FSM_STATE(Req_I_AfterStar) {
+		if (c == '/')
+			__FSM_I_MOVE(Req_I_StarSlashStar);
+		if (IS_TOKEN(c))
+			__FSM_I_JMP(Req_I_Type);
+		return CSTR_NEQ;
+	}
+
+	__FSM_STATE(Req_I_StarSlashStar) {
+		if (c == '*')
+			__FSM_I_MOVE(Req_I_AcceptHtml);
+		return CSTR_NEQ;
 	}
 
 	__FSM_STATE(Req_I_AcceptHtml) {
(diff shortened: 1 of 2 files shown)


tempesta_fw/http_parser.c
```diff
@@ -1844,8 +1921,8 @@ __req_parse_accept(TfwHttpReq *req, unsigned char *data, size_t len)
 			__FSM_I_MOVE(I_EoT);
```
krizhanovsky: It seems we allow `Accept: text/html; q=0.5, , ,,,, text/plain`. Please add the case to the tests. And this is allowed case, see https://tools.ietf.org/html/rfc7230#section-7 .

tempesta_fw/http_parser.c
```diff
@@ -1812,26 +1812,103 @@ __req_parse_accept(TfwHttpReq *req, unsigned char *data, size_t len)
...
+			return __data_off(p + __fsm_sz);
+		}
+		return CSTR_NEQ;
+	}
+
+	__FSM_STATE(Req_I_QValue) {
+		if (('0' <= c && c <= '9') || c == '.')
+			__FSM_I_MOVE(Req_I_QValue);
+		if (IS_WS(c) || c == ',' || c == ';' || IS_CRLF(c))
+			__FSM_I_JMP(I_EoT);
```
krizhanovsky: JMP looks incorrect since it allows `text/plain; q=0.5 *` (without required `,`) since on the space you jump to `IoT`, moves to `IoT` again on the space, and jump on `*` to `Req_I_Accept`. The same problem we had in `Req_I_AcceptHtml` state. Please also fix and add the two test cases for both the states that we do not allow such values.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -1485,6 +1485,40 @@ TEST(http_parser, accept)
 	{
 		EXPECT_TRUE(test_bit(TFW_HTTP_B_ACCEPT_HTML, req->flags));
 	}
+
+	FOR_REQ("GET / HTTP/1.1\r\n"
+		"Accept:  invalid/invalid;  q=0.5;    key=val, */* \r\n"
+		"\r\n")
+	{
+		EXPECT_TRUE(test_bit(TFW_HTTP_B_ACCEPT_HTML, req->flags));
```
ikoveshnikov: The bit shouldn't be set for unknown values. The `TFW_HTTP_B_ACCEPT_HTML` means, that client can process HTMLs, JS code, and we can send JS code to the client and it will evaluate the code and pass our JS challenge.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -1485,6 +1485,40 @@ TEST(http_parser, accept)
...
+		"Accept: invalid\r\n"
+		"\r\n");
+
+	EXPECT_BLOCK_REQ("GET / HTTP/1.1\r\n"
+		"Accept: /invalid\r\n"
+		"\r\n");
+
+	EXPECT_BLOCK_REQ("GET / HTTP/1.1\r\n"
+		"Accept: invalid/\r\n"
+		"\r\n");
```
ikoveshnikov: Some cases are missing and the parsed can't pass them: 
* `text/html; q=0.5`, `text/html ; q=0.5` - those with valid type-sybtype and parameter, but with different number of  OWS between them.
* `*/html` - forbidden by RFC, but allowed by parser (bug).
* `text/html; =0.5` - no token in  parameter, `text/html; q = 0.5` - OWS inside parameter
* `text/html; param="value"`  - quoted-string inside accept extension, the string may include ows and other stuff.

Please, take a look into [RFC](https://tools.ietf.org/html/rfc7231#section-5.3.2) and add all possible Accept header mutations, that can still be parsed successfully.

ikoveshnikov: Mozilla also tells about [Cross-Origin Resource Sharing safety for Accept header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept);
>  additional restriction that values can't contain a CORS-unsafe request header byte: "():<>?@[\]{}, Delete, Tab and control characters: 0x00 to 0x19.

 I don't know, whether it should be taken into account, while parsing the quoted strings. Anyway, the quoted-strings is a good candidate to have attacks in.
> accept-ext = OWS ";" OWS token [ "=" ( token / quoted-string ) ]

 Attacker may use the difference between quoted string processing between Tempesta and backends and use it fot request-splitting attacks.

tempesta_fw/t/unit/test_http_parser.c
ikoveshnikov: The comment is invalid, didn't noticed the `*/*` type at the end.

tempesta_fw/http_parser.c
```diff
@@ -1812,26 +1812,103 @@ __req_parse_accept(TfwHttpReq *req, unsigned char *data, size_t len)
...
 		if (IS_WS(c) || c == ',' || c == ';' || IS_CRLF(c)) {
 			__set_bit(TFW_HTTP_B_ACCEPT_HTML, req->flags);
 			__FSM_I_JMP(I_EoT);
 		}
-		__FSM_I_MOVE(Req_I_AcceptOther);
+		__FSM_I_JMP(Req_I_Subtype);
+	}
+
+	__FSM_STATE(Req_I_Type) {
+		__FSM_I_MATCH_MOVE(token, Req_I_Type);
```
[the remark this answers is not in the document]
ikoveshnikov: My mistake here. discussed in the chat.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -1485,6 +1485,78 @@ TEST(http_parser, accept)
...
+
+	FOR_REQ("GET / HTTP/1.1\r\n"
+		"Accept:  text/html,  invalid/invalid  ;  key=val;   q=0.5 \r\n"
+		"\r\n")
+	{
+		EXPECT_TRUE(test_bit(TFW_HTTP_B_ACCEPT_HTML, req->flags));
+	}
+
+	FOR_REQ("GET / HTTP/1.1\r\n"
+		"Accept: */html\r\n"
```
ikoveshnikov: Why the `*/html` is allowed? RFC prohibits it:
```
media-range    = ( "*/*"
                      / ( type "/" "*" )
                      / ( type "/" subtype )
                      ) *( OWS ";" OWS parameter )
```
No `"*/" subtype` in list.

tempesta_fw/t/unit/test_http_parser.c
```diff
@@ -836,7 +836,7 @@ TEST(http_parser, casesense)
 TEST(http_parser, hdr_token_confusion)
 {
 	FOR_REQ("GET / HTTP/1.1\r\n"
-		"Accept: */*text/html\r\n"
+		"Accept: */html*\r\n"
```
ikoveshnikov: See my other comment, `*/html` is prohibited, the `text/html*` should be also prohibited, since mixing asterisks and tokens is not allowed.

tempesta_fw/t/unit/test_http_parser.c
avbelov23: I did not read the comment correctly, я прочитал `forbidden by RFC, but allowed by browser`

tempesta_fw/t/unit/test_http_parser.c
krizhanovsky: In `Req_I_AcceptOther` state we use `uri` alphabet to check the input. The alphabet (see `str_avx2.S:94`) doesn't allow `DEL`, `TAB`, any control characters (`0x00-0x20`), `<`, `>`, `\`, but does allow `():?@[]`. The alphabet was choosen according to the RFC, so I'm not sure how important to exclude the rest of allowed chars.

tempesta_fw/http_parser.c
```diff
@@ -1812,41 +1812,144 @@ __req_parse_accept(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		if (IS_WS(c) || c == ',' || c == ';')
+			__FSM_I_MOVE_n(I_EoT, __fsm_sz);
 		if (IS_CRLF(c)) {
 			return __data_off(p + __fsm_sz);
 		}
 		return CSTR_NEQ;
 	}
 
+	__FSM_STATE(Req_I_QuotedString) {
+		if (c != '"')
```
ikoveshnikov: We can speed up quoted string matching by adding `__FSM_I_MATCH_MOVE(token, Req_I_QuotedString);` to previous line, or other alphabe will be more useful.  Idea is to grab all allowed symbols in batched mode, but stop right before closing quote.

tempesta_fw/http_parser.c
ikoveshnikov: The comment is not addressed. Jump is not accurate: `,` means that a new type/subtype is expected, `;` - a new parameter is expected.

tempesta_fw/http_parser.c
```diff
@@ -1812,41 +1812,144 @@ __req_parse_accept(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		if (IS_WS(c) || c == ',' || c == ';')
+			__FSM_I_MOVE_n(I_EoT, __fsm_sz);
+		if (IS_CRLF(c)) {
+			return __data_off(p + __fsm_sz);
+		}
+		return CSTR_NEQ;
+	}
+
+	__FSM_STATE(Req_I_QValue) {
+		if (('0' <= c && c <= '9') || c == '.')
```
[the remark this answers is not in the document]
avbelov23: @krizhanovsky

tempesta_fw/http_parser.c
avbelov23: We do not eat the current character now, when we go to `I_EoT` and analyze it in `I_EoT`

tempesta_fw/t/unit/test_http_parser.c
avbelov23: Now we used `token` alphabet. We parse params on `token`=`token`

tempesta_fw/http_parser.c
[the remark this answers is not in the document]
krizhanovsky: `'0' <= c && c <= '9'` is essentially `isdigit()`.

We wanted `q` validation to avoid tricking a target server about content type. `999.888.777.888` is definitely obscure case - I have no idea whether the value will be rejected or treated as `1` or whatever. However, I don't know attack vector using such `q` values (w/o control or `;` like characters) and it doesn't make sense to develop huge FSM here, so I think we're good with these small improvements.

tempesta_fw/http_parser.c
krizhanovsky: The conditions `IS_WS(c) || c == ',' || c == ';' || IS_CRLF(c)` just doubles conditions in `I_EoT` state, so an unconditional jump should be here - there is no sense to check the same `c` against the same condition twice.

tempesta_fw/http_parser.c
```diff
@@ -1812,41 +1812,144 @@ __req_parse_accept(TfwHttpReq *req, unsigned char *data, size_t len)
...
-			__FSM_I_MOVE_n(I_EoT, __fsm_sz + 1);
+		if (c == '=')
+			__FSM_I_MOVE_n(Req_I_ParamValue, __fsm_sz + 1);
+		return CSTR_NEQ;
+	}
+
+	__FSM_STATE(Req_I_ParamValue) {
+		if (c == '\"')
+			__FSM_I_MOVE(Req_I_QuotedString);
+		__FSM_I_MATCH_MOVE(token, Req_I_ParamValue);
```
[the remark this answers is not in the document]
krizhanovsky: That's fine, we didn't care about this before

tempesta_fw/http_parser.c
```diff
@@ -1812,41 +1812,145 @@ __req_parse_accept(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		c = *(p + __fsm_sz);
+		if (IS_WS(c) || c == ',' || c == ';')
+			__FSM_I_MOVE_n(I_EoT, __fsm_sz);
 		if (IS_CRLF(c)) {
 			return __data_off(p + __fsm_sz);
 		}
```
krizhanovsky: Again: these 5 lines are just equal `__FSM_I_JMP(I_EoT);`

tempesta_fw/http_parser.c
```diff
@@ -1812,41 +1812,145 @@ __req_parse_accept(TfwHttpReq *req, unsigned char *data, size_t len)
...
+	}
+
+	__FSM_STATE(Req_I_Subtype) {
+		__FSM_I_MATCH_MOVE(token, Req_I_Subtype);
+		c = *(p + __fsm_sz);
+		if (IS_WS(c) || c == ',' || c == ';')
+			__FSM_I_MOVE_n(I_EoT, __fsm_sz);
+		if (IS_CRLF(c)) {
+			return __data_off(p + __fsm_sz);
+		}
```
krizhanovsky: These 5 lines are just equal `__FSM_I_JMP(I_EoT)`. Moreover, the direct just is better because it doesn't allow sequences like `,;` (you allow them since you eat the separators here and in `I_EoT`).

BTW I asked to add test case for multiple consequent separators in https://github.com/tempesta-tech/tempesta/pull/1351/files#r326119365 , but the new version still has no the test.

PR #1353, opened 2019-09-27
TLS record parsing fixes

### LINKED ISSUE #1318: BUG at ttls_recv() on invalid record
Tempesta master at fabe6755c7049ab446330c8bd522fddb1c1e1165 . Run the fuzzer from (https://github.com/tempesta-tech/tempesta-test/pull/103/commits/4dd72f677c79764e1e1f02a9e2cda04d59c1c79e):
```
# ./run_tests.py -n tls.test_tls_handshake.TlsHandshakeTest.test_fuzzing
```
and watch the crash:
```
[58700.147845] kernel BUG at /root/tempesta/tls/ttls.h:993!
[58700.150499] invalid opcode: 0000 [#1] SMP PTI
[58700.154247] Modules linked in: tempesta_fw(O) tempesta_db(O) tempesta_tls(O) tempesta_lib(O) tcp_diag inet_diag binfmt_misc bochs_drm ttm drm_kms_helper drm ppdev crct10dif_pclmul button fb_sys_fops syscopyarea sysfillrect sg sysimgblt parport_pc serio_raw pcspkr parport ip_tables x_tables ext4 crc16 mbcache jbd2 fscrypto sd_mod sr_mod cdrom ata_generic ata_piix libata i2c_piix4 e1000 psmouse scsi_mod [last unloaded: tempesta_lib]
[58700.154477] CPU: 1 PID: 9282 Comm: python2 Tainted: G        W  O    4.14.32-kdump+ #142
[58700.154477] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.12.0-2.fc30 04/01/2014
[58700.154477] task: ffff932a42576100 task.stack: ffff932a2f110000
[58700.154477] RIP: 0010:ttls_recv+0x756/0x920 [tempesta_tls]
[58700.154477] RSP: 0018:ffff932a7fd039e8 EFLAGS: 00010293
[58700.154477] RAX: 0000000000000017 RBX: ffff932a2f32c370 RCX: 0000000000000000
[58700.154477] RDX: 0000000000000000 RSI: 0000000000000003 RDI: ffff932a7fd15558
[58700.154477] RBP: 0000000000000005 R08: 0000000000000000 R09: 0000000000004b45
[58700.154477] R10: 0000000000000000 R11: 0000000000000001 R12: 0000000000000049
[58700.154477] R13: ffff932a7fd03a6c R14: 0000000000000000 R15: 0000000000000000
[58700.154477] FS:  00007f8acd840700(0000) GS:ffff932a7fd00000(0000) knlGS:0000000000000000
[58700.154477] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[58700.154477] CR2: 00005651bf46d128 CR3: 0000000055c5a004 CR4: 00000000003606e0
[58700.154477] Call Trace:
[58700.154477]  <IRQ>
[58700.154477]  ? ttls_decrypt+0xa50/0xa50 [tempesta_tls]
[58700.154477]  ss_skb_process+0x187/0x1f0 [tempesta_fw]
[58700.154477]  tfw_tls_msg_process+0xe3/0x360 [tempesta_fw]
[58700.154477]  ? tfw_tls_msg_process+0x6a/0x360 [tempesta_fw]
[58700.154477]  __gfsm_fsm_exec+0x6d/0xb0 [tempesta_fw]
[58700.154477]  tfw_connection_recv+0x49/0x70 [tempesta_fw]
[58700.154477]  ? tfw_connection_send+0x20/0x20 [tempesta_fw]
[58700.154477]  ss_tcp_process_data+0x1d6/0x440 [tempesta_fw]
[58700.154477]  ss_tcp_data_ready+0x3e/0x80 [tempesta_fw]
[58700.154477]  tcp_data_queue+0x4f0/0xc50
[58700.154477]  tcp_rcv_established+0x277/0x570
[58700.154477]  tcp_v4_do_rcv+0x124/0x1c0
[58700.154477]  tcp_v4_rcv+0x93f/0xa40
[58700.154477]  ip_local_deliver_finish+0x95/0x1c0
[58700.154477]  ip_local_deliver+0x66/0xe0
[58700.154477]  ? nf_hook_slow+0x3e/0xc0
[58700.154477]  ip_rcv+0x284/0x3b0
[58700.154477]  ? inet_del_offload+0x30/0x30
[58700.154477]  __netif_receive_skb_core+0x84a/0xb30
[58700.154477]  ? process_backlog+0x92/0x120
[58700.154477]  process_backlog+0x92/0x120
[58700.154477]  net_rx_action+0x261/0x3a0
[58700.154477]  __do_softirq+0x104/0x297
[58700.154477]  do_softirq_own_stack+0x2a/0x40
[58700.154477]  </IRQ>
```

i-rinat:
There are two related fixes here.

First one is what triggered BUG_ON mentioned in #1318: we were parsing application data records even if they appear early. That doesn't make any sense, since cipher is not ready yet. Conforming client will never send such record before handshake is completed, so it's safe to completely reject connections where such record is seen too early.

Second one was not mentioned explicitly in #1318, but was found later. We were postponing record processing despite all required data were already received. The data were processed when next skb arrived and enqueued into `tls->io_in.skb_list`. But since that list is cleared after a record is successfully processed, that next skb was thrown away. Later, that results in out-of-buffer accesses during scatter-gather operations inside a cipher.

(part of #1318-related changes)

### DIFF
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -2226,8 +2226,6 @@ next_record:
 	}
 	WARN_ON_ONCE(!io->msglen);
 	delta = *read - parsed;
-	if (delta == len)
-		return T_POSTPONE;
 	len -= delta;
 	buf += delta;
 	parsed = *read;
@@ -2248,6 +2246,8 @@ next_record:
 	case TTLS_MSG_CHANGE_CIPHER_SPEC:
 		/* Parsed as part of handshake FSM. */
 	case TTLS_MSG_HANDSHAKE:
+		if (len == 0)
+			return T_POSTPONE;
 		if (unlikely(tls->state == TTLS_HANDSHAKE_OVER)) {
 			T_DBG("refusing renegotiation, sending alert\n");
 			ttls_send_alert(tls, TTLS_ALERT_LEVEL_FATAL,
@@ -2288,6 +2288,9 @@ next_record:
 			goto skip_record;
 		}
 	}
+
+	if (len == 0)
+		return T_POSTPONE;
 
 	/* After the handshake the crypto context must be ready. */
 	if (unlikely(!ttls_xfrm_ready(tls))) {
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -2330,7 +2330,8 @@ skip_record:
 	bzero_fast(io->__initoff, sizeof(*io) - offsetof(TlsIOCtx, __initoff));
 
 	delta = *read - parsed;
-	WARN_ON_ONCE(delta > len);
+	if (WARN_ON_ONCE(delta > len))
+		return T_DROP;
 	len -= delta;
 	if (len) {
 		buf += delta;
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -1171,6 +1171,8 @@ ttls_parse_record_hdr(TlsCtx *tls, unsigned char *buf, size_t len,
 		 */
 
 	case TTLS_MSG_APPLICATION_DATA:
+		if (unlikely(!ready))
+			return TTLS_ERR_INVALID_RECORD;
 		ivahs_len = ttls_expiv_len(&tls->xfrm);
 		break;
 
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -2282,11 +2282,7 @@ next_record:
 		return r;
 
 	case TTLS_MSG_APPLICATION_DATA:
-		if (!io->msglen) {
-			/* OpenSSL sends empty messages to randomize the IV. */
-			T_DBG("empty application TLS record - skip\n");
-			goto skip_record;
-		}
+		/* Fall through. */
 	}
 
 	if (len == 0)


tls/ttls.c
```diff
@@ -2226,7 +2228,7 @@ ttls_recv(void *tls_data, unsigned char *buf, size_t len, unsigned int *read)
 	}
 	WARN_ON_ONCE(!io->msglen);
 	delta = *read - parsed;
-	if (delta == len)
+	if (delta == len && *read < io->msglen)
```
krizhanovsky: `msglen` doesn't include handshake header size, so don't we need to account for the header size in the condition? Also this is the second fix in the piece of code for the last weeks - the piece is tricky, so please leave a comment in which cases, after header parsing a new record rf reentering the function for the next chunk of a record, we return `TFW_POSTPONE`.

i-rinat: You are right, we need to add TLS_HEADER_SIZE to the `io->msglen` here.
Moreover the way I tried to use `*read` is not right at all. Its value has no sense, only difference matter. That's why `delta` is computed as a delta. But we don't need a delta, we need an absolute value. At first I was going to change condition to `io->rlen + len < io->msglen + TLS_HEADER_SIZE` and that looks kind of OK. There is even similar condition `io->msglen > io->rlen + len` later in the code.

But at the same time, `ttls_parse_record_hdr()` which is called a bit earlier, is changing `io->rlen`, for application data records and for alerts, but latter only if `ttls_xfrm_ready(tls)` returns truth. So `io->rlen + len` cannot be used for the purpose of measuring amount of data read.

Now I doubt the `io->msglen > io->rlen + len` condition is the right one. Either there is a trick of some kind which I don't get yet, or the condition is actually wrong.

tls/ttls.c
```diff
@@ -2287,6 +2289,9 @@ ttls_recv(void *tls_data, unsigned char *buf, size_t len, unsigned int *read)
 		}
 	}
 
+	if (len == 0)
+		return T_POSTPONE;
+
```
ikoveshnikov: Not an issue of this patch, but later at `skip_record` label we have `WARN_ON_ONCE(delta > len)` statement. It's good to catch the issue and throw a warning, but more unpredictable situations may happen in that case. Better to close the connection in that case:
```
if (WARN_ON_ONCE(delta > len))
    return T_BLOCK;
```

tls/ttls.c
```diff
@@ -2246,6 +2246,8 @@ ttls_recv(void *tls_data, unsigned char *buf, size_t len, unsigned int *read)
 	case TTLS_MSG_CHANGE_CIPHER_SPEC:
 		/* Parsed as part of handshake FSM. */
 	case TTLS_MSG_HANDSHAKE:
+		if (len == 0)
+			return T_POSTPONE;
```
krizhanovsky: Ok, we received only header of a handshake record, return here and having `TTLS_F_ST_HDRIV` in flags we won't read a record header again. Won't we be tricked if we receive a handshake record with zero length?

i-rinat: > Won't we be tricked if we receive a handshake record with zero length?

As far as I understand, no.

`ttls_parse_record_hdr()` returns an error if `msglen` is less than `TTLS_HS_HDR_LEN` (which is 4). That's enough to read a handshake length. And then there is a check for `hslen` which returns an error if we have too short ClientHello message, that is less than 38 bytes. We also have a strict limit of at least 1 byte length in `ttls_hdr_check()`.

i-rinat:
As a result of an off-site discussion, I also removed `msglen == 0` special case handling in https://github.com/tempesta-tech/tempesta/pull/1353/commits/f5836a101e6891006850aae58b0cc1dd98b7f7f5.
Rationale: (a) That workaround is not used in OpenSSL since about 2002. (b) Doesn't apply to TLS 1.2, only TLS 1.0 and SSL 3.0, which we do not support. (c) The workaround [was talking](https://www.openssl.org/~bodo/tls-cbc.txt) about empty plaintext, which can not translate to empty ciphertext in our implementation, as we require encryption.

PR #1358, opened 2019-10-09
closer to rfc parsing date

### LINKED ISSUE #1202: Inaccurate date parsing
Originally from #1050 Date case - a request with incorrect date field with normally processed by Tempesta:
```
# curl -v -H 'If-modified-since: invalid' http://192.168.100.4:80/
*   Trying 192.168.100.4...
* TCP_NODELAY set
* Connected to 192.168.100.4 (192.168.100.4) port 80 (#0)
> GET / HTTP/1.1
> Host: 192.168.100.4
> User-Agent: curl/7.61.1
> Accept: */*
> If-modified-since: invalid
> 
< HTTP/1.1 200 OK
< Date: Mon, 04 Mar 2019 21:01:01 GMT
< Last-Modified: Sat, 26 Jan 2019 12:04:30 GMT
< ETag: "3-5805b3efa7793"
< Accept-Ranges: bytes
< Content-Length: 3
< Content-Type: text/html
< Server: Tempesta FW/0.7.0
< Via: 1.1 tempesta_fw (Tempesta FW 0.7.0)
< 
He
* Connection #0 to host 192.168.100.4 left intact
```
Debug output shows
```
[20927.916386] [tempesta fw]       parser at Req_HdrIf_Modified_SinceV: c=0x69(i), p_off=94
[20927.918140] [tempesta fw]       store field chunk len=94 data=ffff96db6ed4e8c2(G) field=<0x0,0,ffff96db6ed4e90d>
[20927.921975] [tempesta fw]       enter FSM at state 0000000000000000
[20927.922190] [tempesta fw]       parser at I_Date: c=0x69(i), p_off=0
[20927.926175] [tempesta fw]       enter FSM at state 0000000000000001
[20927.926760] [tempesta fw]       parser at I_EoL: c=0x69(i), p_off=0
[20927.929892] [tempesta fw]       tfw_match_nctl: str[0]=0x69 len=11 r=7 cust=0
[20927.930188] [tempesta fw]       __parse_http_date: parsed date 0
[20927.930188] [tempesta fw]       parse raw header __req_parse_if_msince: ret=7 data_len=11
```
i.e. `invalid` string is just ignored as a week day in `__skip_weekday()` loop. The bad thing is that `__req_parse_if_msince()` sets seems zero `req->cond.m_date` and `TFW_HTTP_COND_IF_MSINCE` flag, so Tempesta FW can be tricked.

Following must be done:
1. match exactly 3 letters in `__skip_weekday()`, so we can unroll the loop and that's good. Probably it's fine to pass `inv` and we don't need to spend CPU cycles on verifying skipped info;
2. `__parse_http_date()` in all the cases must return success value only if the date is correct.

Also please update the HTTP parser unit test by the Date cases and select appropriate checkboxes in #1059 corresponindinf to date headers.

No need to port the fix to 0.6.

avbelov23:
Fix #1202

- We must set the HTTP flags, if only the date is successfully parse.
- Parsing date according to rfc
- Fix `__year_day_secs()`.
- If-Modified-Since header parsing tests.
- Fix date format in tests.

### DIFF
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -2764,7 +2764,7 @@ __req_parse_if_msince(TfwHttpMsg *msg, unsigned char *data, size_t len)
 		r = __parse_http_date(msg, data, len);
 	}
 
-	if (r >= 0) {
+	if (r >= 0 && parser->_date != 0) {
 		req->cond.m_date = parser->_date;
 		req->cond.flags |= TFW_HTTP_COND_IF_MSINCE;
 	}
@@ -4459,7 +4459,7 @@ __resp_parse_expires(TfwHttpMsg *msg, unsigned char *data, size_t len)
 		r = __parse_http_date(msg, data, len);
 	}
 
-	if (r >= 0) {
+	if (r >= 0 && parser->_date != 0) {
 		resp->cache_ctl.expires = parser->_date;
 		resp->cache_ctl.flags |= TFW_HTTP_CC_HDR_EXPIRES;
 	}
@@ -4488,7 +4488,7 @@ __resp_parse_date(TfwHttpMsg *msg, unsigned char *data, size_t len)
 		r = __parse_http_date(msg, data, len);
 	}
 
-	if (r >= 0) {
+	if (r >= 0 && parser->_date != 0) {
 		resp->date = parser->_date;
 		__set_bit(TFW_HTTP_B_HDR_DATE, resp->flags);
 	}
@@ -4517,7 +4517,7 @@ __resp_parse_if_modified(TfwHttpMsg *msg, unsigned char *data, size_t len)
 		r = __parse_http_date(msg, data, len);
 	}
 
-	if (r >= 0) {
+	if (r >= 0 && parser->_date != 0) {
 		resp->last_modified = parser->_date;
 		__set_bit(TFW_HTTP_B_HDR_LMODIFIED, resp->flags);
 	}
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -232,17 +232,23 @@ do {									\
 #define __FSM_I_chunk_flags(flag)					\
 	__FSM_I_field_chunk_flags(&msg->stream->parser.hdr, flag)
 
-#define __FSM_I_MOVE_n(to, n)						\
+#define __FSM_I_MOVE_BY_REF_n(to, n)					\
 do {									\
-	parser->_i_st = &&to;						\
+	parser->_i_st = to;						\
 	p += n;								\
 	if (unlikely(__data_off(p) >= len)) {				\
 		/* Close currently parsed field chunk. */		\
 		__msg_hdr_chunk_fixup(data, len);			\
 		__FSM_EXIT(TFW_POSTPONE);				\
 	}								\
-	goto to;							\
+	goto *to;							\
 } while (0)
+
+#define __FSM_I_MOVE_n(to, n)						\
+	__FSM_I_MOVE_BY_REF_n(&&to, n)
+
+#define __FSM_I_MOVE_BY_REF(to)						\
+	__FSM_I_MOVE_BY_REF_n(to, 1)
 
 #define __FSM_I_MOVE(to)		__FSM_I_MOVE_n(to, 1)
 /* The same as __FSM_I_MOVE_n(), but exactly for jumps w/o data moving. */
@@ -725,7 +731,7 @@ mark_trailer_hdr(TfwHttpMsg *hm, TfwStr *hdr)
  * Parsing helpers.
  * @str in TRY_STR_LAMBDA must be in lower case.
  */
-#define TRY_STR_LAMBDA_finish(str, lambda, finish, state)		\
+#define TRY_STR_LAMBDA_BY_REF_finish(str, lambda, finish, state)	\
 	if (!chunk->data)						\
 		chunk->data = p;					\
 	__fsm_n = __try_str(&parser->hdr, chunk, p, __data_remain(p),	\
@@ -734,12 +740,15 @@ mark_trailer_hdr(TfwHttpMsg *hm, TfwStr *hdr)
 		if (chunk->len == sizeof(str) - 1) {			\
 			lambda;						\
 			TRY_STR_INIT();					\
-			__FSM_I_MOVE_n(state, __fsm_n);			\
+			__FSM_I_MOVE_BY_REF_n(state, __fsm_n);		\
 		}							\
 		__msg_hdr_chunk_fixup(data, len);			\
 		finish;							\
 		return CSTR_POSTPONE;					\
 	}
+
+#define TRY_STR_LAMBDA_finish(str, lambda, finish, state)		\
+	TRY_STR_LAMBDA_BY_REF_finish(str, lambda, finish, &&state)
 
 /*
  * Store current state if we're going to exit in waiting for new data
@@ -754,6 +763,11 @@ mark_trailer_hdr(TfwHttpMsg *hm, TfwStr *hdr)
 #define TRY_STR(str, curr_st, next_st)					\
 	TRY_STR_LAMBDA_finish(str, { }, {				\
 			parser->_i_st = &&curr_st;			\
+		}, next_st)
+
+#define TRY_STR_BY_REF(str, curr_st, next_st)				\
+	TRY_STR_LAMBDA_BY_REF_finish(str, { }, {			\
+			parser->_i_st = curr_st;			\
 		}, next_st)
 
 /**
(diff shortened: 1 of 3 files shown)


tempesta_fw/http_parser.c
```diff
@@ -2625,59 +2643,137 @@ __parse_http_date(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
 		if (likely(c == ' '))
-			__FSM_I_MOVE(I_DateHour);
+			__FSM_I_MOVE(I_DateHourBeg);
 		return CSTR_NEQ;
 	}
 
+	__FSM_STATE(I_DateHourBeg) {
+		size_t n;
+		__fsm_sz = __data_remain(p);
+		n = __fsm_sz < 3 ? __fsm_sz: 3;
```
krizhanovsky: That's just `n = __fsm_sz & 3`. Please avoid conditions in the parser - it's too hot.

tempesta_fw/http_parser.c
```diff
@@ -2625,59 +2643,137 @@ __parse_http_date(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+		parser->_date += parser->_acc * 3600;
+		parser->_acc = 0;
+		__FSM_I_MOVE_n(I_DateMinCln, __fsm_n);
+	}
+
 	__FSM_STATE(I_DateHour) {
+		size_t n;
 		__fsm_sz = __data_remain(p);
-		__fsm_n = parse_int_a(p, __fsm_sz, colon_a, &parser->_acc);
+		n = __fsm_sz < 3 ? __fsm_sz: 3;
```
krizhanovsky: same as above

tempesta_fw/http_parser.c
i-rinat: Ehm... no. For example, if `__fsm_sz` is 4, the statement `n = __fsm_sz & 3` will result in `0` rather than expected `3`.

krizhanovsky: Yeah, my bad

tempesta_fw/http_parser.c
```diff
@@ -2518,12 +2518,15 @@ __parse_http_date(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+		size_t n;
 		__fsm_sz = __data_remain(p);
 		/* Parse a 2-digit day. */
-		__fsm_n = parse_int_ws(p, __fsm_sz, &parser->_acc);
+		n = __fsm_sz < 3 ? __fsm_sz: 3;
+		__fsm_n = parse_int_ws(p, n, &parser->_acc);
 		if (__fsm_n == CSTR_POSTPONE) {
-			parser->_i_st = &&I_DateDay;
-			__msg_hdr_chunk_fixup(data, len);
+			if (n == 3)
```
i-rinat: If we are parsing one byte at a time, `n` will always be `1`, so this check won't catch long numbers. So it will sometimes catch longer numbers, sometimes not. It's not like there will be a security breach, but something like `0016` would be possible if chunk border is between `00` and `16`.

I think it will be easier to make two explicit states, for the first and for second digits. Resulting code may end up even smaller. Day of month may be represented with a single digit, but that should not be an issue.

tempesta_fw/http_parser.c
```diff
@@ -2468,61 +2468,69 @@ __year_day_secs(unsigned int year, unsigned int day_sec)
...
+
+	__FSM_STATE(I_WDateSP) {
+		if (likely(c == ' '))
+			__FSM_I_MOVE(I_DateDay1);
+		return CSTR_NEQ;
+	}
+
+	__FSM_STATE(I_DateDay1) {
+		if (unlikely(!isdigit(c)))
+			return CSTR_BADLEN;
```
ikoveshnikov: It's not clear why some return codes are `CSTR_BADLEN` while others are `CSTR_NEQ`. I see no special processing of the `CSTR_BADLEN` case for date values. In most other cases we return `CSTR_BADLEN` when the parsed value is longer than underlying data structures (e.g. integers). This return values are not required and we can leave only one - `CSTR_NEQ`

tempesta_fw/http_parser.c
```diff
@@ -2468,61 +2468,69 @@ __year_day_secs(unsigned int year, unsigned int day_sec)
...
-		__fsm_n = __skip_weekday(p, __fsm_sz);
-		if (__fsm_sz == __fsm_n)
-			__FSM_I_MOVE_n(I_Date, __fsm_sz);
-		if (unlikely(__fsm_n == CSTR_NEQ))
+	/*
+	 * Skip a weekday with comma (e.g. "Sun,") as redundant
+	 * information.
+	 */
+	__FSM_STATE(I_WDate1) {
+		unsigned char lc = TFW_LC(c);
```
ikoveshnikov: The [RFC 7231 section  Date/Time Formats](https://tools.ietf.org/html/rfc7231#section-7.1.1.1) talks about three formats, but only `IMF-fixdate` is preferred:
```
   An example of the preferred format is

     Sun, 06 Nov 1994 08:49:37 GMT    ; IMF-fixdate

   Examples of the two obsolete formats are

     Sunday, 06-Nov-94 08:49:37 GMT   ; obsolete RFC 850 format
     Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
```
In the same time `A recipient that parses a timestamp value in an HTTP header field MUST accept all three HTTP-date formats.` The proposed solution parses only ` IMF-fixdate`, but we have to support all of them, so extra steps are required.

tempesta_fw/http_parser.c
```diff
@@ -2468,61 +2468,69 @@ __year_day_secs(unsigned int year, unsigned int day_sec)
...
-		/* Parse a 2-digit day. */
-		__fsm_n = parse_int_ws(p, __fsm_sz, &parser->_acc);
-		if (__fsm_n == CSTR_POSTPONE) {
-			parser->_i_st = &&I_DateDay;
-			__msg_hdr_chunk_fixup(data, len);
-		}
-		if (__fsm_n < 0)
-			return __fsm_n;
+	__FSM_STATE(I_WDate2) {
+		unsigned char lc = TFW_LC(c);
```
ikoveshnikov: All the three supported formats are _case-sensitive_, so We don't need `TFW_LC` calls at all. The first letter of the day name is upper case, the rest are lowercase.

ikoveshnikov:
It seems like all parts of date except day name are fixed sized, so we can make two-step parsing routine that will perform better but will require a bit of copy operations:

Changes:
1. Add to TfwHttpParser a char array that can handle the longest valid Date characters (except week day name), let it be `char date_str[]`.
2. Add to parser a bit array (or reuse any of the currently unused members) to store a date format indicator, let it be `int date_format` in this example, with possible values `0x1` for `IMF-fixdate`, `0x2` for `obsolete RFC 850` and `0x4` for `ANSI C's asctime()`. 
3. Until the date is fully parsed, the `parser->_acc` will show the current empty position in `date_str`

Algorithm:
Preparations:
1. Statically define three couples of char arrays: `min[]` and `max[]` for all the date formats. Since all the formats are fixed sized except day name and case-sensitive, no extra `SP` are allowed,  for each position of date value we can define minimum and maximum character possible. This will be our compare mark.
2. set `parser->_acc = 0`, Clear the `parser->date_str` (optionally, just to get rid of possible errors). 

Validate the date value:
1. Eat the weekday name. First letter is upper case, the rest is lowercase,
2. `,` after weekday name ->  Set `parser->date_format` to `(IMF-fixdate) | (obsolete RFC 850)`, or `SP` after weekdayname ->  Set `parser->date_format` to `ANSI C's asctime`
3. The rest of the date value is the fixed-sized value, so compare every new character `c` to be `min[parser->_acc] <= c <= max[parser->_acc]`, if the condition is true, then copy the `c` to `parser->date_str[parser->_acc]`, increment the `parser->_acc` and repeat step `3`.  Since we may have three possible date formats, we will need three  pairs of `min[]` and `max[]`.
4. If the condition from the point above failed, then drop the format id from the `parser->date_format`, if no left, then return the `CSTR_NEQ`
5. Formats are very different, and only one can be parsed successfully, if one of the formats if fully parsed, then the validation step is over, eat the rest of the line.

Parse validated value into `time_t`:
1. Now we know the exact format and hate the full copy of the date value inside the `parser->date_str[]`. If the `parser->date_format == 0`, then the date format is invalid and we can't parse the value.
2. We can directly access day number, month name, year etc. and directly get numerical values, we even can use binary search to parse month name.

No `TRY_STR()`, no complicated state machines to parse all the three possible date formats, but we have to copy ~25 bytes and can't reject messages with invalid date value, until the full date string is delivered.

Any thoughts?

tempesta_fw/http_parser.c
i-rinat: Addressed.

krizhanovsky:
In discussion for https://github.com/tempesta-tech/tempesta/pull/1358#issuecomment-540567439 :

For current implementation (before introducing additional states to handle obsolete dates) we have 29 states and we saw (https://www.slideshare.net/AlexanderKrizhanovsky1/fast-http-string-processing-algorithms , page 22) that plain switch driven FSM behaves poorly.

Basically, I'm fine with the algorithm proposed by @ikoveshnikov , however couple notes :

1. we do need to shring current FSM - 29 states is too much. I'd say that the FSM should not be more that 20 states.

2. the other option is to introduce goto driven DSL for nested parsers as we have for the main parsers. The benefit is that we can win in different places, however, this particular case won't be faster having greater number of states even goto-driven.

All in all the propsed algorithms looks good if it can shrink FSM to below 20 states or lower.

i-rinat:
There is still discrepancies regarding leap years: there are sometimes gaps or overlapping in epoch time that comes out of the parser.

As there is no way such jumps will occur during the days, assuming we don't touch daylight saving time at all. We multiply hours and minutes by constants and add with seconds. There is no way there will be gaps or overlaps. But jumps may occur when date changes. So here is a sample code that iterates through a period and prints dates at which jumps occur. In the ideal case, the should be only one such apparent jump, at 1 Jan 1970, since we start at 0.

```patch
diff --git a/tempesta_fw/http_parser.c b/tempesta_fw/http_parser.c
index 69527805..7e0371fc 100644
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -5093,3 +5093,64 @@ tfw_http_parse_is_done(TfwHttpMsg *hm)
 {
 	return test_bit(TFW_HTTP_B_FULLY_PARSED, hm->flags);
 }
+
+void
+hp_startup(void)
+{
+    static const int days_in_month[12] =
+        { 31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
+    static const char *mname[12] =
+        {"Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct",
+         "Nov", "Dec"};
+    int year;
+    int month;
+    int day;
+    int r;
+    static char buf[200];
+    time_t prev_day_premidnight = 0;
+
+    printk("hp_startup begin\n");
+    for (year = 1970; year < 2020; year++) {
+        int leap = 0;
+        if (year % 4 == 0)
+            leap = 1;
+        if (year % 100 == 0 && year % 400 != 0)
+            leap = 0;
+
+        for (month = 1; month <= 12; month++) {
+            int days_count = days_in_month[month - 1];
+            if (month == 2 && leap)
+                days_count = 29;
+            for (day = 1; day <= days_count; day++) {
+                {
+                    TfwStream stream = {};
+                    TfwHttpMsg hm = {.stream = &stream};
+
+                    sprintf(buf, "aaa, %02d %s %04d 00:00:00 GMT\r\n",
+                            day, mname[month-1], year);
+                    r = __parse_http_date(&hm, (unsigned char *)buf, strlen(buf));
+                    if (r <= 0)
+                        printk("error1, r=%d\n", r);
+
+                    if (hm.stream->parser._date - prev_day_premidnight != 1)
+                        printk("%02d %s %04d 00:00:00, delta=%ld\n", day,
+                               mname[month-1], year,
+                               hm.stream->parser._date - prev_day_premidnight);
+                }
+                {
+                    TfwStream stream = {};
+                    TfwHttpMsg hm = {.stream = &stream};
+
+                    sprintf(buf, "aaa, %02d %s %04d 23:59:59 GMT\r\n",
+                            day, mname[month-1], year);
+
+                    r = __parse_http_date(&hm, (unsigned char *)buf, strlen(buf));
+                    if (r <= 0)
+                        printk("error2, r=%d\n", r);
+                    prev_day_premidnight = hm.stream->parser._date;
+                }
+            }
+        }
+    }
+    printk("hp_startup end\n");
+}
diff --git a/tempesta_fw/main.c b/tempesta_fw/main.c
index b7d8b2e4..fbfbc334 100644
--- a/tempesta_fw/main.c
+++ b/tempesta_fw/main.c
@@ -417,6 +417,9 @@ tfw_exit(void)
 	unregister_net_sysctl_table(tfw_sysctl_hdr);
 }
 
+void
+hp_startup(void);
+
 static int __init
 tfw_init(void)
 {
@@ -463,6 +466,7 @@ tfw_init(void)
 	DO_INIT(sched_hash);
 	DO_INIT(sched_ratio);
 
+	hp_startup();
 	return 0;
 err:
 	tfw_exit();
```

krizhanovsky:
> We are already using computed goto jumps in the __parse_http_date(), so there is no switch on entry. Does it matter how many states there are then?

Oh, forgot about this. Yes, it's less important then. However, the FSM is still sensitive to the number of states.

tempesta_fw/http_parser.c
```diff
@@ -2428,274 +2428,237 @@ __req_parse_referer(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+		TfwHttpDateFmt *fmt = parser->date_fmt;
+
+		if (parser->_acc >= fmt->len) {
+			time_t date = __parse_date(parser->date_str, fmt);
+			if (date < 0)
+				return CSTR_NEQ;
+			parser->_date = date;
+			__FSM_I_JMP(I_EoL);
+		}
+		if (c < fmt->min[parser->_acc] || c > fmt->max[parser->_acc]) {
```
i-rinat: This is a possible point for further optimizations. However, it's not immediately clear if vectorization will gain anything here.

tempesta_fw/http_parser.c
```diff
@@ -2428,274 +2428,237 @@ __req_parse_referer(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+	days += year * 365 + year / 4 - year / 100 + year / 400;
+	/* 31 and 28 days were in January and February 1970 */
+	return (days - EPOCH_DAYS + 31 + 28) * SEC24H +
+	       hour * 3600 + min * 60 + sec;
 }
 
-static size_t
-__skip_weekday(unsigned char *p, size_t len)
+static time_t
+__parse_date(char *date_str, TfwHttpDateFmt *fmt)
```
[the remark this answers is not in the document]
krizhanovsky: Yeah, agree - usually it's more beneficial to split a hot path (the most used data format) from the others (2 others) - it's more straightforward to implement and works faster.

tempesta_fw/http_parser.c
```diff
@@ -2428,274 +2428,237 @@ __req_parse_referer(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+	       (date_str[fmt->time_pos + 1] - '0');
+	min  = (date_str[fmt->time_pos + 3] - '0') * 10 +
+	       (date_str[fmt->time_pos + 4] - '0');
+	sec  = (date_str[fmt->time_pos + 6] - '0') * 10 +
+	       (date_str[fmt->time_pos + 7] - '0');
+
+	month_int = TFW_CHAR4_INT(' ',
+	                          date_str[fmt->month_pos + 0],
+	                          date_str[fmt->month_pos + 1],
+	                          date_str[fmt->month_pos + 2]);
```
krizhanovsky: I thinks compiler is smart enough to place `date_str[fmt->time_pos]` in a register, but I don't see any reason for the such massive code copy & paste - why not to use something like (please chose more appropriate names):
```C
unsigned char *p = date_str + fmt->year_pos;
year = (time_t)p[0] * 1000 + (p[1] - '0') * 100 + (p[2] - '0') * 10 + (p[2] - '0');
```
-- much less typing, isn't it?

tempesta_fw/http_parser.h
```diff
@@ -98,6 +131,8 @@ typedef struct {
 	TfwStr		_tmp_chunk;
 	TfwStr		hdr;
 	TfwHttpHbhHdrs	hbh_parser;
+	TfwHttpDateFmt	*date_fmt;
+	char		date_str[MAX_DATE_LEN];
```
krizhanovsky: There is not reason to keep the temporal buffer - we don't appreciate Nginx-like copies and multiple data traversals.

tempesta_fw/http_parser.c
```diff
@@ -2493,18 +2480,41 @@ __parse_http_date(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
-		if (__fsm_sz == __fsm_n)
-			__FSM_I_MOVE_n(I_Date, __fsm_sz);
-		if (unlikely(__fsm_n == CSTR_NEQ))
+	/*
+	 * Skip a weekday with comma (e.g. "Sun,") as redundant
+	 * information.
+	 */
+	__FSM_STATE(I_WDate1) {
+		unsigned char lc = TFW_LC(c);
+		if ((unsigned)(lc - 'a') > (unsigned)('z' - 'a'))
```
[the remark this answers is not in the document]
krizhanovsky: Ok, thank you @i-rinat . Then forget my comment about Hacker's delight trick

tempesta_fw/http_parser.c
i-rinat: Yeah, I was talking earlier about compilers already doing the optimization, and about why one shouldn't do that by hand, since it makes code harder to understand at a first glance. But in this particular case compiler skips the optimize. (We use GCC 6.3, but it's the same for more recent versions too). Even simple
```c
void func_simple2(char c, char min, char max) {
    if (min < c && c < max)
        puts("check");
}
```
are not optimized.
Turned out, the reason is in unpredictable `min` and `max` values. I assumed that `min <= max` just from the names. It's a reasonable assumption for a human, but a compiler cannot afford that. So since in the case of `min > max` optimization breaks the control flow, it's not used.

So, rule of thumb: if borders are immediate constants, two comparisons will likely be optimized to one. Otherwise, a manual optimization is required.

krizhanovsky: Thanks @i-rinat . This means that we do need the hacker's delight trick here, but don't need it in comparisons like `'a' <= c && c <= 'z'`. @avbelov23 please make the optimization and leave a comment about compiler optimizations and why we do need manual optimization here.

tempesta_fw/http_parser.c
```diff
@@ -2428,274 +2428,237 @@ __req_parse_referer(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+		if (c < fmt->min[parser->_acc] || c > fmt->max[parser->_acc]) {
+			if (likely(fmt->type != ISOC ||
+			           parser->_acc != fmt->day_pos ||
+			           c != ' '))
+			{
+				return CSTR_NEQ;
+			}
+		}
+		parser->date_str[parser->_acc++] = c;
+		__FSM_I_MOVE(I_Loop);
```
krizhanovsky: I don't like what happens in the state at all.

Please use bits [56-61] of `_acc` (bits [62-62] are used to index array of formats) for current parser position on the data string. If it reaches `fmt->len`, then return `CSTR_NEQ`. In `TfwHttpDateFmt` you have positions of the year, month, day and time, so by reaching each of the position we should move to appropriate states like `I_DateYear`, `I_DateMonth` etc. Your format templates also have separators, so we can use them in only one state for all the separators. States like `I_DateYear` just use `_acc` as previously to accumulate current integer values and do appropriate part of the logic from `__parse_date()`.

For me it looks like we can keep very similar code as the original `__parse_http_date()` with the extension by the format templates.

tempesta_fw/http_parser.c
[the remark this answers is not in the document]
krizhanovsky: I though about this little bit more and it turns out that we'll still have to keep all the states, for all the 3 formats, in the same FSM. We can use `hot` states for the work flow of the most probable format, but total number of states in the function will be large and it's likely that GCC won't make a good job on optimization of it.

@i-rinat do you have an idea how to split the states or reduce number of states? My concern is that we can go to the same version of the code as the original one and have the same problem https://github.com/tempesta-tech/tempesta/pull/1358#issuecomment-540815974

tempesta_fw/http_parser.h
```diff
@@ -85,19 +85,34 @@ typedef struct {
  * @hbh_parser	- list of special and raw headers names to be treated as
  *		  hop-by-hop
  * @_date	- currently parsed http date value;
+ * @st		- template showing in what state the i-th character in the
```
ikoveshnikov: The structure documentation is outdated and doesn't reflect exact members. E.g. there is no `st` member.

tempesta_fw/http_parser.c
```diff
@@ -2430,274 +2444,310 @@ __req_parse_referer(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+			return CSTR_NEQ;
+
+		if (parser->date.year < 100)
+			parser->date.year += (parser->date.year < 70) ? 2000
+			                                              : 1900;
+
+		date = __date_secs(parser->date.year, month,
+		                   parser->date.day, parser->date.hour,
+		                   parser->date.min, parser->date.sec);
+		if (date < 0)
```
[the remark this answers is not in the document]
i-rinat: Won't this make `Thu,  1 Jan 1970 00:00:00 GMT` invalid? Some client may use that date in `If-Modified-Since`, and that looks like a correct date to me.

PR #1361, opened 2019-10-15
tls: move io context resetting to tfw_tls_msg_process()

i-rinat:
Investigation of #1318 revealed an inherent issue with the previous design of the IO context resets. `ttls_recv()` reset the whole IO context, which in turn dropped skb list collected in a IO context so far, and then continued to parse the supplied data. This was an optimization that allowed `ttls_recv()` to greedily parse multiple records at a time. But if during TCP fragmentation a single skb ended up containing both an alert (or a handshake-related record) and a part of application data record, the code could forget a first part of the record. Then, when remaining parts of an application data record arrived, scatterlist was filled incorrectly, and reading uninitialized data could occur. As part of  essential enciphered data were discarded without processing, no further data exchange were possible.

The proposed solution is to move context resets out of `ttls_recv()` to `tfw_tls_msg_process()`. That makes `ttls_recv()` process a single record at a time, but also allows `tfw_tls_msg_process()` to reinitialize `tls->io_in->skb_list` before each record, thus eliminating the issue.

### DIFF
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -28,7 +28,6 @@
 #include <linux/module.h>
 #include <net/tls.h>
 
-#include "lib/str.h"
 #include "config.h"
 #include "debug.h"
 #include "crypto.h"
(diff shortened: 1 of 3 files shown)


tls/ttls.c
```diff
@@ -2310,33 +2308,12 @@ ttls_recv(void *tls_data, unsigned char *buf, size_t len, unsigned int *read)
...
 	}
 
-	/*
-	 * At this point we have fully prepared data for the upper layer.
-	 * Once we return the data is passed for processing to the upper layer,
-	 * so we reinitialize I/O context for a next message.
-	 */
-	bzero_fast(io->__initoff, sizeof(*io) - offsetof(TlsIOCtx, __initoff));
-
+done:
```
krizhanovsky: It seems we don't need the label any more and can just `return T_OK` instead of `goto done`.

Next, the comments for `T_POSTPONE` and `T_OK` in `tfw_tls_msg_process()` become misleading, since now after the `switch (r) { ... }` in the function we may have an alert of handshake message which should not be passed to upper layer. And this is why you check for `tls->io_in.msgtype == TTLS_MSG_APPLICATION_DATA` - having the old work flow in mind, I couldn't figure this out quickly.

tls/ttls.c
```diff
@@ -2239,7 +2237,7 @@ ttls_recv(void *tls_data, unsigned char *buf, size_t len, unsigned int *read)
 	case TTLS_MSG_ALERT:
```
krizhanovsky: I believe the block
```C
        delta = *read - parsed;                                                        
        len -= delta;                                                                  
        buf += delta;
        parsed = *read; 
```
should be under the `if (!(io->st_flags & TTLS_F_ST_HDRIV))` statement and before `if (io->msgtype == TTLS_MSG_HANDSHAKE...)`, so we can use `delta` instead of two `*read - parsed` expressions in another `if` - trivial change, but it makes the code simpler.

tls/ttls.c
i-rinat: > It seems we don't need the label any more and can just `return T_OK` instead of `goto done`.

Replaced `goto done` with `return T_OK`.

> the comments <...> become misleading

Updated the comments so that they now explain the current code.

tls/ttls.c
i-rinat: Indeed. Moved the block into the if statement body.

i-rinat:
Changed the patch to address review comments so far.

PR #1362, opened 2019-10-16
fix #1221: chunked encoding may be not-final encoding for response

### LINKED ISSUE #1221: Inaccurate Transfer-Encoding parsing
From [RFC](https://tools.ietf.org/html/rfc7230#section-3.3.1):
>   If any transfer coding
   other than chunked is applied to a _request_ payload body, the sender
   MUST apply chunked as the final transfer coding to ensure that the
   message is properly framed.  If any transfer coding other than
   chunked is applied to a _response_ payload body, the sender MUST either
   apply chunked as the final transfer coding or terminate the message
   by closing the connection.

But Tempesta's parsing is inaccurate. It uses the same rules for both response and requests, thus `chunked` encoding is not allowed at any place except the last one in the response. This is not the correct behaviour.

ikoveshnikov:
RFC allows to add more encodings to chunked response, but correct
framing is possible only by connection close.

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -3873,16 +3873,20 @@ static void
 tfw_http_resp_terminate(TfwHttpMsg *hm)
 {
 	TfwFsmData data;
-	int r;
+	int r = 0;
 
 	/*
-	 * Add absent message framing information. If there is no body, add
-	 *  Content-length header: it requires less modifications.
+	 * Add absent message framing information. It's possible to add a
+	 * 'Content-Length: 0' header, if the Transfer-Encoding header is not
+	 * set, but keep more generic solution and transform to chunked.
+	 * It's the only possible modification for future proxy mode.
+	 * If the framing information can't be added, then close client
+	 * connection after response is forwarded.
 	 */
-	r = (hm->body.len)
-		? tfw_http_msg_to_chunked(hm)
-		: TFW_HTTP_MSG_HDR_XFRM(hm, "Content-Length", "0",
-					TFW_HTTP_HDR_CONTENT_LENGTH, 0);
+	if (!test_bit(TFW_HTTP_B_CHUNKED_APPLIED, hm->flags))
+		r = tfw_http_msg_to_chunked(hm);
+	else
+		set_bit(TFW_HTTP_B_CONN_CLOSE, hm->req->flags);
 
 	if (r) {
 		TfwHttpReq *req = hm->req;
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -1040,6 +1040,8 @@ __FSM_STATE(RGen_BodyInit) {						\
 			TFW_PARSER_BLOCK(RGen_BodyInit);		\
 		if (test_bit(TFW_HTTP_B_CHUNKED, msg->flags))		\
 			__FSM_MOVE_nofixup(RGen_BodyStart);		\
+		/* Process the body until the connection is closed. */	\
+		__set_bit(TFW_HTTP_B_UNLIMITED, msg->flags);		\
 		__FSM_MOVE_nofixup(Resp_BodyUnlimStart);		\
 	}								\
 	if (!TFW_STR_EMPTY(&tbl[TFW_HTTP_HDR_CONTENT_LENGTH])) {	\
@@ -1719,7 +1721,8 @@ STACK_FRAME_NON_STANDARD(__req_parse_content_type);
  * Parse Transfer-Encoding header value, RFC 2616 14.41 and 3.6.
  */
 static int
-__parse_transfer_encoding(TfwHttpMsg *hm, unsigned char *data, size_t len)
+__parse_transfer_encoding(TfwHttpMsg *hm, unsigned char *data, size_t len,
+			  bool client)
 {
 	int r = CSTR_NEQ;
 	__FSM_DECLARE_VARS(hm);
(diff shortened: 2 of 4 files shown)


tempesta_fw/http_parser.c
```diff
@@ -1764,19 +1765,39 @@ __parse_transfer_encoding(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+				if (client)
+					return CSTR_NEQ;
+				/*
+				 * Prepare to receive 'unlimited' body, close
+				 * client connection after response is forwarded
+				 * to preserve correct message framing on client
+				 * side.
+				 */
+				__clear_bit(TFW_HTTP_B_CHUNKED, msg->flags);
+				set_bit(TFW_HTTP_B_CONN_CLOSE, msg->pair->flags);
```
i-rinat: We get to this line only for responses, and only where `Transfer-Encoding` had a "chunked" somewhere in the field value. So, as far as I understand, if backend sends us `Transfer-Encoding: gzip`, we won't be ready for the "unlimited body mode", since there were no "chunked", and therefore TFW_HTTP_B_CHUNKED bit wasn't set. Right?

ikoveshnikov: The "unlimited body mode" is turned on in two cases: https://github.com/tempesta-tech/tempesta/blob/1d9a5758b966ed93a78ae4264a592f67023e68a2/tempesta_fw/http_parser.c#L1022

1. Transfer-Encoding header is not empty and the `CHUNKED` flag is missing
2. No Transfer-Encoding and no Content-Length headers are provided. 

So if backend sends us Transfer-Encoding: gzip, we will be ready for the "unlimited body mode".

But there is an issue in the PR: `TFW_HTTP_B_UNLIMITED` flag is not set, so the response is actually blocked and dropped after receive. I will need to move `TFW_HTTP_B_UNLIMITED` flag to the `Resp_BodyUnlimStart` parser state.

ikoveshnikov:
The whole PR was reworked, tests was added in tempesta-tech/tempesta-test#145

Every thing is clear with requests: request is blocked if `chunked` is not last encoding.

It's allowed to use chunked encoding as not the final encoding for response. When a "unlimited" body is received from backend server TempestaFW needs some how indicate the message end for the client. In most cases it's possible to apply the `chunked` encoding as the final encoding and pass the response to a client. But if the `chunked` encoding was already applied to the payload, RFC doesn't allow to apply it second time, even if the `chunked` wasn't the final encoding. In this case we have to close the connection with a client to indicate the message end.

The following message flags are used:
- `TFW_HTTP_B_CHUNKED` - `chunked` is the final encoding, message framing is valid, no need to modify response.  
- `TFW_HTTP_B_CHUNKED_APPLIED` - `chunked` encoding is not the final encoding, so we can't modify the message to indicate the message end, and have to close the client connection. To accomplish that the `TFW_HTTP_B_CONN_CLOSE` flag is set for corresponding request in `tfw_http_resp_terminate()`.  
- `TFW_HTTP_B_UNLIMITED` - the response body is to be parsed in `unlimited` mode - fin up the response on connection error event if the flag set, otherwise drop the response as incompletely received.

tempesta_fw/http_parser.c
ikoveshnikov: Thanks to pointing to that. I have reworked the PR, the comment below provides the explanations about flags and their usage.

tempesta_fw/http.c
```diff
@@ -3873,16 +3873,20 @@ static void
...
+	 * set, but keep more generic solution and transform to chunked.
+	 * It's the only possible modification for future proxy mode.
+	 * If the framing information can't be added, then close client
+	 * connection after response is forwarded.
 	 */
-	r = (hm->body.len)
-		? tfw_http_msg_to_chunked(hm)
-		: TFW_HTTP_MSG_HDR_XFRM(hm, "Content-Length", "0",
-					TFW_HTTP_HDR_CONTENT_LENGTH, 0);
+	if (!test_bit(TFW_HTTP_B_CHUNKED_APPLIED, hm->flags))
```
i-rinat: `TFW_HTTP_B_CHUNKED_APPLIED` is set only if there were `chunked` somewhere in `Transfer-Encoding`, but it wasn't the last one. So it won't be set for `Transfer-Encoding: deflate`, for example. Shouldn't we switch `TFW_HTTP_B_CONN_CLOSE` on in this case too?

tempesta_fw/http.c
```diff
@@ -3873,16 +3873,20 @@ static void
...
+		r = tfw_http_msg_to_chunked(hm);
```
i-rinat: I believe there may be an issue with `tfw_http_msg_to_chunked()` with zero-length body responses. It prepends `<length>\r\n` from the start if the length is non-zero and always appends `\r\n0\r\n\r\n` at the end. If body length is `0`, a body will be replaced with `\r\n0\r\n\r\n` instead of `0\r\n\r\n`. We had no issues with that before since the code called `tfw_http_msg_to_chunked()` only if `hm->body.len > 0`.

tempesta_fw/http.c
ikoveshnikov: No, we needn't. We are trying our best to keep connections open if possible, it actually save our resources:
- Less handshakes TCP or more costly TLS ones,
- There can be more pipelined requests from that client, that can be sent to other connections, and we may store that request and corresponding responses. But if we chose to close the client connections all that requests and responses will be discarded. It's very likely that the client will reopen its connection and request the same resources one more time. So keeping client connection open may help us to reduce backed traffic.

In that particular case with `Transfer-Encoding: deflate` we can add our own encoding to keep the message framed with out connection close - we can encode the message to chunked encoding: `Transfer-Encoding: deflate, chunked`. The function `tfw_http_msg_to_chunked()` makes that, which called from tfw_http_resp_terminate.

If the flag `TFW_HTTP_B_CHUNKED_APPLIED` is set, then we can't encode the message to chunked encoding once again, and the only thing we can do to keep the framing - is to close the client connection.

tempesta_fw/http.c
ikoveshnikov: I have already cared of that case: https://github.com/tempesta-tech/tempesta/pull/1362/files#diff-d2642ed1c8bec1f31e559ba4871eaf51R817
Only `0\r\n\r\n` is added to message with zero body. And I have a functional test for that too: https://github.com/tempesta-tech/tempesta/pull/1362/files#diff-d2642ed1c8bec1f31e559ba4871eaf51R817

tempesta_fw/http.c
ikoveshnikov: I also have a test case for  `Transfer-Encoding: deflate` -> `Transfer-Encoding: deflate, chunked` conversion: https://github.com/tempesta-tech/tempesta-test/pull/145/files#diff-d1f29d9179db6bfb793086c219087719R198

i-rinat: Ah, now I see, I was wrong. Thanks for the clarifications.

tempesta_fw/http.c
i-rinat: Yeah, I was looking into the current (master branch) version of `tfw_http_msg_to_chunked()`, and completely forgot that this patch changes `tfw_http_msg_to_chunked()` too.

Issue #1335, opened 2019-08-11
TLS crypto extensions


krizhanovsky:
## Scope

Following algorithms must be implemented or optimized in Tempesta TLS:

- [ ] ChaCha20_Poly1305 (already is in the kernel, also required for QUIC, usually preferred by mobile devices) must be implemented. MbedTLS uses additional callback stream_func in mbedtls_cipher_base_t which is used by ChaCha20 and ARC4 only, but maybe we can find a better solution. 

- [ ] #1064 improves MPI, but doesn't takes specific step sto improve RSA performance, so the algorithm must be optimized.

## Testing

No need for a specific test for RSA, but a test for ChaCha20_Poly1305 ciphersuite is required.

krizhanovsky:
Updated benchmarks for ECDSA (performance core on i9-12900HK):
```
$ openssl version
OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022)

$ taskset --cpu-list 2 openssl speed ecdsa
....
compiler: gcc -fPIC -pthread -m64 -Wa,--noexecstack -Wall -Wa,--noexecstack -g -O2 -ffile-prefix-map=/build/openssl-olCZw9/openssl-3.0.2=. -flto=auto -ffat-lto-objects -flto=auto -ffat-lto-objects -fstack-protector-strong -Wformat -Werror=format-security -DOPENSSL_TLS_SECURITY_LEVEL=2 -DOPENSSL_USE_NODELETE -DL_ENDIAN -DOPENSSL_PIC -DOPENSSL_BUILDING_OPENSSL -DNDEBUG -Wdate-time -D_FORTIFY_SOURCE=2
CPUINFO: OPENSSL_ia32cap=0x7ffaf3ffffebffff:0x98c007bc239ca7eb
                              sign    verify    sign/s verify/s
 160 bits ecdsa (secp160r1)   0.0001s   0.0001s  10476.5   9825.9
 192 bits ecdsa (nistp192)   0.0001s   0.0001s   8490.5   8223.9
 224 bits ecdsa (nistp224)   0.0000s   0.0001s  34613.6  16034.2
 256 bits ecdsa (nistp256)   0.0000s   0.0000s  63743.3  20365.9
 384 bits ecdsa (nistp384)   0.0005s   0.0004s   2097.6   2455.4
 521 bits ecdsa (nistp521)   0.0002s   0.0003s   5812.7   2880.5
[...]

Issue #1054, opened 2018-08-11
TLS sessions resumption


krizhanovsky:
### Scope

Session tickets defined by [RFC 5077](https://tools.ietf.org/html/rfc5077) must be enabled. TLS cache is [considered](https://www.youtube.com/watch?v=nNrVjHaRFSk) as a bad practice due to unnecessary server memory usage, so I completely removed `TTLS_CACHE_C` option.

A new config option with 2 attributes must be introduced:
1. `random` or `secret` - define how a ticket key must be generated - using random generator or the static data for [tickets sharing](https://vincent.bernat.ch/en/blog/2011-ssl-session-reuse-rfc5077#sharing-tickets) among cluster nodes (a hash function must be used to not to send the secret as is).
2. the ticket lifetime in seconds

In case of using shared secret, the perfect forward secrecy must be kept and the secret must be regenerated in pseudo-random (to make sure that all cluster nodes have the same state) fashion.

It makes sense to use the same secret as for `sticky_secret`, so I propose to rename the option to `secret` and use it for the TLS tickets as well as for the session tickets. (Please update Wiki).

*Notice (probably good to add to Wiki)* it's attractive to reuse TLS Session ID instead of sticky cookie or vice versa, but Session ID is generated (and controlled) by a client, so we can not rely on it and can not force a client to use our cookie.

AES_128_GCM should be used (bedtls/programs/ssl/ssl_server2.c) for encryption.

Note and fix `TODO #1054` comments around the code.

### Handshakes limits

It's supposed that normal clients use TLS sessions instead of establish new TLS connections, so as part of the issue, a rate limit for the new handshakes (not only RSA as in #1038) must be introduced. This is the right way to [limit TLS](https://github.com/tempesta-tech/tempesta/issues/832). 

The limit must be per-vhost in sense of #715.

### Code reworking

Use `ttls_ticket_write()` and `ttls_ticket_parse()` directly instead of indirect calls of`f_ticket_write()` and `f_ticket_parse()` from `tls->conf`. This allows to fix record size estimation in `ttls_write_new_session_ticket()`.

`ssl_load_session()` must not parse X.509 each time, instead it must cache only the required information from the certificate.

I believe `ttls_ticket_context->mutex` in `ttls_ticket_parse()` should experience a serious contention, so the lock must be acquired only to update the key in `ssl_ticket_update_keys()` - all the other operations with the context are read only.

[...]

PR #1363, opened 2019-10-20
#1064: small MPI cleanups and improvements

krizhanovsky:
All changes are small, except bignum.[ch] where some small pieceis of logic were fully reworked.
    
1. MPI small optimizations and improvements:
1.1. introduce `used` to avoid zero skipping loops;
1.2. remove copies in several places;
1.3. reduce number and sizes of memset() calls    
2. Reworked MPI debugging facilities and debug.[ch]    
3. Remove secp192r1 as deprecated by RFC8422 5.1.1.
4. Finally get rid of config.h
5. Add MPI unit tests for basic operations to simplify debugging of new changes and adjust ECP and RSA tests.
6. Massive coding style fixes.    
7. Proper reference to mbed TLS in licensing.


tls/bignum.c
```diff
@@ -1,1009 +1,865 @@
...
-			TTLS_MPI_CHK(ttls_mpi_sub_abs(X, A, B));
-			X->s =  s;
-		}
-		else
-		{
-			TTLS_MPI_CHK(ttls_mpi_sub_abs(X, B, A));
+	int r, s = A->s;
+
+	if (A->s * B->s > 0) {
+		if ((r = ttls_mpi_cmp_abs(A, B)) >= 0) {
```
i-rinat: Value of `r` is not used afterwards, as it will be overwritten in both cases.

tls/oid.c
```diff
@@ -383,14 +382,6 @@ typedef struct {
 
 static const oid_ecp_grp_t oid_ecp_grp[] =
 {
-	{
-		{ ADD_LEN(TTLS_OID_EC_GRP_SECP192R1), "secp192r1",	"secp192r1" },
```
i-rinat: Definitions of TTLS_OID_EC_GRP_SECP192R1, TTLS_OID_EC_GRP_SECP224R1, TTLS_OID_EC_GRP_SECP192K1, TTLS_OID_EC_GRP_SECP224K1, and TTLS_OID_EC_GRP_SECP256K1 were left undeleted in the code. As the curves are deleted, those definitions are not needed.

tls/pk.h
```diff
@@ -136,7 +138,6 @@ int ttls_pk_decrypt(ttls_pk_context *ctx,
 int ttls_pk_encrypt(ttls_pk_context *ctx,
 		    const unsigned char *input, size_t ilen,
 		    unsigned char *output, size_t *olen, size_t osize);
-int ttls_pk_check_pair(const ttls_pk_context *pub, const ttls_pk_context *prv);
 int ttls_pk_debug(const ttls_pk_context *ctx, ttls_pk_debug_item *items);
```
i-rinat: There is no such function anymore, as it was inlined. So this declaration should be removed too.

tls/tls_srv.c
```diff
@@ -1094,9 +1095,8 @@ ttls_parse_client_hello(TlsCtx *tls, unsigned char *buf, size_t len,
 	if (ttls_ciphersuite_uses_ecdh(tls->xfrm.ciphersuite_info) ||
 	    ttls_ciphersuite_uses_ecdhe(tls->xfrm.ciphersuite_info))
 	{
-		unsigned char pf = tls->hs->ecdh_ctx.point_format;
 		ttls_ecdh_init(&tls->hs->ecdh_ctx);
-		tls->hs->ecdh_ctx.point_format = pf;
+		tls->hs->ecdh_ctx.point_format = tls->hs->ecdh_ctx.point_format;
```
i-rinat: This optimization introduces a bug: `.point_format` is cleared by `ttls_ecdh_init()`. We use `.point_format` to store point format during ClientHello parsing.

tls/ttls.c
```diff
@@ -261,44 +264,40 @@ ttls_xfrm_need_encrypt(TlsCtx *tls)
...
 	memcpy_fast(dst, src, sizeof(TlsSess));
 
 	if (src->peer_cert) {
 		int r;
 
-		dst->peer_cert = ttls_calloc(1, sizeof(ttls_x509_crt));
-		if (dst->peer_cert == NULL)
-			return(TTLS_ERR_ALLOC_FAILED);
+		dst->peer_cert = kmalloc(sizeof(ttls_x509_crt), GFP_ATOMIC);
+		if (dst->peer_cert)
```
i-rinat:
```suggestion
		if (!dst->peer_cert)
```

tls/ttls.c
```diff
@@ -2002,75 +2005,55 @@ int ttls_conf_dh_param_ctx(TlsCfg *conf, ttls_dhm_context *dhm_ctx)
...
-		ttls_free(tls->hostname);
+		kfree(tls->hostname);
 	}
 
-	/* Passing NULL as hostname shall clear the old one */
-
-	if (hostname == NULL)
-	{
+	/* Passing NULL as hostname shall clear the old one. */
+	if (hostname) {
```
i-rinat:
```suggestion
	if (!hostname) {
```

tls/ttls.c
```diff
@@ -2101,31 +2088,47 @@ ttls_conf_version(TlsCfg *conf, int min_minor, int max_minor)
...
+ * corresponding ttls_ticket_write_t function, and, if the ticket is authentic
+ * and valid, load the session.
+ *
+ * The implementation is allowed to modify the first len bytes of the input
+ * buffer, eg to use it as a temporary area for the decrypted ticket contents.
+ *
+ * On server, session tickets are enabled by providing non-NULL callbacks.
+ */
+void
+ttls_conf_session_tickets_cb(TlsCfg *conf,
```
i-rinat: Inconsistency: this code guarded by TTLS_SESSION_TICKETS is left intact, while at other places code is deleted.

tls/ttls.c
```diff
@@ -2101,31 +2088,47 @@ ttls_conf_version(TlsCfg *conf, int min_minor, int max_minor)
...
+int
+ttls_get_session(const TlsCtx *tls, TlsSess *dst)
 {
-	if (tls == NULL ||
-		dst == NULL ||
-		tls->conf->endpoint != TTLS_IS_CLIENT)
-	{
-		return(TTLS_ERR_BAD_INPUT_DATA);
-	}
+	if (!tls || dst || tls->conf->endpoint != TTLS_IS_CLIENT)
```
i-rinat:
```suggestion
	if (!tls || !dst || tls->conf->endpoint != TTLS_IS_CLIENT)
```

tls/x509_crt.c
```diff
@@ -72,7 +69,6 @@ const ttls_x509_crt_profile ttls_x509_crt_profile_next =
 	TTLS_X509_ID_FLAG(TTLS_ECP_DP_BP256R1) |
 	TTLS_X509_ID_FLAG(TTLS_ECP_DP_BP384R1) |
 	TTLS_X509_ID_FLAG(TTLS_ECP_DP_BP512R1) |
```
i-rinat:
```suggestion
	TTLS_X509_ID_FLAG(TTLS_ECP_DP_BP512R1),
```

tls/bignum.h
```diff
@@ -71,546 +48,120 @@
...
- * \brief		  Modulo: r = A mod b
+ * There are a lot of MPI operations used around, so Tempesta TLS becomes
+ * unusable if all MPIs are dumped, so following pattern should be used:
  *
- * \param r		Destination ttls_mpi_uint
- * \param A		Left-hand MPI
- * \param b		Integer to divide by
+ *	__mpi_do_dump = true;
+ *	....
+ *	some crypto operatrion to be debugged, e.g. ttls_ecp_mul()
```
ikoveshnikov:
```suggestion
 *	some crypto operation to be debugged, e.g. ttls_ecp_mul()
```

tls/ttls.c
```diff
@@ -2101,31 +2088,47 @@ ttls_conf_version(TlsCfg *conf, int min_minor, int max_minor)
...
-#if defined(TTLS_CLI_C)
-int ttls_get_session(const ttls_context *tls, ttls_ssl_session *dst)
+/**
+ * Save session in order to resume it later (client-side only).
+ * Session data is copied to presented session structure.
+ *
+ * WARNING Currently, peer certificate is lost in the operation.
+ */
+int
+ttls_get_session(const TlsCtx *tls, TlsSess *dst)
```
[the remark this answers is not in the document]
i-rinat: We already had
```c
typedef struct ttls_context {
   ...
} TlsCtx;
```
and
```c
typedef struct ttls_context ttls_context;
```
So `ttls_context` and `TlsCtx` are interchangeable. But that's confusing I must say.

tls/bignum.h
```diff
@@ -71,546 +48,120 @@
...
- * \param G		Destination MPI
- * \param A		Left-hand MPI
- * \param B		Right-hand MPI
- *
- * \return		 0 if successful,
- *				 TTLS_ERR_MPI_ALLOC_FAILED if memory allocation failed
- */
-int ttls_mpi_gcd(ttls_mpi *G, const ttls_mpi *A, const ttls_mpi *B);
+#define MPI_GROW_COPY		0x1
+#define MPI_GROW_ZERO		0x1
```
i-rinat:
```suggestion
#define MPI_GROW_ZERO		0x2
```

tls/tls_srv.c
krizhanovsky: Oops, my bad. Thank you!

tls/ttls.c
krizhanovsky: Yeah, the function must be deleted in #1054 (added a comment).

tls/asn1parse.c
```diff
@@ -249,8 +250,8 @@ int ttls_asn1_get_sequence_of(unsigned char **p,
 		/* Allocate and assign next pointer */
 		if (*p < end)
 		{
-			cur->next = (ttls_asn1_sequence*)ttls_calloc(1,
-		sizeof(ttls_asn1_sequence));
+			cur->next = kmalloc(sizeof(ttls_asn1_sequence),
```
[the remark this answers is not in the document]
krizhanovsky: We're good here: `cur->next` is assigned after the `while` loop, `buf->len` is assigned w/o reading in `ttls_asn1_get_tag()`, `buf->tag` and `buf->p` are also assigned in the loop.

tls/rsa.c
```diff
@@ -65,9 +66,9 @@ static inline int ttls_safer_memcmp(const void *a, const void *b, size_t n)
 }
 
 int ttls_rsa_import(ttls_rsa_context *ctx,
-			const ttls_mpi *N,
-			const ttls_mpi *P, const ttls_mpi *Q,
-			const ttls_mpi *D, const ttls_mpi *E)
+			const TlsMpi *N,
```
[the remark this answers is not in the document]
krizhanovsky: RSA is to be reworked in #1335 and it has a lot of coding style issues, so I thinks it's better to postpone coding style cleanups until we do something useful with the file. For now we care about ecp* , so let's focus on these files only.

tls/bignum.c
```diff
@@ -1025,1146 +881,794 @@ int ttls_mpi_sub_int(ttls_mpi *X, const ttls_mpi *A, ttls_mpi_sint b)
...
 	 */
-	if (_RR == NULL || _RR->p == NULL)
-	{
-		TTLS_MPI_CHK(ttls_mpi_lset(&RR, 1));
-		TTLS_MPI_CHK(ttls_mpi_shift_l(&RR, N->n * 2 * biL));
-		TTLS_MPI_CHK(ttls_mpi_mod_mpi(&RR, &RR, N));
-
-		if (_RR != NULL)
-			memcpy(_RR, &RR, sizeof(ttls_mpi));
+	if (!RR || !RR->p) {
```
i-rinat: This part makes no sense anymore. Previously, one could pass `NULL` as `_RR`, and in that case this function did the needed calculation. If `_RR` wasn't `NULL`, it was used as `RR`. Now the function requires a pre-computed `RR` and won't work correctly if there is `NULL`. Moreover, in that case `ttls_mpi_lset(RR, 1)` will be called, which will dereference `NULL`.

tls/bignum.c
```diff
@@ -1,1009 +1,865 @@
...
-	*olen = p - buf;
+	if (WARN_ON_ONCE(size > TTLS_MPI_MAX_SIZE))
+		return -EINVAL;
 
-cleanup:
+	if (__mpi_realloc(X, limbs, 0))
+		return -ENOMEM;
 
-	ttls_mpi_free(&T);
+	ttls_rnd(X->p, size);
```
i-rinat: This will fill `X->p` by `size` bytes of random data. However, remaining bytes in the last limb will be kept as they were after memory allocation. So it's possible to have random number larger than expected.

tls/bignum.c
```diff
@@ -1,1009 +1,865 @@
...
+	/* Modular comparison. */
+	return z == X->p[0] ? 0 : X->p[0] > z ? X->s : -X->s;
+}
 
-	if (X->n * biL < i)
-		TTLS_MPI_CHK(ttls_mpi_grow(X, BITS_TO_LIMBS(i)));
+/**
+ * Unsigned addition: X = |A| + |B|
+ *
+ * @A and @B must be different, but either of them can accept the result @X.
```
i-rinat: This may be a neat interface for a general purpose library, but that flexibility usually is not needed, as one knows whenever they do addition `X = A + B` where all numbers are in different memory blocks, or accumulation `A += B'. Flexibility makes code slower and more complex. I'd prefer to have two separate functions for the two cases.

tls/bignum.c
```diff
@@ -1025,1146 +881,794 @@ int ttls_mpi_sub_int(ttls_mpi *X, const ttls_mpi *A, ttls_mpi_sint b)
...
-		if (Q != NULL) TTLS_MPI_CHK(ttls_mpi_lset(Q, 0));
-		if (R != NULL) TTLS_MPI_CHK(ttls_mpi_copy(R, A));
+	if (!ttls_mpi_cmp_int(B, 0)) {
+		T_DBG_MPI1("Division by zero", B);
+		TTLS_MPI_DUMP_ONCE(B, "B/zero");
+		return -EINVAL;
+	}
+	if (!ttls_mpi_cmp_int(B, 1)) {
+		if (Q)
+			if (ttls_mpi_lset(Q, 1))
```
i-rinat: It's a division by one, right? Shouldn't `Q` receive the value of `A` then?

tls/ecp_curves.c
```diff
@@ -976,203 +735,87 @@ static int ecp_mod_p384(ttls_mpi *N)
...
+	if (M.used > P255_WIDTH + 1)
+		M.used = P255_WIDTH + 1;
 	M.p = Mp;
-	memset(Mp, 0, sizeof Mp);
-	memcpy(Mp, N->p + P255_WIDTH - 1, M.n * sizeof(ttls_mpi_uint));
-	TTLS_MPI_CHK(ttls_mpi_shift_r(&M, 255 % (8 * sizeof(ttls_mpi_uint))));
-	M.n++; /* Make room for multiplication by 19 */
+	n = M.used * CIL;
+	memcpy(Mp, N->p + P255_WIDTH - 1, n);
+	memset(Mp + n, 0, sizeof(Mp) - n);
```
i-rinat: `n` is number of bytes, but `Mp` is coerced to `unsigned long *`, so we may write beyond the `Mp` here.

PR #1371, opened 2019-11-12
Process method override

### LINKED ISSUE #1369: Detect Method Override in requests
According to HTTP standards client sets request method in the first line of its request. But some WAFs, anti-DDoS providers and other intermediates block requests with relatevely small whitelists of known methods. To allow passing not allowed method over WAFs some backends allow to override request methods. In this case a request will contain a whitelisted `GET` or `POST` method, but uri or header can define absolutely different request method.

E.g. from [one of IBM products](https://www.ibm.com/support/knowledgecenter/en/SSQP76_8.8.1/com.ibm.odm.dserver.rules.res.managing/topics/con_res_restapi_rsrcmng_methods.html)
```
Use the X-Method-Override or X-HTTP-Method-Override HTTP header fields to channel a PUT or DELETE request through a POST request.
Use the x-method-override or the x-http-method-override URI parameters.
For example: POST /api/v1/ruleapps?...&x-method-override=PUT
``` 

The [CPDoS](https://cpdos.org/#HMO) tells about cache poisoning using method override, but they don't tell that client may force intermediate to cache response with wrong method.

## Affect on cache

Attacker may override `GET` method but make us cache it as `GET` so all other clients will receive incorrect cache version. 

At this moment we have a list of 'known' methods, all other methods are marked as `_TFW_HTTP_METH_UNKNOWN`. But in https://github.com/tempesta-tech/tempesta/blob/97f974fcb2e1da51eac4bdbb5d80496988e4458a/tempesta_fw/cache.c#L746 we don't use string comparison to match the unknown methods. Its a bug.

## Affect on message pipeline

Attacker may cover `non-idempotent` method under `idempotent` one. Don't know why, but...

## Affect on frang

System administrator may whitelist some methods. Overriding methods looks line an attack in this case.

## How to fix

Detect method override on parser stage and override `req->method`. Moreover, add string HTTPMessage member to point on unknown method string in the request, and evaluate it in `tfw_cache_entry_key_eq`.

ikoveshnikov:
fix #1369 

Method override masks real request properties, non-idempotent methods can hide behind idempotent, method is used as a key in cache subsystem to store and look up cached responses. Thus hiding real method can spoil responses for other clients. Use the real method for accurate processing.

Don't rewrite the method string and don't remove override header since there can be additional intermediates between TempestaFW and backend.

### DIFF
--- a/etc/tempesta_fw.conf
+++ b/etc/tempesta_fw.conf
@@ -984,6 +984,7 @@
 #       http_host_required true|false;
 #       http_methods [METHOD]...;
 #       http_ct_required true|false;
+#       http_method_override_allowed true|false;
 #       http_ct_vals ["CONTENT_TYPE"]...;
 #       http_resp_code_block RESPONSE_CODE [RESPONSE_CODE]...
 #                            LIMIT TIME_FRAME_IN_SECONDS;
--- a/tempesta_fw/http_limits.c
+++ b/tempesta_fw/http_limits.c
@@ -539,6 +539,25 @@ frang_http_methods(const TfwHttpReq *req, FrangAcc *ra, unsigned long m_mask)
 }
 
 static int
+frang_http_methods_override(const TfwHttpReq *req, FrangAcc *ra,
+			    FrangVhostCfg *f_cfg)
+{
+	unsigned long mbit = (1UL << req->method_override);
+
+	if (!req->method_override)
+		return TFW_PASS;
+	if (!f_cfg->http_method_override
+	    || (f_cfg->http_methods_mask && !(f_cfg->http_methods_mask & mbit)))
+	{
+		frang_msg("restricted overridden HTTP method",
+			  &FRANG_ACC2CLI(ra)->addr, ": %u (%#lxu)\n",
+			  req->method_override, mbit);
+		return TFW_BLOCK;
+	}
+	return TFW_PASS;
+}
+
+static int
 frang_http_ct_check(const TfwHttpReq *req, FrangAcc *ra, FrangCtVals *ct_vals)
 {
 	TfwStr field, *s;
@@ -1013,6 +1032,10 @@ frang_http_req_process(FrangAcc *ra, TfwConn *conn, TfwFsmData *data,
 		{
 			T_FSM_EXIT();
 		}
+		/* Ensure overridden HTTP method suits restrictions. */
+		r = frang_http_methods_override(req, ra, f_cfg);
+		if (r)
+			T_FSM_EXIT();
 		/*
 		* Ensure presence of Content-Type: header field.
 		* Ensure that the value is one of those defined by a user.
--- a/tempesta_fw/http_parser.c
+++ b/tempesta_fw/http_parser.c
@@ -3204,7 +3204,6 @@ __parse_m_override(TfwHttpReq *req, unsigned char *data, size_t len)
 	__FSM_STATE(I_Meth_Unknown) {
 		__FSM_I_MATCH_MOVE(token, I_Meth_Unknown);
 		req->method_override = _TFW_HTTP_METH_UNKNOWN;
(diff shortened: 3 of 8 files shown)


etc/tempesta_fw.conf
```diff
@@ -984,6 +984,7 @@
 #       http_host_required true|false;
 #       http_methods [METHOD]...;
 #       http_ct_required true|false;
+#       http_method_override_allowed true|false;
```
krizhanovsky: Please describe the option in the Wiki, preferably with the reference to https://cpdos.org/

tempesta_fw/http_limits.c
```diff
@@ -538,6 +538,25 @@ frang_http_methods(const TfwHttpReq *req, FrangAcc *ra, unsigned long m_mask)
...
+static int
+frang_http_methods_override(const TfwHttpReq *req, FrangAcc *ra,
+			    FrangVhostCfg *f_cfg)
+{
+	unsigned long mbit = (1UL << req->method_override);
+
+	if (!req->method_override)
+		return TFW_PASS;
+	if (!f_cfg->http_method_override
+	    || (f_cfg->http_methods_mask && !(f_cfg->http_methods_mask & mbit)))
```
krizhanovsky: The second part of the expression is equal to `!(f_cfg->http_methods_mask & mbit)`, i.e. `f_cfg->http_methods_mask` is unnecessary

ikoveshnikov: The `f_cfg->http_methods_mask` part in condition is actually required. If the `http_methods` directive under `frang_limits` section isn't declared, then all methods are allowed, and `f_cfg->http_methods_mask`  is `0`.  That's not great and it seems like the mask should be `0xffffffff` by default, but this option follows frang global configuration rules: if something is disabled it's zero.

Thus we have three options how to deal with overridden method:
- block a request if overriding is prohibited.,
- pass a request if overriding is allowed and method restriction is not used.
- block a request if overridden method is prohibited, pass the request otherwise.

tempesta_fw/http_parser.c
```diff
@@ -3072,6 +3072,157 @@ __parse_uri_mark(TfwHttpReq *req, unsigned char *data, size_t len)
...
+			req->method_override = TFW_HTTP_METH_UNLOCK;
+		} , I_Meth_U, I_EoT);
+		TRY_STR_INIT();
+		__FSM_I_JMP(I_Meth_Unknown);
+	}
+
+	__FSM_STATE(I_Meth_Unknown) {
+		__FSM_I_MATCH_MOVE(token, I_Meth_Unknown);
+		req->method_override = _TFW_HTTP_METH_UNKNOWN;
+		c = *(p + __fsm_sz);
```
aleksostapenko: It seems that `c` is not used here.

tempesta_fw/http_parser.c
```diff
@@ -3072,6 +3072,157 @@ __parse_uri_mark(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		__FSM_I_MOVE_n(I_EoT, __fsm_sz);
```
aleksostapenko: Looks like we can do `p += __fsm_sz` here and then just fall through to the `I_EoT` state (instead of `__FSM_I_MOVE_n()` using) to avoid of repeating of the check that we had already done in `__FSM_I_MATCH_MOVE()` above.

tempesta_fw/http_parser.c
ikoveshnikov: Thanks!

tempesta_fw/http_parser.c
ikoveshnikov: Yes, you're right, fall-trough would have absolutely the same behaviour here. But this particular case is not a hot-path, so I prefered to have explicit `move` statement here.

PR #1372, opened 2019-11-24
tls: mark-up temporary storage for clienthello parser

### LINKED ISSUE #1265: TLS: ClientHello parser simplifications
(current version: df9088483e51c12513c04bc2f4b941f5d7eee2dd)

There are couple cosmetic issues with the `ttls_parse_client_hello()`.
* It uses `tls->hs->tmp` array a lot with explicitly calculated offsets. That makes it harder to inspect code to ensure there are no buffer overruns. That also makes changes cumbersome. It would be better to just use structure fields. Consider using an unnamed struct (one or more) for such fields.

* Two-byte numbers are stored in network order in a memory buffer that never goes outside. That forces to use `htons()` and `ntohs()` more often than necessary. Consider storing such values in native byte order.

i-rinat:
Transform the uniform `tls->hs->tmp` array into an object with named fields, to (hopefully) make its usage more explicit in the code. Additionally, all stored values are now stored in host byte order to avoid calling `ntohs()` or `htons()` each time we want to access the numbers.

Previously, a single buffer was used for both ciphersuite list and temporary extension linear buffer. Ciphersuite list could occupate almost the whole buffer, however 32 bytes were reserved for the temporary extension buffer. If ciphersuite list took less amount of storage, extension buffer had more capacity. It turns out that 32 bytes is not enough, and even in tests there are extensions which are at least 34 bytes in length. So if a client sends an unusually large list of ciphersuites, parse will fail at extension parsing stage.

This patch modifies code behavior a little bit. Now both ciphersuite list buffer and extension buffer are fixed-length. Extension buffer is 256 bytes long, and ciphersuite list takes what remains of about 1k of space. Should be enough for both.

Fixes #1265.

### DIFF
--- a/tls/tls_internal.h
+++ b/tls/tls_internal.h
@@ -168,7 +168,27 @@ struct tls_handshake_t {
 	};
 	union {
 		unsigned char		premaster[TTLS_PREMASTER_SIZE];
-		unsigned char		tmp[TTLS_HS_RBUF_SZ];
+		struct {
+			union {
+				unsigned short cs_cur_len;
+				struct {
+					unsigned char compr_n;
+					unsigned char compr_has_null;
+				};
+				struct {
+					unsigned short ext_rem_sz;
+					unsigned short ext_type;
+					unsigned short ext_sz;
+				};
+				unsigned char *cert_page_address;
+			};
+			unsigned short cs_total_len;
+			struct {
+				u16 css[(TTLS_HS_RBUF_SZ - 10 - 256)/2];
+				unsigned char ext[256];
+			};
+		};
+		unsigned char key_exchange_tmp[TTLS_HS_RBUF_SZ];
 	};
 };
 
@@ -293,8 +313,6 @@ int ttls_check_cert_usage(const ttls_x509_crt *cert,
 			  int cert_endpoint,
 			  uint32_t *flags);
 
-void ttls_read_version(TlsCtx *tls, const unsigned char ver[2]);
-
 int ttls_get_key_exchange_md_tls1_2(TlsCtx *tls,
 		unsigned char *output,
 		unsigned char *data, size_t data_len,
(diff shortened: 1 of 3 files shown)


tls/tls_internal.h
```diff
@@ -168,7 +168,29 @@ struct tls_handshake_t {
 	};
 	union {
 		unsigned char		premaster[TTLS_PREMASTER_SIZE];
-		unsigned char		tmp[TTLS_HS_RBUF_SZ];
+		struct {
+			union {
+				struct {
+					unsigned short cs_cur_len;
+				};
```
krizhanovsky: the wrapping unnamed `struct` seems unnecessary

tls/tls_srv.c
```diff
@@ -730,22 +727,20 @@ ttls_parse_client_hello(TlsCtx *tls, unsigned char *buf, size_t len,
...
 		BUG_ON(io->rlen >= 2);
 		if (unlikely(io->rlen)) {
-			tls->hs->tmp[TTLS_HS_TMP_STORE_SZ + 1] = *p;
-			p++;
+			tls->hs->cs_total_len += *p++;
 		}
 		else if (unlikely(buf + len - p == 1)) {
-			tls->hs->tmp[TTLS_HS_TMP_STORE_SZ] = *p;
-			p++;
+			tls->hs->cs_total_len = *p++ << 8;
```
krizhanovsky: Is `*p++` converted to `unsigned short` before the shift? Isn't it better to convert the type explicitly?

tls/tls_internal.h
i-rinat: Removed unnamed struct here, as it's not needed.

tls/tls_srv.c
i-rinat: > Is `*p++` converted to `unsigned short` before the shift?

Well, yes and no.
Integer promotions are applied to both operands of a bit-shift operator. Since `*p` is `unsigned char`, which has lower rank than `int`, it's converted to `int`. Then result of the operation is clamped to fit into `tls->hs->cs_total_len` which is `unsigned short`. As result is always representable in `unsigned short`, no information loss occurs.
So although it's not converted to `unsigned short`, but to `int`, operation behaves as if it's `unsigned short` was here instead.

Would you like me to put explicit casts to `unsigned short` in all similar cases in this patch?

krizhanovsky: No, I think we're good.

PR #1374, opened 2019-12-24
Fix request h2 -> http1.1 conversion

ikoveshnikov:
Main changes:
1. Keep h_tbl always valid since we need to read the table during
   response caching. Keep the unsent skbs within request to avoid huge
   updates of h_tbl
2. Optimize header alteration for h2, use compatibility mode for http1.1
3. Use tfw_msg_write functions to take care of all possible caveats at
   message creation stage. (E.g. headers may be longer than one page)

Known issues:
1. Body and trailer conversion is not tested due to parsing issues:
   body is not appended to the request and WARN_ON is triggered.


tempesta_fw/http.c
```diff
@@ -2916,125 +2949,145 @@ tfw_h2_req_set_loc_hdrs(TfwHttpReq *req)
...
 {
 	int r;
-	char *dst;
-	struct page *pg;
-	unsigned int hid;
-	TfwStr method;
-	const TfwStr *fld, *fld_end, *hdr, *dup_end, *chunk, *end;
-	TfwHttpMsg *hm = (TfwHttpMsg *)req;
+	TfwMsgParseIter *pit = &req->pit;
+	ssize_t h1_hdrs_sz = 0;
```
aleksostapenko: Seems, that this zeroing is redundant now.

tempesta_fw/http.c
```diff
@@ -2916,125 +2949,145 @@ tfw_h2_req_set_loc_hdrs(TfwHttpReq *req)
...
+		 * field.
+		 * AND
+		 * Clients that generate HTTP/2 requests directly SHOULD use
+		 * the :authority pseudo-header field instead of the Host
+		 * header field.
+		 */
+		if (host) {
+			h1_hdrs_sz -= ht->tbl[TFW_HTTP_HDR_HOST].len;
+			h1_hdrs_sz -= SLEN(S_H2_AUTH);
+			h1_hdrs_sz += SLEN(S_F_HOST);
```
aleksostapenko: Looks like excessive `SLEN(S_DLM)` counted here for `Host` header; it seems, that this size has already counted above, in the following code:
```
h1_hdrs_sz = pit->hdrs_len
		+ (pit->hdrs_cnt - pseudo_num) * (SLEN(S_DLM) + SLEN(S_CRLF))
		- req->mark.len;
```

tempesta_fw/http.c
```diff
@@ -3043,124 +3096,138 @@ do {									\
...
-	       it->hdrs_len, it->hdrs_cnt, dst, (char *)page_address(pg));
+	old_head = req->msg.skb_head;
+	req->msg.skb_head = new_head;
 
-	WARN_ON_ONCE(dst - (char *)page_address(pg) != it->hdrs_len);
+	/* Http chains might add a mark for the message, keep it. */
+	new_head->mark = old_head->mark;
 
-	r = ss_skb_replace_page(&req->msg.skb_head, pg, it->hdrs_len,
-				it->hb_len);
```
aleksostapenko: The `hb_len` member of `TfwMsgParseIter` structure is not used any more, and can be removed.

tempesta_fw/http.c
```diff
@@ -3043,124 +3096,138 @@ do {									\
...
+			trailer = trailer->next;
+
+		} while ((trailer != b_skbs) || (len != req->body.len));
+		ss_skb_queue_tail(&req->msg.skb_head, b_skbs);
+		if (trailer != b_skbs) {
+			ss_skb_queue_split(req->msg.skb_head, trailer);
+			ss_skb_queue_tail(&old_head, trailer);
+		}
 	}
+	ss_skb_queue_purge(&old_head);
```
aleksostapenko: The `ss_skb_replace_page()` function is not used any more, and looks like it can be removed.

tempesta_fw/http.c
```diff
@@ -2916,125 +2949,145 @@ tfw_h2_req_set_loc_hdrs(TfwHttpReq *req)
...
+			h1_hdrs_sz += SLEN(S_F_HOST);
+		}
+		else {
+			h1_hdrs_sz -= SLEN(S_H2_AUTH);
+			h1_hdrs_sz += SLEN(S_F_HOST) + SLEN(S_CRLF);
+		}
+	}
+
+	/* 'x-forwarded-for' header must be updated. */
+	xff_hid = __h2_hdr_lookup((TfwHttpMsg *)req, TFW_STR_CHUNK(&h_xff, 0));
```
aleksostapenko: `xff` is a special header, and looks like we can access it by index.

tempesta_fw/http.c
ikoveshnikov: Fixed, please check once more time.

aleksostapenko: I mean - from the previous code:
```
if (host) {
                        h1_hdrs_sz -= ht->tbl[TFW_HTTP_HDR_HOST].len;
                        h1_hdrs_sz -= SLEN(S_H2_AUTH);
                        h1_hdrs_sz += SLEN(S_F_HOST);
                }
                else {
                        h1_hdrs_sz -= SLEN(S_H2_AUTH);
                        h1_hdrs_sz += SLEN(S_F_HOST) + SLEN(S_CRLF);
                }
```
just to make 
```
if (host) {
                        h1_hdrs_sz -= ht->tbl[TFW_HTTP_HDR_HOST].len;
                        h1_hdrs_sz -= SLEN(S_H2_AUTH);
                        h1_hdrs_sz += SLEN(S_F_HOST) - SLEN(S_DLM);
                }
                else {
                        h1_hdrs_sz -= SLEN(S_H2_AUTH);
                        h1_hdrs_sz += SLEN(S_F_HOST) + SLEN(S_CRLF);
                }

```
to evict `SLEN(S_DLM)` length from `SLEN(S_F_HOST)`, since `SLEN(S_DLM)` for `Host` header (as well as `SLEN(S_CRLF)`) is already contained in `h1_hdrs_sz` in case of `Host` header existence in source HTTP/2-request.

ikoveshnikov: Let me clear up the calculations:
- `:authority` psedo-header _is counted_ in `h1_hdrs_sz`,
- if `host` header is present it is counted in `h1_hdrs_sz` and `SLEN(S_DLM) + SLEN(S_CRLF)` was added.

So here we need to make some alterations (assuming that the `:authority` pseudo header exists):

If the `host` header exists, we replace its value with the `:authority` ones. So we  need to:
- subtract  original length of `host` header and length of `S_DLM` and `S_CRLF` added to it;
- we need subtract difference between `:authority` and `host` length: 
```
h1_hdrs_sz -= SLEN(S_H2_AUTH);
/* S_F_HOST already contains S_DLM, but we already calculated `S_DLM` for authority header. */
h1_hdrs_sz += SLEN(S_F_HOST) - SLEN(S_DLM);
```

If host header doesn't exist, the we only need to rename `:authority` into `host`.

That's why che calculations are done in current way.

aleksostapenko: >:authority psedo-header is counted in h1_hdrs_sz

Yes, but in previous version of code the `:authority` pseudo-header had not been counted in `pit->hdrs_cnt` (i.e. subtracted from it), so the `SLEN(S_DLM) + SLEN(S_CRLF)` was not added for it and  - that's why my previous comment appeared. But with last correction in ['fix header calculations'](https://github.com/tempesta-tech/tempesta/pull/1374/commits/b4e2719509d7afd010fba99ae7556781d1c0b056) commit - it seems the problem is gone.
By the way: the `pseudo_num` local variable is constant now and used only once, thus maybe it should be better to move its definition to the beginning of the function, or just evict it (and replace with `3` in the place of usage with adding appropriate comment there).

tempesta_fw/http.c
```diff
@@ -849,6 +849,15 @@ tfw_h2_send_resp(TfwHttpReq *req, int status, unsigned int stream_id)
 	mit = &resp->mit;
 	skb_head = &resp->msg.skb_head;
 	body = TFW_STR_BODY_CH(msg);
+	if (WARN_ON_ONCE(body->len > ctx->rsettings.max_frame_sz)) {
+		/*
+		 * TODO: split body to frames it it's too long.
+		 * Since full-featured response is not constructed construct
+		 * here, it's possible to avoid framing body for each h2 client
+		 * individually. Every client must support 16384-byte frames.
+		 */
```
krizhanovsky: Please add issue numbers for TODO - it's handy to see when exactly and in context of which issues a particular TODO must be done

tempesta_fw/http.c
```diff
@@ -4632,30 +4765,43 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, const TfwFsmData *data)
...
-				  &data_up);
-		T_DBG3("TFW_HTTP_FSM_REQ_CHUNK return code %d\n", r);
-		if (r == TFW_BLOCK) {
-			TFW_INC_STAT_BH(clnt.msgs_filtout);
-			tfw_http_req_parse_block(req, 403,
-						 "postponed request has been"
-						 " filtered out");
-			return TFW_BLOCK;
+		if (TFW_MSG_H2(req) && tfw_h2_stream_req_complete(req->stream)) {
+			if (tfw_h2_parse_req_finish(req)) {
```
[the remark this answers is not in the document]
ikoveshnikov: We need this check here, because of all that skb fragment processing effects inside of `ss_skb_process()`.

How do we process h2 traffic:
- First we try to find frame borders in `tfw_h2_frame_process()`. We use  `ss_skb_process()`  to walk among skb fragments and split skb at frame end. At the end of this stage we have skb which contain only h2 headers. Let's assume that the headers is fully received, no body required for that request and the skb is fragmented. So the `STREAM_END`  flag will be received there.
- Since the `STREAM_END` flag is received, we mark the stream as finished to indicate request end for parser layer.
- Then the fragmented skb is passed into `tfw_http_req_process()` to parse the request. With `ss_skb_process()` we call `tfw_h2_parse_req()` _for each_ skb fragment.

Since the `tfw_h2_parse_req()` is called for every skb fragment, we can't have `tfw_h2_parse_req_finish()` inside it. The stream is _already_ marked as finished and after the very first fragment was parsed, the `tfw_h2_parse_req_finish()` would be called. The remaining fragments will remain unparsed and glued with the parsed ones.

I understand your wish to hide all the parser routines (and h2 specific processing) under single `actor` for `ss_skb_process()`. But since in h2  _the framing_ is done on framing level, not on parser one, we either need to explicitly call a parser function for the message end (in current patch), or make it implicitly by some extra state sharing.

I decided that the next parser API looks a bit cleaner:
- one function that parses a message from contiguous buffer - `tfw_h2_parse_req()`. It never return `T_OK` on success, only `T_POSTPONE`, since in h2 _the framing_ is done on framing level, not on parser one. 
- one function to explicitly indicate the message end and fixup the parser routines - `tfw_h2_parse_req_finish()`

tempesta_fw/http_frame.c
```diff
@@ -1177,6 +1183,17 @@ tfw_h2_frame_type_process(TfwH2Ctx *ctx)
...
+	if (unlikely(ctx->hdr.length > ctx->lsettings.max_frame_sz))
+		goto conn_term;
+
+	/*
+	 * TODO: RFC 7540 Section 6.2:
+	 * A HEADERS frame without the END_HEADERS flag set MUST be followed
+	 * by a CONTINUATION frame for the same stream. A receiver MUST treat
+	 * the receipt of any other type of frame or a frame on a different
+	 * stream as a connection error (Section 5.4.1) of type PROTOCOL_ERROR.
+	 */
```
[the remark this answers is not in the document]
ikoveshnikov: > The verification of HEADERS/CONTINUATION frame sequence is done in tfw_h2_stream_fsm()

Yes, no problem there. 

> The different streams intersection

Yes, that is what I meant. If peer has sent a `HEADER` frame without `HEADERS_END` flag, then he can't sent messages on other streams until the headers are closed with `CONTINUATION` frames followed by `HEADERS_END` flag.

>  e.g. via implementing the active_stream number in TfwH2Ctx structure, and setting/zeroing it for appropriate stream states in tfw_h2_stream_fsm().

Seems like it's enough to set some `ON_CONTINUATION` flag for the whole connection and check that the current stream ID is the same as the `lstream_id`

tempesta_fw/http.c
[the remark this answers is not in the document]
aleksostapenko: Yeah, that's about skb fragments - I forgot about them. Thanks!

tempesta_fw/http.c
```diff
@@ -4102,7 +4229,13 @@ tfw_http_req_block(TfwHttpReq *req, int status, const char *msg)
...
 }
 
-
+/*
+ * TODO: http1 headers usually use mixed case, while h2 uses only lower case
+ * header names. It seems that usual browsers are OK with mixed case in h2
+ * responses, but this behaviour is not defined by RFC and implementation
+ * dependent. Curl is not happy with the mixed case and return parser errors.
+ * The headers should be converted to lower case before transmission.
+ */
```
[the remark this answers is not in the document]
ikoveshnikov: Oh, yes. I was confused by `Transfer-Encoding` filtered by curl. It wasn't because of the case, but because the header itself.

From the [RFC 7540 8.1.2](https://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2)
 > Just as in HTTP/1.x, header field names are strings of ASCII characters that are compared in a case-insensitive fashion. However, header field names MUST be converted to lowercase prior to their encoding in HTTP/2. A request or response containing uppercase header field names MUST be treated as malformed (Section 8.1.2.6).

According the RFC an end point must close a connection (send error) on receiving our responses, but major browsers and curl _ignore_ this requirement. Well, it's a problem for us, actually. Converting header names to lower case definitely has performance impact. But even if browsers are OK, we can't just brake the RFC requirement and leave headers names in mixed case.

tempesta_fw/http.c
```diff
@@ -2916,124 +2976,163 @@ tfw_h2_req_set_loc_hdrs(TfwHttpReq *req)
...
+	}
 
 	/*
-	 * Conditional substitution/addition/deletion or appending of statically
-	 * configured headers.
+	 * TODO: remove from header table all hop-by-hop headers, including
+	 * 'Connection:' header (hdrs with flag TFW_STR_HBH_HDR). If do this
+	 * before tfw_h2_req_set_loc_hdrs(), we may free some space in the table
+	 * and avoid the table reallocation.
+	 */
```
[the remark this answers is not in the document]
ikoveshnikov: Oh, my fault here. H2 requests mustn't contain HBH-headers, but HBH-headers may be present in forwarded responses and mustn't be sent to a h2 client. I looked into wrong function and thought that something is missed. But everything is good with the HBH-headers. I will remove the misleading TODO comment.

tempesta_fw/http_parser.c
```diff
@@ -7999,23 +7994,56 @@ tfw_h2_parse_req(void *req_data, unsigned char *data, size_t len,
 	 * procedures yet, it must be deleted during corresponding stream
 	 * removal.
 	 */
-	if (tfw_h2_stream_req_complete(req->stream)) {
-		TfwHttpHdrTbl *ht = req->h_tbl;
```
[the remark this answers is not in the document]
ikoveshnikov: Comment is reworked, and second part is dropped: it repeats already existing comments.

tempesta_fw/http_frame.c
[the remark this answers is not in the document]
aleksostapenko: The problem with `lstream_id` - is that it will not allow to handle this case for requests (i.e. streams) with HEADERS/CONTINUATION frames after DATA frames.

PR #1375, opened 2019-12-26
#1064: TLS performance imporovements

krizhanovsky:
### #1064 point 1.

Introduce memory pools for the MPIs. MPI memory profile is an MPI memory pool filled at configuration time, so that we can use stream copy to quickly create ready to use MPI image for a particular PK computation.

### #1064 point 2

Reduce number of public key checkings and assembly implementations for the core math.

### TODO

The PR is buggy and to be fixed during the review.

Sorry for the large commits of mixed logical changes and cleanups.


tls/tls_srv.c
```diff
@@ -239,15 +232,10 @@ ttls_parse_supported_point_formats(TlsCtx *tls, const unsigned char *buf,
...
-		{
-			tls->hs->ecdh_ctx.point_format = p[0];
-			T_DBG("ClientHello: point format selected: %d\n", p[0]);
-			return 0;
-		}
-	}
+	/*
+	 * Uncompressed is the only point format supported by RFC 8422,
+	 * so just ignore the list.
+	 */
```
i-rinat: The control flow of this function looks wrong — returned code is always zero regardless of the actual buffer contents. It doesn't matter now in our fork when all cases were stripped. I get it.
But in the mbedtls code they process the buffer, and if no supported point format was found, return success leaving `.point_format` uninitialized. I wonder why?

tls/tls_srv.c
```diff
@@ -2160,6 +1993,23 @@ ttls_handshake_server_hello(TlsCtx *tls)
 		if ((r = ttls_write_server_hello(tls, &sgt, &p)))
 			T_FSM_EXIT();
 		CHECK_STATE(128);
+		if ((tls->state = TTLS_SERVER_CHANGE_CIPHER_SPEC)) {
```
i-rinat: Do we really need an assignment here? That looks suspicious.
If for some reason this is what was intended, condition is always true, and need not to be checked.

tls/ciphersuites.h
```diff
@@ -115,101 +80,19 @@ typedef struct {
 	int				max_major_ver;
 	int				max_minor_ver;
 	unsigned char			flags;
+	TlsMpiPool			*mpi_profile[__TTLS_ECP_DP_N];
```
i-rinat: Although type says "pool", variable name is about "profile". That may be interpreted as equivalence of the two terms. But as I understood from the code, profiles are not really about memory, but about the blueprints of the data, if I can put it that way.
It would be nice to have a little more clarity here. Unfortunately, I can't offer an acceptable solution.

tls/bignum.h
```diff
@@ -65,27 +71,41 @@ do {									\
...
+ * so @_off is typically quite small.
  */
 typedef struct {
-	int		s;
+	short		s;
 	unsigned short	used;
 	unsigned short	limbs;
-	unsigned long	*p;
+	unsigned short	_off;
 } __attribute__((packed)) TlsMpi;
```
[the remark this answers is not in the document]
krizhanovsky: Now it's not packed when I added 12 more bytes in debug mode

tls/tls_srv.c
krizhanovsky: All the memory is zeroized in mbed TLS, so technically the point format is equal to `MBEDTLS_ECP_PF_UNCOMPRESSED`. Probably, the constant is zero to be the default value.

tls/tls_srv.c
krizhanovsky: Yeah, that was a bug, there must be `if (tls->state == TTLS_SERVER_CHANGE_CIPHER_SPEC)`. Thank you!

tls/ciphersuites.h
krizhanovsky: Profiles are pools actually: we have the cipher suites static MPI pools and use then on the module initialization - allocate all necessary PK contexts. In run time, durinag a handshake, we create per-handshake MPI pool and just copy an appropriate cipher suite profile. Since we copy using SIMD, this is much faster than initialize each MPI, even with the allocation using the pools.

I wrote small comment about this at the header of `mpool.c`:
> The pool also used for MPI profiles (see below) to initialize a profile
> implicitly for MPI math. Dynamically allocated pages are used instead of
> static per-cpu ones.

tls/ecdsa.c
```diff
@@ -251,15 +165,16 @@ static int ecdsa_signature_to_asn1(const TlsMpi *r, const TlsMpi *s,
 	TTLS_ASN1_CHK_ADD(len, ttls_asn1_write_tag(&p, buf,
 			   TTLS_ASN1_CONSTRUCTED | TTLS_ASN1_SEQUENCE));
 
-	memcpy(sig, p, len);
+	memcpy_fast(sig, p, len);
```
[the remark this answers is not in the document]
krizhanovsky: Yeah, I added TODO comment here. The thing is that we write from the end to the begin and don't know the resulting buffer size. Meantime the signature length is relatively small and probably we won't win here anything.

tls/mpool.c
```diff
@@ -0,0 +1,432 @@
...
+	if (WARN_ON_ONCE(sp < a && a < sp + 2 * PAGE_SIZE))
+		return NULL;
+
+	mp = *this_cpu_ptr(&g_tmp_mpool);
+	mp_name = "temporary";
+	if ((unsigned long)mp < a && a < (unsigned long)mp + 2 * PAGE_SIZE)
+		goto check;
+
+	/* FIXME need a safer determination of the pool address. */
+	mp = (TlsMpiPool *)(a & ~((PAGE_SIZE << TTLS_MPOOL_ORDER) - 1));
```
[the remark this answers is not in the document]
krizhanovsky: I added comment for `ttls_mpi_pool_create()` that the pools are allocated using direct call to the buddy system, so higher order pages are buddies and always aligned. I also added `WARN_ON_ONCE()` contract to `ttls_mpi_pool_create()` for the alignment guarantees,

tls/mpool.c
```diff
@@ -0,0 +1,432 @@
...
+	mp = ci->mpi_profile[ec];
+	if (WARN_ON_ONCE(!mp)) {
+		T_WARN("Try to do %s handshake w/o an MPI profile\n",
+		       ci->name);
+		return -ENOENT;
+	}
+
+	p_sz = MPI_PROFILE_SZ(mp);
+	/* TODO #1064 doesn't it make sense to use full pages? */
+	ptr = pg_skb_alloc(p_sz, GFP_ATOMIC, NUMA_NO_NODE);
```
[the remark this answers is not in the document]
krizhanovsky: Yeah, I moved to the page allocator: in fact the PK memory requirements are quite large, i.e. the typical EC handshake requires about 6KB and there is no really sense to use the page chunks.

tls/pk.c
```diff
@@ -23,18 +23,192 @@
...
-#include "ecdsa.h"
+#include "mpool.h"
+#include "pk.h"
+#include "rsa.h"
+
+/* Defined in ecdsa.c */
+extern int ttls_ecdsa_write_signature(TlsEcpKeypair *ctx,
+				      const unsigned char *hash, size_t hlen,
+				      unsigned char *sig, size_t *slen);
+extern int ttls_ecdsa_read_signature(TlsEcpKeypair *ctx,
```
i-rinat: Placing prototypes here is dangerous. If a prototype is in the header file, and during some refactoring parameter list in the function definition changes, compiler will warn about the ismatch. But if the prototype only defined here, no warnings and no errors will occur, but arguments won't be passed as intended.
So, why?

tls/bignum.c
```diff
@@ -135,57 +143,40 @@ mpi_fixup_used(TlsMpi *X, size_t n)
...
+		X->s = 1;
 
-	memcpy(p, X->p, X->used * CIL);
-	kfree(X->p);
+		return 0;
+	}
 
-	X->limbs = nblimbs;
-	X->p = p;
+	if (need_alloc)
```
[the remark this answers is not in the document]
krizhanovsky: Yeah. I added `BUG_ON()`.

tls/pk.c
krizhanovsky: I just don't see much sense to keep a separate `.h` file just for 2 function prototypes and which is used in one place only.

tls/rsa.c
```diff
@@ -1449,54 +1154,34 @@ static int rsa_rsassa_pkcs1_v15_encode(ttls_md_type_t md_alg,
...
+	if (!(sig_try = kzalloc(ctx->len * 2, GFP_ATOMIC)))
 		return -ENOMEM;
 
-	verif = kzalloc(ctx->len, GFP_ATOMIC);
-	if (verif == NULL)
-	{
-		kfree(sig_try);
-		return -ENOMEM;
-	}
+	verif = sig_try + ctx->len;
```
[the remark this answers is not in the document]
krizhanovsky: There is slow and even worse `kzalloc()` call. In general this PR doesn't care much about RSA performance, which is the subject for #1335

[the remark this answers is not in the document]
i-rinat: Yep, not really an issue.

tls/pk.c
i-rinat: tl;dr it makes compiler check what otherwise needs to be checked manually

---

Separate header files make code a little bit more robust during future changes.
So for example, if code look like this:
```c
/* func.h */
#ifndef _header_h
#define _header_h

int func(int param1);

#endif
```
```c
/* func.c */
#include "func.h"

int func(int param1) {
  return param1 * param1;
}
```
```c
/* main.c */
#include "func.h"
int main(void) {
  return func(1);
}
```
Then change in any of the files will result in compiler error.
But if `func()` prototype is declared in `main.c`, and `func.h` doesn't exist, it's possible to change `func()` definition to
```c
int func(int param1, int param2) {
  return param1 * param1 + param2 * param2;
}
```
without realizing that `main.c` still calls `func(1)`.

krizhanovsky: Yeah, makes sense, thank you. Let's move back to the header later if we really change the function signatures - 2 lines files look quite weird :)

tempesta_fw/t/unit/test_http_sticky.c
```diff
@@ -1,139 +0,0 @@
...
-
-/* Custom version for testing purposes. */
-void
-tfw_http_resp_build_error(TfwHttpReq *req)
-{
-	(void)req;
-}
-
-int
-test_helper_sticky_start(unsigned int misses)
```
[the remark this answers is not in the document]
krizhanovsky: I left the sticky test removed and fixed the tests in a proper way.

tls/bignum.c
```diff
@@ -377,24 +367,20 @@ ttls_mpi_shift_l(TlsMpi *X, size_t count)
...
-	if ((X->limbs << BSHIFT) < i) {
-		if (!(p = kmalloc(BITS_TO_LIMBS(i) * CIL, GFP_ATOMIC)))
-			return -ENOMEM;
-	}
+	if (WARN_ON_ONCE((X->limbs << BSHIFT) < i))
+		return -ENOSPC;
+
 	X->used = BITS_TO_LIMBS(i);
+	if (old_used < X->used)
+		bzero_fast(p + old_used, (X->used - old_used) * CIL);
```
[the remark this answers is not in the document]
krizhanovsky: The function has be rewriten

tls/ecp_curves.c
```diff
@@ -815,7 +375,156 @@ ecp_mod_p255(TlsMpi *N)
...
+		break;
+	default:
+		T_WARN("Trying to load unsupported curve %d\n", id);
+		return -EINVAL;
+	}
+
+	grp->id = id;
+	ec_groups[id] = grp;
+
+	return 0;
```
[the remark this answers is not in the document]
krizhanovsky: There is no `r` any more in the function

ikoveshnikov:
With Tempesta configuration if RSA key is used (never reproduced with EC keys):
```
listen 80;
listen 443 proto=https;
cache 0;

tls_certificate /home/user/cert/tfw-root.crt;
tls_certificate_key /home/user/cert/tfw-root.key;

srv_group default {
    server 127.0.0.1:8080 conns_n=1;
}

block_action attack reply;

vhost default {
    proxy_pass default;
}

http_chain {
     -> default;
}
```


After running several attempts of tls-perf
```
./tls-perf -T 10  192.168.122.12 443 
```

```
debian login: [193427.134548] [tempesta tls] Warning: None of the common ciphersuites is usable (e.g. no suitable certificate)
[193427.136826] [tempesta fw] Warning: Unrecognized TLS receive return code -22, drop packet
[193527.406298] [tempesta fw] Warning: Unrecognized TLS receive return code -28288, drop packet
[193527.409931] [tempesta fw] Warning: Unrecognized TLS receive return code -28288, drop packet
[193527.413461] ------------[ cut here ]------------
[193527.415771] WARNING: CPU: 1 PID: 16 at /home/user/qtc/release/tempesta/tls/bignum.c:1223 ttls_mpi_exp_mod+0x6b3/0x9e0 [tempesta_tls]
[193527.421587] Modules linked in: tempesta_fw(O) tempesta_db(O) tempesta_tls(O) tempesta_lib(O) sha256_ssse3 sha512_ssse3 sha512_generic ccm kvm_intel binfmt_misc snd_hda_codec_generic iTCO_wdt iTCO_vendor_support kvm snd_hda_intel irqbypass crct10dif_pclmul crc32_pclmul snd_hda_codec ghash_clmulni_intel snd_hda_core snd_hwdep snd_pcm snd_timer snd sg soundcore virtio_console virtio_balloon virtio_gpu ttm pcspkr evdev lpc_ich serio_raw drm_kms_helper mfd_core drm shpchp button ip_tables x_tables autofs4 ext4 crc16 mbcache jbd2 fscrypto ecb aesni_intel crypto_simd cryptd glue_helper aes_x86_64 crc32c_generic sr_mod cdrom virtio_blk virtio_net crc32c_intel psmouse ahci i2c_i801 libahci libata sym53c8xx ehci_pci scsi_transport_spi uhci_hcd ehci_hcd scsi_mod virtio_pci virtio_ring virtio usbcore usb_common
[193527.448064]  [last unloaded: tempesta_lib]
[193527.449526] CPU: 1 PID: 16 Comm: ksoftirqd/1 Tainted: G           O    4.14.0-tempesta-kmemleak-amd64 #1 Debian 4.14.32-tfw7-1
[193527.452856] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS ?-20191223_100556-anatol 04/01/2014
[193527.455129] task: ffff91cb415d3c00 task.stack: ffffa01d006b0000
[193527.456308] RIP: 0010:ttls_mpi_exp_mod+0x6b3/0x9e0 [tempesta_tls]
[193527.457894] RSP: 0018:ffffa01d006b3408 EFLAGS: 00010206
[193527.459354] RAX: 0000000000000000 RBX: ffff91caf5ac1010 RCX: 0000000000000021
[193527.461390] RDX: 0000000000000020 RSI: ffff91caf5ac1018 RDI: ffff91caf5ac1028
[193527.464588] RBP: ffffa01d006b34c8 R08: ffff91caf5ac1040 R09: 0000000000000008
[193527.468112] R10: 0000000000000148 R11: ffff91cb0d740010 R12: ffff91caf5ac1020
[193527.471518] R13: ffffffffffffffff R14: ffff91caf5ac1040 R15: ffff91caf5ac1018
[193527.474295] FS:  0000000000000000(0000) GS:ffff91cb65500000(0000) knlGS:0000000000000000
[193527.476345] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[193527.477465] CR2: 00007fbd3f3c9018 CR3: 000000004ac0a006 CR4: 00000000003606e0
[193527.478678] Call Trace:
[193527.479456]  ? ttls_mpool+0x18/0x110 [tempesta_tls]
[193527.480394]  ? ttls_mpi_sub_int+0x4f/0xc0 [tempesta_tls]
[193527.481289]  ttls_dhm_make_params+0xa0/0x210 [tempesta_tls]
[193527.482261]  ttls_handshake_server_hello+0xd7c/0x1080 [tempesta_tls]
[193527.483772]  ? ttls_update_checksum+0x96/0x160 [tempesta_tls]
[193527.486626]  ttls_handshake_server_step+0x378/0x1a80 [tempesta_tls]
[193527.489689]  ? shash_async_import+0x40/0x40
[193527.490587]  ttls_recv+0x147/0x680 [tempesta_tls]
[193527.491529]  ? ttls_decrypt+0x5d0/0x5d0 [tempesta_tls]
[193527.493000]  ss_skb_process+0x11b/0x140 [tempesta_fw]
[193527.495499]  ? ip_finish_output2+0x19f/0x3b0
[193527.497245]  tfw_tls_msg_process+0xc0/0x4d0 [tempesta_fw]
[193527.499405]  ? ip_output+0x71/0xe0
[193527.501001]  ? ip_queue_xmit+0x5c/0x3b0
[193527.502922]  __gfsm_fsm_exec+0x56/0x90 [tempesta_fw]
[193527.504117]  tfw_connection_recv+0x41/0x60 [tempesta_fw]
[193527.506733]  ? tfw_connection_send+0x30/0x30 [tempesta_fw]
[193527.508318]  ss_tcp_process_data+0x1ea/0x480 [tempesta_fw]
[193527.509647]  ? dev_hard_start_xmit+0x31/0x1f0
[193527.511152]  ss_tcp_data_ready+0x43/0x90 [tempesta_fw]
[193527.512477]  tcp_data_queue+0x4f5/0xc50
[193527.513134]  tcp_rcv_established+0x27c/0x570
[193527.513847]  tcp_v4_do_rcv+0x129/0x1d0
[193527.514474]  tcp_v4_rcv+0x947/0xa50
[193527.515295]  ip_local_deliver_finish+0x9a/0x1c0
[193527.517538]  ip_local_deliver+0x6b/0xe0
[193527.519457]  ? tcp_v4_early_demux+0x112/0x150
[193527.521606]  ? ip_rcv_finish+0x17a/0x400
[193527.523205]  ip_rcv+0x289/0x3c0
[193527.524626]  ? inet_del_offload+0x40/0x40
[193527.525545]  __netif_receive_skb_core+0x84f/0xb30
[193527.526328]  ? lock_timer_base+0x74/0x90
[193527.527340]  ? process_backlog+0xa3/0x160
[193527.528969]  process_backlog+0xa3/0x160
[193527.530352]  net_rx_action+0x28e/0x3f0
[193527.531587]  __do_softirq+0x10f/0x2a8
[193527.532913]  run_ksoftirqd+0x1c/0x40
[193527.533918]  smpboot_thread_fn+0x10e/0x160
[193527.535624]  kthread+0xff/0x130
[193527.536592]  ? sort_range+0x20/0x20
[193527.537851]  ? kthread_create_on_node+0x70/0x70
[193527.539564]  ret_from_fork+0x35/0x40
[193527.540512] Code: 45 b6 66 44 89 4d b0 66 44 89 55 b2 e8 f7 d9 ff ff 48 8b 8d 70 ff ff ff 41 bb 01 00 00 00 66 44 89 5d b0 49 89 cf e9 da fa ff ff <0f> 0b b9 f4 ff ff ff e9 d4 fe ff ff 48 8b 85 78 ff ff ff 48 8b 
[193527.545381] ---[ end trace 00f20e676f79f25b ]---
```

Can't give more details this happend once ocasionally.

tls/bignum.c
```diff
@@ -644,225 +633,155 @@ ttls_mpi_cmp_int(const TlsMpi *X, long z)
...
+	if (B->used > A->used) {
 		const TlsMpi *T = A;
-		A = X;
+		A = B;
 		B = T;
 	}
 
+	X->used = mpi_add_x86_64(MPI_P(X), X->limbs, MPI_P(B), B->used,
+				 MPI_P(A), A->used);
+	BUG_ON(X->used <= 0);
```
ikoveshnikov, line 658: `X->used` is unsigned, so condition is never false. Either need to convert to `int` (as `mpi_add_x86_64()` return value), or use additional variable. Now the error returned by `mpi_add_x86_64()` is ignored.

krizhanovsky:
The call trace https://github.com/tempesta-tech/tempesta/pull/1375#issuecomment-619010725 is quite weird. `tls-perf` produces only ECC handshakes, so `ClientHello` step should fail and it does fail - `Warning: None of the common ciphersuites is usable` is thrown exactly on the step. Meantime the call trace is from the later `ServerHello` step.

The script `while :; do ./tls-perf -T 10 192.168.100.4 443; sleep 1; done` even w/o the `sleep 1` can not reveal the problem.

Having that the problem has appeared on a5c51434030e413ccf3313f46289325c30d88b0f commit, before the last memory leak and corruption fixes, I assume that the problem is in the corrupted memory.

PR #1385, opened 2020-03-01
A few review comment fixes for http2 parser

ikoveshnikov:


tempesta_fw/sock_clnt.c
```diff
@@ -473,10 +473,11 @@ tfw_sock_check_lst(TfwServer *srv)
 static int
```
krizhanovsky: Please update the file copyrights

tempesta_fw/sock_clnt.c
```diff
@@ -507,25 +508,36 @@ tfw_cfgop_listen(TfwCfgSpec *cs, TfwCfgEntry *ce)
 		goto parse_err;
 
 	if (!ce->attr_n)
-		return tfw_listen_sock_add(&addr, TFW_FSM_HTTP);
+		goto done;
 
 	in_str = tfw_cfg_get_attr(ce, "proto", NULL);
 	if (!in_str)
 		goto parse_err;
```
krizhanovsky: `default` parameter is `NULL`, so it seems now we don't have default value for `proto` parameter which is inconsistent with https://github.com/tempesta-tech/tempesta/wiki/Handling-clients#listening-address . I don't see a reason why don't use `h2` as a default one, especially with that `https` and `http` are deprecated, so it seem the code fix is required. Also the Wik says that `proto=http` is the default one, so the Wiki adjustment is also required.

tempesta_fw/http_parser.c
```diff
@@ -6771,271 +6771,231 @@ tfw_h2_parse_req_hdr(unsigned char *data, unsigned long len, TfwHttpReq *req,
```
krizhanovsky: Please update the file copyrights

tempesta_fw/sock_clnt.c
ikoveshnikov: After some discussions in the chat we decided to postpone the change. Currently Tempesta listens for http/1.1 connections over TCP on 0.0.0.0:80 by defaults. This mostly done to simplify jump-start and configuration files for new installations. Setting h2 as default protocol requires TLS section configuration and secret keys creations (and most browsers hates self-signed certificates more than unencrypted http).
So we decided to keep  using http/1.1 over 80 port as default value for now.

PR #1401, opened 2020-04-10
Process multiple 'Cookie:' header in http2 requests

### LINKED ISSUE #1396: Sticky cookies doesn't work over h2
## Scope

In HTTP/1.x `Cookie:` header was singular: it could present in the request only once. If more than one cookie are passed to server, all of them are stored in in the same header. 

In HTTP/2 [it was revised](https://httpwg.org/specs/rfc7540.html#CompressCookie). Non-usual delimiter between values (`;` instead of `,`) cause problems to update only one cookie between requests. 

So both cookie codings are valid:
```
cookie: a=b; c=d; e=f

cookie: a=b
cookie: c=d
cookie: e=f
```

Browsers prefer the second case. And that cause Tempesta to crash on `BUG_ON()` statement. When sticky cookie is enabled, the chain of functions is called:  `tfw_http_sticky_get_req()` -> `tfw_http_msg_clnthdr_val()` -> `__h2_msg_hdr_val()` -> `BUG_ON(TFW_STR_DUP(hdr))`.

The `http_sess` code must be updated to support duplicated `Cookie` headers on h2 protocol.

Steps to reproduce:
- populate several (>>1) permanent cookies for protected uri
- try to reach the uri via Tempesta over h2 protocol
- enjoy the crash

## Testing

Test sticky cookie via h2 protocol, when multiple cookies are send by client.

ikoveshnikov:
Fix #1396. 

A degradation was found in sticky module, introduced in 0937fed631a1b1b954bcd47c52e0873ddff6eaf5 : a redirection mark is always added to 302 redirect even if disabled. Functional tests doesn't validate redirect uri, thus the degradation passed through CI tests. Functional tests are to be updated shortly. Fixed that.

The PR doesn't fix  https://github.com/tempesta-tech/tempesta/issues/1396#issuecomment-611550309 yet.

### DIFF
--- a/tempesta_fw/http_sess.c
+++ b/tempesta_fw/http_sess.c
@@ -683,14 +683,15 @@ tfw_http_sticky_challenge_start(TfwHttpReq *req, TfwStickyCookie *sticky,
 				StickyVal *sv)
 {
 	int r;
-	RedirMarkVal mv = {};
+	RedirMarkVal mv = {}, *mvp = NULL;
 
 	/*
 	 * If configured, ensure that limit for requests without
 	 * cookie and timeout for redirections are not exhausted.
 	 */
 	if (sticky->max_misses) {
-		if ((r = tfw_http_sess_check_redir_mark(req, &mv)))
+		mvp = &mv;
+		if ((r = tfw_http_sess_check_redir_mark(req, mvp)))
 			return r;
 	}
 	if (!sv->ts) {
@@ -698,7 +699,7 @@ tfw_http_sticky_challenge_start(TfwHttpReq *req, TfwStickyCookie *sticky,
 			return TFW_HTTP_SESS_FAILURE;
 	}
 
-	return tfw_http_sticky_build_redirect(req, sv, &mv);
+	return tfw_http_sticky_build_redirect(req, sv, mvp);
 }
 
 /*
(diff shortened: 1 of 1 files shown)


tempesta_fw/http_sess.c
```diff
@@ -318,22 +318,33 @@ search_cookie(TfwStickyCookie *sticky, const TfwStr *cookie, TfwStr *val,
 static int
 tfw_http_sticky_get_req(TfwHttpReq *req, TfwStr *cookie_val)
```
krizhanovsky, line 319: What was the sense to crash on duplicate header in `__h2_msg_hdr_val()`? If the duplicated header doesn't lead to later crash and/or memory corruption, then please replace it with `WARN_ON_ONCE()`.

ikoveshnikov: The `__h2_msg_hdr_val()`, `__http_msg_hdr_val()` and `tfw_http_msg_clnthdr_val()` functions
can be called only against special headers. Since we block messages with duplicated singular special headers at early parsing stage, we assume that singular header can be processed by direct access, while non-singular require `TFW_STR_FOR_EACH_DUP` loops.

We've got a crash here, because `cookie` header is the only exception in h2 and the h2 RFC explicitly allow to duplicate `cookie` header.

I don't sure we really can replace `BUG_ON(TFW_STR_DUP(hdr))` to `WARN_ON()` for every header and string read and write operation.   Either we throw a error and check return code in all such operations, which are very frequent, or we use a kind of save fallback - first duplicate or similar solution. I don't like the later case, since attackers will be able to force us to skip header processing by adding empty header before used one, and processing may differ for duplicated and compounded (united with `,`) headers.

I've checked all the singular spec headers, and we have no issues while working with them, but we have some with nonsingular ones: #1403

PR #1405, opened 2020-05-03
NIST p256 EC performance optimizations

krizhanovsky:
**This pull request isn't ready for code review yet.**

- Improve TLS unit tests with more testing on EC point multiplication
- Benchmark for ECDSA and ECDHE server operations
- Specialized and optimized computation of `m * G`
- Fast optimized non-constant time variant of modular inversion by [Bernstein and Yang](https://tches.iacr.org/index.php/TCHES/article/view/8298/7648)
- Optimized FIPS 186-3 D.2 modular reduction
- Get rid of most MPI wrappers on the hot path


krizhanovsky:
@ikoveshnikov I merged the PR, but please do the review - I'll fix the issues in a new PR

PR #1407, opened 2020-05-12
Adding Content-Length to the request during transforming h2 to h1

avbelov23:
https://github.com/tempesta-tech/tempesta/issues/1409

When transforming h2 to h1, the request may contain a body, so Content-Length must be added for it

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -3300,6 +3300,31 @@ tfw_h2_adjust_req(TfwHttpReq *req)
 			+ req->multipart_boundary_raw.len + SLEN(S_CRLF)
 	};
 	int h_ct_replace = 0;
+	TfwStr h_cl = {0};
+	char cl_data[TFW_ULTOA_BUF_SIZ] = {0};
+	size_t cl_len = 0;
+	/*
+	 * The Transfer-Encoding header field cannot be in the h2 request, because
+	 * requests with Transfer-Encoding are blocked.
+	 */
+	bool need_cl = req->body.len &&
+	               TFW_STR_EMPTY(&ht->tbl[TFW_HTTP_HDR_CONTENT_LENGTH]);
+
+	if (need_cl) {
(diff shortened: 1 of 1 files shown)


tempesta_fw/http.c
```diff
@@ -3475,6 +3491,7 @@ tfw_h2_adjust_req(TfwHttpReq *req)
 	}
 
 	r |= tfw_msg_write(&it, &h_via);
+	r |= tfw_msg_write(&it, &h_cl);
```
ikoveshnikov, line 3494: Unconditional additions actually breaks rules for message framing in http1:

> A user agent SHOULD NOT send a
   Content-Length header field when the request message does not contain
   a payload body and the method semantics do not anticipate such a
   body.

> If a message is received that has multiple Content-Length header
   fields...
   ... recipient MUST either reject the message as invalid or replace the
   duplicated field-values 

Also Content-Length mustn't appear if chunked transfer encoding is used. This is not our use case since chunked encoding is prohibited for H2, but we actually never check that.

To fix that:
- add appropriate condition here
- add check for chunked encoding during request parsing, and reject request or connection if chunked encoding is used.

tempesta_fw/http.c
```diff
@@ -3300,6 +3300,28 @@ tfw_h2_adjust_req(TfwHttpReq *req)
 			+ req->multipart_boundary_raw.len + SLEN(S_CRLF)
 	};
 	int h_ct_replace = 0;
+	TfwStr h_cl = {0};
+	char cl_data[TFW_ULTOA_BUF_SIZ] = {0};
+	size_t cl_len = 0;
+	bool need_cl = req->body.len &&
+	               TFW_STR_EMPTY(&ht->tbl[TFW_HTTP_HDR_CONTENT_LENGTH]) &&
+	               !test_bit(TFW_HTTP_B_CHUNKED, req->flags);
```
ikoveshnikov, line 3308: Nope, that's not right. Сhunked encoding is not allowed to be received on h2 connections, we need to _block_ them during parsing. And here the flag must be always unset.

Issue #1428, opened 2020-07-02
BUG at ttls.c:2243


avbelov23:
## Scope

Caught the BUG at commit `19cf09a`
```
[13221.290092] kernel BUG at /home/avb/Projects/tempesta/tls/ttls.c:2243!
[13221.291223] invalid opcode: 0000 [#1] SMP PTI
[13221.291959] Modules linked in: tempesta_fw(O) tempesta_db(O) tempesta_tls(O) tempesta_lib(O) mcryptd sha256_ssse3 sha512_ssse3 sha512_generic crc32_pclmul crc32c_intel ghash_clmulni_intel pcbc aesni_intel aes_x86_64 crypto_simd glue_helper cryptd vboxvideo(O) ttm vboxguest(O) [last unloaded: tempesta_lib]
[13221.292203] CPU: 0 PID: 7 Comm: ksoftirqd/0 Tainted: G        W  O    4.14.32+ #1
[13221.292203] Hardware name: innotek GmbH VirtualBox/VirtualBox, BIOS VirtualBox 12/01/2006
[13221.292203] task: ffff9157b3555280 task.stack: ffffa7fec18cc000
[13221.292203] RIP: 0010:ttls_close_notify+0x40/0x50 [tempesta_tls]
[13221.292203] RSP: 0018:ffffa7fec18cfd58 EFLAGS: 00010246
[13221.292203] RAX: 0000000000000000 RBX: ffff91571d13cba8 RCX: 0000000000000000
[13221.292203] RDX: 0000000000000001 RSI: 0000000000000000 RDI: ffff91571d13cf18
[13221.292203] RBP: ffff91571d13cf18 R08: 0000000000000000 R09: 00000000010594a1
[13221.292203] R10: ffff9157dfc195f0 R11: ffffffff9f4ad80d R12: 0000000000000000
[13221.292203] R13: ffffffffc030c540 R14: ffff91571d13cba8 R15: 0000000000000000
[13221.292203] FS:  0000000000000000(0000) GS:ffff9157dfc00000(0000) knlGS:0000000000000000
[13221.292203] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[13221.292203] CR2: 00007fbfd802d018 CR3: 000000007a60a001 CR4: 00000000000606f0
[13221.292203] Call Trace:
[13221.292203]  tfw_tls_conn_close+0x25/0x90 [tempesta_fw]
[...]

PR #1429, opened 2020-07-07
Delete keep-alive timer on connection_drop hook

avbelov23:
#1428 

The keep-alive timer may work after https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/tls.c#L582, but before https://github.com/tempesta-tech/tempesta/blob /master/tempesta_fw/tls.c#L584 (where the timer is deleted) and we will catch the bug https://github.com/tempesta-tech/tempesta/blob/master/tls/ttls.c#L2242 (tfw_sock_cli_keepalive_timer_cb () tfw_connection_close () -> tfw_tls_conn_close () -> ttls_close_notify ()), because The tls context will be filled with zeros and tls-> conf == NULL after https://github.com/tempesta-tech/tempesta/blob/master/tempesta_fw/tls.c#L582

### DIFF
--- a/tempesta_fw/connection.h
+++ b/tempesta_fw/connection.h
@@ -147,12 +147,14 @@ typedef struct {
  * @seq_queue	- queue of client's messages in the order they came;
  * @seq_qlock	- lock for accessing @seq_queue;
  * @ret_qlock	- lock for serializing sets of responses;
+ * @timer_lock	- lock for serializing of deleting/modifing keep-alive timer
  */
 typedef struct {
 	TFW_CONN_COMMON;
 	struct list_head	seq_queue;
 	spinlock_t		seq_qlock;
 	spinlock_t		ret_qlock;
+	spinlock_t		timer_lock;
 } TfwCliConn;
 
 /*
--- a/tempesta_fw/connection.h
+++ b/tempesta_fw/connection.h
@@ -342,9 +342,6 @@ tfw_connection_get(TfwConn *conn)
 	atomic_inc(&conn->refcnt);
 }
 
-#define tfw_cli_conn_get(c)	tfw_connection_get((TfwConn *)(c))
-#define tfw_srv_conn_get(c)	tfw_connection_get((TfwConn *)(c))
-
 /**
  * Increment reference counter and return true if @conn is not in
  * failovering process, i.e. @refcnt wasn't less or equal to zero.
@@ -382,7 +379,6 @@ tfw_connection_put(TfwConn *conn)
 		conn->destructor(conn);
 }
 
-#define tfw_cli_conn_put(c)	tfw_connection_put((TfwConn *)(c))
 #define tfw_srv_conn_put(c)	tfw_connection_put((TfwConn *)(c))
 
 static inline void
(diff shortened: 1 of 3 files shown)


tempesta_fw/sock_clnt.c
```diff
@@ -128,9 +126,11 @@ tfw_cli_conn_send(TfwCliConn *cli_conn, TfwMsg *msg)
 	int r;
 
 	r = tfw_connection_send((TfwConn *)cli_conn, msg);
-	mod_timer(&cli_conn->timer,
-		  jiffies +
-		  msecs_to_jiffies((long)tfw_cli_cfg_ka_timeout * 1000));
+
+	if (timer_pending(&cli_conn->timer))
+		mod_timer(&cli_conn->timer,
```
krizhanovsky, line 131: Generally speaking, this check is unreliable: the timer can be deactivated right between the check and `mod_timer()`. Why we ever call `tfw_cli_conn_send()` on dropped connection? There are only 2 places calling the function:

1. ` tfw_http_resp_fwd()` -> `__tfw_http_resp_fwd()` - the former one does `tfw_cli_conn_get()`, so the connection is alive;

2. `tfw_h2_resp_fwd()` - this one seems doesn't care about the client connection state.

The interesting thing is that there are only two places where connection reference counter is incremented: the 1st one in HTTP/1 forwarding and the 2nd one is `tfw_http_conn_msg_alloc()`. Nobody calls `tfw_srv_conn_get()`.

So it seems HTTP/2 is prone for the this and I think other similar referencing a dead connection bugs. Please test HTTP/2 code against #1428 

There are two macros over `tfw_connection_get()`, which are not needed.

tempesta_fw/sock_clnt.c
```diff
@@ -128,9 +127,13 @@ tfw_cli_conn_send(TfwCliConn *cli_conn, TfwMsg *msg)
...
-	mod_timer(&cli_conn->timer,
-		  jiffies +
-		  msecs_to_jiffies((long)tfw_cli_cfg_ka_timeout * 1000));
+
+	spin_lock_bh(&cli_conn->timer_lock);
+	if (timer_pending(&cli_conn->timer))
+		mod_timer(&cli_conn->timer,
+		          jiffies +
+		          msecs_to_jiffies((long)tfw_cli_cfg_ka_timeout * 1000));
+	spin_unlock_bh(&cli_conn->timer_lock);
```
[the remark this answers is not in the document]
ikoveshnikov: Same here. I also tried to find another solution, e.g. for server connections we have a special value `TFW_CONN_DEATHCNT`  in reference counter and state `TFW_CONN_B_DEL` in `TfwSrvConn->flags`. But this doesn't solve unordered `del_timer/mod_timer` calls, which is the problem root.

tempesta_fw/connection.h
```diff
@@ -380,7 +379,10 @@ tfw_connection_put(TfwConn *conn)
 		conn->destructor(conn);
 }
 
-#define tfw_cli_conn_put(c)	tfw_connection_put((TfwConn *)(c))
+/*
+ * Paired with tfw_srv_conn_get_if_live() via tfw_http_get_srv_conn() or
```
ikoveshnikov, line 383: The comment is misplaced. Not the `tfw_srv_conn_put()` function itself is paired with ` tfw_srv_conn_get_if_live()`. Instead commit must be placed on that place where the function is used.

PR #1444, opened 2020-08-19
Restore the response date from the cache

avbelov23:


ikoveshnikov (approved):
Good catch!

ikoveshnikov:
Seems we ignored the test case in functional test. Could you either add it or add a task for the tests?

krizhanovsky (approved):
LGTM, I confirm that the bug is gone.

Please, add an appropriate test in tempesta-tests - this should be a quick task.

Issue #1403, opened 2020-04-20
Possible crashes on accessing nonsingular spec headers


ikoveshnikov:
## Scope

We always access special header directly from headers table `h_tlb` by their ID. But some string functions cant be used against duplicated strings. If the header is  duplicated, it always must be processed in `TFW_STR_FOR_EACH_DUP` loop or crash will happen.

Spotted at:

https://github.com/tempesta-tech/tempesta/blob/3a09b1929a01079a378a6332f5bc6fdf70f481ff/tempesta_fw/http.c#L4845-L4846

There can be other cases, need to check thoroughly.

## Testing

Need to test all possible cases of nonsingular headers in every context they are used. Probably better to do it with fuzzing testing.

PR #1445, opened 2020-08-19
 Use only the first duplicate X-Forwarded-For for client accounting

avbelov23:
#1403

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -4954,15 +4954,19 @@ tfw_http_hm_drop_resp(TfwHttpResp *resp)
 static TfwStr
 tfw_http_get_ip_from_xff(TfwHttpReq *req)
 {
-	TfwStr s_xff, s_ip, *c, *end;
+	const TfwStr *s_xff, *c, *end;
+	TfwStr s_ip;
 	unsigned int nchunks;
 
 	/*
 	 * If a client works through a forward proxy, then a proxy can pass it's
 	 * IP address by the first value in X-Forwarded-For
 	 */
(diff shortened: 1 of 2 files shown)


tempesta_fw/http.c
```diff
@@ -4962,6 +4962,11 @@ tfw_http_get_ip_from_xff(TfwHttpReq *req)
 	 * IP address by the first value in X-Forwarded-For
 	 */
 	s_xff = req->h_tbl->tbl[TFW_HTTP_HDR_X_FORWARDED_FOR];
+	/*
+	 * Get only the first duplicate header
+	 */
```
ikoveshnikov, line 4967:
```suggestion
	 /* Pick only end client address, ignore all proxies between it and us.  */
```
First, don't explain what expressions do, explain why they do that.
Second, it's just a single line comment, don't need to extend it to take three lines.

tempesta_fw/http.c
```diff
@@ -4962,6 +4962,11 @@ tfw_http_get_ip_from_xff(TfwHttpReq *req)
...
+	if (TFW_STR_DUP(&s_xff))
+		s_xff = *__TFW_STR_CH(&s_xff, 0);
```
ikoveshnikov, line 4969: `s_xff` should be a pointer to `TfwStr`, not `TfwStr`. The next function gets `const TfwStr *` as the argument and can't modify underlying data, thus there is no need to build a copy (twice) of XFF header in `s_xff` variable.

ikoveshnikov (approved):
Didn't spot any other mistreating non-singular header as header with no possible duplicates. Good to merge after minor changes.

Issue #1398, opened 2020-03-30
Can't pass redirect mark, when browser restores old session


ikoveshnikov:
## Scope

This issue describes issues with redirect mark (`cookie max_misses` directive) with browsers on session restores.

Tempesta config:
```
sticky {
    cookie name=jdkey enforce max_misses=3 timeout=3;
}

srv_group default {
    server 127.0.0.1:8080;
}

http_chain {
     -> default;
}

```

### Can't reload page in restored session

Browsers may save and restore user sessions on restart. In this case browsers preserve sticky cookie between reboots and also uri redirect.

To reproduce: 
- start Tempesta;
- start browser and open a new session to the protected server with path `/`;
- a sticky cookie challenge will be performed with the redirection mark injection;
- browser gets 302 redirect with sticky cookie and uri as follows: `/__tfw=00000001000000010005cdca7ecbf845cc8fa0151e704f7ae26fd6414a423749/`
- browser follows the redirect with sticky cookie set;
- Tempesta grants access and return content of `/` uri.
- The whole page is displayed correctly, *but the uri in address bar is* `/__tfw=00000001000000010005cdca7ecbf845cc8fa0151e704f7ae26fd6414a423749/`. 
- Exit firefox with session save and reload Tempesta to clear sessions DB;
- Start firefox once agsin, restore the latest session.
- Firefox tries to load the uri `/__tfw=00000001000000010005cdca7ecbf845cc8fa0151e704f7ae26fd6414a423749/` with last sticky cookie.
- Since the cookie is bit outdated and `timeout` is set as 2 seconds, Tempesta treats redirection mark as greatly outdated and blocks client by ip.

Client can't access the site any more. This happen because  cached sticky cookie was too ancient. Seems like directive `sticky { cookie `timeout=3` ) issue. Probably it should be treated as global timeout? Maximum 3 attempts in 3 seconds? Not cookie lifetime?

### Can't reload page in restored session and clean cookies

Same situation as above, but sticky cookie is cleared now after reboot (cookie `jdkey` is missing). Also imagine that we don't block ancient sessions but restart challenge for them (previous part of the issue is resolved).

- browser  tries to load the uri `/__tfw=00000001000000010005cdca7ecbf845cc8fa0151e704f7ae26fd6414a423749/` with no sticky cookie set.
- Tempesta checks the challenge, treats challenge as failed and sends a new redirection mark and cookie.
[...]

PR #1448, opened 2020-08-28
Resetting the counter of redirected requests after a timeout during which the client must block

avbelov23:
#1398 
- counting the timeout as a time interval after which the counter of redirected requests is reset
- adding the ability to set the path directive for cookies
- adding reconfiguration capability for `js_challenge`

### DIFF
--- a/tempesta_fw/http_sess.c
+++ b/tempesta_fw/http_sess.c
@@ -47,6 +47,7 @@
 #include "hash.h"
 #include "http_msg.h"
 #include "http_sess.h"
+#include "http_sess_conf.h"
 #include "vhost.h"
 #include "filter.h"
 #include "tdb.h"
@@ -174,7 +175,8 @@ tfw_http_redir_mark_prepare(RedirMarkVal *mv, char *buf, unsigned int buf_len,
 }
 
 static int
-tfw_http_sticky_build_redirect(TfwHttpReq *req, StickyVal *sv, RedirMarkVal *mv)
+tfw_http_sticky_build_redirect(TfwHttpReq *req, StickyVal *sv, RedirMarkVal *mv,
+                               bool jsch_allow)
 {
 	unsigned long ts_be64 = cpu_to_be64(sv->ts);
 	TfwStr c_chunks[3], m_chunks[3], r_chunks[5], cookie = { 0 }, rmark = { 0 };
@@ -182,7 +184,7 @@ tfw_http_sticky_build_redirect(TfwHttpReq *req, StickyVal *sv, RedirMarkVal *mv)
 	char c_buf[sizeof(*sv) * 2], m_buf[sizeof(*mv) * 2];
 	TfwStr body = { 0 };
 	TfwStickyCookie *sticky;
-	int r;
+	int r, code;
 
 	WARN_ON_ONCE(!list_empty(&req->fwd_list));
 	WARN_ON_ONCE(!list_empty(&req->nip_list));
@@ -205,7 +207,7 @@ tfw_http_sticky_build_redirect(TfwHttpReq *req, StickyVal *sv, RedirMarkVal *mv)
 					    sizeof(m_chunks), &rmark);
 
 	sticky = req->vhost->cookie;
-	if (sticky->js_challenge) {
+	if (sticky->js_challenge && jsch_allow) {
 		if (mv) {
 			if (sticky->js_challenge->body.nchunks != 2)
 				return TFW_HTTP_SESS_FAILURE;
@@ -250,11 +252,10 @@ tfw_http_sticky_build_redirect(TfwHttpReq *req, StickyVal *sv, RedirMarkVal *mv)
 		cookie.nchunks++;
 	}
 
+	code = jsch_allow ? sticky->redirect_code: TFW_REDIR_STATUS_CODE_DFLT;
 	r = TFW_MSG_H2(req)
-		? tfw_h2_prep_redirect(resp, sticky->redirect_code, &rmark,
-				       &cookie, &body)
-		: tfw_h1_prep_redirect(resp, sticky->redirect_code, &rmark,
-				       &cookie, &body);
+		? tfw_h2_prep_redirect(resp, code, &rmark, &cookie, &body)
+		: tfw_h1_prep_redirect(resp, code, &rmark, &cookie, &body);
 	if (r) {
 		tfw_http_msg_free((TfwHttpMsg *)resp);
 		return TFW_HTTP_SESS_FAILURE;
@@ -719,7 +720,7 @@ tfw_http_sticky_challenge_start(TfwHttpReq *req, TfwStickyCookie *sticky,
 			return TFW_HTTP_SESS_FAILURE;
 	}
 
-	return tfw_http_sticky_build_redirect(req, sv, mvp);
+	return tfw_http_sticky_build_redirect(req, sv, mvp, true);
 }
 
 /*
@@ -1145,6 +1146,7 @@ tfw_http_sess_obtain(TfwHttpReq *req)
 	TfwStr *c_val = &ctx.cookie_val;
 	TdbRec *rec;
 	TdbGetAllocCtx tdb_ctx = { 0 };
+	TfwStr mark_val = {};
 
 	/*
 	 * If vhost is not known, then request is to be dropped. Don't save the
@@ -1212,6 +1214,9 @@ tfw_http_sess_obtain(TfwHttpReq *req)
 		 * bucket with the session as soon as possible.
 		 */
 		tdb_rec_put(rec);
+
+	if (tfw_http_redir_mark_get(req, &mark_val))
+		return tfw_http_sticky_build_redirect(req, sv, NULL, false);
 
 	return TFW_HTTP_SESS_SUCCESS;
 }
--- a/tempesta_fw/http_sess.c
+++ b/tempesta_fw/http_sess.c
@@ -177,10 +177,10 @@ static int
 tfw_http_sticky_build_redirect(TfwHttpReq *req, StickyVal *sv, RedirMarkVal *mv)
 {
 	unsigned long ts_be64 = cpu_to_be64(sv->ts);
-	TfwStr c_chunks[3], m_chunks[3], cookie = { 0 }, rmark = { 0 };
+	TfwStr c_chunks[3], m_chunks[3], r_chunks[5], cookie = { 0 }, rmark = { 0 };
 	TfwHttpResp *resp;
 	char c_buf[sizeof(*sv) * 2], m_buf[sizeof(*mv) * 2];
-	TfwStr *body = NULL;
+	TfwStr body = { 0 };
 	TfwStickyCookie *sticky;
 	int r;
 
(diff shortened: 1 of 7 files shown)


tempesta_fw/http_sess.c
```diff
@@ -659,10 +659,12 @@ tfw_http_sess_check_redir_mark(TfwHttpReq *req, RedirMarkVal *mv)
 	unsigned long tmt = HZ * (unsigned long)req->vhost->cookie->tmt_sec;
 
 	if (tfw_http_redir_mark_get(req, &mark_val)) {
-		if (tfw_http_redir_mark_verify(req, &mark_val, mv)
-		    || ++mv->att_no > max_misses
-		    || (tmt && mv->ts + tmt < jiffies))
-		{
+		int r = tfw_http_redir_mark_verify(req, &mark_val, mv);
+
+		if (tmt && mv->ts + tmt < jiffies)
```
ikoveshnikov, line 664: Possible access out of variable borders. `tfw_http_redir_mark_verify()` can fail because of its size. Rmark can be smaller than `mv->ts`.

tempesta_fw/http_sess.c
```diff
@@ -659,10 +659,12 @@ tfw_http_sess_check_redir_mark(TfwHttpReq *req, RedirMarkVal *mv)
...
+			mv->att_no = 0;
+
+		if (r || ++mv->att_no > max_misses) {
```
ikoveshnikov, line 667: This works incorrect. If a cookie was taken too long ago and rmark is already expired at this moment, we always treat rmark as valid, no matter, how many attempts was done. Instead we should restart the challenge.

tempesta_fw/http_sess_conf.c
```diff
@@ -159,12 +159,30 @@ tfw_cfgop_cookie_set_options(TfwStickyCookie *sticky, const char *options)
...
+{
+	size_t len = strlen(path);
+	if (len + 2 > STICKY_OPT_MAXLEN) {
+		T_ERR_NL("http_sess: too long cookie path length.\n");
+		return -EINVAL;
+	}
+
+	sticky->options_str[sticky->options.len + 0] = ';';
+	sticky->options_str[sticky->options.len + 1] = ' ';
+	memcpy(&sticky->options_str[sticky->options.len + 2], path, len);
```
ikoveshnikov, line 183: The path is copied into buffer here. But only attribute value, starting `path=` string is missing.

tempesta_fw/http_sess_conf.c
```diff
@@ -211,13 +232,29 @@ tfw_cfgop_cookie_set(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
 			return -EINVAL;
 		}
 	}
 
+	if (!was_path) {
+		sticky->options_str[sticky->options.len + 0] = ';';
+		sticky->options_str[sticky->options.len + 1] = ' ';
+		memcpy(&sticky->options_str[sticky->options.len + 2], "path=/", 6);
+
+		sticky->options.len += 6 + 2;
```
ikoveshnikov, line 255: Please, no magic numbers, use `SLEN` macro.

ikoveshnikov (changes requested):

tempesta_fw/http_sess_conf.c
```diff
@@ -211,13 +235,31 @@ tfw_cfgop_cookie_set(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
 				 cs->name, key, val);
 			return -EINVAL;
 		}
 	}
 
+	if (!was_path) {
+		static const char *dflt_path = "path=/";
+		sticky->options_str[sticky->options.len + 0] = ';';
+		sticky->options_str[sticky->options.len + 1] = ' ';
+		memcpy(&sticky->options_str[sticky->options.len + 2], dflt_path,
```
ikoveshnikov, line 257: Why not combine dilimeters and the attribute name? `dflt_path = "; path=/";`. A single `memcpy` will handle that.

tempesta_fw/http_sess_conf.c
```diff
@@ -159,12 +159,33 @@ tfw_cfgop_cookie_set_options(TfwStickyCookie *sticky, const char *options)
...
+
+	return 0;
+}
+
+int
+tfw_cfgop_cookie_set_path(TfwStickyCookie *sticky, const char *path_val)
+{
+	static const char *path = "path=";
+	size_t path_val_len = strlen(path_val);
+	if (path_val_len + 2 > STICKY_OPT_MAXLEN) {
```
ikoveshnikov, line 177: Before the patch `tfw_cfgop_cookie_set_options()` was the only place where `sticky->options_str` was modified. The options was a string with both value names and values, thus `strlen(options) + 2 > STICKY_OPT_MAXLEN` worked fine. (The same argument is not allowed to be listed multiple times, thanks to `cfg.c`).

But now the string is modified in three places: when parsing `option`, when parsing `path` and on default `path` value. Only first two have free space check for the buffer, and all of them don't account data already pushed to buffer.

ikoveshnikov (approved):
Looks good to me.

krizhanovsky (commented):
I tested my demo scenario and now it works fine, so I'm good with the changes.

However, there is some mess with redirection mark, which we send in Location header:
1. it seems it's just ignored at the moment
2. it seems we can not track attempts any more - we have max_misses in the sticky cookies, why not to use it?

Probably, we can/should just remove the rmark logic as a dead code. Also https://github.com/tempesta-tech/tempesta/wiki/Sticky-Cookie#javascript-challenge doesn't mention Location logic at all, and, if we leave the logic, the Wiki must be updated.

ikoveshnikov:
About redirection mark:

> it seems it's just ignored at the moment

It is ignored *only* when JSCH is enabled. Just like in your scenario. This happens because JSCH is sent in 503 response, while Location header is meaningful only for 3xx and 201 responses

> it seems we can not track attempts any more

Only attempts for basic cookie challenge can be accounted now. Attempts for JSCH are not accounted, that's true.

>  rmark logic as a dead code

No, it perfectly works for the basic cookie challenge.

krizhanovsky (changes requested):
The minutes from the discussion.

We add a template variable into js_challenge.js.tpl and remember its position via TfwStr chunks (there is no need to actually separate a contiguous memory). The redirect construction should rewrite the variable body with the value of the sticky cookie plus the attempt number, so JS can decode the value instead current Cookie. This way JSCH should not depend on cookies at all.

With the variable rewriting we should copy the response body and mark it for in-place encryption at the TLS layer, so we won't pay extra for the copying at HTTP sessions layer.

Please add a corresponding comment for the template variable and do not use any spaces to allow to use JS minifiers.

>>    rmark logic as a dead code

> No, it perfectly works for the basic cookie challenge.

This is the documentation Wiki: please update https://github.com/tempesta-tech/tempesta/wiki/Sticky-Cookie with what is really going on, that the cookie challenge actually can work without cookies just using rmarks.

With this changes in Wiki, we probably don't need to move js_challenge out of the sticky configuration.

krizhanovsky:
I noticed `__tfw` token in URI after redirection by JSCH. It seems we shouldn't see it with the last modifications
Screenshot_2020-09-10_23-41-15

krizhanovsky:
Another usability issue: if a browser doesn't support JavaScript, then we just show the message
```
 Tempesta FW is verifying your browser, please wait a little bit...
```
Please add a line like "Please check that you browser supports JavaScript"

tempesta_fw/http_sess.c
```diff
@@ -207,6 +203,14 @@ tfw_http_sticky_build_redirect(TfwHttpReq *req, StickyVal *sv, RedirMarkVal *mv)
 	if (mv)
 		tfw_http_redir_mark_prepare(mv, m_buf, sizeof(m_buf), m_chunks,
 					    sizeof(m_chunks), &rmark);
+
+	sticky = req->vhost->cookie;
+	if (sticky->js_challenge) {
+		body = &sticky->js_challenge->body;
+		if (mv && tfw_strcpy(&sticky->js_challenge->rmark, &rmark))
```
krizhanovsky, line 210: If many CPUs run the code, then they mangle the `rmark` by rewriting each other. We can not do copy here as I proposed in https://github.com/tempesta-tech/tempesta/pull/1448#pullrequestreview-483017256 because `tfw_h1_prep_redirect()` and `tfw_h2_prep_redirect()` -> `tfw_h2_append_predefined_body()` copy the body: we don't want extra copies. However, the functions accept `body` as `TfwStr`, so it seems we should do little bit more work here with `TfwCfgJsCh->body`: `body` should be allocated within `TfwCfgJsCh` as 3-chunks `TfwStr` (e.g. `rmark` becomes the middle chunk). This way we can just copy the 2-chunks TfwStr in each `tfw_http_sticky_build_redirect()` call (only `TfwStr` descriptors with different middle chunk pointers!) and pass the string descriptors for copying to the HTTP functions constructing the redirects.

krizhanovsky (changes requested):
It seems there is a concurrency issue as well as not all #1398 issues were fixed (the `__tfw` redirection mark persists in the browser address bar).

tempesta_fw/http_sess.c
```diff
@@ -207,6 +206,28 @@ tfw_http_sticky_build_redirect(TfwHttpReq *req, StickyVal *sv, RedirMarkVal *mv)
 	if (mv)
 		tfw_http_redir_mark_prepare(mv, m_buf, sizeof(m_buf), m_chunks,
 					    sizeof(m_chunks), &rmark);
+
+	sticky = req->vhost->cookie;
+	if (sticky->js_challenge && jsch_allow) {
+		if (mv) {
+			body = tfw_strdup_desc(req->pool, &sticky->js_challenge->body);
```
krizhanovsky, line 213: It seems here and at the below are extra memory allocations, which can be replaced with automatic variables

tempesta_fw/http_sess.c
```diff
@@ -235,10 +256,12 @@ tfw_http_sticky_build_redirect(TfwHttpReq *req, StickyVal *sv, RedirMarkVal *mv)
...
 	r = TFW_MSG_H2(req)
-		? tfw_h2_prep_redirect(resp, sticky->redirect_code, &rmark,
-				       &cookie, body)
-		: tfw_h1_prep_redirect(resp, sticky->redirect_code, &rmark,
-				       &cookie, body);
+		? tfw_h2_prep_redirect(resp, jsch_allow ? sticky->redirect_code:
+		                                          tfw_cfg_redirect_st_code_dflt,
+		                       &rmark, &cookie, body)
+		: tfw_h1_prep_redirect(resp, jsch_allow ? sticky->redirect_code:
+		                                          tfw_cfg_redirect_st_code_dflt,
```
krizhanovsky, line 263: Isn't it better to use some automatic variable instead of the `?:` repetition?

tempesta_fw/http_sess_conf.h
```diff
@@ -24,6 +24,8 @@
 #include "cfg.h"
 #include "http_types.h"
 
+extern const unsigned short tfw_cfg_redirect_st_code_dflt;
+
```
krizhanovsky, line 28: Maybe just use #define or `static const short` instead of exporting the single integer?

krizhanovsky (approved):
OK, looks good now: the demo scenario passes smoothly.

Please fix the minor comments and merge it.

Issue #1441, opened 2020-08-11
block_action reply doesn't work as expected with frang


krizhanovsky:
## Scope

Tempesta config:
```
listen 192.168.100.4:443 proto=https;
listen 192.168.100.4:80;

frang_limits {
        # Conenction level limits.
        client_header_timeout 20;
        client_body_timeout 10;
        http_header_chunk_cnt 10;
        http_body_chunk_cnt 0;

        # Message level limits (per-vhost)
        http_methods GET;
        http_uri_len 512;
        http_resp_code_block 400 403 404 3 10;
}

http_uri_brange 0x2f 0x41-0x5a 0x61-0x7a 0x30-0x39 0x2d 0x2e 0x5f; # /a-zA-Z0-9-._

block_action attack reply;
block_action error reply;

srv_group default {
        server 127.0.0.1:8080 conns_n=4;
}

vhost default {
        tls_certificate /root/tempesta/etc/tfw-root.crt;
        tls_certificate_key /root/tempesta/etc/tfw-root.key;

        resp_hdr_set Strict-Transport-Security "max-age=31536000; includeSubDomains";

        proxy_pass default;
}

cache 0;

http_chain {
        -> default;
}
```

If I send a request with invalid character, say `*`, then I receive `400 Bad Request` response as expected. However, if I reach `http_resp_code_block` limit, then Tempesta resets the connection.

## Testing

Please add the test case to functional tests, maybe as part of #673

PR #1449, opened 2020-09-11
Handling the return code of tfw_gfsm_move()

avbelov23:
#1441

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -5603,7 +5603,12 @@ tfw_http_resp_cache(TfwHttpMsg *hmresp)
 	data.skb = NULL;
 	data.req = (TfwMsg *)req;
 	data.resp = (TfwMsg *)hmresp;
-	tfw_gfsm_move(&hmresp->conn->state, TFW_HTTP_FSM_RESP_MSG_FWD, &data);
+	if (tfw_gfsm_move(&hmresp->conn->state, TFW_HTTP_FSM_RESP_MSG_FWD, &data)) {
+		tfw_http_conn_msg_free(hmresp);
+		tfw_http_req_block(req, 403, "response blocked: filtered out");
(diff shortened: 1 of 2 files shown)


tempesta_fw/http.c
```diff
@@ -5603,7 +5603,12 @@ tfw_http_resp_cache(TfwHttpMsg *hmresp)
 	data.skb = NULL;
 	data.req = (TfwMsg *)req;
 	data.resp = (TfwMsg *)hmresp;
-	tfw_gfsm_move(&hmresp->conn->state, TFW_HTTP_FSM_RESP_MSG_FWD, &data);
+	if(tfw_gfsm_move(&hmresp->conn->state, TFW_HTTP_FSM_RESP_MSG_FWD, &data)) {
+		tfw_http_conn_msg_free(hmresp);
+		tfw_http_req_block(req, 403, "response blocked: filtered out");
```
krizhanovsky, line 5608: I see that there are other places with `tfw_http_conn_msg_free()` followed by `tfw_http_req_block()`, but I believe we don't need to call `tfw_http_conn_msg_free()` if request is paired with the response. The other similar places are: `tfw_http_resp_gfsm()`, `tfw_http_resp_terminate()`, and seems `tfw_http_resp_process()`.

Nothing really bad happens since `tfw_http_msg_free()` calls `tfw_http_msg_unpair()`, so there is no double freeing. But, `tfw_http_cli_error_resp_and_log()` frees the response as `req->pair` and there is just no reason to call the extra function and conditional statements inside it.

Please check my assumption and, it it's correct (**make an instrumentation patch insuring that the responses are really freed and there is no memory leak**), replace the `tfw_http_conn_msg_free()` calls preceding  `tfw_http_req_block()` with a comment like `/* The response is freed by tfw_http_req_block(). */`

krizhanovsky (approved):
I confirm that the problem is gone. However, there is possible code improvement which must be done before the merge.

Issue #1404, opened 2020-04-22
BUG at http_limits.c:406


krizhanovsky:
## Scope

Just caught the BUG at current #1375 :
```
[  945.117448] kernel BUG at /root/tempesta/tempesta_fw/http_limits.c:406! 
[  945.120362] invalid opcode: 0000 [#1] SMP PTI
[  945.122355] Modules linked in: tempesta_fw(O) tempesta_db(O) tempesta_tls(O) tempesta_lib(O) binfmt_misc bochs_drm ttm drm_kms_helper drm crct10dif_pclmul fb_sys_fops syscopyarea ppdev sysfillrect sysimgblt parport_pc sg serio_raw parport button ip_tables x_tables ext4 crc16 mbcache jbd2 fscrypto sr_mod cdrom sd_mod ata_generic ata_piix libata psmouse i2c_piix4 e1000 scsi_mod [last unloaded: tempesta_lib]
[  945.133647] CPU: 1 PID: 0 Comm: swapper/1 Tainted: G           O    4.14.32-kdump+ #152
[  945.136338] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS ?-20190727_073836-buildvm-ppc64le-16.ppc.fedoraproject.org-3.fc31 04/01/2014
[  945.140970] task: ffff980e1d3f50c0 task.stack: ffff980e1c854000 
[  945.143251] RIP: 0010:frang_conn_close+0x31/0x40 [tempesta_fw]
[  945.145477] RSP: 0018:ffff980e3fd03e78 EFLAGS: 00010246 
[  945.148106] RAX: 0000000000000000 RBX: ffff980e3a405048 RCX: 000000000000091b
[  945.152067] RDX: 0000000000000001 RSI: 0000000000010080 RDI: ffff980e3a40504c
[  945.156196] RBP: ffff980e1c5f0dc0 R08: 00000000000027ec R09: 00000000000cce66
[  945.168760] R10: 0000000000000000 R11: 0000000000016346 R12: 0000000000000000
[  945.173241] R13: ffffffffae176710 R14: ffff980e1c5f0dc0 R15: 0000000000000001
[  945.178035] FS:  0000000000000000(0000) GS:ffff980e3fd00000(0000) knlGS:0000000000000000
[  945.187984] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[...]

PR #1456, opened 2020-10-14
frang: during stop procedure, wait until all in-flight sockets are done

i-rinat:
It was possible to stop Tempesta while some sockets were still processed by the kernel network code. If Tempesta was started right after it was stopped, as it happens in restart procedure, a socket from a previous run may be freed during a current run. Rather than changing code logic to handle such ghosts from the past, let's wait in stop procedure until all related sockets are done.

### DIFF
--- a/tempesta_fw/http_limits.c
+++ b/tempesta_fw/http_limits.c
@@ -53,6 +53,8 @@ static struct {
 
 static TfwClassifier __rcu *classifier = NULL;
 
+static atomic64_t inflight_socket_count = ATOMIC64_INIT(0);
+
 /**
  * Shrink client connections hash and/or reduce QoS for blocked clients to
  * lower back-end servers or local system load.
@@ -127,6 +129,8 @@ tfw_classify_conn_estab(struct sock *sk)
 	return TFW_PASS;
 
 ours:
+	atomic64_inc(&inflight_socket_count);
+
 	rcu_read_lock();
 
 	clfr = rcu_dereference(classifier);
@@ -146,6 +150,8 @@ tfw_classify_conn_close(struct sock *sk)
 
 	if (clfr && clfr->classify_conn_close)
 		clfr->classify_conn_close(sk);
+
+	atomic64_dec(&inflight_socket_count);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ikoveshnikov:
The PR is intended to close the issue #1404 . A wait cycle is added to assure that all the client connections are closed during close.

At the first sight we already has the cycle that does the same: а `tfw_sock_clnt_stop()` spins until all client connections will receive `SS_close()` signal. The resource management relies on preemption here: `stop()` procedure is done in a process context while actual connection close procedures are performed in SoftIRQ. It looks like a week synchronisation and there might be a possibility that some sockets may be closed during `stop()` operations of further  modules. 

Since `TfwClient` entries may live longer than their connections (to keep Frang accounting data between client reconnects), there is no direct way to assure, that all the client connections were closed. From this perspective an atomic counter of client connection is the only way to assure that all the connections were closed before stop. This is what the PR does.

Happily we already has this counter implemented, the function called `ss_synchronize()` and called on stop. It  assures that all the open connections used by Tempesta was closed. Some objects can leak there though. But it's a kind of eternal loop protection: if none of the connections was closed during last 5 seconds, then there is an issue in reference counters and there is no sense to wait for release. Thus I don't think we need a new counter in `http_limits.c`.


_By the way, `tfw_client_for_each()` can cause a scheduler stall message in kernel log. Earlier `TfwClients` were organised as a list and looping through all the clients  involved a `schedule()` call like all other long loops in process context. So it was safe to dive into a long  `tfw_client_for_each()` loop. But after TDB become the storage for `TfwClient` instances, a new procedure `tdb_entry_walk()` was introduced, that can grab the scheduler for a long time while interrupts are disabled._


Rinat also described in chat a possible appearance of fantom connections left from the previous run:
- Tempesta is started
- A `client0.tdb` file is created and filled with zeros
- An ingress connection increments counter of current active connections in client entry in the `client0.tdb` table
- Tempesta restart is requested
- An `kernel_write()` call fails during `client0.tdb` dumping on disk
- Tempesta is started
- The `client0.tdb` is read from disk, current connection counter there - `0`
- A socket for the ingress connection (established few steps earlier) is destroyed, a hook for `sk_free` is called.
- Finally the `frang_conn_close()` is called, where the `BUG_ON(!ra->conn_curr)` is triggered. 
`kernel_write()` are rare but possible. So does connections created on the previous run. Both conditions required to trigger the issue.

krizhanovsky:
The problem is with sockets on the TCP layer, below the Tempesta FW layer https://github.com/tempesta-tech/tempesta/issues/1404#issuecomment-830868534

PR #1457, opened 2020-10-17
Fix framing on long msg

ikoveshnikov:
Although #1400 already fixed the issue, a bug remained hidden there thus client received 500 response instead of huge files over H2 connections. This happened because after H2 frame header is inserted, the message iterator and index pointer in whole represent invalid pointer to skb data. Now the pointer is fixed before iterator move and huge responsed are served as normal.

The second issue fixed in the PR is header modifications of http/1 responses served from cache. Detailed description is provided in commit messages.

### DIFF
--- a/tempesta_fw/hpack.c
+++ b/tempesta_fw/hpack.c
@@ -1834,7 +1834,7 @@ check_name_text:
 			/* All duplicate headers must be skipped by caller. */
 			WARN_ON_ONCE(test_bit(i, mit->found));
 			__set_bit(i, mit->found);
-			if (!TFW_STR_CHUNK(dc_iter->desc->hdr, 2)) {
+			if (!TFW_STR_CHUNK(dc_iter->desc->hdr, 1)) {
 				dc_iter->skip = true;
 				goto out;
 			}
@@ -1915,10 +1915,10 @@ get_value_text:
 			TfwStr n_val = {
 				.chunks = (TfwStr []){
 					{ .data = ", ", .len = 2 },
-					{ .data = __TFW_STR_CH(h, 2)->data,
-					  .len = __TFW_STR_CH(h, 2)->len }
+					{ .data = __TFW_STR_CH(h, 1)->data,
+					  .len = __TFW_STR_CH(h, 1)->len }
 				},
-				.len = __TFW_STR_CH(h, 2)->len + 2,
+				.len = __TFW_STR_CH(h, 1)->len + 2,
 				.nchunks = 2
 			};
 
--- a/tempesta_fw/hpack.c
+++ b/tempesta_fw/hpack.c
@@ -1,7 +1,7 @@
 /**
  *		Tempesta FW
  *
- * Copyright (C) 2019 Tempesta Technologies, Inc.
+ * Copyright (C) 2019-2020 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
@@ -1834,6 +1834,12 @@ check_name_text:
 			/* All duplicate headers must be skipped by caller. */
 			WARN_ON_ONCE(test_bit(i, mit->found));
 			__set_bit(i, mit->found);
+			/*
+			 * Header modifications format: 0 chunk - header name,
+			 * optional 1st chunk - header value. If the value is
+			 * empty, then the header is about to be removed,
+			 * don't write it.
+			 */
 			if (!TFW_STR_CHUNK(dc_iter->desc->hdr, 1)) {
 				dc_iter->skip = true;
 				goto out;
@@ -1912,6 +1918,10 @@ get_value_text:
 
 		if (dc_iter->desc) {
 			TfwStr *val, *h = dc_iter->desc->hdr;
+			/*
+			 * Header value is stored in chunk 1, see
+			 * tfw_cfgop_mod_hdr_add().
+			 */
 			TfwStr n_val = {
 				.chunks = (TfwStr []){
 					{ .data = ", ", .len = 2 },
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -1448,6 +1448,7 @@ tfw_http_msg_insert(TfwMsgIter *it, char **off, const TfwStr *data)
 	}
 
 	*off = dst.data;
+	it->skb = dst.skb;
 
 	return tfw_strcpy(&dst, data);
 }
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -1462,17 +1462,30 @@ this_chunk:
 /**
  * Insert data from string @data to message at offset defined by message
  * iterator @it and @off. This function doesn't maintain message structure.
- * After insertion message iterator will point at the end of fragment right
- * before the inserted data.
+ * After insertion message iterator and @data will point at the start of
+ * inserted data fragment.
  */
 int
-tfw_http_msg_insert(TfwMsgIter *it, char *off, const TfwStr *data)
+tfw_http_msg_insert(TfwMsgIter *it, char **off, const TfwStr *data)
 {
 	int r;
 	TfwStr dst = {};
 
-	if ((r = ss_skb_get_room(it->skb_head, it->skb, off, data->len, &dst)))
+	if ((r = ss_skb_get_room(it->skb_head, it->skb, *off, data->len, &dst)))
 		return r;
+
+	/*
+	 * @dst is always a plain string. If the @data wasn't inserted into
+	 * current skb fragment, then it was appended to a previous one.
+	 * Insertions into linear part doesn't require iterator updates.
+	 */
+	*off = dst.data;
+	if (it->frag >= 0) {
+		skb_frag_t *f = &skb_shinfo(it->skb)->frags[it->frag];
+		long offset = dst.data - (char *)skb_frag_address(f);
+		if (offset < 0 || offset > skb_frag_size(f))
+			it->frag--;
+	}
 
 	return tfw_strcpy(&dst, data);
 }
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -1443,7 +1443,9 @@ tfw_http_msg_insert(TfwMsgIter *it, char **off, const TfwStr *data)
 	int r;
 	TfwStr dst = {};
 
-	if ((r = ss_skb_get_room_w_frag(it->skb_head, it->skb, *off, data->len, &dst, &it->frag))) {
+	if ((r = ss_skb_get_room_w_frag(it->skb_head, it->skb, *off, data->len,
+					&dst, &it->frag)))
+	{
 		return r;
 	}
 
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -952,34 +952,6 @@ tfw_http_msg_del_hbh_hdrs(TfwHttpMsg *hm)
 	return 0;
 }
 
-int
-tfw_http_msg_del_eol(struct sk_buff *skb_head, TfwStr *hdr)
-{
-	int r;
-	TfwStr it = {};
-	int eolen = tfw_str_eolen(hdr);
-	TfwStr *last = TFW_STR_LAST(hdr);
-
-	if ((r = skb_next_data(last->skb, last->data + last->len - 1, &it)))
-		return r;
-
-	while (eolen) {
-		char *ptr = it.data;
-		struct sk_buff *skb = it.skb;
-
-		bzero_fast(&it, sizeof(TfwStr));
-		if ((r = skb_fragment(skb_head, skb, ptr, -eolen, &it)))
-			return r;
-
-		if (WARN_ON_ONCE(r > eolen))
-			return -EINVAL;
-
-		eolen -= r;
-	}
-
-	return 0;
-}
-
 /**
  * Modify message body, add string @data to the end of the body (if @append) or
  * to its beginning.
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -1443,21 +1443,11 @@ tfw_http_msg_insert(TfwMsgIter *it, char **off, const TfwStr *data)
 	int r;
 	TfwStr dst = {};
 
-	if ((r = ss_skb_get_room(it->skb_head, it->skb, *off, data->len, &dst)))
+	if ((r = ss_skb_get_room_w_frag(it->skb_head, it->skb, *off, data->len, &dst, &it->frag))) {
 		return r;
-
-	/*
-	 * @dst is always a plain string. If the @data wasn't inserted into
-	 * current skb fragment, then it was appended to a previous one.
-	 * Insertions into linear part doesn't require iterator updates.
-	 */
-	*off = dst.data;
-	if (it->frag >= 0) {
-		skb_frag_t *f = &skb_shinfo(it->skb)->frags[it->frag];
-		long offset = dst.data - (char *)skb_frag_address(f);
-		if (offset < 0 || offset > skb_frag_size(f))
-			it->frag--;
 	}
+
+	*off = dst.data;
 
 	return tfw_strcpy(&dst, data);
 }
--- a/tempesta_fw/msg.c
+++ b/tempesta_fw/msg.c
@@ -132,10 +132,18 @@ tfw_msg_iter_move(TfwMsgIter *it, unsigned char **data, unsigned long sz)
 
 		/* Linear skb part. */
 		if (it->frag < 0) {
+			unsigned char *begin = it->skb->data;
+			unsigned char *end = begin + skb_headlen(it->skb);
+			if (unlikely(addr < begin || addr > end))
+				addr = begin;
 			f_sz_rem = skb_headlen(it->skb) + it->skb->data - addr;
 		}
 		else {
 			skb_frag_t *f = &skb_shinfo(it->skb)->frags[it->frag];
+			unsigned char *begin = skb_frag_address(f);
+			unsigned char *end = begin + skb_frag_size(f);
+			if (unlikely(addr < begin || addr > end))
+				addr = begin;
 			f_sz_rem = skb_frag_size(f) +
 				(unsigned char *)skb_frag_address(f) - addr;
 		}
--- a/tempesta_fw/msg.c
+++ b/tempesta_fw/msg.c
@@ -132,18 +132,10 @@ tfw_msg_iter_move(TfwMsgIter *it, unsigned char **data, unsigned long sz)
 
 		/* Linear skb part. */
 		if (it->frag < 0) {
-			unsigned char *begin = it->skb->data;
-			unsigned char *end = begin + skb_headlen(it->skb);
-			if (unlikely(addr < begin || addr > end))
-				addr = begin;
 			f_sz_rem = skb_headlen(it->skb) + it->skb->data - addr;
 		}
 		else {
 			skb_frag_t *f = &skb_shinfo(it->skb)->frags[it->frag];
-			unsigned char *begin = skb_frag_address(f);
-			unsigned char *end = begin + skb_frag_size(f);
-			if (unlikely(addr < begin || addr > end))
-				addr = begin;
 			f_sz_rem = skb_frag_size(f) +
 				(unsigned char *)skb_frag_address(f) - addr;
 		}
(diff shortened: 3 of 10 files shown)


tempesta_fw/msg.c
```diff
@@ -132,10 +132,18 @@ tfw_msg_iter_move(TfwMsgIter *it, unsigned char **data, unsigned long sz)
 
 		/* Linear skb part. */
 		if (it->frag < 0) {
+			unsigned char *begin = it->skb->data;
+			unsigned char *end = begin + skb_headlen(it->skb);
+			if (unlikely(addr < begin || addr > end))
+				addr = begin;
```
keshonok, line 138: Frankly, with the introduction of new variable `begin` the unchanged statement below now looks somewhat strange. It would be more logical if it looked like this now:
```
f_sz_rem = skb_headlen(it->skb) + begin - addr;
```

Same is true for a similar statement in the `else` branch.

tempesta_fw/hpack.c
```diff
@@ -1834,7 +1834,7 @@ do {									\
 			/* All duplicate headers must be skipped by caller. */
 			WARN_ON_ONCE(test_bit(i, mit->found));
 			__set_bit(i, mit->found);
-			if (!TFW_STR_CHUNK(dc_iter->desc->hdr, 2)) {
+			if (!TFW_STR_CHUNK(dc_iter->desc->hdr, 1)) {
```
krizhanovsky, line 1837: There is comment `All duplicate headers must be skipped by caller.` at the above - please update it with the words why do we check the 2nd chunk, i.e. why strings with only one chunks are treated as duplicate headers.

krizhanovsky (changes requested):
Good fixes, but some h2 code improvements are required.

tempesta_fw/hpack.c
ikoveshnikov: The comment about duplicates is applied only for two following lines, while `if (!TFW_STR_CHUNK(dc_iter->desc->hdr, 1))...` has nothing to do about duplicates. I agree, that is looks misleading and messed up, so I've added a new comment here, describing chunk indexes.

tempesta_fw/msg.c
ikoveshnikov: Thank you both, guys, this was dirty fixed. I have updated the `tfw_http_msg_insert()` function, so now it updates data pointer and following `tfw_msg_iter_move()` call gives correct iterator offset (not yet uploaded). But...

But I have spot a bug in `tfw_http_msg_insert()`, which is very simple wrapper around `ss_skb_get_room()`. The last function is called in many places. Look: `tfw_http_msg_insert()` inserts a 9-byte data piece at `TfwMsgIter {skb=ffff88977d9bd000, frag=8}`, at offset `ffff88976d004000`.

This the skb before insertion:
```
SKB (ffff88977d9bd000) DUMP: len=64858 data_len=64858 truesize=65882 users=1
  head=ffff88977d9bd100 data=ffff88977d9bd260 tail=160 end=1c0
  nr_frags=17 frag_list=          (null) next=ffff88977d241c00 prev=ffff88977d241000
  head data (0):
  frag 0 (addr=ffff8897713b92af pg_off=687 size=3409 pg_ref=3):
  frag 1 (addr=ffff88976d002000 pg_off=0 size=4096 pg_ref=2):
  frag 2 (addr=ffff88976d003000 pg_off=0 size=4096 pg_ref=2):
  frag 3 (addr=ffff88971c0a5000 pg_off=0 size=9 pg_ref=1):
  frag 4 (addr=ffff88977726c000 pg_off=0 size=4096 pg_ref=2):
  frag 5 (addr=ffff88977726d000 pg_off=0 size=4096 pg_ref=2):
  frag 6 (addr=ffff88977726e000 pg_off=0 size=4096 pg_ref=2):
  frag 7 (addr=ffff88977726f000 pg_off=0 size=4096 pg_ref=2):
  frag 8 (addr=ffff88976d004000 pg_off=0 size=4096 pg_ref=2):
  frag 9 (addr=ffff88976d005000 pg_off=0 size=4096 pg_ref=2):
  frag 10 (addr=ffff88976d006000 pg_off=0 size=4096 pg_ref=2):
  frag 11 (addr=ffff88976d007000 pg_off=0 size=4096 pg_ref=2):
  frag 12 (addr=ffff88974df1c000 pg_off=0 size=4096 pg_ref=2):
  frag 13 (addr=ffff88974df1d000 pg_off=0 size=4096 pg_ref=2):
  frag 14 (addr=ffff88974df1e000 pg_off=0 size=4096 pg_ref=2):
  frag 15 (addr=ffff88974df1f000 pg_off=0 size=4096 pg_ref=2):
  frag 16 (addr=ffff889770340000 pg_off=0 size=4096 pg_ref=2):
```
And after insertion:
```
SKB (ffff88977d9bd000) DUMP: len=60771 data_len=60771 truesize=61795 users=1
  head=ffff88977d9bd100 data=ffff88977d9bd260 tail=160 end=1c0
  nr_frags=17 frag_list=          (null) next=ffff88977d241c00 prev=ffff88977d241000
  head data (0):
  frag 0 (addr=ffff8897713b92af pg_off=687 size=3409 pg_ref=3):
  frag 1 (addr=ffff88976d002000 pg_off=0 size=4096 pg_ref=2):
  frag 2 (addr=ffff88976d003000 pg_off=0 size=4096 pg_ref=2):
  frag 3 (addr=ffff88971c0a5009 pg_off=9 size=9 pg_ref=2):
  frag 4 (addr=ffff88971c0a5000 pg_off=0 size=9 pg_ref=2):
  frag 5 (addr=ffff88977726c000 pg_off=0 size=4096 pg_ref=2):
  frag 6 (addr=ffff88977726d000 pg_off=0 size=4096 pg_ref=2):
  frag 7 (addr=ffff88977726e000 pg_off=0 size=4096 pg_ref=2):
  frag 8 (addr=ffff88977726f000 pg_off=0 size=4096 pg_ref=2):
  frag 9 (addr=ffff88976d004000 pg_off=0 size=4096 pg_ref=2):
  frag 10 (addr=ffff88976d005000 pg_off=0 size=4096 pg_ref=2):
  frag 11 (addr=ffff88976d006000 pg_off=0 size=4096 pg_ref=2):
  frag 12 (addr=ffff88976d007000 pg_off=0 size=4096 pg_ref=2):
  frag 13 (addr=ffff88974df1c000 pg_off=0 size=4096 pg_ref=2):
  frag 14 (addr=ffff88974df1d000 pg_off=0 size=4096 pg_ref=2):
  frag 15 (addr=ffff88974df1e000 pg_off=0 size=4096 pg_ref=2):
  frag 16 (addr=ffff88974df1f000 pg_off=0 size=4096 pg_ref=2):
```

The last insertion was not done at expected fragment, it was inserted as  fragment 3, instead of 8. This leads to broken h2 framing. Still looking for exact conditions. Happens only sometimes and only for some files for me.

ikoveshnikov: The insertion happens in a wrong place because of optimisation in `skb_fragment()`. 

Let's again take a look into the faulty skb:
```
SKB (ffff88977d9bd000) DUMP: len=64858 data_len=64858 truesize=65882 users=1
  head=ffff88977d9bd100 data=ffff88977d9bd260 tail=160 end=1c0
  nr_frags=17 frag_list=          (null) next=ffff88977d241c00 prev=ffff88977d241000
  head data (0):
  frag 0 (addr=ffff8897713b92af pg_off=687 size=3409 pg_ref=3):
  frag 1 (addr=ffff88976d002000 pg_off=0 size=4096 pg_ref=2):
  frag 2 (addr=ffff88976d003000 pg_off=0 size=4096 pg_ref=2):
  frag 3 (addr=ffff88971c0a5000 pg_off=0 size=9 pg_ref=1):
  frag 4 (addr=ffff88977726c000 pg_off=0 size=4096 pg_ref=2):
  frag 5 (addr=ffff88977726d000 pg_off=0 size=4096 pg_ref=2):
  frag 6 (addr=ffff88977726e000 pg_off=0 size=4096 pg_ref=2):
  frag 7 (addr=ffff88977726f000 pg_off=0 size=4096 pg_ref=2):
  frag 8 (addr=ffff88976d004000 pg_off=0 size=4096 pg_ref=2):
  frag 9 (addr=ffff88976d005000 pg_off=0 size=4096 pg_ref=2):
  frag 10 (addr=ffff88976d006000 pg_off=0 size=4096 pg_ref=2):
  frag 11 (addr=ffff88976d007000 pg_off=0 size=4096 pg_ref=2):
  frag 12 (addr=ffff88974df1c000 pg_off=0 size=4096 pg_ref=2):
  frag 13 (addr=ffff88974df1d000 pg_off=0 size=4096 pg_ref=2):
  frag 14 (addr=ffff88974df1e000 pg_off=0 size=4096 pg_ref=2):
  frag 15 (addr=ffff88974df1f000 pg_off=0 size=4096 pg_ref=2):
  frag 16 (addr=ffff889770340000 pg_off=0 size=4096 pg_ref=2):
```

 In the example we needed to insert a new data piece at `TfwMsgIter {skb=ffff88977d9bd000, frag=8}`, at offset `ffff88976d004000`. The `skb_fragment()` function doesn't take the `frag` as a hint (it's mostly unknown for callers anyway), and tries  to find the exact fragment. The function is happy, if the fragment ends just before the requested pointer, and makes an insertion there. In this example an insertion is requested before the  `ffff88976d004000` address, fragment 8. But fragment 2 ends just before the the same address. That is why insertion happens as fragment at unexpected place.

Why we never seen the issue before? We was lucky enough not to get skbs with such memory map in those rare functional tests which involves modification of extremely huge number/size of headers. 

We have no code depending on that behaviour, so I've removed the optimisation. Checked many times, the data received by a client now always match the original data.

tempesta_fw/ss_skb.c
```diff
@@ -788,18 +788,13 @@ __skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
...
-				/*
-				 * @pspt is at the end of the frag (zero tail
-				 * length): append if @len > 0 or move to the
-				 * next frag for deletion.
-				 */
-				if (len > 0)
-					goto append;
-				continue;
-			}
 			len = max(len, -t_size);
```
krizhanovsky, line 798: We don't need `t_size` any more, just use `max_t(long, len, offset - d_size)` and the same in the `if (offset >= 0 && offset < d_size)` above

tempesta_fw/ss_skb.c
```diff
@@ -788,18 +788,13 @@ __skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
...
-			if (!t_size) {
-				/*
-				 * @pspt is at the end of the frag (zero tail
-				 * length): append if @len > 0 or move to the
-				 * next frag for deletion.
-				 */
-				if (len > 0)
-					goto append;
-				continue;
-			}
```
krizhanovsky, line 802: Very subtle place and `__split_pgfrag_add()` also optimizes zero offset, so quite a good fix.

krizhanovsky (changes requested):
We either need more discussion or work on the PR besides the minor cleanups.

tempesta_fw/http_msg.c
```diff
@@ -1462,17 +1434,30 @@ tfw_h2_msg_rewrite_data(TfwHttpTransIter *mit, const TfwStr *str,
...
+	 * current skb fragment, then it was appended to a previous one.
+	 * Insertions into linear part doesn't require iterator updates.
+	 */
+	*off = dst.data;
+	if (it->frag >= 0) {
+		skb_frag_t *f = &skb_shinfo(it->skb)->frags[it->frag];
+		long offset = dst.data - (char *)skb_frag_address(f);
+		if (offset < 0 || offset > skb_frag_size(f))
+			it->frag--;
+	}
```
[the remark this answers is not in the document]
ikoveshnikov: Didn't think about it as move back. Just thought about is as the laziest and assumption-free way to obtain valid `(it, off)` pair:  `ss_skb_get_room()` can split skbs, if there are too many fragments, and the original iterator fragment may be moved to a next skb. Happens quite frequently on large bodies proxying. 

The `ss_skb_get_room()` always return plain string with exactly pointed beginning, only fragment number is unclear. It's easy to get a pointer to the first or the last byte of the string. But pointer to the byte just behind the string is bit more complicated: `it->skb`, `it->frag`, `off` - they're all will be updated, and we also need a check, that `it->skb` doesn't point to `it->skb_head`. This semantically equals to a single `iter_move()` call. With this in mind `iter_move()` will be called twice: first to step behind the inserted string, and second  - to skip next `max_sz` bytes.

Since the data is always inserted as a plain string, `iter_move()` will jump over the whole string by one iteration. That's why I see no practical reason to move the iterator in `tfw_http_msg_insert()` behind the inserted string.

tempesta_fw/ss_skb.c
ikoveshnikov: Yeah, we happily jumped over this bug. I was very lucky to catch it. Data corruption is possible on large response headers, but too many conditions must met. Happily this cannot be exploited by an attacker.

tempesta_fw/ss_skb.c
```diff
@@ -863,7 +858,14 @@ ss_skb_get_room(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
 	int r = skb_fragment(skb_head, skb, pspt, len, it);
 	if (r == len)
 		return 0;
-	return r;
+	/*
+	 * TODO: skb_fragment() returns a number of processed data, which can
```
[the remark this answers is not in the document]
ikoveshnikov: Well, there is no such direct scenario, because `ss_skb_get_room()` is called with `len > 0` and `skb_fragment()` now is never called by other modules. But when it's called, the `len` can be modified e.g. here (if `len < 0`): https://github.com/tempesta-tech/tempesta/blob/8d90875697968ddbeccdcdc008503f9fe2e290e4/tempesta_fw/ss_skb.c#L803
and `skb_fragment()` will return a value different from the `|len|`.

The function removed in this PR as dead one -- `tfw_http_msg_del_eol()` -- could do that. 

Now this never happens, but I decided to add this return value modifications to prevent such cases.

tempesta_fw/http_msg.c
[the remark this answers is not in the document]
krizhanovsky: My point isn't about DRY principle and basically I'm OK with you explanation. My concern is that the current h2 code is overcomplicated with the fragments and I'm afraid that we won't see good enough performance. We should have been benefited from the zero copy approach and avoiding memory allocations, but in fact we do a lot of _senseless_ logic (in general meaning, not the current patch). For example, `tfw_msg_iter_move()` for DATA frame is called after `tfw_http_msg_insert()` for HEADERS and both the functions check `it->frag` for zero. The worst thing about both the functions is that both of them were written to be called from `__tfw_h2_make_frames()` only, which is called from only 2 pieces of the same function `tfw_h2_make_frames()`. So there is generic code with extra conditions (and also complexity!), but I don't see enough motivation for the generic functions.

I ask you to review this place carefully and rework in in simpler and more efficient way. Otherwise we might find ourselves as Nginx: there is a long tail in perf and the whole code must be rewritten to make it more or less faster. We defenitelly won't see these two function in perf top, but we do need to remove such ineficiencies as we observe them.

tempesta_fw/ss_skb.c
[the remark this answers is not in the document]
krizhanovsky: So do we really need `TODO` something on this as pointed out in the comment? What about just to add `WARN_ON()` instead?

tempesta_fw/http_msg.c
```diff
@@ -1443,21 +1443,11 @@ tfw_http_msg_insert(TfwMsgIter *it, char **off, const TfwStr *data)
 	int r;
 	TfwStr dst = {};
 
-	if ((r = ss_skb_get_room(it->skb_head, it->skb, *off, data->len, &dst)))
+	if ((r = ss_skb_get_room_w_frag(it->skb_head, it->skb, *off, data->len, &dst, &it->frag))) {
```
ikoveshnikov, line 1446: Good choice to pass the exact modified fragment to the caller! 

It seems like `skb_head`, `skb`, `frag` parameters  represent `TfwMsgIter` structure, and in most places where the `get_room()` is called right in a context where an iterator is defined and used to provide values for `get_room()`. But I bet, most functions doesn't expect that the iterator may change after `get_room()` call. Same I can say about all the similar functions in this PR.

avbelov23: Not all users of the function `ss_skb_get_room...()` use `TfwMsgIter`, besides, I did not want to introduce the dependency ss_skb.h on msg.h

avbelov23:
DONE

tempesta_fw/ss_skb.c
```diff
@@ -788,20 +804,15 @@ __skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
 		d_size = skb_frag_size(frag);
```
krizhanovsky, line 804: We do not update `fragn` for the case `len > 0 && !offset` above. Is this correct? Need a review from @i-rinat since he fixed many bugs recently with iterator frags.

tempesta_fw/ss_skb.c
```diff
@@ -860,10 +871,26 @@ int
 ss_skb_get_room(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
 		unsigned int len, TfwStr *it)
 {
-	int r = skb_fragment(skb_head, skb, pspt, len, it);
+	int _;
+
+	return ss_skb_get_room_w_frag(skb_head, skb, pspt, len, it, &_);
```
krizhanovsky, line 876: Please move `ss_skb_get_room()` using `ss_skb_get_room_w_frag()` after the `ss_skb_get_room_w_frag()` definition.

tempesta_fw/ss_skb.c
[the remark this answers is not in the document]
krizhanovsky: OK, but it seems there is nothing to do for now, so please just remove `TODO` from the comment.

krizhanovsky (approved):
I have only minor comments for the PR. An accurate review from @i-rinat is required since very subtle things with the skb iterator are changed in the PR.

tempesta_fw/http_msg.c
```diff
@@ -1462,17 +1434,21 @@ tfw_h2_msg_rewrite_data(TfwHttpTransIter *mit, const TfwStr *str,
...
  * Insert data from string @data to message at offset defined by message
  * iterator @it and @off. This function doesn't maintain message structure.
- * After insertion message iterator will point at the end of fragment right
- * before the inserted data.
+ * After insertion message iterator and @data will point at the start of
+ * inserted data fragment.
  */
 int
-tfw_http_msg_insert(TfwMsgIter *it, char *off, const TfwStr *data)
+tfw_http_msg_insert(TfwMsgIter *it, char **off, const TfwStr *data)
```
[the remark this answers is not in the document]
avbelov23: No. `tfw_msg_iter_move()` use

tempesta_fw/http_msg.c
```diff
@@ -1462,17 +1434,21 @@ tfw_h2_msg_rewrite_data(TfwHttpTransIter *mit, const TfwStr *str,
...
 	int r;
 	TfwStr dst = {};
 
-	if ((r = ss_skb_get_room(it->skb_head, it->skb, off, data->len, &dst)))
+	if ((r = ss_skb_get_room_w_frag(it->skb_head, it->skb, *off, data->len, &dst, &it->frag))) {
 		return r;
+	}
+
+	*off = dst.data;
+	it->skb = dst.skb;
```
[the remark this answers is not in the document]
avbelov23: `tfw_msg_iter_move()` use. Here `it` is `TfwMsgIter`. In  `skb_fragment()` `it` is `TfwStr`
In `__tfw_h2_make_frames()` a loop and `tfw_http_msg_insert()` changes `data`, and `tfw_msg_iter_move()` uses `data`

tempesta_fw/ss_skb.c
```diff
@@ -247,15 +247,17 @@ __skb_data_address(struct sk_buff *skb)
...
 {
 	struct skb_shared_info *si = skb_shinfo(skb);
 	if (i < si->nr_frags) {
 		it->data = skb_frag_address(&si->frags[i]);
 		it->skb = skb;
+		*fragn = i;
 	} else {
 		it->skb = skb->next;
 		it->data = __skb_data_address(it->skb);
+		*fragn = -1;
```
i-rinat, line 260: `__skb_data_address()` can return either a pointer to skb's linear part, or it can return a pointer to the first fragment. `*fragn` may be `0` sometimes.

i-rinat: The fix looks good.

PR #1459, opened 2020-10-23
Merge TLS messages sent on session resumption scenario

ikoveshnikov:
Second part of the changes required by the #1054 

Merge TLS messages together when sending them to the client. The `ttls_handshake_server_hello()` and `ttls_handshake_finished()` declared their own buffers, which represented two separate message sets. In this PR both functions use a single buffer declared in `ttls_handshake_server_step()`. Now both functions only fills the buffer, while   `ttls_handshake_server_step()` triggers the send process.

### DIFF
--- a/tls/tls_srv.c
+++ b/tls/tls_srv.c
@@ -1887,7 +1887,7 @@ ttls_parse_certificate_verify(TlsCtx *tls, unsigned char *buf, size_t len,
 }
 
 static int
-ttls_write_new_session_ticket(TlsCtx *tls, struct sg_table *_,
+ttls_write_new_session_ticket(TlsCtx *tls, struct sg_table *sgt,
 			      unsigned char **in_buf)
 {
 	int r;
@@ -1895,15 +1895,6 @@ ttls_write_new_session_ticket(TlsCtx *tls, struct sg_table *_,
 	uint32_t lifetime = 0;
 	TlsIOCtx *io = &tls->io_out;
 	unsigned char *p = *in_buf;
-	/*
-	 * FIXME: #1054 Don't send the message alone, combine it with
-	 * CHANGE_CIPHER_SPEC and TTLS_HS_FINISHED. For now there are digest
-	 * mismatch errors on client side.
-	 */
-	struct scatterlist sg[1];
-	struct sg_table sgt = { .sgl = sg };
-
-	sg_init_table(sgt.sgl, 1);
 
 	/*
 	 * struct {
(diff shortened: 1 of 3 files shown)


tls/tls_srv.c
```diff
@@ -2246,5 +2235,11 @@ ttls_handshake_server_step(TlsCtx *tls, unsigned char *buf, size_t len,
 	}
 	T_FSM_FINISH(r, tls->state);
 
+	if (unlikely(pg)) {
```
krizhanovsky, line 2238: Please leave the comment `/* If we exit here, then something went wrong. */`

tls/ttls.c
```diff
@@ -1670,16 +1670,28 @@ ttls_parse_certificate(TlsCtx *tls, unsigned char *buf, size_t len,
...
 }
 void
-ttls_write_change_cipher_spec(TlsCtx *tls)
-{
-	TlsIOCtx *io = &tls->io_out;
+ttls_write_change_cipher_spec(TlsCtx *tls, struct sg_table *sgt,
+			      unsigned char **in_buf)
+{
+	/* The ChangeCipherSpec message is added after NewSessionTicket one. */
+	if (unlikely(in_buf)) {
```
krizhanovsky, line 1677: Why `unlikely()`? I think the most TLS handshakes support the session tickets extension.

Note, that I added 2 more test cases into #1054. Please make sure that this branch is properly covered with the tests.

krizhanovsky (approved):
The change is good in general, but more work is required.

tls/tls_srv.c
```diff
@@ -2065,62 +2022,71 @@ ttls_handshake_server_hello(TlsCtx *tls)
...
 
+static int
+ttls_handshake_init_out_buffers(struct sg_table *sgt, unsigned char **p,
+				struct page **pg)
+{
+	if (*p)
+		return 0;
+
+	*p = pg_skb_alloc(2048, GFP_ATOMIC, NUMA_NO_NODE);
+	if (!*p)
```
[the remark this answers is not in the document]
krizhanovsky: Please explore whether it makes sense to aggregate the memory chunks with `TlsHandshake` allocation-  probably we can do everything within single page.

tls/ttls.c
ikoveshnikov: Unlikely was replaced with likely - with tickets enabled (mostly true) ChangeCipherSpec message will always be the second message in the sequence. Comment was updated.

Issue #1452, opened 2020-10-09
Inconsistent vhost & JSCH configuration


krizhanovsky:
## Scope

https://github.com/tempesta-tech/tempesta/wiki/Sticky-Cookie#http-session-management makes false statement that vhost can inherit default sticky configuration. There is quite nasty inconsistency. The config:
```
listen 192.168.100.4:80;

sticky {
        cookie name=my_js_cookie enforce max_misses=3 timeout=3;
        js_challenge resp_code=503 delay_min=1000 delay_range=1000 delay_limit=3000 /root/tempesta/etc/js_challenge.html;
}

srv_group default {
        server 127.0.0.1:9090;
}

vhost default {
        proxy_pass default;
}

cache 1;
cache_fulfill * *;

http_chain {
        -> default;
}
```
Works fine, but
```
listen 192.168.100.4:80;

sticky {
        cookie name=my_js_cookie enforce max_misses=3 timeout=3;
        js_challenge resp_code=503 delay_min=1000 delay_range=1000 delay_limit=3000 /root/tempesta/etc/js_challenge.html;
}

srv_group default {
        server 127.0.0.1:9090;
}

vhost debian {
        proxy_pass default;
}

cache 1;
cache_fulfill * *;

http_chain {
        -> debian;
}
```
does not work. Any non default vhost must use stick inside vhost configuration only, so to make the last config work we have to rewrite it as
```
listen 192.168.100.4:80;

srv_group default {
        server 127.0.0.1:9090;
}

vhost debian {
        sticky {
                cookie name=my_js_cookie enforce max_misses=3 timeout=3;
                js_challenge resp_code=503 delay_min=1000 delay_range=1000 delay_limit=3000 /root/tempesta/etc/js_challenge.html;
        }

        proxy_pass default;
}

cache 1;
cache_fulfill * *;

http_chain {
        -> debian;
}
```

It seems that the only reason for the behaviour is that `update_template.pl` can not track sticky state across the configuration file.

## Testing

- [x] The default sticky in a named vhost
- [x] Double default sticky for a named host - the second sticky should be used
- [x] Double default sticky for the default host - the second sticky should be used
- [x] Internal vhost sticky cookie overrides the default one

## Documentation

Update https://github.com/tempesta-tech/tempesta/wiki/Sticky-Cookie#http-session-management with words that default `sticky` must be always *before* vhost, otherwise the sticky section will be *ignored*.

[...]

PR #1460, opened 2020-11-04
Inherit sticky cookie configuration as stated in wiki

ikoveshnikov:
Inherit sticky cookie configuration as stated in wiki

Main changes:
- 'sticky' directive at top-level now updates default values instead of configuring implicit default vhost. Named vhosts inherit 'sticky' directive if it was defined at top-level.
- implicit default vhost now inherits effective defaults from 'sticky'
- sticky cookie secret is not stored as plain text for release configuration, but is still available in debug builds.
- Sticky cookie secret length is relaxed to 1kb when it overrides default values.   If configured for named vhost, there is no limitation at all. Previously it had length of 20 bytes with very limited entropy since it was defined as a string in the configuration file. Now the length is relaxed, so keys with better entropy can be used.

### DIFF
--- a/tempesta_fw/http_sess_conf.c
+++ b/tempesta_fw/http_sess_conf.c
@@ -1,7 +1,7 @@
 /**
  *		Tempesta FW
  *
- * Copyright (C) 2019 Tempesta Technologies, Inc.
+ * Copyright (C) 2019-2020 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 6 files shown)


tempesta_fw/http_sess_conf.c
```diff
@@ -32,17 +32,29 @@ static TfwVhost *cur_vhost;
 static const unsigned int tfw_cfg_jsch_code_dflt = 503;
 #define TFW_CFG_JS_PATH "/etc/tempesta/js_challenge.html"
```
krizhanovsky, line 33: Please update the file copyrights

krizhanovsky (approved):
LGTM and I can confirm that #1452 is fixed with the patch. Please just update the wiki with the configuration examples and generic descriptions as well as with https://github.com/tempesta-tech/enterprise/issues/18

PR #1461, opened 2020-11-13
Write header names in lower case when transform h1 responses to h2

### LINKED ISSUE #1380: Write header names in lower case when transform h1 responses to h2
From the [RFC 7540 8.1.2](https://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2)
 > Just as in HTTP/1.x, header field names are strings of ASCII characters that are compared in a case-insensitive fashion. However, header field names MUST be converted to lowercase prior to their encoding in HTTP/2. A request or response containing uppercase header field names MUST be treated as malformed (Section 8.1.2.6).

According the RFC an end point must close a connection (send error) on receiving our responses, ~~but major browsers and curl _ignore_ this requirement. Well, it's a problem for us, actually.~~ **All modern browsers treat mixed case in headers names as protocol error.** Converting header names to lower case definitely has performance impact. ~~But even if browsers are OK, we can't just break the RFC requirement and leave headers names in mixed case.~~

From https://github.com/tempesta-tech/tempesta/pull/1374#discussion_r368960033

avbelov23:
Fix #1380 

When placing in the cache and when transforming header names into h2, we transform them to lowercase

### DIFF
--- a/tempesta_fw/http_msg.c
+++ b/tempesta_fw/http_msg.c
@@ -1360,9 +1360,10 @@ this_chunk:
 	return 0;
 }
 
-int
-tfw_h2_msg_rewrite_data(TfwHttpTransIter *mit, const TfwStr *str,
-			const char *stop)
+static int
+tfw_h2_msg_rewrite_data_common(TfwHttpTransIter *mit, const TfwStr *str,
+                               const char *stop,
+                               void cpy(void *dest, const void *src, size_t n))
 {
 	const TfwStr *c, *end;
 	TfwMsgIter *it = &mit->iter;
@@ -1417,8 +1418,7 @@ this_chunk:
 
 			return -E2BIG;
 		}
-
-		memcpy_fast(mit->curr_ptr, c->data + c_off, n_copy);
+		cpy(mit->curr_ptr, c->data + c_off, n_copy);
 
 		T_DBG3("%s: acc_len=%lu, n_copy=%u, mit->curr_ptr='%.*s',"
 		       " ptr_diff=%ld\n", __func__, mit->acc_len, n_copy,
@@ -1457,6 +1457,20 @@ this_chunk:
 	}
 
 	return 0;
+}
+
+int
+tfw_h2_msg_rewrite_data(TfwHttpTransIter *mit, const TfwStr *str,
+                        const char *stop)
+{
+	return tfw_h2_msg_rewrite_data_common(mit, str, stop, memcpy_fast);
+}
+
+int
+tfw_h2_msg_rewrite_data_lc(TfwHttpTransIter *mit, const TfwStr *str,
+                           const char *stop)
+{
+	return tfw_h2_msg_rewrite_data_common(mit, str, stop, tfw_cstrtolower);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


krizhanovsky (approved):
LGTM

ikoveshnikov (changes requested):
The changes looks good, but an addition is required for configuration handling: we have directives as `resp_hdr_set|resp_hdr_add` which are defined by end user. It worth to convert them to lowercase even if user used mixed case. The same should be done for request header modification to have uniform configuration handling.

i-rinat (approved):
Looks good.

tempesta_fw/http_msg.c
```diff
@@ -42,18 +42,39 @@ tfw_http_msg_make_hdr(TfwPool *pool, const char *name, const char *val)
...
+	++h_c;
+
+	memset(h_c, 0, sizeof(*h_c));
+	h_c->data = data;
+	h_c->len = v_len;
+	h_c->flags = TFW_STR_HDR_VALUE;
+	memcpy(data, val, v_len);
 
-	return tfw_strdup(pool, &tmp_hdr);
+	return hdr;
```
ikoveshnikov, line 77: There are only two places, where `tfw_strdup()` is called. Here and in unit tests. Seems we can use there `
tfw_cstrtolower_wo_avx2` instead of `memcpy`, if the flag `TFW_STR_HDR_VALUE` is not set on source chunk. What do you say? 

If we remove the `tfw_strdup` call here, then there is no reason to keep the function, and corresponding tests... 

You have to chose one of the ways, or we will have two implementations of `tfw_strdup`, less and more generic, the later would be never called.

ikoveshnikov (approved):

tempesta_fw/http_msg.c
```diff
@@ -42,18 +42,39 @@ tfw_http_msg_make_hdr(TfwPool *pool, const char *name, const char *val)
...
+
+	TFW_STR_INIT(h_c);
+	h_c->data = data;
+	h_c->len = n_len;
+	tfw_cstrtolower_wo_avx2(data, name, n_len);
+
+	data += h_c->len;
+	++h_c;
+
+	TFW_STR_INIT(h_c);
```
ikoveshnikov, line 71: The reason why `resp_hdr_set|add` tests fails on your branch is here. The allocated memory segment is used like this:
`| compound str | sub str for name | sub str for value | name buf | value buf |` . But if the value is missing, the string becomes shorter: `| compound str | sub str for name | name buf | ` . 

In this code segment substring for value in unconditionally written to memory segment after name substring. But on short allocations the whole value substring is missing and name buffer is overwriten with TfwStr header.

tempesta_fw/http_msg.c
```diff
@@ -42,18 +42,39 @@ tfw_http_msg_make_hdr(TfwPool *pool, const char *name, const char *val)
...
+	h_c->data = data;
+	h_c->len = v_len;
+	h_c->flags = TFW_STR_HDR_VALUE;
+	memcpy(data, val, v_len);
```
krizhanovsky, line 75: If the code works in softirq, then we should avoid using `memcpy()` and similar functions: they're quite slow in the Linux kernel (much slower than in the user space with GCC/glibc), so if possible use our versions of the functions.

avbelov23: This func call from process context wo `fpu_begin()`/`fpu_end()`

PR #1462, opened 2020-11-14
Hide autogenerated code from git

i-rinat:
The project now uses autogenerated tables and a generator to make the tables. They are not tracked by Git, so it would be nice to tell Git to ignore them. It would also be nice to remove the generator `tgen_ec256` when `make clean` is called.


ikoveshnikov (approved):

krizhanovsky (approved):
LGTM

PR #1465, opened 2020-11-19
Close tls connection after encryption is complete

### LINKED ISSUE #1431: Can't read large files from SSL socket on client side if 'Connection: close' is set by client
## Scope

In my preset Tempesta is started on VM, `wrk` - on host machine.
Configure Tempesta from current master to serve TLS connections: 
```
server 127.0.0.1:8080;

listen 443;
listen 80;

tls_match_any_server_name;
tls_certificate /home/user/cert/tfw-root.crt;
tls_certificate_key /home/user/cert/tfw-root.key;
```

And run wrk client with 'connection: close' header against not too small resource, e.g. 18Kb GPL2 licence shipped with Tempesta.
When a HTTPS protocol is used a large amount of read errors appears in wrk statistics, and RPS counters are zero: 
```
 % wrk -d 1 -t 1 -c 1 -H "connection: close" https://192.168.122.12:443/LICENSE
Running 1s test @ https://192.168.122.12:443/LICENSE
  1 threads and 1 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     0.00us    0.00us   0.00us    -nan%
    Req/Sec     0.00      0.00     0.00      -nan%
  0 requests in 1.10s, 4.36MB read
  Socket errors: connect 0, read 317, write 0, timeout 0
Requests/sec:      0.00
Transfer/sec:      3.96MB
```

With plain HTTP no errors appear in the statistics:
```
% wrk -d 1 -t 1 -c 1 -H "connection: close" http://192.168.122.12/LICENSE 
Running 1s test @ http://192.168.122.12:8082/LICENSE
  1 threads and 1 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   364.22us  208.33us   3.92ms   95.52%
    Req/Sec     2.06k   192.22     2.25k    80.00%
  2049 requests in 1.00s, 35.89MB read
Requests/sec:   2048.70
Transfer/sec:     35.89MB
```

In the same time other HTTP servers does not have the issue described:
```
% wrk -d 1 -t 1 -c 1 -H "connection: close" https://192.168.122.12:8081/LICENSE
Running 1s test @ https://192.168.122.12:8081/LICENSE
  1 threads and 1 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   469.09us  320.03us   3.61ms   94.34%
    Req/Sec   693.73    147.07     0.88k    81.82%
  759 requests in 1.10s, 13.26MB read
Requests/sec:    690.20
Transfer/sec:     12.06MB
```

## Testing

Make a test, checking that both large and small resources can be gathered even if client asks to close a connection once response is sent.

### LINKED ISSUE #1435: Nullptr dereference in tcp_schedule_loss_probe()
## Scope

The crash happens on wrk load under heavy load on baremetal server, couldn't reproduce it in my VM. 
To reproduce:
- build Tempesta on master
- set configuration as follows:
```
 server 127.0.0.1:8080;

listen 443 proto=https;
listen 80;

tls_match_any_server_name;
tls_certificate /home/ik/tempesta/etc/tfw-root.crt;
tls_certificate_key /home/ik/tempesta/etc/tfw-root.key;

cache_fulfill * *;
cache 2;
```
- run `wrk` over https protocol with `connection: close` header
```
wrk -d 120 -c 1000 -t 8 -H 'connection: close' https://192.168.76.7:443/1k"
``` 

- Get the crash: 
```
[  474.460446] net_ratelimit: 10760 callbacks suppressed
[  474.460447] TCP: Tempesta: cannot encrypt data (-22), reset a TLS connection.
[  474.461447] TCP: Tempesta: cannot encrypt data (-22), reset a TLS connection.
[  474.461455] TCP: Tempesta: cannot encrypt data (-22), reset a TLS connection.
[  474.461460] TCP: Tempesta: cannot encrypt data (-22), reset a TLS connection.
[  474.461466] TCP: Tempesta: cannot encrypt data (-22), reset a TLS connection.
[  474.461474] TCP: Tempesta: cannot encrypt data (-22), reset a TLS connection.
[  474.461478] TCP: Tempesta: cannot encrypt data (-22), reset a TLS connection.
[  474.461484] TCP: Tempesta: cannot encrypt data (-22), reset a TLS connection.
[  474.461487] TCP: Tempesta: cannot encrypt data (-22), reset a TLS connection.
[  474.461496] TCP: Tempesta: cannot encrypt data (-22), reset a TLS connection.
[  474.750531] BUG: unable to handle kernel NULL pointer dereference at 0000000000000010
[  474.758504] IP: tcp_schedule_loss_probe+0xc2/0x170
[  474.763445] PGD 0 P4D 0 
[  474.766367] Oops: 0000 [#1] SMP PTI
[  474.769988] Modules linked in: tempesta_fw(O) tempesta_db(O) sha256_ssse3 sha512_ssse3 sha512_generic ccm tempesta_tls(O) tempesta_lib(O) binfmt_misc nft_chain_route_ipv4 xt_CHECKSUM nft_chain_nat_ipv4 nf_nat_ipv4 ipt_MASQUERADE nf_nat_masquerade_ipv4 nf_nat nf_conntrack_ipv4 nf_defrag_ipv4 xt_conntrack nf_conntrack ipt_REJECT nf_reject_ipv4 nft_counter nft_meta xt_tcpudp nft_compat tun bridge cpuid nf_tables_bridge nf_tables_ipv6 nf_tables_ipv4 nf_tables nfnetlink intel_rapl x86_pkg_temp_thermal intel_powerclamp coretemp kvm_intel kvm irqbypass crct10dif_pclmul crc32_pclmul ghash_clmulni_intel intel_cstate intel_uncore intel_rapl_perf joydev evdev sg ast iTCO_wdt ttm drm_kms_helper drm iTCO_vendor_support video ie31200_edac mei_me intel_pch_thermal shpchp mei pcspkr ipmi_si ipmi_devintf ipmi_msghandler
[  474.842613]  button macvlan knem(O) 8021q garp mrp stp llc nfsd auth_rpcgss nfs_acl lockd grace sunrpc ip_tables x_tables autofs4 ext4 crc16 mbcache jbd2 fscrypto ecb btrfs zstd_decompress zstd_compress xxhash raid10 raid456 async_raid6_recov async_memcpy async_pq async_xor async_tx xor raid6_pq libcrc32c crc32c_generic raid1 raid0 multipath mlx4_ib linear md_mod ib_core mlx4_en hid_generic usbhid hid sd_mod crc32c_intel ahci megaraid_sas libahci xhci_pci igb xhci_hcd i2c_algo_bit i2c_i801 mlx4_core devlink aesni_intel dca aes_x86_64 crypto_simd ptp cryptd pps_core glue_helper libata usbcore usb_common scsi_mod fan thermal
[  474.899633] CPU: 0 PID: 7 Comm: ksoftirqd/0 Tainted: G           O    4.14.0-tempesta-amd64 #1 Debian 4.14.32-tfw7-1
[  474.910510] Hardware name: Supermicro Super Server/X11SSL-CF, BIOS 2.0a 03/15/2017
[  474.918444] task: ffff982973f09f00 task.stack: ffffa37903198000
[  474.924735] RIP: 0010:tcp_schedule_loss_probe+0xc2/0x170
[  474.930429] RSP: 0018:ffffa3790319b990 EFLAGS: 00010246
[  474.936094] RAX: 0000000000000000 RBX: ffff9828ebb1dd80 RCX: 0000000000000001
[  474.943638] RDX: 0000000000000000 RSI: 0000000000000000 RDI: 0000000000000035
[  474.951210] RBP: 0000000000000000 R08: ffff9829b7825b30 R09: 0000000000000001
[  474.958748] R10: 0000000000000001 R11: 00000000529d64b1 R12: 0000000000000007
[  474.966544] R13: 0000000000000002 R14: ffff9828ec342000 R15: 00000000ffffffea
[  474.974087] FS:  0000000000000000(0000) GS:ffff9829b7800000(0000) knlGS:0000000000000000
[  474.982586] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[  474.988826] CR2: 0000000000000010 CR3: 000000064ac0a003 CR4: 00000000003606f0
[  474.996393] DR0: 0000000000000000 DR1: 0000000000000000 DR2: 0000000000000000
[  475.003950] DR3: 0000000000000000 DR6: 00000000fffe0ff0 DR7: 0000000000000400
[  475.011516] Call Trace:
[  475.014408]  tcp_write_xmit+0x2a7/0x11c0
[  475.018808]  __tcp_push_pending_frames+0x2d/0xd0
[  475.023878]  tcp_rcv_state_process+0x418/0xec0
[  475.028785]  ? tfw_classify_conn_close+0x30/0x30 [tempesta_fw]
[  475.035148]  ? security_sock_rcv_skb+0x3c/0x50
[  475.040051]  ? tcp_v4_do_rcv+0xb4/0x1d0
[  475.044345]  tcp_v4_do_rcv+0xb4/0x1d0
[  475.048475]  tcp_v4_rcv+0x947/0xa50
[  475.052438]  ip_local_deliver_finish+0x9a/0x1c0
[  475.057442]  ip_local_deliver+0x6b/0xe0
[  475.061742]  ? ip_rcv_finish+0x400/0x400
[  475.066403]  ip_rcv+0x289/0x3c0
[  475.070028]  ? inet_del_offload+0x40/0x40
[  475.074524]  __netif_receive_skb_core+0x84f/0xb30
[  475.079714]  ? inet_gro_receive+0x204/0x2b0
[  475.084411]  ? netif_receive_skb_internal+0xbb/0xe0
[  475.089782]  netif_receive_skb_internal+0xbb/0xe0
[  475.094959]  napi_gro_frags+0xe5/0x210
[  475.099195]  mlx4_en_process_rx_cq+0xa8c/0xbd0 [mlx4_en]
[  475.105077]  ? tcp_transmit_skb+0x535/0x9c0
[  475.109744]  ? tcp_delack_timer_handler+0x1a0/0x1a0
[  475.115083]  mlx4_en_poll_rx_cq+0x59/0x100 [mlx4_en]
[  475.120504]  net_rx_action+0x28e/0x3f0
[  475.124700]  __do_softirq+0x10f/0x2a8
[  475.128788]  run_ksoftirqd+0x1c/0x40
[  475.133039]  smpboot_thread_fn+0x10e/0x160
[  475.137540]  kthread+0xff/0x130
[  475.141071]  ? sort_range+0x20/0x20
[  475.144937]  ? kthread_create_on_node+0x70/0x70
[  475.150026]  ret_from_fork+0x35/0x40
[  475.153947] Code: 45 e0 40 84 ed 0f 85 8a 00 00 00 48 8b 83 58 01 00 00 48 8d 93 58 01 00 00 8b bb 50 04 00 00 48 39 d0 ba 00 00 00 00 48 0f 44 c2 <48> 8b 68 10 e8 a5 af b1 ff 2b 83 c0 05 00 00 8d 3c 28 85 ff 74 
[  475.173531] RIP: tcp_schedule_loss_probe+0xc2/0x170 RSP: ffffa3790319b990
[  475.180650] CR2: 0000000000000010
```

Can be related to #1431

avbelov23:
Fix #1431
Fix #1435 
Close tls connection after encryption is complete, if it is a fatal alert or close notify

### DIFF
--- a/tempesta_fw/tls.c
+++ b/tempesta_fw/tls.c
@@ -515,6 +515,13 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
 	for (p = pages; p < pages_end; ++p)
 		put_page(*p);
 
+	if (type == TTLS_MSG_ALERT &&
+	    (io->alert[1] == TTLS_ALERT_MSG_CLOSE_NOTIFY ||
+	     io->alert[0] == TTLS_ALERT_LEVEL_FATAL))
+	{
+		ss_close(sk, SS_F_SYNC);
+	}
+
 out:
 	if (unlikely(sgt.nents > AUTO_SEGS_N))
 		kfree(sgt.sgl);
@@ -528,7 +535,7 @@ out:
  * initiates a record transmission, e.g. alert or a handshake message.
  */
 static int
-tfw_tls_send(TlsCtx *tls, struct sg_table *sgt, bool close)
+tfw_tls_send(TlsCtx *tls, struct sg_table *sgt)
 {
 	int r, flags = 0;
 	TfwTlsConn *conn = container_of(tls, TfwTlsConn, tls);
@@ -586,10 +593,6 @@ tfw_tls_send(TlsCtx *tls, struct sg_table *sgt, bool close)
 			       sg_virt(sg), sg->length, sg->page_link & 0x3,
 			       skb, skb->len);
 		}
-	}
-	if (close) {
-		flags |= SS_F_CONN_CLOSE;
-		TFW_CONN_TYPE(&conn->cli_conn) |= Conn_Stop;
 	}
 	if (ttls_xfrm_need_encrypt(tls))
 		flags |= SS_SKB_TYPE2F(io->msgtype) | SS_F_ENCRYPT;
--- a/tls/ttls.c
+++ b/tls/ttls.c
@@ -1038,24 +1038,24 @@ __ttls_add_record(TlsCtx *tls, struct sg_table *sgt, int sg_i,
 }
 
 int
-__ttls_send_record(TlsCtx *tls, struct sg_table *sgt, bool close)
+__ttls_send_record(TlsCtx *tls, struct sg_table *sgt)
 {
 	int r;
 
-	if ((r = ttls_send_cb(tls, sgt, close)))
+	if ((r = ttls_send_cb(tls, sgt)))
 		T_DBG("TLS send callback error %d\n", r);
 	return r;
 }
 
 static int
-ttls_write_record(TlsCtx *tls, struct sg_table *sgt, bool close)
+ttls_write_record(TlsCtx *tls, struct sg_table *sgt)
 {
 	/* Change __ttls_add_record() call if you need it for handshakes. */
 	WARN_ON_ONCE(tls->io_out.msgtype == TTLS_MSG_HANDSHAKE);
 
 	__ttls_add_record(tls, NULL, 0, NULL);
 
-	return __ttls_send_record(tls, sgt, close);
+	return __ttls_send_record(tls, sgt);
 }
 
 static int
@@ -1336,7 +1336,6 @@ ttls_handle_alert(TlsCtx *tls)
 int
 ttls_send_alert(TlsCtx *tls, unsigned char lvl, unsigned char msg)
 {
-	bool close = false;
 	TlsIOCtx *io = &tls->io_out;
 
 	T_DBG("send alert level=%u message=%u\n", lvl, msg);
@@ -1348,10 +1347,7 @@ ttls_send_alert(TlsCtx *tls, unsigned char lvl, unsigned char msg)
 	io->alert[0] = lvl;
 	io->alert[1] = msg;
 
-	if (msg == TTLS_ALERT_MSG_CLOSE_NOTIFY || lvl == TTLS_ALERT_LEVEL_FATAL)
-		close = true;
-
-	return ttls_write_record(tls, NULL, close);
+	return ttls_write_record(tls, NULL);
 }
 
 int
--- a/tls/tls_internal.h
+++ b/tls/tls_internal.h
@@ -216,7 +216,7 @@ int ttls_derive_keys(TlsCtx *tls);
 
 void __ttls_add_record(TlsCtx *tls, struct sg_table *sgt, int sg_i,
 		       unsigned char *hdr_buf);
-int __ttls_send_record(TlsCtx *tls, struct sg_table *sgt, bool close);
+int __ttls_send_record(TlsCtx *tls, struct sg_table *sgt);
 int ttls_sendmsg(TlsCtx *tls, const char *buf, size_t len);
 
 int ttls_parse_certificate(TlsCtx *tls, unsigned char *buf, size_t len,
--- a/tls/tls_srv.c
+++ b/tls/tls_srv.c
@@ -2088,7 +2088,7 @@ ttls_handshake_send_out_buffers(TlsCtx *tls, struct sg_table *sgt,
 	*pg = NULL;
 	sg_mark_end(&sgt->sgl[sgt->nents - 1]);
 	/* Exit, enter the FSM on more data from the client. */
-	return __ttls_send_record(tls, sgt, false);
+	return __ttls_send_record(tls, sgt);
 }
 
 /**
--- a/tls/ttls.h
+++ b/tls/ttls.h
@@ -565,7 +565,7 @@ typedef struct ttls_context {
 	char			*hostname;
 } TlsCtx;
 
-typedef int ttls_send_cb_t(TlsCtx *tls, struct sg_table *sgt, bool close);
+typedef int ttls_send_cb_t(TlsCtx *tls, struct sg_table *sgt);
 typedef int ttls_sni_cb_t(TlsCtx *tls, const unsigned char *data, size_t len);
 typedef unsigned long ttls_cli_id_t(TlsCtx *tls, unsigned long hash);
 


ikoveshnikov (changes requested):
I totally disagree with the proposed changes. Please take a look to analysis  in https://github.com/tempesta-tech/tempesta/issues/1431#issuecomment-728981779  . You're working with the scenario described there, but you're not checking object lifetimes. 

We have a race between lazy encryption in `tfw_tls_encrypt()` and `TfwTlsConn` life time. We need to carefully extend it and keep the `TfwTlsConn` object existing until all the scheduled data was send. I have a couple solutions in mind, like grab an extra reference to TLS  connections and drop it after some period of inactivity, but all of them have draw back and issues with leaking references. In the proposed patch a `TfwTlsConn` object  lifetime is increased, but now client controls, when the object is destroyed, this adds a possible DDoS scenario, which we can not allow.

tempesta_fw/tls.c
```diff
@@ -482,6 +482,13 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int limit)
 	for (p = pages; p < pages_end; ++p)
 		put_page(*p);
 
+	if (type == TTLS_MSG_ALERT &&
+	    (io->alert[1] == TTLS_ALERT_MSG_CLOSE_NOTIFY ||
+	     io->alert[0] == TTLS_ALERT_LEVEL_FATAL))
+	{
+		ss_close(sk, SS_F_SYNC);
```
ikoveshnikov, line 489: We've discussed in chat about different problems related to reference counting, and I happy to see a solution without that caveats. At least it looks so for me now. Please check all the cases we have discussed in the chat, but I think we won't have any problems there.

An optimisation is possible here, but it's quite tricky. And can be done as enhancement later.  `ss_close()` adds a task into work_queue, calls `connection_drop()` hook and invokes `ss_do_close()` which will create a new skb with TCP_FIN flag. Instead, we can move some logic from `ss_do_close` right here, and append TCP_FIN to the alert directly.  Then we need to somehow invoke `connection_drop` or `connection_error` callback as `ss_close` does.

UPD: I see one possible issue. Imagine, there are a lot of requests in processing state, and corresponding responses are continuously added to the write_queue. Then an error happens on receiving path, and close notify is inserted between responses in write_queue. So we call `ss_close()` here, which schedules connection close. `TCP_FIN` flag will be added to the last skb in the queue - response after fatal alert. And we will send a little more data, than we wanted. Can't call it critical issue, though. With the enhancement above such befaviour can be coped.

ikoveshnikov (approved):
Looks good for me, but it's very subtile fix, need to test it thoroughly. Please try to make tests upon scenarios we've discussed, or at least try to reproduce them manually and add tasks for appropriate tests. Also review from other teamamates are required.

tls/ttls.c
```diff
@@ -1335,10 +1334,7 @@ ttls_send_alert(TlsCtx *tls, unsigned char lvl, unsigned char msg)
 	io->alert[0] = lvl;
 	io->alert[1] = msg;
 
-	if (msg == TTLS_ALERT_MSG_CLOSE_NOTIFY || lvl == TTLS_ALERT_LEVEL_FATAL)
-		close = true;
-
-	return ttls_write_record(tls, NULL, close);
+	return ttls_write_record(tls, NULL, false);
```
krizhanovsky, line 1337: This is the only place calling `ttls_write_record()`, so we don't need the boolean argument. `__ttls_send_record()` called by `ttls_write_record()` seems also don't need the last argument as well. I didn't check deeper, but please check other calls for the boolean argument.

tempesta_fw/tls.c
krizhanovsky: I agree with @ikoveshnikov and the solution is really elegant . It seems there are actually not so much code to write to mark the `skb` with `FIN` and move the socket to the closing state. Just please make it in a function in `sock.c`. It seems `skb && tcp_send_head(sk)` is always true in this function, but the rest of `ss_do_close()` should be the same, so probably `ss_do_close()` can be reworked to use the same subfunction for the new function plus the code with an skb allocation.

krizhanovsky (changes requested):
It seems more work is required here

i-rinat (approved):
Looks OK.

krizhanovsky:
One more notice about current `ss_do_close()` implementation. At the beginning of the function we just free all peer's data with
```C
        while ((skb = __skb_dequeue(&sk->sk_receive_queue)) != NULL) {
                u32 len = TCP_SKB_CB(skb)->end_seq - TCP_SKB_CB(skb)->seq -            
                          tcp_hdr(skb)->fin;                                           
                data_was_unread += len;
                T_DBG3("[%d]: free rcv skb %p\n", smp_processor_id(), skb);            
                __kfree_skb(skb);                                                      
        }
```
and this isn't what TCP is supposed to do on active closing: we have to be able to read data from the peer. For example, if we forward an HTTP request with `Connection: close` to the server, we should not drop the response. We also need to read the peer TLS close notify to avoid [TLS truncation attacks](https://hal.inria.fr/hal-01102259/file/triple-handshakes-and-cookie-cutters-oakland14.pdf).

krizhanovsky (approved):
Good to merge, but please do the cleanups.

avbelov23:
Using `ss_close_force()` after the end of encryption (directly close a socket from `tcp_write_xmit()`) doesn't solve the problems.
In `tcp_write_xmit()`, at the end, the sending of the loss probe segment is rescheduled, and in the case of a direct `ss_do_close()` we clear all queues via `inet_csk_destroy_sock()` and crashes, because reschedule uses the cleared write queue (`tcp_schedule_loss_probe()`->`tcp_rto_delta_us()`->`tcp_write_queue_head()`).
I think `ss_close_force()` is too rude and I think it is worth using `ss_close()`.

krizhanovsky:
I believe that the crash with `tcp_schedule_loss_probe()` is essentially what I'm working on in https://github.com/tempesta-tech/tempesta/issues/1435#issuecomment-761874121 - please confirm that you have `tcp_tasklet_fun()` on the stack.

krizhanovsky (approved):
LGTM. The crash is to be fixed in #1435

PR #1479, opened 2020-12-12
Fix tls.test_tls_cert.RSA512_SHA256 test

krizhanovsky:
Print warning and exit with error if a user tries to load an RSA certificate with a key smaller than 1024 bits.

Do not check ctx->padding since we check it before the function calls.

Small coding style cleanups.

Note that `__rsa_setup_ctx()` during the `for_each_possible_cpu()` loop uses current CPU stack for all the big integer allocations, including `ttls_mpi_inv_mod()` and `ttls_mpi_exp_mod()`, and that's OK since the function is called on configuration time only and the stack allocations are used for temporal data only.

### DIFF
--- a/tls/rsa.c
+++ b/tls/rsa.c
@@ -101,7 +101,7 @@ ttls_rsa_init(TlsRSACtx *ctx, int padding, int hash_id)
 static int
 __rsa_setup_ctx(TlsRSACtx *ctx)
 {
-	int cpu, count = 0;
+	int r, cpu;
 
 	/*
 	 * Do nothing if the context is already setup or N or E aren't loaded
@@ -134,6 +134,8 @@ __rsa_setup_ctx(TlsRSACtx *ctx)
 	 * Unblinding value: Vf = random number, invertible mod N.
 	 */
 	for_each_possible_cpu(cpu) {
+		int count = 0;
+
 		TlsMpi *vi = per_cpu_ptr(ctx->Vi, cpu);
 		TlsMpi *vf = per_cpu_ptr(ctx->Vf, cpu);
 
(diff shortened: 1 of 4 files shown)


tls/rsa.c
```diff
@@ -148,10 +150,16 @@ __rsa_setup_ctx(TlsRSACtx *ctx)
 		} while (ttls_mpi_cmp_int(vi, 1));
 		/* Blinding value: Vi =  Vf^(-e) mod N */
 		ttls_mpi_inv_mod(vi, vf, &ctx->N);
-		ttls_mpi_exp_mod(vi, vi, &ctx->E, &ctx->N, &ctx->RN);
+		if ((r = ttls_mpi_exp_mod(vi, vi, &ctx->E, &ctx->N, &ctx->RN)))
```
ikoveshnikov, line 153: For many similar places a `TTLS_MPI_CHK()` macro is used. Is it going to be deprecated or it was just lost here?

tls/rsa.c
```diff
@@ -148,10 +150,16 @@ __rsa_setup_ctx(TlsRSACtx *ctx)
 		} while (ttls_mpi_cmp_int(vi, 1));
```
ikoveshnikov, line 150: The cycle above is limited to 10 cpus. Do we actually need a limit. Some modern processors have more cores.

ikoveshnikov (approved):
LGTM

i-rinat (approved):
Looks good to me.

tls/rsa.c
i-rinat: I can't see the limit of 10 tries over all available CPUs as anything other than a mistake (typo?). It will definitely trigger the `WARN_ON_ONCE()` on a machine with more than 10 CPUs.
As this patch now resets `count` for each CPU, it's now 10 tries per CPU, which looks fine. I think this is OK.

krizhanovsky: Yeah, Rinat is right: this is 10 tries per each CPU. Previously, on 32 vCPUs. 10 tries wasn't enough for all of them and I observed the warning.

tls/rsa.c
krizhanovsky: Franky, I've never liked the macros and this is mostly why they use WARN statements. Most of the errors from mbed TLS are about failed memory allocations, but we should not do any memory allocations in fast path, so the code is going to have more void-returning calls and there won't be so many calls checking the status. In general, the macros are needed to handle many calls returning some error codes and the conditions aren't good for performance. So I'd prefer to feel pain writing the code with many barnches.

PR #1484, opened 2021-01-13
Fix invalid error messages if js_challenge points to invalid file

### LINKED ISSUE #1442: JSCH Template Issue
## Scope

With configuration:
```
listen 192.168.100.4:80;

frang_limits {
	# Conenction level limits.
	client_header_timeout 20;
	client_body_timeout 10;
	http_header_chunk_cnt 10;
	http_body_chunk_cnt 0;

	# Message level limits (per-vhost)
	http_methods GET;
	http_uri_len 512;
	http_resp_code_block 400 403 404 3 10;
}

# Allow only following characters in URI (no '%'): /a-zA-Z0-9&?:-._=
http_uri_brange 0x2f 0x41-0x5a 0x61-0x7a 0x30-0x39 0x26 0x3f 0x3a 0x2d 0x2e 0x5f 0x3d;

sticky {
	cookie name=my_js_cookie enforce max_misses=3 timeout=3;
	js_challenge resp_code=503 delay_min=1000 delay_range=1000 delay_limit=3000 /root/tempesta/etc/js_challenge.html;
}

block_action attack reply;
block_action error reply;

srv_group default {
	server 127.0.0.1:8080 conns_n=4;
}

#vhost debian {
vhost default {
	tls_certificate /root/tempesta/etc/tfw-root.crt;
	tls_certificate_key /root/tempesta/etc/tfw-root.key;

	resp_hdr_set Strict-Transport-Security "max-age=31536000; includeSubDomains";

	proxy_pass default;
}

cache 0;
#cache_fulfill * *;

http_chain {
	#hdr "Referer" == "http://badhost.com*" -> block;

	#-> debian;
	-> default;
}
```

I get perl error:
```
Starting Tempesta...
...enable RPS on enp3s0
...load Tempesta modules
Loading Tempesta kernel modules...
Loading module tempesta_lib 
Loading module tempesta_tls 
Loading module tempesta_db 
Loading module tempesta_fw tfw_cfg_path=/home/user/qtc/last-build/etc/tempesta_fw.conf
...compile html templates for JS challenge
Use of uninitialized value $src in -e at /home/user/qtc/last-build/scripts/update_template.pl line 37.
Use of uninitialized value $src in concatenation (.) or string at /home/user/qtc/last-build/scripts/update_template.pl line 38.
JavaScript Challenge template '' not found, generation aborted! at /home/user/qtc/last-build/scripts/update_template.pl line 38.
...start Tempesta FW
Un-loading Tempesta kernel modules...
ERROR: cannot start Tempesta FW (sysctl code: Invalid argument)
```

I have no file `/root/tempesta/etc/js_challenge*` on my VM.

## Testing

Check what cause the problem add add the test to JSCH functional tests.

ikoveshnikov:
fix #1442

### DIFF
--- a/scripts/update_template.pl
+++ b/scripts/update_template.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/env perl
 #
-# Copyright (C) 2017-2020 Tempesta Technologies, Inc.
+# Copyright (C) 2017-2021 Tempesta Technologies, Inc.
 #
 # This program is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License as published by
@@ -29,13 +29,16 @@ if (!$template || $template !~ '.tpl$' || !$sticky_name
 {
 	die "bad args!\n"
 }
+unless (-e $template && -r $template) {
+	die "JavaScript Challenge template '$template' not found, generation aborted!\n";
+}
 
 # Assemble HTML templates and minify resulting files.
 sub assemble
 {
 	my ($src) = @_;
 	unless (-e $src) {
-		die "JavaScript Challenge template '$src' not found, generation aborted!";
+		die "JavaScript Challenge template '$src' not found, generation aborted!\n";
 	}
 
 	my $dir = dirname($src);
--- a/scripts/tempesta.sh
+++ b/scripts/tempesta.sh
@@ -22,11 +22,12 @@
 if [ -z "$TFW_SYSTEMD" ]; then
 	if [ "${TEMPESTA_LCK}" != "$0" ]; then
 		env TEMPESTA_LCK="$0" flock -n -E 254 "/tmp/tempesta-lock-file" "$0" "$@"
-		if [ $? -eq 254 ]; then
+		ret=$?
+		if [ $ret -eq 254 ]; then
 			echo "Cannot operate with Tempesta FW: locked by another process"
 			exit 3
 		fi
-		exit
+		exit $ret
 	fi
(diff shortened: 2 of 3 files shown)


scripts/update_template.pl
```diff
@@ -35,7 +35,7 @@ sub assemble
 {
 	my ($src) = @_;
 	unless (-e $src) {
-		die "JavaScript Challenge template '$src' not found, generation aborted!";
+		die "JavaScript Challenge template '$src' not found, generation aborted!\n";
```
krizhanovsky, line 38: `die` adds line number and LF, so no need for `\n`

scripts/update_template.pl
```diff
@@ -58,6 +58,9 @@ sub assemble
 		|| die $t->error(), "\n";
 }
 
+unless (-e $template && -r $template) {
+	die "JavaScript Challenge template '$template' not found, generation aborted!\n";
+}
```
krizhanovsky, line 63: I think it's better to move the check higher, where we check all the arguments from `@ARGV`

krizhanovsky (approved):
LGTM

scripts/update_template.pl
ikoveshnikov: `die` adds not only LF, but also a trace to faulty file and line. This information is omitted if explicit `\n` was stated. For user the trace message looks like internal error in our scripts, but it actually configuration error.

PR #1492, opened 2021-02-01
Fixes and tools to fight OOM

krizhanovsky:
The PR is in progress.
* skmem.pl script to track HTTP(S) sockets memory usage
* Properly account sk_forward_alloc.
* Add couple of unrelated comments

### DIFF
--- a/scripts/skmem.pl
+++ b/scripts/skmem.pl
@@ -1,0 +1,55 @@
+#!/usr/bin/perl
+#
+# Copyright (C) 2021 Tempesta Technologies, Inc.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License,
+# or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 59
+# Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+use 5.16.0;
+use strict;
+use warnings;
+
+my ($rmem, $rcv_buf, $wmem, $snd_buf, $fwd_alloc) = (0, 0, 0, 0, 0);
+my ($wmem_q, $ropt_mem, $back_log, $sock_drop) = (0, 0, 0, 0);
+
+my @ss = `ss -tm '( dport = :https or sport = :https )' | grep skmem`;
+foreach (@ss) {
+	/\s+skmem:\(r(\d+),rb(\d+),t(\d+),tb(\d+),f(\d+),w(\d+),o(\d+),bl(\d+),
+	 d(\d+)\)/x;
+
+	$rmem += $1;
+	$rcv_buf += $2;
+	$wmem += $3;
+	$snd_buf += $4;
+	$fwd_alloc += $5;
+	$wmem_q += $6;
+	$ropt_mem += $7;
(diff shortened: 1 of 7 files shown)


tempesta_fw/sock.c
```diff
@@ -340,6 +340,23 @@ ss_sock_active(struct sock *sk)
  *  	Server and client connections handling
  * ------------------------------------------------------------------------
  */
+/**
+ * The simplified version of sk_forced_mem_schedule().
+ * We use the forced version of the socket accounting function to send the
+ * data ASAP - we can't wait for memory as process context does.
```
i-rinat, line 346: > we can't wait for memory

As far as I understand, the only difference from `sk_forced_mem_schedule()` is that latter also calls `mem_cgroup_charge_skmem()`. Looks like `mem_cgroup_charge_skmem()` is aware of being called from softirq context:
```c
        /* Don't block in the packet receive path */
        if (in_softirq())
                gfp_mask = GFP_NOWAIT;
```
So it looks like waiting is not an issue?

tls/ttls.c
```diff
@@ -1702,6 +1702,11 @@ ttls_write_change_cipher_spec(TlsCtx *tls, struct sg_table *sgt,
 	}
 }
 
+/**
+ * Do not allow to change the cipher spec actually to prevent CCS injection
```
i-rinat, line 1706: Is it a TODO note, or is it a description of the function that follows?
If it's a TODO, "TODO" string is missing.
If it's a description, then maybe it's better be where `ttls_parse_change_cipher_spec()` is called from. An FSM there calls `ttls_parse_change_cipher_spec()` once per handshake message, so that's where protection against CCS injection is.

i-rinat (approved):
Looks good overall.
Some thoughts I had while reading the code are in the comments below.

tempesta_fw/sock.c
krizhanovsky: Yeah. Firstly, the function is quite small and I though that it makes sense just to copy it instead of to patch the kernel with the function exporting. Also the function doesn't make the unnecessary `mem_cgroup_charge_skmem()` call - not a big deal to call it, but all in all it makes sense to make the short simplified copy of the function.

tls/ttls.c
krizhanovsky: I updated the comment to make it clear. The thing is that the function doesn't changes the cipher specification as it is supposed to do, instead it just validates the input data. This is inherited from mbed TLS, but this comment just explains why the code is written in this way.

tempesta_fw/sock.c
i-rinat: Oh, right. I missed that `sk_forced_mem_schedule()` is not exported.

scripts/skmem.pl
```diff
@@ -0,0 +1,56 @@
...
+# Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+use 5.16.0;
+use strict;
+use warnings;
+
+my ($rmem, $rcv_buf, $wmem, $snd_buf, $fwd_alloc) = (0, 0, 0, 0, 0);
+my ($wmem_q, $ropt_mem, $back_log, $sock_drop) = (0, 0, 0, 0);
+
+# See linux/net/ipv4/inet_diag.c for the fields reporting.
+my @ss = `ss -tm '( dport = :https or sport = :https )' | grep skmem`;
```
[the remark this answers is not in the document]
i-rinat: I use a custom-built kernel with a config that leans towards minimal working configuration (to speed-up rebuild times). It lacked at least CONFIG_INET_DIAG, CONFIG_INET_TCP_DIAG, and CONFIG_NETLINK_DIAG, so `ss` used `/proc/net/tcp` rather than getting information over a netlink socket.

With rebuilt kernel that now have mentioned config options enabled, `ss` shows `skmem:` lines in the VM too.

tls/ttls.c
i-rinat: That makes sense.

PR #1493, opened 2021-02-20
The effect of block_action on the absence of a suitable vhost

### LINKED ISSUE #1453: HTTPtables blocking action ignores block_action directive
# Scope

On configuration
```
listen 192.168.100.4:80;

block_action attack reply;
block_action error reply;
response_body 400 /var/www/tempesta-tech.com/400;

srv_group default {
        server 127.0.0.1:9090;
}

vhost debian {
        proxy_pass default;
}

cache 1;
cache_fulfill * *;

http_chain l7_rules {
        hdr "Referer" == "http://badhost.com*" -> block;
}

http_chain {
        mark == 1 -> l7_rules;
        -> debian;
}
```
Note that there is no default host as well no default rule for `l7_rules` chain. Configure iptables as:
```
iptables -A INPUT -s 192.168.100.1 -j MARK --set-mark 1
```
Next try to send 2 requests from `192.168.100.1`:
1. with `Referer: http://goodhost.com`
2. with `Referer: http://badhost.com/foo`

In both the cases HTTPtables block the requests, but Tempesta does not send any HTTP response as expected. From browsing the code it seems we just ignore `block` argument returned from HTTPtables.

# Testing

For the scenario and configuration as the above:
- [ ] request with `Referer: http://goodhost.com`
- [ ] request with `Referer: http://badhost.com/foo`

- [ ] Check that configuration option `response_body` also does the work. In my demo scenario I used
```
block_action attack reply;
block_action error reply;
response_body 400 /var/www/tempesta-tech.com/400;
```

Please also add tests for the configuration described in https://github.com/tempesta-tech/tempesta/wiki/HTTP-tables#default-rules-and-chains to `./http_rules/test_http_tables.py`
- [ ] Drop a request if there is no matching rule in a chain
- [ ] implicit creation of the main chain
- [ ] Dropping all requests with an empy main chain

avbelov23:
Fix #1453

### DIFF
--- a/tempesta_fw/http.c
+++ b/tempesta_fw/http.c
@@ -5402,7 +5402,7 @@ next_msg:
 	 * error response to client and move on to the next request.
 	 */
 	else if (unlikely(!req->vhost)) {
-		tfw_http_send_resp(req, 404,
+		tfw_http_req_block(req, 403,
 				   "request dropped: cannot find appropriate "
 				   "virtual host");
 		TFW_INC_STAT_BH(clnt.msgs_otherr);


krizhanovsky (approved):
LGTM

PR #1496, opened 2021-03-03
Do expand transformation if add or inplace transformation fail

avbelov23:


tempesta_fw/hpack.c
```diff
@@ -3697,6 +3697,7 @@ tfw_hpack_encode(TfwHttpResp *__restrict resp, TfwStr *__restrict hdr,
 	TfwH2Ctx *ctx = tfw_h2_context(resp->req->conn);
```
krizhanovsky, line 3697: The patch may affect performance, so please add following comment to the header comment of the function:
```
While we do our best to rewrite the HTTP message in-place, sometimes the final HTTP/2 message may be bigger than the original HTTP/1 one, so we have to fall back from TFW_H2_TRANS_ADD, TFW_H2_TRANS_SUB, or TFW_H2_TRANS_INPLACE to TFW_H2_TRANS_EXPAND operation. This basically happen in 2 cases:
1. HTTP/2 overhead is too large in comparison with the original data, which can not be significantly compressed, i.e. with too short original message. This basically artificial and usually the case for benchmarks;
2. too much data is added, e.g. with resp_hdr_add configuration option.
The first case isn't so important and basically the 'false try' is cheap because the traversed data is small. The second one might be a problem. TODO #1103: revise the logic.
```

krizhanovsky (changes requested):
There are several TODOs. Also please unmask the appropriate test  https://github.com/tempesta-tech/tempesta-test/pull/182/files/5c5f6f771661ff9af17243fd24d629aa7df02da2..9070a3e26c3c9b8293679052f3a10f33ab324072

krizhanovsky (approved):
LGTM

Issue #1475, opened 2020-12-10
Sporadic failures on several functional tests


krizhanovsky:
The test fails sporadically. On current master as of 737113eabd593ee082bb1a0016196ee02eee5534
```
# ./run_tests.py sched.test_per_group_lb.RedefineGlobalSched.test_inheritLoading disabled file
....

----------------------------------------------------------------------
Running functional tests (resuming from after sched.test_per_group_lb.RedefineGlobalSched.test_inherit)...
----------------------------------------------------------------------


----------------------------------------------------------------------
Ran 0 tests in 0.000s

OK

# ./run_tests.py sched.test_per_group_lb.RedefineGlobalSched.test_inherit
...........

----------------------------------------------------------------------
Running functional tests...
----------------------------------------------------------------------

test_inherit (sched.test_per_group_lb.RedefineGlobalSched) ... 	Init test case...
	Starting Nginx on nginx_8001
	Starting Nginx on nginx_8000
	Starting Nginx on nginx_8003
	Starting Nginx on nginx_8002
	Starting Nginx on nginx_8005
	Starting Nginx on nginx_8004
	Starting Nginx on nginx_8007
	Starting Nginx on nginx_8006
	Starting Nginx on nginx_8009
	Starting Nginx on nginx_8008
	Starting TempestaFW on localhost
Running client
Running client
	Client exit
	Client exit
Server group custom uses 'ratio dynamic' scheduler
FAIL
	Teardown
Stopping client
	client stdout:
Running 10s test @ http://127.0.0.1:80/
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    20.68ms   10.80ms  60.08ms   56.65%
    Req/Sec     2.40k    90.40     2.60k    90.50%
  47807 requests in 10.07s, 11.81MB read
Requests/sec:   4747.35
Transfer/sec:      1.17MB
---- RESULTS --------
Status 200 : 47807 times
---- END ------------

Client is stopped
Stopping client
	client stdout:
Running 10s test @ http://127.0.0.1:80/
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    75.65ms   33.03ms 399.77ms   92.70%
    Req/Sec   680.57     60.64   777.00     93.00%
  13561 requests in 10.05s, 3.35MB read
Requests/sec:   1349.85
Transfer/sec:    341.42KB
---- RESULTS --------
[...]

PR #1497, opened 2021-03-11
Handling the case if data is inserted at the end of a fragment and after that there is no other fragment that starts after it 

avbelov23:
#1475
Fix for msg_modification.* tests

If, when inserting/deleting data at the end of the fragment, there is no other fragment that starts immediately after that, then we will not be able to find the fragment to insert/delete data, and we need to add/delete data after the fragment at the end of which we wanted to insert.

Handling cases of looking for space in skbs to add/delete data larger than one chunk

### DIFF
--- a/tempesta_fw/ss_skb.c
+++ b/tempesta_fw/ss_skb.c
@@ -752,7 +752,7 @@ static int
 __skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
 	       int len, TfwStr *it, int *fragn)
 {
-	int i = -1, ret;
+	int i = -1, j = -1, ret;
 	long offset;
 	unsigned int d_size;
 	struct skb_shared_info *si = skb_shinfo(skb);
@@ -816,6 +816,30 @@ __skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
 			ret = __split_pgfrag(skb_head, skb, i, offset, len, it, fragn);
 			goto done;
 		}
+		/*
+		 * In case another fragment does not exist, remember the
+		 * fragment number, after which you need to insert/delete data.
+		 */
+		if (offset == d_size)
+			j = i;
+	}
+
+	if (unlikely(j >= 0)) {
+		if (len > 0) {
+			if (!(ret = __new_pgfrag(skb_head, skb, len, j + 1, 1)))
+				__it_next_data(skb, j + 1, it, fragn);
+			goto done;
+		}
+
+		if (j == si->nr_frags - 1) {
+			T_WARN("Cannot delete bytes from skb\n");
+			ret = -EFAULT;
+			goto done;
+		}
+
+		ret = __split_pgfrag_del_w_frag(skb_head, skb, j + 1, 0, -len,
+						it, fragn);
+		goto done;
 	}
 
 	/* The split is not within the SKB. */


tempesta_fw/ss_skb.c
```diff
@@ -806,12 +806,7 @@ __skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
 		d_size = skb_frag_size(frag);
 		offset = pspt - (char *)skb_frag_address(frag);
 
-		/*
-		 * @pspt can be the end of a frag, but it can also be a start of
-		 * another fragment. Not necessary, that both frags are
-		 * neighbours.
-		 */
-		if (offset >= 0 && offset < d_size) {
+		if (offset >= 0 && offset <= d_size) {
```
krizhanovsky, line 809: The original condition was introduced in 0f04bfd29372ac722f818af3874e1bba3f1ef865 , which also removed the special handling of `offset == d_size`. `offse == d_size` means that `pspt` is just *after* the current fragment, so why should we split the current fragment?

krizhanovsky (changes requested):
The fix touches the very subtle place and it seems that we need a more accurate fix because it also hits the `WARN_ON_ONCE(sgt.nents != frags)` warning in `tfw_tls_encrypt()`.

tempesta_fw/ss_skb.c
avbelov23: Split works like adding a new fragment

tempesta_fw/ss_skb.c
```diff
@@ -816,6 +816,20 @@ __skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
 			ret = __split_pgfrag(skb_head, skb, i, offset, len, it, fragn);
 			goto done;
 		}
+		/*
+		 * In case another fragment does not exist, remember the
+		 * fragment number, after which you need to insert a new one.
+		 */
+		if (offset == d_size) {
+			len = max_t(long, len, offset - d_size);
```
krizhanovsky, line 824: Technically, we can be here with `len < 0`. There is no sense to delete `|len|` bytes after current fragment if there is no fragment right after the current one, but I'd prefer to have at least `BUG_ON()` assertion here or return with `-EINVAL` like:
```C
if (WARN_ON_ONCE(len < 0))
    return -EINVAL;
```

krizhanovsky (approved):
LGTM with small adjustment

tempesta_fw/ss_skb.c
```diff
@@ -816,6 +816,50 @@ __skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
...
+
+	if (j >= 0) {
+		skb_frag_t *frag = &si->frags[j + 1];
+
+		if (len > 0) {
+			ret = __new_pgfrag(skb_head, skb, len, j + 1, 1);
+			__it_next_data(skb, j + 1, it, fragn);
+			goto done;
+		}
+
```
krizhanovsky, line 835: When we can have the negative `len` and `offset == d_size` above? The case looks wrong for me.

I tried to run your previous code version with the small patch
```
@@ -821,7 +821,7 @@ __skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
 		 * fragment number, after which you need to insert a new one.
 		 */
 		if (offset == d_size) {
-			len = max_t(long, len, offset - d_size);
+			BUG_ON(len < 0);
 			j = i;
 		}
 	}
```
It doesn't hit the assertion, but it seems there is some memory leak - the full functional tests suite fails with OOM while it passes on the master as of today.

krizhanovsky (changes requested):
It seems more work and/or discussion is required.

tempesta_fw/ss_skb.c
```diff
@@ -816,6 +816,50 @@ __skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
...
+		/*
+		 * In case another fragment does not exist, remember the
+		 * fragment number, after which you need to insert/delete data.
+		 */
+		if (offset == d_size)
+			j = i;
+	}
+
+	if (j >= 0) {
+		skb_frag_t *frag = &si->frags[j + 1];
```
krizhanovsky, line 828: We can get here with j == si->nr_frags - 1, so this could be an out of bound access here and in all the statements till the end of the if statement. Here we can check the `j` value and increment it to deal with `j` till the end of the if statement.

tempesta_fw/ss_skb.c
```diff
@@ -816,6 +816,50 @@ __skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
...
+
+		/* delete the head part of a fragment. */
+		frag->page_offset -= len;
+		skb_frag_size_sub(frag, -len);
+		skb->len += len;
+		skb->data_len += len;
+		it->data = skb_frag_address(frag);
+		it->skb = skb;
+		*fragn = j + 1;
+		goto done;
```
krizhanovsky, line 862: It seems this just equals to `__split_pgfrag_del_w_frag(skb_head, skb, j + 1, 0, -len, it, fragn)`

tempesta_fw/http_msg.c
```diff
@@ -687,25 +687,58 @@ static int
 __hdr_add(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
 {
```
krizhanovsky, line 688: I propose to get rid of the changes introduced by the commit. It's unclear whether do we need to mangle headers longer than 4KB and this is definitely not in the scope of #1475, but I just caught Oops on this branch:
```
[181764.142253] BUG: unable to handle kernel paging request at ffff9e2c1ae3f840
[181764.145584] IP: tfw_msg_iter_move+0x23/0x130 [tempesta_fw]
[181764.148306] PGD 6c126067 P4D 6c126067 PUD 0
[181764.150899] Oops: 0000 [#1] SMP PTI
[181764.152061] Modules linked in: tempesta_fw(O) tempesta_db(O) tempesta_tls(O) tempesta_lib(O) tcp_diag inet_diag xt_tcpudp nft_counter nft_meta nft_chain_route_ipv4 nf_tables_ipv4 xt_mark nft_compat nf_tables nfnetlink bochs_drm ttm drm_kms_helper crct10dif_pclmul drm ppdev fb_sys_fops syscopyarea intel_rapl_perf sg sysfillrect sysimgblt serio_raw parport_pc parport button binfmt_misc ip_tables x_tables ext4 crc16 mbcache jbd2 fscrypto sr_mod sd_mod cdrom ata_generic virtio_net ata_piix libata psmouse virtio_pci virtio_ring virtio i2c_piix4 scsi_mod [last unloaded: tempesta_lib]
[181764.152061] CPU: 1 PID: 12682 Comm: python2 Tainted: G        W  O    4.14.32-kdump+ #217
[181764.152061] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.13.0-3.fc33 04/01/2014
[181764.152061] task: ffff9e31b2c426c0 task.stack: ffffbc648425c000
[181764.152061] RIP: 0010:tfw_msg_iter_move+0x23/0x130 [tempesta_fw]
[181764.152061] RSP: 0018:ffff9e31bfd035f8 EFLAGS: 00010283
[181764.152061] RAX: ffff9e31b2deecc0 RBX: 0000000000000000 RCX: ffff9e2c1ae3f840
[181764.152061] RDX: 0000000000000000 RSI: ffff9e31bfd03618 RDI: ffff9e31bfd03620
[181764.152061] RBP: ffff9e317dbbd020 R08: ffff9e31b2deeb00 R09: ffff9e31b2deea00
[181764.152061] R10: 000000000000000f R11: ffff9e31bfd23620 R12: ffffffffc05a2927
[181764.152061] R13: ffff9e31bfd037d0 R14: ffff9e31bfd03810 R15: 000000000000000f
[181764.152061] FS:  00007f6e7e0c4740(0000) GS:ffff9e31bfd00000(0000) knlGS:0000000000000000
[181764.152061] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[181764.152061] CR2: ffff9e2c1ae3f840 CR3: 0000000078890003 CR4: 00000000003606e0
[181764.152061] Call Trace:
[181764.152061]  <IRQ>
[181764.152061]  __hdr_add+0x135/0x360 [tempesta_fw]
[181764.152061]  ? _raw_spin_unlock+0x1f/0x30
[181764.152061]  tfw_http_msg_hdr_xfrm_str+0xa2/0x430 [tempesta_fw]
[181764.152061]  ? tfw_tls_hs_over+0x43/0x60 [tempesta_fw]
[181764.152061]  tfw_http_msg_hdr_xfrm+0xaf/0x100 [tempesta_fw]
[181764.152061]  tfw_http_req_cache_cb+0x305/0x480 [tempesta_fw]
[181764.152061]  ? tfw_http_req_cache_cb+0x2d9/0x480 [tempesta_fw]
[181764.152061]  ? tfw_http_conn_repair+0xa40/0xa40 [tempesta_fw]
[181764.152061]  tfw_cache_process+0x6d/0x290 [tempesta_fw]
[181764.152061]  tfw_http_msg_process_generic+0x4e3/0xce0 [tempesta_fw]
[181764.152061]  ? ttls_decrypt+0x31f/0x610 [tempesta_tls]
[181764.152061]  tfw_http_msg_process+0x7f/0xc0 [tempesta_fw]
[181764.152061]  ? ttls_recv+0xad/0x640 [tempesta_tls]
[181764.152061]  __gfsm_fsm_exec+0x51/0x90 [tempesta_fw]
[181764.152061]  tfw_gfsm_move+0x146/0x180 [tempesta_fw]
[181764.152061]  tfw_tls_msg_process+0x334/0x510 [tempesta_fw]
[181764.152061]  __gfsm_fsm_exec+0x51/0x90 [tempesta_fw]
[181764.152061]  ? consume_skb+0x39/0x120
[181764.152061]  tfw_connection_recv+0x3c/0x60 [tempesta_fw]
[181764.152061]  ? tfw_connection_send+0x20/0x20 [tempesta_fw]
[181764.152061]  ss_tcp_process_data+0x1e5/0x480 [tempesta_fw]
[181764.152061]  ss_tcp_data_ready+0x4d/0xa0 [tempesta_fw]
```

krizhanovsky (changes requested):
Please see the comments

krizhanovsky (approved):
LGTM

PR #1508, opened 2021-06-30
Initialize sk->sk_cgrp_data in ss_sock_create/ss_inet_create

i-rinat:
When CONFIG_CGROUP_BPF is enabled, but CONFIG_CGROUP_NET_PRIO and/or
CONFIG_CGROUP_NET_CLASSID are not, sock_cgroup_ptr() just returns
sk->sk_cgrp_data.val as is, without checking for NULL. Other functions in
the networking stack, say __cgroup_bpf_run_filter_sock_addr(), may then
try to dereference the pointer, causing the kernel to crash. As sk_alloc()
doesn't initialize sk->sk_cgrp_data when called from an interrupt
context, we need to do the initialization work themselves.


krizhanovsky (approved):
LGTM

PR #1509, opened 2021-06-30
update .gitignore 

i-rinat:
Pull-request #1505 renamed `tempesta_db/` to `db/`, so we need to update paths mentioned. And the newer Kbuild version started to create `.mod` and `.cmd` files that need to be ignored too.


krizhanovsky (approved):
LGTM

PR #1510, opened 2021-07-04
Fix unit test failures

i-rinat:
There is a caching in the pool implementation that uses per-cpu arrays which must be allocated. Caches seems to work without initialization too, but they point to the beginning of %gs segment. That may cause false positive of the GCC's stack protector feature, as it relies on `(%gs:0x28)` keeping the same value all the time.


krizhanovsky (approved):
LGTM.

PR #1512, opened 2021-07-07
muffle UBSAN warnings caused by DECLARE_MPI_AUTO()

i-rinat:
DECLARE_MPI_AUTO() declares limbs just above TlsMpi instance on the stack, and subsequent accesses to limbs through MPI_P() make UB sanitizer to notice an access outside a TlsMpi structure instance. Technically it does cause an undefined behavior, but as I can't figure out how to make the same but without UB, we'll have to live with that.

Initializing limbs early in DECLARE_MPI_AUTO_INITLIMBS() makes UB accesses seem to go away. At least for the current UBSAN implementation.


krizhanovsky (approved):
LGTM

Issue #1513, opened 2021-07-27
Setting cache problems


DagothDeer:
os: Debian-9.13.0
tempesta ver.: 0.6.7

1. Setting caching mode equal 1 or 2 not take difference at all to me.

"https://github.com/tempesta-tech/tempesta/commit/0f5003976ee39a860cb6b80bdc31e87830c052f9"

"Tempesta creates per NUMA node database files, so if you have two
processor packages on modern hardware, then follwoing files will be created
(one for earch processor package) for the example above:
        /opt/tempesta/db/cache0.tdb
        /opt/tempesta/db/cache1.tdb"

But no matter what mode i set 1 or 2 in the directory "/opt/tempesta/db" appears only ONE file: "cache0.tdb".
But in the console the string ".../cache.tdb: size=..." appears two times.
![1](https://user-images.githubusercontent.com/33806067/127068853-8a950d07-8eef-429a-8dfd-c65eb8d56110.png)

And only, when i set caching mode equal = 0 these strings are disappear.

2. Multiple of cache_size

In the config file you are write: 
"# SIZE is specified in bytes, suffixes like 'MB' are not supported yet.
     Also, the number must be a multiple of 2MB (Tempesta DB extent size)."

But, in the console error output you program write that size must be a multiple of 4096 (byte?) and NOT 2MB.
![2](https://user-images.githubusercontent.com/33806067/127069937-6c41401b-70d1-4e40-b05c-542486e81cc3.png)

3. Setting cache_size

In the file "tempesta_fw/cache.c" 1108-1117 strings you set limit of range from PAGE_SIZE (4096 byte?) to 1 « 30 (equal 1073741824 byte == 1GB).
![3](https://user-images.githubusercontent.com/33806067/127071036-c476e91e-6082-4e05-a5a9-867b2645af54.png)

At first, 1 GB of cache is too small for store *.JPG, *.PNG, *.MP4 and etc. as you describe in the config and Wiki. If you have hundreds backends and server with 512 GB of memory you exactly want to have cache more than 1 GB. Some servers have more than 6 TB of memory, so limit cache_size with integer value and parse it with function "TfwCfgSpecInt()" is NOT right.

At second, Tempesta FW could not allow to set "cache_size 1073741824;". I can set maximum 470 MB (492830720 byte).
![4](https://user-images.githubusercontent.com/33806067/127074132-b2028126-6dfd-4fd5-b417-82af32f4b03b.png)

[...]

PR #1514, opened 2021-08-08
Handling web cache size configuration

krizhanovsky:
Fixes for #1513:

1. use unsigned long to handle values up to the maximum TDB table size
   of 128GB;
2. require multiplication of 2MB and range [16MB; 128GB].
3. Pick a huge pages memory region if it's exactly of requested size

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -2531,7 +2531,7 @@ static TfwCfgSpec tfw_cache_specs[] = {
 		.dest = &cache_cfg.db_size,
 		.spec_ext = &(TfwCfgSpecInt) {
 			.multiple_of = 2 * 1024 * 1024,
-			.range = { 2 * 1024 * 1024, (1UL << 37) },
+			.range = { 16 * 1024 * 1024, (1UL << 37) },
 		}
 	},
 	{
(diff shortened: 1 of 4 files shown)


fw/cache.c
```diff
@@ -2527,11 +2527,11 @@ static TfwCfgSpec tfw_cache_specs[] = {
...
 		.name = "cache_size",
 		.deflt = "268435456",
-		.handler = tfw_cfg_set_int,
+		.handler = tfw_cfg_set_long,
 		.dest = &cache_cfg.db_size,
 		.spec_ext = &(TfwCfgSpecInt) {
-			.multiple_of = PAGE_SIZE,
-			.range = { PAGE_SIZE, (1 << 30) },
+			.multiple_of = 2 * 1024 * 1024,
+			.range = { 2 * 1024 * 1024, (1UL << 37) },
```
keshonok, line 2534: In the comment of the PR it says that the range is `[16MB; 128Gb]`. It's appears that either the comment is inaccurate or the line of code above it wrong and it should be `{ 16 * 1024 * 1024, (1UL << 37) }`

krizhanovsky: Yeah, thank you! Fixed.

Issue #1516, opened 2021-08-22
Automatic Platform Optimization


krizhanovsky:
## Scope

CloudFlare provides [APO](https://blog.cloudflare.com/automatic-platform-optimizations-starting-with-wordpress/) which seems just an automatic population of changed content into the cache, i.e. instead of purging content on update, they immediately update the web cache. Very similar set of features is provided by the [LiteSpeed WordPress plugin](https://wordpress.org/plugins/litespeed-cache/). This task will be easily done with the web-content tool. We can do not only web cache population, but also optimizations such as content minification.

Basically, the APO can be easily implemented using `wget -r <domain_name>`, probably with generated headers according to the web server settings for `Vary` values. However, in more advanced (heavy loaded) case wget requires to transfer response bodies and process them on the `wget` side. This is not wished and the Tempesta tool should send some header or introduce a specific method (this requires the Tempesta FW update) to make Tempesta FW just cache the responses and return only status code, without unnecessary headers and body.

This all is too complicated and it's much easier to add a specific header like
```
X-Tempesta-Cache: get
```
which is valid for `PURGE` method only. If the header is present, then Tempesta FW after deletion of the cache entry must send a GET request to the upstream server, cache it and *do not* send body to the client. ~Some of the headers also must be stripped, at least `Set-Cookie`~(this is the subject for #530). This must be a _special_ and _singualr_ header, see [tfw_http_hdr_t](https://github.com/tempesta-tech/tempesta/blob/master/fw/http.h#L182).

The grammar for the header (in terms of RFC 2616 2.1):
```
       X-Tempesta-Cache         = "X-Tempesta-Cache" ":" method

       method                           = "get"
```
At the moment we're going to support only one method for the cache population.

We SHOULD NOT support HTTP/2 for the PURGE logic.

## Testing & documentation 

Develop following functional tests in HTTP/1:

- [x] send `X-Tempesta-Cache` for other method, e.g. GET, and make sure that it doesn't do anything
- [x] Populate the cache with a response body, check that it's in the cache, PURGE the entry, change the response string and resend the request - check that the deproxy server receives the request. Repeat the request and check that the second response string is returned from the cache.
[...]

PR #1518, opened 2021-08-30
Purge cache update

IronBug:
Draft for PURGE with X-Tempesta-Cache flag (issue https://github.com/tempesta-tech/tempesta/issues/1516)


fw/http.c
```diff
@@ -2992,6 +2992,18 @@ tfw_h1_adjust_req(TfwHttpReq *req)
 	int r;
 	TfwHttpMsg *hm = (TfwHttpMsg *)req;
 
+	/* remove X-Tempesta-Cache and change PURGE to GET, if proper flag X-Tempesta-Cache was set */
```
krizhanovsky, line 2995: This is bit old-fashioned even for the kernel, but usialy we fir 80-characters wide lines, maybe couple of characters more

fw/http.c
```diff
@@ -2992,6 +2992,18 @@ tfw_h1_adjust_req(TfwHttpReq *req)
 	int r;
```
krizhanovsky, line 2992: Just a note to keep in mind with the upcoming changes. This function is called in a callback when we're about to forward the request to an upstream, i.e. when it can not be satisfied by the cache. Meantime `tfw_cache_purge_method()` is called without the callback (see `tfw_cache_do_action()`) and do not return a status code. It seems you'll need to update the workflow a little bit.

fw/http_msg.c
```diff
@@ -973,6 +1019,28 @@ tfw_http_msg_del_hbh_hdrs(TfwHttpMsg *hm)
...
+int
+tfw_http_msg_del_tempesta_cache_hdr(TfwHttpMsg *hm)
+{
+	TfwHttpHdrTbl *ht = hm->h_tbl;
+	unsigned int hid = ht->off;
+	int r = 0;
+
+	do {
+		hid--;
+		if (hid == TFW_HTTP_HDR_X_TEMPESTA_CACHE) {
```
krizhanovsky, line 1034: This header was made _special_, i.e. it has fixed index in the headers table and you can just check whether the header by the particular index in the table is present or not. Avoiding such loops is the main reason to have special headers.

fw/http.h
```diff
@@ -205,11 +206,12 @@ typedef enum {
 	TFW_HTTP_HDR_X_FORWARDED_FOR,
 	TFW_HTTP_HDR_KEEP_ALIVE,
 	TFW_HTTP_HDR_TRANSFER_ENCODING,
+	TFW_HTTP_HDR_X_TEMPESTA_CACHE,
 
 	/* Start of list of generic (raw) headers. */
 	TFW_HTTP_HDR_RAW,
 
-	TFW_HTTP_HDR_NUM	= 16,
+	TFW_HTTP_HDR_NUM	= 17,
```
krizhanovsky, line 214: We don't need the constant, `16` or `17`, the compiler will assign the constant to the last enum member automatically.

fw/http_msg.c
```diff
@@ -815,6 +820,47 @@ __hdr_sub(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
...
 }
 
+/**
+ * Substitute request method
+ *
+ * For a while it's just a custom realization for substituting dest method with sub method
+ * (Implemented to substitute PURGE with GET in X-Tempesta-Cache processing)
+ */
+int
+tfw_http_req_sub(TfwHttpMsg *hm, unsigned char dest_meth, unsigned char sub_meth)
```
krizhanovsky, line 830: We use the `tfw_http_req` prefix around the code to reference HTTP requests logic, so it makes sense to add some more sense to the name, e.g. `tfw_http_req_meth_sub()`.

fw/http_msg.c
```diff
@@ -815,6 +820,47 @@ __hdr_sub(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
...
+	/* do we need to look for "get" string in HTTP headers or just set it as stub constant value? */
+	dest_name = tfw_http_tbl_method_get_name(dest_meth);
+	sub_name = tfw_http_tbl_method_get_name(sub_meth);
+
+	/* todo: optimize this later */
+	if (!dest_name || !sub_name)
+		return TFW_BLOCK;
+
+	dest_len = strlen(dest_name);
+	sub_len = strlen(sub_name);
```
krizhanovsky, line 851: We do our best to make as many things computed statically as possible, so we call functions like `strlen()` in very rare cases. You basically don't need the lookup logic and length calculation if you reduce the function to handle substitution to "GET" only.

fw/http_msg.c
```diff
@@ -815,6 +820,47 @@ __hdr_sub(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
...
+
+	if (dest_len<sub_len)
```
krizhanovsky, line 853: We place spaces around operators, i.e. `if (dest_len < sub_len)`

fw/http.c
IronBug: this is draft. I didn't mean this as production code. this comment is for myself.

fw/http_msg.c
```diff
@@ -815,6 +820,47 @@ __hdr_sub(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
...
+		return TFW_BLOCK;
+
+	sub_str = TFW_STR_FROM_CSTR(sub_name);
+
+	if (ss_skb_cutoff_data(hm->msg.skb_head, &sub_str, 0, dest_len-sub_len))
```
krizhanovsky, line 858: This is the place where the things are getting complicated. We receive data in socket buffers, `skb`, which may deliver HTTP messages in many chunks (imagine that `GE`, `T / H` and `TTP/1.1` are received in 3 different TCP segments). We do not copy the input data. Instead, `TfwStr` gathers all the chunks *and* keep references to the original skbs. This way you can work with fragmented data using TfwStr API, but if you need you can go down and adjust the underlying skb data and resend it.

This all means that you need the original skbs under `sub_str` to be able to call `ss_skb_cutoff_data()`

fw/http_parser.c
```diff
@@ -3135,6 +3135,36 @@ __req_parse_x_forwarded_for(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+		if (IS_CRLF(c)) {
+			// do some processing of PURGE X-Tempesta-Cache request
+			hm->cache_ctl.flags |= TFW_HTTP_CC_CACHE_PURGE;
+			return __data_off(p);
+		}
+		__FSM_EXIT(CSTR_NEQ);
+	}
+done:
+	return r;
+}
```
krizhanovsky, line 3166: It'd be handy to add appropriate unit test for this header to `t/unit/test_http_parser.c`

fw/http_tbl.h
```diff
@@ -56,5 +56,6 @@ typedef struct {
 
 TfwVhost *tfw_http_tbl_vhost(TfwMsg *msg, bool *block);
 int tfw_http_tbl_method(const char *arg, tfw_http_meth_t *method);
+const char *tfw_http_tbl_method_get_name(int value);
```
krizhanovsky, line 59: This is different subsystem https://github.com/tempesta-tech/tempesta/wiki/HTTP-tables , which has nothing with cache, so we shouldn't use any routines from it in this patch.

fw/hpack.c
```diff
@@ -1252,6 +1252,9 @@ tfw_hpack_hdr_set(TfwHPack *__restrict hp, TfwHttpReq *__restrict req,
 	case TFW_TAG_HDR_RAW:
 		parser->_hdr_tag = TFW_HTTP_HDR_RAW;
 		break;
+	case TFW_TAG_HDR_X_TEMPESTA_CACHE:
+		parser->_hdr_tag = TFW_HTTP_HDR_X_TEMPESTA_CACHE;
+		break;
```
keshonok, line 1257: So far the case for the tag `TFW_TAG_HDR_RAW` was tried after all other more definitive (_special_) headers were tried. Perhaps, it's best to follow that convention.

fw/http_parser.c
```diff
@@ -3135,6 +3135,36 @@ __req_parse_x_forwarded_for(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+		 *	X-Tempesta-Cache = "X-Tempesta-Cache" ":" method
+		 *	method = "get"
+		 */
+		TRY_STR("get", Req_I_X_Tempesta_Cache, I_Tempesta_Cache_get);
+		__FSM_EXIT(CSTR_NEQ);
+	}
+
+	__FSM_STATE(I_Tempesta_Cache_get) {
+		if (IS_CRLF(c)) {
+			// do some processing of PURGE X-Tempesta-Cache request
```
keshonok, line 3158: I believe a fair attempt was made before to avoid C++ style comments in kernel space code. As I can see now some (just a few) have still infiltrated the TLS code.

fw/http_parser.c
```diff
@@ -6719,6 +6769,38 @@ __h2_req_parse_x_forwarded_for(TfwHttpMsg *hm, unsigned char *data, size_t len,
...
+			hm->cache_ctl.flags |= TFW_HTTP_CC_CACHE_PURGE;
+			return __data_off(p);
+		}
+		__FSM_EXIT(CSTR_NEQ);
```
keshonok, line 6796: I can't help but notice that the code now has a mix of two ways/styles to exit. One is using `__FSM_EXIT()` macro, and the the other is a straight `return` statement. For instance, `__FSM_EXIT(CSTR_NEQ)` and `return CSTR_NEQ`. The two styles can even be seen mixed together in the same functions.

It's not this new code's issue, it's been this way for a while.

fw/http_msg.c
```diff
@@ -815,6 +820,33 @@ __hdr_sub(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
...
+{
+	const char *dest_name;
+	size_t dest_len;
+
+	TfwStr sub_str = TFW_STR_STRING("get");
+
+	BUG_ON(hm->req->method != dest_meth);
+
+	dest_name = tfw_http_tbl_method_get_name(dest_meth);
+	dest_len = strlen(dest_name);
```
keshonok, line 840: I think this is the case where it can be done more effectively at compile time. A static table can be created that has the length of the string for each HTTP method, like `[TFW_HTTP_METH_PURGE] = SLEN("PURGE")`. There was no need for that before, but it looks like now there's a good case for that. There's a similar table for headers.

fw/http_msg.c
```diff
@@ -815,6 +820,33 @@ __hdr_sub(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
 	return TFW_PASS;
 }
 
+/**
+ * Substitute request method with "get"
```
keshonok, line 824: Personally, I am used to treat the term `sub` in the code as one in the series of `add/sub/mul/del`. If the goal is a shortcut name for `substitution`, perhaps a better and more common term would be `subst`. Just a thought.

fw/http_msg.c
```diff
@@ -815,6 +820,33 @@ __hdr_sub(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
...
+ *
+ * Note: replaced method name length is supposed to be shorter
+ *       than the replacement string ("get")
+ */
```
keshonok, line 828: I am sorry but the comment sounds ambiguous and unclear. Suggestion: `the name length of the replaced method must be the same or longer than the length of the replacement string ("GET")` or `the name length of the method being replaced must be the same or longer than the length of the replacement string ("GET")`.

fw/http_msg.c
```diff
@@ -815,6 +820,33 @@ __hdr_sub(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
...
+
+	BUG_ON(hm->req->method != dest_meth);
+
+	dest_name = tfw_http_tbl_method_get_name(dest_meth);
+	dest_len = strlen(dest_name);
+
+	BUG_ON(dest_len < sub_str.len);
+
+	if (ss_skb_cutoff_data(hm->msg.skb_head, &sub_str, 0, dest_len-sub_str.len))
+		return TFW_BLOCK;
```
keshonok, line 845: Just an idea: perhaps, a smarter modification may be considered that would require less modification of the underlying SKB (and thus be faster). In case of HTTP/1.1 this header line is the first line in a request. It can be argued that in many cases it would be the start of SKB data. If we align the new string to the right of the previous method's string (`PURGE` -> PU**GET**), then the modification of the SKB would be a simple adjustment of pointers (move by two chars forward) instead of a heavier and more expensive modification in the middle of the SKB.

Not sure if this trick is applicable to HTTP/2.  Perhaps not.

fw/http.c
```diff
@@ -4119,6 +4131,13 @@ tfw_h2_resp_next_hdr(TfwHttpResp *resp, const TfwHdrMods *h_mods)
 		    || tgt->flags & TFW_STR_HBH_HDR)
 			continue;
 
+		/*
+		 * Remove 'X-Tempesta-Cache' header
+		 * from the HTTP/2 response.
+		 */
```
keshonok, line 4137: This should fit in a single line.

fw/http_msg.c
```diff
@@ -815,6 +820,33 @@ __hdr_sub(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
...
+	TfwStr sub_str = TFW_STR_STRING("get");
+
+	BUG_ON(hm->req->method != dest_meth);
+
+	dest_name = tfw_http_tbl_method_get_name(dest_meth);
+	dest_len = strlen(dest_name);
+
+	BUG_ON(dest_len < sub_str.len);
+
+	if (ss_skb_cutoff_data(hm->msg.skb_head, &sub_str, 0, dest_len-sub_str.len))
```
keshonok, line 844: Just noticed that the comment from @krizhanovsky was lost after the update. That comment is comprehensive. This is not the right use of the SKB modification function. Also, the actual string replacement in the HTTP request (and the underlying SKB) seems to be missing.

fw/cache.c
```diff
@@ -1818,24 +1824,25 @@ tfw_cache_purge_invalidate(TfwHttpReq *req)
...
 
 	/* Deny PURGE requests by default. */
-	if (!(cache_cfg.cache && g_vhost->cache_purge && g_vhost->cache_purge_acl)) {
-		tfw_http_send_resp(req, 403, "purge: not configured");
-		return;
-	}
+		if (!(cache_cfg.cache && g_vhost && g_vhost->cache_purge && g_vhost->cache_purge_acl)) {
+			tfw_http_send_resp(req, 403, "purge: not configured");
+			return res;
+		}
```
krizhanovsky, line 1839: bad identation

fw/cache.c
```diff
@@ -2224,32 +2233,43 @@ cache_req_process_node(TfwHttpReq *req, tfw_http_cache_cb_t action)
...
-		id = tfw_h2_stream_id_close(req, HTTP2_HEADERS,
-					    HTTP2_F_END_STREAM);
-		if (unlikely(!id)) {
-			tfw_http_msg_free((TfwHttpMsg *)resp);
-			tfw_http_conn_msg_free((TfwHttpMsg *)req);
-			goto put;
+	T_DBG("Cache: req->method = %d, req->cache_ctl.flags = 0x%x (is not X-Tempesta-Cache: %d)\n",
+		(int)req->method, req->cache_ctl.flags,
+		(int)((req->cache_ctl.flags & TFW_HTTP_CC_CACHE_PURGE) == 0));
+	if (likely(req->method == TFW_HTTP_METH_GET &&
```
krizhanovsky, line 2239: With this check we break caching for all non-GET requests, which is wrong.

fw/cache.c
```diff
@@ -2224,32 +2233,43 @@ cache_req_process_node(TfwHttpReq *req, tfw_http_cache_cb_t action)
...
+			}
 		}
 	}
 out:
-	if (!resp && (req->cache_ctl.flags & TFW_HTTP_CC_OIFCACHED))
-		tfw_http_send_resp(req, 504, "resource not cached");
-	else
+		if (!resp && (req->cache_ctl.flags & TFW_HTTP_CC_OIFCACHED))
+			tfw_http_send_resp(req, 504, "resource not cached");
+		else {
```
krizhanovsky, line 2261: broken identation

fw/cache.c
```diff
@@ -2224,32 +2233,43 @@ cache_req_process_node(TfwHttpReq *req, tfw_http_cache_cb_t action)
...
 		 * if any with values from cached entries to revalidate stored
 		 * stale responses for both: client and Tempesta.
 		 */
-		action((TfwHttpMsg *)req);
+			if (req->method == TFW_HTTP_METH_GET ||
+				unlikely(req->cache_ctl.flags &
+					 TFW_HTTP_CC_CACHE_PURGE)) {
+				action((TfwHttpMsg *)req);
+			}
+		}
```
krizhanovsky, line 2272: Why to check for GET method? We already has made decision whether to cache the response or not.

The whole logic #1516 is unique for Tempesta and relatively sophisticated, so please leave comments on the steps how `TFW_HTTP_CC_CACHE_PURGE` is processed.

fw/cache.c
```diff
@@ -2265,10 +2285,18 @@ tfw_cache_do_action(TfwHttpMsg *msg, tfw_http_cache_cb_t action)
...
+			return;
+		if (unlikely(req->cache_ctl.flags & TFW_HTTP_CC_CACHE_PURGE)) {
+			cache_req_process_node(req, action);
+		} else {
+			tfw_http_send_resp(req, 200, "purge: success");
+		}
+	}
+	else {
 		cache_req_process_node(req, action);
+	}
```
krizhanovsky, line 2299: This is essencially just
```C
        if (unlikely(req->method == TFW_HTTP_METH_PURGE)) {
                if (tfw_cache_purge_method(req))
                        return;
                if (!(req->cache_ctl.flags & TFW_HTTP_CC_CACHE_PURGE)) {
                        tfw_http_send_resp(req, 200, "purge: success");
                        return;
                }
        }
        cache_req_process_node(req, action);
```
which is simpler. We don't need likely/unlikely here because we're already in `unlikely` branch

fw/cache.c
```diff
@@ -2265,10 +2285,18 @@ tfw_cache_do_action(TfwHttpMsg *msg, tfw_http_cache_cb_t action)
...
 
 	req = (TfwHttpReq *)msg;
-	if (unlikely(req->method == TFW_HTTP_METH_PURGE))
-		tfw_cache_purge_method(req);
-	else
+	if (unlikely(req->method == TFW_HTTP_METH_PURGE)) {
+		if (tfw_cache_purge_method(req))
+			return;
+		if (unlikely(req->cache_ctl.flags & TFW_HTTP_CC_CACHE_PURGE)) {
+			cache_req_process_node(req, action);
```
krizhanovsky, line 2292: This call doesn't look good since the function tries to fetch the TDB cache entry and should not find it since it's already PURGEed and only then we go to the `action()` callback. Why not to call the callback right here?

fw/cache.c
```diff
@@ -1784,8 +1784,14 @@ tfw_cache_add(TfwHttpResp *resp, tfw_http_cache_cb_t action)
...
 out:
-	resp->msg.ss_flags |= keep_skb ? SS_F_KEEP_SKB : 0;
-	action((TfwHttpMsg *)resp);
+	if (likely((req->cache_ctl.flags & TFW_HTTP_CC_CACHE_PURGE) == 0)) {
+		resp->msg.ss_flags |= keep_skb ? SS_F_KEEP_SKB : 0;
+		action((TfwHttpMsg *)resp);
+	}
+	else {
+		tfw_http_conn_msg_free((TfwHttpMsg *)resp);
+		tfw_http_send_resp(req, 200, "purge: success");
```
krizhanovsky, line 1793: We do our best to to minimize memory allocations and copyings but here you just remove the original response and create a new one from scratch. It's more efficient to just drop body from the response. Probably Content-Lenght and Transfer-Encoding headers as well. See the big comment about the proper protocol

krizhanovsky (changes requested):
The code doesn't satisfy the agreed design. Some cleanups and comments are also required.

fw/http_tbl.h
IronBug: So I should double the function of getting string from method MACRO in this module?
The only purpose of this call is to get the "purge" string from TFW_HTTP_METHOD_PURGE. Or I can fix the input method to PURGE only and avoid any transformations.

krizhanovsky: No, I commented at the place of using the function: just hardcode "PURGE" as you already hardcoded "GET"

fw/http.c
```diff
@@ -5417,6 +5444,11 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, const TfwFsmData *data)
 					     " processing error");
 		TFW_INC_STAT_BH(clnt.msgs_otherr);
 	}
+	else if (unlikely(req->resp) && unlikely(req->cache_ctl.flags &
+		TFW_HTTP_CC_CACHE_PURGE)) {
+		tfw_http_send_resp(req, 200, "purge: success");
```
krizhanovsky, line 5449: Well, the design of the pull request isn't so far from the discussed one. This `tfw_http_send_resp(req, 200...)` call is very confusing though. There are also several issues with the caching hooks and calls (also commented previously) and the HTTP requests adjustments. Fixing all the issues will lead to the desired changes.

fw/cache.c
IronBug: I tried to separate processing of PURGE with X-Tempesta-Cache from other cached requests (I wrongly supposed only GET requests were cached).
The difference between processing of common user requests and PURGE with upstream GET is skipping some unnecessary processing like building up the full-body request (we return only HTTP 200 to user).

fw/cache.c
IronBug: done

fw/cache.c
IronBug: done

fw/hpack.c
IronBug: done

fw/http.c
```diff
@@ -2829,6 +2829,8 @@ tfw_http_add_x_forwarded_for(TfwHttpMsg *hm)
 	int r;
 	char *p, *buf = *this_cpu_ptr(&g_buf);
 
+	BUG_ON(!hm->msg.skb_head);
```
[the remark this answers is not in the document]
IronBug: I removed it. needed it for debug when there were errors. 
done

fw/http.c
```diff
@@ -4119,6 +4121,13 @@ tfw_h2_resp_next_hdr(TfwHttpResp *resp, const TfwHdrMods *h_mods)
 		    || tgt->flags & TFW_STR_HBH_HDR)
 			continue;
 
+		/*
+		 * Remove 'X-Tempesta-Cache' header
+		 * from the HTTP/2 response.
+		 */
+		if (hid == TFW_HTTP_HDR_X_TEMPESTA_CACHE)
+			continue;
```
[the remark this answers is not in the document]
IronBug: I removed it.
done.

fw/http.c
IronBug: done

fw/http.c
```diff
@@ -5388,15 +5397,33 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, const TfwFsmData *data)
 	if (!TFW_MSG_H2(req))
 		hmsib = tfw_h1_req_process(stream, skb);
 
+	/* Remove X-Tempesta-Cache and change PURGE to GET,
+	 * if proper flag X-Tempesta-Cache was set
+	 */
+	if (!TFW_MSG_H2(req)) {
```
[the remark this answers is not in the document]
IronBug: I removed it
done

fw/http_msg.c
IronBug: ok. done.

fw/http_msg.c
IronBug: ok. done.

fw/http_msg.c
```diff
@@ -973,6 +1092,29 @@ tfw_http_msg_del_hbh_hdrs(TfwHttpMsg *hm)
...
+	TfwHttpHdrTbl *ht = hm->h_tbl;
+	int r = 0;
+	TfwStr *hdr;
+
+	hdr = &ht->tbl[TFW_HTTP_HDR_X_TEMPESTA_CACHE];
+
+	if(!hdr || TFW_STR_EMPTY(hdr))
+		return 0;
+
+	T_DBG("tfw_http_msg_del_tempesta_cache_hdr: delete header\n");
```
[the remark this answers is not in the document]
IronBug: done.

fw/http_parser.c
```diff
@@ -6719,6 +6771,50 @@ __h2_req_parse_x_forwarded_for(TfwHttpMsg *hm, unsigned char *data, size_t len,
...
+		return CSTR_NEQ;
+	}
+
+done:
+	return r;
+
+}
+STACK_FRAME_NON_STANDARD(__h2_req_parse_x_tempesta_cache);
+
+
```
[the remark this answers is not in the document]
IronBug: done

fw/cache.c
IronBug: done.

fw/cache.c
IronBug: done

fw/http_msg.c
IronBug: ok. I made it plain replace PURGE with GET, no variants.
just at the beginning there was a discussion to create a generic function to rewrite. then we simplified it to multiple targets - single subst. now it's simply designed for one case.

fw/cache.c
IronBug: The usual Tempesta FW HTTP response for PURGE looks like
```
< HTTP/1.1 200 OK
< date: Sun, 26 Sep 2021 13:20:58 GMT
< content-length: 0
< server: Tempesta FW/0.7.0
```

and the underlying server (Nginx in my current settings) returns

```
< HTTP/1.1 200 OK
< Server: nginx/1.10.3
< Date: Sun, 26 Sep 2021 13:19:54 GMT
< Content-Type: text/html
< Content-Length: 3378
< Last-Modified: Thu, 02 Sep 2021 20:40:13 GMT
< Connection: keep-alive
< ETag: "613136ad-d32"
< Expires: Sun, 26 Sep 2021 13:29:54 GMT
< Cache-Control: max-age=600
< Cache-Control: public
< Accept-Ranges: bytes
<
```

I can remove or edit any headers but is it ok to leave all the rest? and how to analyse the headers? The returned headers set is dependent on upstream settings, there may be whatever list of headers.
I think about leaving Date, fixing Content-Length and Setting server header. and deleting the rest. But this all is barely load work than re-creating a short answer.

fw/cache.c
IronBug: The usual Tempesta FW HTTP response for PURGE looks like
```
< HTTP/1.1 200 OK
< date: Sun, 26 Sep 2021 13:20:58 GMT
< content-length: 0
< server: Tempesta FW/0.7.0
```
and the underlying server (Nginx in my current settings) returns
```
< HTTP/1.1 200 OK
< Server: nginx/1.10.3
< Date: Sun, 26 Sep 2021 13:19:54 GMT
< Content-Type: text/html
< Content-Length: 3378
< Last-Modified: Thu, 02 Sep 2021 20:40:13 GMT
< Connection: keep-alive
< ETag: "613136ad-d32"
< Expires: Sun, 26 Sep 2021 13:29:54 GMT
< Cache-Control: max-age=600
< Cache-Control: public
< Accept-Ranges: bytes
<
```
I remove body and rewrite Content-Length to zero, leaving all the rest headers intact.

IronBug: I committed changes for this issue.

fw/cache.c
IronBug: I took this variant. done.

fw/http_msg.c
```diff
@@ -973,6 +1092,29 @@ tfw_http_msg_del_hbh_hdrs(TfwHttpMsg *hm)
...
+tfw_http_msg_del_tempesta_cache_hdr(TfwHttpMsg *hm)
+{
+	TfwHttpHdrTbl *ht = hm->h_tbl;
+	int r = 0;
+	TfwStr *hdr;
+
+	hdr = &ht->tbl[TFW_HTTP_HDR_X_TEMPESTA_CACHE];
+
+	if(!hdr || TFW_STR_EMPTY(hdr))
+		return 0;
```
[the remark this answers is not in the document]
IronBug: done

fw/http_tbl.h
IronBug: ok, done

fw/http.c
```diff
@@ -5417,6 +5444,11 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, const TfwFsmData *data)
 					     " processing error");
 		TFW_INC_STAT_BH(clnt.msgs_otherr);
 	}
+	else if (unlikely(req->resp) && unlikely(req->cache_ctl.flags &
+		TFW_HTTP_CC_CACHE_PURGE)) {
```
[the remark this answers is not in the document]
IronBug: this block was removed, the response is returned in another place now. 
done.

fw/http_msg.c
IronBug: done

fw/http_msg.c
```diff
@@ -815,6 +820,120 @@ __hdr_sub(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
...
+
+	dest_len = strlen(dest_name);
+
+	BUG_ON(dest_len < sub_len);
+
+	tail = dest_len - sub_len;
+
+	BUG_ON(!req->msg.skb_head);
+
+	skb = req->msg.skb_head;
```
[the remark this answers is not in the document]
IronBug: I rewrote the function to get deeper with skb's and chunks.
and renamed it to tfw_http_subst_purge_with_get to describe it better. 
for a while it's designed for one task only, but it can be turned into something generic later. 
I work on tests for it.

fw/http_msg.c
```diff
@@ -815,6 +820,120 @@ __hdr_sub(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
...
+
+	replace.data = sub_name;
+	replace.len = sub_len;
+	replace.tail = tail;
+	replace.skb = skb;
+
+	parsed = 0;
+	r = ss_skb_process(skb, tfw_http_req_meth_sub, &replace,
+			   &unused, &parsed);
```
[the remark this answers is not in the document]
IronBug: rewrote the function. done.

fw/http_msg.c
IronBug: for a while I didn't realize this. but I like the idea, I will consider it for further redesign.

fw/http_msg.c
```diff
@@ -815,6 +820,120 @@ __hdr_sub(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid)
...
+	if (r != T_OK) {
+		return TFW_BLOCK;
+	}
+
+	/* Cut off the tail. */
+	while (tail) {
+		bzero_fast(&it, sizeof(TfwStr));
```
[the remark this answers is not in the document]
IronBug: ok. I used simple memcpy for short memory blocks.

fw/http_parser.c
IronBug: I work on tests.
common generic tests with requests load work fine. I work on unit tests for splitted through multiple skb requests.

fw/ss_skb.c
```diff
@@ -868,7 +868,7 @@ __skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
 	return abs(len);
 }
 
-static int
+int
```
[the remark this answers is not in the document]
IronBug: for a while I use it to cut off the extra bytes in PURGE/GET rewrite. but I will possibly implement the hint that @keshonok offered above: the rewrite of PU...GET and then changing the beginning of the chunk, without rewrite.

fw/cache.c
```diff
@@ -1784,8 +1784,14 @@ tfw_cache_add(TfwHttpResp *resp, tfw_http_cache_cb_t action)
 	 */
 
 out:
-	resp->msg.ss_flags |= keep_skb ? SS_F_KEEP_SKB : 0;
-	action((TfwHttpMsg *)resp);
+	if (likely((req->cache_ctl.flags & TFW_HTTP_CC_CACHE_PURGE)==0)) {
```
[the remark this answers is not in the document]
IronBug: ok, I tried my best to see through my changes to keep the spaces. but sometimes I may miss it.

fw/http_parser.c
IronBug: I added test to test_http_msg.c but I cannot run tests on my VM. vmalloc and vfree in the tests init halt the kernel. it reports ```BUG: scheduling while atomic: insmod/5080/0x00000200``` and then kernel panic happens.  so I cannot check the test here.  it's not finished yet. I wanted to improve it but couldn't run it somehow. if I figure out a way to get rid of calls that kill the kernel maybe this will work out. I will try.

fw/cache.c
```diff
@@ -1752,6 +1752,48 @@ __cache_add_node(TDB *db, TfwHttpResp *resp, unsigned long key)
...
+			TFW_STR_STRING("0")
+		},
+		.nchunks = 3,
+	};
+	TfwStr *c = replacement.chunks;
+	int r;
+
+	if(!TFW_STR_EMPTY(&resp->body)) {
+		/* Delete the underlying body data. */
+		TFW_STR_FOR_EACH_DUP(dup, &resp->body, end) {
```
krizhanovsky, line 1776: We use DUP for duplicate stings, i.e. duplicate headers only. There is no duplicated body, so we can use simpler looping

fw/http_msg.c
krizhanovsky: Yeah, the @keshonok 's proposal is the right way to go. We actuall need to do this in two steps (in any order):
1. call the skb cutting logic (IIRC it's already optimized for the case, when the beginning is cut)
2. rewrite the 3 bytes

fw/cache.c
IronBug: ok. done.

krizhanovsky:
https://github.com/tempesta-tech/tempesta/pull/1533 is the successor of the PR

Issue #1511, opened 2021-07-06
Poor connection error messages


krizhanovsky:
# Scope

- [x] At the moment we see a lot of error messages like

```
[ 7565.987075] [tempesta tls] Warning: TLS: server requested by client is not known.
[ 7565.989528] [tempesta fw] Warning: Unrecognized TLS receive return code -0x7900, drop packet
```
and it's unclear what is the reason: whether it was a malicious user requesting bad host or some Tempesta issue. Need to print the requested host.

- [x] There is another problem message
```
[ 7400.456698] [tempesta fw] Warning: TLS: sni ext: client 104.131.181.:/ requested  unknown server name '93.115.28.125', reject connection.
```
in this case we see that the requested host is invalid, but there is also some issue with the client IP.

- [x] Third bad message is
```
[60052.119404] [tempesta tls] Warning: None of the common ciphersuites is usable (e.g. no suitable certificate)
```
We need to know at least the TLS version and first 1-2 advertised ciphersuites for the client to understand which ciphersuite we missed.

- [x] The warning `TLS: no global TLS certificates provided. Client TLS connections with unknown server name values or with no server name specified will be dropped.` is misleading. In practice golbal certificates are very rare.

- [x] Lastly we should print exactly one line for a connection problem: dmesg throughput is very limited.

# Testing

See current TLS functional tests for various different cases when we expect the warnings. We should log some meaningful human readable message instead of just error code.

Please fix `tls/test_tls_cert.py` tests (see comments), e.g. for `InvalidHash` test (SHA1) Tempesta throws misleading ERROR report on invalid `tls_certificate configuration`, just the same as if there is no certificate file at all, instead of correctly report about not supported SHA1.

snizovtsev:
> 104.131.181.:/

I've found that `tfw_put_dec` converts some numbers with nines  incorrectly:
```
'69'    '7/'
'79'    '8/'
'89'    '9/'
'99'    ':/'
'690'   '7/0'
'691'   '7/1'
....
```

Should also affect `X-Forwarded-For` header insertion and other places where `tfw_addr_fmt` or `tfw_addr_ntop` were used.

krizhanovsky:
@snizovtsev could you please add a unit test for the function to t/unit/test_addr.c ?

PR #1519, opened 2021-09-08
Fix IPv4/port decimals formatting issue

snizovtsev:
A bug in tfw_put_dec() cause a garbage printed on some rare IPv4/port
decimals. May fire in places calling tfw_addr_fmt() or tfw_addr_ntop(),
like log entries or X-Forwarded-For header.

Wrongly formatted address looks like '104.131.181.:/', where ':/'
stands for '99'. On 0..255 range only 69, 79, 89 and 99 affected.

To fix this issue we re-examined original code found in
https://elixir.bootlin.com/linux/v3.10.25/source/lib/vsprintf.c#L131
and check all valid inputs against dumb algorithm.

Found in #1511.

### DIFF
--- a/fw/addr.c
+++ b/fw/addr.c
@@ -419,6 +419,21 @@ tfw_addr_ifmatch(const TfwAddr *server, const TfwAddr *listener)
  * define few low-level helpers here instead of using something like snprintf().
  */
 
+static const u16 decpair[100] = {
+#define _(x) (__force u16) __cpu_to_le16((((x % 10) << 8) | (x / 10)) + 0x3030)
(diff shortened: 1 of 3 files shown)


krizhanovsky (commented):
Looks good especially with `TEST(tfw_put_dec, proof_correctness)`, but I'm wondering isn't it better just to use the newer kernel function from https://github.com/torvalds/linux/commit/7c43d9a30c527d9e06e2c55f82b56f28df43caed. We can just export `put_dec_trunc8()` to be used in modules (it's not inline anyway) or copy it. Probably the first way is more preferable since we'll get all updates and fixes for it in future kernel releases.

krizhanovsky (approved):
Good to merge

Issue #1332, opened 2019-08-10
Load certificates with Linux asymmetric keys API


krizhanovsky:
## Scope

Linux provides `add_key(2)` system call with `asymmetric-key(7)` API which allows to load X509 certificates including ECDSA keys. This API should be used instead of current mbedTLS-inherited X509 infrastructure. C++ programm must be called by `tempesta.sh` for certificates and keys management.

## References

[Kernel key management](https://lwn.net/Articles/210502/)

## Testing

No need for specific testing as current tests already load and manage various certificates.

krizhanovsky:
Move priorities from upstream to the product release, so 1.0 for now

PR #1521, opened 2021-09-13
Fix poor connection error messages

### LINKED ISSUE #1511: Poor connection error messages

snizovtsev:
- Remove `server requested by client is not known` where it is duplicated with `unknown server name` sni warning;
- Otherwise replace `server requested by client is not known` with `client requested misconfigured vhost`;
- Raise TLS server errors to WARN level to complement `0x7900` code with more details;
- Add cipher suite name into `None of the common ciphersuites is usable` message;
- Print unsupported algorithm OIDs;
- Print peer address in catch-all-errors block.

Hint: please review commit by commit, not the whole changes.

Tests modified: tempesta-tech/tempesta-test#190

Closes #1511


fw/tls.c
```diff
@@ -146,8 +146,9 @@ tfw_tls_msg_process(void *conn, TfwFsmData *data)
 	r = ss_skb_process(skb, ttls_recv, tls, &tls->io_in.chunks, &parsed);
 	switch (r) {
 	default:
-		T_WARN("Unrecognized TLS receive return code -0x%X, drop packet\n",
-		       -r);
+		TFW_WITH_ADDR_FMT(&c->peer->addr, TFW_WITH_PORT, addr_str,
+			T_WARN("Session %s rejected with error -0x%X\n",
+				addr_str, -r));
```
krizhanovsky, line 151: At this layer we already can't say enough details about the reason for the error and we can just print it as a second error message, which is not wished. I think we don't need any error messages on this layer. Please just make `ttls_recv()` to print error messages on each return path. We can get the client IP from the `TlsCtx->io_in.skb_list`.

tls/tls_srv.c
```diff
@@ -74,7 +74,7 @@ ttls_parse_servername_ext(TlsCtx *tls, const unsigned char *buf, size_t len)
 
 	servername_list_size = ((buf[0] << 8) | (buf[1]));
 	if (unlikely(servername_list_size + 2 != len)) {
-		T_DBG("ClientHello: bad SNI list size\n");
+		T_WARN("ClientHello: bad SNI list size\n");
```
krizhanovsky, line 77: Why do we need the patch " TLS server: raise handshake errors log severity "? The first claim in #1511 is about 2 meaningless lines in the log: we need not only to make the log messages more meaningful, but also reduce the number of records in the log. The kerne log (dmesg) isn't cheap and even now, experiencing only moderate security scans, we have plenty of log messages. If we increase the number of error messages, then the system becomes hardly monitored under more or less intensive malicious traffic raising many TLS error messages.

fw/tls.c
```diff
@@ -833,7 +834,7 @@ tfw_tls_get_if_configured(TfwVhost *vhost)
 
 #define SNI_WARN(fmt, ...)						\
 	TFW_WITH_ADDR_FMT(&cli_conn->peer->addr, TFW_NO_PORT, addr_str,	\
-			  T_WARN("TLS: sni ext: client %s requested "fmt, \
+			  T_WARN("client %s requested"fmt, \
```
krizhanovsky, line 837: I'm wondering what might be the reason for the broken last octet in the client IP `104.131.181.:/` as cited in #1511 ? Was it some formatting issue or did we experience a broader memory corruption issue?

krizhanovsky (changes requested):
Quite a good job, but more work is required.

fw/tls.c
snizovtsev: The reason was broken `tfw_put_dec()` and we have fixed that in #1519. Here I just shorten it because it's clear that all 3 messages containing vhost is about sni.

tls/tls_srv.c
snizovtsev: We cannot print both details and peer IP address on a single line. That is the reason for 2 lines: one prints details and another prints IP address. SNI code is implemented inside FW layer and avoids this problem. But in TLS layer it's not valid to use something like that:
```
TfwCliConn *cli_conn = &container_of(ctx, TfwTlsConn, tls)->cli_conn;
```

snizovtsev: Just noticed "We can get the client IP from the TlsCtx->io_in.skb_list.". Will try that.

tls/tls_srv.c
```diff
@@ -84,28 +84,34 @@ ttls_parse_servername_ext(TlsCtx *tls, const unsigned char *buf, size_t len)
...
-			return TTLS_ERR_BAD_HS_CLIENT_HELLO;
+			case -ENOENT:
+				ttls_send_alert(tls, TTLS_ALERT_LEVEL_FATAL,
+						TTLS_ALERT_MSG_UNRECOGNIZED_NAME);
+				return TTLS_ERR_BAD_HS_CLIENT_HELLO;
+			default:
+				ttls_send_alert(tls, TTLS_ALERT_LEVEL_FATAL,
+						TTLS_ALERT_MSG_INTERNAL_ERROR);
+				return TTLS_ERR_BAD_HS_CLIENT_HELLO;
+			}
```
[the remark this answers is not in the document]
snizovtsev: We print requested host name on all error paths from `ttls_sni_cb()`. No reason to print something here. The claim in #1511 miss the fact that `server requested by client is not known` duplicate `client X requested  unknown server name Y` in most cases.

tls/error.c
```diff
@@ -140,8 +118,6 @@ void ttls_strerror(int ret, char *buf, size_t buflen)
 			snprintf(buf, buflen, "TLS - Verification of the message MAC failed");
 		if (use_ret == -(TTLS_ERR_INVALID_RECORD))
 			snprintf(buf, buflen, "TLS - An invalid TLS record was received");
-		if (use_ret == -(TTLS_ERR_CONN_EOF))
-			snprintf(buf, buflen, "TLS - The connection indicated an EOF");
 		if (use_ret == -(TTLS_ERR_NO_CLIENT_CERTIFICATE))
 			snprintf(buf, buflen, "TLS - No client certification received from the client, but required by the authentication mode");
```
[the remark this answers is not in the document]
snizovtsev: We cannot eliminate them by simple "move to a warning on a throw side" strategy:

- ASN1/BASE64 errors combines with X509 like "X509 - Input invalid: ASN1 - ASN1 tag was of an unexpected value";
- Errors like TTLS_ERR_DHM_BAD_INPUT_DATA are used a lot of times, no reason to copy-paste string literals;
- Lots of codes are caught on the way, incl catch-all in `ttls_x509_crt_parse()`;
- We don't test TLS corner cases as much as original mbedtls do. It's easy to broke something or spam dmesg accidentally on some exotic conditions.

Also #1332 proposes replacing x509 mbedtls-inherited code with linux infrastructure. I don't see a reason for refactoring code that has to be removed anyway.

Maybe reducing scope to `TTLS_ERR_BAD_HS_*` would be enough for this task?

tls/tls_srv.c
krizhanovsky: We're going to propose Tempesta TLS for the Linux upstream, so it must be slef-sufficient and do not require any modules like Tempesta FW, even for logging. That might be not so trivial to get IP address from the skb list, but the good point is that we pretty sure that at the TLS layer TCP header are already processed (i.e. the kernel already has called `pskb_copy()` and the TCP header is in a contiguous memory).

snizovtsev: Ok, maybe consider refactoring TLS logging as `TlsCtx-> error_log()` indirect call then? Or buffer some limited amount of messages for an upper layers? With # 537 in mind we shouldn't hardcode logging back-end into TLS anyway.

tls/error.c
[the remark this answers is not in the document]
krizhanovsky: OK, let's then leave error.c as it is now after your changes. With moving to the Linux kernel, it's quite likely that we'll migrate to the native Linux asymmetric keys API #1332 , so some of the current Tempesta TLS code relating x509 will be eliminated.

tls/debug.h
```diff
@@ -35,4 +35,21 @@
...
+	if (ttls_extract_peer_addr(tls, _tls_peer_sa))		\
+		_tls_peer_sa = NULL;				\
+	action_expr;						\
+} while (0)
+
+#define TTLS_PEER_FMT "[c=%pISp] "
+#define TTLS_ERR(tls, fmt, ...) TTLS_LOG_WITH_PEER(tls, \
+	T_ERR(TTLS_PEER_FMT fmt, _tls_peer_sa, ##__VA_ARGS__))
+#define TTLS_WARN(tls, fmt, ...) TTLS_LOG_WITH_PEER(tls, \
+	T_WARN(TTLS_PEER_FMT fmt, _tls_peer_sa, ##__VA_ARGS__))
```
snizovtsev, line 51: @krizhanovsky TTLS messages looks like:
```
[  301.549539] [tempesta tls] Warning: [c=127.0.0.1:65280] ClientHello: bad extension size 1032 (tls->hs->ext_rem_sz=1177)
```
Does it look well or you have a better suggestion? I tried to move address part to the end but it breaks explicit `\n` convention.

keshonok: May I have a [generic] suggestion? Unless it's a debug message, it doesn't make sense to have something like internal variable or structure names in a message. It looks cryptic, and it's meaningless to a user. Still, that extra information might still be valuable to Tempesta crew when figuring out what the issue is, so, perhaps, it would make sense to replace some or all of those variable or structure members with somethig expressed with words.

In the above example, I think, **[c=127.0.0.1:65280]** can easily be replaced with just **[127.0.0.1:65280]**, and **(tls->hs->ext_rem_sz=1177)** with something like **(*remaining size* 1177)**. (I don't know what *ext_rem_sz* stands for).

On another note, perhaps, the use of round and/or square brackets in these messages could be unified.

tls/ttls.c
```diff
@@ -2070,6 +2070,39 @@ ttls_get_session(const TlsCtx *tls, TlsSess *dst)
...
+int ttls_extract_peer_addr(TlsCtx *tls, void *ss)
+{
+	struct sockaddr_in *sin = (struct sockaddr_in*) ss;
+	struct sockaddr_in6 *sin6 = (struct sockaddr_in6*) ss;
+	struct sk_buff *skb = tls->io_in.skb_list;
+	const struct iphdr *ih4 = ip_hdr(skb);
+	const struct ipv6hdr *ih6 = ipv6_hdr(skb);
+	const struct tcphdr *tcp = tcp_hdr(skb);
+
+	if (!skb || !ih6 || !tcp)
```
snizovtsev, line 2072: Just realized that `X_hdr` functions never returns NULL. Will replace with proper size check then.

krizhanovsky (commented):
The new changes are pretty good

fw/tls.c
krizhanovsky: Addressed in https://github.com/tempesta-tech/tempesta/pull/1521/commits/f66360340665cd9dda0a6db2f140f5dcbd1a7583

Issue #1530, opened 2021-10-12
Crash on running tempesta-test


voodam:
## Scope

See Testing. Also this patch of `scripts/tempesta.sh` solves the problem, but it's required to understand and fix the root cause.
```
@@ -184,6 +184,7 @@ update_js_challenge_templates()
 start()
 {
 	echo "Starting Tempesta..."
+	sleep 5
 
 	TFW_STATE=$(sysctl net.tempesta.state 2> /dev/null)
 	TFW_STATE=${TFW_STATE##* }
```

## Testing

- [Setup](https://github.com/tempesta-tech/tempesta-test) functional tests environment.
- [Build](https://github.com/tempesta-tech/tempesta/wiki/Install-from-Sources) Tempesta from the `master` branch.
- Run tests from the `master`:
```
$ ./run_tests.py
```
- Kernel panic should have occurred, like in the [crash.log](https://github.com/tempesta-tech/tempesta/files/7331548/crash.log).

ad-tmpst:
This is easy to reproduce with almost any test if you run it long enough (e.g. the cache tests) in a non-debug build.

What seems to happen here, is that the test suite starts the shutdown script, which it turn runs 'rmmod', so

1. tempesta_fw is rmmod-ed, and tfw_exit runs a bunch of exit_hooks
2. procfs, sock_srv & server hooks produce several BUG warnings
3. tempesta_db is rmmod-ed and we get an oops

[   59.272591] BUG: unable to handle page fault for address: ffffffffa006f720
[   59.272838] #PF: supervisor instruction fetch in kernel mode
[   59.273046] #PF: error_code(0x0010) - not-present page

In my case, the logs actually show that the panic is caused by the second "rmmod" command.

PR #1531, opened 2021-10-20
FW shutdown clean-up

ghost:
This patch fixes several problems related to FW shutdown.

1. Run 'tfw_stop' when the module exit function is called. Right now
all kinds of things break, if the module is unloaded before FW is
propery shut down with a sysctl write.

2. Expose the module state using a separate procfs file. This allows
userland utilities and tests to look for this file rather than parse
the dmesg buffer.

3. The start/shutdown banners are written using a non-limited
logger. I think it doesn't make a lot of sense to rate-limit these.

### DIFF
--- a/fw/main.c
+++ b/fw/main.c
@@ -63,7 +63,6 @@ tfw_runstate_is_reconfig(void)
 {
 	return READ_ONCE(tfw_reconfig);
 }
-EXPORT_SYMBOL(tfw_runstate_is_reconfig);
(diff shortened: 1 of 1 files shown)


ad-tmpst:
I added a new file instead of putting more stuff to perfstats. I think this will be easier to work with

snizovtsev:
Fixes #1530 on my setup :+1:

PR #1533, opened 2021-11-03
PURGE w/ update implementation (#1516)

ad-tmpst:
Note that while this seems to work for me, there are some quirks I didn't fix yet. For example:

- "X-Tempesta-Cache" is forwarded to an upstream. Probably need to mark it as an HBH header somewhere.
- The upstream GET request comes with a "Content-Length: 0" header.

There might be other similar problems. I need to properly test everything now.

The original commit message:

1. During the request parsing, if there is an "X-Tempesta-Cache"
   header, set the TFW_HTTP_B_PURGE_GET request flag. If the request
   method isn't PURGE, the flag is cleared at a later processing
   stage.

2. PURGE request handling is mostly the same as before, i.e. it will
   invalidate all matching cache entries, but also invoke a "cache
   action" callback if the request flag (1) is set, so the request may
   be forwarded to an upstream.

3. The message adjustment functions that are called before forwarding
   (in both directions) now support PURGE requests.

   - During downstream to upstream forwarding, we rewrite the request
     method to "GET" directly in an underlying SKB.

   - During upstream to downstream forwarding, the response body is
     removed and the "Content-Length" is set to zero. So clients can
     see original headers and a response status.

HTTP/2 is not supported for such requests.

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -1490,7 +1490,11 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 	}
 
 	/* Request method is a part of the cache record key. */
-	ce->method = req->method;
+	if (unlikely(req->method == TFW_HTTP_METH_PURGE
+	    && test_bit(TFW_HTTP_B_PURGE_GET, req->flags)))
+		ce->method = TFW_HTTP_METH_GET;
+	else
+		ce->method = req->method;
 
 	/* Write ':status' pseudo-header. */
 	ce->status = TDB_OFF(db->hdr, p);
@@ -1840,9 +1844,10 @@ tfw_cache_purge_invalidate(TfwHttpReq *req)
 }
 
 /**
- * Process PURGE request method according to the configuration.
+ * Process PURGE request method according to the configuration. Send a response,
+ * unless there are no errors *and* we're also told to refresh the cache.
  */
-static void
+static int
 tfw_cache_purge_method(TfwHttpReq *req)
 {
 	int ret;
@@ -1852,14 +1857,14 @@ tfw_cache_purge_method(TfwHttpReq *req)
 	/* Deny PURGE requests by default. */
 	if (!(cache_cfg.cache && g_vhost->cache_purge && g_vhost->cache_purge_acl)) {
 		tfw_http_send_resp(req, 403, "purge: not configured");
-		return;
+		return -EINVAL;
 	}
 
 	/* Accept requests from configured hosts only. */
 	ss_getpeername(req->conn->sk, &saddr);
 	if (!tfw_capuacl_match(&saddr)) {
 		tfw_http_send_resp(req, 403, "purge: ACL violation");
-		return;
+		return -EINVAL;
 	}
 
 	/* Only "invalidate" option is implemented at this time. */
@@ -1869,13 +1874,15 @@ tfw_cache_purge_method(TfwHttpReq *req)
 		break;
 	default:
 		tfw_http_send_resp(req, 403, "purge: invalid option");
-		return;
+		return -EINVAL;
 	}
 
 	if (ret)
 		tfw_http_send_resp(req, 404, "purge: processing error");
-	else
+	else if (!test_bit(TFW_HTTP_B_PURGE_GET, req->flags))
 		tfw_http_send_resp(req, 200, "purge: success");
+
+	return ret;
 }
 
 /**
@@ -2282,6 +2289,7 @@ put:
 static void
 tfw_cache_do_action(TfwHttpMsg *msg, tfw_http_cache_cb_t action)
 {
+	int ret;
 	TfwHttpReq *req;
 
 	if (TFW_CONN_TYPE(msg->conn) & Conn_Srv) {
@@ -2290,9 +2298,12 @@ tfw_cache_do_action(TfwHttpMsg *msg, tfw_http_cache_cb_t action)
 	}
 
 	req = (TfwHttpReq *)msg;
-	if (unlikely(req->method == TFW_HTTP_METH_PURGE))
-		tfw_cache_purge_method(req);
-	else
+	if (unlikely(req->method == TFW_HTTP_METH_PURGE)) {
+		ret = tfw_cache_purge_method(req);
+		/* Check if we want to do a GET in addition to PURGE. */
+		if (!ret && test_bit(TFW_HTTP_B_PURGE_GET, req->flags))
+			action((TfwHttpMsg *)req);
+	} else
 		cache_req_process_node(req, action);
 }
 
--- a/fw/http.c
+++ b/fw/http.c
@@ -2983,6 +2983,52 @@ tfw_h1_set_loc_hdrs(TfwHttpMsg *hm, bool is_resp, bool from_cache)
 	return 0;
 }
 
+/*
+ * Rewrite HTTP/1 "PURGE" method to "GET" directly inside a request SKB.
+ */
+static int
+tfw_h1_rewrite_purge_to_get(struct sk_buff *head)
+{
+	const char *q = "GET";
+	char *p;
+	struct skb_shared_info *si;
+	struct sk_buff *skb = head;
+	unsigned int f, z;
+	int ret;
+
+	/* Possible if somehow we already sent a response  */
+	BUG_ON(!head);
+
+	/* Chop two bytes from the beginning of SKB data. */
+	ret = ss_skb_chop_head_tail(head, head, 2, 0);
+	if (ret)
+		return ret;
+
+	do {
+		p = skb->data;
+		z = skb_headlen(skb);
+		while (z--) {
+			*p++ = *q++;
+			if (!*q)
+				return 0;
+		}
+		si = skb_shinfo(skb);
+		for (f = 0; f < si->nr_frags; ++f) {
+			p = skb_frag_address(&si->frags[f]);
+			z = skb_frag_size(&si->frags[f]);
+			while (z--) {
+				*p++ = *q++;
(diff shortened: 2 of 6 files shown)


fw/http.c
```diff
@@ -2983,6 +2983,50 @@ tfw_h1_set_loc_hdrs(TfwHttpMsg *hm, bool is_resp, bool from_cache)
...
+			if (++k == sizeof(to))
+				goto chop;
+		}
+	}
+
+	T_ERR("Not enough skb data for method rewrite?!\n");
+	return -ENOMEM;
+chop:
+	/* Chop two bytes from the beginning of SKB data. */
+	return ss_skb_chop_head_tail(NULL, skb, 2, 0);
```
krizhanovsky, line 3027: If the method name spans over several skbs, then we should pass skb_head here

fw/http.c
```diff
@@ -3503,6 +3553,36 @@ tfw_h2_adjust_req(TfwHttpReq *req)
...
+			TFW_STR_STRING(": "),
+			TFW_STR_STRING("0"),
+		},
+		.nchunks = 3,
+	};
+	TfwStr *c = replacement.chunks;
+
+	if (!TFW_STR_EMPTY(&resp->body)) {
+		ret = ss_skb_chop_head_tail(NULL, resp->msg.skb_head, 0,
+					    tfw_str_total_len(&resp->body));
```
krizhanovsky, line 3575: We also should use skb_head here. Just use `pv` tool to force the kernel to send data with extremely small skbs.

`tfw_str_total_len()` includes EOL - is this `\r\n` just before the body? If yes, then we should not chop it

fw/cache.c
```diff
@@ -2290,9 +2298,12 @@ tfw_cache_do_action(TfwHttpMsg *msg, tfw_http_cache_cb_t action)
 	}
 
 	req = (TfwHttpReq *)msg;
-	if (unlikely(req->method == TFW_HTTP_METH_PURGE))
-		tfw_cache_purge_method(req);
-	else
+	if (unlikely(req->method == TFW_HTTP_METH_PURGE && !TFW_MSG_H2(req))) {
```
krizhanovsky, line 2301: I don't think we need an explicit check here - what's wrong if we receive PURGE on h2? Probably it's better just to remove the method processing from h2 parsing - the parser will be faster and we'll need to support less code.

fw/cache.c
```diff
@@ -2290,9 +2298,12 @@ tfw_cache_do_action(TfwHttpMsg *msg, tfw_http_cache_cb_t action)
...
+		ret = tfw_cache_purge_method(req);
+		/* Check if want to do a GET in addition to PURGE */
```
krizhanovsky, line 2303:
```suggestion
		/* Check if we want to do a GET in addition to PURGE. */
```

krizhanovsky (commented):
So far so good.

> The upstream GET request comes with a "Content-Length: 0" header.

Is there a problem to get 200 response with "Content-Length: 0" for PURGE request? Does curl says anything bad about us?

fw/http.c
ad-tmpst: `tfw_str_total_len` here is the body length, it doesn't seem to include any part of the headers block. Then we chop the body length bytes from the end of the request skb list. I think we're good here.

fw/http.c
```diff
@@ -2983,6 +2983,50 @@ tfw_h1_set_loc_hdrs(TfwHttpMsg *hm, bool is_resp, bool from_cache)
...
+	unsigned int f, z, k = 0;
+	char *p;
+
+	/* Possible if somehow we already sent a response  */
+	BUG_ON(!skb);
+
+	z = skb_headlen(skb);
+	while (z--) {
+		BUG_ON(toupper(skb->data[k]) != from[k]);
+		skb->data[k] = to[k];
```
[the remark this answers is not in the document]
ad-tmpst: TLS doesn't appear to use linear data in my testing.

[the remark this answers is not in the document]
ad-tmpst: IOW, if we send a PURGE request over TLS the method name is still found in an skb fragment (not in the skb->data part).

fw/http.c
```diff
@@ -2983,6 +2983,52 @@ tfw_h1_set_loc_hdrs(TfwHttpMsg *hm, bool is_resp, bool from_cache)
...
+			p = skb_frag_address(&si->frags[f]);
+			z = skb_frag_size(&si->frags[f]);
+			while (z--) {
+				*p++ = *q++;
+				if (!*q)
+					return 0;
+			}
+		}
+		skb = skb->next;
+	} while(skb != head);
```
krizhanovsky, line 3026: I think we should use the standard `skb_store_bits()` here

krizhanovsky (changes requested):
There are still several places for improvement

fw/http.c
ad-tmpst: I'm sorry, how is this supposed to work? `skb_store_bits` tries to copy into the linear area, failing that - into the fragments, failing that - into the fragment list which *I think* we unroll earlier in a socket callback. Otherwise it returns EFAULT. `skb_store_bits` isn't going to walk over our skb list.

fw/http_parser.c
```diff
@@ -3295,15 +3295,26 @@ __req_parse_x_tempesta_cache(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
 		}
-		return CSTR_NEQ;
+		__FSM_I_JMP(I_Tempesta_Cache_skip);
+	}
+
+	__FSM_STATE(I_Tempesta_Cache_skip) {
+		parser->_acc = 0;
+		/* Skip the rest of the line. */
+		__FSM_I_MATCH_MOVE(nctl, I_Tempesta_Cache_skip);
+		if (!IS_CRLF(*(p + __fsm_sz)))
```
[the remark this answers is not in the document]
ad-tmpst: Sorry, I'm super confused now. Judging by the suggested test, we want to parse a comma delimited list of X-Tempesta-Cache methods now?

What about something like `X-Tempesta-Cache: head, get` should this work too?

fw/http_parser.c
```diff
@@ -3295,15 +3295,26 @@ __req_parse_x_tempesta_cache(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
 		TRY_STR_fixup(&TFW_STR_STRING("get"), Req_I_X_Tempesta_Cache,
 			      I_Tempesta_Cache_get);
-		return CSTR_NEQ;
+		/* If the method is not "GET" just ignore the rest of the
+		 * header line. */
+		__FSM_I_JMP(I_Tempesta_Cache_skip);
 	}
 
 	__FSM_STATE(I_Tempesta_Cache_get) {
 		if (likely(IS_CRLF(c))) {
```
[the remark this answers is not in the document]
ad-tmpst: I've changed the parsing logic so that it will also accept a comma or whitespace in addition to CRLF after the "get" string (and ignore the rest of the string as before).

fw/http.c
ad-tmpst: Also, I thought you didn't want us to use `memcpy` just to set three bytes?

fw/http_parser.c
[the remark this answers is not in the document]
ad-tmpst: I've added these tests. It actually just works as long as the method list starts with "get".

fw/http_parser.c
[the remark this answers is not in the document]
krizhanovsky: Why do we need to accept comas? The task requires to accept `get` only as the header value. Allowing spaces is good sine all HTTP headers allow spaces at the end and sometimes headers are generated automatically, so extra spaces may appear. Form the other hand, `X-Tempesta-Cache: get , \r\n` doesn't look correct. I believe we should not allow comas.

fw/http_parser.c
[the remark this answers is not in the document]
krizhanovsky: I wrote "like", not "exactly" :) We may allow or not `get, get` and we do not need to support `head, get`. Since `get , ` isn't correct value and we do not want to overcomplicate parser for senseless values, it makes sense to prohibit `get, get`

fw/t/unit/test_http_parser.c
```diff
@@ -3893,6 +3893,22 @@ TEST(http_parser, vchar)
...
 }
 
+TEST(http_parser, x_tempesta_cache)
+{
+	FOR_REQ_SIMPLE("X-Tempesta-Cache: get ") {
+		EXPECT_TRUE(test_bit(TFW_HTTP_B_PURGE_GET, req->flags));
+	}
+	FOR_REQ_SIMPLE("X-Tempesta-Cache: get, get") {
+		EXPECT_TRUE(test_bit(TFW_HTTP_B_PURGE_GET, req->flags));
+	}
```
krizhanovsky, line 3903: Let's prohibit the test

fw/http.c
krizhanovsky: > skb_store_bits isn't going to walk over our skb list.

Yeah, good point.

> Also, I thought you didn't want us to use memcpy just to set three bytes?

No, `memcpy()` is good since compiler can inline it, but our own `memcpy_fast()` is bad for small copies since it requires a far call. But the current implementation is good as well.

fw/http.c
```diff
@@ -2983,6 +2983,52 @@ tfw_h1_set_loc_hdrs(TfwHttpMsg *hm, bool is_resp, bool from_cache)
...
```
krizhanovsky, line 3026:
```suggestion
	} while (skb != head);
```

krizhanovsky (approved):
Looks good, just make the final cleanups and merge it with https://github.com/tempesta-tech/tempesta-test/pull/193 and close #1516

Issue #1540, opened 2021-12-22
Memory pool weirdness


ad-tmpst:
OK, so I've been reviewing the memory pool implementation and it does some really strange things that may be wrong.

I've started putting together the following test.
```c
	pl = __tfw_pool_new(1234);
	printk("pool_test: pool %px\n", pl);	// ffff888126ede010
	a = tfw_pool_alloc_not_align(pl, 1);
	b = tfw_pool_alloc(pl, 1);
	c = tfw_pool_alloc_not_align(pl, 1);
	printk("pool_test: a = %px\n", a);	// ffff888126ede020
	printk("pool_test: b = %px\n", b);	// ffff888126ede021
	printk("pool_test: c = %px\n", c);	// ffff888126ede029
	a = tfw_pool_alloc_not_align(pl, 1);
	printk("pool_test: a = %px\n", a);	// ffff888126ede02a
	b = tfw_pool_alloc_not_align(pl, 7);
	a = tfw_pool_realloc(pl, a, 1, 17);
	printk("pool_test: a = %px\n", a);	// ffff888126ede02a
```

So what I think happens here is:

`a` is 16 bytes past the pool pointer which makes sense since it's the first allocation. So far so good.

`b` goes right after `a`. Based on the function naming, I would expect the pointer to be aligned. The reason it works like this, is that in https://github.com/tempesta-tech/tempesta/blob/master/fw/pool.c#L125 if `align` is  true, then the size is adjusted to be a multiple of 8, but the returned pointer is not adjusted in any way, is this something expected? It surely isn't obvious.

The next allocation `c` isn't aligned either (although it shouldn't be), it just starts 8 bytes after the pointer to `b` (which we "aligned"). Still unsure what's being aligned here.

Finally, the `realloc` function assumes all sizes were previously aligned. https://github.com/tempesta-tech/tempesta/blob/master/fw/pool.c#L168 So in this case, it simply returns the previous pointer which we now can use to overwrite `b`.

I didn't go any further with this, there might be other problems.

I think it's possible to salvage `alloc` (assuming it actually should simply align pointers rather than... whatever it does.) but I'm not sure what to do with `realloc`. Overall my suggestion for fixing this is to remove the `align` flag entirely and always align pointers at a 8 byte boundary. I'm pretty sure this will almost never lead to a considerable overalignment in practice.

krizhanovsky:
[...]

PR #1541, opened 2021-12-24
fix alignment & realloc bugs #1540

ad-tmpst:
This patch should fix the alignment and realloc() problems described
in issue #1540. It also adds a couple of unit tests.

### DIFF
--- a/fw/pool.c
+++ b/fw/pool.c
@@ -46,10 +46,6 @@
 #include "lib/str.h"
 #include "pool.h"
 
-#define TFW_POOL_CHUNK_SZ(p)	(PAGE_SIZE << (p)->order)
-#define TFW_POOL_CHUNK_BASE(c)	((unsigned long)(c) & PAGE_MASK)
-#define TFW_POOL_CHUNK_END(p)	(TFW_POOL_CHUNK_BASE((p)->curr) + (p)->off)
-#define TFW_POOL_ALIGN_SZ(n)	(((n) + 7) & ~7UL)
 #define TFW_POOL_HEAD_OFF	(TFW_POOL_ALIGN_SZ(sizeof(TfwPool))	\
 				 + TFW_POOL_ALIGN_SZ(sizeof(TfwPoolChunk)))
 #define TFW_POOL_PGCACHE_SZ	512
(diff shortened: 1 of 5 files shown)


fw/pool.c
```diff
@@ -152,8 +151,14 @@ __tfw_pool_alloc(TfwPool *p, size_t n, bool align, bool *new_page)
...
-	a = (char *)TFW_POOL_CHUNK_END(p);
-	p->off += n;
+	if (align) {
+		a = (void *)TFW_POOL_ALIGN_SZ((unsigned long)TFW_POOL_CHUNK_END(p));
+		p->off = TFW_POOL_ALIGN_SZ(p->off) + n;
+	}
+	else {
+		a = (void *)TFW_POOL_CHUNK_END(p);
+		p->off += n;
+	}
```
krizhanovsky, line 161: There are 2 tests for `align` value on the fast path, which are just one after another, but the compiler is smart enough to make only one test. Also the compiler is smart enough to optimize the second assignment to `p->off = p_off + n`. Please write a comment, that we actually don't have 2 conditional jumps here and don't spend anything on double offset computation.

But I think the code would be still simpler with:
```C
p_off = (likely(align) ? TFW_POOL_ALIGN_SZ(p->off) : p->off) + n;
if (unlikely(p_off > TFW_POOL_CHUNK_SZ(p))) {
   // ....
}
if (likely(align))
    a = (void *)TFW_POOL_ALIGN_SZ((unsigned long)TFW_POOL_CHUNK_END(p));
else
   a = (void *)TFW_POOL_CHUNK_END(p);
p->off = p_off;
```
Also it'd be better to use `likely()` for align as the most code uses the aligned version

fw/pool.c
```diff
@@ -152,8 +151,14 @@ __tfw_pool_alloc(TfwPool *p, size_t n, bool align, bool *new_page)
 			: (void *)(c + 1);
 	}
 
-	a = (char *)TFW_POOL_CHUNK_END(p);
-	p->off += n;
+	if (align) {
+		a = (void *)TFW_POOL_ALIGN_SZ((unsigned long)TFW_POOL_CHUNK_END(p));
```
krizhanovsky, line 155: The compiler also realizes that here you use aligned `p->off`, which you recompute on the next line, so it uses the aligned version of the offset. However, the code with two macros doesn't look simple. Maybe it's better to make `p->off = align ? TFW_POOL_ALIGN_SZ(p->off) : p->off` at the begin instead of `p_off` computation? (We don't care about the slow path.)

If you do the code adjustments, please make sure that the code is still well optimized and there are no double conditions and computations. This code is perfomace crucial - we call it on each TfwStr memory chunk.

fw/t/unit/test_pool.c
```diff
@@ -0,0 +1,62 @@
...
+	TfwPool *p;
+	void *a, *b, *c;
+
+	p = __tfw_pool_new(1001);
+	EXPECT_NOT_NULL(p);
+
+	a = tfw_pool_alloc_not_align(p, 1);
+	b = tfw_pool_alloc_not_align(p, 7);
+	c = tfw_pool_realloc(p, a, 1, 17);
+
```
krizhanovsky, line 53: Please add test for the fast path, when we just move `p->off`, ideally if the allocation reaches the end of memory chunk.

fw/t/unit/test_pool.c
```diff
@@ -0,0 +1,62 @@
...
+	TfwPool *p;
+	void *a, *b, *c;
+
+	p = __tfw_pool_new(1001);
+	EXPECT_NOT_NULL(p);
+
+	a = tfw_pool_alloc_not_align(p, 1);
+	b = tfw_pool_alloc(p, 1);
+	c = tfw_pool_alloc_not_align(p, 1);
+
```
krizhanovsky, line 36: Please add a test for the slow path of a new page allocation, probably as a separate test case.

krizhanovsky (changes requested):
Some clanups and/or comments are required as well as it'd be nice to make couple of new tests

ad-tmpst:
OK, I'll add this stuff, but frankly it doesn't make sense to me to branch on `align` in the "fast" path if we care about the performance so much. All callers already know (statically) whether alignment is necessary. So it makes more sense to have three functions: aligned allocation, non-aligned allocation, and move the chunk allocation code to a separate function that accepts the `align` parameter.

krizhanovsky:
Yes, good point. The aligned and non-aligned fast paths can go to pool.h and on the slow path just call the function, which will implement the slow path only.

ad-tmpst:
@krizhanovsky Please take a look at the updated PR. Since this now touches H2 processing a bit I also tried running the "h2" tests from the suite, but half of these doesn't work as usual (even without this patch). If you have suggestions for other testing let me know.

krizhanovsky (approved):
LGTM, thank you! Yes there are known issues in the h2 code and we have separate testing and bugfix issues for it, so don't care about h2 tests in this PR.

PR #1545, opened 2021-12-29
Pk 530 - Preview of the cache control options

byko3y:


krizhanovsky:
I close the PR since it looks like a duplicate of https://github.com/tempesta-tech/tempesta/pull/1550

Issue #1296, opened 2019-07-04
Handle Content-Type in GET requests


krizhanovsky:
There is a [WAF bypass](https://www.youtube.com/watch?v=tSf_IXfuzXk) ([slides](https://www.slideshare.net/SoroushDalili/waf-bypass-techniques-using-http-standard-and-web-servers-behaviour)) technique to send GET requests with attack payload and Content-Type header - some web servers still process the body of GET requests. In `__req_parse_content_type()` we just skip Content-Type parsing and seems just normally forward such requests to a backend.

Need to check RFC for Content-Type requirements - if it prohibits such requests, then we must block them. Otherwise probably we need a new configuration option like `http_strict_mode` which allows to set Tempesta mode for various  HTTP fields in which it blocks such HTTP messages. If the mode isn't set for a HTTP field, then we should just print warning. There are potentially many HTTP fields which behavior are either not clearly defined by RFC or impractical (e.g. single LF character for end of HTTP line).

Please update [HTTP security](https://github.com/tempesta-tech/tempesta/wiki/HTTP-security) and [Web security](https://github.com/tempesta-tech/tempesta/wiki/Web-security) wikis with detailed examples of the attacks (with reference to the original talk) and Tempesta settings.

krizhanovsky:
RFC 7231 4.3.* say:
```
   A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request.
```
```
   A payload within a HEAD request message has no defined semantics;
   sending a payload body on a HEAD request might cause some existing
   implementations to reject the request.
```
```
   A payload within a DELETE request message has no defined semantics;
   sending a payload body on a DELETE request might cause some existing
   implementations to reject the request.
```
```
   A payload within a CONNECT request message has no defined semantics;
   sending a payload body on a CONNECT request might cause some existing
   implementations to reject the request.
```
```
   A client MUST NOT send a message body in a TRACE request.
```
So Content-Type and Content-Length headers must not be allowed for the methods as well as parser must not have a path to body processing for these methods.

We still need to describe the logic for blocking such requests with links to the attacks scenarios in the Wiki

krizhanovsky:
[...]

PR #1546, opened 2022-01-11
Incorrect fixups for Accept and Content-Type in request parser

byko3y:
I really dislike the idea of "saveval" argument in __TFW_HTTP_PARSE_SPECHDR_VAL/__TFW_HTTP_PARSE_RAWHDR_VAL/TFW_H2_PARSE_HDR_VAL macroses, because it spreads a single fixups logic into three remote locations. However, I haven't been able to find an easy solution to avoid a significant refactoring of the parser. So I'm just leaving the saveval problem as is and just repairing the concrete fixup logic, which currently leads to additional 30 assertions in tests.

### DIFF
--- a/fw/http_parser.c
+++ b/fw/http_parser.c
@@ -46,6 +46,11 @@
 #define __data_remain(pos)		(len - __data_off(pos))
 #define __data_available(pos, num)	(num <= __data_remain(pos))
 
+typedef struct {
+	size_t data_fixed_up;
+	size_t data_processed;
+} fsm_result;
+
 /**
  * The following set of macros is for use in generic field processing.
  * @__msg_field_open macro is used for field opening, @__msg_field_fixup
@@ -63,7 +68,8 @@
 do {									\
 	if (unlikely(tfw_http_msg_add_str_data(msg, field, data,	\
 					       __data_off(pos))))	\
-		return CSTR_NEQ;					\
+		__FSM_EXIT(CSTR_NEQ);					\
+	__fsm_last_fixup = data + (len);				\
 } while (0)
 
 #define __msg_field_finish(field, pos)					\
@@ -75,7 +81,8 @@ do {									\
 #define __msg_field_fixup_pos(field, data, len)				\
 do {									\
 	if (unlikely(tfw_http_msg_add_str_data(msg, field, data, len)))	\
-		return CSTR_NEQ;					\
+		__FSM_EXIT(CSTR_NEQ);					\
+	__fsm_last_fixup = data + (len);				\
 } while (0)
 
 #define __msg_field_finish_pos(field, data, len)			\
@@ -97,7 +104,8 @@ do {									\
 do {									\
 	if (unlikely(tfw_http_msg_add_str_data(msg,			\
 			&msg->stream->parser.hdr, data, len)))		\
-		return CSTR_NEQ;					\
+		__FSM_EXIT(CSTR_NEQ);					\
+	__fsm_last_fixup = data + (len);				\
 } while (0)
 
 #define __msg_hdr_set_hpack_index(idx)					\
@@ -117,6 +125,7 @@ do {									\
 	TfwHttpParser	*parser = &msg->stream->parser;			\
 	unsigned char	*p = data;					\
 	unsigned char	c = *p;						\
+	unsigned char	__maybe_unused *__fsm_last_fixup = p;		\
 	int		__maybe_unused __fsm_n;				\
 	size_t		__maybe_unused __fsm_sz;			\
 	TfwStr		__maybe_unused *chunk = &parser->_tmp_chunk;	\
@@ -304,7 +313,8 @@ do {									\
 	__fsm_n = __data_remain(p);					\
 	__fsm_sz = tfw_match_##alphabet(p, __fsm_n);			\
 	if (unlikely(__fsm_sz == __fsm_n)) {				\
-		__msg_hdr_chunk_fixup(data, len);			\
+		__msg_hdr_chunk_fixup(__fsm_last_fixup,			\
+				      __data_remain(__fsm_last_fixup));	\
 		parser->_i_st = &&to;					\
 		r = TFW_POSTPONE;					\
 		finish;							\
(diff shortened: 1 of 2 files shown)


fw/http_parser.c
```diff
@@ -1873,6 +1873,8 @@ __req_parse_content_type(TfwHttpMsg *hm, unsigned char *data, size_t len)
 	__FSM_STATE(I_EoL) {
 		__FSM_I_MATCH_MOVE_fixup(ctext_vchar, I_EoL, 0);
 		if (IS_CRLF(*(p + __fsm_sz))) {
+			/* __TFW_HTTP_PARSE_SPECHDR_VAL is told to not fixup it */
+			__msg_hdr_chunk_fixup(p, __fsm_sz);
```
krizhanovsky, line 1877: Well, this does fix https://github.com/tempesta-tech/tempesta/issues/1296#issuecomment-1008446777 . However, the fix looks like a workaround for `__TFW_HTTP_PARSE_SPECHDR_VAL(..., 0)` and we can also fix the issue simply with calling the function from `TFW_HTTP_PARSE_SPECHDR_VAL()`. However, I see that there are a lot of headers with values parsed with `__TFW_HTTP_PARSE_SPECHDR_VAL(..., 0)` instead of `TFW_HTTP_PARSE_SPECHDR_VAL()` - do we have problems with them as well? Seems not. So need to figure out why some of the headers are parsed with the first macro and others with the second one. Please write a good comments at least for the macros when each of them must be used. If the problem should be fixed exactly in the was as in the patch, then please write a good comment why - at this moment it's not clear.

krizhanovsky (changes requested):
Need to elaborate on the fix

byko3y:
I updated the pull request with a proposal for a more radical fix of the issue: https://github.com/tempesta-tech/tempesta/pull/1546/commits/7811dbae68a84b4c142c7407244db17d731733d2. The fundamental problem is the chaining of different parser's functions i.e. one function parsing the header value calls another function for parsing of a date inside the header value. This chaining currently requires the functions to know each others internals, because a single integer result (total amount of bytes processed) is not enough for communication, it tells the callee function (i.e. __TFW_HTTP_PARSE_SPECHDR_VAL) about the last processed byte, but doesn't tell anything about fixups within the called function (i.e. __req_parse_content_type that caused the bug).

So we either need additional global variable or additional integer in the function's result -- the latter is more convenient in terms of performance and code modularity.

The new mechanism for chaining of parser functions via fsm_result as advantage has the ability to lazily fixup string more often, which will overshadow the small overhead of having additional variable "__fsm_last_fixup" in the function's stack. At the same time, for functions that don't use the "__fsm_last_fixup" this variable will be completely optimized out. My benchmarks show no performance impact (less than precision of my tools, which is approx. 2%) from those changes.

keshonok:
I am not intimately familiar with the parser's internals in its current state, so this may be considered a view of an outsider who was familiar with the parser design before. I would appreciate (benefit from) a clear explanation (comment) of why an additional variable is required, why a subsequent function in a chain needs to know of a fixup that occurred inside an intermediate parsing function (parsing a date?), and so on. The whole idea of a fuxup is to stop a `TfwStr{}` chunk of data at a specific place semantically, so that there's no need to seek the end of that semantic piece of data later on (i.e. a date or an HTTP header name). Perhaps, a carefully constructed and clear example of a typical processing chain would be sufficient instead of a lengthy theoretical explanation of all possible situations.

Perhaps, `__TFW_HTTP_PARSE_SPECHDR_VAL`/`__TFW_HTTP_PARSE_RAWHDR_VAL`/`TFW_H2_PARSE_HDR_VAL` macros should not be considered as three remote locations because the primary reason there are three different macros is to avoid extra branching where it's unnecessary. Otherwise they serve the same purpose and essentially are "the same" semantically, albeit with small variations. The parser uses repetitious code inserts instead of extra branching, be that function calls or `if`/`else` branching.

krizhanovsky (changes requested):
`__TFW_HTTP_PARSE_HDR_VAL()` is a partial solution - will have 3 macros instead of 2 "wrapper" and "saveval" for compatibility only. There must be a good motivation to merge such change and this change doesn't introduce any more performance or security, but, in this unfinished form, increases the code complexity.

Let's focus on accurate fixing of the bug in the patch.

Dmitry-Gouriev:
I's avoid make so radical changes because they require an exhaustive testing and are about "a research".
Probably better to move these changes to some later millestone?

byko3y:
@keshonok wrote:
> I would appreciate (benefit from) a clear explanation (comment) of why an additional variable is required, why a subsequent function in a chain needs to know of a fixup that occurred inside an intermediate parsing function (parsing a date?), and so on.

It's not just about a function calling another function, but also about several sequential FSM states i.e. any non-trivial value processing with possible non-standard chunking. Some blocks delegate the fixup of the chunk to a subsequent code and some fixup immediately. The current code base contains sufficient amount of manually applied glue code to just work e.g. `__msg_hdr_chunk_fixup` calls within a nested function and "saveval" parameter for `__TFW_HTTP_PARSE_RAWHDR_VAL`/`__TFW_HTTP_PARSE_SPECHDR_VAL`. Failing to apply the glue correctly on any of the sides will result in corruption of parser's result -- that's precisely what caused the two corruptions I reported in the https://github.com/tempesta-tech/tempesta/issues/1547

I do agree `TFW_H2_PARSE_HDR_VAL` is a distinct macro, however, `__TFW_HTTP_PARSE_RAWHDR_VAL `and `__TFW_HTTP_PARSE_SPECHDR_VAL` are not. The only new condition in the merged `__TFW_HTTP_PARSE_HDR_VAL` is optimized away by GCC even at "-O0" level.

Considering the "added complexity", I spent like 2 hours rewriting all the request parsing functions and testing them, got zero impact on performance (1% or less). A complete rewrite (h2, responses) would probably take a whole day. Most of the modifications are about replacing `return __data_off(p)` with `__FSM_EXIT(__data_off(p))` and `return FSM_RESULT(__data_off(p))`, so I don't really see much complexity coming from it and don't feel like that's anything radical.

However, I agree the modification does not bring much to the table right now, so it can be considered of low priority.

byko3y:
The changes are already merged as a part of #1538, so I'm closing this pull request. I'm gonna create a new one with documentation-related changes for #1547.

PR #1548, opened 2022-01-12
Incorrect offsets for fast parsing of Accept header in request

byko3y:
Probably a copy-paste typo.


krizhanovsky (approved):
Good catch! LGTM.

It seems all the years we were parsing `Accept:` by the slow path. I didn't figure out how to catch other such bugs in the unit test... If you have an idea, please add a unit test for this header and may be other known headers for the parser.

Issue #530, opened 2016-05-26
[Cache] behavior tuning


krizhanovsky:
Tempesta Web-caching must be tunable in at least following parameters:

1. there should be option to not to cache responses with `Set-Cookie` header regardless `Cache-Control` header (see for example [how does Squid handle this](http://wiki.squid-cache.org/SquidFaq/InnerWorkings#How_does_Squid_deal_with_Cookies)). This is extremely useful for CDNs;

2. Sometimes it's necessary to ignore some or all `Cache-Control` directives;

3. configurable [quick abort](https://www.mnot.net/blog/2009/06/12/cache-win);

The issue is extension of #522. See "Cache Behaviour Tuning" discussion in [What to Look For in a HTTP Proxy/Cache](https://www.mnot.net/blog/2009/06/12/cache-win).

Issue #810, opened 2017-08-22
Missing functional tests for cache


intelfx:
We need more functional tests for correct handling of various cache Tempesta config options:

1. Options
   - [x] `cache_purge_acl`
     - include CIDR masks and single IP specifications
     - ensure overlapping entries are handled properly
     - ~include some dummy entries, ensure they don't influence real ones~
   - [x] `location` sections
     - include `location` sections that contradict rules inside them; ensure they have no effect
   - [x] `nonidempotent`
2. Improvements
   - [x] ~extend multiple-methods test with multiple URLs~
   - [x] run all cache tests (incl. elaborate ones) in all cache modes
3. Cache eviction (#515)
   - [x] ~test workload significantly larger than available RAM~ The test case is invalid since Tempesta DB is an in-memory dtatabase
   - [x] send responses with `Cache-Control: max-age=1` and check that 1 second later the response is fetched from the cache only if a request and our settings allow stale responses (check RFC 7234 chapter 5 and our Wiki)
4. Operation
   - [x] Check that `https://mysite/pic.jpg?myparam=test` and `https://mysite/pic.jpg?myparam2=test2` are cached separately
   - [x] Permanent redirects (e.g. 301 and 308) are cached, while temporal aren't

krizhanovsky:
For heavy tests running in CI (#484) environment please make a tests verifying that each NUMA node depending on the cache mode is a replication of whole Web cache or a shard of particular part of content.

ikoveshnikov:
Resent #801 has added [RFC 7234 Section 4.3.2](https://tools.ietf.org/html/rfc7234#section-4.3.2) support making TempestaFW able to validate cache of it's clients. But only basic functional functional tests was introduced. Here is the full list of required tests:

- [ ] RFC 7232 Section-4.1: The server generating a 304 response MUST generate: Cache-Control, Content-Location, Date, ETag, Expires, and Vary. Last-Modified is generated if the response does not have an ETag header. Currently only _ETag_ and _Date_ are checked. Note, that all the listed headers affects cache behaviour.

If-None-Match
---------------
If-None-Match present, If-Modified-Since header is fully ignored:
- [ ] Client sends GET of HEAD with If-None-Match header, Tempesta replies with 304 if:
    - [x] ETag in If-None-Match equals to ETag of current representation stored in cache;
    - [x] Same as above, but ETag of stored response is tagged as weak;
    - [x] Same as above, but ETag in If-None-Match tagged as weak;
[...]

PR #1550, opened 2022-01-12
Implemented additional caching features for #530

byko3y:
Parsing and application of no-cache and private directives with arguments.
Additional validation of Cache-Control header.
New cache_control_ignore, cache_resp_hdr_del configuration options.
Fixed corruption of cached responses with empty header value.

Fixed all the comments from the https://github.com/tempesta-tech/tempesta/pull/1539 here.

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -1545,7 +1545,6 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 	ce->hdrs = TDB_OFF(db->hdr, p);
 	ce->hdr_len = 0;
 	ce->hdr_num = resp->h_tbl->off;
-	printk("Writing cache");
 	FOR_EACH_HDR_FIELD_FROM(field, end1, resp, TFW_HTTP_HDR_REGULAR) {
 		int hid = field - resp->h_tbl->tbl;
 		/*
@@ -1656,14 +1655,14 @@ check_cfg_ignored_header(const TfwStr *field, TfwCaToken *tokens,
 	TfwCaToken *token = tokens;
 	for (i = 0; i < tokens_sz; i++) {
 		const TfwStr to_del = {
-			.data = &token->str,
+			.data = token->str,
 			.len = token->len - 1
 		};
 		BUG_ON(token->len > 10000);
 		if (tfw_stricmpspn(field, &to_del, ':') == 0) {
 			return true;
 		}
-		bytes_count += offsetof(TfwCaToken, str) + token->len;
+		bytes_count += sizeof(TfwCaToken) + token->len;
 		token = (TfwCaToken *)(bytes_count + (char *)tokens);
 	}
 	return false;
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -307,7 +307,7 @@ tfw_cache_policy(TfwVhost *vhost, TfwLocation *loc, TfwStr *arg)
 {
 	TfwCaPolicy *capo;
 
-	/* Search locations in current vhost. */
+	/* Search locations in current loc. */
 	if (loc && loc->capo_sz) {
 		if ((capo = tfw_capolicy_match(loc, arg)))
 			return capo->cmd;
@@ -397,6 +397,9 @@ static bool
 tfw_cache_employ_resp(TfwHttpResp *resp)
 {
 	TfwHttpReq *req = resp->req;
+	unsigned int ignored_resp_flags =
+		tfw_vhost_get_cache_policy_mask(req->location, req->vhost);
+	resp->cache_ctl.flags &= ~ignored_resp_flags;
 
 #define CC_REQ_DONTCACHE				\
 	(TFW_HTTP_CC_CFG_CACHE_BYPASS | TFW_HTTP_CC_NO_STORE)
@@ -1434,6 +1437,42 @@ __save_hdr_304_off(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *hdr, long off)
 	return false;
 }
 
+static bool
+check_ignored_header(const TfwStr *field, const TfwCaPolicyTokens *tokens,
+		     TfwCacheEntry *ce)
+{
+	int i;
+	for (i = 0; i < tokens->sz; i++) {
+		const TfwStr to_del = {
+			.data = tokens->tokens[i]->arg,
+			.len = tokens->tokens[i]->len
+		};
+		if (tfw_stricmpspn(field, &to_del, ':') == 0) {
+			if (ce)
+				--ce->hdr_num;
+			return true;
+		}
+	}
+	return false;
+}
+
+static bool
+check_ignored_header2(const TfwStr *field, const TfwStrArray *tokens,
+		      TfwCacheEntry *ce)
+{
+	int i;
+	/* TfwStrArray implies invariants:
+	   last < capacity and last = count - 1  */
+	for (i = 0; i < tokens->last; i++) {
+		if (tfw_stricmpspn(field, &tokens->items[i], ':') == 0) {
+			if (ce)
+				--ce->hdr_num;
+			return true;
+		}
+	}
+	return false;
+}
+
 /**
  * Copy response skbs to database mapped area.
  * @tot_len	- total length of actual data to write w/o TfwCStr's etc;
@@ -1450,6 +1489,7 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 	char *p;
 	unsigned short status_idx;
 	TfwStr *field, *h, *end1, *end2;
+	TfwCaPolicyTokens *hdr_del_tokens;
 	TDB *db = node_db();
 	TdbVRec *trec = &ce->trec, *etag_trec = NULL;
 	long n, etag_off = 0;
@@ -1538,9 +1578,12 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 	if (tfw_cache_h2_copy_str(&ce->rph_len, &p, &trec, rph, &tot_len))
 		return -ENOMEM;
 
+	hdr_del_tokens = tfw_vhost_get_cache_policy_tokens(req->location, req->vhost);
+
 	ce->hdrs = TDB_OFF(db->hdr, p);
 	ce->hdr_len = 0;
 	ce->hdr_num = resp->h_tbl->off;
+	printk("Writing cache");
 	FOR_EACH_HDR_FIELD_FROM(field, end1, resp, TFW_HTTP_HDR_REGULAR) {
 		int hid = field - resp->h_tbl->tbl;
 		/*
@@ -1554,6 +1597,23 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 		{
 			--ce->hdr_num;
 			continue;
+		}
+
+		if (hdr_del_tokens) {
+			/* remove headers mentioned in cache_resp_hdr_del */
+			if (check_ignored_header(field, hdr_del_tokens, ce))
+				continue;
+		}
+
+		if (resp->no_cache_tokens.last > 0) {
+			/* remove headers mentioned in Cache-Control: no-cache */
+			if (check_ignored_header2(field, &resp->no_cache_tokens, ce))
+				continue;
+		}
+		if (resp->private_tokens.last > 0) {
+			/* remove headers mentioned in Cache-Control: private */
+			if (check_ignored_header2(field, &resp->private_tokens, ce))
+				continue;
 		}
 
 		if (hid == TFW_HTTP_HDR_ETAG) {
@@ -1651,7 +1711,8 @@ __cache_entry_size(TfwHttpResp *resp)
 	TfwStr *host = &req->h_tbl->tbl[TFW_HTTP_HDR_HOST];
 	unsigned long via_sz = SLEN(S_VIA_H2_PROTO)
 		+ tfw_vhost_get_global()->hdr_via_len;
-
+	TfwCaPolicyTokens *hdr_del_tokens =
+			tfw_vhost_get_cache_policy_tokens(req->location, req->vhost);
 	/* Add compound key size */
 	res_size += req->uri_path.len;
 	tfw_http_msg_clnthdr_val(req, host, TFW_HTTP_HDR_HOST, &host_val);
@@ -1678,6 +1739,20 @@ __cache_entry_size(TfwHttpResp *resp)
 		    || hid == TFW_HTTP_HDR_SERVER
 		    || TFW_STR_EMPTY(hdr))
 			continue;
+
+		if (hdr_del_tokens) {
+			if (check_ignored_header(hdr, hdr_del_tokens, NULL))
+				continue;
+		}
+
+		if (resp->no_cache_tokens.last > 0) {
+			if (check_ignored_header2(hdr, &resp->no_cache_tokens, NULL))
+				continue;
+		}
+		if (resp->private_tokens.last > 0) {
+			if (check_ignored_header2(hdr, &resp->private_tokens, NULL))
+				continue;
+		}
 
 		size = sizeof(TfwCStr);
 
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -397,9 +397,10 @@ static bool
 tfw_cache_employ_resp(TfwHttpResp *resp)
 {
 	TfwHttpReq *req = resp->req;
-	unsigned int ignored_resp_flags =
-		tfw_vhost_get_cache_policy_mask(req->location, req->vhost);
-	resp->cache_ctl.flags &= ~ignored_resp_flags;
+	unsigned int cc_ignore_flags =
+		tfw_vhost_get_cc_ignore(req->location, req->vhost);
+	unsigned int effective_resp_flags =
+		resp->cache_ctl.flags & ~cc_ignore_flags;
 
 #define CC_REQ_DONTCACHE				\
 	(TFW_HTTP_CC_CFG_CACHE_BYPASS | TFW_HTTP_CC_NO_STORE)
@@ -418,18 +419,18 @@ tfw_cache_employ_resp(TfwHttpResp *resp)
 	 */
 	if (req->cache_ctl.flags & CC_REQ_DONTCACHE)
 		return false;
-	if (resp->cache_ctl.flags & CC_RESP_DONTCACHE)
+	if (effective_resp_flags & CC_RESP_DONTCACHE)
 		return false;
 	if (!(req->cache_ctl.flags & TFW_HTTP_CC_IS_PRESENT)
 	    && (req->cache_ctl.flags & TFW_HTTP_CC_PRAGMA_NO_CACHE))
 		return false;
-	if (!(resp->cache_ctl.flags & TFW_HTTP_CC_IS_PRESENT)
-	    && (resp->cache_ctl.flags & TFW_HTTP_CC_PRAGMA_NO_CACHE))
+	if (!(effective_resp_flags & TFW_HTTP_CC_IS_PRESENT)
+	    && (effective_resp_flags & TFW_HTTP_CC_PRAGMA_NO_CACHE))
 		return false;
 	if ((req->cache_ctl.flags & TFW_HTTP_CC_HDR_AUTHORIZATION)
 	    && !(req->cache_ctl.flags & CC_RESP_AUTHCAN))
 		return false;
-	if (!(resp->cache_ctl.flags & CC_RESP_CACHEIT)
+	if (!(effective_resp_flags & CC_RESP_CACHEIT)
 	    && !tfw_cache_status_bydef(resp))
 		return false;
 #undef CC_RESP_AUTHCAN
@@ -1437,42 +1438,6 @@ __save_hdr_304_off(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *hdr, long off)
 	return false;
 }
 
-static bool
-check_ignored_header(const TfwStr *field, const TfwCaPolicyTokens *tokens,
-		     TfwCacheEntry *ce)
-{
-	int i;
-	for (i = 0; i < tokens->sz; i++) {
-		const TfwStr to_del = {
-			.data = tokens->tokens[i]->arg,
-			.len = tokens->tokens[i]->len
-		};
-		if (tfw_stricmpspn(field, &to_del, ':') == 0) {
-			if (ce)
-				--ce->hdr_num;
-			return true;
-		}
-	}
-	return false;
-}
-
-static bool
-check_ignored_header2(const TfwStr *field, const TfwStrArray *tokens,
-		      TfwCacheEntry *ce)
-{
-	int i;
-	/* TfwStrArray implies invariants:
-	   last < capacity and last = count - 1  */
-	for (i = 0; i < tokens->last; i++) {
-		if (tfw_stricmpspn(field, &tokens->items[i], ':') == 0) {
-			if (ce)
-				--ce->hdr_num;
-			return true;
-		}
-	}
-	return false;
-}
-
 /**
  * Copy response skbs to database mapped area.
  * @tot_len	- total length of actual data to write w/o TfwCStr's etc;
@@ -1489,7 +1454,6 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 	char *p;
 	unsigned short status_idx;
 	TfwStr *field, *h, *end1, *end2;
-	TfwCaPolicyTokens *hdr_del_tokens;
 	TDB *db = node_db();
 	TdbVRec *trec = &ce->trec, *etag_trec = NULL;
 	long n, etag_off = 0;
@@ -1578,8 +1542,6 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 	if (tfw_cache_h2_copy_str(&ce->rph_len, &p, &trec, rph, &tot_len))
 		return -ENOMEM;
 
-	hdr_del_tokens = tfw_vhost_get_cache_policy_tokens(req->location, req->vhost);
-
 	ce->hdrs = TDB_OFF(db->hdr, p);
 	ce->hdr_len = 0;
 	ce->hdr_num = resp->h_tbl->off;
@@ -1591,29 +1553,12 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 		 * possible duplicates), since we will substitute it with our
 		 * version of this header.
 		 */
-		if ((field->flags & TFW_STR_HBH_HDR)
+		if ((field->flags & (TFW_STR_HBH_HDR | TFW_STR_NOCCPY_HDR))
 		    || hid == TFW_HTTP_HDR_SERVER
 		    || TFW_STR_EMPTY(field))
 		{
 			--ce->hdr_num;
 			continue;
-		}
-
-		if (hdr_del_tokens) {
-			/* remove headers mentioned in cache_resp_hdr_del */
-			if (check_ignored_header(field, hdr_del_tokens, ce))
-				continue;
-		}
-
-		if (resp->no_cache_tokens.last > 0) {
-			/* remove headers mentioned in Cache-Control: no-cache */
-			if (check_ignored_header2(field, &resp->no_cache_tokens, ce))
-				continue;
-		}
-		if (resp->private_tokens.last > 0) {
-			/* remove headers mentioned in Cache-Control: private */
-			if (check_ignored_header2(field, &resp->private_tokens, ce))
-				continue;
 		}
 
 		if (hid == TFW_HTTP_HDR_ETAG) {
@@ -1702,6 +1647,45 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 	return 0;
 }
 
+static bool
+check_cfg_ignored_header(const TfwStr *field, TfwCaToken *tokens,
+		     unsigned int tokens_sz)
+{
+	int i;
+	int bytes_count = 0;
+	TfwCaToken *token = tokens;
+	for (i = 0; i < tokens_sz; i++) {
+		const TfwStr to_del = {
+			.data = &token->str,
+			.len = token->len - 1
+		};
+		BUG_ON(token->len > 10000);
+		if (tfw_stricmpspn(field, &to_del, ':') == 0) {
+			return true;
+		}
+		bytes_count += offsetof(TfwCaToken, str) + token->len;
+		token = (TfwCaToken *)(bytes_count + (char *)tokens);
+	}
+	return false;
+}
+
+static bool
+check_cc_ignored_header(const TfwStr *field, const TfwStr *tokens)
+{
+	int i;
+	for (i = 0; i < tokens->nchunks; i++) {
+		if (tfw_stricmpspn(field, &tokens->chunks[i], ':') == 0) {
+			return true;
+		}
+	}
+	return false;
+}
+
+/*
+ * __cache_entry_size is supposed to be called before tfw_cache_copy_resp,
+ * to set the TFW_STR_NOCCPY_HDR flag on the headers, so we don't double-check
+ * the ignored header list (this check is O(N*M) currently).
+ */
 static long
 __cache_entry_size(TfwHttpResp *resp)
 {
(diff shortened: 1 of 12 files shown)


byko3y:
The original version of the pull request is not compilable because of excessive curly brackets. Somehow git brought my working copy 1 commit before the final version so I verified the final version itself. Good news is the last commit had only cosmetic changes so it's not a big deal.

fw/http_parser.c
```diff
@@ -2238,13 +2321,22 @@ __req_parse_authorization(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		parser->cache_control.dir_flag = 0;
+		parser->cache_control.filled = false;
+		/* A flag can be precluded either by semicolumn
+		 * (implied in this branch) or by comma. */
+		parser->cache_control.dir_allowed = true;
+		/* fall through */
+	}
 	__FSM_STATE(Req_I_CC) {
-		WARN_ON_ONCE(parser->_acc);
+		WARN_ON_ONCE(parser->cache_control.dir_flag);
```
krizhanovsky, line 2339: This is just not addressed comment https://github.com/tempesta-tech/tempesta/pull/1539/files#r775029875 . Please review the review notes once more. In short: you shouldn't fit new logic into existing states, most likely you need several new states.Introducing new states hurts, but not so much as introducing a lot of complexity with many if's and extra variables.

krizhanovsky, line 2339: This is just not addressed comment https://github.com/tempesta-tech/tempesta/pull/1539/files#r775029875 . Please review the review notes once more. In short: you shouldn't fit new logic into existing states, most likely you need several new states.Introducing new states hurts, but not so much as introducing a lot of complexity with many if's and extra variables.

fw/http_parser.c
```diff
@@ -2265,45 +2359,59 @@ __req_parse_cache_control(TfwHttpReq *req, unsigned char *data, size_t len)
...
-			parser->_acc = TFW_HTTP_CC_OIFCACHED;
+			parser->cache_control.dir_flag = TFW_HTTP_CC_OIFCACHED;
 		}, Req_I_CC_o, Req_I_CC_Flag);
 		TRY_STR_INIT();
 		__FSM_I_JMP(Req_I_CC_Ext);
 	}
 
 	__FSM_STATE(Req_I_CC_Flag) {
-		WARN_ON_ONCE(!parser->_acc);
+		/* A start of a common flag successfully detected */
```
krizhanovsky, line 2388: What is a "common flag"? The comment is very confusing in context of the state name `Req_I_CC_Flag`, where "flag" is about the parser flag, see https://github.com/tempesta-tech/tempesta/pull/1539/files#r774807540

fw/http_parser.c
```diff
@@ -2312,65 +2420,102 @@ __req_parse_cache_control(TfwHttpReq *req, unsigned char *data, size_t len)
...
+			/* reinit for next token */
+			parser->cache_control.dir_flag = 0;
 			__FSM_I_JMP(Req_I_CC);
 		}
-		if (IS_CRLF(c))
-			return __data_off(p);
-		return CSTR_NEQ;
+		if (IS_CRLF(c)) {
+			if (!parser->cache_control.filled)
+				__FSM_EXIT(TFW_BLOCK);
```
krizhanovsky, line 2514: A comment is required - why the success of the whole Cache-Control depends on cache_control.filled ? Actually, I believe the variable is just a variant of push down just to avoid new states, which makes the code overcomplicated and probably slower because we have extremely fast FSM technology

fw/http_parser.c
```diff
@@ -8587,59 +8792,133 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
+		__FSM_I_MOVE_fixup(Resp_I_CC_Flag_Separator, __fsm_sz, 0);
+	}
+
+	__FSM_STATE(Resp_I_CC_Flag_Separator) {
+		if (IS_CRLF(c))
+			__FSM_EXIT(TFW_BLOCK);
+		if (IS_WS(c) || c == ',')
+			__FSM_I_MOVE_fixup(Resp_I_CC_Flag_Argument_Token, 1, 0);
+		if (c == '"')
+			__FSM_I_MOVE_fixup(Resp_I_EoT, 1, 0);
```
krizhanovsky, line 8914: We should move `"` parsing into a separate state

fw/http_parser.c
```diff
@@ -8587,59 +8792,133 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
+			{
+				__FSM_EXIT(TFW_BLOCK);
+			}
+		});
+
+		tokens = parser->cache_control.dir_flag == TFW_HTTP_CC_PRIVATE ?
+			 &resp->private_tokens : &resp->no_cache_tokens;
+		if (__str_array_add_token(msg, tokens, p, __fsm_sz, false))
+			__FSM_EXIT(TFW_BLOCK);
+		/* Header name comparision requires a trailing semicolumn */
```
krizhanovsky, line 8901:
```suggestion
		/* Header name comparison requires a trailing colon */
```

fw/http_parser.c
```diff
@@ -8587,59 +8792,133 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
+		if (__str_array_add_token(msg, tokens, ":", 1, true))
+			__FSM_EXIT(TFW_BLOCK);
```
krizhanovsky, line 8903: Do not make a separate call with the generic allocation and TfwStr logic. You know precisely on the `__str_array_add_token()` call above that you finished to parse a header name and you can call the function with a flag, which will add the required `:`

fw/http_parser.c
```diff
@@ -8587,59 +8792,133 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
 		TRY_STR_LAMBDA_fixup(&TFW_STR_STRING("proxy-revalidate"),
 			&parser->hdr, {
-			parser->_acc = TFW_HTTP_CC_PROXY_REVAL;
+			parser->cache_control.dir_flag = TFW_HTTP_CC_PROXY_REVAL;
 		}, Resp_I_CC_p, Resp_I_Flag);
 		TRY_STR_INIT();
 		__FSM_I_JMP(Resp_I_Ext);
 	}
 
+	__FSM_STATE(Resp_I_Flag_Maybe_Qal) {
```
krizhanovsky, line 8843: Please make a comment for the state referencing the RFC chapters requiring header names parsing on `no-cache` and `private`

fw/http_parser.c
```diff
@@ -8680,56 +8964,77 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
 		resp->cache_ctl.flags |= TFW_HTTP_CC_S_MAXAGE;
 		__FSM_I_MOVE_fixup(Resp_I_EoT, __fsm_n, 0);
 	}
 
 	__FSM_STATE(Resp_I_Ext) {
+		/* Any flag we don't understand.
+		 * Here we just skip all the tokens, double quotes and
+		 * equality signs.
+		 */
 		/* TODO: process cache extensions. */
```
krizhanovsky, line 8988: Please remove the comment as well

krizhanovsky (changes requested):
There are still several bugs and more work is required to for the FSM workflow simplification (basically already commented in https://github.com/tempesta-tech/tempesta/pull/1539). The FSM issues are somewhat linked with https://github.com/tempesta-tech/tempesta/pull/1538 , so please talk with @Dmitry-Gouriev - we already had conversations about automata constructions.

Also #530 requires several functional tests tests including https://github.com/tempesta-tech/tempesta/issues/810#issuecomment-966690711 (note the `cache_bypass` and `cache_fullfil` configuration options, which can not be tested with unit tests) and Wiki update.

fw/http_parser.c
```diff
@@ -767,39 +773,116 @@ __hbh_parser_add_data(TfwHttpMsg *hm, char *data, unsigned long len, bool last)
...
+		if (!__mark_hbh_hdr(hm, hbh_hdr))
+			hbh_hdr->flags |= TFW_STR_HBH_HDR;
+	};
+
+	return 0;
+}
+
+/* similar to tfw_http_msg_grow_hdr_tbl */
+int
+__str_array_grow(TfwHttpMsg *hm, TfwStrArray *array)
```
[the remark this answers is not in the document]
Dmitry-Gouriev: Stange rename hdr -> hbh_hdr.

fw/http_parser.c
```diff
@@ -2265,45 +2359,59 @@ __req_parse_cache_control(TfwHttpReq *req, unsigned char *data, size_t len)
...
-			parser->_acc = TFW_HTTP_CC_NO_TRANSFORM;
+			parser->cache_control.dir_flag = TFW_HTTP_CC_NO_TRANSFORM;
 		}, Req_I_CC_n, Req_I_CC_Flag);
 		TRY_STR_INIT();
 		__FSM_I_JMP(Req_I_CC_Ext);
 	}
 
 	__FSM_STATE(Req_I_CC_o) {
+		if (unlikely(!parser->cache_control.dir_allowed))
+			__FSM_EXIT(TFW_BLOCK);
```
Dmitry-Gouriev, line 2379: Hello! I wonder a little that the conditions `!parser->cache_control.dir_allowed` is tested at several states.
It is set to `true` at `Req_I_CC_start`. It can become `false` during parsing? Or the parser can enter these states avoiding `Req_I_CC_start` and with `false` value for condition? Believing to the given name `dir_allowed`, the check probably better be outside this function at all?

fw/http_parser.c
```diff
@@ -2312,65 +2420,102 @@ __req_parse_cache_control(TfwHttpReq *req, unsigned char *data, size_t len)
...
+				if (unlikely(!parser->cache_control.dir_allowed))
+					__FSM_EXIT(TFW_BLOCK);
+				parser->cache_control.dir_flag = TFW_HTTP_CC_OTHER;
+			}
+			parser->cache_control.filled = true;
+		}
+		if (parser->cache_control.dir_flag != 0) {
+			/* Req_I_EoT will update the field at comma */
+			parser->cache_control.dir_allowed = false;
+		}
```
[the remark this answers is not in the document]
Dmitry-Gouriev: Hello! It seems I do not understand a conception of `dir_allowed` flag (or can hardly agree with it).

The flag is set to `true` in two states `Req_I_CC_start` and `Resp_I_EoT`, reset to `false` in several other states and is checked in many other states.

It seems that the flag is used to disable (make disallowed) a subset of states.

Is not it better to provide the same by hardcoding transitions between states with `__FSM_JPM`, `__FSM_I_MOVE_xxxx`?

We can control state transition hardly and `__FSM_START` is to restore that state on the next call of the function rather than to enter a random unpredictable state.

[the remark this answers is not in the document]
Dmitry-Gouriev: Looking from more general position, I'd use states to encode parser positions inside text being parsed and flags and variables to store information which was detected in the past and needed in the future. However performance consideration could dictate something different.

fw/hpack.c
```diff
@@ -1888,6 +1888,11 @@ do {									\
 
 		NEXT_STATE(HPACK_STATE_VALUE_TEXT);
 
+		if (src >= last && !h2_mode) {
+			/* We are going to skip the HPACK_STATE_VALUE_TEXT,
+			   but still need the newline for a correct HTTP/1 */
+			EXPAND_DATA(S_CRLF, SLEN(S_CRLF));
+		}
```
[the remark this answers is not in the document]
krizhanovsky: This is for retrieving headers with empty values from the cache, right? In other words I believe this is the part of https://github.com/tempesta-tech/tempesta/pull/1551 .

Please move the fix to #1551 AND write a better comment:
```
/*
 * The header has no value, so we are going to skip the HPACK_STATE_VALUE_TEXT,
 * but still need to write CRLF at the end of the header line for a correct HTTP/1
 */
```

fw/hpack.c
```diff
@@ -1888,7 +1888,16 @@ do {									\
 
 		NEXT_STATE(HPACK_STATE_VALUE_TEXT);
 
-		GET_NEXT_DATA(src >= last);
+		if (src >= last) {
```
krizhanovsky, line 1891: It still makes sense to leave the condition with `unlikely`

fw/vhost.h
```diff
@@ -117,6 +141,8 @@ typedef struct {
 	size_t			capo_sz;
 	size_t			nipdef_sz;
 	TfwCaPolicy		**capo;
+	TfwCaPolicyTokens	capo_hdr_del;
```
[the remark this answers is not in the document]
krizhanovsky: the comment is still not addressed

fw/http_parser.h
```diff
@@ -127,7 +127,6 @@ typedef struct {
 		struct {
 			short		dir_flag;
 			char		filled;
-			char		dir_allowed;
 		} cache_control;
 	};
```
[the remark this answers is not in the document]
Dmitry-Gouriev: I finally do not fully understand an idea of the `dir_flag `.
It seems to be unneccessary in request parser (you can make additions (`|=`) directly to `req->cache_ctl.flags` instead of `dir_flag `.
It used in 2 conditions in reqponse parser:
```C
Line 8750: 			if (parser->cache_control.dir_flag == TFW_HTTP_CC_PRIVATE)
Line 8787: 		tokens = parser->cache_control.dir_flag == TFW_HTTP_CC_PRIVATE ?
```
This is because the states `Resp_I_Flag_Maybe_Qal` and `Resp_I_CC_Flag_Argument_Token` are shared between several Cache-Control directives. You can divide these states into several ones to aviod sharing. If you do not like new states... Anyway, it seems, it would be simplier to use the `dir_flag` **only** to cache the directive flag for these checks, not to set `resp->cache_ctl.flags`.

fw/http_parser.c
```diff
@@ -2312,65 +2330,107 @@ __req_parse_cache_control(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		if (c == ',')
+			__FSM_I_MOVE(Req_I_After_Comma);
+		/* Tokens have to be separated by comma */
+		if (IS_TOKEN(c))
+			__FSM_EXIT(TFW_BLOCK);
+
+		if (IS_CRLF(c)) {
+			/* Cache-Control header cannot be empty */
+			if (!parser->cache_control.filled)
+				__FSM_EXIT(TFW_BLOCK);
```
krizhanovsky, line 2411: +1 for @Dmitry-Gouriev 's comments. `dir_flag` is just a replacement for old `_acc` and I'm good with it. But Dmitry pointed out that probably we can directly set the flags, w/o collecting them in `_acc`. `_acc` stands for "accumulator" and the primary target for the field is to accumulate multi-digit number values, which seem not necessary for flags.

`cache_control.filled` still looks unnatural, but there isn't so many code around it and now FSM looks much more consistent (so I closed #1539). However, probably we can determine whether Cache-Control had a directive with introducing only one new state. In this case we won't need `filled`.

If we can get rid of both the `filled` and `dir_flag`, then we can get rid of `Req_I_CC_start` and, with one new state instead of `filled`, we'll have the same number of states and more straightforward and likely faster FSM.

BTW since `Req_I_CC_start` is executed only once on very first FSM entrance, we can initialize the variable just at the beginning of the function and we don't need a state for it.

fw/http_parser.c
```diff
@@ -6013,24 +6089,49 @@ __h2_req_parse_cache_control(TfwHttpReq *req, unsigned char *data, size_t len,
...
+	__FSM_STATE(Req_I_After_Comma) {
+		if (IS_WS(c) || c == ',')
+			__FSM_H2_I_MOVE(Req_I_After_Comma);
+		parser->_acc = 0;
+		if (IS_TOKEN(c)) {
+			parser->cache_control.dir_flag = 0;
 			__FSM_I_JMP(Req_I_CC);
-		return CSTR_NEQ;
+		}
+		/* Should also check for parser->cache_control.filled at CRLF */
```
krizhanovsky, line 6133: we have no CRLFs in h2

fw/http_parser.c
```diff
@@ -8600,45 +8710,116 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
+				resp->cache_ctl.flags |= TFW_HTTP_CC_PRIVATE_QUAL;
+			else
+				resp->cache_ctl.flags |= TFW_HTTP_CC_NO_CACHE_QUAL;
+
+			/* Resp_I_CC_Opening_Quote => Resp_I_CC_Flag_Argument =>
+			 * => Resp_I_EoT */
+			__FSM_I_MOVE_fixup(Resp_I_CC_Opening_Quote, 1, 0);
+		}
+		/* not qualified, fall though */
+	}
```
krizhanovsky, line 8760: an empty line is wished here

fw/http_parser.c
```diff
@@ -8600,45 +8710,116 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
+		 * no-cache or private with arguments as defined by rfc7234
+		 * sections 5.2.2.2 and 5.2.2.6
+		 */
+		parser->cache_control.filled = true;
+		if (c == '=') {
+			/* Qualified form cancels out the general flag */
+			if (parser->cache_control.dir_flag == TFW_HTTP_CC_PRIVATE)
+				resp->cache_ctl.flags |= TFW_HTTP_CC_PRIVATE_QUAL;
+			else
+				resp->cache_ctl.flags |= TFW_HTTP_CC_NO_CACHE_QUAL;
```
krizhanovsky, line 8753: Ok, I see the usage of `dir_flag` versus direct update in `cache_ctl.flags`. Otherwise we'll need at least one more state them. However, neither of the *_QUAL flags are actually used besides the tests, so they should be just removed (and the tests reworked to test the things in a different way).

fw/http_parser.c
```diff
@@ -8600,45 +8710,116 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
+		/* ...but the flag appears to have an unknown suffix */
+		parser->cache_control.dir_flag = TFW_HTTP_CC_OTHER;
 		__FSM_I_JMP(Resp_I_Ext);
 	}
 
+	__FSM_STATE(Resp_I_CC_Opening_Quote) {
+		if (c != '"')
+			__FSM_EXIT(TFW_BLOCK);
+		else
+			__FSM_I_MOVE_fixup(Resp_I_CC_Flag_Argument_Token, 1, 0);
```
krizhanovsky, line 8779: Usually, we don't use else in such cases, just
```C
if (c != '"')
    __FSM_EXIT(TFW_BLOCK);
__FSM_I_MOVE_fixup(Resp_I_CC_Flag_Argument_Token, 1, 0);
```

fw/http_parser.c
```diff
@@ -8600,45 +8710,116 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
+		});
+
+		/* Header name comparison requires a trailing colon */
+		if (tfw_str_array_append_chunk(msg->pool, tokens,
+					       TFW_CACHE_CONTROL_TOKENS_MAX,
+					       p, __fsm_sz,
+					       ":", 1))
+		{
+			__FSM_EXIT(TFW_BLOCK);
+		}
```
krizhanovsky, line 8807: Looks much better now: shorter and more efficient!

fw/http_parser.c
krizhanovsky: It's still worth to move `"` parsing to a separate small state

fw/http_parser.c
```diff
@@ -8680,56 +8864,83 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
 
 	/* End of term. */
 	__FSM_STATE(Resp_I_EoT) {
-		if (c == ',')
-			__FSM_I_MOVE_fixup(Resp_I_EoT, 1, 0);
-
 		if (IS_WS(c))
 			__FSM_I_MOVE_fixup(Resp_I_EoT, 1, TFW_STR_OWS);
 
-		parser->_acc = 0; /* reinit for next token */
```
krizhanovsky, line 8719: Don't you need to reinit `dir_flag` as well?

krizhanovsky (changes requested):
The FSM is much better now and we're almost "there".

Functional tests and wiki are still in TODO. It'd better to make the tests and make sure that they do pass before the final review and merge of the code.

fw/vhost.h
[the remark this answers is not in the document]
byko3y: It is addressed: TfwCaPolicyTokens no longer exists, instead there is "TfwCaToken *capo_hdr_del".

fw/http_parser.c
byko3y: Sorry, I don't understand what part of the algorithm should be separated into the new state. To me it already looks like the most simple character probing possible.

fw/http_parser.c
```diff
@@ -2471,39 +2378,47 @@ __req_parse_cache_control(TfwHttpReq *req, unsigned char *data, size_t len)
...
+		__FSM_EXIT(TFW_BLOCK);
+	}
+
+	__FSM_STATE(Req_I_After_Comma) {
+		if (IS_WS(c) || c == ',')
+			__FSM_I_MOVE(Req_I_After_Comma);
+
 		if (IS_TOKEN(c)) {
 			/* reinit for next token */
 			parser->cache_control.dir_flag = 0;
```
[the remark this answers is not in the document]
byko3y: Agreed, RFC seems to forbid empty tokens, I'll fix that.

fw/http_parser.h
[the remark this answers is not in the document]
byko3y: Actually there is another condition in the Req_I_CC_Flag/Resp_I_Flag, so we cannot set the req->cache_ctl.flags immediately. However, I agree it's not needed anywhere else and Resp_I_Flag_Maybe_Qal should be segregated into two states.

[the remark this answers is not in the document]
Dmitry-Gouriev: @byko3y 
If you mean
```
WARN_ON_ONCE(!parser->cache_control.dir_flag);
```
this is about internal check for algorithm consistency rather than to parse the text.
However this consistence condition could be proved staticaly.
Is there any way to reach these states without `reg_flag` set?

fw/cache.c
```diff
@@ -1679,6 +1724,38 @@ __cache_entry_size(TfwHttpResp *resp)
...
 
+		/*
+		 * TODO #496: assemble all the string patterns into state machines
+		 * (one if possible) to avoid the loops over all configured and
+		 * mentioned in `private` and `no-cache` directives.
+		 */
+		/* remove headers mentioned in cache_resp_hdr_del */
+		if (hdr_del_tokens.tokens) {
+			if (check_cfg_ignored_header(hdr, hdr_del_tokens.tokens,
+						     hdr_del_tokens.sz))
```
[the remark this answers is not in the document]
byko3y: The nesting problem mostly comes from large tab width and small line width limit. However, in this particular case the nested block is small and fits well into 80 character limit.

fw/http_parser.h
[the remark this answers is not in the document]
byko3y: Everything can be proven statically, it's more about how complex the proof is. In this case I do agree the proof is quite simple so there's not much use for this particular statement.

fw/vhost.h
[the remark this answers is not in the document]
krizhanovsky: I meant that why not to combine all caching policy stuff, including the new configuration directives and `cc_ignore`, into TfwCaPolicy. In this sense there is no difference between TfwCaPolicyTokens and TfwCaToken.

However, I see that `capo` and `capo_hdr_del` are arrays of different sizes, not single structures, which can be combined. So we're good with the current design.

fw/http_parser.c
krizhanovsky: Yes, the state is simple, and good enough. I was confused with that it allows e.g. "Cache-Control: public, no-cache="Set-Cookie,"` (`"` just after `,`), but I believe that's OK.

fw/http_parser.c
```diff
@@ -5860,8 +5913,22 @@ __h2_req_parse_cache_control(TfwHttpReq *req, unsigned char *data, size_t len,
 
 	__FSM_START(parser->_i_st);
 
+	parser->cache_control.dir_flag = 0;
+
+	__FSM_STATE(Req_I_CC_start) {
+		if (IS_WS(c))
+			__FSM_I_MOVE(Req_I_CC_start);
```
krizhanovsky, line 5920: yeah, it seems we had the issue in the parsing function because it didn't process WS at the beginning. However, it seems we can move the check and transition to `Req_I_CC`

fw/http_parser.c
```diff
@@ -8600,45 +8699,140 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
 
+	/*
+	 * no-cache or private with arguments as defined by rfc7234
+	 * sections 5.2.2.2 and 5.2.2.6
+	 */
+	__FSM_STATE(Resp_I_Flag_No_Cache) {
+		if (c == '=') {
+			/* Qualified form cancels out the general flag */
+			/*
+			 * Resp_I_CC_Opening_Quote => Resp_I_CC_Dir_Arg_Token =>
```
krizhanovsky, line 8739:
```suggestion
			 * Resp_I_CC_Dir_Opening_Quote => Resp_I_CC_Dir_Arg_Token =>
```

fw/http_parser.c
```diff
@@ -8600,45 +8699,140 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
+		}
+		__FSM_I_JMP(Resp_I_Flag);
+	}
+
+	__FSM_STATE(Resp_I_Flag_Private) {
+		if (c == '=')
+			__FSM_I_MOVE_fixup(Resp_I_CC_Dir_Opening_Quote, 1, 0);
+
+		__FSM_I_JMP(Resp_I_Flag);
+	}
```
krizhanovsky, line 8752: This state is just the same as `Resp_I_Flag_No_Cache`, so the states should be unified

fw/http_parser.c
```diff
@@ -8600,45 +8699,140 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
+		if (IS_CRLF(c))
+			__FSM_EXIT(TFW_BLOCK);
+		/*
+		 * Two consecutive commas in arguments not allowed,
+		 * just like in Cache-Control directives.
+		 */
+		if (c == ',')
+			__FSM_EXIT(TFW_BLOCK);
+		if (c == '"')
+			__FSM_I_MOVE_fixup(Resp_I_EoT, 1, 0);
```
krizhanovsky, line 8829: This state is to allow `no-cache="Set-Cookie, "` - is this intentional? If yes, then comment is required. Otherwise, we can just omit the check and block `"` on the default transition to `TFW_BLOCK`

krizhanovsky (changes requested):
There are still things to fix and clanup

fw/http_parser.c
```diff
@@ -8566,13 +8646,32 @@ STACK_FRAME_NON_STANDARD(__resp_parse_age);
...
 	__FSM_START(parser->_i_st);
 
+	parser->cache_control.dir_flag = 0;
+
+	__FSM_STATE(Req_I_CC_start) {
+		if (IS_WS(c))
+			__FSM_I_MOVE_fixup(Req_I_CC_start, 1, TFW_STR_OWS);
+		/* Two consecutive commas not allowed */
+		if (c == ',')
+			__FSM_EXIT(TFW_BLOCK);
```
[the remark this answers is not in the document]
byko3y: I did some brief benchmarks on this - after every unnecessary branch removed the `__resp_parse_cache_control` function became 0.8% faster. During `TEST_RUN(http_parser, fills_hdr_tbl_for_resp)` the `__resp_parse_cache_control` takes 0.1% of the run time (thus parser is 0.008% faster with modifications), and `TEST_RUN(http_parser, cache_control)`, which feeds nothing but "cache-control" headers to the parser, leads to `__resp_parse_cache_control` working 25% of time (parser 0.2% faster, probably same increase for `__req_parse_cache_control`).
I just don't feel like performance is an issue here. Readability and robustness are though.
Also, trailing pure conditions with same statements in branches are optimized into a single unconditional statement even at "-Og" optimization level.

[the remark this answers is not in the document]
krizhanovsky: First of all, there are many questions about performance: how many times you ran the test? What the test is? Did you pick the *best* times for each of the tests? Did you test different hardware?

There is still no reason to have more code if you can have less code. More code tends to be harder to read and understand, especially if the code does senseless checks.

byko3y:
- Removed unnecessary conditions and reworked leading comma handling in cache-control parsing functions;
- Additional tests for commas in cache-control directives and their arguments;
- Reworked tfw_str_array_append_chunk();
- Reworked __tfw_strcmpspn() to match strings even if one of them lacks a trailing separator;
- Additional tests for both tfw_str_array_append_chunk and __tfw_strcmpspn().

Will update the copyright years once the final list of files is clear.

fw/http_parser.c
```diff
@@ -6014,23 +6080,37 @@ __h2_req_parse_cache_control(TfwHttpReq *req, unsigned char *data, size_t len,
...
+		/* Should also check for parser->cache_control.filled at CRLF */
+		__FSM_EXIT(TFW_BLOCK);
+	}
+
+	__FSM_STATE(Req_I_After_Comma) {
+		if (IS_WS(c))
+			__FSM_H2_I_MOVE(Req_I_After_Comma);
+		if (c == ',')
+			__FSM_EXIT(TFW_BLOCK);
```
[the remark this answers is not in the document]
krizhanovsky: The comment above isn't addressed. Why do we need the check and why it's not enough just to check for token at the below?

fw/str.c
```diff
@@ -670,6 +688,181 @@ tfw_str_add_duplicate(TfwPool *pool, TfwStr *str)
...
+ *                        flags = TFW_STR_COMPLETE }
+ *          chunks[1] = { nchunks = 3,
+ *                        len = undefined,
+ *                        chunks = ["tok", "en2", ":"],
+ *                        flags = TFW_STR_COMPLETE }
+ *          chunks[2] = { nchunks = 0,    <= single chunk stored inside parent
+ *                        len = 3,
+ *                        data = "tok",
+ *                        flags = 0 }     <= last uncompleted item
+ *          nchunks = 3
```
krizhanovsky, line 721: Please update the comment after the simplification. When the function has signature `tfw_str_array_append_chunk(TfwPool *pool, TfwStr *array, char *data, unsigned long len)` that won't be obvious how it's different from `tfw_str_add_compound()`/`__str_grow_tree()`, so please address this question (why do we need the new function) in the comment.

krizhanovsky (changes requested):
Unfortunately at least one more change and one more review is required. The main issue is still overcomplecated `tfw_str_array_append_chunk()`, which must be reworked. Several small cleanups are also required.

Please make the changes in a new commit, just like you did this time.

byko3y:
Hopefully resolved all the comments.

fw/str.c
```diff
@@ -670,6 +688,139 @@ tfw_str_add_duplicate(TfwPool *pool, TfwStr *str)
...
+	 */
+	if ((len || (last_token && !(last_token->flags & TFW_STR_COMPLETE)))
+	    && complete_last)
+	{
+		shall_complete = true;
+	} else {
+		shall_complete = false;
+	}
+
+	if (!len) {
```
krizhanovsky, line 741:
```suggestion
	if (unlikely(!len)) {
```

fw/str.c
```diff
@@ -670,6 +688,139 @@ tfw_str_add_duplicate(TfwPool *pool, TfwStr *str)
...
+	}
+
+	if (!last_chunk)
+		return -ENOMEM;
+
+	added_len += len;
+	last_chunk->len = len;
+	last_chunk->data = data;
+	/*
+	 * We don't care about array->len, but the last_token->len is required
```
krizhanovsky, line 808: I think it makes sense to replace
```
We don't care about array->len, but
```
with
```
array->len has no sense since this is an array of independent strings. However,
```

fw/str.c
```diff
@@ -670,6 +688,139 @@ tfw_str_add_duplicate(TfwPool *pool, TfwStr *str)
...
+		/* Merge single item into parent to avoid reallocation */
+		last_chunk = last_token;
+	} else {
+		last_chunk = __str_grow_tree(pool, last_token, 0, 1);
+	}
+
+	if (!last_chunk)
+		return -ENOMEM;
+
+	added_len += len;
```
krizhanovsky, line 804: `added_len` is always zero here, so just use `len` at the below and remove `added_len` completely

fw/http_parser.c
```diff
@@ -8600,45 +8715,134 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
...
+	__FSM_STATE(Resp_I_CC_Dir_Opening_Quote) {
+		if (c != '"')
+			__FSM_EXIT(TFW_BLOCK);
+
+		__FSM_I_MOVE_fixup(Resp_I_CC_Dir_Arg_Token, 1, 0);
+	}
+
+	__FSM_STATE(Resp_I_CC_Dir_Arg_Token) {
+		/* comma-separated field list (field-name/token) with optional
+		 * linear white space
```
krizhanovsky, line 8783: Please start the sentence form Captital letter and end it with dot.

Also please add following comment here: "We do not set TfwStr->skb for tokens, since the tokens are used in read-only fashion to compare strings and are never rewriten (TfwStr->skb is used for rewriting raw data only)."

krizhanovsky (approved):
Good to merge after the small cleanups

Issue #1535, opened 2021-11-30
Process skb chain in ss_skb_chop_head_tail()


krizhanovsky:
## Scope

#1533 calls `ss_skb_chop_head_tail()` to chop large parts of HTTP messages, e.g. body, while the function processes only one skb. Moreover, the function does not notify that it can cut less than requested number of bytes and just returns 0.

There are other usages of the function, and all of them are safe:
1. `ss_tcp_process_skb()` calls the function for single skb
2. `tfw_tls_chop_skb_rec()` scans the beginning and end of the skb chain for skbs smaller than the requested bytes to delete and only then calls  `ss_tcp_process_skb()`
3. `tfw_h2_frame_process()` does the same, but only for one part of the skb chin

We need to rework the `ss_skb_chop_head_tail()` API in following way:
1. the current implementation of `ss_skb_chop_head_tail()` must be renamed to `__ss_skb_chop_head_tail()` and `ss_tcp_process_skb()` must call `__ss_skb_chop_head_tail()`
2. `__ss_skb_chop_head_tail()` must do `WARN_ON_ONCE(trail || head)` just before `return 0` at the end.
3. `ss_skb_chop_head_tail()` must check the beginning and end of the skb chain (if both the `head` and `trail` are non-zero) and next call `__ss_skb_chop_head_tail()`
4. The h2 and tls code must be cleaned up to remove the skb chain checks since they're going to `ss_skb_chop_head_tail()`
5. The buggy HTTP calls from #1533 must remain the same.

There is a gotcha in `ss_skb_chop_head_tail()`:
```C
        while (head) {
                frag = &skb_shinfo(skb)->frags[0];
                n = min_t(int, skb_frag_size(frag), head);
                if ((r = __split_pgfrag_del(skb_head, skb, 0, 0, n, &it)))
                        return r;
                head -= n;
        }

        while (trail && si->nr_frags) {
                i = si->nr_frags - 1;
                frag = &skb_shinfo(skb)->frags[i];
                n = min_t(int, skb_frag_size(frag), trail);
                r = __split_pgfrag_del(skb_head, skb, i,
                                       skb_frag_size(frag) - n, n, &it);
```
The first loop supposes that a called checked that we try to cut off less bytes than `skb->len`, so it's sure that we have enough frags and do not care about them. It seems the second loop can do the same, but the function `ss_skb_chop_head_tail()` must check that `head + trail <= skb->len`. *Please double check the conditions and write good comments for the function.*

[...]

Issue #195, opened 2015-08-20
Repetitive HTTP 404 responses on fragmented request


i-rinat:
At version 173990895dcfc362387243f7f79cf2e6036e352a, if I make a request like that:
```
echo -en "GET / HTTP/1.1\r\nHost: loc\r\nConnection: close\r\n\r\n\r\n" | pv -L 300 | nc -q -1 192.168.122.152 80
```
with disabled cache and no backend configured, I get HTTP 404 responses one after another. Eventually they stop, when virtual machine with Tempesta runs out of memory.

There are two notable parts in the request. First, it's fragmented. Passing string through `pv -L 300` make request split in two chunk, `...loc\r\nCon` and `nection: clo...`. Second, there are additional `\r\n` at the end. Removing any of those makes reply normal.

keshonok:
Fixed in 39ddef7

Issue #115, opened 2015-06-15
Broken request/response pairing when response is blocked


keshonok:
A request is kept around until the response comes. That's done for multiple reasons, like caching, etc. When response is blocked for any reason, both response and the corresponding request should be removed and released. 

That's not done now in all cases, which may lead to incorrect pairing when responses may be paired with wrong requests. That invalid pairing then finds its way into Tempesta cache, and Tempesta may give out wrong responses to requests.

`tfw_http_resp_process()` should be reworked to keep pairing correct at all times.

krizhanovsky:
I'm wondering why responses can be blocked? It's probable case, but Tempesta should not block responses from regular Nginx or Apache HTTPD servers...

keshonok:
As described, this issue is solved in #154.

PR #1552, opened 2022-01-25
Solution for #1535 issue

Dmitry-Gouriev:
The PR contains (probably final) proposals for solution for issue #1535 .
The new function `ss_skb_list_chop_head_tail()` was intoduced.
The function chops given number of bytes from head and tail of skb list, iterating over the list if neccessary.
The function is a wrapper around older function `ss_skb_chop_head_tail()`, which remains unchanged.

Examples of usage:
- the temporary function `tfw_h1_chop_leading_crlf()` was replaced completely with  `ss_skb_list_chop_head_tail()`
- in the   the `tfw_h1_rewrite_purge_to_get` function an erroneous call to  `ss_skb_chop_head_tail()` was replaced with correct call to `ss_skb_list_chop_head_tail()`. This replacement fixes the bug https://github.com/tempesta-tech/tempesta/issues/1535#issuecomment-988478106 which was the formal reason to open the #1535  isddue.

Correctness is checked by the `tests test_malformed_crlfs.py` and `test_purge_hch.py`; both tests passed.
(Both tests reside in #195 https://github.com/tempesta-tech/tempesta-test/pull/195#issue-1073979863  PR for now :( ).

However chopping from the tail of the list and both from the head and the tail is still not tested.

### DIFF
--- a/fw/ss_skb.c
+++ b/fw/ss_skb.c
@@ -1061,10 +1061,10 @@ multi_buffs:
 	while (unlikely(skb->len <= head))
 	{
 		head -= skb->len;
-		/* We do not use ss_skb_unlink() here to prevent it
-	 	 * to remove the last skb in the list and to skip
-	 	 * unneccessary checks and actions inside it, here
-		 * and in the similar loop for tail below.
+		/* We do not use ss_skb_unlink() here and in
+		 * in the similar loop for tail below to prevent
+	 	 * removing the last skb in the list and to skip
+	 	 * unneccessary checks and actions inside the func.
 	 	 */
 		skb->next->prev = skb->prev;
 		skb->prev->next = skb->next;
--- a/fw/ss_skb.c
+++ b/fw/ss_skb.c
@@ -1017,18 +1017,18 @@ ss_skb_chop_head_tail(struct sk_buff *skb_head, struct sk_buff *skb,
  * otherwise an error is retuned.
  */
 int
-ss_skb_list_chop_head_tail(struct sk_buff **skb_list_head, 
+ss_skb_list_chop_head_tail(struct sk_buff **skb_list_head,
 			   size_t head, size_t trail)
 {
 	struct sk_buff *skb, *skb_hd;
 	size_t sum;
 	int ret;
-	
+
 	skb_hd = *skb_list_head;
 	if (likely(skb_hd->next == skb_hd))
 		goto single_buff;
 	goto multi_buffs;
-	
+
 	/* Everywhere in the function we perform 'redundant'
 	 * checks for head and tail values to avoid unneccessary
 	 * calls to underlying ss_skb_chop_head_tail() for
@@ -1049,14 +1049,14 @@ single_buff:
 	return ss_skb_chop_head_tail(NULL, skb_hd, head, trail);
 
 multi_buffs:
-    /* skb_list contains more than 1 skb &&
+	/* skb_list contains more than 1 skb &&
 	 * skb_hd points to head element of the list
 	 */
-	 
+
 	/* Here below we delete heading skbs which size
 	 * is less than chop demand at the head,
 	 * switching to single_buff: in case
-	 */ 
+	 */
 	skb = skb_hd;
 	while (unlikely(skb->len <= head))
 	{
@@ -1074,15 +1074,15 @@ multi_buffs:
 		if (unlikely(skb->next == skb))
 			goto single_buff;
 	}
-	
-    /* skb_list still contains more than 1 skb &&
+
+	/* skb_list still contains more than 1 skb &&
 	 * skb_hd points to head element of the list
 	 */
 
 	/* Here below we delete trailing skbs which size
 	 * is less than chop demand at the tail,
 	 * switching to single_buff: in case
-	 */ 
+	 */
 	skb = skb_hd->prev;
 	while (unlikely(skb->len <= trail)) {
 		trail -= skb->len;
@@ -1093,23 +1093,21 @@ multi_buffs:
 		if (unlikely(skb == skb_hd))
 			goto single_buff;
 	}
-	
-    /* skb_list still contains more than 1 skb &&
+
+	/* skb_list still contains more than 1 skb &&
 	 * skb_hd points to head element of the list &&
 	 * skb points to last element of the list
 	 */
-	 
+
 	/* Here we remove remaining head and trail bytes, if any */
-	if (head)
-	{
+	if (head) {
 		ret = ss_skb_chop_head_tail(NULL, skb_hd, head, 0);
 		if (unlikely(ret))
 			return ret;
 	}
-	
 	if (trail)
 		return ss_skb_chop_head_tail(NULL, skb, 0, trail);
-	
+
 	return 0;
 }
 
--- a/fw/ss_skb.c
+++ b/fw/ss_skb.c
@@ -1025,9 +1025,8 @@ ss_skb_list_chop_head_tail(struct sk_buff **skb_list_head,
 	int ret;
 
 	skb_hd = *skb_list_head;
-	if (likely(skb_hd->next == skb_hd))
-		goto single_buff;
-	goto multi_buffs;
+	if (unlikely(skb_hd->next != skb_hd))
+		goto multi_buffs;
 
 	/* Everywhere in the function we perform 'redundant'
 	 * checks for @head and @trail values to avoid unneccessary
@@ -1041,7 +1040,7 @@ single_buff:
 	 */
 	sum = head + trail;
 	if (WARN_ON_ONCE(skb_hd->len <= sum))
-		return -EBADMSG;
+		return -EINVAL;
 	if (unlikely(sum == 0))
 		/* Nothing to chop */
 		/* This check is mostly for jumps from branches below */
--- a/fw/ss_skb.c
+++ b/fw/ss_skb.c
@@ -1061,7 +1061,7 @@ multi_buffs:
 	while (unlikely(skb->len <= head)) {
 		head -= skb->len;
 		/* We do not use ss_skb_unlink() here and in
-		 * in the similar loop for tail below to prevent
+		 * the similar loop for tail below to prevent
 	 	 * removing the last skb in the list and to skip
 	 	 * unneccessary checks and actions inside the func.
 	 	 */
--- a/fw/ss_skb.c
+++ b/fw/ss_skb.c
@@ -1058,8 +1058,7 @@ multi_buffs:
 	 * switching to single_buff: in case
 	 */
 	skb = skb_hd;
-	while (unlikely(skb->len <= head))
-	{
+	while (unlikely(skb->len <= head)) {
 		head -= skb->len;
 		/* We do not use ss_skb_unlink() here and in
 		 * in the similar loop for tail below to prevent
--- a/fw/ss_skb.c
+++ b/fw/ss_skb.c
@@ -1100,12 +1100,12 @@ multi_buffs:
 	 */
 
 	/* Here we remove remaining head and trail bytes, if any */
-	if (head) {
+	if (likely(head)) {
 		ret = ss_skb_chop_head_tail(NULL, skb_hd, head, 0);
 		if (unlikely(ret))
 			return ret;
 	}
-	if (trail)
+	if (likely(trail))
 		return ss_skb_chop_head_tail(NULL, skb, 0, trail);
 
 	return 0;
--- a/fw/ss_skb.c
+++ b/fw/ss_skb.c
@@ -1009,7 +1009,7 @@ ss_skb_chop_head_tail(struct sk_buff *skb_head, struct sk_buff *skb,
 }
 
 /**
- * Chop @head and @tail bytes at head and end of the message (skb circular
+ * Chop @head and @trail bytes at head and end of the message (skb circular
  * list), iterating over the list if nessessary.
  * @return 0 on success and -(error code) on error.
  *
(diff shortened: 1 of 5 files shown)


fw/ss_skb.c
```diff
@@ -1008,6 +1008,94 @@ ss_skb_chop_head_tail(struct sk_buff *skb_head, struct sk_buff *skb,
...
+				head = 0;
+				goto trail_only;
+			}
+			return 0;
+		}
+		head -= skb->len;
+		/* We do not use ss_skb_unlink() here to prevent it
+	 	* to remove the last skb in the list and to skip
+	 	* some actions using our apriori knowledge
+	 	*/
```
krizhanovsky, line 1064: I stuck again on the code, why we don't use `ss_skb_unlink()` in particular. I.e. why it's bad to have empty list after its call. In https://github.com/tempesta-tech/tempesta/pull/1534#discussion_r765725584 I commented that `ss_skb_unlink()` just can silently make the skb list empty and we'll need to check this along with other extra checks in `ss_skb_unlink()`. Please improve the comment to make it more clear. But it's not necessary to duplicate the comment in single small function twice.

fw/ss_skb.c
```diff
@@ -1008,6 +1008,94 @@ ss_skb_chop_head_tail(struct sk_buff *skb_head, struct sk_buff *skb,
...
+single_buff:
+	/* There is the only 1 buffer in the list */
+	sum = head + trail;
+	if (WARN_ON(skb->len <= sum))
+		return -EBADMSG;
+	if (sum == 0)
+		/* Nothing to chop */
+		/* This check is mostly for jumps from branches below */
+		return 0;
+	return ss_skb_chop_head_tail(skb, skb, head, trail);
```
krizhanovsky, line 1046: Calling `ss_skb_chop_head_tail()` only once looks like a good optimization: we really can expect that in most the cases we'll have only one skb here.

fw/ss_skb.c
```diff
@@ -1008,6 +1008,94 @@ ss_skb_chop_head_tail(struct sk_buff *skb_head, struct sk_buff *skb,
...
+	 	* some actions using our apriori knowledge
+	 	*/
+		skb->next->prev = skb->prev;
+		skb->prev->next = skb->next;
+		*skb_list_head = skb_hd = skb->next;
+		__kfree_skb(skb);
+		skb = skb_hd;
+		if (unlikely(skb->next == skb))
+			goto single_buff;
+	} while (head);
```
krizhanovsky, line 1072: Each loop you check two conditions - why not to rewrite the look with precondition? Just like it's done in `tfw_tls_chop_skb_rec()`?

fw/ss_skb.c
```diff
@@ -1008,6 +1008,94 @@ ss_skb_chop_head_tail(struct sk_buff *skb_head, struct sk_buff *skb,
...
+	if (!trail)
+		return 0;
```
krizhanovsky, line 1074: We don't need the check if we rewrite the loop for `trail` in precondition form

krizhanovsky (changes requested):
The main issue with the PR is that now we have many places, which I enumerated in #1535, with very similar logic. Even worse that the logic isn't just copy & paste, but logically the same. This makes the whole code harder to maintain. This must be fixed. The code should unified among several places, e.g. the new `ss_skb_list_chop_head_tail()` and `tfw_tls_chop_skb_rec()`.

I thought that you'd just reuse some code, e.g. from `tfw_tls_chop_skb_rec()` with some adjustments, e.g. replacing `ss_skb_unlink()` with our own code.

However, the optimization for single skb case at the beginning of the function looks good. But it can be done just by adding the check between the two loops in `tfw_tls_chop_skb_rec()` and the `ss_skb_chop_head_tail()` call.

fw/ss_skb.c
```diff
@@ -1008,6 +1008,94 @@ ss_skb_chop_head_tail(struct sk_buff *skb_head, struct sk_buff *skb,
...
+
+single_buff:
+	/* There is the only 1 buffer in the list */
+	sum = head + trail;
+	if (WARN_ON(skb->len <= sum))
+		return -EBADMSG;
+	if (sum == 0)
+		/* Nothing to chop */
+		/* This check is mostly for jumps from branches below */
+		return 0;
```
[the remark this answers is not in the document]
Dmitry-Gouriev: @krizhanovsky 
The check seems redundant. However skipping the funct call if possible will be faster.
I believe this redundancy is reasonable.

fw/ss_skb.c
```diff
@@ -1008,6 +1008,94 @@ ss_skb_chop_head_tail(struct sk_buff *skb_head, struct sk_buff *skb,
...
+	return ss_skb_chop_head_tail(skb, skb, head, trail);
+
+head_and_probably_trail:
+	do {
+		if (likely(skb->len > head)) {
+			R = ss_skb_chop_head_tail(skb, skb, head, 0);
+			if (unlikely(R < 0))
+				return R;
+			if (unlikely(trail)) {
+				head = 0;
```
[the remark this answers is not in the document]
Dmitry-Gouriev: @krizhanovsky  True. But I still want to keep sum computation (see above).

fw/ss_skb.c
Dmitry-Gouriev: Seen. Thinknig...

Dmitry-Gouriev: Probably we can use skb_unlink(), at least now, when we guarantee that the list does not contain the only one skb/

fw/ss_skb.c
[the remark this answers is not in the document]
krizhanovsky: Yes, we definitely should have your optimization with only one `ss_skb_chop_head_tail()` call for one skb case. However, if we see at `tfw_tls_chop_skb_rec()` it does the things with two loops:
```C
while (unlikely(skb->len <= off)) {
   // ...
}

while (unlikely(skb->len <= tail)) {
   // ....
}

ss_skb_chop_head_tail();
```
So there is essentially just two and only two checks for the head and tail and on the fast path (with only one skb) we have only these 2 checks. If we have you optimization between the loops and the two `ss_skb_chop_head_tail()` calls (which we have in current `tfw_tls_chop_skb_rec()`, then won't such code work?

fw/ss_skb.c
Dmitry-Gouriev: @krizhanovsky 
The initail version of the func has exactly the procondition.
I changed it to poscondition for optimization purpose.
We enter the loop with knowledge that `head` is positive and I do not want to check it once more time.

fw/ss_skb.c
krizhanovsky: Yeah, need to handle the question carefully - the code is crucial and we need to accurately rework it and properly document with comments

fw/ss_skb.c
krizhanovsky: In `tfw_tls_chop_skb_rec()` we have only one `off` (`head` in your code) check, in the loop condition - I don't see why can't we construct this function in similar way and just add the optimization with one skb

fw/ss_skb.c
Dmitry-Gouriev: @krizhanovsky 
But I still believe in the postcondition form.
In general - there are many redundant condition in the functions - textually,
but they are not redundant in run.

In particular, there is initialization action
```
skb_hd = *skb_list_head;
```
which should be done before loop and so independent of pre/post condition result,
currently I avoid it.

In even more general, we can produce more fast code or more maintenable code, the decision.
Here I tried to follow traditions shown in other funcs in ss_skb.c.

___
I see more comments on code comparison which do not appear here.
Moving to them...

fw/ss_skb.c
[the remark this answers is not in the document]
Dmitry-Gouriev: @krizhanovsky Well, I need some time to think on it.

fw/ss_skb.c
Dmitry-Gouriev: Well, I'll try to make smth similar.

Dmitry-Gouriev: @krizhanovsky 
It seems we are to check the second condition `skb->next == skb` anyway, at least if we want to route the run to single skb case when appropriate and to keep the skb containing several skbs.

Dmitry-Gouriev: @krizhanovsky Again, looking into the  `tfw_tls_chop_skb_rec() ` I found sone strange thing.
```C
	skb = data->skb->prev;
	while (unlikely(skb->len <= tail)) {
		tail -= skb->len;
		ss_skb_unlink(&data->skb, skb);
		if (WARN_ON_ONCE(!data->skb))
			return -EIO;
		__kfree_skb(skb);
		skb = data->skb->prev;
	}
```
It seems the `__kfree_skb(skb);` call should be 2 lines upper, otherwise here is a memory leak on error condition raise.

Dmitry-Gouriev: > It seems we are to check the second condition skb->next == skb anyway

As in `tfw_tls_chop_skb_rec()` we are to check the second condition `WARN_ON_ONCE(!skb)` (or `(WARN_ON_ONCE(!data->skb))` in tail loop).

And it is realy the same condition.
The difference is the only that `ss_skb_list_chop_head_tail()` check condition _before_ "the fault" trying to prevent it whereas `tfw_tls_chop_skb_rec()` check yhe condition _after_ "the fault" to produce correct reaction and indication.

Dmitry-Gouriev: > Again, looking into the tfw_tls_chop_skb_rec() 

Fixed

Dmitry-Gouriev:
@krizhanovsky 
I borrowed an idea from `tfw_tls_chop_skb_rec()` and in this way have reduced a number of condition checks per loop from 3 to 2. I believe, in can not be lower.

I decided to keep `single_buf:` branch as is.

Dmitry-Gouriev:
Still remains a question to test chopping from tail and both from head and tail.
I still do not understand is this required.
And I still have no idea how to.
Probably I should create a kind of unit test?

Anyway the programming itself is near to finish however testing will take additional time.

fw/ss_skb.c
```diff
@@ -1008,6 +1008,94 @@ ss_skb_chop_head_tail(struct sk_buff *skb_head, struct sk_buff *skb,
...
+ * otherwise an error is retuned.
+ */
+int
+ss_skb_list_chop_head_tail(struct sk_buff **skb_list_head, 
+			   size_t head, size_t trail)
+{
+
+	struct sk_buff *skb, *skb_hd;
+	size_t sum;
+	int R;
```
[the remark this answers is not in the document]
Dmitry-Gouriev: Fixed

fw/ss_skb.c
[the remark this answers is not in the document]
Dmitry-Gouriev: Fixed in new version

fw/ss_skb.c
Dmitry-Gouriev: Fixed in new version

fw/ss_skb.c
Dmitry-Gouriev: > Please improve the comment to make it more clear.

I have tried

Dmitry-Gouriev:
> The main issue with the PR is that now we have many places, which I enumerated in #1535, with very similar logic. Even worse that the logic isn't just copy & paste, but logically the same. This makes the whole code harder to maintain. This must be fixed. The code should unified among several places, e.g. the new ss_skb_list_chop_head_tail() and tfw_tls_chop_skb_rec().

I think the way could be to debug end test the new `ss_skb_list_chop_head_tail()` thorously and then replace all similar code parts with calls of the function. However all these replacements must be thorously tested as well (probably it can
be the same tests).

My initial understanding of the #1535 was just like this idea.
However the timing of this work seems to exceed 5 days.
The most time-cost part is, of course, testing rather than programming.

fw/ss_skb.c
krizhanovsky: > It seems the __kfree_skb(skb); call should be 2 lines upper, otherwise here is a memory leak on error condition raise.

Good catch!

fw/ss_skb.c
```diff
@@ -1008,6 +1008,108 @@ ss_skb_chop_head_tail(struct sk_buff *skb_head, struct sk_buff *skb,
...
+	if (unlikely(sum == 0))
+		/* Nothing to chop */
+		/* This check is mostly for jumps from branches below */
+		return 0;
+	return ss_skb_chop_head_tail(NULL, skb_hd, head, trail);
+
+multi_buffs:
+	/* skb_list contains more than 1 skb &&
+	 * skb_hd points to head element of the list
+	 */
```
krizhanovsky, line 1054: I think there is no sense to copy and paste the comment 3 times: just comment it in one place and reader will get the idea for the other places

krizhanovsky (commented):
The new `ss_skb_list_chop_head_tail()` looks good for me, just several minor cleanups are required.

IIRC you already have the test which revealed #1535, right? `tfw_tls_chop_skb_rec()` is also very easy to test - any TLS record is truncated at the beginning and end for about 16 bytes. The only possible thing to do extra is to send a TLS record with about 5-byte chunks to make the function firstly delete couple of skbs and next remove the rest of 1 byte. Probably this can be done with existing tempesta-test/tls tests and you even don't need to employ the new features of #115.

In #1535 we have the list of only 3 functions:

1. `ss_tcp_process_skb()` should not be affected by the patch
2. `tfw_tls_chop_skb_rec()` should call the new `ss_skb_list_chop_head_tail()`  and is easy to test
3. `tfw_h2_frame_process()` is about buggy http/2, which we'll deal a lot in following next weeks. Also if you have well tested `tfw_tls_chop_skb_rec()` for the TLS and purge code, then it's also work good for HTTP/2.

Unit tests with sk_buff are crazy: we used to live with them and it was quite a challenge to mock all the sk_buff fields and related kernel functions.

P.S. We do care about the code quality more than the development time ;)

fw/ss_skb.c
Dmitry-Gouriev: @krizhanovsky 
These comments are not exactly the same.
The main idea is that they are formal preconditions for the code below.
They make the future maintenance of the code more easy, including other programmers.
So let they remain as is.

fw/ss_skb.c
```diff
@@ -1008,6 +1008,108 @@ ss_skb_chop_head_tail(struct sk_buff *skb_head, struct sk_buff *skb,
...
+			   size_t head, size_t trail)
+{
+	struct sk_buff *skb, *skb_hd;
+	size_t sum;
+	int ret;
+
+	skb_hd = *skb_list_head;
+	if (likely(skb_hd->next == skb_hd))
+		goto single_buff;
+	goto multi_buffs;
```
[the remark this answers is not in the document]
Dmitry-Gouriev: Fixed

Dmitry-Gouriev:
@krizhanovsky 
In reply to https://github.com/tempesta-tech/tempesta/pull/1552#pullrequestreview-865169136 

If I understand correctly,
- the `ss_tcp_process_skb()` should not patched with `ss_skb_list_chop_head_tail()` at all,
- the patch for `tfw_h2_frame_process()` is defered until you (with @byko3y ?) fix more serious bugs in it,
- and the only patch of `tfw_tls_chop_skb_rec()` with `ss_skb_list_chop_head_tail()` is desired for now.
Is this true?

Of course tesing is requred.
I have tested with predesigned tests for patches which I just have done, i.e. for gone `tfw_h1_chop_leading_crlf()` and for `tfw_h1_rewrite_purge_to_get()`.

What about tests for `tfw_tls_chop_skb_rec()`...
I guess somewhere the test for it exists.
However I still not know where it is.

Dmitry-Gouriev:
> `tfw_tls_chop_skb_rec()` is also very easy to test - any TLS record is truncated at the beginning and end for about 16 bytes. The only possible thing to do extra is to send a TLS record with about 5-byte chunks to make the function firstly delete couple of skbs and next remove the rest of 1 byte. Probably this can be done with existing tempesta-test/tls tests and you even don't need to employ the new features of #115.

Thinking on it.

krizhanovsky:
> the ss_tcp_process_skb() should not be patched with ss_skb_list_chop_head_tail() at all,

Yes, because `ss_tcp_process_skb()` is always called for single skb

> the patch for tfw_h2_frame_process() is defered until you (with @byko3y ?) fix more serious bugs in it,

Please call `ss_skb_list_chop_head_tail(&h2->skb_head, h2->data_off, 0);` in `tfw_h2_frame_process()` and remove the code, which is currently duplicated with the call. There is no need to test the function, since we'll test and fix h2 code in the near weeks.

> and the only patch for tfw_tls_chop_skb_rec() with ss_skb_list_chop_head_tail() is desired for now.

yes, `tfw_tls_chop_skb_rec()` seems will be just a tiny wrapper for `ss_skb_list_chop_head_tail()`

> What about tests for tfw_tls_chop_skb_rec()...
> I guess somewhere the test for it exists.

Just run the tests from `tempesta-test/tls` - I believe they cover the code. Basically, TLS record with an application data will test the function call. Just to be more confident that the fix is good, I'd just recommend to deploy our web site under Tempesta FW with https:// and browse it a little bit using normal browser.

Dmitry-Gouriev:
@krizhanovsky I found that `ss_tcp_process_skb()` calls `ss_skb_unroll()` wich converts a single skb into skb list and tnen implements a scenario similar to `ss_skb_list_chop_head_tail(skb_list, offset, 0)`. So there is a reason to replace it's part with `ss_skb_list_chop_head_tail()` as well as in other places.
Hosever I'd better leave this function untouched because it is more crucial and is more difficult to test than other candidates for replaements.

Dmitry-Gouriev:
@krizhanovsky  on `tfw_h2_frame_process()`
There is a special check
```C
			/*
			 * Special case when the frame is postponed just
			 * in the beginning of the app data, after all
			 * frame header fields processed.
			 */
			if (!h2->skb_head) {
				WARN_ON_ONCE(h2->data_off);
				return T_OK;
			}
```
(lines 1777-1785 at master)
With poor knowledge of HTTP2 I can not estimate an importance of this check.
Of course, it is not and should not implemented in more general purpose `ss_skb_list_chop_head_tail()`.
So there are 2 alternatives:
- to drop the check at all
- to implement it in a separate preprocessing loop.
The 1st lead to semantic changes and the 2nd have negative impact on performace.
And of course there is a way to leave `tfw_h2_frame_process()` as is.
A suggestion needed.

krizhanovsky:
> I found that ss_tcp_process_skb() calls ss_skb_unroll() wich converts a single skb into skb list and tnen implements a scenario similar to ss_skb_list_chop_head_tail(skb_list, offset, 0).

There are a lot of other things done inside the loop over the skb list and removing the overlapping TCP segment is one of the smallest. Also we call `ss_skb_list_chop_head_tail()` for the application data and non zero `sum` is quite likely, form the other hand overlapping TCP segments is quite unlikely. The functions are very different actually.

Regarding `tfw_h2_frame_process()` let's leave it as is. Maybe just add words why don't we use `ss_skb_list_chop_head_tail() into the comment for the check.

krizhanovsky (approved):
LGTM

PR #1553, opened 2022-02-01
Make adx and bmi2 non-mandatory, disable tls build in absence

ttaym:
I have not so ancient, but relatively old hardware.

My processor family is IvyBridge. Wiki says that it is not supported and tempesta build fails.

But it seems that it is strongly requires only for Tempesta TLS. Just made simple PR that render adx and bmi2 feature non-mandatory, but instead of failure disables Tempesta TLS build in hope that Tempesta FW is usable without TLS too.

This PR is WIP. I'll update and finish it if it make sense overall.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,26 +96,38 @@ ifneq (, $(findstring bmi2, $(PROC)))
 	BMI2 = "y"
 	TFW_CFLAGS += -DBMI2=1
 else
-	ERROR = "BMI2 CPU extension is required for Tempesta TLS"
+	WARNING = "BMI2 CPU extension is required for Tempesta TLS. Disabling it"
 endif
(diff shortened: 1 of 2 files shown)


krizhanovsky:
We had requests to support pre-Broadwell hardware, e.g. https://github.com/tempesta-tech/tempesta/issues/1515 . The most modern CPU extensions are required for TLS crypto, so the PR makes sense just to build a Tempesta FW on such hardware. However, since TLS is mandatory nowadays even in microservices communications inside single server, it seems that such build is just a toy.

The most my concern about compatibility issues is that we'll need to support more code or even two versions of mathematics if we go with a proper implementation of cryptography for the pre-Broadwell CPUs.

PR #1554, opened 2022-02-02
Ethtool output parsing fix #1542

### LINKED ISSUE #1542: tfw_lib.sh: handle the new ethtool output
## Scope
```
# ./scripts/tempesta.sh --restart
Stopping Tempesta...
...unload Tempesta modules
Un-loading Tempesta kernel modules...
rmmod: ERROR: Module tempesta_fw is not currently loaded
rmmod: ERROR: Module tempesta_db is not currently loaded
rmmod: ERROR: Module tempesta_tls is not currently loaded
...revert irqbalance config
done
Starting Tempesta...
./scripts/tfw_lib.sh: line 156: [: n/a: integer expression expected
...enable RPS on ens3
...load Tempesta modules
```

The thing is that with Debian 11 `ethtool` prints `n/a` instead of the number (I reproduce it in my KVM):
```
# ethtool -l ens3
Channel parameters for ens3:
Pre-set maximums:
RX:		n/a
TX:		n/a
Other:		n/a
Combined:	1
Current hardware settings:
RX:		n/a
TX:		n/a
Other:		n/a
Combined:	1
```

## Testing

Just fix the script and make sure that ethtool output is handled correctly.

ttaym:


scripts/tfw_lib.sh
```diff
@@ -151,8 +151,9 @@ tfw_set_net_queues()
 	cpu_mask=$(perl -le 'printf("%x", (1 << '$CPUS_N') - 1)')
 
 	for dev in $devs; do
-		queues=$(ethtool -l $dev 2>/dev/null \
+		queues_str=$(ethtool -l $dev 2>/dev/null \
 				| grep -m 1 RX | sed -e 's/RX\:\s*//')
+		queues=$(printf '%d' "$queues_str")
```
krizhanovsky, line 156: We still have error from `printf()` this time
```
./scripts/tfw_lib.sh: line 156: printf: n/a: invalid number
```
You can check the script even manually - just run the commands by hands

krizhanovsky (changes requested):
The problem is still here

ttaym:
@krizhanovsky 
This message from `printf` is not preventing script from functioning. Can be safely redirected to `/dev/null` (done already).

`Printf` give `0` if argument seems to be not integer, that is right thing for script logic.

krizhanovsky (approved):
LGTM. I tested that the patch fixes #1542

PR #1555, opened 2022-02-02
Fix #1473

### LINKED ISSUE #1473: Incorrect client port in the TFW_WITH_ADDR_FMT, if multiple listen ports are configured
## Scope

When a new client connection is accepted, a new `TfwClient` instance is created. `TfwClient` has an `TfwAddr` attribute which describes client address. But with a little detail: ip address of the `addr` field is set to client source address, but listen port address is used in the `addr` (which is client's destination port). Thus, client with ip `192.168.1.1` accepted on port `443` will have an `TfwAddr` equals to `192.168.1.1:443`.

When a request from the client is filtered, or an error happens, the following message is written to the log:
```
[tempesta fw] Warning: parsed request has been filtered out: 192.168.1.1:443
``` 

It's a bit confusing, since it really means `request  from client 192.168.1.1 received on port 443 has been filtered out`. But this is not the only problem. Imagine, multiple ports are configured to listen:
```
listen 80;
listen 443;
listen 8081;
```

When a client opens a new connection, a new `TfwClient` instance is created  with the corresponding listen port, say  443. But all the concurrent connections to _other_ listen ports will share the same `TfwClient` instance and the port accessed on the first try. E.g. if an error occurred on a new connection to port `80` the following message with the port `443` will be logged:
```
[tempesta fw] Warning: parsed request has been filtered out: 192.168.1.1:443
``` 

Since a `TfwClient` instance is stored in the tdb for some time even after all the client connections was closed, a new connection to previously unused port (`8081` from the configuration above) will still address previous port number (`443` in our example)

dmpetroff:
Fixed `ss_getpeername` so it uses source IP and port now.

`TfwClient` is cached across multiple connections and its `addr` field contains address with port from the first encounter of that client. This means that if we use said address for printing, port part will eventually get outdated. This is why we disabling client port printing from `tfw_http_cli_error_resp_and_log()`.

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -1310,7 +1310,7 @@ tfw_http_send_resp(TfwHttpReq *req, int status, const char *reason)
 {
 	if (!(tfw_blk_flags & TFW_BLK_ERR_NOLOG)) {
 		T_WARN_ADDR_STATUS(reason, &req->conn->peer->addr,
-				   TFW_WITH_PORT, status);
+				   TFW_NO_PORT, status);
 	}
 
 	if (TFW_MSG_H2(req))
@@ -4595,7 +4595,7 @@ static void
 tfw_http_conn_error_log(TfwConn *conn, const char *msg)
 {
 	if (!(tfw_blk_flags & TFW_BLK_ERR_NOLOG))
-		T_WARN_ADDR(msg, &conn->peer->addr, TFW_WITH_PORT);
+		T_WARN_ADDR(msg, &conn->peer->addr, TFW_NO_PORT);
 }
 
 static void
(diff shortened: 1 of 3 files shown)


fw/http.c
```diff
@@ -4746,8 +4746,10 @@ tfw_http_cli_error_resp_and_log(TfwHttpReq *req, int status, const char *msg,
 		nolog = tfw_blk_flags & TFW_BLK_ERR_NOLOG;
 	}
 
+	/* Do not log client port as it doesn't provide useful information
+	 * and could contain outdated cached data (see #1473). */
 	if (!nolog)
-		T_WARN_ADDR(msg, &req->conn->peer->addr, TFW_WITH_PORT);
+		T_WARN_ADDR(msg, &req->conn->peer->addr, TFW_NO_PORT);
```
krizhanovsky, line 4752: That's not the only place of printing with the client port, e.g. we also have similar printing in the warning at `tfw_http_req_fwd_send()`. There are also other macros with similar logic, e.g. `T_WARN_ADDR_STATUS()`. All the places must be reviewed whether it makes sense to print the port (e.g. there is an error message when we can't open listening port - we definitely need port in the message) and fix all the places.

krizhanovsky (changes requested):
Need to review other usages of similar macros printing client port.

dmpetroff:
Carefully looked through code for `TFW_WITH_PORT` and replaced it to `TFW_WITHOUT_PORT` where appropriate.

Finding out how to run tests to make sure that changes haven't broke anything else.

krizhanovsky (approved):
LGTM

PR #1556, opened 2022-02-02
Add runtime check for not present adx, may be false negative

ttaym:
There are some cloud providers that do support adx instructions in their VPS but do not show this support in /proc/cpuinfo.

I suppose there is need for runtime check of supporting theas instruct if their support not present in /proc/cpuinfo by running binary that contain adcx instruction for example emmited by inline asm.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,9 +104,11 @@ else
 ifdef M
 	DIR = $(M)/
 endif
-	# Some cloud providers hide ADX support bit in vCPU, but it still present, make run-rime check to discard false negative cases
+	# Some cloud providers hide ADX support bit in vCPU, but it still present,
+	# make run-rime check to discard false negative cases
 	CHECK_CONF = $(DIR)scripts/check_conf.pl
-	ADX_SUPPORTED := $(shell $(CHECK_CONF) 2>/dev/null | grep ADX | if grep -q ': found'; then echo y; fi)
+	ADX_SUPPORTED := $(shell $(CHECK_CONF) 2>/dev/null | \
+	grep ADX | if grep -q ': found'; then echo y; fi)
 ifeq ($(ADX_SUPPORTED), y)
 	ADX = "y"
 else
--- a/Makefile
+++ b/Makefile
@@ -100,9 +100,21 @@ else
 endif
 ifneq (, $(findstring adx, $(PROC)))
 	ADX = "y"
-	TFW_CFLAGS += -DADX=1
+else
+ifdef M
+	DIR = $(M)/
+endif
+	# Some cloud providers hide ADX support bit in vCPU, but it still present, make run-rime check to discard false negative cases
+	CHECK_CONF = $(DIR)scripts/check_conf.pl
+	ADX_SUPPORTED := $(shell $(CHECK_CONF) 2>/dev/null | grep ADX | if grep -q ': found'; then echo y; fi)
+ifeq ($(ADX_SUPPORTED), y)
+	ADX = "y"
 else
 	ERROR = "ADX CPU extension is required for Tempesta TLS"
+endif
+endif
+ifeq ($(ADX),y)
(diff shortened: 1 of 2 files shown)


check/Makefile
```diff
@@ -0,0 +1,2 @@
+clean:
+	find . \( -name \*.exe -exec rm -f {} \; \)
```
krizhanovsky, line 2: we don't have exe's

Makefile
```diff
@@ -101,9 +101,24 @@ endif
 ifneq (, $(findstring adx, $(PROC)))
 	ADX = "y"
 	TFW_CFLAGS += -DADX=1
+else
+ifdef M
+	DIR = $(M)/
+endif
+	ADX_TEST_SRC = $(DIR)check/adx_test.c
+	ADX_TEST_BIN = $(DIR)check/adx_test.exe
```
krizhanovsky, line 109:
```suggestion
	ADX_TEST_BIN = $(DIR)check/adx_test
```

Makefile
```diff
@@ -101,9 +101,24 @@ endif
...
+	ADX_BUILD_OUTPUT := $(shell gcc -madx $(ADX_TEST_SRC) -o $(ADX_TEST_BIN))
+	ADX_SUPPORTED := $(shell $(ADX_TEST_BIN) | if grep -q "ADX supported"; then echo -n "y"; fi)
```
krizhanovsky, line 111: We already have `scripts/check_conf.pl` for this purpose, so I'd suggest to move the compilation into the script. We also can embed `adx_test.c` code right into the script and compile it in `/tmp` with a generated file name and remove it after the check.

The Makefile can run the `scripts/check_conf.pl` to make sure that configuration is good.

krizhanovsky (changes requested):
Basically good to be merged after several improvements

ttaym:
@krizhanovsky 
I have made all requested changes and resolve all discusstions.

Makefile
```diff
@@ -100,10 +100,22 @@ else
...
 ifneq (, $(findstring adx, $(PROC)))
 	ADX = "y"
-	TFW_CFLAGS += -DADX=1
+else
+ifdef M
+	DIR = $(M)/
+endif
+	# Some cloud providers hide ADX support bit in vCPU, but it still present, make run-rime check to discard false negative cases
+	CHECK_CONF = $(DIR)scripts/check_conf.pl
+	ADX_SUPPORTED := $(shell $(CHECK_CONF) 2>/dev/null | grep ADX | if grep -q ': found'; then echo y; fi)
```
krizhanovsky, line 109: Too wide lines, we keep lines in 80 characters long https://github.com/tempesta-tech/tempesta/blob/master/CodingStyle

scripts/check_conf.pl
```diff
@@ -18,12 +18,37 @@
...
+  __asm__ __volatile__("adcx %%rbx,%%rax" : "=a"(op1) : "a"(op1), "b"(op2));
+  printf("%s\n", "ADX supported");
+}|;
+    my $tmp_src = File::Temp->new(SUFFIX => '.c');
+    print $tmp_src $adx_test_src;
+    my $tmp_exe_filename = $tmp_src->filename;
+    $tmp_exe_filename =~ s{\.[^.]*(?:\.c)?$}{};
+    my $exit_status = system("gcc", $tmp_src->filename, "-o", $tmp_exe_filename);
+    my $test_output = `$tmp_exe_filename`;
+    print "'Intel ADX Instruction Extensions' support: ", index($test_output, "ADX supported") == -1 || $exit_status != 0 ? "NOT " : "", "found\n";
```
krizhanovsky, line 47: Also too wide line. And please update the copyright year for the file to 2022.

krizhanovsky (approved):
LGTM after minor cleanups.

PR #1557, opened 2022-02-02
Fixed memory leak in tfw_tls_msg_process()

Dmitry-Gouriev:
It seems there is a memory leak in the function `tfw_tls_msg_process()` (file `fw/tls.c`):
```C
		r = tfw_tls_chop_skb_rec(tls, tls->io_in.skb_list, &data_up);
		if (r) {
			spin_unlock(&tls->lock);
			return r;
		}
```
(lines 189-193 at master).
On error code returned from `tfw_tls_chop_skb_rec()` neither skbs in `tls->io_in.skb_list` nor `nskb` are freed.

The PR contains proposal to fix this bug.

The changes were tested against `tempesta-test/tls` test suite just like preceding PR #1552

### DIFF
--- a/fw/tls.c
+++ b/fw/tls.c
@@ -49,20 +49,6 @@ static struct {
 /* Temporal value for reconfiguration stage. */
 static bool allow_any_sni_reconfig;
 
-/**
- * Chop skb list with begin at @skb by TLS extra data at the begin and end of
- * the list after decryption and write the right pointer at the first skb and
- * offset to @data for upper layers processing.
- */
-static int
-tfw_tls_chop_skb_rec(TlsCtx *tls)
-{
-	size_t off = ttls_payload_off(&tls->xfrm);
-	size_t tail = TTLS_TAG_LEN;
-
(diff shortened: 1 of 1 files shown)


fw/tls.c
```diff
@@ -186,12 +178,15 @@ tfw_tls_msg_process(void *conn, TfwFsmData *data)
 		 * Pass tls->io_in.skb_list to data_up ownership for the upper
 		 * layer processing.
 		 */
-		r = tfw_tls_chop_skb_rec(tls, tls->io_in.skb_list, &data_up);
+		r = tfw_tls_chop_skb_rec(tls);
```
krizhanovsky, line 181: Do we need the function at all? It seems we can call `ss_skb_list_chop_head_tail()` just here

krizhanovsky (approved):
Good catch! LGTM

fw/tls.c
Dmitry-Gouriev: @krizhanovsky Dropped

PR #1566, opened 2022-02-15
Corrected fixups for cache-control and authority

byko3y:
Extracted from #1562

Also added more headers into fills_hdr_tbl_for_resp test

__req_parse_cache_control() bug:

__FSM_I_MATCH_MOVE and __FSM_I_MATCH_MOVE_fixup only differ in calling
__msg_hdr_chunk_fixup() with data + len and p + __fsm_size
correspondingly. This creates a problem when p != data e.g. we are
parsing multiple tokens.
The whole __req_parse_cache_control() function normally uses
__FSM_I_MATCH_MOVE, thus doing lazy fixups over complete chunks with
TFW_POSTPONE return value and the final fixup is performed by parent
TFW_HTTP_PARSE_RAWHDR_VAL with return value greater or equal 0.

The fills_hdr_tbl_for_req test shows a mishandling of Cache-Control
header value when it's equal to

Cache-Control: max-age=1, dummy, no-store, min-fresh=30

Let's suppose the string is chunked into

max-age=1, dum
my, no-store, min-fresh=30

blocks. The first chunk will be processed via labels

 m  -> Req_I_CC_start, Req_I_CC
max-age=  -> Req_I_CC_m, Req_I_CC_MaxAgeVBeg
 1  -> Req_I_CC_MaxAgeVBeg, Req_I_CC_MaxAgeV
 ,  -> Req_I_EoT
" " -> Req_I_After_Comma
 d  -> Req_I_CC
dum -> Req_I_CC_Ext (returns TFW_POSTPONE)

When the wrong fixup macro is used, the parsing of the first chunk
will reach the Req_I_CC_Ext label and trigger the __fsm_sz == __fsm_n
condition (chunk exhaustion) in the __FSM_I_MATCH_MOVE_fixup, thus
doing a __msg_hdr_chunk_fixup("dum", 3) fixup and returning
TFW_POSTPONE. This way the starting portion of the first chunk will
never be fixed up. Parsing of the second chunk, will be resumed on
the same __FSM_I_MATCH_MOVE_fixup, which will effectively return a
__fsm_sz value and continue normal processing in the
__req_parse_cache_control().

This way the final value for msg->stream->parser.hdr string will be

Cache-Control: dummy, no-store, min-fresh=30

Which might look fine, aprat from the missing starting portion.

__h2_req_parse_authority() bug:

This function has an inverse problem: a lazy __FSM_H2_I_MOVE_n_flag
fixup is used instead of explicit __FSM_H2_I_MOVE_fixup.

Suppose the header is

Authority: [::1]:5001

and its value is chunked as

[:
:1]:5001

The processing will go through the following FSM states:

"[" -> Req_I_A_Start (fixups "[")
":" -> Req_I_A_v6 (fixups ???, returns TFW_POSTPONE \ CSTR_NEQ)
":" -> Req_I_A_v6 (fixups ":")
"1" -> Req_I_A_v6 (fixups "1")
"]" -> Req_I_A_v6 (fixups "]")
":" -> Req_I_A_End (fixups ":")
5000 -> Req_I_A_Port (fixups "5000", returns 0 \ CSTR_EQ)

The use of __FSM_H2_I_MOVE_n_flag macro inside Req_I_A_v6 leads to
a __msg_hdr_chunk_fixup("[:", 2) call instead of a correct
__msg_hdr_chunk_fixup(":", 1). Thus the final parsed string will look
like

Authority: [[::1]:5001

### DIFF
--- a/fw/http_parser.c
+++ b/fw/http_parser.c
@@ -5422,9 +5422,6 @@ do {									\
 #define __FSM_H2_I_MOVE_LAMBDA_n(to, n, lambda)				\
 	__FSM_H2_I_MOVE_LAMBDA_n_flag(to, n, lambda, 0)
 
-#define __FSM_H2_I_MOVE_n_flag(to, n, flag)				\
-	__FSM_H2_I_MOVE_LAMBDA_n_flag(to, n, {}, flag)
-
 #define __FSM_H2_I_MOVE_n(to, n)					\
 	__FSM_H2_I_MOVE_LAMBDA_n(to, n, {})
 
--- a/fw/http_parser.c
+++ b/fw/http_parser.c
@@ -2379,7 +2379,7 @@ __req_parse_cache_control(TfwHttpReq *req, unsigned char *data, size_t len)
 		/* Any directive we don't understand.
 		 * Here we just skip all the tokens, double quotes and equal signs.
 		 */
-		__FSM_I_MATCH_MOVE_fixup(qetoken, Req_I_CC_Ext, 0);
+		__FSM_I_MATCH_MOVE(qetoken, Req_I_CC_Ext);
 
 		__FSM_I_MOVE_n(Req_I_EoT, __fsm_sz);
 	}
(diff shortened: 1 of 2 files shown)


fw/http_parser.c
```diff
@@ -5689,7 +5689,7 @@ __h2_req_parse_authority(TfwHttpReq *req, unsigned char *data, size_t len,
 	__FSM_STATE(Req_I_A_v6) {
 		/* See Req_UriAuthorityIPv6 processing. */
 		if (likely(isxdigit(c) || c == ':'))
-			__FSM_H2_I_MOVE_n_flag(Req_I_A_v6, 1, TFW_STR_VALUE);
+			__FSM_H2_I_MOVE_fixup(Req_I_A_v6, 1, TFW_STR_VALUE);
```
krizhanovsky, line 5692: That was the only usage of `__FSM_H2_I_MOVE_n_flag()`, so we can remove the macro for now.

The other usages of `__FSM_H2_I_MOVE_LAMBDA_n_flag()` are `__FSM_H2_I_MOVE_LAMBDA_n()` and `__FSM_H2_I_MOVE_n`. `__FSM_H2_I_MOVE_LAMBDA_n()` is used in only 2 places: `__h2_req_parse_authority()` in `Req_I_A` and it fixes up a chunk via lambda, and in `__h2_req_parse_cache_control()` without any fixups. `__FSM_H2_I_MOVE_n` is used in a lot of places and it's likely that there are no fixups at all - how the things work if we reach the end of data? Don't we have the same problem in all the places? It seems we just don't set the current TfwStr chunk length in all the cases or do we?

PR #1567, opened 2022-02-15
Simple access log

dmpetroff:
Add simple access logging (see #1154)

Access logging is enabled/disabled globally via config option
```
access_log on;
```
Messages are being printed to klog with printk and truncated to 990 bytes with about ~300 bytes limit per uri/referer/user-agent.
Missing entries are replaced with `-`. In case of parser errors request method, uri, and version may be "missing separately" and resulting request line could look like `"GET /some-uri -"` or even `"- - -"`.

Access logging works for http/1.x and http/2 for frang-rejected requests too (though I fake http response status in `tfw_h2_error_resp`).

### DIFF
--- a/fw/access_log.c
+++ b/fw/access_log.c
@@ -76,6 +76,7 @@ get_http_header_value(char http_version, TfwStr *line)
 	size_t len = line->len;
 	TfwStr *chunk = line->chunks, *end = chunk + line->nchunks;
 	static TfwStr empty_hdr = TFW_STR_STRING("-");
+
 	switch (http_version) {
 	case TFW_HTTP_VER_09:
 	case TFW_HTTP_VER_10:
@@ -108,8 +109,10 @@ get_http_header_value(char http_version, TfwStr *line)
 	default:
 		return empty_hdr;
 	}
+
 	if (chunk == end)
 		return empty_hdr;
+
 	TFW_STR_INIT(&result);
 	result.len = len;
 	result.chunks = chunk;
@@ -118,7 +121,7 @@ get_http_header_value(char http_version, TfwStr *line)
 }
 
 
-/* Helpers for const=>name conversions for http methods and versions {{{ */
+/* Helpers for const=>name conversions for http methods and versions */
 static const struct {
 #	define MAX_HTTP_METHOD_NAME_LEN	10
 	char name[MAX_HTTP_METHOD_NAME_LEN];
@@ -164,7 +167,7 @@ enum {
 #undef ENUM
 #undef IGNORE
 	TRUNCATABLE_FIELDS_COUNT
-};/*}}}*/
+};
 
 struct str_ref {
 	const char *data;
@@ -219,6 +222,8 @@ process_truncated(TfwStr *in, struct str_ref *out, char *p, char *end,
 	/* Unhappy path: evenly distribute available buffer space across all
 	 * strings that do not fit */
 	buf_avail = (ACCESS_LOG_BUF_SIZE - used_chars);
+	/* This division by constant usually gets optimized by compiler with
+	 * multiplication/shifts */
 	max_untruncated_len = buf_avail / TRUNCATABLE_FIELDS_COUNT;
 	
 	for (i = 0; i < TRUNCATABLE_FIELDS_COUNT; i++) {
@@ -375,12 +380,11 @@ do_access_log_req(TfwHttpReq *req, int resp_status, unsigned long resp_content_l
 		ACCESS_LOG_LINE(FMT_FIXED, FMT_UNTRUNCATABLE, FMT_TRUNCATABLE) "\n"
 		ACCESS_LOG_LINE(ARG_FIXED, ARG_UNTRUNCATABLE, ARG_TRUNCATABLE)
 	);
-#if 0
+
 	/* Undefine all locally defined macros.
 	 * You can use following oneliner to regenerate this list
+	 * sed -nre '/^do_access_log_req/,/^\}/{s@^#[[:space:]]*define[[:space:]]*([^([:space:]]+).*@#undef \1@p}' fw/access_log.c | tac
 	 */
-sed -nre '/^do_access_log_req/,/^\}/{s/^#[[:space:]]*define[[:space:]]*([^([:space:]]+).*/#undef \1/p}' fw/access_log.c | tac
-#endif
 #undef DO_PR_INFO
 #undef ARG_TRUNCATABLE
 #undef ARG_UNTRUNCATABLE
--- a/fw/access_log.c
+++ b/fw/access_log.c
@@ -1,8 +1,7 @@
 /**
  *		Tempesta FW
  *
- * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
- * Copyright (C) 2015-2022 Tempesta Technologies, Inc.
+ * Copyright (C) 2022 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
@@ -24,10 +23,46 @@
 #include "http.h"
 #include "lib/common.h"
 
+/* This thing describes access log format.
+ * - FIXED => fixed string, passed as is
+ * - UNTRUNCATABLE => this expression will never be truncated
+ * - TRUNCATABLE => this expression will be truncated if it does not fit
+ *   log buffer
+ *
+ * !!!
+ * If you add new UNTRUNCATABLE/TRUNCATABLE field don't forget to also
+ * set appropriate variable/array value in do_access_log_req()
+ !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+ ! Adding TRUNCATABLE field will compile !
+ !  but won't work without changing the  !
+ !       code that fills its value       !
+ !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+ */
+#define ACCESS_LOG_LINE(FIXED, UNTRUNCATABLE, TRUNCATABLE) \
+	FIXED("[tempesta fw] ")                            \
+	UNTRUNCATABLE(client_ip)                           \
+	FIXED(" \"")                                       \
+	UNTRUNCATABLE(vhost)                               \
+	FIXED(" \"")                                       \
+	UNTRUNCATABLE(method)                              \
+	FIXED(" ")                                         \
+	TRUNCATABLE(uri)                                   \
+	FIXED(" ")                                         \
+	UNTRUNCATABLE(version)                             \
+	FIXED("\" ")                                       \
+	UNTRUNCATABLE(status)                              \
+	FIXED(" ")                                         \
+	UNTRUNCATABLE(content_length)                      \
+	FIXED(" \"")                                       \
+	TRUNCATABLE(referer)                               \
+	FIXED("\" \"")                                     \
+	TRUNCATABLE(user_agent)                            \
+	FIXED("\"")
+
 static bool access_log_enabled = false;
 
 /* Use small buffer because printk won't display strings longer that ~1000 bytes */
-#define ACCESS_LOG_BUF_SIZE 990
+#define ACCESS_LOG_BUF_SIZE 960
 static DEFINE_PER_CPU_ALIGNED(char[ACCESS_LOG_BUF_SIZE], access_log_buf);
 
 /** Build string consists of chunks that belong to the header value.
@@ -42,13 +77,13 @@ get_http_header_value(char http_version, TfwStr *line)
 	size_t len = line->len;
 	TfwStr *chunk = line->chunks, *end = chunk + line->nchunks;
 	static TfwStr empty_hdr = TFW_STR_STRING("-");
-
-	if (http_version >= TFW_HTTP_VER_09 && http_version <= TFW_HTTP_VER_11) {
-
+	switch (http_version) {
+	case TFW_HTTP_VER_09:
+	case TFW_HTTP_VER_10:
+	case TFW_HTTP_VER_11:
 		/* Should never get plain string from parser */
 		if (TFW_STR_PLAIN(line))
 			return TFW_STR_EMPTY(line) ? empty_hdr : *line;
-
 		/* Skip over until ':' marker */
 		for (; chunk < end; chunk++) {
 			len -= chunk->len;
@@ -58,23 +93,24 @@ get_http_header_value(char http_version, TfwStr *line)
 		if (chunk == end)
 			return empty_hdr;
 		chunk++;
-
 		/* Skip possible whitespace blocks */
 		while (chunk < end && (chunk->flags & TFW_STR_OWS) != 0) {
 			len -= chunk->len;
 			chunk++;
 		}
-	} else {
+		break;
+	case TFW_HTTP_VER_20:
 		/* skip over chunks until HDR_VALUE one */
 		while (chunk < end && (chunk->flags & TFW_STR_HDR_VALUE) == 0) {
 			len -= chunk->len;
 			chunk++;
 		}
+		break;
+	default:
+		return empty_hdr;
 	}
-
 	if (chunk == end)
 		return empty_hdr;
-
 	TFW_STR_INIT(&result);
 	result.len = len;
 	result.chunks = chunk;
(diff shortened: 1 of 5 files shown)


fw/access_log.c
```diff
@@ -0,0 +1,246 @@
...
+	CONCAT_CASE_STR_XX(TFW_HTTP_METH_POST,       "POST");
+	CONCAT_CASE_STR_XX(TFW_HTTP_METH_PROPFIND,   "PROPFIND");
+	CONCAT_CASE_STR_XX(TFW_HTTP_METH_PROPPATCH,  "PROPPATCH");
+	CONCAT_CASE_STR_XX(TFW_HTTP_METH_PUT,        "PUT");
+	CONCAT_CASE_STR_XX(TFW_HTTP_METH_TRACE,      "TRACE");
+	CONCAT_CASE_STR_XX(TFW_HTTP_METH_UNLOCK,     "UNLOCK");
+	CONCAT_CASE_STR_XX(TFW_HTTP_METH_PURGE,      "PURGE");
+	default: CONCAT_STR("-");
+#undef XX
+	}
```
krizhanovsky, line 179: I think if we place all the methods and their lengths into a static array, then we can just pick the required string and length using `req->method`. This should give us simpler and probably faster code. We can check the arrays size using static assertion `BUILD_BUG_ON()`.

fw/access_log.c
```diff
@@ -0,0 +1,246 @@
...
+	if (!TFW_STR_EMPTY(&req->uri_path))
+		CONCAT_TFW_STR(&req->uri_path, 3,
+			       8 + 2 + 10 + 1 + 20 + 2 + 3 + 1);
```
krizhanovsky, line 182: Please comment the constants or use defines instead. Also for the same call with with the constant `40` above

fw/access_log.c
```diff
@@ -0,0 +1,246 @@
...
+		while (chunk < end && (chunk->flags & TFW_STR_OWS) != 0) {
+			len -= chunk->len;
+			chunk++;
+		}
+	} else {
+		/* skip over chunks until HDR_VALUE one */
+		while (chunk < end && (chunk->flags & TFW_STR_HDR_VALUE) == 0) {
+			len -= chunk->len;
+			chunk++;
+		}
```
krizhanovsky, line 72: I guess we should do something different for `__TFW_HTTP_VER_INVALID`, probably not try to print any value. When this can happen? If it's unexpected, then it makes sense to do `WARN_ON_ONCE()` for this case.

krizhanovsky (changes requested):
I didn't see any major issues with the code, but several improvements are wished

fw/access_log.c
```diff
@@ -0,0 +1,246 @@
...
+overflow:
+	*(p < end ? p : end - 1) = 0;
+	pr_info("%s\n", buf);
+#undef CONCAT_CASE_STR_XX
+#undef CONCAT_CASE_STR
+#undef CONCAT_STR
+#undef CONCAT_PRINTF
+#undef CONCAT_HDR
+#undef CONCAT_TFW_STR
+}
```
krizhanovsky, line 210: One more wish: the access log messages must start with `[tempesta fw]` that we can grep them out

fw/access_log.c
dmpetroff: Such things only occur on malformed requests (i.e. when there was an error detected by parser before or in that field). I've decided  to go with `'-'` replacement to make it easier to understand what fields were malformed — this could be useful for debugging I think.
Ideally it would be nice to get original unmodified request line, but that doesn't seem to be easily possible.

dmpetroff:
I've addressed all problems mentioned and re-implemented access logging with some metaprogramming.

Now log line structure is set within `ACCESS_LOG_LINE` and require some minor code adjustments (to serialze new field to the string).

Also as a side-effect now there's a bit of "zero-copying" things (like plain strings are sent unmodified to printk) and this would make this code more suitable for proper logging in the future.

fw/access_log.c
```diff
@@ -0,0 +1,430 @@
...
+	XX(UNLOCK)       \
+	XX(PURGE)
+
+static const struct {
+#	define MAX_HTTP_METHOD_NAME_LEN	10
+	char name[MAX_HTTP_METHOD_NAME_LEN];
+	u8 len;
+} http_methods[] = {
+#define XX(name) [TFW_HTTP_METH_ ## name] = { #name, sizeof(#name) - 1 },
+	HTTP_METHODS(XX)
```
krizhanovsky, line 152: `HTTP_METHODS` is used only here, so there is not sense to define the macro.

`XX` is essentially a constant string initializer, so the code will be more clear with macro like `STR_INIT()` instead of `XX`.

fw/access_log.c
```diff
@@ -0,0 +1,430 @@
...
+	XX(PROPFIND)     \
+	XX(PROPPATCH)    \
+	XX(PUT)          \
+	XX(TRACE)        \
+	XX(UNLOCK)       \
+	XX(PURGE)
+
+static const struct {
+#	define MAX_HTTP_METHOD_NAME_LEN	10
+	char name[MAX_HTTP_METHOD_NAME_LEN];
```
krizhanovsky, line 148:
```suggestion
	const char *name;
```
At the below you initialize the strings with pointers to constant strings, so there is no sense to allocate fixed space for the strings

fw/access_log.c
```diff
@@ -0,0 +1,430 @@
...
+	result.nchunks = end - chunk;
+	return result;
+}
+
+
+void
+do_access_log(TfwHttpResp *resp)
+{
+	do_access_log_req(resp->req, resp->status, resp->content_length);
+}
```
krizhanovsky, line 126: We organize source files with placing callers under the callees, i.e. `do_access_log()` should be under `do_access_log_req()`.

fw/access_log.c
```diff
@@ -0,0 +1,430 @@
...
+#define ARG_FIXED(str)
+#define ARG_UNTRUNCATABLE(id) ARG_ ## id
+#define ARG_TRUNCATABLE(id) , truncated_out[idx_ ## id].len, \
+		truncated_out[idx_ ## id].data
+#ifdef XXX
+BEGIN
+#endif
+
+	/* we have to enforce some macro expansion here */
+#define DO_PR_INFO(...) pr_info(__VA_ARGS__)
```
krizhanovsky, line 362: We don't need to define a macro just for one usage. The code is already overloaded with macros and it'd be good to reduce the number of macros

fw/access_log.c
```diff
@@ -0,0 +1,430 @@
...
+ * If you add new UNTRUNCATABLE/TRUNCATABLE field don't forget to also
+ * set appropriate variable/array value in do_access_log_req()
+ !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+ ! Adding TRUNCATABLE field will compile !
+ !  but won't work without changing the  !
+ !       code that fills its value       !
+ !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+ */
+#define ACCESS_LOG_LINE(FIXED, UNTRUNCATABLE, TRUNCATABLE) \
+	FIXED("[tempesta fw] ")                            \
```
krizhanovsky, line 42: We already have `[tempesta fw]` defined as `__BNR` in `log.h`.

BTW maybe we should move the contents of access_log.h to log.h to keep all logging declarations in the same place.

fw/access_log.c
```diff
@@ -0,0 +1,430 @@
...
+	for (i = 0; i < TRUNCATABLE_FIELDS_COUNT; i++) {
+		/* we loose some chars due to string "less than", but
+		 * tfw_str_to_cstr accounts terminating NUL to total buffer
+		 * length and would truncate strings anyways */
+		if (in[i].len < max_untruncated_len)
+			total_len -= in[i].len;
+		else
+			truncated_count++;
+	}
+	max_untruncated_len = buf_avail / truncated_count;
```
krizhanovsky, line 238: division by zero is possible

fw/access_log.c
```diff
@@ -0,0 +1,430 @@
...
+	for (i = 0; i < TRUNCATABLE_FIELDS_COUNT; i++) {
+		if (in[i].len < max_untruncated_len) {
+			p = make_plain(p, end, in + i, out + i);
+		} else {
+			/* we need only part of the string + "...", so
+			 * we have to enforce string copy */
+			out[i].data = p;
+			out[i].len = max_untruncated_len;
+			tfw_str_to_cstr(in + i, p, max_untruncated_len);
+			memcpy(p + max_untruncated_len - 3, "...", sizeof("...") - 1);
```
krizhanovsky, line 251: This is relatively big string copying and we call logger from softirq (traffic processing), so it'd be beneficial to use `memcpy_fast()`

fw/access_log.c
```diff
@@ -0,0 +1,430 @@
...
+			p += max_untruncated_len;
+		}
+	}
+	return;
```
krizhanovsky, line 255: It'd be good to have an empty line here

fw/access_log.c
```diff
@@ -0,0 +1,430 @@
...
+		goto no_buffer_space;
+	/* Compute total length of all strings that can be truncated */
+	for (i = 0; i < TRUNCATABLE_FIELDS_COUNT; i++)
+		total_len += in[i].len;
+	/* Check if we're on happy path: all strings fit */
+	if (likely(total_len + used_chars < ACCESS_LOG_BUF_SIZE)) {
+		for (i = 0; i < TRUNCATABLE_FIELDS_COUNT; i++)
+			p = make_plain(p, end, in + i, out + i);
+		return;
+	}
```
krizhanovsky, line 224: It'd be good to have an empty line here

fw/access_log.c
```diff
@@ -0,0 +1,430 @@
...
+	/* Unhappy path: evenly distribute available buffer space across all
+	 * strings that do not fit */
+	buf_avail = (ACCESS_LOG_BUF_SIZE - used_chars);
+	max_untruncated_len = buf_avail / TRUNCATABLE_FIELDS_COUNT;
```
krizhanovsky, line 228: I still don't like the division, but I do like the flexible truncation much more and I don't see how to avoid the division. It's on the slow path, so I guess we're good with it for now

fw/access_log.c
```diff
@@ -0,0 +1,430 @@
...
+	struct str_ref missing = { "-", 1 };
+	TfwStr truncated_in[TRUNCATABLE_FIELDS_COUNT];
+	struct str_ref truncated_out[TRUNCATABLE_FIELDS_COUNT];
+	/* Check if logging is enabled */
+	if (!access_log_enabled)
+		return;
+	/* client_ip */
+	/* this BUG_ON would only trigger if
+	 * ACCESS_LOG_BUF_SIZE < TFW_ADDR_STR_BUF_SIZE
+	 * which should be always false */
```
krizhanovsky, line 282: Typically we write such comments like
```C
/*
 * client_ip
 *
 * This BUG_ON would only trigger if ACCESS_LOG_BUF_SIZE < TFW_ADDR_STR_BUF_SIZ
 * which should be always false.
 * /
 ```

**Review by krizhanovsky — CHANGES_REQUESTED** [review]
The access log isn't simple at all :) Very clever code with automatic generation of the format string and arguments assembling.

In general the code is good and I love the approach, but several clanups are required to make the code consistent with the rest of the project. Also a small bug must be fixed.

File: fw/access_log.c (hunk 14 above)
Comment (dmpetroff) [reply fw/access_log.c:228]: This division could be changed into multiplication by gcc: https://gcc.godbolt.org/z/e1M5ae3hf, but the second one will occur anyway.
But I don't think this would be an issue as we're printing decimal numbers which already involves multiple divisions.

File: fw/access_log.c (hunk 6 above)
Comment (dmpetroff) [reply fw/access_log.c:148]: I believe this is a bit more compact (cache-friendly) version. Also there's one indirection less to compute string address.
```c
#define ESTIMATE_STR(str) + sizeof(#str) + 8
#define ESTIMATE_INLINE(str) +10
	printf("pointers: %d\n", (int)(HTTP_METHODS(ESTIMATE_STR)));
	printf("inline:   %d\n", HTTP_METHODS(ESTIMATE_INLINE));
/* Gives
pointers: 226
inline:   160
*/
```

fw/access_log.c
dmpetroff: I've added explaining comment. The macro could be more generic, but it is absolutely necessary to expand arguments to invoke another macro. This is required only because `pr_info` is a macro itself.

fw/access_log.c
dmpetroff: Precondition that sum of `{len[i] | i=0..n-1} >= buf_avail` implies that there must be a string that is longer or equal than `buf_avail/n` => `truncated_count` is always non-zero at this point.

Note that this could be false it there was '<=' used instead of '<' inside that loop.

dmpetroff:
I've addressed mentioned problems and found one small bug, where `vhost` field wasn't enclosed into double quotes on the right side.

fw/access_log.c
dmpetroff: This optimization is enabled even on `-Og`. Here's this fragment from `objdump -S` from object file compiled with kernel cflags (which enables -O2):
```
        max_untruncated_len = buf_avail / TRUNCATABLE_FIELDS_COUNT;
 234:   41 b8 ab aa aa aa       mov    $0xaaaaaaab,%r8d
 23a:   89 c7                   mov    %eax,%edi
 23c:   49 0f af f8             imul   %r8,%rdi                    <= multiplication
                if (in[i].len < max_untruncated_len)
 240:   48 c1 ef 21             shr    $0x21,%rdi
 244:   48 39 cf                cmp    %rcx,%rdi
 247:   41 0f 96 c1             setbe  %r9b
                        truncated_count++;
 24b:   48 39 fe                cmp    %rdi,%rsi
 24e:   41 83 d9 ff             sbb    $0xffffffff,%r9d
 252:   48 39 fa                cmp    %rdi,%rdx
 255:   41 83 d9 ff             sbb    $0xffffffff,%r9d
        max_untruncated_len = buf_avail / truncated_count;
 259:   31 d2                   xor    %edx,%edx
 25b:   41 f7 f1                div    %r9d                        <= division
```

fw/access_log.c
```diff
@@ -0,0 +1,449 @@
...
+	DO_PR_INFO(
+		ACCESS_LOG_LINE(FMT_FIXED, FMT_UNTRUNCATABLE, FMT_TRUNCATABLE) "\n"
+		ACCESS_LOG_LINE(ARG_FIXED, ARG_UNTRUNCATABLE, ARG_TRUNCATABLE)
+	);
+#if 0
+	/* Undefine all locally defined macros.
+	 * You can use following oneliner to regenerate this list
+	 */
+sed -nre '/^do_access_log_req/,/^\}/{s/^#[[:space:]]*define[[:space:]]*([^([:space:]]+).*/#undef \1/p}' fw/access_log.c | tac
+#endif
```
krizhanovsky, line 383: I think we don't need the ifdef - we can do this with hands

fw/access_log.c
krizhanovsky: Ok, cool. Could you please add a comment that sometimes, not always, compiler optimizes devision to multiplication,

fw/access_log.c
```diff
@@ -0,0 +1,449 @@
...
+};
+
+enum {
+#define IGNORE(...)
+#define ENUM(id) idx_ ## id,
+	ACCESS_LOG_LINE(IGNORE, IGNORE, ENUM)
+#undef ENUM
+#undef IGNORE
+	TRUNCATABLE_FIELDS_COUNT
+};/*}}}*/
```
krizhanovsky, line 167: we don't use this (vim folding?)

fw/access_log.c
```diff
@@ -0,0 +1,449 @@
...
+ *
+ * @param http_version is TFW_HTTP_VER_xx enum
+ * @param line is the request line from TfwHttpReq::h_tbl */
+static TfwStr
+get_http_header_value(char http_version, TfwStr *line)
+{
+	TfwStr result;
+	size_t len = line->len;
+	TfwStr *chunk = line->chunks, *end = chunk + line->nchunks;
+	static TfwStr empty_hdr = TFW_STR_STRING("-");
```
krizhanovsky, line 78: please place an empty line

krizhanovsky (approved):
LGTM with minor cleanups

Issue #755, opened 2017-06-18
Websocket


krizhanovsky:
Need to support [Websocket](https://en.wikipedia.org/wiki/WebSocket) (RFCs [6455](https://tools.ietf.org/html/rfc6455)), both the *ws://* and *wss://* schemes.

# HTTP/1

## Parser extensions

- `Upgrade: websocket` header [RFC 7230 6.7](https://tools.ietf.org/html/rfc7230#section-6.7) must be implemented. We support only `websocket`, so values like `h2c` or `h2c, websocket` should be just [ignored](https://tools.ietf.org/html/rfc7540#section-3.2).
- `Upgrade` directive for `Connection` header. This must be done for requests (HTTP/1 and HTTP/2) and responses. It seems we shouldn't do anything if a server just advertises `Upgrade` w/o preceding client `Upgrade` request.

Since `Upgrade` and `Connection` is a hop by hop header, if we ignore their values, then a backend receives a request w/o the headers.

If all the headers are good and we can upgrade to websocket, then a new websocket flag for the request must be set and `tfw_h1_adjust_req()` must regenerate the headers. Note that `Sec-WebSocket-*` headers aren't hop-by-hop and shouldn't be stripped.

All the extensions must be tested for HTTP/1 in `test_http_parser.c`

## Scheme parsing

`Req_Uri` state at the moment handles only `http://` scheme, which must be fixed to allow `https://` and extend with the new `ws://` and `wss://`. Since we'll need to match 4 strings, it makes sense to introduce new states with switch on prefixes (see as headers parsing is done). It seems that browsers never (or almost) never send full URLs, so the states should be `cold` and `unlikely`.

# HTTP/2

Websockets for HTTP/2 are described in [8441](https://tools.ietf.org/html/rfc8441) and we have a very special case of proxying HTTP/2 CONNECT to HTTP/1 websocket.

Nginx just proxies [CONNECT method](https://trac.nginx.org/nginx/ticket/1992). I don't understand how reliable websockets in these setups, but there are issues in the Internet with websockets over HTTP/2 proxying by Nginx, e.g. https://github.com/RocketChat/Rocket.Chat/issues/15028 . The problem is that RFC 8441 does not require an HTTP/2 client to set `Sec-WebSocket-Key` header, but maybe the most implementations actually do set the header.

[...]

Issue #775, opened 2017-07-26
Unify the representation of IP addresses in Tempesta


keshonok:
IP addresses can be specified in the configuration either in IPv4 or in IPv6 format. In addition, in some cases they may be specified in CIDR format. The addresses are stored in Tempesta, and in some cases they are stored in Tempesta in the same format they are specified, and in some cases they are converted to IPv6. The lack of unification may lead to bugs when comparing addresses. For unification, all addresses should be converted to IPv6 representation internally in Tempesta.

There's one thing however. Sometimes addresses are printed in log messages. It's highly desirable that they are printed in the same form they are specified in configuration. That way the addresses can be visually recognized and matched by a human operator and/or system administrator.

krizhanovsky:
There is agreement that Tempesta works with IPv6 addresses only and all IPv4 addresses are converted to IPv6 (IPv4-mapped IPv6 addresses), e.g. we call `ipv6_addr_set_v4mapped()` from `tfw_addr_pton_addr()` for IPv4 addresses. @keshonok do you mean that no all IPv4 addresses are converted to IPv6 in internal Tempesta's operations?

At least logging must be fixed. For example Frang limits generate records like

        Warning: frang: Content-Type header field for ::ffff:c0a8:6401 is missed

for IPv4 clients and the records are very cryptic for a user.

krizhanovsky:
In short it seems that we just need properly write IPv6-mapped IPv4 addresses to log, this should fix the issue.

i-rinat:
Implemented in https://github.com/tempesta-tech/tempesta/pull/1081.

PR #1571, opened 2022-02-24
Websocket upgrade directive implementation

ttaym:
Contributes to #755

Considerations.

It seems that to confrom to [RFC7230#section-6.6](https://datatracker.ietf.org/doc/html/rfc7230#section-6.6) we have to be prepared to process requests with `Connection: upgrade` and `Expect: 100-continue` set at the same time and respond with 100 before going further with the request. But for websocket upgrade case it is obscure and rare (if ever) seen. For now implemented logic does not have that.

It seems that standard does not explicitly forbid `Upgrade` header in request trailer field. But for now implementation does not support that scenario. 

Tempesta FW MUST block requests with `Upgrade` header but without `upgrade` option in `Connection` header. Tempesta FW MUST ignore `Upgrade` header for HTTP version less then `HTTP/1.1`. See [RFC7230#section-6.1](https://datatracker.ietf.org/doc/html/rfc7230#section-6.1):

> When Upgrade is sent, the sender MUST also send a Connection header
   field ([Section 6.1](https://datatracker.ietf.org/doc/html/rfc7230#section-6.1)) that contains an "upgrade" connection option, in
   order to prevent Upgrade from being accidentally forwarded by
   intermediaries that might not implement the listed protocols.  A
   server MUST ignore an Upgrade header field that is received in an
   HTTP/1.0 request.

### DIFF
--- a/fw/http_limits.c
+++ b/fw/http_limits.c
@@ -565,6 +565,49 @@ frang_http_methods_override(const TfwHttpReq *req, FrangAcc *ra,
 }
 
 static int
+frang_http_upgrade_websocket(const TfwHttpReq *req, FrangAcc *ra,
+			     FrangVhostCfg *f_cfg)
+{
+	BUG_ON(!req);
+
+	switch (req->version) {
+	/*
+	 * TODO upgrade websocket checks for h2 as described in RFC8441
+	 */
+	case TFW_HTTP_VER_20:
+		break;
+	/*
+	 * Tempesta FW MUST block requests with Upgrade header but without
+	 * upgrade option in Connection header. Tempesta FW MUST ignore
+	 * Upgrade header for HTTP version less then HTTP/1.1.
+	 * See RFC7230#section-6.1.
+	 */
+	case TFW_HTTP_VER_11:
+	case TFW_HTTP_VER_10:
+	case TFW_HTTP_VER_09:
+		if (test_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET, req->flags)
+		    && !test_bit(TFW_HTTP_B_CONN_UPGRADE, req->flags))
+		{
+			frang_msg("upgrade request without connection option",
+				  &FRANG_ACC2CLI(ra)->addr, ": protocol: %s\n",
+				  "websocket");
+			return TFW_BLOCK;
+		}
+		if (req->version == TFW_HTTP_VER_10
+		    || req->version == TFW_HTTP_VER_09)
+		{
+			clear_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET,
+				  ((TfwHttpReq *)req)->flags);
+		}
+		break;
+	default:
+		return TFW_BLOCK;
+	}
+
+	return TFW_PASS;
+}
+
+static int
 frang_http_ct_check(const TfwHttpReq *req, FrangAcc *ra, FrangCtVals *ct_vals)
 {
 	TfwStr field, *s;
@@ -1188,6 +1231,13 @@ frang_http_req_process(FrangAcc *ra, TfwConn *conn, TfwFsmData *data,
 		*/
 		if (f_cfg->http_ct_required || f_cfg->http_ct_vals)
 			r = frang_http_ct_check(req, ra, f_cfg->http_ct_vals);
+
+		/* Do checks for websocket upgrade */
+		if (test_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET, req->flags)
+		    && (r = frang_http_upgrade_websocket(req, ra, f_cfg)))
+		{
+			T_FSM_EXIT();
+		}
 
 		__FRANG_FSM_MOVE(Frang_Req_Body_Start);
 	}
--- a/fw/http_limits.c
+++ b/fw/http_limits.c
@@ -4,7 +4,7 @@
  * Interface to classification modules.
  *
  * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
- * Copyright (C) 2015-2021 Tempesta Technologies, Inc.
+ * Copyright (C) 2015-2022 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
@@ -572,7 +572,7 @@ frang_http_upgrade_websocket(const TfwHttpReq *req, FrangAcc *ra,
 
 	switch (req->version) {
 	/*
-	 * TODO upgrade websocket checks for h2 as described in RFC8441
+	 * TODO #755: upgrade websocket checks for h2 as described in RFC8441
 	 */
 	case TFW_HTTP_VER_20:
 		break;
@@ -593,8 +593,7 @@ frang_http_upgrade_websocket(const TfwHttpReq *req, FrangAcc *ra,
 				  "websocket");
 			return TFW_BLOCK;
 		}
-		if (req->version == TFW_HTTP_VER_10
-		    || req->version == TFW_HTTP_VER_09)
+		if (req->version < TFW_HTTP_VER_11)
 		{
 			clear_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET,
 				  ((TfwHttpReq *)req)->flags);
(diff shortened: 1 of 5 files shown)


fw/http_parser.c
```diff
@@ -3320,6 +3332,167 @@ __parse_pragma(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+					 TFW_STR_NAME);
+		__FSM_I_JMP(I_UpgradeProtocolEnd);
+	}
+
+	__FSM_STATE(I_UpgradeProtocolEnd) {
+		if (likely(__fsm_sz)) {
+			/* Save protocol name */
+			__msg_hdr_chunk_fixup(p, __fsm_sz);
+			__FSM_I_chunk_flags(TFW_STR_NAME);
+		}
```
krizhanovsky, line 3382: Usually we achieve this with some MOVE fixup macros, which moves the FSM and fixes up current string if we reach end of data

fw/http_parser.c
```diff
@@ -3320,6 +3332,167 @@ __parse_pragma(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+		/*
+		 * Only ',' is available now: fixup ',' as independent chunk,
+		 * SP will be fixed up at next enter to the FSM.
+		 */
+		__FSM_I_MOVE_fixup(I_UpgradeSP, 1, 0);
+	}
+
+	__FSM_STATE(I_UpgradeSP) {
+		if (unlikely(*p != ' '))
+			return CSTR_NEQ;
```
krizhanovsky, line 3485: The RFC doesn't allow many SPs, right? Could you please leave comment if so

krizhanovsky (commented):
Just several notes how to make the parser move efficient

fw/http_parser.c
ttaym: In fact this part of patch inconsistent with the RFC. Lists have to be matched a little different. I'll fix that.

fw/http_parser.c
```diff
@@ -3320,6 +3332,167 @@ __parse_pragma(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+			__FSM_I_chunk_flags(TFW_STR_VALUE);
+		}
+
+		if (likely(IS_CRLF(*(p + __fsm_sz)))) {
+			__FSM_EXIT(__data_processed(p + __fsm_sz));
+		}
+
+		p += __fsm_sz;
+		if (likely(*p == ','))
+			__FSM_I_JMP(I_UpgradeComma);
```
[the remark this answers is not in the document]
ttaym: Whitespace and comma matching should be reworked here. See https://github.com/tempesta-tech/tempesta/pull/1571/commits/8d9320042c2c3d7a0218c9870f5777b234d1d64c#r819362089

fw/http_parser.c
```diff
@@ -3320,6 +3332,167 @@ __parse_pragma(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+	}
+
+	/*
+	 * At this state we know that we saw at least one character in
+	 * protocol name and now we can pass zero length token.
+	 */
+	__FSM_STATE(I_UpgradeProtocol) {
+		__FSM_I_MATCH_MOVE_fixup(token, I_UpgradeProtocol,
+					 TFW_STR_NAME);
+		__FSM_I_JMP(I_UpgradeProtocolEnd);
```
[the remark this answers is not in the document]
ttaym: This pattern (`I_.*`, `I_.*Start`, `I_.*End` with just jumps) used in not so few locations in parser. See for example `__h2_req_parse_authority()`. I designed this parser after it. And pay attention to the fact that `Version` states are in fact very rare and coded just to comply with the RFC.

fw/http_parser.c
ttaym: The `__req_parse_upgrade()` function is low level and use manual chunking facilities. See comment before the function:

>  Nested FSM with explicit fine-grained fixups, should employ
 __FSM_I_MOVE_fixup()/__FSM_I_MATCH_fixup()/TRY_STR_fixup() everywhere.

All functions marked with this comments use the pattern you commented on here.

fw/http_parser.c
```diff
@@ -3320,6 +3332,167 @@ __parse_pragma(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+	 */
+	__FSM_STATE(I_UpgradeProtocolStart) {
+		__FSM_I_MATCH_MOVE_fixup(token, I_UpgradeProtocol,
+					 TFW_STR_NAME);
+		/*
+		 * Protocol name should contain at least 1 character.
+		 */
+		if (likely(__fsm_sz))
+			__FSM_I_JMP(I_UpgradeProtocolEnd);
+
```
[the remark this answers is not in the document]
ttaym: In fact in `I_UpgradeProtocolEnd` state `__fsm_sz` bytes swallowed immediately. But we can come to the state from different places and if we in each of this places fixes up chunk than code locality in fact degraded. And jumpes to labels will be in any cases.

fw/http_parser.c
```diff
@@ -3320,6 +3332,167 @@ __parse_pragma(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+						   0);
+			return CSTR_NEQ;
+		}
+
+		/*
+		 * Only ',' is available now: fixup ',' as independent chunk,
+		 * SP will be fixed up at next enter to the FSM.
+		 */
+		__FSM_I_MOVE_fixup(I_UpgradeSP, 1, 0);
+	}
```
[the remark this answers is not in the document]
ttaym: Whitespace and comma matching should be reworked here. See https://github.com/tempesta-tech/tempesta/commit/8d9320042c2c3d7a0218c9870f5777b234d1d64c#r819362089

fw/http_limits.c
```diff
@@ -564,6 +564,49 @@ frang_http_methods_override(const TfwHttpReq *req, FrangAcc *ra,
 	return TFW_PASS;
```
krizhanovsky, line 564: Please update the copyright year for the file

fw/http_limits.c
```diff
@@ -564,6 +564,49 @@ frang_http_methods_override(const TfwHttpReq *req, FrangAcc *ra,
...
 
+static int
+frang_http_upgrade_websocket(const TfwHttpReq *req, FrangAcc *ra,
+			     FrangVhostCfg *f_cfg)
+{
+	BUG_ON(!req);
+
+	switch (req->version) {
+	/*
+	 * TODO upgrade websocket checks for h2 as described in RFC8441
```
krizhanovsky, line 575: Usually we add the issue number for TODO comments and when we start on an issue it's handy to grep the code for something like `TODO #775`

fw/http_limits.c
```diff
@@ -564,6 +564,49 @@ frang_http_methods_override(const TfwHttpReq *req, FrangAcc *ra,
...
+		if (test_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET, req->flags)
+		    && !test_bit(TFW_HTTP_B_CONN_UPGRADE, req->flags))
+		{
+			frang_msg("upgrade request without connection option",
+				  &FRANG_ACC2CLI(ra)->addr, ": protocol: %s\n",
+				  "websocket");
+			return TFW_BLOCK;
+		}
+		if (req->version == TFW_HTTP_VER_10
+		    || req->version == TFW_HTTP_VER_09)
```
krizhanovsky, line 597: This is equal to `req->version < TFW_HTTP_VER_11`

fw/http_parser.c
```diff
@@ -938,16 +940,26 @@ process_trailer_hdr(TfwHttpMsg *hm, TfwStr *hdr, unsigned int id)
...
 	}
 
+#define TRY_STR_LAMBDA_fixup(str, field, lambda, curr_st, next_st)	\
+	TRY_STR_LAMBDA_fixup_flag(str, field, lambda, curr_st,		\
+				  next_st, 0)
+
 #define TRY_STR_fixup(str, curr_st, next_st)				\
 	TRY_STR_LAMBDA_fixup(str, &parser->hdr, { }, curr_st, next_st)
 
+#define TRY_STR_fixup_flag(str, curr_st, next_st, flag)			\
```
krizhanovsky, line 959: `TRY_STR_fixup_flag` is never used

krizhanovsky (approved):
Good to merge after the couple of cleanups

PR #1590, opened 2022-03-15
Content length zero in h2 get request fix

ttaym:
Fixes discovered false-pass unit test.

```patch
	/* But with content-length will be block for http2 too */
	EXPECT_BLOCK_REQ_H2(":authority: debian\n"
			    ":method: GET\n"
-			    ":scheme: http\n"
+			    ":scheme: https\n"
			    ":path: /\n"
			    "content-length: 0");
```

Http2 only viable when https scheme, so was block, but not on `content-length` bot on `scheme`. PR fixes that and `content-legnth` h2 parsing code.

### DIFF
--- a/fw/http_parser.c
+++ b/fw/http_parser.c
@@ -6308,6 +6308,15 @@ __h2_req_parse_content_length(TfwHttpMsg *msg, unsigned char *data, size_t len,
 			      bool fin)
 {
 	int ret;
+	TfwHttpReq *req = (TfwHttpReq *)msg;
+
+	if (unlikely(req->method == TFW_HTTP_METH_GET
+		     || req->method == TFW_HTTP_METH_HEAD
+		     || req->method == TFW_HTTP_METH_DELETE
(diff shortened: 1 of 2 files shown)


fw/http_parser.c
```diff
@@ -6308,6 +6308,15 @@ __h2_req_parse_content_length(TfwHttpMsg *msg, unsigned char *data, size_t len,
 			      bool fin)
 {
 	int ret;
+	TfwHttpReq *req = (TfwHttpReq *)msg;
+
+	if (unlikely(req->method == TFW_HTTP_METH_GET
```
alegz-filomaphitskay, line 6313: Can't method override occur?

ttaym: That's possible. We should take it into account.

alegz-filomaphitskay (approved):

ttaym:
Merge now as is. Method overrides will be handled in separate PR.

Issue #1103, opened 2018-11-17
Optimizer for HTTP messages adjustment


krizhanovsky:
Tightly linked with #634 which is required for #515.

## Current logic

Currently we adjust HTTP messages like:
```C
tfw_http_adjust_req(TfwHttpReq *req)
{
        r = tfw_http_sess_req_process(req);
 
        r = tfw_http_add_x_forwarded_for(hm);
 
        r = tfw_http_add_hdr_via(hm);
 
        r = tfw_http_msg_del_hbh_hdrs(hm);
 
        r = tfw_http_set_loc_hdrs(hm, req);
 
        return tfw_http_set_hdr_connection(hm, TFW_HTTP_F_CONN_KA);
}
```

This is a bunch of adjustment calls and each of the call does (1) relatively expensive skb fragmentation and (2) generates independent set of new skb frags. So at the end of the call there are many skb fragments.

For example consider we have following HTTP request:
```
GET / HTTP/1.1\r\n
Host: example.com\r\n
X-Forwarded-For: 1.1.1.1\r\n
Connection: keep-alive\r\n
Accept: */*\r\n
Keep-Alive: timeout=600, max=65526\r\n
Cookie: foo\r\n
\r\n
```

And we need to:

1. add `2.2.2.2` to `X-Forwarded-For`
2. rewrite `Connection` to `close`
3. Add `Via: 1.1 foo.com (Tempesta FW)`
4. remove header `Keep-Alive`

Currently we do following fragmentation steps:

1. generate a new fragment for `2.2.2.2` and split the request to 2 frags - we have 3 frags now;
2. `keep-alive` is shorter than `close`, so the last current fragment is split into two around the gap - 4 frags now;
3. `Via` is inserted before the trailing CRLF - one more split and one more new frag - we come to 6 frags;
4. Removal of the `Keep-Alive` header produces a gap, so we finish with 7 frags for very small HTTP request.

We do not move trailing CRLF and do the fragmentation stuff to save data pointers in HTTP message representation, `TfwHttpMsg`.

### Preliminary logic

Actually we have some rudiments for the logic, in particular we treat all user specified headers modifications in `TfwHdrMods` and process them at once in functions like `tfw_h2_resp_adjust_fwd()` and `tfw_h2_resp_next_hdr()`.

## Proposed design

The target of the function is to make only 1 memory (skb frament) allocation in worst case and use current skb overhead in most cases (see Testing section).

The adjustment logic must be split into 3 phases:

[...]

PR #1592, opened 2022-03-16
Recreate upgrade headers for websocket request

ttaym:
Contributes to https://github.com/tempesta-tech/tempesta/issues/755

Signed-off-by: Aleksey Mikhaylov [aym@tempesta-tech.com](mailto:aym@tempesta-tech.com)

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -2581,10 +2581,21 @@ tfw_http_set_hdr_upgrade(TfwHttpMsg *hm, bool is_resp)
 	int r = 0;
 
 	if (test_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET, hm->flags)) {
+		/*
+		 * RFC7230#section-6.7:
+		 * A server that sends a 101 (Switching Protocols) response
+		 * MUST send an Upgrade header field to indicate the new
+		 * protocol(s) to * which the connection is being switched; if
+		 * multiple protocol layers are being switched, the sender MUST
+		 * list the protocols in layer-ascending order.
+		 *
+		 * We do expect neither upgrades besides 'websocket' nor
+		 * multilayer upgrades. So we consider extra options as error.
+		 */
 		if (is_resp && ((TfwHttpResp *)hm)->status == 101
 		    && test_bit(TFW_HTTP_B_UPGRADE_EXTRA, hm->flags))
 		{
-			T_ERR("Unable to add uncompliant Upgrade: header "
+			T_ERR("Unable to add uncompliant 'Upgrade:' header "
 			      "to msg [%p]\n", hm);
 			return -EINVAL;
 		}
@@ -2714,6 +2725,12 @@ tfw_http_set_hdr_connection(TfwHttpMsg *hm, unsigned long conn_flg)
 		return 0;
 	}
 
+	/*
+	 * We can see `TFW_HTTP_B_CONN_CLOSE` here only in case of 4XX
+	 * response with 'Connection: close' option.
+	 *
+	 * For requests conn_flg by default is TFW_HTTP_B_CONN_KA.
+	 */
 	if (unlikely(conn_flg == BIT(TFW_HTTP_B_CONN_CLOSE)))
 		return TFW_HTTP_MSG_HDR_XFRM(hm, "Connection", "close",
 					     TFW_HTTP_HDR_CONNECTION, 0);
--- a/fw/http.c
+++ b/fw/http.c
@@ -2573,6 +2573,33 @@ tfw_http_set_hdr_date(TfwHttpMsg *hm)
 }
 
 /*
+ * Add 'Upgrade:' header for websocket upgrade messages
+ */
+static int
+tfw_http_set_hdr_upgrade(TfwHttpMsg *hm, bool is_resp)
+{
+	int r = 0;
+
+	if (test_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET, hm->flags)) {
+		if (is_resp && ((TfwHttpResp *)hm)->status == 101
+		    && test_bit(TFW_HTTP_B_UPGRADE_EXTRA, hm->flags))
+		{
+			T_ERR("Unable to add uncompliant Upgrade: header "
+			      "to msg [%p]\n", hm);
+			return -EINVAL;
+		}
+		r = tfw_http_msg_hdr_xfrm(hm, "upgrade", SLEN("upgrade"),
+				  "websocket", SLEN("websocket"),
+				  TFW_HTTP_HDR_UPGRADE, 0);
+		if (r)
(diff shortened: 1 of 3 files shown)


fw/http.c
```diff
@@ -2572,6 +2572,33 @@ tfw_http_set_hdr_date(TfwHttpMsg *hm)
...
+ * Add 'Upgrade:' header for websocket upgrade messages
+ */
+static int
+tfw_http_set_hdr_upgrade(TfwHttpMsg *hm, bool is_resp)
+{
+	int r = 0;
+
+	if (test_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET, hm->flags)) {
+		if (is_resp && ((TfwHttpResp *)hm)->status == 101
+		    && test_bit(TFW_HTTP_B_UPGRADE_EXTRA, hm->flags))
```
krizhanovsky, line 2585: I'm not sure if I understand this. It seems this is about https://datatracker.ietf.org/doc/html/rfc7230#section-6.7

> A server that sends a 101 (Switching Protocols) response MUST send an
   Upgrade header field to indicate the new protocol(s) to which the
   connection is being switched; if multiple protocol layers are being
   switched, the sender MUST list the protocols in layer-ascending
   order.  A server MUST NOT switch to a protocol that was not indicated
   by the client in the corresponding request's Upgrade header field.
 
This piece of code is for response with something extra plus to `websocket` in `Upgrader` header. Since we can only generate `Upgrade: websocket` in adjusted client request and server MUST NOT add anything non-requested, then what could be in the response for the extra? Can't we just always send `Upgrade: websocket` from the server to the client?

If I'm wrong, then a good comment with the RFC cite is required in the place.

fw/http.c
```diff
@@ -2677,23 +2704,38 @@ tfw_http_expand_hbh(TfwHttpResp *resp, unsigned short status)
...
 					    TFW_HTTP_HDR_CONNECTION);
 	}
+
+	if (r < 0)
+		return r;
+
+	if (test_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET, hm->flags)
+	    && test_bit(TFW_HTTP_B_CONN_UPGRADE, hm->flags))
+		r = TFW_HTTP_MSG_HDR_XFRM(hm, "Connection", "upgrade",
+				      TFW_HTTP_HDR_CONNECTION, 1);
```
krizhanovsky, line 2736: Headers modification is a heavy operation, probably involving memory allocations and skb frarmentation (see https://github.com/tempesta-tech/tempesta/issues/1103), so it'd be more beneficial to do bit more complex condition and call do only one transformation call.

krizhanovsky (changes requested):
It seems there are several places for improvement

fw/http.c
```diff
@@ -2572,6 +2572,33 @@ tfw_http_set_hdr_date(TfwHttpMsg *hm)
...
+			      "to msg [%p]\n", hm);
+			return -EINVAL;
+		}
+		r = tfw_http_msg_hdr_xfrm(hm, "upgrade", SLEN("upgrade"),
+				  "websocket", SLEN("websocket"),
+				  TFW_HTTP_HDR_UPGRADE, 0);
+		if (r)
+			T_ERR("Unable to add Upgrade: header to msg [%p]\n", hm);
+		else
+			T_DBG2("Added Upgrade: header to msg [%p]\n", hm);
```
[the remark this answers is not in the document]
ttaym: As i see `Documentation/core-api/printk-formats.rst` say the oppposite. Format `%p` is hashed and hidden, but `%pK` is unhidden plain pointers.

> This modifier [%pK] is only intended when producing content of a file read by userspace from e.g. procfs or sysfs, not for dmesg. Please refer to the section about %p above for discussion about how to manage hashing pointers in printk().

fw/http.c
ttaym: My intention was as follows:

We always strip upgrade header from response and re-add it when needed. So we do not just pass it through tempesta as is. But when we recreate headers would be beneficially to do simple sane check because we essentially hide backend and act on behalf on it.

Imaging erroneous backend that we on our discretion silently transform into compliant backend. That can would be destructive if you ask me.

But this intention is, i understand, subtle and subjective. It is up to you if we need just alway send `Upgrade: websocket`.

fw/http.c
ttaym: That make sense. I'll fix it.

fw/http.c
```diff
@@ -2677,23 +2704,48 @@ tfw_http_expand_hbh(TfwHttpResp *resp, unsigned short status)
...
+						  "keep-alive",
+						  TFW_HTTP_HDR_CONNECTION, 0);
+		}
+	} else {
+		if (test_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET, hm->flags)
+		    && test_bit(TFW_HTTP_B_CONN_UPGRADE, hm->flags))
+		{
+			r = TFW_HTTP_MSG_HDR_XFRM(hm, "Connection",
+						  "upgrade",
+						  TFW_HTTP_HDR_CONNECTION, 0);
```
krizhanovsky, line 2740: If we have `Connection: close, upgrade`, then it seems we should send just `Connection: close`. Also could you please check whether a connection is considered keep-alive by default (there is neither `close` nor `keep-alive`)?

ttaym: By default passed 0 in the function for responses, `BIT(TFW_HTTP_B_CONN_KA)` passed to function for responses only if paired request has `TFW_HTTP_B_CONN_KA` bit.

But for requests by default passed `BIT(TFW_HTTP_B_CONN_KA)`.

And semantically connection is keep-alive by default for `HTTP/1.1` and later.

ttaym: If we have `Connection: close, upgrade` in response than now result depends on status. If status is `4XX` than connection is closed without upgrade. If `101` than upgrade is finished and connection state than depends on actual closing on no-closing from backend.

May be that is right enough.

fw/http.c
```diff
@@ -2572,6 +2572,44 @@ tfw_http_set_hdr_date(TfwHttpMsg *hm)
...
+tfw_http_set_hdr_upgrade(TfwHttpMsg *hm, bool is_resp)
+{
+	int r = 0;
+
+	if (test_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET, hm->flags)) {
+		/*
+		 * RFC7230#section-6.7:
+		 * A server that sends a 101 (Switching Protocols) response
+		 * MUST send an Upgrade header field to indicate the new
+		 * protocol(s) to * which the connection is being switched; if
```
krizhanovsky, line 2588:
```suggestion
		 * protocol(s) to which the connection is being switched; if
```

krizhanovsky (approved):
LGTM

PR #1593, opened 2022-03-16
1589 include directive in the configuration file

pale-emperor:
From now - config generated from template


scripts/tempesta.sh
```diff
@@ -39,6 +39,7 @@ tfw_path=${TFW_PATH:="$TFW_ROOT/fw"}
 tls_path=${TLS_PATH:="$TFW_ROOT/tls"}
 lib_path=${LIB_PATH:="$TFW_ROOT/lib"}
 tfw_cfg_path=${TFW_CFG_PATH:="$TFW_ROOT/etc/tempesta_fw.conf"}
+tfw_cfg_template=${TFW_CFG_TMPL:="$TFW_ROOT/etc/tempesta_tmpl.conf"}
```
krizhanovsky, line 42: If after the first run a user changes tempesta_fw.conf, then on the restart the changes will be rewritten by the old template. I think it'd be more robust to hide the template games:
1. just keep one `tempesta_fw.conf`
2. `templater()` creates a temporary file and assembles the final config there
3. start/restart/reload Tempesta FW
4. when Tempesta FW restarted, remove the temporary file

Probably we don't need to care about [filesystem race conditions](https://tldp.org/HOWTO/Secure-Programs-HOWTO/avoid-race.html) in the single-user edge server.

krizhanovsky (changes requested):
Beautiful and short bash scripting! But there might be issues for users with rewritten config file

Issue #881, opened 2018-01-08
WebSocket test


krizhanovsky:
Need to develop a functional test for Tempesta FW websockets (#755). Python [websockets](https://websockets.readthedocs.io/en/stable/index.html) can be used for both the client and server sides. However, it seems all Python websockets libraries (also [websocket-client](https://pypi.org/project/websocket-client/)) depend on Python 3, so the issue is blocked by [migration of our testing framework to Python3](https://github.com/tempesta-tech/tempesta-test/issues/56).

Following tests for HTTP/1 ~and HTTP/2~ client must be done (**UPD** it seems the modern HTTP/2 applications [prefer SSE over WebSockets](https://www.mnot.net/blog/2022/02/20/websockets), so it's hard to find an HTTP/2 WebSocket client):
- [x] ping pong over ws:// (for HTTP/1 only)
- [x] ping pong over wss:// (client - tfw - backend)
- [x] ping pong over wss:// (client - tfw - proxy - backend, see at the below)
- [x] ~HTTP tables rules to [load balance websockets](https://www.haproxy.com/blog/websockets-load-balancing-with-haproxy/) traffic by `Host` or `Connection: upgrade` headers~ We have other HTTPtable tests and it seems there is nothing special about websockets
- [x] cache: check that we never cache 101 responses with `Upgrade` header. Error responses for invalid client upgrade requests still can be cached
- [x] (stress test) establish many wss connections against several server groups using all current connections schedulers, cause severe load and do several Tempesta restarts during the workload
- [x] ~(scheduler stress test) we have stress tests for the schedulers, at least https://github.com/tempesta-tech/tempesta-test/blob/master/sched/test_hash_stress.py , so we need a similar tests for this and other schedulers to make sure that we can normally create a websocket under high scheduler load~ The test case seems covered by the previous test.

### wss://

Tempesta FW doesn't have client side TLS (#769 ) yet. However, websockets have nothing about TLS and just use the same upgrade scheme for both ws:// and wss:// . In the default scheme a client connects Tempesta on HTTPS and Tempesta has plain HTTP connection to the backend, which leads to wss:// between the client and Tempesta and ws:// between Tempesta and the backend server.

One wishing to have wss:// to the backend should use an additional HTTPS server, e.g. Nginx or HAproxy, supporting client side TLS. Tempesta FW config must specify the server (typically running on the same host with Tempesta FW) as the backend server.
```
                       +---------------------------+
[...]

Issue #710, opened 2017-03-18
Dynamically allocate and shrink server connections


krizhanovsky:
The issue was originally described in https://github.com/tempesta-tech/tempesta/issues/76#issuecomment-185295563 . The issue is crucial due to recent https://github.com/tempesta-tech/tempesta/pull/660/ : if there are a lot of non-idempotent requests, then we have limited throughput to upstream servers. Thus static number of connections can be wrong and it'd be good to establish a new server connection if current connections are blocked by non-idempotent requests. Meantime, if APM shows that the server is overloaded, then it could has sense to reduce number of server connections. Probably unused connections also can be shrunk - it could improve performance of schedulers. However, too frequent changing of connections number is also not wished due to relatively expensive schedulers updates.

krizhanovsky:
Crucial note from #498:

> Long polling is another question that bothers me. Imagine we have 32k connections to backend servers, and have 32k clients that use long polling In this case server connections are depleted and we can't serve new client connections.

PR #1594, opened 2022-03-17
Upgrade connection to websocket and provision new connection

ttaym:
Contributes to https://github.com/tempesta-tech/tempesta/issues/755

Signed-off-by: Aleksey Mikhaylov [aym@tempesta-tech.com](mailto:aym@tempesta-tech.com)

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -2339,7 +2339,6 @@ static int
 tfw_http_conn_init(TfwConn *conn)
 {
 	T_DBG2("%s: conn=[%p]\n", __func__, conn);
-
 	if (TFW_CONN_TYPE(conn) & Conn_Srv) {
 		TfwSrvConn *srv_conn = (TfwSrvConn *)conn;
 		if (!list_empty(&srv_conn->fwd_queue)) {
@@ -5931,6 +5930,7 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 	unsigned int chunks_unused, parsed;
 	TfwHttpReq *bad_req;
 	TfwHttpMsg *hmresp, *hmsib;
+	TfwHttpResp *resp;
 	TfwFsmData data_up;
 	bool conn_stop, filtout = false;
 
@@ -5953,6 +5953,7 @@ next_msg:
 	parsed = 0;
 	hmsib = NULL;
 	hmresp = (TfwHttpMsg *)stream->msg;
+	resp = (TfwHttpResp *)hmresp;
 
 	r = ss_skb_process(skb, tfw_http_parse_resp, hmresp, &chunks_unused,
 			   &parsed);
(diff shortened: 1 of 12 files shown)


fw/http_sched_hash.c
```diff
@@ -432,13 +435,38 @@ tfw_sched_hash_del_srv(TfwServer *srv)
...
+
+	tfw_sched_hash_add_conns(srv, cl_copy, &seed, seed_inc);
+
+	rcu_assign_pointer(srv->sched_data, cl_copy);
+
+	if (cl)
+		call_rcu(&cl->rcu, tfw_sched_hash_put_srv_data);
+
+	return 0;
+}
```
krizhanovsky, line 460: This is more or less copy & paste from `tfw_sched_hash_add_srv()` except allocation flags and freeing temoprorary data at the end, so it makes sense to rework both the functions to reuse the same code.

fw/http.c
```diff
@@ -6098,6 +6099,44 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
...
+			return TFW_BLOCK;
+		}
+
+		set_bit(TFW_CONN_B_UNSCHED,
+			&((TfwSrvConn *)hmresp->conn)->flags);
+
+		tfw_sock_srv_conn_activate(srv, srv_conn);
+		tfw_sock_srv_connect_try(srv_conn);
+
+		srv->sg->sched->upd_srv(srv);
```
krizhanovsky, line 6137: The implementation looks robust, but previously we had fixed a lot of bugs in the lock-free schedulers and server connections lifetime, so I added a new test case to https://github.com/tempesta-tech/tempesta/issues/881 about scheduler stress test. @pale-emperor FYI

fw/http_sched_ratio.c
```diff
@@ -1291,13 +1312,48 @@ tfw_sched_ratio_del_srv(TfwServer *srv)
...
+		if (tfw_srv_conn_unscheduled(srv_conn))
+			continue;
+		if (unlikely(ci++ == srv->conn_n))
+			goto err;
+		cl_copy->conns[ci-1] = srv_conn;
+	}
+	if (unlikely(ci != srv->conn_n))
+		goto err;
+	cl->conn_n = ci;
+	rcu_assign_pointer(srvdesc->cl, cl_copy);
```
krizhanovsky, line 1338: This is almost a copy & paste from `tfw_sched_ratio_srvdesc_setup_srv()`, so let's unify the functions.

krizhanovsky (changes requested):
Quite a good and accurate work, especially in the subtle server connection schedulers code! Besides the minor cleanups, there is only one place with `tfw_sock_srv_connect_try()` instead of `tfw_sock_srv_connect_try_later()`, which confuses me.

fw/http.c
```diff
@@ -6098,6 +6099,44 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
...
+						      "connection for websocket"
+						      " upgrade response");
+			return TFW_BLOCK;
+		}
+
+		set_bit(TFW_CONN_B_UNSCHED,
+			&((TfwSrvConn *)hmresp->conn)->flags);
+
+		tfw_sock_srv_conn_activate(srv, srv_conn);
+		tfw_sock_srv_connect_try(srv_conn);
```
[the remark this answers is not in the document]
ttaym: I see.

`tfw_sock_srv_connect_try_later` really should be used here.

fw/http_sched_ratio.c
```diff
@@ -1035,32 +1045,52 @@ tfw_sched_ratio_del_grp(TfwSrvGroup *sg)
...
+		if (tfw_srv_conn_unscheduled(srv_conn))
+			continue;
 		if (unlikely(ci++ == srv->conn_n))
 			goto err;
-		*conn++ = srv_conn;
+
+		cl_copy->conns[ci-1] = srv_conn;
 	}
 	if (unlikely(ci != srv->conn_n))
 		goto err;
```
krizhanovsky, line 1082: It seems I was wrong with the proposal of the functions unification in both the schedulers.

Previously we called `tfw_sock_srv_new_conn()`, which adds a new connection to the connection list, only from process context (since we have constant number of server connections and just failover TCP connections inside them). Moreover,  `tfw_peer_add_conn()` uses a spin lock to add a connection. I believe we have a mutex on concurrent configuraion reloads, so we rely on the fact that `tfw_sched_ratio_add_srv()` always works in single context.

Now we can call this function concurrently with other websocket creation and with the configuration reload, so we definitely have races in all the places in the function. We can back to the separate functions and use the spin lock in the new function, but we'll also need to use the spin lock in all other functions iterating the connections list, which introduces lock contention.

What you're doing in the patch is almost solving https://github.com/tempesta-tech/tempesta/issues/710 , which is a hard task on it's own to upgrade the schedulers data structures without locks.

Initially with #755 I assumed that you just "steal" `sk` from `TfwSrvConn` and initiate the existing connection failovering mechanism.

krizhanovsky (changes requested):
It seems we have severe race, which may lead to crashes on multi-core workload with concurrent websocket creations, and need to find a simpler solution leaving #710 for later phases.

krizhanovsky:
The code and the only one not-addressed comment were moved to #1595

PR #1595, opened 2022-03-18
Websocket simple proxy protocol implementation

ttaym:
Contributes to #755 


fw/http.c
```diff
@@ -6236,25 +6249,29 @@ tfw_http_msg_process_generic(TfwConn *conn, TfwStream *stream,
...
 	{
 		if (likely(r == T_OK || r == T_POSTPONE)) {
-			data->skb->next = data->skb->prev = NULL;
-			r = TFW_CONN_H2(conn)
-				? tfw_h2_frame_process(conn, data->skb)
-				: tfw_http_msg_process_generic(conn, stream,
-							       data);
+			skb->next = skb->prev = NULL;
+			if (unlikely(TFW_CONN_PROTO(conn) & TFW_FSM_WS))
+				r = tfw_ws_msg_process(conn, skb);
```
krizhanovsky, line 6266: This looks hacky and there is no sense to call WS through HTTP FSM handler. We should just call `tfw_gfsm_register_fsm()` for `tfw_ws_msg_process()` just as we do this for this handler for the HTTP FSM.

fw/websocket.c
```diff
@@ -0,0 +1,263 @@
...
+		if ((r = tfw_connection_send(conn->pair, (TfwMsg *)hm)))
+			T_DBG("%s: cannot send data to backend via websocket\n",
+			      __func__);
+	}
+
+	if (r) {
+		tfw_connection_close(conn, true);
+	}
+
+	tfw_http_msg_free(hm);
```
krizhanovsky, line 84: I think we don't need the dummy `hm` and call `tfw_connection_send()`, which just wraps our `tfw_ws_conn_send()`, which is in turn wraps `ss_send()`, and just can proxy the `skb` with `ss_send()` on the paired connection.

krizhanovsky (changes requested):
Looks like quite a few places should be reworked

fw/websocket.c
```diff
@@ -0,0 +1,263 @@
...
+	if (!pair)
+		return;
+
+	tfw_connection_close(pair, true);
+}
+
+static void
+tfw_ws_conn_drop(TfwConn *conn)
+{
+	tfw_conn_hook_call(TFW_CONN_HTTP_TYPE(conn), conn, conn_drop);
```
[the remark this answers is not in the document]
ttaym: No recursion in fact. See https://github.com/tempesta-tech/tempesta/pull/1595#discussion_r836592418

fw/http.c
ttaym: For now we call `tfw_gfsm_register_fsm()` for HTTP FSM only to call frang hooks. But `tfw_http_msg_process()` called directly from ss_hooks.

So if you prefer full separation of websocket protocol from `tfw_http_msg_process()` than we need separate websocket ss_hooks.

fw/websocket.c
ttaym: We need dummy http message at least for secure websockets case to properly call `tfw_tls_conn_send()`.

All websocket hooks downcall to proper http or https hooks.

fw/websocket.c
```diff
@@ -0,0 +1,263 @@
...
+
+/**
+ * Send the @msg skbs as is.
+ */
+static int
+tfw_ws_conn_send(TfwConn *conn, TfwMsg *msg)
+{
+	int r;
+
+	r = tfw_conn_hook_call(TFW_CONN_HTTP_TYPE(conn), conn, conn_send, msg);
```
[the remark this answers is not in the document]
ttaym: The same here, see https://github.com/tempesta-tech/tempesta/pull/1595#discussion_r836592418

fw/http.c
ttaym: I have factored out common code for recv in `tfw_connection_recv()` in `connection.c`. These (websocket and http processing callbacks) are now separate.

fw/sock_srv.c
```diff
@@ -708,6 +713,30 @@ tfw_sock_srv_start_srv(TfwSrvGroup *sg, TfwServer *srv, void *hm)
 	return 0;
 }
 
+TfwSrvConn *
+tfw_sock_srv_new_steal_sk(TfwSrvConn *srv_conn)
+{
+	TfwSrvConn *conn;
+	TfwServer *srv = (TfwServer *)srv_conn->peer;
+
+	if (!(conn = tfw_sock_srv_new_conn(srv))) {
```
krizhanovsky, line 722: `tfw_sock_srv_new_conn()` -> `tfw_srv_conn_alloc()` -> `might_sleep()` and `kmalloc(GFP_KERNEL)`, so there is a bug. I don't think we need to allocate a new server connection since websockets don't have any server logic, which we have for HTTP connections. What we need to do here is to allocate a new `TfwConn` and probably just need a new memory cache for it.

fw/sock_srv.c
```diff
@@ -708,6 +713,30 @@ tfw_sock_srv_start_srv(TfwSrvGroup *sg, TfwServer *srv, void *hm)
...
+		T_WARN_ADDR("Can't create new connection for socket stealing",
+			    &srv->addr, TFW_NO_PORT);
+		return NULL;
+	}
+
+	tfw_connection_revive((TfwConn *)conn);
+	conn->destructor = (void *)tfw_srv_conn_release;
```
krizhanovsky, line 729: With stealing I meant exactly the call of `tfw_srv_conn_release()` for the current server connection, but we don't need to set the callback for the websocket connection `conn`.

See the `ss_linkerror()` calls in sock.c - it handles socket errors and initiates server reconnection via `tfw_srv_conn_release()`. Note that the function handles `srv_conn->sk == NULL`, so probably we can just

```C
tfw_connection_unlink_from_sk(srv_conn->sk);
conn->sk = srv_conn->sk;
srv_conn->sk = NULL;
tfw_srv_conn_release(srv_conn);
```
But be careful with other server connection users and check that we're under the socket lock just like `ss_linkerror()`.

fw/http_sched_ratio.c
```diff
@@ -1035,32 +1045,52 @@ tfw_sched_ratio_del_grp(TfwSrvGroup *sg)
...
+		if (tfw_srv_conn_unscheduled(srv_conn))
+			continue;
 		if (unlikely(ci++ == srv->conn_n))
 			goto err;
-		*conn++ = srv_conn;
+
+		cl_copy->conns[ci-1] = srv_conn;
 	}
 	if (unlikely(ci != srv->conn_n))
 		goto err;
```
krizhanovsky, line 1082: Just copy and paste comment https://github.com/tempesta-tech/tempesta/pull/1594/files#r833823157 :

It seems I was wrong with the proposal of the functions unification in both the schedulers.

Previously we called `tfw_sock_srv_new_conn()`, which adds a new connection to the connection list, only from process context (since we have constant number of server connections and just failover TCP connections inside them). Moreover,  `tfw_peer_add_conn()` uses a spin lock to add a connection. I believe we have a mutex on concurrent configuraion reloads, so we rely on the fact that `tfw_sched_ratio_add_srv()` always works in single context.

Now we can call this function concurrently with other websocket creation and with the configuration reload, so we definitely have races in all the places in the function. We can back to the separate functions and use the spin lock in the new function, but we'll also need to use the spin lock in all other functions iterating the connections list, which introduces lock contention.

What you're doing in the patch is almost solving https://github.com/tempesta-tech/tempesta/issues/710 , which is a hard task on it's own to upgrade the schedulers data structures without locks.

Initially with #755 I assumed that you just "steal" `sk` from `TfwSrvConn` and initiate the existing connection failovering mechanism.

krizhanovsky (commented):
Just reviewed #1594 comments in application to this PR.

fw/websocket.c
```diff
@@ -0,0 +1,292 @@
...
+ *
+ * So this function must be called manually on bootstrap finished.
+ */
+static int
+tfw_ws_conn_init(TfwConn *conn)
+{
+	T_DBG2("%s: conn=[%p]\n", __func__, conn);
+
+	return 0;
+}
```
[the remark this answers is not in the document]
ttaym: Yes, it seems that we can get rid of it.

fw/http.c
```diff
@@ -6067,12 +6101,30 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
...
+		websocket = true;
+	}
+
 	/*
 	 * If @skb's data has not been processed in full, then
 	 * we have pipelined responses. Create a sibling message.
 	 * @skb is replaced with a pointer to a new SKB.
 	 */
-	if (skb) {
+	if (skb && !websocket) {
```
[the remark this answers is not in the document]
ttaym: I'll leave comment here.

fw/http.c
```diff
@@ -6067,12 +6101,30 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
...
+	if (unlikely(test_bit(TFW_HTTP_B_CONN_UPGRADE, hmresp->flags)
+		     && test_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET, hmresp->flags)
+		     && ((TfwHttpResp *)hmresp)->status == 101))
+	{
+		r = tfw_http_websocket_upgrade(hmresp);
+		if (unlikely(r < TFW_PASS))
+			return TFW_BLOCK;
+
+		websocket = true;
+	}
```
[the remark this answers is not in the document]
ttaym: There is no specific reason for do upgrade before `tfw_http_resp_cache()`.

ttaym:
This answers https://github.com/tempesta-tech/tempesta/pull/1595#discussion_r841253216

While for backend connection `state` field not needed, but we use client connection as is with adapted websocket type. And we call for it for example `tfw_http_conn_drop()` -> `tfw_http_resp_terminate()` -> `tfw_http_resp_cache()` -> `tfw_gfsm_move()` that use `state`. I think here not safe to union with it.

fw/sock_srv.c
ttaym: I addressed this by moving connection allocation and stealing into websocket module. And now backend connection is simple connection, but client connection leaved `TfwCliConn`.

fw/sock_srv.c
ttaym: Ok. I use now connection destructor for reestablishing that is in essence the same as you suggest here. But i cannot see why we should unlink from socket, it would nullify all sync socket callbacks from stealed socket and prevent further socket operations.

Please re-review recent changes.

fw/http_sched_ratio.c
ttaym: Ok. Now socket stealing logic implemented instead of meddling with lock-free schedulers algorithms.

fw/connection.h
```diff
@@ -105,6 +107,7 @@ enum {
 	struct timer_list	timer;			\
 	TfwStream		stream;			\
 	TfwPeer 		*peer;			\
+	TfwConn			*pair;			\
```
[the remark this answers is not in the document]
ttaym: See https://github.com/tempesta-tech/tempesta/pull/1595#issuecomment-1091437259

fw/http_sched_hash.c
```diff
@@ -392,34 +395,46 @@ tfw_sched_hash_add_grp(TfwSrvGroup *sg, void *data)
...
+tfw_sched_hash_add_srv(TfwServer *srv)
 {
-	TfwHashConnList *cl = container_of(rcu, TfwHashConnList, rcu);
-	kfree(cl);
+	TfwHashConnList *cl = rcu_dereference_check(srv->sched_data, 1);
+
+	if (unlikely(cl))
+		return -EEXIST;
+
+	return tfw_sched_hash_srv_setup(srv);
```
krizhanovsky, line 437: All the changes in the scheduler besides the `tfw_srv_conn_unscheduled()` calls are rudiments from the PR prior the current socket stealing logic. Now we don't call the scheduler from the softirq context, so the changes should be reverted.

fw/server.h
```diff
@@ -163,6 +163,7 @@ typedef struct {
  *		  Called in process context at re-configuration time.
  * @del_srv	- delete single server added via add_srv.
  *		  Called in SoftIRQ context.
+ *
  * @sched_sg_conn	- virtual method. Schedule a request to a server from
  *			  given server group. Returns a server connection;
  * @sched_srv_conn	- schedule a request to the given server.
```
krizhanovsky, line 169: Probably the changes don't have much sense

fw/websocket.c
```diff
@@ -0,0 +1,356 @@
...
+	}
+	conn->peer = (TfwPeer *)srv;
+	conn->sk = srv_conn->sk;
+	conn->sk->sk_user_data = conn;
+
+	conn->destructor = tfw_ws_conn_release;
+	tfw_connection_revive(conn);
+
+	/* Connection destructor does failover for server connections */
+	srv_conn->sk = NULL;
```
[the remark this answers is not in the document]
krizhanovsky: Regarding https://github.com/tempesta-tech/tempesta/pull/1595#discussion_r841268795 the new logic looks robust and you're right, we don't need to call `tfw_connection_unlink_from_sk()`.

krizhanovsky (changes requested):
There are still several issues and several cleanups are also desired.

fw/cfg.c
```diff
@@ -125,7 +125,7 @@ __alloc_and_copy_literal(const char *src, size_t len, bool keep_bs)
 	}
 
 	/* Copy the string. Eat escaping backslashes if @keep_bs is not set. */
-	/* FIXME: the logic looks like a tiny FSM,
+	/* TODO: the logic looks like a tiny FSM,
```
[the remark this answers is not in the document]
ttaym: Through code i saw FIXMEs and TODOs. To be consistent it make sense to use only one of it, and another leave to private use. Through code TODOs are used much more frequently, so i rewrite all FIXME with TODO.

If you accept this change, then i'll amend https://github.com/tempesta-tech/tempesta/wiki/Development-guidelines in accordance with it.

fw/websocket.c
```diff
@@ -0,0 +1,356 @@
...
+	tfw_http_msg_free(hm);
+
+	return r;
+}
+
+/**
+ * Websocket connection hooks.
+ *
+ * These hooks unified between wss and ws. For wss called tls and for
+ * plain ws called http conn_hook. No self recursion here just downcall.
```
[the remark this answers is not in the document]
ttaym: If we call even through hook machinery lower layer, as here, i'd say that we do downcall, not upcall.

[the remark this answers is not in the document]
ttaym: Do a bit more extensive comment here.

fw/websocket.c
```diff
@@ -0,0 +1,356 @@
...
+#undef DEBUG
+#if DBG_WS > 0
+#define DEBUG DBG_WS
+#endif
+
+#include "cfg.h"
+#include "connection.h"
+#include "websocket.h"
+#include "http.h"
+#include "http_msg.h"
```
krizhanovsky, line 31: Now we don't need the HTTP headers:

```suggestion
#include "server.h"
```

fw/cfg.c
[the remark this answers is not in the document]
krizhanovsky: Frankly, I don't see any problem with using FIXME and TODO in the comments and we can accept or not the change - it's up to you: if you like TODO more, then I'm fine with it :) At least, I don't think it makes sense to make a statement about this in the guideline.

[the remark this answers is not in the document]
keshonok: I would say that TODO is more for something that it NOT implemented yet, whereas FIXME is more for something that HAS been implemented but not in the most desirable form. Also, I think I may have used FIXME for something that is mostly implemented but with small caveats, like cutting a few corners.

fw/http.c
```diff
@@ -5039,6 +5042,12 @@ tfw_http_req_cache_cb(TfwHttpMsg *msg)
 	/* Account current request in APM health monitoring statistics */
 	tfw_http_hm_srv_update((TfwServer *)srv_conn->peer, req);
```
krizhanovsky, line 5044: At this place, between getting the `srv_conn` and setting `TFW_CONN_B_UNSCHED`, we have a race: another CPU can get the same server connection and send an HTTP request to the same server socket right after our Upgrade request. Moreover, `tfw_http_popreq()` forwards pipelined requests, so the race is very probable. I spent quite a time to think how to fix the problem properly and didn't find an elegant solution.

It seem I also was wrong with my statement in the last review:
> Connection upgrade requests are non-idempotent by their nature. We can schedule (but not send!) HTTP requests to connections processing non-idempotent requests, but we should not schedule new requests for a connection processing upgrade message. This means that we can no just use TFW_CONN_B_HASNIP for upgrade requests and do need TFW_CONN_B_UNSCHED.

We schedule requests to `TfwSrvConn`, which remains intact during connection upgrade and only underlying socket is stolen. After revision it seems that we actually can treat connection upgrade requests as NIP and `tfw_http_conn_on_hold()` will do the work for us.

`tfw_http_conn_on_hold()` "unlocks" a connection, when a nip request is removed, i.e. by `tfw_http_propreq()` called from `tfw_http_resp_cache()`. `tfw_http_websocket_upgrade()` is called right after `tfw_http_resp_cache()`.

`tfw_http_resp_cache()` may send an error response to the client (by HTTP limits decision in the GFSM call), so the function must return error code an we should not upgrade to websockets on error.

It looks like the best way to fix the issue is to move `tfw_http_popreq()` our from `tfw_http_resp_cache()` and call it after `tfw_http_websocket_upgrade()`. Don't forget to call it also after `tfw_http_resp_cache()`  in `tfw_http_resp_terminate()`. We don't need TFW_CONN_B_UNSCHED in this case

fw/http_sched_ratio.c
```diff
@@ -1047,6 +1048,8 @@ tfw_sched_ratio_srvdesc_setup_srv(TfwServer *srv, TfwRatioSrvDesc *srvdesc)
 
 	conn = srvdesc->conn;
 	list_for_each_entry(srv_conn, &srv->conn_list, list) {
+		if (tfw_srv_conn_unscheduled(srv_conn))
+			continue;
```
krizhanovsky, line 1052: It seems we don't need `tfw_srv_conn_unscheduled()` calls in the schedulers and can just totally revert all scheduler changes.

krizhanovsky (changes requested):
One more race in the new design, hopefully the last one. Please carefully verify my conclusions for any other hidden issues.

fw/http.c
ttaym: Found bug is legit, but to me nip request concept is unrelated here. Semantically nip requests cannot be pipelined, but idempotent requests _can be_ pipelined after it. In our case that is exactly the situation we want to avert. No one request must be sent after upgrade request forwarded to backend.

Overall your suggested fix can help, but to me it seams a little brittle. May be all races arise from a fact that we meddle with connection while changes to it drastically change its semantics. May be original approach with provisioning new connection instead of socket steal was not so wrong.

Note, that with your approach upgrade request in fact would block connection with backend until it respond to it. And what if `conns_n=1` in config?

You discard approach with provisioning new connection on a basis of locking server connection list argument. But what if we do not add it to the list or any other list, but give ownership over it to client connection. Specifically if we create separate connection outside the list only to forward upgrade request to it and may be upgrade it. It will be unavailable to any other request.

fw/http_sched_ratio.c
ttaym: See https://github.com/tempesta-tech/tempesta/pull/1595/files#r857456928

fw/http.c
ttaym: BTW, for now i implemented your approach in last comment.

krizhanovsky: > Semantically nip requests cannot be pipelined, but idempotent requests can be pipelined after it.
Actually vise versa: we do pipline a non-idempotent request after idempotent requests, but do not pipline after it.

> what if conns_n=1 in config?

Such configuration works, but I'd say this is 'misconfiguration'. Traditional servers like Nginx or HAproxy establish a new connection with backend for each concurrent request, so a real life Tempesta FW configuration must have way more connections, like 100 and more.

ttaym: I have not found a fix that at the same time was not invasive. So i implemented approach. I have leaved UNSCHED flag on place but introduced marking upgrade request as non-idempotent. This as i can see prevent race condition found by you.

fw/http.c
```diff
@@ -1541,6 +1543,13 @@ tfw_http_req_fwd_send(TfwSrvConn *srv_conn, TfwServer *srv, TfwHttpReq *req,
 	req->jtxtstamp = jiffies;
 	tfw_http_req_init_ss_flags(srv_conn, req);
 
+	/* We set TFW_CONN_B_UNSCHED on server connection. New requests must
+	 * not be scheduled to it. It will be used only for websocket transport.
+	 * If upgrade will fail, we clear it. */
```
krizhanovsky, line 1548: Looks good for met. Could you please just add a comment why we don't have two racing `tfw_http_req_fwd_send()` here? (In my understanding because the function is called under `fwd_qlock`).

krizhanovsky (approved):
Good to merge. Just several minor cleanups/fixes are required.

fw/http_frame.h
```diff
@@ -200,7 +200,7 @@ typedef struct {
 	unsigned char	data_off;
 } TfwH2Ctx;
 
-typedef struct TfwConn TfwConn;
+typedef struct tfw_conn_t TfwConn;
```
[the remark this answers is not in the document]
ttaym: This line of code i had introduced in my earlier commits in previous PRs. It is go againts widly used practice in tempesta code. Mostly type named in snake case with `_t` suffix. I have just fixed style.

fw/websocket.c
```diff
@@ -0,0 +1,355 @@
...
+	conn->sk->sk_user_data = conn;
+
+	conn->destructor = tfw_ws_conn_release;
+	tfw_connection_revive(conn);
+
+	/* Connection destructor does failover for server connections */
+	srv_conn->sk = NULL;
+	if (srv_conn->destructor)
+		srv_conn->destructor(srv_conn);
+	clear_bit(TFW_CONN_B_UNSCHED, &srv_conn->flags);
```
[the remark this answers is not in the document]
ttaym: Done

PR #1596, opened 2022-03-21
Update tempesta.sh

pale-emperor:
Fix deb-package - cant reach default template path at first start


krizhanovsky (approved):
LGTM but it makes sense to redirect error messages in the directory exists, `&2>/dev/null`

PR #1600, opened 2022-03-24
Fix tempesta.sh and tempesta_installer.sh

pale-emperor:
Fix install script with right default grub option
Fix !include pattern to get only *.conf files


krizhanovsky (approved):
LGTM

PR #1605, opened 2022-04-08
Fixed use after destroy

avbelov23:
An array of chunks was created on the stack in the block, and when
the block exits, it was destroyed

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -3381,6 +3381,7 @@ tfw_h2_adjust_req(TfwHttpReq *req)
 	int h_ct_replace = 0;
 	TfwStr h_cl = {0};
 	char cl_data[TFW_ULTOA_BUF_SIZ] = {0};
+	size_t cl_data_len = 0;
 	size_t cl_len = 0;
 	/*
 	 * The Transfer-Encoding header field cannot be in the h2 request, because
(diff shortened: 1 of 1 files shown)


fw/http.c
```diff
@@ -3393,18 +3393,19 @@ tfw_h2_adjust_req(TfwHttpReq *req)
...
-				{ .data = S_DLM, .len = SLEN(S_DLM) },
-				{ .data = cl_data, .len = cl_len },
-				{ .data = S_CRLF, .len = SLEN(S_CRLF) }
-			},
-			.len = 14 + SLEN(S_DLM) + cl_len + SLEN(S_CRLF),
-			.nchunks = 4
-		};
 	}
 
+	h_cl = (TfwStr) {
```
krizhanovsky, line 3398: OK, but we don't need to initialize `h_cl` with zeros any more.

krizhanovsky (approved):
Good to merge with the small cleanup

avbelov23:
DONE

fw/http.c
```diff
@@ -3490,7 +3481,7 @@ tfw_h2_adjust_req(TfwHttpReq *req)
 	h1_hdrs_sz += h_xff.len;
 	h1_hdrs_sz += h_via.len;
 	if (need_cl)
-		h1_hdrs_sz += h_cl.len;
+		h1_hdrs_sz += cl_len;
```
krizhanovsky, line 3484: By default `cl_len` is zero, so there is no need for the condition

fw/http.c
```diff
@@ -3581,8 +3572,19 @@ tfw_h2_adjust_req(TfwHttpReq *req)
...
-	if (need_cl)
+	if (need_cl) {
+		h_cl = (TfwStr) {
+			.chunks = (TfwStr []) {
+				{ .data = "Content-Length", .len = SLEN("Content-Length") },
+				{ .data = S_DLM, .len = SLEN(S_DLM) },
+				{ .data = cl_data, .len = cl_len },
+				{ .data = S_CRLF, .len = SLEN(S_CRLF) }
+			},
+			.len = cl_len,
```
krizhanovsky, line 3583: Oops, now the both the length are initialized into `cl_len`, while they were different in the original code..

krizhanovsky (changes requested):
It seems this version is buggy

avbelov23:
DONE

krizhanovsky (approved):
LGTM!

PR #1609, opened 2022-04-10
Handle SAN certificates

### LINKED ISSUE #1608: SNI matching for SAN certificates
# Scope

Tempesta TLS performs strict matching of SNI against configured vhost name. With this scheme [SAN](https://en.wikipedia.org/wiki/Subject_Alternative_Name) certificates do not work.

The current implementation is wrong to match SNI against vhosts. Instead we need a separate SAN/CN matcher, employed for SNI only.

Besides `tfw_tls_sni()` fix, the Host validation by [`http_host_required` option](https://github.com/tempesta-tech/tempesta/wiki/HTTP-security#message-level-limits) must be fixed. There is no sense to match vhost against SNI in runtime since we can simply check that the vhost name corresponds to SNA/CN from the certificate and print warning.

# Testing

There is a separate issue https://github.com/tempesta-tech/tempesta-test/issues/212

krizhanovsky:
See commit messages. Fixes #1608.

### DIFF
--- a/fw/access_log.c
+++ b/fw/access_log.c
@@ -169,15 +169,12 @@ enum {
 	TRUNCATABLE_FIELDS_COUNT
 };
 
-struct str_ref {
-	const char *data;
-	unsigned len;
-};
-
-/** Copies string to buffer if it is not plain. Returns pointer to the
- * first unused character in the "planar" buffer. */
+/**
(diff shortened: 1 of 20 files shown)


avbelov23 (approved):
LGTM

Issue #1297, opened 2019-07-06
[unittest] HTTP/2 parser


krizhanovsky:
Need to extent `t/unit/test_http_parser.c` to run the same tests against HTTP/2 requests (we do not process HTTP/2 responses yet). The test must carefully test and fuzz the framing layer.

#1560 introduced h2 tests in `t/unit/test_http_parser.c` to not to duplicate all strings parsing and basic tests helping routines. Need to extend the tests to cover the same test cases in the h2 parser as in the http/1 parser. Whether to keep tests for both the parsers in the same file is TDB.

The task is very large, so please make it with a small pull requests, that we can make a smooth progress on it.

- [x] **(crucial!)** Process chunked messages, especially in the headers part to stress hpack decoder. `Authorization` header and other HTTP/2 headers mentioned in https://github.com/tempesta-tech/tempesta/pull/1566/files#r807475198 must be tested with chunked mode to test the fix #1566.
- [x] Body-less methods with Content-Type, Content-Length and/or body (see #1296 )
- [x] All the cases for the [static table](https://datatracker.ietf.org/doc/html/rfc7541#appendix-A)
- [x] dymamic table (this is might be not so easy to do in the unit test, TBD )

There are concerns, that in fact chunk fuzzing not run (or at least run not in full). See https://github.com/tempesta-tech/tempesta/pull/1576#issuecomment-1057878436 . Need to ensure 'with instrumentation and/or parser debugging' facilities that chunk fuzzing fully run as intended and give tangible evidence on this. 

See also #1059.

voodam:
What also is good to be checked ana maybe fixed:
1. This code in `tfw_h2_parse_req_finish` is probably wrong:
```
	if (req->content_length
	    && req->content_length != req->body.len)
	{
		return T_DROP;
	}
	/*
	 * RFC 7540 8.1.2.6: A request or response that includes a payload
	 * body can include a content-length header field.
	 *
	 * Since the h2 provides explicit message framing, the content-length
	 * header is not required at all. But our code may rely on
	 * req->content_length field value, fill it.
	 */
	req->content_length = req->body.len;
```
First time `req->content_length != req->body.len` is checked, then `req->content_length = req->body.len` is set.

Also for now `content_length` have limit `LONG_MAX` to fit `parser->to_read` field, so maybe `req->body.len` also need to be limited in the same way.

2. Is it OK that in `tfw_h2_parse_body` on this line `long` type is passed when `len` is `unsigned long`?
[...]

PR #1616, opened 2022-04-19
Separate hpack parser tests

alegz-filomaphitskay:
Contributes to #1297

What was done here:
- move hpack_static_table tests to individual file;
- loop that used for iterating chunks for TRY_PARSE_EXPECT_PASS/TRY_PARSE_EXPECT_BLOCK was moved inside defines;
  - after that fuzzer's tests became use chunks for requests. Before that, all  fuzzer's requests were run on the size of the first chunk;
- now validate_data_fully_parsed() doesn't use static variable hm_exp_len (it became local);
- added FOR_REQ_H2_HPACK/EXPECT_BLOCK_REQ_H2_HPACK that not call test_case_parse_prepare_h2 (this will be needed for dynamic table tests).


fw/t/unit/test_http2_parser_hpack.c
```diff
@@ -0,0 +1,269 @@
+/**
+ *		Tempesta FW
+ *
+ * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
+ * Copyright (C) 2015-2022 Tempesta Technologies, Inc.
```
krizhanovsky, line 5: This is a new file, so please leave only `Copyright (C) 2022 Tempesta Technologies, Inc.`

fw/t/unit/test_http_parser_common.h
```diff
@@ -572,32 +554,39 @@ do_split_and_parse(int type, int chunk_mode)
...
 static
 int __attribute__((unused))
-validate_data_fully_parsed(int type)
+validate_data_fully_parsed(int type, size_t sz_diff)
 {
 	TfwHttpMsg *hm = (type == FUZZ_REQ || type == FUZZ_REQ_H2)
 			? (TfwHttpMsg *)req
 			: (TfwHttpMsg *)resp;
 
+	size_t hm_exp_len = GET_FRAMES_TOTAL_SZ() - sz_diff;
```
krizhanovsky, line 563: Just use `frames_total_sz` - there is no reason to mask it to a function. The same for `GET_FRAMES_MAX_SZ` and `GET_CURRENT_FRAME()`, the last one should be just
```C
#define CURRENT_FRAME     frames[frame_index]
```

fw/t/unit/test_http_parser_common.h
```diff
@@ -572,32 +554,39 @@ do_split_and_parse(int type, int chunk_mode)
...
+	    || !validate_data_fully_parsed(type, sz_diff))		\
 		TEST_FAIL("can't parse %s (code=%d)\n",			\
 			  (type == FUZZ_REQ	    			\
 			   || type == FUZZ_REQ_H2			\
 			   ? "request" : "response"),			\
 			  _err);					\
 	__fpu_schedule();						\
 	_err == TFW_PASS;						\
-})
+}))
```
krizhanovsky, line 582: There is no usage of the macro, when the `while` loop has body, so please reorganize into a normal loop with a body, not just do everything in the condition

fw/t/unit/test_http_parser_common.h
```diff
@@ -649,19 +648,25 @@ do {									\
 	    FUZZ_REQ_H2, CHUNK_ON)
 #define EXPECT_BLOCK_REQ_H2_CHUNK_OFF(str)				\
 	__EXPECT_BLOCK_REQ(str, FUZZ_REQ_H2, CHUNK_OFF)
+#define EXPECT_BLOCK_REQ_H2_HPACK(H2_FRAMES_DEF_BLOCK)			\
+	INIT_FRAMES();							\
+	SET_FRAMES_BUF(frames_buf);					\
+	H2_FRAMES_DEF_BLOCK;						\
+	RESET_FRAMES_BUF();						\
+	TRY_PARSE_EXPECT_BLOCK(FUZZ_REQ_H2, CHUNK_ON)
```
krizhanovsky, line 656: Please use `do { .. } while (0)` for multi-line defines here and for all the code.

Also please empty lines between multi-line macro definitions

krizhanovsky (approved):
Good to merge after coding style cleanups

fw/t/unit/test_http_parser_common.h
alegz-filomaphitskay: I didn't quite understand the comment. We must use that ugly construction in condition because in some cases we add check-block after implicit EXPECT-macro. Like this:
```
 	FOR_REQ_H2(
	    HEADERS_FRAME_BEGIN();
		// :method = "GET"
		HEADER(INDEX(2));
		// :scheme = "https"
		HEADER(INDEX(7));
		// :path = "/"
		HEADER(INDEX(4));
	    HEADERS_FRAME_END();
	)
	{
		EXPECT_EQ(req->method, TFW_HTTP_METH_GET);
		EXPECT_TFWSTR_EQ(&req->h_tbl->tbl[TFW_HTTP_HDR_H2_SCHEME],
				 ":scheme" "https");
		EXPECT_TFWSTR_EQ(&req->uri_path, "/");
	}
```
I can't take and move all the operations from the condition block to the loop body. So what do I need to do?

krizhanovsky: I just didn't find a place where the macro has a body. If there is a place, where the macro has a body, then my comment is wrong

fw/t/unit/test_http_parser_common.h
alegz-filomaphitskay: In the case of FOR_REQ_H2_HPACK and others like it, this is not possible. Because after FOR_REQ_H2_HPACK you can add a check block, which is actually the body of the while loop from TRY_PARSE_EXPECT_PASS.
In the current PR I can use `do { .. } while (0)` for the EXPECT_BLOCK_REQ_H2_HPACK case because TRY_PARSE_EXPECT_BLOCK cannot have a check block like TRY_PARSE_EXPECT_PASS. But in some next PR, EXPECT_BLOCK_REQ_H2_HPACK already has the ability to use a check block similar to FOR_REQ_H2_HPACK.

Issue #856, opened 2017-11-06
Custom HTTP redirects


krizhanovsky:
While [HSTS](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) ([RFC 6797](https://tools.ietf.org/html/rfc6797)) can be implemented just by adding `Strict-Transport-Security` header, there also must be implemented redirect from HTTP port to the same URL by HTTPS (HTTP redirects are common practice for HSTS implementation).

Please add a functional test for HSTS checking that the right header is added to all the responses and HTTP request is correctly redirected to HTTPS.

The [Wiki](https://github.com/tempesta-tech/tempesta/wiki/HTTP-security#http-strict-transport-security-hsts) update is required -  we have to add redirect statement.

krizhanovsky:
Adding the header is just enough, see Nginx's [doc](https://www.nginx.com/blog/http-strict-transport-security-hsts-and-nginx/) for example. So I added the header description to the Wiki pages

- https://github.com/tempesta-tech/tempesta/wiki/Modify-HTTP-Messages
- https://github.com/tempesta-tech/tempesta/wiki/HTTP-security

and close the task.

krizhanovsky:
Actually, while we're send HTTP redirects on JS and cookie challenges, we're still unable to send custom redirects, e.g. if a user wants to redirect all HTTP traffic to some other HTTPS resource. Anyway HTTP redirects are very useful in for HTTP proxies, so I redefine the issue and reopen it.

In particular HTTP tables must be able to redirect an HTTP request by a new redirect action. Also see how Nginx can redirect all the requests for a particular location (you can find an example on the link above).

Please update the wiki pages mentioned above and the page for HTTP tables. Please provide full example for HSTS (like there is a full configuration for Nginx in the  doc).

Issue #907, opened 2018-02-10
Variables and conditions for custom HTTP headers


krizhanovsky:
Further development of #511 (HTTP: transform custom HTTP headers). Depends on the new rule syntax introduced in #731 (HTTP tables) and #531 (Upgrade/downgrade HTTP messages).

### Use case (for Wiki)

If several of the backend servers in the [WAF acceleratiotn](https://github.com/tempesta-tech/tempesta/wiki/WAF-acceleration#waf-acceleration) scenario have public IPs and allows HTTP/1.0, then, even announcing the public IPs as virtual IPs (VIPs) on Tempesta FW intefaces, the WAF, configured in reverse proxy mode, doesn't have an idea which backend to send an HTTP/1.0 requests to.

### Conditional HTTP upgrade

The first solution is to upgrade forwarded HTTP/1.0 request without Host header to HTTP/1.1 or HTTP/2 with correct setting of Host header. We can do this using a rule to translate the backend public IP to it's Host value (the condition construction is the subject got multi-pattern matching optimization in #732, so keep the further optimization in mind):
```shell
$dst_ip == 1.1.1.1 -> req_hdr_add Host "foo.com";
$dst_ip == 2.2.2.2 -> req_hdr_add Host "bar.com";
```
`$dst_ip` is the matching variable for skb destination IP address.

### Setting X-Server-IP

Another solution is just to add an extension header with the backend public IP:
```shell
req_hdr_add X-Server-IP $dst_ip
```
and the the WAF to make the correct message forwarding based on the header.

### The variables

Following variables must be implemented:

1. `$proto` - a protocol (`HTTP/1.0`, `HTTP/1.1` etc) used by a client in a request;
1. `$dst_addr` - client skb destination address;
1. `$dst_port` - client skb destination port;
1. `$src_addr` - client skb source address;

### References

* Amazon Elasitc Load Balancing, [HTTP Headers and Classic Load Balancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html)

PR #1618, opened 2022-04-22
Custom HTTP redirects

avbelov23:
#856

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -115,8 +115,10 @@
 #define S_H2_STAT		":status"
 
 #define RESP_BUF_LEN		128
+#define URL_BUF_LEN		1024
 
 static DEFINE_PER_CPU(char[RESP_BUF_LEN], g_buf);
+static DEFINE_PER_CPU(char[URL_BUF_LEN], g_url_buf);
 
 #define TFW_CFG_BLK_DEF		(TFW_BLK_ERR_REPLY)
 unsigned short tfw_blk_flags = TFW_CFG_BLK_DEF;
@@ -1117,6 +1119,17 @@ err:
 	tfw_http_resp_build_error(req);
 }
 
+/*
+ * Perform operations to sending an custom HTTP2 response to a client.
+ * Set current date in the header of an HTTP response.
+ * If memory allocation error or message setup errors occurred, then
+ * client connection should be closed, because response-request
+ * pairing for pipelined requests is violated.
+ *
+ * NOTE: The first chunk is a status line, and then every odd chunk is a header
+ * field name starting with CRLF and ending with ': ', and every even chunk is
+ * a value.
+ */
 static void
 tfw_h2_send_resp2(TfwHttpReq *req, TfwStr *msg, int status,
 	              unsigned int stream_id)
@@ -1190,6 +1203,17 @@ err:
 	tfw_http_resp_build_error(req);
 }
 
+/*
+ * Perform operations to sending an custom HTTP1 response to a client.
+ * Set current date in the header of an HTTP response.
+ * If memory allocation error or message setup errors occurred, then client
+ * connection should be closed, because response-request pairing for pipelined
+ * requests is violated.
+ *
+ * NOTE: The first chunk is a status line, and then every odd chunk is a header
+ * field name starting with CRLF and ending with ': ', and every even chunk is
+ * a value.
+ */
 static void
 tfw_h1_send_resp2(TfwHttpReq *req, TfwStr *msg, int status)
 {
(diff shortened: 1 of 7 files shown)


avbelov23:
Perhaps it would be convenient for ease of inserting variable values into placeholders (`$host`, `$request_uri`) at the config stage to break the redir url into chunks

fw/http_tbl.h
```diff
@@ -23,6 +23,25 @@
...
 
+typedef enum {
+    TFW_HTTP_RES_VHOST,
+    TFW_HTTP_RES_REDIR,
+    _TFW_HTTP_RES_COUNT
+} tfw_http_res_act_t;
+
+typedef struct {
+    char *url;
+    unsigned int resp_code;
```
krizhanovsky, line 34: `unsigned short` is enough. The structure should be extended - see other comments, but also remember to not have memory holes.

fw/http.c
```diff
@@ -1320,6 +1478,112 @@ tfw_http_send_resp(TfwHttpReq *req, int status, const char *reason)
...
+	}                                                                    \
+	url_p = url + dif;                                                   \
+	TFW_STR_FOR_EACH_CHUNK(c, (from), end) {                             \
+		memcpy_fast(url_p, c->data, c->len);                             \
+		url_p += c->len;                                                 \
+	}                                                                    \
+} while (0)
+
+	for (; *pattern; ++pattern) {
+		if (*pattern == '$') {
```
krizhanovsky, line 1529: This is not good to do all this stuff in runtime, especially in softirq and especially with `GFP_KERNEL`. Let's introduce helping flags to `TfwHttpRedir`: which variables are used in the rule (say up to 8) and an array of variable positions and types (the array is of variable length, which isn't a problem on config type). `TfwHttpRedir->url` having variables should be `TfwStr`.

We need to extend `tfw_http_req_t` with the array of variable (see https://github.com/tempesta-tech/tempesta/issues/907). At some point we might arrange all the fields of the the request, which can be referenced by variables, in a special headers array (this is quite intuitive for HTTP/2), but now just let's keep references to TfwStr and please write a comment that it's good to optimize to avoid indirect references.

Variables in a compiled pattern for `TfwHttpRedir` should be just static indexes in the variables array in `tfw_http_req_t`.

In this function we just need to check the flags and merge TfwStr by the positions where the variables are presented.

fw/http.c
```diff
@@ -1320,6 +1478,112 @@ tfw_http_send_resp(TfwHttpReq *req, int status, const char *reason)
...
+			if (pattern_end - pattern > SLEN("request_uri") &&
+				!strncmp(pattern + 1, "request_uri", min(SLEN("request_uri"), len)))
+			{
+				TFW_STRCPY(&req->uri_path);
+				pattern += SLEN("request_uri");
+			} else if (pattern_end - pattern > SLEN("host") &&
+				       !strncmp(pattern + 1, "host", min(SLEN("host"), len)))
+			{
+				TFW_STRCPY(&req->host);
+				pattern += SLEN("host");
```
krizhanovsky, line 1541: I don't see any bug in the code, but I tested `uri == "/blog/" -> 301 = https://$host$request_uri;` - `$request_uri` does work, but `$host` is always empty. Maybe we need to consider which host to use - from URI or from Host header (you use the first one, the second one is stored in the headers table).

fw/http.c
```diff
@@ -1320,6 +1478,112 @@ tfw_http_send_resp(TfwHttpReq *req, int status, const char *reason)
...
+
+		tfw_h1_send_resp2(req, msg, status);
+	}
+}
+
+static void
+tfw_http_req_redir(TfwHttpReq *req, int status, const char *pattern)
+{
+	size_t url_len = strlen(pattern) + 1;
+	char *url = kmalloc(url_len, GFP_KERNEL);
```
krizhanovsky, line 1504: Ideally, if you can rewrite the code (having the comment at the below in mind) without the dynamic allocation, just using `TfwStr` fragments. However, this might be too complex and we also discussed that too many small fragments hurt performance, so, if you end up with the need of dynamic memory allocation, then just allocate a per-cpu static buffer, just like the current `g_buf` (maybe you'll have an idea how to aggregate them in an elegant way).

krizhanovsky (changes requested):
The code does work except `$host` variable, but there are things to optimize and improve, which are good to make before the merge.

fw/http.c
```diff
@@ -1076,6 +1117,123 @@ tfw_h1_send_resp(TfwHttpReq *req, int status)
 	tfw_http_resp_build_error(req);
 }
```
krizhanovsky, line 1119: There are already `tfw_h2_prep_redirect()` and `tfw_h1_prep_redirect()`, so please comment the new redirection functions (allocating a new responses) and these two old ones (rewriting server responses) about the differences.

fw/http.c
```diff
@@ -1307,17 +1489,133 @@ tfw_http_nip_req_resched_err(TfwSrvConn *srv_conn, TfwHttpReq *req,
...
+		WARN_ONCE(!list_empty_careful(&req->msg.seq_list),
+			  "Request is already in seq_queue\n");
+		tfw_stream_unlink_msg(req->stream);
+		spin_lock(&cli_conn->seq_qlock);
+		list_add_tail(&req->msg.seq_list, &cli_conn->seq_queue);
+		spin_unlock(&cli_conn->seq_qlock);
+
+		tfw_h1_send_resp(req, msg, status);
+	}
+}
```
krizhanovsky, line 1522: Shouldn't the functions be `tfw_http_send_redir()`, `tfw_h1_send_redir()` and `tfw_h2_send_redir()`? They send redirects only, no any HTTP response

krizhanovsky (approved):
Let's merge it and do refactoring on the next week

fw/http_tbl.h
avbelov23: Parsing wants integers. Making an intermediate variable looks redundant

fw/http.c
```diff
@@ -6259,13 +6533,20 @@ tfw_http_hm_srv_send(TfwServer *srv, char *data, unsigned long len)
...
 			    TFW_WITH_PORT);
 		goto cleanup;
 	}
+	if (res.type == TFW_HTTP_RES_REDIR) {
+		goto cleanup;
+	} else if (res.type == TFW_HTTP_RES_VHOST) {
+		req->vhost = res.vhost;
+	} else {
+		BUG();
+	}
```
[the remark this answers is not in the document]
avbelov23: It's not obvious when not all action types might need to wait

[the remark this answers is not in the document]
krizhanovsky: For the call of `tfw_http_tbl_action()` in `tfw_http_req_process()` we should pass only `TFW_HTTP_RES_VHOST` in the mask, so HTTPtables will care only about these rules ignoring for example `TFW_HTTP_RES_REDIR`. In `tfw_http_hm_srv_send()` we pass something like `TFW_HTTP_RES_REDIR | TFW_HTTP_RES_VHOST`, so all these rules will be processed.

PR #1620, opened 2022-04-28
Use unsigned_int instead of size_t in ss_skb_actor_t

alegz-filomaphitskay:
Contributes to https://github.com/tempesta-tech/tempesta/issues/1297.

[@krizhanovsky](https://github.com/krizhanovsky) said very well:
> `len` in `ss_skb_actor_t()` comes from `skb_headlen()`, which is `unsigned int`. `unsigned int headlen` also makes the code more robust.

What was done here:

- used `unsigned int` instead of `size_t` in `ss_skb_actor_t`;
- also `unsigned int` is used instead of `size_t` at major network data processing points.

### DIFF
--- a/fw/http_frame.c
+++ b/fw/http_frame.c
@@ -1481,7 +1481,7 @@ conn_term:
  * Main FSM for processing HTTP/2 frames.
  */
 static int
-tfw_h2_frame_recv(void *data, unsigned char *buf, size_t len,
+tfw_h2_frame_recv(void *data, unsigned char *buf, unsigned int len,
 		  unsigned int *read)
 {
 	int n, r = T_POSTPONE;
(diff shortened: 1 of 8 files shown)


keshonok:
Frankly, I would have added an explanatory text to the commit message to explain why the change was necessary. Otherwise it would be totally obscure to anyone tracking the changes in the code (including any future contributors).

alegz-filomaphitskay:
> Frankly, I would have added an explanatory text to the commit message to explain why the change was necessary. Otherwise it would be totally obscure to anyone tracking the changes in the code (including any future contributors).

Thanks for the advice. But it is just only draft for view files changes.

krizhanovsky (approved):
Good and actually clear patch: `len` in `ss_skb_actor_t()` comes from `skb_headlen()`, which is `unsigned int`. `unsigned int headlen` also makes the code more robust.

PR #1621, opened 2022-04-29
Pretty print for H2 requests in log

alegz-filomaphitskay:
Contributes to https://github.com/tempesta-tech/tempesta/issues/1297.

What was done here:
- added pretty print for H2 requests;
- the historical usage of the __FOR_REQ macro is redundant for macros of the FOR_REQ_H2-family. It is more correct to use the __FOR_REQ macro only for HTTP/1 requests;
- removed unused FOR_REQ_H2_CHUNK_OFF and EXPECT_BLOCK_REQ_H2_CHUNK_OFF macros;
- some cosmetic changes.


krizhanovsky (approved):
LGTM

PR #1622, opened 2022-05-02
EXPECT_BLOCK-macro post-checks

alegz-filomaphitskay:
Thus we can use post-checks inside while-body related to EXPECT_BLOCK-macro for every chunk size but not only for the last.

### DIFF
--- a/fw/t/unit/test_http_parser_common.h
+++ b/fw/t/unit/test_http_parser_common.h
@@ -979,26 +979,20 @@ do {\
 	TRY_PARSE_EXPECT_PASS(FUZZ_REQ_H2, CHUNK_ON)
 
 #define EXPECT_BLOCK_REQ(str)						\
-do {									\
 	PRINT_REQ(str);							\
 	test_case_parse_prepare_http(str);				\
-	TRY_PARSE_EXPECT_BLOCK(FUZZ_REQ, CHUNK_ON);			\
-} while (0)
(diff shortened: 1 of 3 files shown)


fw/t/unit/test_http_parser_common.h
```diff
@@ -669,11 +663,9 @@ do {									\
 #define FOR_RESP(str)	__FOR_RESP(str, 0, CHUNK_ON)
 
 #define EXPECT_BLOCK_RESP(str)						\
-do {									\
 	TEST_LOG("=== response: [%s]\n", str);				\
 	test_case_parse_prepare_http(str);				\
-	TRY_PARSE_EXPECT_BLOCK(FUZZ_RESP, CHUNK_ON);			\
-} while (0)
+	TRY_PARSE_EXPECT_BLOCK(FUZZ_RESP, CHUNK_ON)
```
krizhanovsky, line 668: I believe there is no reason to remove `do { } while (0)` statements.

krizhanovsky (approved):
The first part of PR is good to merge, but please revert the changes in the second part

fw/t/unit/test_http_parser_common.h
alegz-filomaphitskay: No, if `do { } while (0)` remains, then the checks from the check block will always be performed only for the last chunk size. More precisely, these checks will always work, regardless of the result of running TRY_PARSE_EXPECT_BLOCK.

PR #1623, opened 2022-05-02
Additional fix and the code documentation improvements for #1597

krizhanovsky:
Please see commit messages and code comments for the details


alegz-filomaphitskay (approved):
Have to agree with every change.

Issue #1377, opened 2020-01-10
TfwStream objects leaks if errors happen on serving h2 clients 


vankoven:
## Scope

- [x] To reproduce force Tempesta to close client connection with error. Or force client to close the connection with Tempesta with e.g. PROTOCOL_ERROR code. If this happens, connection is closed, but TfwStream object is not destroyed.

```
[tdb] Close table 'client0.tdb'
[tdb] Close table 'sessions0.tdb'
[tdb] Close table 'cache0.tdb'
[tdb] Close table 'filter0.tdb'
[tempesta fw] modules are stopped
[tempesta fw] exiting...
=============================================================================
BUG tfw_stream_cache (Tainted: G    B   W  O   ): Objects remaining in tfw_stream_cache on __kmem_cache_shutdown()
-----------------------------------------------------------------------------

INFO: Slab 0xffffd48381f8a800 objects=23 used=1 fp=0xffffa0a63e2a1340 flags=0xffffc000008100
CPU: 0 PID: 28679 Comm: rmmod Tainted: G    B   W  O    4.14.0-tempesta-amd64 #1 Debian 4.14.32-tfw7-1
Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS ?-20191223_100556-anatol 04/01/2014
Call Trace:
 dump_stack+0x5c/0x85
 slab_err+0xad/0xd0
 ? __kmalloc+0x164/0x1e0
 __kmem_cache_shutdown+0x1c2/0x3a0
 ? wait_for_completion+0x140/0x190
 shutdown_cache+0x1a/0x1b0
 kmem_cache_destroy+0x1e5/0x210
 tfw_tls_exit+0x2e/0x40 [tempesta_fw]
 ? tfw_tls_free_alpn_protos+0x30/0x30 [tempesta_fw]
 tfw_exit+0x3f/0x70 [tempesta_fw]
 SyS_delete_module+0x1b2/0x280
 do_syscall_64+0x8d/0x120
 entry_SYSCALL_64_after_hwframe+0x3d/0xa2
RIP: 0033:0x7f22a23a85e7
RSP: 002b:00007fff4d6d79a8 EFLAGS: 00000206 ORIG_RAX: 00000000000000b0
RAX: ffffffffffffffda RBX: 00007fff4d6d79f8 RCX: 00007f22a23a85e7
RDX: 000000000000000a RSI: 0000000000000800 RDI: 000055ece9d31808
[...]

PR #1632, opened 2022-05-29
Fix minor warning for #1377

krizhanovsky:
If a server sends empty header (e.g. `add_header X-Crash-1377 ' ';` for Nginx), then `tfw_hpack_node_compare()` may raise `WARN_ON_ONCE(state != HPACK_HDR_VALUE_FOUND)`.

In case of empty header `tfw_hpack_add_node()` -> `tfw_http_hdr_split()` strips spaces and ':' so we may fall to the warning statement with matching the header name only, but `HPACK_HDR_OWS_PROCESS()` didn't move `state` to `HPACK_HDR_NAME_FOUND` (beacuse no spaces and no ':').

The patch moves the `hdr` processing tail out of the loop, so hlen might be non-zero after the main loop and `WARN_ON_ONCE(hlen)` is also eliminated.

### DIFF
--- a/fw/hpack.c
+++ b/fw/hpack.c
@@ -2233,17 +2233,13 @@ typedef enum {
 									\
 	BUG_ON(len < (part_len));					\
 	if (state != HPACK_HDR_NAME_SEARCH				\
-	    && (state != HPACK_HDR_VALUE_FOUND				\
-		|| chunk != TFW_STR_LAST(hdr)				\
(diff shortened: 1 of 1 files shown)


HDR:
Going to assume the mention was unintentional 😄

krizhanovsky:
> Going to assume the mention was unintentional smile

@hdr, Yes, sorry - just copy & pasted comment with doxygen's `@`.

s0nx (approved):

Issue #1617, opened 2022-04-19
HPACK dyn tables do not update special fields in req structure


alegz-filomaphitskay:
See second request below that use indexes from hpack dynamic table:
```
	FOR_REQ_H2_HPACK(
	    HEADERS_FRAME_BEGIN();
		// :method = "GET"
		HEADER(INDEX(2));
		// :scheme = "https"
		HEADER(INDEX(7));
		// :path = "/"
		HEADER(INDEX(4));
		// host = "developer.mozilla.org:5588"
		HEADER(INC_IND(INDEX(38), VALUE("developer.mozilla.org:5588")));
		// if-modified-since = "Inv, 31 Jan 2012 15:02:53"
		HEADER(INC_IND(INDEX(40), VALUE("Inv, 31 Jan 2012 15:02:53 GMT")));
	    HEADERS_FRAME_END();
	)
	{
		EXPECT_TFWSTR_EQ(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST],
				 "host" "developer.mozilla.org:5588");
		EXPECT_EQ(req->host_port, 5588);
		EXPECT_TFWSTR_EQ(&req->h_tbl->tbl[TFW_HTTP_HDR_RAW],
				 "if-modified-since"
				 "Inv, 31 Jan 2012 15:02:53 GMT");
		EXPECT_TRUE(req->cond.m_date == 1328022173);
		EXPECT_TRUE(req->cond.flags & TFW_HTTP_COND_IF_MSINCE);
	}

#define TOTAL_CNT_INC_IND 2

#define HPACK_DYN_TBL_INDEX 62
#define DYN_IND(entry_seq_no) \
	(HPACK_DYN_TBL_INDEX + TOTAL_CNT_INC_IND - entry_seq_no)

	FOR_REQ_H2_HPACK(
	    HEADERS_FRAME_BEGIN();
		// :method = "GET"
		HEADER(INDEX(2));
		// :scheme = "https"
		HEADER(INDEX(7));
		// :path = "/"
		HEADER(INDEX(4));
		// host = "developer.mozilla.org:5588"
		HEADER(INDEX(DYN_IND(1)));
		// if-modified-since = "Inv, 31 Jan 2012 15:02:53"
		HEADER(INDEX(DYN_IND(2)));
	    HEADERS_FRAME_END();
	)
	{
[...]

PR #1635, opened 2022-06-03
HPACK dyn tables do not update special fields in request

s0nx:


s0nx:
Fix to the issues mentioned in #1617

fw/t/unit/test_http2_parser_hpack.c
```diff
@@ -845,6 +840,37 @@ TEST(http2_parser_hpack, increment_all_static_indexes_for_request)
...
+		HEADER(DYN_TBL_INDEX(1));
+		// :scheme = "https"
+		HEADER(DYN_TBL_INDEX(2));
+		// :path = "/"
+		HEADER(DYN_TBL_INDEX(3));
+	    HEADERS_FRAME_END();
+	)
+	{
+		EXPECT_EQ(req->method, TFW_HTTP_METH_GET);
+	}
```
krizhanovsky, line 873: The test case is artificial. Index 2 in static table means `:method GET`, so while I didn't find explicit prohibition of sending `GET` value for index 2 in the RFC, I still don't think that we need to make our code to accept this kind of requests. I believe we must block such requests.

However, `$ ./h2spec http2/8.1.2.3 -t -h ubuntu -p 443 -k` (`ubuntu` is the name of my VM) raises the same warning. It seems this test https://github.com/summerwind/h2spec/blob/master/http2/8_1_2_1_pseudo_header_fields.go#L54 leads to the warning. Maybe because we have `TFW_TAG_HDR_H2_METHOD = TFW_TAG_HDR_H2_STATUS`. Need to investigate the problem more.

In general, we should keep our parser as strict as possible and allow only direct RFC requirements. I don't see in
>  If the header field name matches the header field name of an entry stored in the static table or the dynamic table, the header field name can be represented using the index of that entry.  In this case, the index of the entry is represented as an integer with a 6-bit prefix (see Section 5.1).  This value is always non-zero.

requirements to process the requests as in the test above. In my understanding this RFC citation is about name only indexes, e.g. 1.

We must satisfy common tools like `curl`, browsers, `h2spec` and so on. We should not process all imaginary requests, which aren't directly contradict RFCs. In this sense I believe this part of #1617 is incorrect.

fw/hpack.c
```diff
@@ -1191,13 +1191,17 @@ tfw_hpack_hdr_set(TfwHPack *__restrict hp, TfwHttpReq *__restrict req,
...
+		 * to parse the method here.
+		 */
 		if (hp->index == 2) {
 			req->method = TFW_HTTP_METH_GET;
 		} else if (hp->index == 3) {
 			req->method = TFW_HTTP_METH_POST;
 		} else {
-			WARN_ON_ONCE(1);
-			return T_DROP;
+			tfw_idx_hdr_parse_method(req, entry->hdr);
```
krizhanovsky, line 1204: I think the change is incorrect and dangerous, see the comment about the test case from #1617 and `h2spec`. `h2spec`, `curl` and browsers like Firefox or Chrome are the good references to check which cases we should support.

By the way, it makes sense to run `h2spec` before the PR and with the PR and compare whether some issues are gone or some new issues were introduced. Let's run the tool for all h2 specific changes to make sure that we're on the right track. We can't switch the test on for CI because at the moment (with this PR) we have `146 tests, 126 passed, 1 skipped, 19 failed`, so several tests are failing. At some point we need to fix all the issues besides the issues about the features, which are in the TODO/doubtful list, e.g. server PUSH. @const-t FYI also.

fw/http_parser.c
```diff
@@ -9039,6 +9039,63 @@ tfw_h2_parse_req_finish(TfwHttpReq *req)
...
+
+void
+tfw_idx_hdr_parse_method(TfwHttpReq *req, TfwStr *hdr)
+{
+	TfwStr *c, *end;
+
+	TFW_STR_FOR_EACH_CHUNK(c, hdr, end) {
+		if (c->flags & TFW_STR_HDR_VALUE) {
+			int ret = tfw_h2_parse_req_hdr(c->data, c->len, req,
+							true, true);
```
krizhanovsky, line 9091: The parser function is heavy and it's not good to call the whole parser function just for method. If we 100% need to call the parser from hpack, then we probably need to move the method parsing into a separate small function. However, it seems this is a misconception since `tfw_hpack_hdr_set()` is called from `HPACK_STATE_ALL_INDEXED`, which implies that the whole header is fully indexed.
 
The current approach is `tfw_h2_parse_req()` -> `tfw_hpack_decode()` (decode indexes and Huffman) -> `HPACK_DECODE_PROCESS_STRING()`, which leads to `__hpack_process_hdr_value()` or `__hpack_process_hdr_name()` for Huffman encoded strings, or `HPACK_PROCESS_STRING(), which leads to `tfw_h2_parse_req_hdr()` for plain strings.

It seems the bug is in the state machine inside `tfw_hpack_decode()` and we should still process `Host` and `If-Modifies-Since` for the test as `HPACK_STATE_VALUE_TEXT`.

krizhanovsky (changes requested):
There are many questions. Besides seems incorrect test in #1617 it seems we a different way to fix the issues for `Host` and `If-Modified-Since` is required by the original design of HPACK and the parser interactions.

fw/t/unit/test.h
```diff
@@ -77,6 +77,17 @@ do {				\
 #define TEST_DBG3(...)
 #endif
 
+#if defined(DEBUG) && (DEBUG >= 3)
+#define SLEEP_TIME	15*1000
+#define TEST_WAIT(...)				\
+do {						\
+	TEST_LOG("postpone exec for 15s...\n");	\
+	msleep(SLEEP_TIME);			\
```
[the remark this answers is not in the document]
s0nx: This was added here to ease the debugging. The test actually begins right from the init_module(), so this msleep() here would make it possible to pause the execution (if using Tempesta in a VM) and load needed symbols, set breakpoints and so on.

fw/t/unit/test_http2_parser_hpack.c
s0nx: I agree. I've checked sources of nghttp2 and Golang std lib related to http2 and this test case is wrong.
Furthermore, I've found the reason of that warning. It has nothing to with method. As you've suggested, there was an issue with the HPACK hdr tags enum.

fw/hpack.c
```diff
@@ -1253,13 +1261,47 @@ tfw_hpack_hdr_set(TfwHPack *__restrict hp, TfwHttpReq *__restrict req,
...
+	case TFW_TAG_HDR_PRAGMA:
+	case TFW_TAG_HDR_REFERER:
+	case TFW_TAG_HDR_X_FORWARDED_FOR:
+	case TFW_TAG_HDR_USER_AGENT:
+	case TFW_TAG_HDR_RAW:
+		is_valid = true;
+		break;
+	default:
+		T_WARN("%s: HTTP/2 request dropped: invalid HPACK entry tag"
+			" = %d\n", __func__, entry_tag);
```
krizhanovsky, line 1300: I think this should be a debug message: we print warnings to users and users may struggle on watching function names. Also the wild Internet traffic is full of bots scanning IP ranges in security vulnerabilities, so we might see a lot of such messages.

fw/hpack.c
```diff
@@ -1253,13 +1261,47 @@ tfw_hpack_hdr_set(TfwHPack *__restrict hp, TfwHttpReq *__restrict req,
...
 
 	return T_OK;
 }
 
+static bool
+tfw_hpack_entry_tag_valid(unsigned int entry_tag)
+{
+	bool is_valid = false;
+
+	switch(entry_tag) {
```
krizhanovsky, line 1277:
```suggestion
	switch (entry_tag) {
```

fw/hpack.c
```diff
@@ -1435,7 +1477,8 @@ tfw_hpack_decode(TfwHPack *__restrict hp, unsigned char *__restrict src,
 			       __func__, hp->index);
 			WARN_ON_ONCE(!hp->index);
 			entry = tfw_hpack_find_index(&hp->dec_tbl, hp->index);
-			if (!entry || tfw_hpack_hdr_name_set(hp, req, entry)) {
+			if (!entry || !tfw_hpack_entry_tag_valid(entry->tag) ||
+				tfw_hpack_hdr_name_set(hp, req, entry)) {
```
krizhanovsky, line 1481: `tfw_hpack_entry_tag_valid()` is always and only called after `tfw_hpack_find_index()`, so I propose to call it from `tfw_hpack_find_index()` and make the function not to return invalid `entry`

krizhanovsky (approved):
LGTM with the couple of cleanups

krizhanovsky:
BTW I really appreciate the small commits with good messages - it's quite trivial to walk over the each of the commit and analyze the idea of each small change!

s0nx:
We might have a problem with my latest change. Some HPACK tests are failing now.
The problem is that `tfw_hpack_entry_tag_valid()` checks `entry->tag` value against the list of 'allowed' tags, and it's not there,
so the `entry == NULL` and the test fails.

I've pushed a small [patch](https://github.com/tempesta-tech/tempesta/pull/1635/commits/15637ea7bbe067b4b449e989d29c10f443bf52e2) and tests do pass now, but I'm not sure if it's correct to add `transfer-encoding` to the list of allowed tags.

fw/hpack.c
```diff
@@ -927,6 +927,48 @@ tfw_hpack_add_index(TfwHPackDTbl *__restrict tbl,
...
+	case TFW_TAG_HDR_TRANSFER_ENCODING:
+	case TFW_TAG_HDR_RAW:
+		is_valid = true;
+		break;
+	default:
+		T_DBG3("%s: HPACK entry tag (%d) is not valid for HTTP/2 req\n",
+			__func__, entry_tag);
+	}
+	return is_valid;
+}
```
krizhanovsky, line 970: `entity->tag` is used to reference HTTP "special" headers (see `tfw_http_hdr_t` in http.h) and set `parser->_hdr_tag`, which is used by `tfw_http_msg_hdr_close()` for direct reusage of the static table entries for our special hesders table (see `tfw_hpack_hdr_set()`). Thus, `tfw_hpack_entry_tag_valid()` must validate all the used `tfw_http_hdr_t` constants for `tag` value in `static_table` from hapck.c. 

In this sense `set-cookie` should have `TFW_HTTP_HDR_SET_COOKIE` tag in `static_table`. Please also add test for this and maybe other headers if I missed some of them.

krizhanovsky (changes requested):
Good question about `entity->tag` - it seems we have one more issue.

krizhanovsky (approved):
OK, it seems there is actually no need to check Set-Cookie. LGTM

PR #1636, opened 2022-06-06
TLS ALPN small fix and improvements

krizhanovsky:
tfw_tls_cfg_alpn_protos() always were setting only one ALPN for the global TLS configuration, even if there were many listen configuraiton options with different protocols (e.g. 'h2' and 'https'). While there is only 2 supported ALPNs the function allocated dynamic memory for them. This patch uses static size array for ALPNs and places h2 always in front of http/1.1 as more preferable.

The plan to deprecate HTTP/1.1 doesn't make sense since the protocol is widely used, so I also delete the deprecation warning.

The patch also improves ALPN debugging and readability of the SNI warning.


fw/tls_conf.c
```diff
@@ -218,9 +218,10 @@ tfw_tls_set_cert(TfwVhost *vhost, TfwCfgSpec *cs, TfwCfgEntry *ce)
 	{
 		/* None of the SANs match the vhost. */
 		T_WARN("Vhost %s doesn't have certificate with matching SAN/CN.\n"
-		       "Maybe that's fine, but it's worth checking the config -\n"
-		       "if there is no relations between the names, then host\n"
-		       "name confussion attack is possible.\n", vhost->name.data);
+		       "    Maybe that's fine, but it's worth checking the\n"
+		       "    config - if there is no relations between the\n"
+		       "    names, then host name confussion attack is possible.\n",
```
keshonok, line 223: Perhaps, this should be **confusion**?

krizhanovsky: Without this adjustments we see the warning in `dmesg` as
```
[334746.437379] [tempesta fw]   Create new server group: 'websocket'
[334746.439277] [tempesta fw]   Add new server group: 'websocket'
[334746.441520] [tempesta fw]   begin srv_group: websocket
[334746.444089] [tempesta fw]   finish srv_group: websocket
[334746.520879] [tempesta tls]   setup pk context for RSA(1) key
[334746.531999] [tempesta fw] Warning: Vhost * doesn't have certificate with matching SAN/CN.
[334746.531999] Maybe that's fine, but it's worth checking the
[334746.531999] config - if there is no relations between the
[334746.531999] names, then host name confussion attack is possible.
[334746.570507] [tempesta tls]   setup pk context for RSA(1) key
[334746.630463] [tempesta fw]   http_tbl: begin http_chain
[334746.632917] [tempesta fw]   http_tbl: finish http_chain
```
With this change the warning is more visible:
```
[334746.437379] [tempesta fw]   Create new server group: 'websocket'
[334746.439277] [tempesta fw]   Add new server group: 'websocket'
[334746.441520] [tempesta fw]   begin srv_group: websocket
[334746.444089] [tempesta fw]   finish srv_group: websocket
[334746.520879] [tempesta tls]   setup pk context for RSA(1) key
[334746.531999] [tempesta fw] Warning: Vhost * doesn't have certificate with matching SAN/CN.
[334746.531999]     Maybe that's fine, but it's worth checking the
[334746.531999]     config - if there is no relations between the
[334746.531999]     names, then host name confussion attack is possible.
[334746.570507] [tempesta tls]   setup pk context for RSA(1) key
[334746.630463] [tempesta fw]   http_tbl: begin http_chain
[334746.632917] [tempesta fw]   http_tbl: finish http_chain
```

keshonok: I am sorry that I didn't make it clear enough. The logic of the change is transparent and the change is excellent and welcome. What I meant to say was that most probably there's a typo, and **confussion attack** should read **confusion attack** - with a single **s**.

s0nx (approved):

Issue #1637, opened 2022-06-07
Crash in __http_msg_hdr_val() and __h2_msg_hdr_val()


const-t:
# Scope
Tempesta crashes while matching duplicate headers. X-Forwarded-For has been tested.

```
[   70.773313] kernel BUG at /home/const/tempesta/fw/http_msg.c:246!
[   70.773321] invalid opcode: 0000 [#1] SMP NOPTI
[   70.773324] CPU: 0 PID: 1387 Comm: nc Kdump: loaded Tainted: G           OE     5.10.35+ #5
[   70.773325] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS ArchLinux 1.16.0-1 04/01/2014
[   70.773343] RIP: 0010:__http_msg_hdr_val+0xf8/0x110 [tempesta_fw]
[   70.773345] Code: 18 00 00 00 00 c3 49 39 f8 0f 83 77 ff ff ff 0f 0b 4c 8b 01 45 0f b6 00 41 80 f8 20 74 12 41 80 f8 09 74 0c 48 89 0a c3 0f 0b <0f> 0b 0f 0b 0f 0b 4c 8b 42 08 49 29 f0 eb 93 66 0f 1f 84 00 00 00
[   70.773347] RSP: 0018:ffffb0d740003b58 EFLAGS: 00010202
[   70.773350] RAX: ffff8fb43aa60470 RBX: ffff8fb4f6a53080 RCX: 0000000000000001
[   70.773351] RDX: ffffb0d740003b60 RSI: 000000000000000d RDI: 0000000000000002
[   70.773352] RBP: 0000000000000002 R08: ffff8fb43aa60020 R09: ffffb0d740003730
[   70.773354] R10: ffffb0d740003728 R11: ffffffff9eccbaa8 R12: ffffb0d740003b60
[   70.773355] R13: 0000000000000008 R14: ffffffffc0b072a0 R15: 0000000000000000
[   70.773360] FS:  00007f855d1ac580(0000) GS:ffff8fb4fbc00000(0000) knlGS:0000000000000000
[   70.773361] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[   70.773363] CR2: 00007f855cfdc003 CR3: 000000005a86a000 CR4: 0000000000350ef0
[   70.773366] Call Trace:
[   70.773369]  <IRQ>
[...]

PR #1638, opened 2022-06-10
Fixed crash on match by HTTP Tables duplicated headers (#1637).

const-t:


const-t:
If changes will be approved I suggest to use `tfw_str_match` throughout the file.

PR #1651, opened 2022-06-28
Detached chunks mode for tests

s0nx:

### DIFF
--- a/fw/t/unit/test_http_parser_common.c
+++ b/fw/t/unit/test_http_parser_common.c
@@ -83,7 +83,8 @@ do {										\
 } while(0)
 
 static void
-tfw_free_chunks(TfwTestChunk *chunks, uint32_t chunk_cnt) {
+tfw_free_chunks(TfwTestChunk *chunks, uint32_t chunk_cnt)
+{
 	int i;
 	TEST_DBG4("%s: chunks %pK, cnt %u\n", __func__, chunks, chunk_cnt);
 	for (i = 0; i < chunk_cnt; i++)
@@ -94,7 +95,8 @@ tfw_free_chunks(TfwTestChunk *chunks, uint32_t chunk_cnt) {
 }
 
 void
-tfw_frames_chunks_free(void) {
+tfw_frames_chunks_free(void)
+{
 	FOR_EACH_FRAME({
 		TfwFrameRec *frame = GET_CURRENT_FRAME();
 		if (frame->chunks != NULL && frame->chunk_cnt != 0) {
@@ -116,7 +118,8 @@ tfw_frames_chunks_free(void) {
  * If there were some chunk buffers assigned to the frame, deallocate them here.
  */
 void
-tfw_init_frames(void) {
+tfw_init_frames(void)
+{
 	tfw_frames_chunks_free();
 	frames_cnt = 0;
 	frames_max_sz = 0;
@@ -476,8 +479,7 @@ do_split_and_parse(int type, int chunk_mode)
 			test_req_free(req);
 
 		req = test_req_alloc(frames_total_sz);
-	}
-	else if (type == FUZZ_REQ_H2) {
+	} else if (type == FUZZ_REQ_H2) {
 		/*
 		 * During the processing of a request, the HPACK dynamic table
 		 * is modified. The same query is used for each chunk size.
@@ -506,15 +508,13 @@ do_split_and_parse(int type, int chunk_mode)
 		req->pit.pool = __tfw_pool_new(0);
 		BUG_ON(!req->pit.pool);
 		__set_bit(TFW_HTTP_B_H2, req->flags);
-	}
-	else if (type == FUZZ_RESP) {
+	} else if (type == FUZZ_RESP) {
 		if (resp)
 			test_resp_free(resp);
 
 		resp = test_resp_alloc(frames_total_sz);
 		tfw_http_msg_pair(resp, sample_req);
-	}
-	else {
+	} else {
 		BUG();
 	}
 
@@ -536,12 +536,9 @@ do_split_and_parse(int type, int chunk_mode)
 
 		if (type == FUZZ_REQ_H2
 		    && frame->subtype == HTTP2_DATA
-		    && !frame->len)
-		{
+		    && !frame->len) {
 			r = TFW_POSTPONE;
-		}
-		else
-		{
+		} else {
 			r = split_and_parse_n(frame->str, type, frame->len,
 						chunk_size, &chunks);
 			if (frame->chunks != NULL) {
--- a/fw/t/unit/test_http_parser_common.c
+++ b/fw/t/unit/test_http_parser_common.c
@@ -1,0 +1,669 @@
+/**
+ *		Tempesta FW
+ *
+ * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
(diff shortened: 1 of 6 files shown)


s0nx:
Each buffer passed to the parser is allocated separately (though technically it can share the same page).
The main purpose of the patch is to catch use-after-free, out-of-bounds, etc. bugs using `KASAN` and `kmemleak` infra.

The following kernel config options are required:
```
CONFIG_KASAN=y
CONFIG_KASAN_GENERIC=y
CONFIG_KASAN_INLINE=y
CONFIG_KASAN_STACK=1
CONFIG_KASAN_VMALLOC=y
```

const-t (dismissed):
As for me, code looks well, but requires some tiny style fixes.

fw/t/unit/test_http_parser_common.h
```diff
@@ -672,48 +734,169 @@ do {										\
...
+	last_chunk_len = str_len % chunk_size;
+	if (last_chunk_len != 0) {
+		chunk_size = last_chunk_len;
+	}
+
+	chunks[chunk_cnt - 1].buf = (char *)kmalloc(chunk_size, GFP_ATOMIC);
+	if (!chunks[chunk_cnt - 1].buf) {
+		TEST_DBG("%s: Failed to allocate last chunk page(s)\n", __func__);
+		goto err_page_alloc;
+	}
```
krizhanovsky, line 788: I think we might get the same problem as before and, worse, get the tests blinking depending on how `kmalloc()` allocates memory - if (e.g. on a fresh system) it allocates all the chunks from sibling buddies, then we're in exactly the same case as before.

I propose to do something like
```C
char *tmp_buf = (char *)kmalloc(chunk_size + 16, GFP_ATOMIC);
memset(tmp_buf, 0x55, chunk_size + 16);
chunks[...].buf = tmp_buf + 8;
```

Maybe we also should consider to set non-aligned addresses for chunks - TBD.

fw/t/unit/test_http_parser_common.h
```diff
@@ -672,48 +734,169 @@ do {										\
...
+				goto err_page_alloc;
+			}
+		}
+	}
+
+	/* last chunk size may differ, so set it up separately */
+	last_chunk_len = str_len % chunk_size;
+	if (last_chunk_len != 0) {
+		chunk_size = last_chunk_len;
+	}
```
krizhanovsky, line 782:
```suggestion
	if (last_chunk_len != 0)
		chunk_size = last_chunk_len;
```
just coding style

fw/t/unit/test_http_parser_common.h
```diff
@@ -672,48 +734,169 @@ do {										\
...
+
+	chunks[chunk_cnt - 1].buf = (char *)kmalloc(chunk_size, GFP_ATOMIC);
+	if (!chunks[chunk_cnt - 1].buf) {
+		TEST_DBG("%s: Failed to allocate last chunk page(s)\n", __func__);
+		goto err_page_alloc;
+	}
+
+	return chunks;
+
+err_page_alloc:
```
krizhanovsky, line 792: "page" doesn't make sense with the new code

fw/t/unit/test_http_parser_common.h
```diff
@@ -672,48 +734,169 @@ do {										\
...
+	tfw_free_chunks(chunks, chunk_cnt);
+	return ERR_PTR(-ENOMEM);
+}
+
+#endif
+
 static int
```
krizhanovsky, line 799: also should be inline (that was just inaccurately moved from .c code)

fw/t/unit/test_http_parser_common.h
```diff
@@ -847,29 +1037,47 @@ do_split_and_parse(int type, int chunk_mode)
...
 		if (type == FUZZ_REQ_H2
-		    && frame.subtype == HTTP2_DATA
-		    && !frame.len)
+		    && frame->subtype == HTTP2_DATA
+		    && !frame->len)
 		{
 			r = TFW_POSTPONE;
 		}
 		else
 		{
```
krizhanovsky, line 1055:
```suggestion
		} else {
```

fw/t/unit/test_http_parser_common.h
```diff
@@ -847,29 +1037,47 @@ do_split_and_parse(int type, int chunk_mode)
...
+						chunk_size, &chunks);
+#ifdef CHUNK_MODE_DETACHED
+			if (frame->chunks != NULL) {
+				tfw_free_chunks(frame->chunks,
+						frame->chunk_cnt);
+				frame->chunks	 = NULL;
+				frame->chunk_cnt = 0;
+			}
+			if (chunks) {
+				frame->chunks	= chunks;
```
krizhanovsky, line 1066:
```suggestion
				frame->chunks = chunks;
```

fw/t/unit/test_http_parser_common.h
```diff
@@ -847,29 +1037,47 @@ do_split_and_parse(int type, int chunk_mode)
...
 		else
 		{
-			r = split_and_parse_n(frame.str, type, frame.len, chunk_size);
+			r = split_and_parse_n(frame->str, type, frame->len,
+						chunk_size, &chunks);
+#ifdef CHUNK_MODE_DETACHED
+			if (frame->chunks != NULL) {
+				tfw_free_chunks(frame->chunks,
+						frame->chunk_cnt);
+				frame->chunks	 = NULL;
```
krizhanovsky, line 1062:
```suggestion
				frame->chunks = NULL;
```

krizhanovsky (changes requested):
Besides the coding style adjustments, the changes look good for me except the concern about effectiveness of the chunking mode. I propose to use 8 bytes poisoning to avoid an issue with sibling buddies from kmalloc

fw/t/unit/test_http_parser_common.h
krizhanovsky: OK, I see the requirement for KASAN, so with KASAN this poisoning might be not necessary. However, IMO the tests should be correct and efficient regardless kernel options. Otherwise with each parser change, we'll need a KASAN-enabled kernel.

s0nx: I believe using KASAN infra is more reliable than just adding guard buffers to the payload.
If I understood correctly, allocating chunks from sibling buddies is not an issue to KASAN. It would still be able to track out-of-bounds accesses and so on. More than that, KASAN can produce pretty informative stack traces to find where the problem is.
We could technically always run the CI tests on a VM with KASAN enabled kernel. It would definitely slow everything down a little, but it's still worth it IMHO.

s0nx:
I've added guard buffers around the payload if `CONFIG_KASAN` is not set.
I've also moved some test functions from `test_http_parser_common.h` to `test_http_parser_common.c`.

fw/t/unit/test_http_parser_common.c
```diff
@@ -0,0 +1,693 @@
...
+		}
+
+		if (type == FUZZ_REQ_H2
+		    && frame->subtype == HTTP2_DATA
+		    && !frame->len)
+		{
+			r = TFW_POSTPONE;
+		}
+		else
+		{
```
krizhanovsky, line 541:
```suggestion
		} else {
```

krizhanovsky (approved):
Good to merge with the couple of small cleanups.

Please also investigate the reason for failed Jenkins job before the merge.

Issue #1565, opened 2022-02-10
WARN_ON is ss_skb.c when running H2 functional tests when accessing nginx via TUN/TAP network interface from qemu guest


dmpetroff:
# Scope

When running tempesta_fw inside qemu with nginx server running on host OS, h2 tests causing `WARN_ON` in ss_skb.c.

```
[16684.224107] ------------[ cut here ]------------
[16684.224401] WARNING: CPU: 0 PID: 0 at /mnt/work/tempesta/tempesta/fw/ss_skb.c:1737 ss_skb_cut_extra_data+0x1c6/0x1f0 [tempesta_fw]
[16684.224874] Modules linked in: tempesta_fw(O) tempesta_db(O) tempesta_tls(O) tempesta_lib(O) sha256_generic libsha256 sha512_generic sr_mod cdrom button psmouse rtc_cmos sg crc32c_intel i2c_piix4 i2c_core evdev configfs ip_tables x_tables ipv6 [last unloaded: tempesta_lib]
[16684.225658] CPU: 0 PID: 0 Comm: swapper/0 Tainted: G        W  O      5.10.35-tfw #12
[16684.225992] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS d55cb5a 04/01/2014
[16684.226342] RIP: 0010:ss_skb_cut_extra_data+0x1c6/0x1f0 [tempesta_fw]
[16684.226593] Code: 29 41 8b 4e 70 41 bf ff ff ff ff 41 2b 4e 74 49 8b 96 c8 00 00 00 eb 92 8b 4e 70 48 8b be c8 00 00 00 2b 4e 74 e9 a3 fe ff ff <0f> 0b 48 83 c4 28 b8 ea ff ff ff 5b 5d 41 5c 41 5d 41 5e 41 5f c3
[16684.227225] RSP: 0018:ffff88807dc05a68 EFLAGS: 00010246
[16684.227455] RAX: ffff888025823b40 RBX: ffff888025811300 RCX: ffff888000000000
[16684.227722] RDX: 0000000000000003 RSI: ffff888025823ba0 RDI: ffff888025823b90
[16684.227983] RBP: ffff888024f3a34d R08: 00001f020000044b R09: ffff88807dc05a70
[16684.228247] R10: 0000000000000000 R11: ffffea000093ce00 R12: 0000000000000000
[...]

PR #1654, opened 2022-07-08
ss_skb_cut_extra_data() fixed fragmented skb processing

const-t:


fw/ss_skb.c
```diff
@@ -1739,6 +1739,10 @@ ss_skb_cut_extra_data(struct sk_buff *skb_head, struct sk_buff *skb,
 
 			frag_num = -1;
 			skb = skb->next;
+			/* new skb also can be fragmented */
+			if (skb_is_nonlinear(skb))
+				goto frag;
```
krizhanovsky, line 1744: If we have fragments, this doesn't mean that we can and should skip the linear data, what the change actually does.

#1565 warning apparently raised because we didn't find the stop pointer and just ran out of the fragments. Why this happened? What we have missed? These questions had to be addressed in the commit message. Sine you made this fix, it seems we missed the fragments of the chained skb, but I don't see how that happened. Once we process the linear data with `__split_linear_data()` we're walking over the grags...

Also it seems your patch addresses the problem of an skb with no linear data at all. In this case we have `size == 0` and, correspondingly, `len == 0`, so `__split_linear_data()` is called with zero length. It seems the function doesn't expect such length - see for example that `__skb_fragment()` has `BUG_ON(!len)`. Maybe the fix should be in checking zero length linear data and _only then_ do `goto frag`

krizhanovsky (changes requested):
The fix seems invalid and much more detailed commit message is required.

const-t:
Maybe we need to add comment to pay attention that `ss_skb_cut_extra_data()` might change the number of fragments.It's obvious, but count changes not at every call and it can be confusing

fw/ss_skb.c
```diff
@@ -1708,12 +1709,21 @@ ss_skb_cut_extra_data(struct sk_buff *skb_head, struct sk_buff *skb,
...
+		if (frag_num >= 0) {
 			ret = __split_pgfrag_del(skb_head, skb, frag_num,
 						 offset, len, &it);
-		else
+			/*
+			 * split_pgfrag_del() can delete whole fragment and
+			 * then decrease @skb->nr_frags, therefore we need to
+			 * update @frag_num.
+			 */
+			frag_num -= fragn - skb_shinfo(skb)->nr_frags;
```
krizhanovsky, line 1720: It seems the name `frag_num` is confusing since it's actually fragment ID, the fragment where we need to start to search. Suppose we have skb with 5 fragments and frag_num = 2. If `__split_pgfrag_del()` deletes the fragment, then fragment 3 is moved to position 2 and 4 to 3. Thus, we continue to work with the same frament ID, which points to a new fragment.

keshonok: Yes, I think along similar lines. Also, `fragn` is one more name you need to remember reading this code. I would just name it `nr_frags` to assert the origin and the relevance.

const-t: As for me rename `fragn` to `nr_frags` is good idea. And I suggest to rename `frag_num` to `frag_index`, but for consistency with other part of code, we should use name `i`.

krizhanovsky (approved):
LGTM, but probably we still need the warning

fw/ss_skb.c
```diff
@@ -1704,41 +1703,43 @@ ss_skb_cut_extra_data(struct sk_buff *skb_head, struct sk_buff *skb,
...
 			return ret;
 
 		if (tail)
 			return 0;
 
+		/*
+		 * Use SKB which iterated in __split_pgfrag_del_w_frag()
+		 * and __split_linear_data().
+		 */
+		skb = it.skb;
```
krizhanovsky, line 1728:
```suggestion
                if (skb != it.skb) {
                        if (WARN_ON_ONCE(skb_head == it.skb))
                                return -EINVAL;
                        skb = it.skb; 
                }
```

fw/ss_skb.c
```diff
@@ -1704,41 +1703,47 @@ ss_skb_cut_extra_data(struct sk_buff *skb_head, struct sk_buff *skb,
...
 		if (unlikely(ret))
 			return ret;
 
 		if (tail)
 			return 0;
 
+		/*
+		 * Use SKB which iterated in __split_pgfrag_del_w_frag()
+		 * and __split_linear_data().
+		 */
```
s0nx, line 1727:
```suggestion
		/* 
		 * While searching for a place to put the remaining data
		 * there might be the case when we would either use existing 
		 * @sk_buff->next of the current SKB or completely new SKB,
		 * which would be allocated and inserted right after the current one.
		 * When this happens, we need to reacquire the correct @skb.
		 * See __split_pgfrag_del_w_frag() / __split_linear_data() for details.
		 */
```

s0nx:
LGTM

s0nx (approved):

PR #1656, opened 2022-07-11
Fix #1650: configuration defaults

### LINKED ISSUE #1650: "Unable to find a backend server" in the default server group
# Motivation

https://github.com/tempesta-tech/tempesta/wiki/Backend-servers makes example for default server group. Meantime the test https://github.com/tempesta-tech/tempesta-test/pull/230#discussion_r896941645 exhibits that the logic is broken.

# Scope

```bash
root@ubuntu:~/tempesta# git log |head -1
commit cda8ef83b803bf0ac5d93ec87947d45595c7a98c
root@ubuntu:~/tempesta# cat etc/tempesta_fw.conf 
server 192.168.100.4:8000 conns_n=4;
```

If Tempesta FW is compiled with `DEBUG=1`, then dmesg shows `Unable to find a backend server` on a browser request. 

`Unable to find a backend server` is printed for `tfw_http_get_srv_conn()` returning `NULL`. We don't have a session, so this seems means that `tfw_vhost_get_srv_conn()` returns `NULL`. On `DEBUG=3` I didn't see `vhost: use server group` message from the last function, so the problem is in empty `main_sg` server group. IIUC for the above config `tfw_vhost_cfgend()` should have set the default server group.

# Testing

Test modified in https://github.com/tempesta-tech/tempesta-test/pull/230 must work without the config modifications

# Documentation

Need to carefully analyze what's going on with the configuration process and request handling, probably we should adjust our documentation instead of changing the code. Anyway, let's discuss what's going on before changing the code.

krizhanovsky:
Please see the commit message for the details.

I did test the patch with just `server 192.168.100.4:8000;` in `tempesta_fw.conf`. Also I have tested
```

server 127.0.0.1:8000;

tls_match_any_server_name;
tls_certificate /root/tempesta/etc/cert_with_san.crt;
tls_certificate_key /root/tempesta/etc/the.key;
```
to make sure that SAN certificates still work and they do work with the default server group, vhost and no http_chain rules.

### DIFF
--- a/fw/tls.c
+++ b/fw/tls.c
@@ -801,17 +801,17 @@ static int
 fw_tls_apply_sni_wildcard(BasicStr *name)
 {
 	int n;
-	char *p = strchr(name->data, '.');
+	char *p = strnchr(name->data, name->len, '.');
 	if (!p)
 		return -ENOENT;
 	n = name->data + name->len - p;
 
 	/* The resulting name must be lower than a top level domain. */
-	if (n < 3 || !strchr(p + 1, '.'))
+	if (n < 3 || !strnchr(p + 1, n, '.'))
 		return -ENOENT;
 
 	/*
-	 * Store leading dot to match against chpped wildcard (see
+	 * Store leading dot to match against chopped wildcard (see
 	 * tfw_tls_add_cn()) and do not confuse the name with a CN.
 	 */
 	name->data = p;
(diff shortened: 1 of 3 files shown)


const-t (approved):
LGTM. Also I have tested with configurations, that I had problems before. Works fine.

pale-emperor (approved):
I run few testcases - works fine
Now tempesta configs looks right and work properly

nickzaev (approved):
LGTM

s0nx (approved):

PR #1662, opened 2022-07-13
Fix hang when many headers(large) received from backend

const-t:


krizhanovsky (approved):
Good to merge with the small test fix

s0nx (approved):

PR #1665, opened 2022-07-26
Update Jenkinsfile tests timeout

pale-emperor:


pale-emperor:
After install 5.10.35+ with kmemleak enabled - test runs out of 30M threshold
So i change timeout to 45M - now works fine

krizhanovsky (approved):
LGTM

Issue #1661, opened 2022-07-13
Multiple failures in unit tests for the HTTP/1 parser


krizhanovsky:
Build the current kernel with exactly config 
[linux-config.txt](https://github.com/tempesta-tech/tempesta/files/9102395/linux-config.txt)
and run Tempesta FW unit tests. Note that I use Ubuntu 22, so binutils and GCC might be different.
```
[  333.098576] tfw_test: /root/tempesta/fw/t/unit/test_http1_parser.c:218: error:
[  333.101716] tfw_test: FAIL:
[  333.104433] tfw_test:   test__http1_parser__parses_req_uri()
[  333.107460] tfw_test:   EXPECT_TRUE((tfw_str_eq_cstr(&req->host, "natsys-lab.com", strlen("natsys-lab.com"), 0))) => 0x0
[  333.111469] tfw_test: h1 req: [OPTIONS * HTTP/1.1^M
[  333.111469] ^M
[  333.111469] ]
[  333.120062] tfw_test: h1 req: [GET sch://userame@natsys-lab.com HTTP/1.1^M
[  333.120062] ^M
[  333.120062] ]
[  333.128773] tfw_test: h1 req: [GET ^? HTTP/1.1^M
[  333.128773] Host: test^M
[  333.128773] ^M
[  333.128773] ]
[  333.139320] tfw_test: h1 req: [GET /^Curi HTTP/1.1^M
[  333.139320] Host: test^M
[  333.139320] ^M
[  333.139320] ]
[  333.149128] tfw_test: TEST_RUN(http1_parser, mangled_messages)
[  333.151772] tfw_test: h1 req: [GET / HTTP/1.1^M
[  333.151772] POST / HTTP/1.1^M
[  333.151772] Host: test^M
[  333.151772] ^M
[  333.151772] ]
[  333.162852] tfw_test: h1 req: [GET / HTTP/1.1^M
[  333.162852] Host: test^M
[  333.162852] ^_X-Foo: test^M
[  333.162852] ^M
[  333.162852] ]
[  333.173212] tfw_test: h1 req: [GET / HTTP/1.1^M
[  333.173212] Host: test^M
[  333.173212] Connection: close, "foo"^M
[  333.173212] ^M
[  333.173212] ]
[  333.182871] tfw_test: === response: [HTTP/1.0 200 OK^M
[  333.182871] Content-Type: foo/aa-^Ynp^M
[  333.182871] ^M
[  333.182871] ]
[...]

PR #1666, opened 2022-07-28
fw/t/unit: Unit tests fixes

s0nx:
* Fixed the issue, when only 1 byte chunk was actually used for testing
* Fixed memory leakage
* Fixed early chunks deallocation for http1 tests causing multiple failures

### DIFF
--- a/fw/t/unit/test_http1_parser.c
+++ b/fw/t/unit/test_http1_parser.c
@@ -3125,7 +3125,6 @@ resp:
 		}
 	}
 end:
-	tfw_frames_chunks_free();
 	kernel_fpu_end();
 	vfree(str);
 	kernel_fpu_begin();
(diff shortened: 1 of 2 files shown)


s0nx:
Fixes #1661

const-t (approved):
Looks good.

fw/t/unit/test_http1_parser.c
```diff
@@ -3125,6 +3125,7 @@ TEST(http1_parser, fuzzer)
 		}
 	}
 end:
+	tfw_frames_chunks_free();
```
krizhanovsky, line 3128: I didn't get why do we need to free the chunks here? It seems all the chunks users, like `FOR_REQ()`, just rely on the fact that any following test frees the chunks in `tfw_init_frames()`, then why this doesn't work for this test case? Shouldn't we move the chunks freeing to the end of `TEST_SUITE()`?

krizhanovsky (approved):
Looks good and I confirm that the problem is gone. Just have a question about memory freeing.

fw/t/unit/test_http1_parser.c
s0nx: Yep, seems excessive.

PR #1672, opened 2022-08-09
Fix bug in response cache processing

s0nx:


fw/http_msg.c
```diff
@@ -652,17 +652,16 @@ __tfw_http_msg_add_str_data(TfwHttpMsg *hm, TfwStr *str, void *data,
 
 	if (TFW_STR_EMPTY(str)) {
 		if (!str->data)
-			__tfw_http_msg_set_str_data(str, data, skb);
+			__tfw_str_set_data(str, data, skb);
```
keshonok, line 655: Before this goes further, let me bring up this thought.
The change seems to break the naming schema to a certain degree. `str` functions are just that - they work on strings in their `TFW` implementation. These functions are intentionally written in the way that avoids direct dealing with `skb`. Now, `skb` appear whenever there's a HTTP message, and skb carries data, hence the appropriate naming for functions that deal with strings that are set on data in skb. Granted, the current naming may be too heavy and excessive (meaning - too long names), but they reflect the idea correctly.
New naming breaks this convention. What's more, it brings `skb` to `str.c` where all functions are _pure_ so far.
So, perhaps, there's a better way? Perhaps, `tfw_strcpy_prep()` doesn't belong there?
Hopefully I am not way off with this rant. Thanks.

s0nx: Though i do kind of agree with you, i must admit that `tfw_strcpy_prep()` still belongs here IMO. Yes, it sets `skb` explicitly, but other than that it's not that different from the other `tfw_str*` functions. It deals with chunks and other `TfwStr` members just like the others. `skb` is a legit member of `TfwStr` and no additional `sk_buff` manipulations have been introduced except for explicit assignment.
@krizhanovsky, @const-t what do you guys think?

const-t: `tfw_strcpy_prep` in my opinion not a good name, because it can be used not only for preparing structure for `tfw_strcpy`, but it suitable to reside in `str.c`. `tfw_strcpy` almost copy of `tfw_strdup` except copy and `__tfw_str_set_data` call. `__tfw_str_set_data` can be replaced with explicit assign of skb or something like this `*dc = *data_str;`. Maybme I am wrong, but `__tfw_str_set_data` really can misled, name indicates we can set data only with skb.

fw/http.c
```diff
@@ -4313,7 +4318,6 @@ tfw_http_hdr_split(TfwStr *hdr, TfwStr *name_out, TfwStr *val_out, bool inplace)
 
 		if (!name_found) {
 			if (chunk->data[0] == ':') {
-				WARN_ON_ONCE(chunk->len != 1);
```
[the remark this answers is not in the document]
s0nx: When Tempesta appends additional headers, which were not present in the original response, the length of the "middle" chunk (": ") is 2 most of the time. It does contain a WS.

fw/str.c
```diff
@@ -993,6 +993,50 @@ tfw_strdup_desc(TfwPool *pool, const TfwStr *src)
...
+ */
+TfwStr *
+tfw_strcpy_prep(TfwPool *pool, const TfwStr *data_str, const TfwStr *src)
+{
+	size_t sz;
+	char *data;
+	const TfwStr *sc, *end;
+	TfwStr *dst, *dc;
+
+	BUG_ON(!TFW_STR_PLAIN(data_str));
```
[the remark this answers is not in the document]
s0nx: What do you mean exactly? `data_str` is expected to be plain here. AFAIK `TfwStr` returned from the `ss_skb_get_room()` would always be plain.

fw/cache.c
```diff
@@ -2313,10 +2313,6 @@ tfw_cache_build_resp(TfwHttpReq *req, TfwCacheEntry *ce, long lifetime,
 		if (tfw_cache_build_resp_body(db, trec, it, p, ce->body_len,
 					      TFW_MSG_H2(req), stream_id))
 			goto free;
-		if (!TFW_MSG_H2(req)
-		    && test_bit(TFW_HTTP_B_CHUNKED, resp->flags)
-		    && tfw_http_msg_expand_data(it, skb_head, &g_crlf, NULL))
-			goto free;
```
krizhanovsky, line 2314: I created testing issue for this https://github.com/tempesta-tech/tempesta-test/issues/289 - could you please review it and add any test cases, which I could miss?

krizhanovsky (changes requested):
I believe we should implement an ad-hoc string copying to not to slow down significantly headers addition.

fw/cache.c
s0nx: It looks like all potential cases are pretty much covered in there.

krizhanovsky (approved):
LGTM

s0nx:
I've enhanced CE debug prints a little bit as part of this PR. Cache entry dump looks like this now:
```
[178929.266850] [tempesta fw]   Tdb CE [000000003536bd5e]:                                                                                                     
                           TdbVRec: key b602c3125fb0b161, chunk_next 0, len 372                                                                                
                           key_len: 10                                                                                                                         
                        status_len: 1                                                                                                                          
                           rph_len: 3                                                                                                                          
                           hdr_num: 9                                                                                                                          
                        hdr_h2_off: 8                                                                                                                          
                           hdr_len: 142                                                                                                                        
                          body_len: 0                                                                                                                          
                            method: 3                                                                                                                          
                             flags: 0                                                                                                                          
                               age: 0                                                                                                                          
                              date: 1661769328                                                                                                                 
                          req_time: 1661769327                                                                                                                 
                         resp_time: 1661769328                                                                                                                 
                     last_modified: 0                                                                                                                          
                           key off: 50f8                                                                                                                       
                        status off: 5102                                                                                                                       
                          hdrs off: 5106                                                                                                                       
                          body off: 5194                                                                                                                       
                           hmflags: 6000 [13-14]                                                                                                               
                           version: 3                                                                                                                          
                       resp_status: 200                                                                                                                        
[178929.298990] [tempesta fw]   Cache: service request [00000000012f0cae] w/ key=b602c3125fb0b161, ce=000000003536bd5e    
```

krizhanovsky (approved):
Looks good. I'd only rename `__tfw_dump_ce()` to something like `__tfw_dbg_dump_ce()` to say explicitly that this is about debugging

Issue #1677, opened 2022-08-18
OOM on h2load


krizhanovsky:
# Scope

Tempesta FW on the current master as the date of the issue as well as on https://github.com/tempesta-tech/tempesta/pull/1643 causes OOM on `h2load -t 2 -c 1000 -D 60 https://ubuntu/` in a VM with 4GB RAM:
```
[38759.502337] Mem-Info:
[38759.504398] active_anon:228 inactive_anon:3 isolated_anon:0
[38759.504398]  active_file:46 inactive_file:39 isolated_file:0
[38759.504398]  unevictable:0 dirty:0 writeback:0
[38759.504398]  slab_reclaimable:3508 slab_unreclaimable:30742
[38759.504398]  mapped:8 shmem:12 pagetables:597 bounce:0
[38759.504398]  free:10508 free_pcp:48 free_cma:0
[38759.518162] Node 0 active_anon:912kB inactive_anon:12kB active_file:184kB inactive_file:156kB unevictable:0kB isolated(anon):0kB isolated(file):0kB mapped:32kB dirty:0kB writeback:0kB shmem:48kB shmem_thp: 0kB shmem_pmdmapped: 0kB anon_thp: 0kB writeback_tmp:0kB kernel_stack:4288kB all_unreclaimable? no
[38759.526902] Node 0 DMA free:12828kB min:328kB low:408kB high:488kB reserved_highatomic:2048KB active_anon:0kB inactive_anon:0kB active_file:0kB inactive_file:0kB unevictable:0kB writepending:0kB present:15992kB managed:15908kB mlocked:0kB pagetables:0kB bounce:0kB free_pcp:0kB local_pcp:0kB free_cma:0kB
[38759.537401] lowmem_reserve[]: 0 2747 3180 3180
[38759.540160] Node 0 DMA32 free:25044kB min:58096kB low:72620kB high:87144kB reserved_highatomic:2048KB active_anon:380kB inactive_anon:0kB active_file:32kB inactive_file:0kB unevictable:0kB writepending:0kB present:3129204kB managed:3022552kB mlocked:0kB pagetables:112kB bounce:0kB free_pcp:0kB local_pcp:0kB free_cma:0kB
[38759.550732] lowmem_reserve[]: 0 0 433 433
[...]

PR #1687, opened 2022-08-30
Fix memory leaks across the code

s0nx:


s0nx:
At the moment `kmemleak` still reports that several objects have leaked (http chains configuration):
```
unreferenced object 0xffff888123797400 (size 512):
  comm "sysctl", pid 71793, jiffies 4303451488 (age 6533.964s)
  hex dump (first 32 bytes):
    00 c5 20 3a 81 88 ff ff 34 00 00 00 00 00 00 00  .. :....4.......
    00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
  backtrace:
    [<000000000b8de27b>] tfw_cfgop_http_rule+0x95d/0x1990 [tempesta_fw]
    [<00000000ffef506a>] spec_handle_entry+0xe8/0x170 [tempesta_fw]
    [<000000008cc20cae>] tfw_cfg_handle_children+0x2a3/0x440 [tempesta_fw]
    [<00000000ffef506a>] spec_handle_entry+0xe8/0x170 [tempesta_fw]
    [<0000000012d943b2>] tfw_cfg_parse_mods+0x2c1/0x3d0 [tempesta_fw]
    [<00000000a3aaf719>] tfw_cfg_parse+0x78/0xb0 [tempesta_fw]
    [<000000005e3599a3>] tfw_ctlfn_state_io+0x1bd/0x5d0 [tempesta_fw]
    [<00000000fcb08e9e>] proc_sys_call_handler+0x293/0x420
    [<00000000cb31df24>] new_sync_write+0x364/0x5e0
    [<00000000446deaaa>] vfs_write+0x4d5/0x710
    [<00000000c8ee35ff>] ksys_write+0xdd/0x1a0
    [<00000000f9139a8e>] do_syscall_64+0x33/0x80
    [<00000000bb27a0eb>] entry_SYSCALL_64_after_hwframe+0x44/0xa9
unreferenced object 0xffff88813a20c500 (size 64):

```
I've rechecked http chains configuration code and seems like these are false-positives. `spec_cleanup()` and the actual cleanup handler (`tfw_cfgop_rules_cleanup()`) deallocates all the objects correctly. More than that if one would unload `tempesta_fw` before the `kmemleak` scan, there won't be any leaks reported.

const-t (approved):
LGTM

fw/http_frame.c
```diff
@@ -641,10 +646,20 @@ tfw_h2_conn_streams_cleanup(TfwH2Ctx *ctx)
...
 		tfw_h2_stream_unlink(ctx, cur);
-		tfw_h2_stream_clean(ctx, cur);
+
+		/* The streams tree is about to be destroyed and
+		 * we don't want to trigger rebalancing.
+		 * No further actions regarding streams dependencies/prio
+		 * is required at this stage.
+		 */
+		tfw_h2_delete_stream(cur);
+		--ctx->streams_num;
```
krizhanovsky, line 660: I'm wondering why do we traverse the tree and delete all the streams one by one instead of using TfwPool for the connection? `TfwStream` looks not so small, so probably we can easily utilize a page for a connection. Are they any other allocations with the connection life time?

Can we avoid the loop at all and just do `ctx->streams_num = 0`?

krizhanovsky (approved):
The fixes look good. But please collect `perf` data for `h2load` from https://github.com/tempesta-tech/tempesta/issues/1677 and let's have a look onto the streams allocation and freeing routines. It'd be good to use at least 8 CPU VM for the tests, more is better.

fw/http_frame.c
s0nx: I believe it was just an initial design decision to use streams `kmem_cache` for all connections. I would create a separate task to evaluate performance.

PR #1693, opened 2022-09-17
fw/cache: Fix a bug processing `cache_resp_hdr_del` tokens

s0nx:
Split duplicate headers while filtering out specific headers which are meant not to reside in cache entry
(provided by `cache_resp_hdr_del` cfg directive)

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -1732,18 +1732,20 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 
 static bool
 check_cfg_ignored_header(const TfwStr *field, TfwCaToken *tokens,
-		     unsigned int tokens_sz)
+			 unsigned int tokens_sz)
 {
 	int i;
 	int bytes_count = 0;
 	TfwCaToken *token = tokens;
+	const TfwStr *hdr = TFW_STR_DUP(field) ? TFW_STR_CHUNK(field, 0) : field;
+
 	for (i = 0; i < tokens_sz; i++) {
 		const TfwStr to_del = {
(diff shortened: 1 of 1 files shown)


fw/cache.c
```diff
@@ -1732,18 +1732,20 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
...
 static bool
 check_cfg_ignored_header(const TfwStr *field, TfwCaToken *tokens,
-		     unsigned int tokens_sz)
+			 unsigned int tokens_sz)
 {
 	int i;
 	int bytes_count = 0;
 	TfwCaToken *token = tokens;
+	const TfwStr *hdr = (field->flags & TFW_STR_DUPLICATE) ?
+			    TFW_STR_CHUNK(field, 0) : field;
```
const-t, line 1741: Add please empty line after variables definition block

const-t (approved):
LGTM. But need small changes.

fw/cache.c
krizhanovsky, line 1741:
```suggestion
        const TfwStr *hdr = TFW_STR_DUP(field) ? TFW_STR_CHUNK(field, 0) : field;              

```

krizhanovsky (approved):
LGTM with the small cleanup

krizhanovsky:
@byko3y good point! @s0nx I'm wondering how many more such places do we have?

s0nx:
That's might be the problem indeed. I've created an issue to review the code carefully.

byko3y:
I looked through the code, but it seems that all other instances of tfw_stricmpspn invocation are already checking for TFW_STR_DUP.

PR #1694, opened 2022-09-18
Handle 0-bits left shift.

krizhanovsky:
ttls_mpi_div_mpi() may call the function with the zero count value if X and Y are of the same size, which doesn't look exceptional.

Fixes https://github.com/tempesta-tech/tempesta-test/issues/302


b3b:
Related issue: #1683

const-t (approved):
LGTM

PR #1701, opened 2022-09-23
Use separate object file copies for unit tests

i-rinat:
Kernel build scripts may rewrite object files in place, and so during parallel builds there is a possibility of having concurrent writers changing files `str_avx2.o`, `http_parser.o`, `http_sess.o`, `str.o`, `ss_skb.o`, `msg.o`, and `http_msg.o`. This race condition may cause builds to fail with cryptic messages.

The solution is to add symlinks to make kernel build script treat affected files as separate sources.


fw/t/unit/Makefile
```diff
@@ -29,15 +29,15 @@ obj-m += tfw_test.o
...
-	../../str.o		\
-	../../ss_skb.o		\
-	../../msg.o		\
-	../../http_msg.o
+	http_parser.o	\
+	http_sess.o	\
+	str.o		\
+	ss_skb.o	\
+	msg.o		\
+	http_msg.o
```
krizhanovsky, line 43: The fix looks quite subtle, so I think at least a comment in the Makefile is required or we'll be confused by the files with just includes.

krizhanovsky (approved):
LGTM

fw/t/unit/Makefile
i-rinat: Added a description into this Makefile.
During the process I realized that there is no need to add anything to `tfw_test-y`, as `.c` symlinks are already being handled by `$(wildcard $(obj)/*.c)` some lines above. So that unnecessary part was removed.

Issue #1647, opened 2022-06-23
Kernel crash on ws.test_ws_ping.CacheTest.test_ping_websockets


pale-emperor:
# Steps to reproduce

Run tests couple of times:

```
for i in `seq 40`; do ./run_tests.py ws.test_ws_ping.CacheTest.test_ping_websockets; done
```

Result:

Kernel panic and reboot

pale-emperor:
DMESG

[   73.074562] Start test: ws.test_ws_ping.CacheTest.test_ping_websockets
[   73.136746] [tdb] Start Tempesta DB
[   73.153563] [tempesta fw] Initializing Tempesta FW kernel module...
[   73.221991] [tempesta fw] Configuration processing is completed.
[   73.242323] [tdb] Opened table /opt/tempesta/db/filter0.tdb: size=16777216 rec_size=20 base=000000007a36f7b5
[   73.309753] [tdb] Opened table /opt/tempesta/db/cache0.tdb: size=268435456 rec_size=0 base=000000005ebb1814
[   73.314848] [tdb] Opened table /opt/tempesta/db/sessions0.tdb: size=16777216 rec_size=312 base=000000003b67df5e
[   73.319858] [tdb] Opened table /opt/tempesta/db/client0.tdb: size=16777216 rec_size=624 base=000000008efba8a1
[   73.425914] [tempesta fw] Open listen socket on: 0.0.0.0:82
[   73.426520] [tempesta fw] Open listen socket on: 0.0.0.0:81
[   73.553889] [tempesta fw] Tempesta FW is ready
[   73.582015] [tempesta fw] 127.0.0.1 "localhost" "GET / HTTP/1.1" 101 0 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/99.0.4844.51 Safari/537.36"
[   75.538030] net_ratelimit: 24 callbacks suppressed
[   75.538032] [tempesta fw] Warning: sock_srv: cannot establish connection for 127.0.0.1:8099: 6 tries, keep trying...
[   75.538938] [tempesta fw] Warning: sock_srv: cannot establish connection for 127.0.0.1:8099: 6 tries, keep trying...
[   75.539365] [tempesta fw] Warning: sock_srv: cannot establish connection for 127.0.0.1:8099: 6 tries, keep trying...
[...]

PR #1709, opened 2022-10-04
Kernel crash on ws.test_ws_ping.CacheTest.test_ping_websockets

avbelov23:
https://github.com/tempesta-tech/tempesta/issues/1647

~Increment the reference count of connections in case of an attempt to destroy them.~
~`tfw_ws_conn_drop()` and `tfw_ws_msg_process()` can run in parallel.~

~We do not nullify paired connections - this can lead to null dereference.~

~We get a client connection for the websocket upgrade so that it is not
destroyed by the original drop callback before the upgrade is completed.~

Eliminate the race between client, server tfw_ws_conn_drop() and
tfw_ws_msg_process()

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -6265,7 +6265,7 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 	TfwHttpMsg *hmresp, *hmsib;
 	TfwCliConn *cli_conn;
 	TfwFsmData data_up;
-	bool conn_stop, filtout = false, websocket = false;
+	bool conn_stop, filtout = false, websocket = false, was_upgrade = false;
 
 	BUG_ON(!stream->msg);
 	/*
@@ -6455,21 +6455,30 @@ next_msg:
 		goto next_resp;
 	}
 
+	/* Do upgrade if correct websocket upgrade response detected earlier */
+	if (websocket && !test_bit(TFW_HTTP_B_REQ_DROP, hmresp->req->flags)) {
+		tfw_connection_get((TfwConn *)(cli_conn));
+
+		r = tfw_http_websocket_upgrade((TfwSrvConn *)conn, cli_conn);
+		if (unlikely(r < TFW_PASS)) {
+			tfw_connection_put((TfwConn *)(cli_conn));
+			return TFW_BLOCK;
+		}
+		was_upgrade = true;
+
+		tfw_connection_put((TfwConn *)(cli_conn));
+	}
+
 	/*
 	 * Pass the response to cache for further processing.
 	 * In the end, the response is sent on to the client.
 	 * @hmsib is not attached to the connection yet.
 	 */
 	r = tfw_http_resp_cache(hmresp);
+	if (was_upgrade)
+		conn->destructor(conn);
 	if (unlikely(r < TFW_PASS))
 		return TFW_BLOCK;
-
-	/* Do upgrade if correct websocket upgrade response detected earlier */
-	if (websocket) {
-		r = tfw_http_websocket_upgrade((TfwSrvConn *)conn, cli_conn);
-		if (unlikely(r < TFW_PASS))
-			return TFW_BLOCK;
-	}
 
 next_resp:
 	if (skb && websocket) {
--- a/fw/http.c
+++ b/fw/http.c
@@ -1967,7 +1967,8 @@ tfw_http_conn_treatnip(TfwSrvConn *srv_conn, struct list_head *eq)
 	TfwServer *srv = (TfwServer *)srv_conn->peer;
 	TfwHttpReq *req_sent = (TfwHttpReq *)srv_conn->msg_sent;
 
-	if (tfw_http_conn_on_hold(srv_conn)
+	if (req_sent
+	    && tfw_http_conn_on_hold(srv_conn)
 	    && !(srv->sg->flags & TFW_SRV_RETRY_NIP))
 	{
 		BUG_ON(list_empty(&req_sent->nip_list));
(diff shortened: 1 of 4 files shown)


fw/websocket.c
```diff
@@ -163,6 +163,11 @@ tfw_ws_msg_process(TfwConn *conn, struct sk_buff *skb)
 
 	T_DBG2("%s: conn=[%p], skb=[%p]\n", __func__, conn, skb);
 
+	if (!tfw_srv_conn_get_if_live(conn))
+		return TFW_BLOCK;
+	if (!tfw_srv_conn_get_if_live(conn->pair))
+		return TFW_BLOCK;
```
krizhanovsky, line 169: I don't think that this is reliable. `tfw_ws_conn_drop()` calls `tfw_connection_put()`, which can just free `conn`, so if we have a race between the function and `conn` isn't protected, then we can just try to increment the refcaunter in a dead memory area.

The main question is who calls `tfw_ws_conn_drop()` - the server shutdown process (process context) or a socket error event (softirq with locked socket)? I guess the second one. The first one is easy to protect from - there just should not be active underlying sockets, when it frees the connections.

The second question is who exactly freed due to the race - `conn` or `conn->pair`? I think `conn->pair` since we're should be under socket lock for `conn`. If I'm right, then we should use the same mechanism as for forwarding responses to client connections - client connections can also die from under us.

fw/websocket.c
```diff
@@ -227,19 +234,37 @@ tfw_ws_conn_unpair(TfwConn *conn)
...
+		 */
+		if (atomic_read(&conn->refcnt) == 2) {
+			tfw_conn_hook_call(TFW_CONN_HTTP_TYPE(conn), conn, conn_drop);
+			tfw_connection_put(conn);
+		}
+		tfw_connection_put(conn);
+	}
 
 	if (!pair)
 		return;
```
const-t, line 263: We must `spin_unlock` before `return`.

fw/websocket.c
```diff
@@ -227,19 +234,37 @@ tfw_ws_conn_unpair(TfwConn *conn)
...
+		 * If original drop callback wasn't called then we call connection drop
+		 * callback and compensate for the getting of the connection for the
+		 * upgrade. If original drop callback was called then we put connection
+		 * in him.
+		 */
+		if (atomic_read(&conn->refcnt) == 2) {
+			tfw_conn_hook_call(TFW_CONN_HTTP_TYPE(conn), conn, conn_drop);
+			tfw_connection_put(conn);
+		}
+		tfw_connection_put(conn);
```
const-t, line 259: We can simplify this part by ridding of first `tfw_connection_put(conn);`

fw/websocket.c
```diff
@@ -227,19 +234,37 @@ tfw_ws_conn_unpair(TfwConn *conn)
...
-		tfw_conn_hook_call(TFW_CONN_HTTP_TYPE(conn), conn, conn_drop);
+	}
+	else {
+		/*
+		 * If original drop callback wasn't called then we call connection drop
+		 * callback and compensate for the getting of the connection for the
+		 * upgrade. If original drop callback was called then we put connection
+		 * in him.
+		 */
+		if (atomic_read(&conn->refcnt) == 2) {
```
const-t, line 255: This condition can be evaded by sending 3 pipelined upgrade requests.

s0nx (commented):
Nothing else from me, except for the issues mentioned by @const-t.

fw/sock_clnt.c
```diff
@@ -88,10 +88,15 @@ tfw_cli_conn_alloc(int type)
 		return NULL;
 
 	tfw_connection_init((TfwConn *)cli_conn);
+	cli_conn->ws_lock = kzalloc(sizeof(tfw_ws_spinlock_t), GFP_ATOMIC);
```
s0nx, line 91: We need to check for allocation failure here.

keshonok: Yeah. Still, this whole lock/refcnt dancing business looks fishy at first glance. I am not convinced it does reliably what it's intended to do. I didn't analyze it deeply though so there's always a chance I am wrong.

avbelov23: The main problem is that we operate with a paired connection in the drop of another and a race can occur.

avbelov23: Hi, @keshonok!
It seems to have been reasonable. look diagonally please

fw/http.c
```diff
@@ -5906,6 +5907,21 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
...
+		     && test_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET, req->flags)))
+	{
+		/*
+		 * We take an additional refcnt one to prevent access to the
+		 * freed connection when racing between client and server
+		 * tfw_ws_conn_drop()
+		 */
+		if (!conn->get_refcnt) {
+			tfw_connection_get(conn);
+			conn->get_refcnt = true;
```
const-t, line 5921: My comment not about current section. However, pipelining of "upgrade" requests are leads to ws connection "object" leaks. I believe we should reject all upgrade requests after first successful upgrade. You can to reproduce it using this [test](https://github.com/tempesta-tech/tempesta-test/blob/kt-1647-poc/ws/test_ws_ping.py#L473) from my branch.

const-t:
I found two issues that not relates to this PR, but should be considered to fix. Reproduce it using: `test_ws_ping.WS2.test_ws_resched` and `test_ws_ping.WS2.test_ws_stop` from https://github.com/tempesta-tech/tempesta-test/blob/kt-1647-poc/ branch.

const-t (changes requested):

fw/http.c
avbelov23: What does it mean (reject)? Send a specific response?

krizhanovsky:
@avbelov23 `tfw_http_conn_msg_alloc()` takes a reference on `TfwHttpMsg->conn` when it allocates a new HTTP message on the connection, so `TfwConn` has as many references as messages were received on it.

krizhanovsky:
I close this in respect to https://github.com/tempesta-tech/tempesta/pull/1786 . https://github.com/tempesta-tech/tempesta/pull/1786/commits/9b4a03a11d165df7d9ce999d12e3642a58e503b8 is ported from this PR though

Issue #1697, opened 2022-09-20
Requests with single-character header names should be accepted


b3b:
# Scope

Currently, requests containing single character header names are rejected with a 400 code.
Request example:
  ```http
GET / HTTP/1.1
X: test
```
Corresponding parser warning: `Warning: Parser error: state=RGen_HdrOtherN input(-8)=0x3a('P/1.1`

Related issues: #1412

# Testing
Test to reproduce: [http_general.test_headers.TestSmallHeader](https://github.com/tempesta-tech/tempesta-test/blob/4bcfd6e2993214de14bc57fc318f2f651a1989da/http_general/test_headers.py#L220)

PR #1721, opened 2022-10-13
#1697 Fix header field name length check

avbelov23:
https://github.com/tempesta-tech/tempesta/issues/1697

~In checking the length of the header field name, we take into account the processed unfixed characters (__FSM_MOVE) and fixed, and not just parsed at the current iteration of the state~
Add a check that we haven't passed a single character since the beginning
of the header

Zero length header field name for http/2 is forbidden in tfw_hpack_decode.

### DIFF
--- a/fw/http_parser.c
+++ b/fw/http_parser.c
@@ -1254,8 +1254,8 @@ __FSM_STATE(RGen_HdrOtherN) {						\
 		 * ':' can be found at the beginning of a new skb or	\
 		 * fragment, it's ok.					\
 		*/							\
-		if (unlikely(!__fsm_sz					\
-			     && TFW_STR_EMPTY(&msg->stream->parser.hdr))) \
+		if (unlikely(!__fsm_sz && TFW_STR_EMPTY(&parser->hdr)	\
+			     && p == (unsigned char *)parser->hdr.data))\
 		{							\
 			TFW_PARSER_BLOCK(RGen_HdrOtherN);		\
 		}							\
--- a/fw/t/unit/test_http1_parser.c
+++ b/fw/t/unit/test_http1_parser.c
@@ -93,6 +93,13 @@ TEST(http1_parser, leading_eol)
 	FOR_RESP("\n\n\n" EMPTY_RESP);
 }
 
+TEST(http1_parser, short_name)
+{
+	FOR_REQ_SIMPLE("X: test");
+	FOR_REQ_SIMPLE("Z: test");
+	EXPECT_BLOCK_REQ_SIMPLE(": test");
+}
+
 TEST(http1_parser, parses_req_method)
 {
 #define TEST_REQ_METHOD(METHOD)					\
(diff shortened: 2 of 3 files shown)


fw/http_parser.c
```diff
@@ -1249,12 +1249,18 @@ __FSM_STATE(st_curr) {							\
 __FSM_STATE(RGen_HdrOtherN) {						\
 	__FSM_MATCH_MOVE(token, RGen_HdrOtherN, 0);			\
 	if (likely(*(p + __fsm_sz) == ':')) {				\
+		size_t n = 0;						\
+		if (TFW_STR_EMPTY(&parser->hdr)) {			\
+			n = p - (unsigned char *)parser->hdr.data;	\
+		} else {						\
+			n = parser->hdr.len + __data_off(p);		\
+		}							\
```
krizhanovsky, line 1257: We need `n` only to not to get into the `if` statement at the below, so the `else` branch is senseless and this `if` condition can be aggregated with the below one. Moreover, we don't need `n` value, just need to for 0 or not, i.e. this means "whether we saw at least one character in the header?". The last thing is that we really can use `parser` and don't need `&msg->stream->parser.hdr`. I propose to rewrite the fix as
```C
--- a/fw/http_parser.c
+++ b/fw/http_parser.c
@@ -1252,10 +1252,11 @@ __FSM_STATE(RGen_HdrOtherN) {                                           \
                /*                                                      \
                 * Header name must contain at least one character, but \
                 * ':' can be found at the beginning of a new skb or    \
-                * fragment, it's ok.                                   \
+                * fragment, it's ok. We also might get into the state  \
+                * either on the fist character of the header or not.   \
                */                                                      \
-               if (unlikely(!__fsm_sz                                  \
-                            && TFW_STR_EMPTY(&msg->stream->parser.hdr))) \
+               if (unlikely(!__fsm_sz && TFW_STR_EMPTY(parser->hdr)    \
+                            && p != (unsigned char *)parser->hdr.data))\
                {                                                       \
                        TFW_PARSER_BLOCK(RGen_HdrOtherN);               \
                }                                                 
```

krizhanovsky (changes requested):
Need to rework the fix, add fix for HTTP/2 and tests for both the HTTP/1 and HTTP/2

fw/http_parser.c
```diff
@@ -1254,8 +1254,8 @@ __FSM_STATE(RGen_HdrOtherN) {						\
 		 * ':' can be found at the beginning of a new skb or	\
 		 * fragment, it's ok.					\
 		*/							\
-		if (unlikely(!__fsm_sz					\
-			     && TFW_STR_EMPTY(&msg->stream->parser.hdr))) \
+		if (unlikely(!__fsm_sz && TFW_STR_EMPTY(&parser->hdr)	\
+                            && p == (unsigned char *)parser->hdr.data))\
```
krizhanovsky, line 1258: Seems misaligned `&&` and trailing slash

krizhanovsky (approved):
LGTM, just please make the small cleanup

s0nx (approved):
LGTM

PR #1727, opened 2022-10-19
fw/cache: Split `cache-control: no-cache/private` duplicate headers

s0nx:
while filtering out

headers which are meant not to reside in cache entry


krizhanovsky (approved):
LGTM

PR #1733, opened 2022-10-26
Increase cache_misses counter after using PURGE

avbelov23:
Increase cache.misses counter if cache entry isn't live

https://github.com/tempesta-tech/tempesta-test/pull/346


RomanBelozerov (approved):
LGTM. It works for https://github.com/tempesta-tech/tempesta-test/pull/346

PR #1734, opened 2022-10-26
fw/cache: Do not take an additional reference on paged sk_buff fragment while copying http/2 response body

s0nx:

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -2251,7 +2251,8 @@ tfw_cache_add_body_page(TfwMsgIter *it, char *p, int sz, TfwFrameHdr *frame_hdr,
 
 	++it->frag;
 	skb_fill_page_desc(it->skb, it->frag, page, off, sz);
-	skb_frag_ref(it->skb, it->frag);
+	if (!h2)
+		skb_frag_ref(it->skb, it->frag);
 	ss_skb_adjust_data_len(it->skb, sz);
 
 	return 0;
--- a/fw/http.c
+++ b/fw/http.c
@@ -4807,7 +4807,6 @@ tfw_h2_append_predefined_body(TfwHttpResp *resp, unsigned int stream_id,
 		++it->frag;
 		skb_fill_page_desc(it->skb, it->frag, page, 0,
 				   copy + FRAME_HEADER_SIZE);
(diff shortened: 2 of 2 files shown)


fw/cache.c
```diff
@@ -2081,7 +2081,6 @@ tfw_cache_add_body_page(TfwMsgIter *it, char *p, int sz, TfwFrameHdr *frame_hdr,
 
 	++it->frag;
 	skb_fill_page_desc(it->skb, it->frag, page, off, sz);
-	skb_frag_ref(it->skb, it->frag);
```
krizhanovsky, line 2084: It seems we don't need this reference counter because we just allocated the page with `alloc_page()` and the page already has refcounter =1. However, this code works also for non HTTP/2 branch, were we just get the page from `p` and it seems we need the refcounter. Or need a good explanation what was wrong with HTTP/1 code as well.

A good commit message explaining the problem and how it is fixed and why the fix is correct, would help.

krizhanovsky (changes requested):
It seems there is an issue with the fix. @const-t review is definitely required, since the skb refcounting is quite a subtle topic.

fw/cache.c
const-t: @krizhanovsky @s0nx We can't remove `skb_frag_ref(it->skb, it->frag);` for h1. Because fragment uses same page as TBD, unrefing the fragment will cause TDB memory corruption.

s0nx:
Taking additional reference on paged sk_buff fragment is not needed when building `http/2` response body because each fragment is represented by newly allocated page with `page->_refcount == 1`. Extra page reference would result in memory leak during `sk_buff` freeing because of non-zero `page->_refcount`, so final `__put_page()` would not be called.

For `http/1` responses we still need an additional page reference due to the fact that the page in question has actually been allocated previously by TDB (see tdb_file_open()). Missing an additional reference here would lead to TDB owned pages freeing by kfree_skb(), which in turn would cause memory corruption.

`tfw_h2_append_predefined_body()` is obviously called for `http/2` responses only, so an additional reference has been deleted.

krizhanovsky (approved):
LGTM

const-t (approved):
LGTM

PR #1746, opened 2022-11-08
JSCH usability and documentation

krizhanovsky:
There was a user request that Tempesta FW with the config like (my testing config)
```
listen 192.168.100.4:80;

block_action attack reply;
block_action error reply;

srv_group default {
        server 192.168.100.4:8000;
}

#vhost default {
vhost ubuntu {
        tls_certificate /root/tempesta/etc/tfw-root.crt;
        tls_certificate_key /root/tempesta/etc/tfw-root.key;

        proxy_pass default;

        sticky {
                cookie enforce options="Max-Age=900";
                js_challenge delay_min=1000 delay_range=1000;
        }
}

cache 1;
cache_fulfill * *;

access_log on;

http_chain {
        -> ubuntu;
}
```
sends 503 error codes regardless any cookies and with normal browser.

Firstly, I was surprised seeing 503 codes in redirects, but it does make sense (see the commit message and the code comment). Next, I saw a Tempesta FW warning, which didn't help me to debug the problem.

This PR adds a comment and makes reporting more administrator-friendly. I also updated https://github.com/tempesta-tech/tempesta/wiki/Sticky-Cookie#configuration with better description of all the `delay` parameters and troubleshooting.

P.S. The original problem was due to too low default value for the `delay_limit` and was fixed by adding ` delay_limit=10000` to `js_challenge`.

### DIFF
--- a/fw/http_sess.c
+++ b/fw/http_sess.c
@@ -555,7 +555,7 @@ end_##f:								\
 				char buf[sizeof(hmac) * 2];		\
 				bin2hex(buf, hmac, sizeof(hmac));	\
 				sess_warn("bad received HMAC value",	\
-					  addr, ": %c(pos=%d),"		\
+					  addr, " %c(pos=%d),"		\
 					  " ts=%#lx orig_hmac=[%.*s]\n", \
 					  *tr, i, ts,			\
 					  (int)sizeof(hmac) * 2, buf);	\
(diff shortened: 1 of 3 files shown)


fw/http_sess.c
```diff
@@ -605,8 +605,10 @@ tfw_http_redir_mark_verify(TfwHttpReq *req, TfwStr *mark_val, RedirMarkVal *mv)
 	HEX_STR_TO_BIN_GET(mv, att_no);
 	HEX_STR_TO_BIN_GET(mv, ts);
 
-	if (__redir_hmac_calc(req, mv))
+	if (__redir_hmac_calc(req, mv)) {
+		T_DBG("can not compute HAMC for a redirection mark\n");
```
s0nx, line 609:
```suggestion
		T_DBG("can not compute HMAC for a redirection mark\n");
```

fw/http_sess_conf.c
```diff
@@ -29,6 +29,14 @@ static TfwVhost *cur_vhost;
 
 #define STICKY_NAME_DEFAULT	"__tfw"
 
+/*
+ * JavaScript challenge requires a browser to execut a code from the response
```
s0nx, line 33:
```suggestion
 * JavaScript challenge requires a browser to execute a code from the response
```

s0nx (approved):

PR #1747, opened 2022-11-08
Fix ClientHello: bad extension size

const-t:
-Added "list" of supported extension. They must be parsed during
 processing ClientHello. Other extensions will be omitted.
 -Unsupported extensions not copies to handshake's temporary
 buffer now, since we must not to parse them, therefore they might
 have large size.

### DIFF
--- a/tls/tls_srv.c
+++ b/tls/tls_srv.c
@@ -244,7 +244,7 @@ ttls_parse_extended_ms_ext(TlsCtx *tls, const unsigned char *buf, size_t len)
 }
 
 static void
-ttls_parse_session_ticket_ext(TlsCtx *tls, unsigned char *buf, size_t len)
+ttls_parse_session_ticket_ext(TlsCtx *tls, const unsigned char *buf, size_t len)
 {
 	T_DBG("ClientHello: ticket length: %lu\n", len);
 	/* Remember the client asked us to send a new ticket */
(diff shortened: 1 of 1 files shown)


tls/tls_srv.c
```diff
@@ -912,7 +989,13 @@ ttls_parse_client_hello(TlsCtx *tls, unsigned char *buf, size_t len,
...
 		io->hslen -= 2;
 		n = tls->hs->ext_sz;
-		if (n + 4 > tls->hs->ext_rem_sz || n > sizeof(tls->hs->ext)) {
+                /*
+                 * Do not account size of @ext for unsupportred extensions.
+                 * They must not be copied to it.
+                 */
+		if (n + 4 > tls->hs->ext_rem_sz
+                    || (ttls_ext_is_supported(tls->hs->ext_type)
+                        && n > sizeof(tls->hs->ext))) {
```
krizhanovsky, line 998: The Linux kernel uses tabs for alignment with not more than 7 spaces at the end for alignment. We also usually place `{` on a new line after multi-line conditions

tls/tls_srv.c
```diff
@@ -945,63 +1031,20 @@ ttls_parse_client_hello(TlsCtx *tls, unsigned char *buf, size_t len,
...
-			tmp = tls->hs->ticket_ctx.ticket;
-		memcpy_fast(tmp + io->rlen, p, n);
+                if (ext_supported) {
+			if (unlikely(ext_type == TTLS_TLS_EXT_SESSION_TICKET))
+                              tmp = tls->hs->ticket_ctx.ticket;
+		        memcpy_fast(tmp + io->rlen, p, n);
+                }
 		p += n;
-		if (unlikely(io->rlen + n < ext_sz))
+                if (unlikely(io->rlen + n < ext_sz))
```
krizhanovsky, line 1040: Tab replaced with spaces

krizhanovsky (approved):
Looks good to merge with the couple cleanups

PR #1765, opened 2022-12-08
Changed condition to met RFC requirement

const-t:
The "max-age" request directive indicates that the client is
  unwilling to accept a response whose age is greater than the
  specified number of seconds. RFC 7234.


krizhanovsky (approved):
LGTM

PR #1772, opened 2022-12-22
Remove pfl optimization

const-t:


fw/ss_skb.c
```diff
@@ -379,24 +288,16 @@ __new_pgfrag(struct sk_buff *skb_head, struct sk_buff *skb, int size,
 	     int i, int shift)
 {
 	int off = 0;
```
krizhanovsky, line 290: We don't need the initialization any more

krizhanovsky (approved):
LGTM with the small cleanups

PR #1787, opened 2023-01-16
Fix invalid parsing of multiple Set-Cookie headers

### LINKED ISSUE #1690: Multiple Set-Cookie headers break the HTTP parser
# Scope
Backend responses with multiple `Set-Cookie` headers are dropped.
Multiple `Set-Cookie` headers support required to run WordPress backend https://github.com/tempesta-tech/tempesta-test/issues/290

Example error output:
```
[267822.051861] [tempesta fw] Warning: Parser error: state=Resp_HdrSet_CookieV input(-8)=0x77('Cookie: wordpress_86a9106ae65') data_len=330 off=309
```

# Testing
Test to reproduce: [http_general/test_headers.py](https://github.com/tempesta-tech/tempesta-test/blob/16199c604168ae5f0d349e6442ecf264d4c363c6/http_general/test_headers.py)

EvgeniiMekhanik:
According to RFS 7230 3.2.2 all duplicates of special singular headers must be blocked, except Set-Cookie headers (See appropriate note).
Previously Set-Cookie headers was also blocked. This patch fixes this behaviour and appropriate unit test is implemented.

Closes #1690


fw/http_msg.c
```diff
@@ -561,6 +561,10 @@ tfw_http_msg_hdr_close(TfwHttpMsg *hm)
 		 * RFC 7230 3.2.2: all duplicates of special singular
 		 * headers must be blocked as early as possible,
 		 * just when parser reads them.
+		 * Since it cannot be combined into a single field-value,
```
krizhanovsky, line 564: Please update the copyright year in this and all the changed source files

fw/http_msg.c
```diff
@@ -570,9 +574,8 @@ tfw_http_msg_hdr_close(TfwHttpMsg *hm)
 		 * For H2 literal headers, @parser->_hdr_tag would be set
 		 * during actual HTTP parsing.
 		 */
-		if (id < TFW_HTTP_HDR_NONSINGULAR) {
-			if (!TFW_MSG_H2(hm) || id != TFW_HTTP_HDR_COOKIE)
```
krizhanovsky, line 574: Why the H2 case is removed? Need review from @s0nx as it was his code. @s0nx BTW I didn't get why this code is about trailer headers as stated in the comment - it seems the code is executed for all header, not trailers only.

krizhanovsky (commented):
The change is probably good and I appreciate the unit test, but I didn't get what's going on with the H2 statement, so need review from @s0nx . If @const-t and @s0nx approve it, then I good with the change as well

fw/http_msg.c
EvgeniiMekhanik: Previously we check that if message is not http2 we should return TFW_BLOCK. But if you look at RFS 7230 3.2.2 Note you see that we should not drop message if it has duplicate cookie header for http1 also.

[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed

s0nx: @krizhanovsky Yes, it's executed for all headers. Duplicate h2 pseudo-headers would be filtered out here, hence the comment.
@const-t raised concerns about this piece of code previously: https://github.com/tempesta-tech/tempesta/pull/1705#discussion_r1015460760

const-t (approved):
LGTM

s0nx (approved):

Issue #1748, opened 2022-11-09
Corrupted location header on H2 sticky cookie redirect


b3b:
# Motivation

When Sticky cookie is enabled, HTTP/2 redirect responses has
* wrong schema, HTTP/2 protocol is not taken into account in https://github.com/tempesta-tech/tempesta/blob/546b03dd36b0640c81ef9e4e5316f2738fee2749/fw/http.c#L717 
* wrong location with non-printable characters

Related issues: #1396
Test to reproduce: `sessions.test_h2_sticky_scheduler` - https://github.com/tempesta-tech/tempesta-test/pull/282

# Scope

## Tempesta config
```nginx
listen 127.0.0.1:80 proto=http;
listen 127.0.0.1:8443 proto=https;
listen 127.0.0.1:443 proto=h2;

srv_group default {
  server 127.0.0.1:8000;
}

vhost default-host {
  proxy_pass default;
  sticky {
    cookie enforce;
  } 
}

http_chain {
  -> default-host;
}

tls_match_any_server_name;
tls_certificate /tmp/tempesta/tempesta.crt;
tls_certificate_key /tmp/tempesta/tempesta.key;
```

##  HTTP: correct response

```bash
curl -v http://127.0.0.1:8443/
...
< HTTP/1.1 302 Found
< date: Wed, 09 Nov 2022 11:56:45 GMT
< content-length: 0
< location: http://127.0.0.1/
< server: Tempesta FW/pre-0.7.0
< set-cookie: __tfw=000000010025e86f38f5048ab071d5d672714c02a30fd0894f85ae6b; path=/
```
## HTTPS: correct response
```bash
curl -kv https://127.0.0.1:8443/
...
< HTTP/1.1 302 Found
< date: Wed, 09 Nov 2022 11:58:37 GMT
< content-length: 0
< location: https://127.0.0.1:8443/
< server: Tempesta FW/pre-0.7.0
< set-cookie: __tfw=0000000100265577493ba1b338b519231032fcd3b55a5c44e476df35; path=/
```

## HTTP/2: corrupted responses

``` bash
curl -kv https://127.0.0.1/
...
> GET / HTTP/2
> Host: 127.0.0.1
> user-agent: curl/7.85.0-DEV
> accept: */*
> 
* Connection state changed (MAX_CONCURRENT_STREAMS == 4294967295)!
< HTTP/2 302 
< date: Wed, 09 Nov 2022 10:42:38 GMT
< content-length: 0
[...]

PR #1790, opened 2023-01-19
Fix corrupted header location on H2 sticky cookie redirect

EvgeniiMekhanik:
There was two problems with header `location` on H2 sticky cookie redirect:
- wrong schema, because we need to check that the TFW_FSM_HTTPS bit is set in the connection protocol type field, not that the connection protocol type is equal to TFW_FSM_HTTPS.
- corrupted location, because we should use PLAIN TfwStrs in url chunks, but host and req->uri_path are not PLAIN TfwStrs and should be converted. This patch fixed both this problems.

Closes: #1748 #1396


EvgeniiMekhanik:
Associated PR https://github.com/tempesta-tech/tempesta-test/pull/385

const-t (changes requested):
There is potential bug in line 743.
```
tfw_http_msg_clnthdr_val(req, &req->h_tbl->tbl[TFW_HTTP_HDR_HOST], TFW_HTTP_HDR_HOST, &host);
```
Response can be sent for http1 and http2, but `TFW_HTTP_HDR_HOST` may absence in http2 request. I believe we should use `:authority` in such cases.

fw/http.c
```diff
@@ -714,9 +714,11 @@ tfw_http_prep_redir(TfwHttpResp *resp, unsigned short status, TfwStr *rmark,
 	char *date_val = *this_cpu_ptr(&g_buf);
 	char *cl_val = *this_cpu_ptr(&g_buf) + SLEN(S_V_DATE);
 	char *body_val = NULL;
-	TfwStr *proto = &protos[TFW_CONN_PROTO(req->conn) == TFW_FSM_HTTPS];
+	char *host_val = NULL;
+	char *uri_path_val = NULL;
+	TfwStr *proto = &protos[TFW_CONN_PROTO(req->conn) & TFW_FSM_HTTPS];
```
krizhanovsky, line 719: This requires that `TFW_FSM_HTTPS` is 1 and if we change it, then we get a memory access violation here.

```suggestion
	TfwStr *proto = &protos[!!(TFW_CONN_PROTO(req->conn) & TFW_FSM_HTTPS)];
```

Also `protos` should be `const`

fw/http.c
```diff
@@ -747,27 +747,35 @@ tfw_http_prep_redir(TfwHttpResp *resp, unsigned short status, TfwStr *rmark,
 	if (host.len) {
 		url.chunks[url.nchunks++] = *proto;
 		url.len += proto->len;
-
-		url.chunks[url.nchunks++] = host;
-		url.len += host.len;
+		host_val = *this_cpu_ptr(&g_buf) + SLEN(S_V_DATE) + cl_len;
+		host_len = tfw_str_to_cstr(&host, host_val, RESP_BUF_LEN - SLEN(S_V_DATE) - cl_len);
```
krizhanovsky, line 751: Too wide line. We keep our lines not wider than 80, maybe a few characters more.

fw/http.c
```diff
@@ -747,27 +747,35 @@ tfw_http_prep_redir(TfwHttpResp *resp, unsigned short status, TfwStr *rmark,
...
 
 	if (rmark->len) {
 		url.chunks[url.nchunks++] = *rmark;
 		url.len += rmark->len;
 	}
 
-	url.chunks[url.nchunks++] = req->uri_path;
-	url.len += req->uri_path.len;
+	uri_path_val = *this_cpu_ptr(&g_buf) + SLEN(S_V_DATE) + cl_len + host_len;
+	uri_path_len = tfw_str_to_cstr(&req->uri_path, uri_path_val, RESP_BUF_LEN - SLEN(S_V_DATE) - cl_len - host_len);
```
krizhanovsky, line 764: too wide lines, please reference https://github.com/tempesta-tech/tempesta/blob/master/CodingStyle

fw/http.c
```diff
@@ -747,27 +747,35 @@ tfw_http_prep_redir(TfwHttpResp *resp, unsigned short status, TfwStr *rmark,
...
+	url.len += uri_path_len;
+	url.nchunks++;
 
 	/*
 	 * We have to copy the body since tfw_h2_append_predefined_body() called
 	 * from tfw_h2_frame_local_resp() expects body as plain string.
 	 * At the moment this function is used for sticky session redirects, so
 	 * there is no big difference wehre to copy the body.
 	 */
+	body_val = *this_cpu_ptr(&g_buf) + SLEN(S_V_DATE) + cl_len + host_len + uri_path_len;
```
krizhanovsky, line 776: Why not to get a pointer and move it forward after each write to `g_buf` instead of repetition of the shift calculation?

Also in the origin code we had only date value and the body size, which are definitely smaller than 2KB, but this is doubtful with the Host value, so in this patch we have to track whether we fit into `g_buf`.

krizhanovsky (changes requested):
Some cleanups are needed, but more importantly a more accurate buffer handling is required.

fw/http.c
```diff
@@ -714,9 +714,11 @@ tfw_http_prep_redir(TfwHttpResp *resp, unsigned short status, TfwStr *rmark,
 	char *date_val = *this_cpu_ptr(&g_buf);
```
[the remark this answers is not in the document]
EvgeniiMekhanik: fixed

fw/http.c
```diff
@@ -714,9 +714,11 @@ tfw_http_prep_redir(TfwHttpResp *resp, unsigned short status, TfwStr *rmark,
...
 	char *cl_val = *this_cpu_ptr(&g_buf) + SLEN(S_V_DATE);
 	char *body_val = NULL;
-	TfwStr *proto = &protos[TFW_CONN_PROTO(req->conn) == TFW_FSM_HTTPS];
+	char *host_val = NULL;
+	char *uri_path_val = NULL;
```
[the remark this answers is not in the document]
EvgeniiMekhanik: fixed

fw/http.c
EvgeniiMekhanik: fixed

fw/http.c
EvgeniiMekhanik: fixed

fw/http.c
EvgeniiMekhanik: fixed

nmmmnu:
I am sorry for this comment, but this is really awful code and should be
splitted to several lines to become more readable.

EvgeniiMekhanik:
> There is potential bug in line 743.
> 
> ```
> tfw_http_msg_clnthdr_val(req, &req->h_tbl->tbl[TFW_HTTP_HDR_HOST], TFW_HTTP_HDR_HOST, &host);
> ```
> 
> Response can be sent for http1 and http2, but `TFW_HTTP_HDR_HOST` may absence in http2 request. I believe we should use `:authority` in such cases.

Yes, we should first check :authority and if it is empty check TFW_HTTP_HDR_HOST

Issue #1490, opened 2021-01-26
All Frang checks must be finished before request can be processed


ikoveshnikov:
## Scope

Frang is called by gfsm hooks right after a new request (response) portion is parsed. A request is blocked if Frang returns `TFW_BLOCK` decision, decisions `TFW_PASS` and `TFW_POSTPONE` allow further processing. To be sure that a request has passed all the Frang checks before it's forwarded to backend server, Frang hooks need to follow this contract:

- [X] A request is validated against the set of Frang checks exactly once. I.e. there is no situations when Frang fsm has reached `DONE` state before request is fully parsed, and a new request portion invokes a new Frang scan from the very first state. Current Frang state follows this contract condition, but in one of previous implementations we could spin through the Frang fsm several times. 

- [ ] Frang never returns `TFW_PASS` decision if a request is not fully parsed yet.
- [ ] FRang fsm has reached `DONE` state if `TFW_PASS` decision is returned on `TFW_HTTP_FSM_REQ_MSG` hook. 

If the last two contract terms are not fulfilled, situations like #1477  are very likely. In that issue an http/1.1-specific condition was checked before moving to the next Frang state, but on h2 requests the condition was always false. So Frang fsm never reached `DONE`  state and a half of frang checks were skipped.

Filtering requests that didn't pass through the full set of frang checks is acceptable and desired behaviour.  

## Testing

This section is mandatory. Please, specify name of the test revealing the problem or at least generic considerations how to make a test for our QA. A link to a new issue in https://github.com/tempesta-tech/tempesta-test/issues would be just perfect.

PR #1798, opened 2023-01-25
Fix response absence in case of invalid http2 request 

EvgeniiMekhanik:


EvgeniiMekhanik:
Patch to tempesta linux kernel sources
https://github.com/tempesta-tech/linux-5.10.35-tfw/pull/6

EvgeniiMekhanik:
Previous one https://github.com/tempesta-tech/tempesta/pull/1790

fw/http.c
```diff
@@ -5857,7 +5857,7 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 		T_DBG2("Block invalid HTTP request\n");
 		TFW_INC_STAT_BH(clnt.msgs_parserr);
 		tfw_http_req_parse_drop(req, 400, "failed to parse request");
-		return TFW_BLOCK;
```
const-t, line 5860: @krizhanovsky  @EvgeniiMekhanik  Parsing errors should not be considered as security event?

krizhanovsky: We filter a lot of attack on the parser level (e.g. injection attacks). However, at the moment we don't distinguish innocent parser errors from  clear attacks and just return `TFW_BLOCK` everywhere. I'd prefer to reply by default with something telling to a user, than just brutally reset connection in many cases. So let's treat parser errors as just errors, not attacks for now.

It seems we need to fix the header comment for `tfw_http_req_parse_drop()` since it states "Caller must return TFW_BLOCK...", which isn't consistent with the current code.

fw/tls.c
```diff
@@ -170,7 +170,7 @@ tfw_tls_connection_recv(TfwConn *conn, struct sk_buff *skb)
 
 		/* Do upcall to http or websocket */
 		r = tfw_connection_recv(conn, data_up.skb);
-		if (r == TFW_BLOCK) {
+		if (r == TFW_BLOCK || r == TFW_BAD) {
 			kfree_skb(nskb);
```
krizhanovsky, line 174: I'm wondering if there are other similar places. It'd be good to run the testsuite with some memory leakage detector, like kmemleak. However, it doesn't track buddy allocator while we use it for skbs. IIRC @s0nx recently shared a link to a kernel debugging tool, which catches page leaks, but I can't find a link...

It'd be good to add a chapter about memory leak detection to https://github.com/tempesta-tech/tempesta/wiki/Debugging-and-troubleshooting with the link and probably link to kmemleak

fw/sock.c
```diff
@@ -938,28 +940,32 @@ ss_tcp_data_ready(struct sock *sk)
...
-		if (!(SS_CONN_TYPE(sk) & Conn_Stop))
-			ss_close(sk, SS_F_SYNC);
-		break;
 	case SS_DROP:
-		ss_close(sk, SS_F_ABORT);
+		flags = SS_F_ABORT;
+		break;
+	case SS_BAD:
+		fallthrough;
+	default:
```
krizhanovsky, line 948: Which `default` values we can observe here? If this is just a safety, then it make sense to place `WARN_ON_ONCE()` here. The `switch` implements a very subtle logic, so a default path make the things less clear.

fw/sock.c
```diff
@@ -938,28 +940,32 @@ ss_tcp_data_ready(struct sock *sk)
...
-		 *
-		 * Closing a socket should go through the queue and
-		 * should be done after all pending data has been sent.
-		 */
-		if (!(SS_CONN_TYPE(sk) & Conn_Stop))
-			ss_close(sk, SS_F_SYNC);
-		break;
 	case SS_DROP:
-		ss_close(sk, SS_F_ABORT);
+		flags = SS_F_ABORT;
```
krizhanovsky, line 944: Please add the comment from the commit message that we do not abort a TCP connection if the higher layer decides that there is an attack, but we have to send an HTTP response (i.e. `Conn_Stop` flag is set).

krizhanovsky (approved):
I don't see any severe issues with the PR, but couple of a cleanups and memory leak testing are very wished.

const-t (approved):
Good to merge after fix mentioned corrections

fw/http.c
EvgeniiMekhanik: Fixed

fw/sock.c
EvgeniiMekhanik: Fixed

fw/sock.c
EvgeniiMekhanik: Fixed

fw/http.c
```diff
@@ -5956,8 +5961,11 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 		skb = NULL;
 	}
 
-	if ((r = tfw_http_req_client_link(conn, req)))
+	if ((r = tfw_http_req_client_link(conn, req))) {
+		tfw_http_req_parse_block(req, 403, "request dropped: "
+					 "incorrect X-Forwarded-For header");
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes, you right

fw/http.c
s0nx: I believe, we should also check for `T_BAD` now inside `tfw_h2_frame_process()`. 
For example here:
```
(fw/http_frame.c:1880) 
		h2->data_off = 0;
		h2->skb_head = data_up.skb->next = data_up.skb->prev = NULL;
		r = tfw_http_msg_process_generic(c, h2->cur_stream, data_up.skb);
		if (r == T_DROP) {
			kfree_skb(nskb);
			goto out;
		}
	} else {
```

fw/tls.c
s0nx: I used this tool to track page allocations: [page_owner](https://www.kernel.org/doc/html/v5.0/vm/page_owner.html)

s0nx (changes requested):

fw/gfsm.h
```diff
@@ -139,6 +139,12 @@ enum {
 	 */
 	TFW_BLOCK	= SS_DROP,
 
+        /*
+         * Same as TFW_BLOCK, but we send FIN not RST in this case
+         * to the client.
+         */
+        TFW_BAD         = SS_BAD,
```
s0nx, line 146: For some reason, it still show spaces here, not tabs.

fw/http.c
EvgeniiMekhanik: Yes, and we also should check for all `r` values not only T_BAD, because `tfw_http_msg_process_generic` can return also system errors.

fw/http_frame.c
```diff
@@ -1878,9 +1878,11 @@ tfw_h2_frame_process(TfwConn *c, struct sk_buff *skb)
 		h2->data_off = 0;
 		h2->skb_head = data_up.skb->next = data_up.skb->prev = NULL;
 		r = tfw_http_msg_process_generic(c, h2->cur_stream, data_up.skb);
-		if (r == T_DROP) {
+		if (r && r != T_POSTPONE) {
 			kfree_skb(nskb);
 			goto out;
+		} else if (r == T_POSTPONE) {
```
EvgeniiMekhanik, line 1884: Please pay attention to this code !!!

fw/tls.c
krizhanovsky: @EvgeniiMekhanik did you have a chance to test the code for memory leaks? @pale-emperor @RomanBelozerov do we have a memory leakage and/or stress tests for HTTP/2?

krizhanovsky (approved):
LGTM

fw/tls.c
EvgeniiMekhanik: I try to use page_owner, but i can't find some useful information in it's logs. I will ask about it @s0nx on our meeting

fw/http_frame.c
const-t: Determining of postponing occurs in `tfw_h2_frame_recv`. `tfw_http_msg_process_generic` just process current frame and can't return `T_POSTPONED`. In `tfw_http_msg_process_generic`  I can't see any codepath that returns `T_POSTPONED`. Even if `tfw_http_msg_process_generic` could return `T_POSTPONED` it would be senseless, because we already determined postponing above. Checking of `T_POSTPONE` here doesn't have a sense.

const-t (approved):
LGTM

fw/http_frame.c
krizhanovsky: Can't this path `tfw_http_msg_process_generic()` -> `tfw_http_req_process()` -> `ss_skb_process()` -> `tfw_http_parse_req()` lead to `T_POSTPONED`  ?

fw/http_frame.c
```diff
@@ -1878,7 +1878,9 @@ tfw_h2_frame_process(TfwConn *c, struct sk_buff *skb)
 		h2->data_off = 0;
 		h2->skb_head = data_up.skb->next = data_up.skb->prev = NULL;
 		r = tfw_http_msg_process_generic(c, h2->cur_stream, data_up.skb);
-		if (r == T_DROP) {
+		/* TODO Check this place, when working on #1490 */
```
krizhanovsky, line 1881: Not a big deal, but we usually we write such comments like `/* TODO #1490: Check this place, when working on */`

pale-emperor:
@krizhanovsky not sure about http2, i suppose  - we havent
Also for now kmemleak tests disabled for run acceleration

s0nx (approved):

PR #1799, opened 2023-01-25
Fix corrupted header location on H2 sticky cookie redirect

EvgeniiMekhanik:
There were two problems with header `location` on H2 sticky cookie redirect:
- wrong schema, because we need to check that the TFW_FSM_HTTPS bit is set in the connection protocol type field, not that the connection protocol type is equal to TFW_FSM_HTTPS.
- corrupted location, because we should use PLAIN TfwStrs in url chunks, but host and req->uri_path are not PLAIN TfwStrs and should be converted. This patch fixed both this problems.

Closes: #1748 #1396

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -66,7 +66,7 @@
  * created HTTP/1.1-message.
  *
  * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
- * Copyright (C) 2015-2022 Tempesta Technologies, Inc.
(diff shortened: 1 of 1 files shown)


EvgeniiMekhanik:
Previous one https://github.com/tempesta-tech/tempesta/pull/1790

krizhanovsky (approved):
LGTM

const-t (approved):
LGTM

s0nx (approved):

Issue #1807, opened 2023-02-02
SETTINGS_HEADER_TABLE_SIZE > 4096 bytes does not work as expected


RomanBelozerov:
# Scope

If client sends settings frame with SETTINGS_HEADER_TABLE_SIZE > 4096, we must:

- or does not send settings frame with `ack` flag;
- or change settings with new value for table size and encode header > 4096 bytes in dynamic table (now Tempesta does not encode headers > 4096 bytes).

# Testing
Need test for one of cases

krizhanovsky:
It seems we can expand the table, but we need to consider security to not get DDoSed with too many and too large such requests. If we do expand the table, then we need to have some limit for this, maybe a configuration option.

Also need to check whether popular browsers, like Chorme and Firefox, request the extensions.

krizhanovsky:
Closed by https://github.com/tempesta-tech/tempesta/pull/1815

PR #1815, opened 2023-02-09
HTTP2: report SETTINGS_HEADER_TABLE_SIZE in initial SETTINGS frame.

dmpetroff:
This mimics nginx behaviour who also has fixed size dynamic table buffer of 4096 bytes and reports it in initial SETTINGS message.

See #1807 for more context.

### DIFF
--- a/fw/http_frame.c
+++ b/fw/http_frame.c
@@ -394,16 +394,26 @@ tfw_h2_send_wnd_update(TfwH2Ctx *ctx, unsigned int id, unsigned int wnd_incr)
 static inline int
 tfw_h2_send_settings_init(TfwH2Ctx *ctx)
 {
-	unsigned char key_buf[SETTINGS_KEY_SIZE];
-	unsigned char val_buf[SETTINGS_VAL_SIZE];
+	/* XX(id, val) where
+	 * `id` is one of HTTP2_SETTINGS_xxx constants
+	 * `val` is of the size uint32_t
+	 */
+#define INITIAL_HEADERS(XX)                                     \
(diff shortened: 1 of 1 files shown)


EvgeniiMekhanik (approved):

fw/http_frame.c
```diff
@@ -394,30 +394,50 @@ tfw_h2_send_wnd_update(TfwH2Ctx *ctx, unsigned int id, unsigned int wnd_incr)
...
+	 * `val` is of the size uint32_t
+	 */
+#define INITIAL_HEADERS(XX)                                     \
+	XX(HTTP2_SETTINGS_TABLE_SIZE, HPACK_ENC_TABLE_MAX_SIZE) \
+	XX(HTTP2_SETTINGS_INIT_WND_SIZE, ctx->lsettings.wnd_sz)
+
+#define PLUS_ONE(...) + 1
+	/* Automatically adjusted payload buffer */
+	unsigned char payload[(SETTINGS_KEY_SIZE + SETTINGS_VAL_SIZE)
+		* (0 INITIAL_HEADERS(PLUS_ONE))];
```
krizhanovsky, line 408: The macro usage is fancy, but makes the code harder to read and doesn't reduce copy & paste since the most macros are used only once and only one of them twice.

krizhanovsky (changes requested):
Please simplify the code, like it was before. Tempesta FW's code is already sophisticated and there is no sense to introduce just fancy C.

krizhanovsky (approved):
LGTM

PR #1820, opened 2023-02-15
Reworked HTTP1 -> HTTP2 message transformation

const-t:
This patch introduce another approach to HTTP1 -> HTTP2 message transformation. We had following problems with inplace transformation: Each added header was added to new skb fragment, that caused skb fragmentation. Appending value to existing in message header was limited by size which remains after transformation. We could use mixed approach, encode headers inplace while message has room and then expand skb by new fragments, but maintain both approaches and code complexity not worth it.

Now transformation consist of following steps:
1. Find skb and fragment contains body or last crlf, if body not exists.
2. Delete found data from skb. Keep skb.
3. Hpack encode headers then add to skb from previous step. Headers adds as paged fragments, now we don't use headroom. HTTP2_MAX_OFFSET not used anymore during HTTP1 -> HTTP2 transformation without cache. All paged data in skb allocates from response's TfwPool.
4. Forward message to client.
5. Cleanup all data that was deleted from skbs in first step.

TfwPool was patched to statisfy new requirments. Now we use compound pages for order above zero and `put_page` insted of `free_pages`, because during free pages might be still used by skbs.

Fixed CONTINUATION frame.


fw/http.c
```diff
@@ -4591,17 +4591,28 @@ tfw_h2_resp_add_loc_hdrs(TfwHttpResp *resp, const TfwHdrMods *h_mods,
...
 		return 0;
 
 	for (i = 0; i < h_mods->sz; ++i) {
 		const TfwHdrModsDesc *desc = &h_mods->hdrs[i];
 		int r;
+		unsigned short hid = desc->hid;
 
 		/* FIXME: this is a temporary WA for GCC12, see #1695 for details */
-		if (test_bit(i, mit->found) || (TFW_STR_CHUNK(desc->hdr, 1) == NULL))
+		if (test_bit(i, mit->found)
```
const-t, line 4603: After fixing `resp_hdr_add` for cached responses `test_bit(i, mit->found)` can be deleted. Because it's not used anymore for http2 headers addition.

fw/http_msg.c
```diff
@@ -1313,115 +1329,253 @@ tfw_http_msg_expand_data(TfwMsgIter *it, struct sk_buff **skb_head,
...
+				goto end;
+			}
+		}
+
+		skb = it->skb;
+		it->skb = it->skb->next;
+		ss_skb_unlink(&it->skb_head, skb);
+		ss_skb_queue_tail(&cleanup->skb_head, skb);
+
+	} while (it->skb != NULL);
```
krizhanovsky, line 1543: @RomanBelozerov I'm wondering if we have a test, which sends an HTTP response to Tempesta with very small chunks, especially headers, and Tempesta forwards the response to a client via HTTP/2? I believe we have some test for this, but I'm not sure about HTTP/2 client in this scenario...

fw/http.h
```diff
@@ -543,17 +527,18 @@ typedef struct {
  * @curr_ptr	- pointer in the skb to write the current header;
  * @bnd		- pointer to the boundary data (which should not be
  *		  overwritten);
```
krizhanovsky, line 529: `bnd` is removed now, as well as `curr` and `next`

fw/http_msg.c
```diff
@@ -1313,115 +1329,253 @@ tfw_http_msg_expand_data(TfwMsgIter *it, struct sk_buff **skb_head,
...
 
-		f_room = f_size - offset;
-		n_copy = min(c_size, f_room);
-		next_ptr = mit->curr_ptr + n_copy;
+/**
+ * Add first paged fragment using TfwPool and reserve room for HEADERS frame.
+ *
+ * This function must be used as start point of message transformation. After
+ * calling this you must use @pool only for allocating paged fragments during
+ * message trasformation. After it @pool can be used for anything else.
```
krizhanovsky, line 1364: I didn't get
> After calling this you must use @pool only for allocating paged fragments during message trasformation. After it @pool can be used for anything else.

Why? What happened if after the first paged fragment allocation we say reallocate some TfwStr? What "after it" means exactly - single paged fragment allocation?

fw/http_msg.c
```diff
@@ -1313,115 +1329,253 @@ tfw_http_msg_expand_data(TfwMsgIter *it, struct sk_buff **skb_head,
...
+			      const TfwStr *str)
 {
-	return tfw_h2_msg_rewrite_data_common(mit, str, stop, memcpy_fast);
+	return __tfw_http_msg_expand_from_pool(mit, pool, str, memcpy_fast);
 }
 
 int
-tfw_h2_msg_rewrite_data_lc(TfwHttpTransIter *mit, const TfwStr *str,
-			   const char *stop)
+tfw_http_msg_expand_from_pool_lc(TfwHttpTransIter *mit, TfwPool* pool,
```
krizhanovsky, line 1456: `tfw_h2_msg_rewrite_data()` and `tfw_h2_msg_rewrite_data_lc()` should be removed from http_msg.h

fw/ss_skb.h
```diff
@@ -236,5 +236,6 @@ int ss_skb_to_sgvec_with_new_pages(struct sk_buff *skb, struct scatterlist *sgl,
                                    struct page ***old_pages);
 int ss_skb_cut_extra_data(struct sk_buff *skb_head, struct sk_buff *skb,
 			  int frag_num, char *curr, const char *stop);
```
krizhanovsky, line 238: The function is removed by the patch

krizhanovsky (changes requested):
I didn't notice any severe issue with the pull request, but there are still some things to do and fix.

Please make all the following changes by new separate patches, do not rewrite the current patches, to simplify next reviews. Later, on the merge time, you can squash senseless patches.

EvgeniiMekhanik:
There are several tests which doesn't work on this branch https://github.com/tempesta-tech/tempesta-test/commit/59db00a398359fb1872801977a909215585378e5#diff-46cbe3acf14e83bddf0360bdc9ab952e6673ea0360c67f246da3cbdae49a6eb1R493 (test_big_header_after_setting_header_table_size, test_big_body_after_setting_header_table_size) (very big body or very big header)

fw/http_msg.c
```diff
@@ -912,8 +912,9 @@ tfw_http_msg_hdr_xfrm_str(TfwHttpMsg *hm, const TfwStr *hdr, unsigned int hid,
 	}
 
 	if (unlikely(append && hid < TFW_HTTP_HDR_NONSINGULAR)) {
-		T_WARN("Appending to nonsingular header %d\n", hid);
-		return -ENOENT;
+		T_WARN("Appending to singular header '%.*s'\n",
+		       PR_TFW_STR(TFW_STR_CHUNK(hdr, 0)));
+		return 0;
```
EvgeniiMekhanik, line 917: Why we don't drop response in this case now?

const-t: We shouldn't, it is a legit case. Just print a warning.

fw/hpack.c
```diff
@@ -3592,43 +3592,53 @@ tfw_hpack_hdr_add(TfwHttpResp *__restrict resp, TfwStr *__restrict hdr,
...
 		return r;
 
 	if (!hdr)
 		return 0;
 
-	if (WARN_ON_ONCE(TFW_STR_PLAIN(hdr)))
+	if (WARN_ON_ONCE(TFW_STR_PLAIN(hdr) || TFW_STR_DUP(hdr)))
 		return -EINVAL;
 
+	tfw_http_hdr_split(hdr, &s_name, &s_val, inplace);
```
EvgeniiMekhanik: We should check return code here

fw/hpack.c
```diff
@@ -3592,43 +3592,53 @@ tfw_hpack_hdr_add(TfwHttpResp *__restrict resp, TfwStr *__restrict hdr,
...
-		if (WARN_ON_ONCE(!c))
-			return -EINVAL;
-	}
+	r = tfw_http_msg_expand_from_pool(mit, resp->pool, &s_vlen);
 
-	end = hdr->chunks + hdr->nchunks;
-	tfw_str_collect_cmp(c, end, &s_val, NULL);
+	if (unlikely(r))
+		return r;
+	if (!TFW_STR_EMPTY(&s_val)) {
```
EvgeniiMekhanik, line 3601: extra "{"

s0nx (changes requested):

EvgeniiMekhanik:
Please see https://github.com/tempesta-tech/tempesta-test/commit/465eecf5d0416d64c6c0e8d0477d6731133ace5f i add several tests, which doesn't work. Also test_flow_control_window_with_long_headers_and_body_FOR_1820 crash on this branch

fw/http_msg.c
```diff
@@ -1313,115 +1329,253 @@ tfw_http_msg_expand_data(TfwMsgIter *it, struct sk_buff **skb_head,
...
+ * as source for message trasformation.
+ */
+void
+tfw_h2_msg_cutoff_headers(TfwHttpResp *resp, TfwHttpRespCleanup* cleanup)
+{
+	char *begin, *end;
+	TfwMsgIter *it = &resp->mit.iter;
+	char* body = TFW_STR_CHUNK(&resp->body, 0)->data;
+	TfwStr *crlf = TFW_STR_LAST(&resp->crlf);
+	char *off = resp->body.len ? body : crlf->data + crlf->len;
```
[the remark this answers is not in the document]
const-t: Yes, they can live in different chunk, however I use only last chunk.

fw/http_msg.c
const-t: For fragments we trying to keep memory continuous. If we allocate some TfwStr after first paged fragment allocation we break continuous. 

For instance:  We have 2kb http headers, these headers have to be encoded into single paged fragment which is continuous and pool has enough space in current chunk. If we allocate TfwStr in the middle of encoding process, we got a gap between data, which will split the paged fragment. As a result we have 2 paged fragments.

fw/http_msg.c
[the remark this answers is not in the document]
const-t: How? `ss_skb_unlink` calls every iteration and eventually `it->skb` will become NULL.

fw/http_msg.c
```diff
@@ -1312,6 +1327,251 @@ tfw_http_msg_expand_data(TfwMsgIter *it, struct sk_buff **skb_head,
...
+	else
+		addr = tfw_pool_alloc_not_align_np(pool, FRAME_HEADER_SIZE,
+						   &np);
+
+	if (unlikely(!addr))
+		return -ENOMEM;
+
+	r = ss_skb_add_frag(it->skb_head, it->skb, addr, ++it->frag,
+			    FRAME_HEADER_SIZE);
+	if (unlikely(r))
```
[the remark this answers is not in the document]
const-t: It's unnecessary,  if we can't allocate fragment response will be dropped and destroyed, during destroying response `pool` also be destroyed.

fw/http.c
```diff
@@ -4921,82 +4859,60 @@ tfw_h2_make_frames(TfwHttpResp *resp, unsigned int stream_id,
...
-				     fr_flags);
+		mit->iter.frag = 0;
+		r = __tfw_h2_make_frames(h_len, data, sizeof(buf), iter, skew,
+					 max_sz, &frame_hdr_str, &frame_hdr,
+					 HTTP2_F_END_HEADERS);
+		if (unlikely(r))
+			return r;
 	}
 
 	if (local_response)
```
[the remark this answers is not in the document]
const-t: > Also we discussed that we should not parse body with TfwStr to not to spend resources in vain, but it seems an issue for this was never created.

Implemented in PR #1837

fw/http_sess.c
```diff
@@ -425,10 +425,8 @@ tfw_http_sticky_add(TfwHttpResp *resp, bool cache)
 	      PR_TFW_STR(&sticky->name), len, buf);
 
 	if (to_h2) {
-		TfwH2TransOp op = cache ? TFW_H2_TRANS_EXPAND : TFW_H2_TRANS_ADD;
-
 		set_cookie.hpack_idx = 55;
-		r = tfw_hpack_encode(resp, &set_cookie, op, !cache);
+		r = tfw_hpack_encode(resp, &set_cookie, !cache, !cache);
```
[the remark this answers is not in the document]
const-t: I'm not sure if this will give any positive result. For cache responses we allocate `page` and "expand" fragments using this page, this approach avoids fragmentation.

fw/http.c
```diff
@@ -4584,63 +4589,76 @@ tfw_h2_resp_add_loc_hdrs(TfwHttpResp *resp, const TfwHdrMods *h_mods,
...
+	unsigned int idx;
+
+	if (!h_mods)
+		return false;
+
+	for (idx = 0; idx < h_mods->sz; ++idx) {
+		TfwHdrModsDesc *desc = &h_mods->hdrs[idx];
+
+		if (!desc->append && ((hid >= TFW_HTTP_HDR_RAW
+		    && !__hdr_name_cmp(hdr, desc->hdr))
```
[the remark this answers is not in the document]
const-t: It's planned as part of #1831

fw/http_msg.c
krizhanovsky: @RomanBelozerov I created https://github.com/tempesta-tech/tempesta-test/issues/431 for this

fw/http.c
```diff
@@ -5396,6 +5328,7 @@ tfw_h2_resp_adjust_fwd(TfwHttpResp *resp)
 			dup->flags |= TFW_STR_HBH_HDR;
 	}
 
+	/* TODO: now we can do it without copy to buffer */
```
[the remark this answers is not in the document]
krizhanovsky: I created https://github.com/tempesta-tech/tempesta/issues/1854 for this

fw/ss_skb.h
```diff
@@ -166,6 +166,21 @@ ss_skb_queue_split(struct sk_buff *skb_head, struct sk_buff *skb)
...
 
+/*
+ * Insert @nskb in the list after @skb. Note that standard
+ * kernel 'skb_insert()' function does not suit here, as it
+ * works with 'sk_buff_head' structure with additional fields
+ * @qlen and @lock; we don't need these fields for our skb
+ * list, so a custom function had been introduced.
+ */
+static inline void
+skb_insert_after(struct sk_buff *skb, struct sk_buff *nskb)
```
krizhanovsky, line 177:
```suggestion
ss_skb_insert_after(struct sk_buff *skb, struct sk_buff *nskb)
```
just to make the name consistent with the other functions in the file

fw/http_msg.c
```diff
@@ -1312,6 +1327,251 @@ tfw_http_msg_expand_data(TfwMsgIter *it, struct sk_buff **skb_head,
...
+	struct skb_shared_info *si = skb_shinfo(it->skb);
+
+	addr = tfw_pool_alloc_not_align_np(pool, size, &np);
+	if (!addr)
+		return -ENOMEM;
+
+	if (np) {
+		r = ss_skb_add_frag(it->skb_head, it->skb, addr, ++it->frag,
+				    size);
+		if (unlikely(r))
```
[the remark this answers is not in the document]
krizhanovsky: In my understanding we drop the whole pool when we finalize processing on the error handling path. I.e. we do not try to continue normal processing with the pool.

fw/http_msg.c
```diff
@@ -1429,6 +1439,28 @@ __tfw_http_msg_expand_from_pool(TfwHttpTransIter *mit, TfwPool* pool,
...
+			{
+				struct sk_buff *nskb;
+
+				nskb = ss_skb_alloc(0);
+				if (!nskb)
+					return -ENOMEM;
+				skb_shinfo(nskb)->tx_flags =
+					skb_shinfo(it->skb)->tx_flags;
+				skb_insert_after(it->skb, nskb);
+				skb_shinfo(nskb)->nr_frags = 0;
```
krizhanovsky, line 1457: We have the same code in `__extend_pgfrags()` - maybe move it to an inline function in `ss_skb.h`?

s0nx (approved):
LGTM

EvgeniiMekhanik (approved):
LGTM, but please squash all fixes in appropriate commits

Issue #1813, opened 2023-02-08
ETag Header break parser


pale-emperor:
# Motivation
When Wordpress works behind tempesta-fw as reverse-proxy. Some functional unaviable.
For example when trying to open page editor got following error in dmesg:

```
[1295796.234477] [tempesta fw] Warning: Parser error: state=Resp_HdrEtagV input(-8)=0x36('
                 Etag: 6.1.1
                 Expires: Thu, 08 Feb 2024 15:12:5') data_len=2834 off=30
[1295796.234771] [tempesta fw] Warning: response dropped: processing error: 1.20.146.249
[1295796.235038] [tempesta fw] Warning: Paired request missing, HTTP Response Splitting attack?
```

# Testing

It reproduces every time when open WordPress page-editor behind tempesta-reverse proxy

PR #1821, opened 2023-02-16
Implement parsing Etag without double quotes

EvgeniiMekhanik:
According to RFC #7232 Etag should be enclosed in
double quotes, but unfortunately many do not follow RFC and send Etag without double quotes (wordpress for example). Implement parsing of such Etag values.

Closes: #1813

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -1261,14 +1261,41 @@ tfw_cache_h2_copy_int(unsigned int *acc_len, unsigned long src,
 	return 0;
 }
 
+static int
+tfw_cache_copy_str_with_extra_quotes(TfwCacheEntry *ce, char **p, TdbVRec **trec,
+				     TfwStr *src, size_t *tot_len,
+				     bool need_extra_quotes)
+{
+#define ADD_ETAG_QUOTE(flag)                                            \
+do {                                                                    \
+        TfwStr quote = { .data = "\"", .len = 1, .flags = flag };       \
+        if (tfw_cache_h2_copy_str(&ce->hdr_len, p, trec, &quote,        \
+                                  tot_len))                             \
+                return -ENOMEM;                                         \
+} while(0)
+
+	if (need_extra_quotes)
+		ADD_ETAG_QUOTE(0);
+
+	if (tfw_cache_h2_copy_str(&ce->hdr_len, p, trec, src, tot_len))
+		return -ENOMEM;
+
+	if (need_extra_quotes)
+		ADD_ETAG_QUOTE(TFW_STR_VALUE);
+
+	return 0;
+
+#undef ADD_ETAG_QUOTE
+}
+
 /**
  * Deep HTTP header copy to TdbRec.
  * @hdr is copied in depth first fashion to speed up upcoming scans.
  * @return number of copied bytes on success and negative value otherwise.
  */
 static long
-tfw_cache_h2_copy_hdr(TfwCacheEntry *ce, char **p, TdbVRec **trec, TfwStr *hdr,
-		      size_t *tot_len)
+tfw_cache_h2_copy_hdr(TfwCacheEntry *ce, TfwHttpResp *resp, int hid, char **p,
+		      TdbVRec **trec, TfwStr *hdr, size_t *tot_len)
 {
 	TfwCStr *cs;
 	long n = sizeof(TfwCStr);
@@ -1276,6 +1303,10 @@ tfw_cache_h2_copy_hdr(TfwCacheEntry *ce, char **p, TdbVRec **trec, TfwStr *hdr,
 	bool dupl = TFW_STR_DUP(hdr);
 	unsigned int init_len = ce->hdr_len;
 	TfwStr s_nm, s_val, *dup, *dup_end;
+	unsigned long s_val_len;
+	const bool need_extra_quotes =
+		test_bit(TFW_HTTP_B_HDR_ETAG_HAS_NO_QOUTES, resp->flags)
+		&& (hid == TFW_HTTP_HDR_ETAG);
 
 	T_DBG3("%s: ce=[%p] p=[%p], trec=[%p], tot_len='%zu'\n", __func__, ce,
 	       *p, *trec, *tot_len);
@@ -1288,7 +1319,8 @@ tfw_cache_h2_copy_hdr(TfwCacheEntry *ce, char **p, TdbVRec **trec, TfwStr *hdr,
 		tfw_http_hdr_split(hdr, &s_nm, &s_val, true);
 
 		st_index = hdr->hpack_idx;
-		h_len = tfw_h2_hdr_size(s_nm.len, s_val.len, st_index);
+		h_len = tfw_h2_hdr_size(s_nm.len, s_val.len, st_index) +
+			2 * need_extra_quotes;
 
 		/* Don't split short strings. */
 		if (sizeof(TfwCStr) + h_len <= L1_CACHE_BYTES)
@@ -1333,10 +1365,12 @@ tfw_cache_h2_copy_hdr(TfwCacheEntry *ce, char **p, TdbVRec **trec, TfwStr *hdr,
 				return -ENOMEM;
 		}
 
-		if (tfw_cache_h2_copy_int(&ce->hdr_len, s_val.len, 0x7f, p,
+		s_val_len =  s_val.len + 2 * need_extra_quotes;
+		if (tfw_cache_h2_copy_int(&ce->hdr_len, s_val_len, 0x7f, p,
 					  trec, tot_len)
-		    || tfw_cache_h2_copy_str(&ce->hdr_len, p, trec, &s_val,
-					     tot_len))
+		    || tfw_cache_copy_str_with_extra_quotes(ce, p, trec,
+							    &s_val, tot_len,
+							    need_extra_quotes))
 			return -ENOMEM;
 
 		CSTR_WRITE_HDR(0, ce->hdr_len - prev_len);
@@ -1501,6 +1535,13 @@ __set_etag(TfwCacheEntry *ce, TfwHttpResp *resp, long h_off, TdbVRec *h_trec,
 	tfw_http_hdr_split(h, &s_dummy, &s_val, true);
 	c_size = 2 + tfw_hpack_int_size(s_val.len, 0x7F);
 	CHECK_REC_SPACE();
+
+	if (test_bit(TFW_HTTP_B_HDR_ETAG_HAS_NO_QOUTES, resp->flags)) {
+		len += 1;
+		c_size = 1;
+		CHECK_REC_SPACE();
+	}
+
 	TFW_STR_FOR_EACH_CHUNK(c, &h_val, end) {
 		if (c->flags & TFW_STR_VALUE) {
 			flags = c->flags;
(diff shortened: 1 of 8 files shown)


krizhanovsky:
It also makes sense to check our current functional and unit tests and adjust them to test the new ETag behavior. And could you please comment the code that we violate RFC in ETag parsing to satisfy WordPress.

I think Nginx just don't parse ETags, i.e. accept almost anything for it. Typically it doesn't check the headers grammar.

EvgeniiMekhanik:
> It also makes sense to check our current functional and unit tests and adjust them to test the new ETag behavior. And could you please comment the code that we violate RFC in ETag parsing to satisfy WordPress.
> 
> I think Nginx just don't parse ETags, i.e. accept almost anything for it. Typically it doesn't check the headers grammar.

You right about Nginx.

pale-emperor (approved):
Wordpress Etag problem is gone - so LGTM

fw/str.h
```diff
@@ -239,6 +239,8 @@ basic_stricmp_fast(const BasicStr *s1, const BasicStr *s2)
 #define TFW_STR_NOCCPY_HDR	0x20
 /* Weak identifier was set for Etag value. */
 #define TFW_STR_ETAG_WEAK	0x40
+/* Etag value isn't enclosed in double quotes */
+#define TFW_STR_ETAG_NO_DOUBLE_QUOTES 0x80
```
krizhanovsky, line 243: I see that you introduced the flag because there is already `TFW_STR_ETAG_WEAK`. Apparently it was stored in the `TfwStr` descriptor of the header for (efficient) strong matching. However, according to RFC 7234 4.3.2 caches deal with `If-None-Match` and `If-Modified-Since` only, so use weak matching only (see `tfw_handle_validation_req()`).

So I believe storing both of the flags in `TfwStr` is wrong. After all, `TfwStr` is a very generic string descriptor, not only for the etag values and we need to have good reasons to introduce very specific flags to the very generic layer.

Both of the flags are used for the parser only and should be handler only there, probably in `TfwHttpParser->flags`

fw/t/unit/test_http1_parser.c
```diff
@@ -2833,15 +2833,18 @@ TEST(http1_parser, etag)
 	EXPECT_BLOCK_RESP(RESP_ETAG_START header "\r\n" RESP_ETAG_END)
 
 	FOR_ETAG("ETag:   \"dummy\"  ",  "dummy\"", EXPECT_FALSE);
+	FOR_ETAG("ETag:   dummy  ", "dummy\"", EXPECT_FALSE);
+	FOR_ETAG("Etag: 'dummy'", "'dummy'\"", EXPECT_FALSE);
+	FOR_ETAG("Etag:   *", "*\"", EXPECT_FALSE);
```
krizhanovsky, line 2838: Shouldn't this be
```suggestion
        FOR_ETAG("Etag:   *\"", "*", EXPECT_FALSE);
```
or
```suggestion
        ETAG_BLOCK("Etag:   *\"");
```
, i.e. we pass `Etag: *"` to the parser and expect that it doesn't match to `*` (i.e. we don't allow single quote)?

krizhanovsky (changes requested):
Besides the scary place, the PR looks OK, but it follows the existing and wrong way to place ETag flags in `TfwStr`, which makes sense to fix

fw/http_parser.c
```diff
@@ -2917,6 +2925,24 @@ __parse_etag_or_if_nmatch(TfwHttpMsg *hm, unsigned char *data, size_t len)
 
 		if (IS_WS(c))
 			__FSM_I_MOVE_fixup(I_Etag, 1, 0);
+
+		/*
+		 * According to RFC 7232 2.3:
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Fixed

fw/http_parser.c
```diff
@@ -2917,6 +2925,24 @@ __parse_etag_or_if_nmatch(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+		 * An entity-tag consists of an opaque quoted string, possibly
+		 * prefixed by a weakness indicator.
+		 * But unfortunately many do not follow RFC and send Etag
+		 * without double quotes (wordpress for example), so here
+		 * we write warning for this case and process this Etag.
+		 */
+		if (!if_nmatch) {
+			T_WARN("Etag header isn't enclosed in double quotes");
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Fixed)

fw/http_parser.c
```diff
@@ -458,6 +458,14 @@ do {									\
 #define __FSM_I_MOVE_fixup(to, n, flag)					\
 	__FSM_I_MOVE_fixup_f(to, n, &msg->stream->parser.hdr, flag)
 
+#define __FSM_I_MOVE_fixup_quote(to, flag)				\
+do {									\
+	__msg_field_fixup_pos(&msg->stream->parser.hdr, "\"", 1);	\
+	__FSM_I_field_chunk_flags(&msg->stream->parser.hdr, flag);	\
+	parser->_i_st = &&to;						\
+	goto to;							\
+} while (0)
```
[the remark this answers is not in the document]
EvgeniiMekhanik: fixed

fw/t/unit/test_http1_parser.c
EvgeniiMekhanik: Fixed to ETAG_BLOK

fw/http_parser.c
```diff
@@ -2917,6 +2925,24 @@ __parse_etag_or_if_nmatch(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+		 * According to RFC 7232 2.3:
+		 * An entity-tag consists of an opaque quoted string, possibly
+		 * prefixed by a weakness indicator.
+		 * But unfortunately many do not follow RFC and send Etag
+		 * without double quotes (wordpress for example), so here
+		 * we write warning for this case and process this Etag.
+		 */
+		if (!if_nmatch) {
+			T_WARN("Etag header isn't enclosed in double quotes");
+			parser->hdr.flags |= TFW_STR_ETAG_NO_DOUBLE_QUOTES;
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Fixed

fw/str.h
EvgeniiMekhanik: I fix it for TFW_STR_ETAG_NO_DOUBLE_QUOTES, but i don't agree with you about TFW_STR_ETAG_WEAK. For 'If-Match' header we should use a strong comparison. I don't know why we don't check this header, but later when we implement it we need to store this flag for strong matching.

EvgeniiMekhanik: Fixed, also remove TFW_STR_ETAG_WEAK at all

fw/http_msg.c
```diff
@@ -256,11 +256,14 @@ __http_msg_hdr_val(TfwStr *hdr, unsigned id, TfwStr *val, bool client)
...
-	 * Only Host header is allowed to be empty.
+	 * Only Host header is allowed to be empty but because
+	 * we don't follow RFC and allow Etag header to be not
+	 * enlosed in double quotes it also can be empty.
 	 * If header string is plain, it is always empty header.
 	 * Not empty headers are compound strings.
 	 */
-	BUG_ON(id == TFW_HTTP_HDR_HOST ? nlen > hdr->len : nlen >= hdr->len);
+	BUG_ON(id == TFW_HTTP_HDR_HOST || id == TFW_HTTP_HDR_ETAG ?
+	       nlen > hdr->len : nlen >= hdr->len);
```
krizhanovsky, line 266:
```suggestion
	BUG_ON(id == TFW_HTTP_HDR_HOST
               || id == TFW_HTTP_HDR_ETAG ? nlen > hdr->len : nlen >= hdr->len);
```

fw/http_parser.c
```diff
@@ -2863,13 +2863,20 @@ __FSM_STATE(st) {							\
...
+	int no_quotes, r = CSTR_NEQ;
 	bool if_nmatch = TFW_CONN_TYPE(hm->conn) & Conn_Clnt;
 	__FSM_DECLARE_VARS(hm);
 
+#define ETAG_MOVE_fixup_flag(flag, to)					\
+do {									\
+	__set_bit(flag, msg->flags);					\
+	parser->_i_st = &&to;						\
+	goto to;							\
+} while (0)
```
krizhanovsky, line 2875: We don't need the macro since it's used in only one place

krizhanovsky (changes requested):
Maybe I'm wrong, but I propose another, looks more efficient and simple, way to store ETag with quotes, in the cache

fw/http_parser.c
EvgeniiMekhanik: fixed

fw/http_msg.c
EvgeniiMekhanik: fixed

fw/http_msg.c
```diff
@@ -256,11 +256,14 @@ __http_msg_hdr_val(TfwStr *hdr, unsigned id, TfwStr *val, bool client)
 
 	nlen = hdr_lens[client][id];
 	/*
-	 * Only Host header is allowed to be empty.
+	 * Only Host header is allowed to be empty but because
+	 * we don't follow RFC and allow Etag header to be not
+	 * enlosed in double quotes it also can be empty.
```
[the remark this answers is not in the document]
EvgeniiMekhanik: fixed

fw/cache.c
```diff
@@ -1332,10 +1365,12 @@ tfw_cache_h2_copy_hdr(TfwCacheEntry *ce, char **p, TdbVRec **trec, TfwStr *hdr,
 				return -ENOMEM;
 		}
 
-		if (tfw_cache_h2_copy_int(&ce->hdr_len, s_val.len, 0x7f, p,
+		s_val_len = (need_extra_quotes ? s_val.len + 2 : s_val.len);
```
krizhanovsky, line 1368:
```suggestion
		s_val_len =  s_val.len + 2 * need_extra_quotes;
```
In my understanding since `bool` isn't a native C type, GCC still can do the optimization and generates a conditional jump

krizhanovsky (approved):
I'm good with the new approach and still like it more than the previous one. It's bit clumsy to introduce 3 length updates, but it isn't worse that to add and remove quotes. Also I think it's more efficient: the compiler and CPU are good with condition branches while we might hit reallocations in the TfwStr operations.

fw/cache.c
```diff
@@ -1261,21 +1261,52 @@ tfw_cache_h2_copy_int(unsigned int *acc_len, unsigned long src,
...
+#define ADD_ETAG_QUOTE(flag)                                            \
+do {                                                                    \
+        TfwStr quote = { .data = "\"", .len = 1, .flags = flag };       \
+        if (tfw_cache_h2_copy_str(&ce->hdr_len, p, trec, &quote,        \
+                                  tot_len))                             \
+                return -ENOMEM;                                         \
+} while(0)
+
+	if (need_extra_quotes)
+		ADD_ETAG_QUOTE(0);
```
const-t, line 1278: Why do we add first quote without `TFW_STR_VALUE` flag?

EvgeniiMekhanik: Since we don't do it when we parse Etag value in `__parse_etag_or_if_nmatch` in http_parser.c. Here i implement the same behaviour

const-t: It looks strange, part of value has "value" flag and part not. Do we have some places dependent of such behavior? I suggest: Find all places where it used and ensure that after changing behaviour we not broke something. If we don't broke anything, then add `TFW_STR_VALUE` for first quote and remove condition from `__set_etag`.

EvgeniiMekhanik: No, you are wrong here! Etag value doesn't contains quotes! Quotes are only used to separate etag values, We save last quote value in etag only as a stop symbol for `tfw_strcmpspn`!

Issue #1411, opened 2020-05-21
H2 related corrections and optimizations.


aleksostapenko:
### Bugs

1. The header 'age' is added unconditionally [here](https://github.com/tempesta-tech/tempesta/blob/50ce6e289b989e396d7a6d17797b23f843efa21c/tempesta_fw/cache.c#L2044); thus, looks like it should be skipped in `tfw_cache_copy_resp()`;

2. [data = tfw_pool_alloc_not_align(it->pool, sz)](https://github.com/tempesta-tech/tempesta/blob/50ce6e289b989e396d7a6d17797b23f843efa21c/tempesta_fw/hpack.c#L1098) should be moved down - after the static index processing;

3. Header 'host' can have empty field-value [here](https://github.com/tempesta-tech/tempesta/blob/50ce6e289b989e396d7a6d17797b23f843efa21c/tempesta_fw/hpack.c#L1480) (this possibility is mentioned in [RFC 7230 section 5.4](https://tools.ietf.org/html/rfc7230#section-5.4));

4. Due to the nature of the internal implementation of encoder dynamic table in TempestaFW (for performance purposes), its maximum size is limited - it cannot be greater than HPACK_ENC_TABLE_MAX_SIZE; so for now, if client sends SETTINGS frame with new max window size greater than HPACK_ENC_TABLE_MAX_SIZE, the [table inconsistency](https://github.com/tempesta-tech/tempesta/blob/50ce6e289b989e396d7a6d17797b23f843efa21c/tempesta_fw/http_frame.c#L932) between endpoints may arise; to avoid this problem - the `WinodwUpdate` HEADERS frame with HPACK_ENC_TABLE_MAX_SIZE should be sent to client's decoder (within the first following HEADERS frame of the following SETTINGS acknowledgment, see [RFC 7541 section 4.2](https://tools.ietf.org/html/rfc7541#section-4.2));

5. The condition [here](https://github.com/tempesta-tech/tempesta/blob/50ce6e289b989e396d7a6d17797b23f843efa21c/tempesta_fw/http_frame.c#L771) should be 
```c
if (hclosed_streams->num <= TFW_MAX_CLOSED_STREAMS
   && (max_streams > ctx->streams_num
      || !hclosed_streams->num))
{
    . . .
}
```

[...]

PR #1831, opened 2023-03-01
Fixed framing error for resposnses received from cache with header modification

const-t:
tfw_hpack_cache_decode_expand: fixed http2 framing. However, function looks overcomplicated and combines http1 and http2 related logic.

tfw_cache_h2_decode_write: after decoding header part of `TfwDecodeCacheIter` was cleared including `acc_len` and
because of this caller of `tfw_cache_h2_decode_write` couldn't receive a total length of header. Fixed by clearing only `hdr_data` and `h2_data`.

Removed `__off` from `TfwDecodeCacheIter`, because we need to clear only two last members in this structure and it can be done via inner struct.

Removed `desc` from `TfwDecodeCacheIter`. It's not used.

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -189,6 +189,7 @@ __tfw_dbg_dump_ce(const TfwCacheEntry *ce)
  * @flags	- TFW_CSTR_DUPLICATE and TFW_CSTR_HPACK_IDX or zero;
  * @name_len	- Header name length. Used for raw headers;
  * @name_len_sz	- HPACK int size of @name_len;
+ * @len		- Total string length or number of duplicates;
  * @idx		- HPACK static index or index of special header;
  */
 typedef struct {
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -798,7 +798,6 @@ tfw_cache_skip_hdr(const TfwCStr *str, char *p, const TfwHdrMods *h_mods)
 	}
 
 	for (i = h_mods->spec_num; i < h_mods->sz; ++i) {
-		int cmplen;
 		char* mod_hdr_name;
 		size_t mod_hdr_len;
 
@@ -808,9 +807,8 @@ tfw_cache_skip_hdr(const TfwCStr *str, char *p, const TfwHdrMods *h_mods)
 			continue;
 
 		mod_hdr_name = TFW_STR_CHUNK(desc->hdr, 0)->data;
-		cmplen = min(hdr.len, mod_hdr_len);
 
-		if (!tfw_cstricmp(hdr.data, mod_hdr_name, cmplen))
+		if (!tfw_cstricmp(hdr.data, mod_hdr_name, hdr.len))
 			return true;
 	}
 
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -692,9 +692,7 @@ tfw_cache_h2_decode_write(TDB *db, TdbVRec **trec, TfwHttpResp *resp,
 		acc += m_len;
 		*data += m_len;
 		if (acc == len) {
-			bzero_fast(dc_iter->__off,
-				   sizeof(*dc_iter)
-				   - offsetof(TfwDecodeCacheIter, __off));
+			TFW_STR_INIT(&dc_iter->hdr_data);
 			break;
 		}
 
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -178,15 +178,22 @@ __tfw_dbg_dump_ce(const TfwCacheEntry *ce)
 #define __tfw_dbg_dump_ce(...)
 #endif
 
+/* TfwCStr contains duplicated header. */
+#define TFW_CSTR_DUPLICATE	TFW_STR_DUPLICATE
+/* TfwCStr contains special header and its id. */
+#define TFW_CSTR_SPEC_IDX	0x2
+
 /**
  * String header for cache entries used for TfwStr serialization.
  *
- * @flags	- only TFW_STR_DUPLICATE or zero;
+ * @flags	- TFW_CSTR_DUPLICATE and TFW_CSTR_HPACK_IDX or zero;
  * @len		- string length or number of duplicates;
+ * @idx		- HPACK static index or index of special header;
  */
 typedef struct {
 	unsigned long	flags : 8,
-			len : 56;
+			len : 48,
+			idx : 8;
 } TfwCStr;
 
 #define TFW_CSTR_MAXLEN		(1UL << 56)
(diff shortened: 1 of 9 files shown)


s0nx (approved):

krizhanovsky (changes requested):
We should not fix the code, we must rewrite it. At some point we decided to store HTTP/2 encoded responses in the cache, i.e. HTTP/2 must be the most optimized in our code, probably sacrificing HTTP/1 performance.

However, for now we **decode** responses fetched from the cache if we have a headers modification logic (quite useful option - it's even used in our current static website for Strict-Transport-Security) and **encode** them back in `tfw_hpack_encode()`, i.e. storing responses in HTTP/2 format is not just senseless, but probably makes performance even worse if we'd store responses in HTTP/1 format.

This is essentially #1411 point 15. I think we don't need to introduce a reverse index for now: on the configuration phase we can even just scan the static table (probably starting from `accept-charset`) to determine the static indexes. The modified headers, which aren't in the static table, we still need to check the dynamic table, but only in `tfw_h2_resp_adjust()`, and there we can check for duplicate header.

The main point is that we must not decode and encode responses from the cache.

const-t:
Mentioned in https://github.com/tempesta-tech/tempesta/pull/1820#discussion_r1125738001

fw/cache.c
```diff
@@ -178,15 +178,20 @@ __tfw_dbg_dump_ce(const TfwCacheEntry *ce)
...
- * @flags	- only TFW_STR_DUPLICATE or zero;
+ * @flags	- only TFW_CSTR_DUPLICATE and TFW_CSTR_HPACK_IDX or zero;
  * @len		- string length or number of duplicates;
+ * @idx		- HPACK static index or index of special header;
  */
 typedef struct {
 	unsigned long	flags : 8,
-			len : 56;
+			len : 48,
+			idx : 8;
```
const-t, line 194: As for me, we can use `unsigned int` here. In this case `len` becomes 2 bytes and it's enough to store value of the header. If so for consistency we must introduce such limit to non-cacheable responses. I believe adding limits may  decrease risk of resource exhaustion attacks or overflow attacks.

fw/vhost.c
```diff
@@ -1336,9 +1373,20 @@ tfw_location_init(TfwLocation *loc, tfw_match_t op, const char *arg,
...
+				    + TFW_USRHDRS_ARRAY_SZ);
+
 	loc->mod_hdrs[TFW_VHOST_HDRMOD_RESP].hdrs =
-			loc->mod_hdrs[TFW_VHOST_HDRMOD_REQ].hdrs + TFW_USRHDRS_ARRAY_SZ;
+		(TfwHdrModsDesc *)(loc->mod_hdrs[TFW_VHOST_HDRMOD_REQ].spec_hdrs
+				   + TFW_HTTP_HDR_RAW);
+
+	loc->mod_hdrs[TFW_VHOST_HDRMOD_RESP].spec_hdrs =
+		(TfwHdrModsDesc **)(loc->mod_hdrs[TFW_VHOST_HDRMOD_RESP].hdrs
+				    + TFW_USRHDRS_ARRAY_SZ);
```
krizhanovsky, line 1388: @RomanBelozerov I think it makes sense to test in https://github.com/tempesta-tech/tempesta-test/pull/414 a configuration to modify 64 headers to test the configuration code.

krizhanovsky (changes requested):
This version looks good. I'd only love to make several optimizations to make the code faster

fw/vhost.c
const-t: @krizhanovsky Seems configuration works fine, but when we use such configuration in tests we have problems in another places that covered by few issues.

fw/http_parser.c
```diff
@@ -1544,6 +1547,11 @@ __FSM_STATE(RGen_HdrOtherN) {						\
 			TFW_PARSER_BLOCK(RGen_HdrOtherN);		\
 		}							\
 		__msg_hdr_chunk_fixup(data, __data_off(p + __fsm_sz));	\
+		if (parser->hdr.len > HTTP_MAX_HDR_NAME_LEN) {		\
```
krizhanovsky, line 1550:
```suggestion
		if (unlikely(parser->hdr.len > HTTP_MAX_HDR_NAME_LEN)) {		\
```

fw/http_parser.c
```diff
@@ -9621,6 +9629,12 @@ tfw_h2_parse_req_hdr(unsigned char *data, unsigned long len, TfwHttpReq *req,
 		 * previous states trying known header names.
 		 */
 		__msg_hdr_chunk_fixup(data, len);
+		if (parser->hdr.len > HTTP_MAX_HDR_NAME_LEN) {
```
krizhanovsky, line 9632:
```suggestion
		if (unlikely(parser->hdr.len > HTTP_MAX_HDR_NAME_LEN)) {
```

krizhanovsky (approved):
LGTM with a couple of cleanups.

fw/hpack.h
```diff
@@ -30,6 +30,11 @@
 #define HPACK_TABLE_DEF_SIZE		4096
 #define HPACK_ENC_TABLE_MAX_SIZE	HPACK_TABLE_DEF_SIZE
 
+/* Limit for the HPACK variable-length integer. */
+#define HPACK_INT_LIMIT			(1 << 20)
+
+/* Static table offset where starts regular headers (not pseudo-headers) */
+#define HPACK_S_TABLE_REGULAR		14
```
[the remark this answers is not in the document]
const-t: Maybe I misunderstood, but HPACK static table defined as static scope array in hpack.c. We don't have enum for static table.

fw/vhost.c
const-t: Yes. You are right, this is `TestManyRequestHeaders`.

fw/hpack.c
```diff
@@ -181,6 +178,42 @@ do {								\
...
+	const TfwStr *h;
+
+	if (!TFW_STR_DUP(hdr))
+		h = hdr;
+	else
+		h = hdr->chunks;
+	fc = tolower(*(unsigned char *)(TFW_STR_CHUNK(h, 0)->data));
+
+	while (start < end) {
+		unsigned short mid = start + (end - start) / 2;
```
EvgeniiMekhanik, line 195: simple is (start + end) / 2

fw/cache.c
```diff
@@ -806,11 +806,19 @@ tfw_cache_skip_hdr(TfwCStr *str, char *p, const TfwHdrMods *h_mods)
...
+
 		desc = &h_mods->hdrs[i];
-		if (desc->append)
+		mod_hdr_len = TFW_STR_CHUNK(desc->hdr, 0)->len;
+		if (desc->append || mod_hdr_len != hdr.len)
 			continue;
 
-		if (!__hdr_name_cmp(&hdr, desc->hdr))
+		mod_hdr_name = TFW_STR_CHUNK(desc->hdr, 0)->data;
+		cmplen = min(hdr.len, mod_hdr_len);
```
EvgeniiMekhanik, line 810: as i see here hdr.len is always equal to mod_hdr_len, so we don't need cmplen

EvgeniiMekhanik (approved):

PR #1837, opened 2023-03-09
Fixed chunked body parsing error

const-t:
In our implementation of parsing chunked body we try to make as less TfwStr's chunks as possible, merging continuous memory TfwStr's chunks in single chunk. However, we didn't take into account that continuous memory can be used by different SKBs.

Body chunk "\r\n8fb1" can be located in continuous memory, but first part of chunk - "\r\n" can be used by one skb and second part "8fb1" by another.


krizhanovsky (approved):
LGTM

EvgeniiMekhanik (approved):

PR #1842, opened 2023-03-21
Fix various problems reported in #1703

s0nx:
- perform SKB linear data split during http/1 -> http/2 response transformation
- with the attached linux patch, we shouldn't get low-headroom SKBs anymore

### DIFF
--- a/linux-5.10.35.patch
+++ b/linux-5.10.35.patch
@@ -506,10 +506,10 @@ index 0dcd51fee..9a09576a8 100644
  struct kvec;
  
 diff --git a/include/linux/netdevice.h b/include/linux/netdevice.h
-index e37480b5f..617f4e76b 100644
+index e37480b5f..8236d5929 100644
 --- a/include/linux/netdevice.h
 +++ b/include/linux/netdevice.h
-@@ -154,11 +154,32 @@ static inline bool dev_xmit_complete(int rc)
+@@ -154,11 +154,22 @@ static inline bool dev_xmit_complete(int rc)
  # define LL_MAX_HEADER 32
  #endif
  
@@ -521,26 +521,16 @@ index e37480b5f..617f4e76b 100644
 + * to allocate 16 more bytes (5 - TLS header, 8 - IV, 3 - alignment).
 + */
 +#define TLS_MAX_HDR		16
-+/*
-+ * For fast transformation of HTTP/1.1 responses into HTTP/2 format, Tempesta
-+ * uses zero-copy in-place rewriting of the response data, right in original
-+ * skb. HTTP/2 data is almost always smaller of its source HTTP/1.1 data, but
-+ * for the sake of robustness we use 32-byte initial offset in front of skb
-+ * data. Thus, in order to guarantee the stack headers to fit, we should
-+ * increase the total space for them.
-+ */
-+#define HTTP2_MAX_OFFSET	32
 +#else
 +#define TLS_MAX_HDR		0
-+#define HTTP2_MAX_OFFSET	0
 +#endif
  #if !IS_ENABLED(CONFIG_NET_IPIP) && !IS_ENABLED(CONFIG_NET_IPGRE) && \
      !IS_ENABLED(CONFIG_IPV6_SIT) && !IS_ENABLED(CONFIG_IPV6_TUNNEL)
 -#define MAX_HEADER LL_MAX_HEADER
-+#define MAX_HEADER (LL_MAX_HEADER + TLS_MAX_HDR + HTTP2_MAX_OFFSET)
++#define MAX_HEADER (LL_MAX_HEADER + TLS_MAX_HDR)
  #else
 -#define MAX_HEADER (LL_MAX_HEADER + 48)
-+#define MAX_HEADER (LL_MAX_HEADER + 48 + TLS_MAX_HDR + HTTP2_MAX_OFFSET)
++#define MAX_HEADER (LL_MAX_HEADER + 48 + TLS_MAX_HDR)
  #endif
  
  /*
@@ -2040,6 +2030,64 @@ index 45fb450b4..48da5be43 100644
  	}
  
  	offset++;
+diff --git a/net/ipv4/ip_output.c b/net/ipv4/ip_output.c
+index 97975bed4..672f21290 100644
+--- a/net/ipv4/ip_output.c
++++ b/net/ipv4/ip_output.c
+@@ -82,6 +82,9 @@
+ #include <linux/netfilter_bridge.h>
+ #include <linux/netlink.h>
+ #include <linux/tcp.h>
++#ifdef CONFIG_SECURITY_TEMPESTA
++#include <net/tcp.h>
++#endif
+ 
+ static int
+ ip_fragment(struct net *net, struct sock *sk, struct sk_buff *skb,
+@@ -702,7 +705,31 @@ struct sk_buff *ip_frag_next(struct sk_buff *skb, struct ip_frag_state *state)
+ 	}
+ 
+ 	/* Allocate buffer */
++#ifdef CONFIG_SECURITY_TEMPESTA
++	/*
++	 * Since Tempesta FW tries to reuse incoming SKBs containing the response
++	 * from the backend, sometimes we might encounter an SKB with quite a small
++	 * head room, which is not big enough to accommodate all the transport headers
++	 * and TLS overhead.
++	 * It usually the case when working over loopback, tun/tap, bridge or similar
++	 * interfaces with small MTU. The issue is specific to aforementioned ifaces
++	 * because the outgoing SKB would be injected back to the stack.
++	 * In order not to reallocate sk_buffs' headroom on RX path,
++	 * allocate and reserve a little bit more memory on TX path.
++	 * Even though it would introduce some memory overhead, it's still
++	 * cheaper than doing transformation.
++	 *
++	 * It seems like no such actions are required for IPv6 counterparts:
++	 * ip6_fragment() / ip6_frag_next() due to the fact that the
++	 * lowest acceptable MTU (1280) is sufficient to fit all the headers.
++	 *
++	 * When receiving SKBs from the outter world, the NIC driver should
++	 * allocate and reserve all necessary space by itself.
++	 */
++	skb2 = alloc_skb(len + state->hlen + MAX_TCP_HEADER, GFP_ATOMIC);
++#else
+ 	skb2 = alloc_skb(len + state->hlen + state->ll_rs, GFP_ATOMIC);
++#endif
+ 	if (!skb2)
+ 		return ERR_PTR(-ENOMEM);
+ 
+@@ -711,7 +738,11 @@ struct sk_buff *ip_frag_next(struct sk_buff *skb, struct ip_frag_state *state)
+ 	 */
+ 
+ 	ip_copy_metadata(skb2, skb);
++#ifdef CONFIG_SECURITY_TEMPESTA
++	skb_reserve(skb2, MAX_TCP_HEADER);
++#else
+ 	skb_reserve(skb2, state->ll_rs);
++#endif
+ 	skb_put(skb2, len + state->hlen);
+ 	skb_reset_network_header(skb2);
+ 	skb2->transport_header = skb2->network_header + state->hlen;
 diff --git a/net/ipv4/tcp.c b/net/ipv4/tcp.c
 index 2384ac048..920b1f01f 100644
 --- a/net/ipv4/tcp.c


fw/http_msg.c
```diff
@@ -1525,24 +1525,36 @@ __tfw_h2_msg_shrink_frags(TfwMsgIter *it, const char *body,
 	}
 }
 
+static void
+tfw_h2_body_skb_set(TfwHttpResp *resp, struct sk_buff *nskb)
+{
+	if (test_bit(TFW_HTTP_B_CHUNKED, resp->flags)) {
+		TfwStr *h2_body = __tfw_h2_get_body_start(resp);
+		h2_body->skb = nskb;
```
const-t, line 1533: There may be few problems:
 
1.  You replace skb pointer only for "data chunks"(`__tfw_h2_get_body_start` skips "cut chunks"), although you need to replace skb for both "cut" and "data chunks". 
2.  `__tfw_h2_get_body_start` can't determine whether the found body start is in currently transformed skb.
3. `resp->body` must point to new skb anyway, because we will cutoff all cut chunks start from `resp->body`.

You can use `encoding` test suite with low mtu for debugging this part.

const-t (changes requested):

fw/ss_skb.c
```diff
@@ -1609,3 +1609,289 @@ ss_skb_add_frag(struct sk_buff *skb_head, struct sk_buff *skb, char* addr,
...
+				ss_set_skb_tso_segs(nskb, mss);
+
+				__skb_header_release(nskb);
+				__skb_queue_after(&sk->sk_write_queue,
+						  skb->next, nskb);
+
+				__skb_unlink(skb, &sk->sk_write_queue);
+				tcp_skb_tsorted_anchor_cleanup(skb);
+				kfree_skb(skb);
+			}
```
krizhanovsky, line 1883: If we need a larger headroom, why not to use something like `__pskb_copy_fclone()` to allocate a new skb with larger headroom and copy `skb` into it? Do we really need all the sophisticated code above?

fw/ss_skb.h
```diff
@@ -211,6 +216,44 @@ ss_skb_adjust_data_len(struct sk_buff *skb, int delta)
...
+do {							\
+	int i = 0;					\
+	struct sk_buff *skb = *queue;			\
+	if (likely(skb)) {				\
+		do {					\
+			lambda;				\
+			skb = skb->next;		\
+		} while (skb != *queue);		\
+	}						\
+} while(0)
```
krizhanovsky, line 245: Should also be under `DEBUG` macro - in this way a reader won't be nervous about an skb list traversal hidden somewhere in the code.

fw/http.c
```diff
@@ -1,4 +1,5 @@
-/**
+/*
+ *
```
krizhanovsky, line 2: senseless change

fw/http_msg.c
```diff
@@ -1551,11 +1563,39 @@ tfw_h2_msg_cutoff_headers(TfwHttpResp *resp, TfwHttpRespCleanup* cleanup)
 
 			if ((begin <= off) && (end >= off)) {
```
krizhanovsky, line 1564:
```suggestion
			if ((begin > off) || (end < off))
			        continue;
```

fw/ss_skb.c
```diff
@@ -1609,3 +1609,289 @@ ss_skb_add_frag(struct sk_buff *skb_head, struct sk_buff *skb, char* addr,
...
+		}
+	}
+	return 0;
+
+out_err:
+	T_WARN("%s: can't allocate new skb for response"
+	       " transformation(orig skb %pK)\n",
+	       __func__, skb);
+	return -ENOMEM;
+}
```
krizhanovsky, line 1709: This looks like a specialized version of `__split_linear_data()`: we should either use `__split_linear_data()` or unify both the functions to make them share the code. It's not good to have two sets of functions doing nearly the same.

fw/ss_skb.c
```diff
@@ -1609,3 +1609,289 @@ ss_skb_add_frag(struct sk_buff *skb_head, struct sk_buff *skb, char* addr,
...
+		/* Avoid the costly divide in the normal
+		 * non-TSO case.
+		 */
+		tcp_skb_pcount_set(skb, 1);
+		TCP_SKB_CB(skb)->tcp_gso_size = 0;
+	} else {
+		tcp_skb_pcount_set(skb, DIV_ROUND_UP(skb->len, mss_now));
+		TCP_SKB_CB(skb)->tcp_gso_size = mss_now;
+	}
+}
```
krizhanovsky, line 1761: We discussed this in another recent pull request - it's better to just export a Linux function instead of copying it

krizhanovsky (changes requested):
The patch introduces a lot of sophisticated code, some of which seems duplicate of existing code, and additional computation complexity reducing performance of any traffic. All of this is just to fix a rare problem. Either an implementation of the idea must be much more polished or we should go with a simpler solution like catch the case wherever we can catch it, log it and close a TCP connection.

fw/ss_skb.c
s0nx: Yes, we can technically use `__pskb_copy_fclone()` or `skb_cow() -> __pskb_expand_head()`, but those functions perform linear data copy, which would be smth like `rep; movsl`. I was trying to avoid explicit copying. Since we would allocate new SKB anyway, i thought moving the linear part as a fragment would be cheaper.

fw/sock.c
```diff
@@ -1113,9 +1115,155 @@ ss_set_callbacks(struct sock *sk)
...
 }
 EXPORT_SYMBOL(ss_set_callbacks);
 
+/* FIXME: the code related to HPACK dynamic table size update has been
+ * temporarily commented out. The reason is that it would not work
+ * on top of 1820 as of now and it would be substantially reworked in 1394.
+ */
+struct sk_buff *
+ss_sk_clnt_wxmit(struct sock *sk, struct sk_buff *skb, unsigned int mss,
+		 unsigned int *limit)
```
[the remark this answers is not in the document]
s0nx: Passing the limit by ptr was mostly to align the changes with future changes in #1845.

fw/http_msg.c
s0nx: I believe we can't skip the iteration here because there might be fragments in the current skb.

fw/ss_skb.c
```diff
@@ -1571,3 +1571,28 @@ ss_skb_add_frag(struct sk_buff *skb_head, struct sk_buff *skb, char* addr,
...
+int
+ss_skb_linear_transform(struct sk_buff *skb_head, struct sk_buff *skb,
+			unsigned char *split_point)
+{
+	int fpos, r;
+	TfwStr _;
+
+	if (!split_point) {
+		/* Usage of linear portion of SKB is not expected */
+		skb_pull(skb, skb_headlen(skb));
```
const-t, line 1588: IIUC skb that has "payload" for instance http headers will contain these headers in "tail part". Such skbs looks like this: `MAX_HEADERS + HTTP_HEADERS`, where `MAX_HEADERS` was reserved using `skb_reserve()` and `HTTP_HEADERS` placed into skb using `skb_put`. It implies the part contains `HTTP_HEADERS` must be removed using `ss_skb_put`, like you doing in code below. Also would be better to set `tail_lock` flag.

linux-5.10.35.patch
```diff
@@ -2040,6 +2030,63 @@ index 45fb450b4..48da5be43 100644
...
++#ifdef CONFIG_SECURITY_TEMPESTA
++/*
++ * Since Tempesta FW tries to reuse incoming SKBs containing the response
++ * from the backend, sometimes we might encounter an SKB with quite a small
++ * head room, which is not big enough to accommodate all the transport headers
++ * and TLS overhead.
++ * It usually the case when working over loopback, tun/tap, bridge or similar
++ * interfaces with small MTU. The issue is specific to aforementioned ifaces
++ * because the outgoing SKB would be injected back to the stack.
++ * In order not to reallocate sk_buffs' headroom on on RX path,
```
EvgeniiMekhanik, line 2050: "headroom on on RX path" should be changed to "headroom on RX path"

linux-5.10.35.patch
```diff
@@ -2040,6 +2030,63 @@ index 45fb450b4..48da5be43 100644
...
++ */
++
++/* Redefine this constant from include/net/tcp.h here, not to bring
++ * the whole header */
++#define MAX_TCP_HEADER	L1_CACHE_ALIGN(128 + MAX_HEADER)
++#endif
++
+ 	/* Allocate buffer */
++#ifdef CONFIG_SECURITY_TEMPESTA
++	skb2 = alloc_skb(len + state->hlen + state->ll_rs
```
EvgeniiMekhanik, line 2070: We can allocate only MAX_TCP_HEADER here and reserve MAX_TCP_HEADER - len - state->hlen

fw/ss_skb.h
```diff
@@ -288,4 +301,30 @@ int ss_skb_to_sgvec_with_new_pages(struct sk_buff *skb, struct scatterlist *sgl,
...
+
+#define SS_SKB_QUEUE_DUMP(queue)			\
+	ss_skb_queue_for_each_do(queue, {		\
+		pr_warn("#%2d skb => %pK\n", i++, skb);	\
+		skb_dump(KERN_WARNING, skb, true);	\
```
EvgeniiMekhanik, line 324: May be use "KERN_DEBUG" here? If this is debug print

fw/ss_skb.c
```diff
@@ -1571,3 +1571,28 @@ ss_skb_add_frag(struct sk_buff *skb_head, struct sk_buff *skb, char* addr,
 
 	return 0;
 }
+
+/* Using @split_point transform the remaining linear portion of original @skb
+ * to the firsst fragment of the same SKB. Existing fragments of @skb
```
EvgeniiMekhanik, line 1576: first

fw/ss_skb.h
```diff
@@ -256,6 +256,19 @@ ss_skb_alloc(size_t n)
 	return skb;
 }
 
+/* Move @cnt existing fragment descriptors within the SKB.
+ */
+static inline void
+ss_skb_frags_move(struct sk_buff *skb, unsigned int new_pos,
```
EvgeniiMekhanik, line 262: I implement such function in PR 1845, but for moving frags from one skb to another. May be think about one version which is good for both this cases?

fw/http_msg.c
```diff
@@ -1550,11 +1550,18 @@ tfw_h2_msg_cutoff_headers(TfwHttpResp *resp, TfwHttpRespCleanup* cleanup)
...
 				it->frag = -1;
-				/* TODO: Handle this case, most likely for low MTU
-				 * it will be implemented as part of #1703 */
-				return;
+				/* We would end up here if the start of the body or
+				 * the end of CRLF lies within the linear data area
+				 * of the current @it->skb
+				 */
+				ret = ss_skb_linear_transform(it->skb_head,
+							      it->skb, body);
```
EvgeniiMekhanik, line 1558: why we use body as a split point not "off"?

fw/ss_skb.c
```diff
@@ -1571,3 +1571,28 @@ ss_skb_add_frag(struct sk_buff *skb_head, struct sk_buff *skb, char* addr,
...
+{
+	int fpos, r;
+	TfwStr _;
+
+	if (!split_point) {
+		/* Usage of linear portion of SKB is not expected */
+		skb_pull(skb, skb_headlen(skb));
+	} else {
+		unsigned int off = split_point - skb->data;
+		r = __split_linear_data(skb_head, skb, split_point, 0, &_, &fpos);
```
EvgeniiMekhanik, line 1591: why we pass 0 as a four argument? not -off?

linux-5.10.35.patch
const-t: I believe we can't do this. For instance, if `len + state->hlen` above than `MAX_TCP_HEADER` we will reserve for headroom only `state->ll_rs` that can be not enough to contain all necessary TCP + TLS headers. At this place we can't predict required size of headers. If we could determine whether this skb will be reused, then we could avoid reserving MAX_TCP_HEADER and use only `ll` room.

krizhanovsky (approved):
I just don't have anything to add to @const-t and @EvgeniiMekhanik reviews, so I just approve the PR.

fw/ss_skb.h
s0nx: I would leave it as it is for now. You can extend it as part of #1845 later.

fw/http_msg.c
s0nx: In case, when there is no body, we don't really want to split anything. We just zap the linear part and pretend that this SKB is kind of empty and start adding the fragments.

PR #1846, opened 2023-03-24
Script added

pale-emperor:
Step by step described script which builds kernel or just pull it from repository - based on patch file version

### DIFF
--- a/scripts/kernel_installer.py
+++ b/scripts/kernel_installer.py
@@ -16,6 +16,19 @@ if os.geteuid() != 0:
     print("You aren't root")
     os._exit(1)
 
+def add_repo(file_path: str, string_to_add: str) -> None:
+    # Add tempesta repo if not present
+    with open(file_path, "r+") as file:
+        lines = file.readlines()
+        if string_to_add + "\n" not in lines:
+            file.write("\n" + string_to_add)
+    # Also add GPG key
+    print('Add GPG key')
+    os.system('apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 36E626F4BEBEB6A8')
+
+add_repo('/etc/apt/sources.list',
+                   'deb http://tempesta-vm.cloud.cherryservers.net:8081/repository/tempesta/ focal main')
+
 script_path = os.path.dirname(os.path.realpath(__file__))
 tempesta_path = os.path.dirname(script_path)
 
@@ -56,6 +69,8 @@ else:
             print('Need to get another version of kernel')
             print(f'{current_patch_hash} != {current_kernel_hash}')
             
+            
+
             # Step 1: Download sources
             print('Step 1: Downloading 5.10.35 sources...')
             url = 'https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/snapshot/linux-5.10.35.tar.gz'
--- a/scripts/kernel_installer.py
+++ b/scripts/kernel_installer.py
@@ -66,7 +66,7 @@ else:
 
         # Else - Run local build
         else:
-            print('Need to get another version of kernel')
+            print('Need build new kernel image')
             print(f'{current_patch_hash} != {current_kernel_hash}')
             
             
--- a/scripts/kernel_installer.py
+++ b/scripts/kernel_installer.py
@@ -1,6 +1,6 @@
 #! /usr/bin/python3
 
-import subprocess, requests, tarfile, shutil, os, glob, argparse
+import subprocess, requests, tarfile, shutil, os, glob, argparse, sys
 
 parser = argparse.ArgumentParser(description='Kernel install script for tempesta-fw')
 parser.add_argument('--reboot', type=bool, default=True,
@@ -45,7 +45,7 @@ current_kernel_hash = os.uname().release.strip().split('-')[-1][:7]
 
 if current_patch_hash == current_kernel_hash:
     print('Same kernel version installed')
-
+    sys.exit(0)
 else:
     print('New kernel version found')
     # Step 0: Check if kernel already present in repo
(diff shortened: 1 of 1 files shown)


scripts/kernel_installer.py
```diff
@@ -0,0 +1,170 @@
...
+            shutil.copyfile(f'{tempesta_path}/linux-5.10.35.patch', 'linux-5.10.35/linux-5.10.35.patch')
+            with subprocess.Popen(["patch -p1 < linux-5.10.35.patch"],
+                                  stdout=subprocess.PIPE,
+                                  shell=True,
+                                  cwd=f"linux-5.10.35") as p:
+                output, err = p.communicate()
+                patch_out = output.decode("utf-8").strip()
+
+            # Step 4: config
+            if os.path.exists('.config'):
```
s0nx, line 105: When downloading vanilla sources, there won't be a `.config` in an archive. `.config` is basically a result of kernel configuration, e.g. someone called `make menuconfig`. Which `.config` are you referring to here?

scripts/kernel_installer.py
```diff
@@ -0,0 +1,170 @@
...
+                print('Step 4: Predefined .config found...')
+                subprocess.Popen(['cp', '.config', 'linux-5.10.35/config'], stdout=subprocess.PIPE)
+            else:
+                print('Step 4: Use OS .config...')
+                subprocess.Popen(['cp', f'/boot/config-{os.uname().release}', '.config'],
```
s0nx, line 110: If, for some reason, the current running kernel is much newer than the one we're trying to build, we might have problems reusing the config from `/boot`. We need to check that it's not `too new` :). 
Also, we should check `CONFIG_SECURITY_TEMPESTA` defines are present in `.config` as described [here](https://github.com/tempesta-tech/tempesta/wiki/Install-from-Sources)

scripts/kernel_installer.py
```diff
@@ -0,0 +1,170 @@
...
+            print('Step 5: Make deb-pkg...')
+            with subprocess.Popen(["make", "deb-pkg"],
+                                  stdout=subprocess.PIPE,
+                                  bufsize=1, universal_newlines=True,
+                                  cwd=f"linux-5.10.35") as p:
+                for line in p.stdout:
+                    print(line, end='') # process line here
+
+            if p.returncode != 0:
+                raise subprocess.CalledProcessError(p.returncode, p.args)
```
s0nx, line 124: If the building process has failed, we need to collect the artifacts (the console output of the part where compilation stopped)

scripts/kernel_installer.py
```diff
@@ -0,0 +1,170 @@
...
+                print('Step 5.1: Push to registry...')
+                for file in glob.glob('*.deb'):
+                    os.system(f'curl -u "{args.registry_creds}" \
+                            -H "Content-Type: multipart/form-data" --data-binary \
+                            "@./{file}" http://tempesta-vm.cloud.cherryservers.net:8081/repository/tempesta/')
+
+            # Step 6: Install and reboot
+            os.system('sudo dpkg -i *.deb')
+
+    if args.make_default:
```
s0nx, line 137: I'm not quite familiar with Debian-based distros, but on Fedora, one might just use `grubby --set-default=/boot/vmlinuz-5.10.35mykernel` to set the default kernel. Is there something similar for Ubuntu?

s0nx (changes requested):

s0nx:
@pale-emperor, the main question here: should we really have this script in the main repo? I thought it would be custom-tailored for CI, so it would be possible to collect artifacts for Jenkins and perform some other actions internally.

scripts/kernel_installer.py
```diff
@@ -0,0 +1,170 @@
...
+        else:
+            print('Need to get another version of kernel')
+            print(f'{current_patch_hash} != {current_kernel_hash}')
+            
+            
+
+            # Step 1: Download sources
+            print('Step 1: Downloading 5.10.35 sources...')
+            url = 'https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/snapshot/linux-5.10.35.tar.gz'
+            r = requests.get(url, allow_redirects=True)
```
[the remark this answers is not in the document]
pale-emperor: Very good point. I had uploaded archive directly to our nexus to be ensure archive will be reacheable

scripts/kernel_installer.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # Else - Run local build
+        else:
+            print('Need to get another version of kernel')
+            print(f'{current_patch_hash} != {current_kernel_hash}')
+            
+            
+
+            # Step 1: Download sources
+            print('Step 1: Downloading 5.10.35 sources...')
```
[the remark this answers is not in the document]
pale-emperor: I changed the code. 
It may be less readable but it looks properly right

scripts/kernel_installer.py
pale-emperor: I add that functional in case somebody can use this script to build from original non-tempesta kernel. And just if .config file is exists in current location - it will be used. 
Just in case

scripts/kernel_installer.py
```diff
@@ -0,0 +1,170 @@
...
+            if p.returncode != 0:
+                raise subprocess.CalledProcessError(p.returncode, p.args)
+
+            # Push to registry
+            if args.registry_creds is not None:
+                print('Step 5.1: Push to registry...')
+                for file in glob.glob('*.deb'):
+                    os.system(f'curl -u "{args.registry_creds}" \
+                            -H "Content-Type: multipart/form-data" --data-binary \
+                            "@./{file}" http://tempesta-vm.cloud.cherryservers.net:8081/repository/tempesta/')
```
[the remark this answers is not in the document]
pale-emperor: Replaced to arg

scripts/kernel_installer.py
pale-emperor: For example `grubby` is not default package in Ubuntu

PR #1851, opened 2023-03-28
Adjustable cache ttl via HTTP tables

dmpetroff:
[WIP]

Added new http tables action — `cache_ttl` that allows setting lifetime for cache in seconds. This is applied only when upstream hasn't send any cache-related headers.

`$cache -> 1/0` action is replaced by `disable_cache` action without any arguments. It works exactly like `$cache = 0`.

Added global option `cache_ttl` that sets default cache lifetime instead of hard-coded constant UINT_MAX.


dmpetroff:
@krizhanovsky @EvgeniiMekhanik please note that old code that handling '$cache' headers is kept almost untouched in order to not make tests migration to the new syntax painless.

fw/http_tbl.c
```diff
@@ -491,24 +491,36 @@ tfw_cfgop_http_rule(TfwCfgSpec *cs, TfwCfgEntry *e)
 		}
```
krizhanovsky, line 491: Please update the copyright year in all the touched files

fw/http_tbl.c
```diff
@@ -491,24 +491,36 @@ tfw_cfgop_http_rule(TfwCfgSpec *cs, TfwCfgEntry *e)
...
+		if (act_val_parsed != 1 && act_val_parsed != 0) {
+			T_ERR_NL("http_tbl: '$cache' action value "
+					"must 0 or 1: '%s'\n",
+					action_val);
+			return -EINVAL;
+		}
+		rule->act.type = TFW_HTTP_MATCH_ACT_FLAG;
+		rule->act.flg.set = (act_val_parsed == 0);
+		rule->act.flg.fid = TFW_HTTP_B_CHAIN_NO_CACHE;
+	}
```
krizhanovsky, line 515: We do not need legacy code. Let's just remove the code and update the tests to evaluate the new logic - we need to touch the tests in any way to test the logic

fw/http_tbl.c
```diff
@@ -491,24 +491,36 @@ tfw_cfgop_http_rule(TfwCfgSpec *cs, TfwCfgEntry *e)
...
+		rule->act.type = TFW_HTTP_MATCH_ACT_FLAG;
+		rule->act.flg.fid = TFW_HTTP_B_CHAIN_NO_CACHE;
+		if (action_val && *action_val) {
+			rule->act.flg.set = action_val[0] == '1';
 		} else {
-			T_ERR_NL("http_tbl: only '$cache' flag setting action "
-				 "supported for now: '%s'\n",
-				 action);
+			rule->act.flg.set = 1;
+		}
```
krizhanovsky, line 501: Now we have only global cache TTL setting and HTTP tables rules to disable caching. However, it seems for WordPress we need to be able to set custom TTLs, e.g. for `knowledge-book` vs blog comments vs the rest of almost static content. @pale-emperor could you please also check this comment in sense of https://github.com/tempesta-tech/tempesta-tech.com/issues/64 and https://github.com/tempesta-tech/tempesta-tech.com/issues/68 ?

pale-emperor: I already tested last commit and configure tempesta-fw as:

`cache_ttl 3600;` - on top level

And specify cache_ttl on http tables like
```
http_chain {        
	uri == "/blog/*" -> cache_ttl = 1;
        uri == "/knowledge-base/*" -> cache_ttl = 600;
       -> default;
}
```

And it works properly as expected. So i suppose everything works fine

PR #1857, opened 2023-04-06
Fix warning under heavy segmentation

const-t:
There was a case when after resizing fragment
to zero size fragment still hang in skb.

It can be reproduced under heavy segmentation
when the fragment is single fragment in skb
and response headers also ends in this fragment.

__tfw_h2_msg_shrink_frag - good candidate for moving to ss_skb.h, but it used only in one place thus moving looks redundant.

### DIFF
--- a/fw/http_msg.c
+++ b/fw/http_msg.c
@@ -1492,36 +1492,57 @@ tfw_http_msg_expand_from_pool_lc(TfwHttpTransIter *mit, TfwPool* pool,
 	return __tfw_http_msg_expand_from_pool(mit, pool, str, tfw_cstrtolower);
 }
 
-static void
-__tfw_h2_msg_shrink_frags(TfwMsgIter *it, const char *body,
-			  TfwHttpRespCleanup *cleanup)
+static inline void
+__tfw_h2_msg_move_frags(struct sk_buff *skb, int frag_idx,
+			TfwHttpRespCleanup *cleanup)
 {
-	struct skb_shared_info *si = skb_shinfo(it->skb);
-	skb_frag_t *frag = &si->frags[it->frag];
-	char* frag_addr = skb_frag_address(frag);
-	int len;
 	int i;
+	int len;
+	struct page *page;
+	struct skb_shared_info *si = skb_shinfo(skb);
 
-	/* Delete all fragments from skb. We will use this skb as skb_head. */
-	if (!body) {
-		for (i = 0; i < si->nr_frags; i++) {
-			struct page *pg = skb_frag_page(&si->frags[i]);
-
-			cleanup->pages[i] = compound_head(pg);
-		}
-		len = it->skb->data_len;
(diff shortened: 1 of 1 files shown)


const-t:
Fix for [this warning](https://github.com/tempesta-tech/tempesta-test/issues/431#issuecomment-1497821149).

EvgeniiMekhanik (approved):

fw/http_msg.c
```diff
@@ -1492,36 +1492,57 @@ tfw_http_msg_expand_from_pool_lc(TfwHttpTransIter *mit, TfwPool* pool,
...
-	/* Delete all fragments from skb. We will use this skb as skb_head. */
-	if (!body) {
-		for (i = 0; i < si->nr_frags; i++) {
-			struct page *pg = skb_frag_page(&si->frags[i]);
+	for (i = 0; i < frag_idx; i++) {
+		page = skb_frag_page(&si->frags[i]);
+		cleanup->pages[i] = compound_head(page);
+		cleanup->pages_sz++;
+		len = skb_frag_size(&si->frags[i]);
+		ss_skb_adjust_data_len(skb, -len);
```
krizhanovsky, line 1509: It'd be better to accumulate all the frag length and call only one `ss_skb_adjust_data_len()` after the loop

fw/http_msg.c
```diff
@@ -1492,36 +1492,57 @@ tfw_http_msg_expand_from_pool_lc(TfwHttpTransIter *mit, TfwPool* pool,
...
+static inline void
+__tfw_h2_msg_move_frags(struct sk_buff *skb, int frag_idx,
+			TfwHttpRespCleanup *cleanup)
 {
-	struct skb_shared_info *si = skb_shinfo(it->skb);
-	skb_frag_t *frag = &si->frags[it->frag];
-	char* frag_addr = skb_frag_address(frag);
-	int len;
 	int i;
+	int len;
```
krizhanovsky, line 1500: Why to keep 2 int variables in separate lines?  The same for `__tfw_h2_msg_rm_all_frags()`

fw/http_msg.c
```diff
@@ -1561,11 +1582,38 @@ tfw_h2_msg_cutoff_headers(TfwHttpResp *resp, TfwHttpRespCleanup* cleanup)
...
+			 */
+			if (!body || (si->nr_frags == 1 && off == end)) {
+				__tfw_h2_msg_rm_all_frags(it->skb, cleanup);
 				goto end;
+			} else if (off != begin) {
+				/*
+				 * Fragment contains headers and body.
+				 * Set beginning of frag as beginning of body.
+				 */
+
```
krizhanovsky, line 1604: extra empty line

krizhanovsky (commented):
I see that the PR is merged and it looks good for me, but could you please jsut make couple of the cleanups?

Issue #1630, opened 2022-05-24
Unify HTTP/1 `req->host` and HTTP/2 `:authority` processing


krizhanovsky:
At the moment we store URI authority part in `req->host` for HTTP/1, but we have a designated `TFW_HTTP_HDR_H2_AUTHORITY` special header for HTTP/2 for `:authority`. This complicates the logic (e.g. see https://github.com/tempesta-tech/tempesta/pull/1629#discussion_r880487824 discussion). Need to store URI authority for HTTP/1 in the special header, just as we do this for HTTP/2, and adjust all the places processing `req->host`.

No need any special testing - these changes are well covered by the existing test suite.

const-t:
Current issue may contribute to https://github.com/tempesta-tech/tempesta/issues/1630

RomanBelozerov:
If send h2 request with `host: ' (empty), we will receive response status code - 400 and `Warning: failed to parse request`. #1705 hides warning. 
But http 1.1 request will be processed successfully.

krizhanovsky:
Please also fix https://github.com/tempesta-tech/tempesta/wiki/HTTP-tables

> `host` - the host part from URI in HTTP request line, or the value of "Host" header field or value of "host" parameter from "Forwarded" header field. The Host part in URI takes priority over the Host header field value and "host" parameter from "Forwarded" header.

Which makes false statement about URI priority over `Host` header value and does not consider http2. See better description for `http_host_required` in https://github.com/tempesta-tech/tempesta/wiki/HTTP-security

Issue #1667, opened 2022-07-29
Wrong cached response with fake host


const-t:
# Scope
There is a possibility to replace original http2 response by response with the same URI from another non http2 vhost. To achieve this need to send request with specific `Host` while cache is clear.

# Testing
Can be reproduced by the next steps:
1. Clear tempesta cache.
2. curl -vk https://debian/file.html -H'Host: frontend'
3. curl -vk https://debian/file.html

On third step `file.html` will be received from `frontend` vhost, instead of `debian` vhost.

Test config:
```
listen 443 proto=h2;
listen 80;

srv_group main {
    server 127.0.0.1:8080;
}

srv_group front {
    server 127.0.0.1:8081;
}

vhost debian {
    tls_certificate /etc/tempesta/tfw-root.crt;
    tls_certificate_key /etc/tempesta/tfw-root.key;

    proxy_pass main;
}

vhost frontend {
    proxy_pass front;
}

http_chain {
  host == "frontend" -> frontend;
  host == "debian" -> debian;
-> block;
}

cache_fulfill * *;
```
Also need to implement functional test for this issue.

krizhanovsky:
Linked issue #1630

dmpetroff:
Fixed in #1862

PR #1862, opened 2023-04-07
Added small hacks to work around tests disabled by #1630

dmpetroff:
Current (27.04.2023) state of PR:
`req->host` aka `picked_authority` is selected from request data in the following priority:
- from `absoluteURI` (HTTP/1.1 and earlier versions)
- from `:authority` header (HTTP/2)
- from `Host` header (all version)

`req->host` is now used for cache key calculation.

Hard-coded checks (`check_authority_correctness()`):
- `Host` header MUST be present even when request comes in form of an absolute-URI (HTTP/1.1)
- authority information MUST be supplied (ie non-empty) for HTTP/1.1 and HTTP/2

Frang check had been update accordingly:
- `:authority == Host` (HTTP/2) or (`host_from_absolute_uri == Host`) (HTTP/1.1)
- `host_from_forwarded_header == picked_authority` (both HTTP/1.1 and HTTP/2)

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -5644,24 +5644,13 @@ check_authority_correctness(TfwHttpReq *req)
 		 * A client MUST send a Host header field in an HTTP/1.1
 		 * request even if the request-target is in the absolute-form
 		 */
-		return test_bit(TFW_HTTP_B_ABSOLUTE_URI, req->flags) ||
-			!TFW_STR_EMPTY(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST]);
-	case TFW_HTTP_VER_20: {
-		TfwStr authority, host;
-		__h2_msg_hdr_val(&req->h_tbl->tbl[TFW_HTTP_HDR_H2_AUTHORITY],
-		                 &authority);
-		__h2_msg_hdr_val(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST], &host);
-		/* no-authority case */
-		if (TFW_STR_EMPTY(&authority) && TFW_STR_EMPTY(&host))
+		if (test_bit(TFW_HTTP_B_ABSOLUTE_URI, req->flags) &&
+		    TFW_STR_EMPTY(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST]))
 			return false;
-		/* https://datatracker.ietf.org/doc/html/rfc9113#section-8.3.1
-		 * A server SHOULD treat a request as malformed if it contains
-		 * a Host header field that identifies an entity that differs
-		 * from the entity in the ":authority" pseudo-header field.
-		 */
-		return TFW_STR_EMPTY(&authority) || TFW_STR_EMPTY(&host)
-			|| tfw_strcmp(&authority, &host) == 0;
-		}
+		fallthrough;
+	case TFW_HTTP_VER_20:
+		/* HTTP/1.1 and HTTP/2 requires authority information */
+		return !TFW_STR_EMPTY(&req->host);
 	}
 	return true;
 }
--- a/fw/http.c
+++ b/fw/http.c
@@ -5644,7 +5644,7 @@ check_authority_correctness(TfwHttpReq *req)
 		 * A client MUST send a Host header field in an HTTP/1.1
 		 * request even if the request-target is in the absolute-form
 		 */
-		return req->uri_host ||
+		return test_bit(TFW_HTTP_B_ABSOLUTE_URI, req->flags) ||
 			!TFW_STR_EMPTY(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST]);
 	case TFW_HTTP_VER_20: {
 		TfwStr authority, host;
--- a/fw/http.c
+++ b/fw/http.c
@@ -5654,10 +5654,11 @@ validate_authority_rfc9113(TfwHttpReq *req)
 	 * a Host header field that identifies an entity that differs
 	 * from the entity in the ":authority" pseudo-header field.
 	 * */
-	TfwStr *authority = &req->h_tbl->tbl[TFW_HTTP_HDR_H2_AUTHORITY];
-	TfwStr *host = &req->h_tbl->tbl[TFW_HTTP_HDR_HOST];
-	return TFW_STR_EMPTY(authority) || TFW_STR_EMPTY(host)
-		|| tfw_strcmp(authority, host) == 0;
+	TfwStr authority, host;
+	__h2_msg_hdr_val(&req->h_tbl->tbl[TFW_HTTP_HDR_H2_AUTHORITY], &authority);
+	__h2_msg_hdr_val(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST], &host);
+	return TFW_STR_EMPTY(&authority) || TFW_STR_EMPTY(&host)
+		|| tfw_strcmp(&authority, &host) == 0;
 }
 
 /**
--- a/fw/http.c
+++ b/fw/http.c
@@ -5604,6 +5604,24 @@ tfw_h1_req_process(TfwStream *stream, struct sk_buff *skb)
 	return hmsib;
 }
 
+static void
+extract_req_host(TfwHttpReq *req)
+{
+	int hid = 0;
+	TfwHttpHdrTbl *ht = req->h_tbl;
+	if (TFW_MSG_H2(req)) {
+		if (!TFW_STR_EMPTY(&ht->tbl[TFW_HTTP_HDR_H2_AUTHORITY]))
+			hid = TFW_HTTP_HDR_H2_AUTHORITY;
+		else
+			hid = TFW_HTTP_HDR_HOST;
+	} else {
+		if (TFW_STR_EMPTY(&req->host))
+			hid = TFW_HTTP_HDR_HOST;
+	}
+	if (hid)
+		__h2_msg_hdr_val(&ht->tbl[hid], &req->host);
+}
+
 /**
  * @return zero on success and negative value otherwise.
  * TODO enter the function depending on current GFSM state.
@@ -5744,6 +5762,7 @@ next_msg:
 		}
 	}
 
+	extract_req_host(req);
 	/*
 	 * The message is fully parsed, the rest of the data in the
 	 * stream may represent another request or its part.
--- a/fw/http.c
+++ b/fw/http.c
@@ -756,21 +756,7 @@ do { 								\
 	if (!cl_len)
 		return TFW_BLOCK;
 
-	if (req->host.len) {
-		host = req->host;
-	} else {
-		/* Invalid id value */
-		unsigned id = TFW_HTTP_HDR_NUM;
-
-		if (!TFW_STR_EMPTY(&req->h_tbl->tbl[TFW_HTTP_HDR_H2_AUTHORITY]))
-			id = TFW_HTTP_HDR_H2_AUTHORITY;
-		else if (!TFW_STR_EMPTY(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST]))
-			id = TFW_HTTP_HDR_HOST;
-
-		if (id != TFW_HTTP_HDR_NUM)
-			tfw_http_msg_clnthdr_val(req, &req->h_tbl->tbl[id],
-						 id, &host);
-	}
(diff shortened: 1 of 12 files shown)


fw/http_match.c
```diff
@@ -231,18 +231,16 @@ match_host_forwarded(const TfwHttpReq *req, const TfwHttpMatchRule *rule)
...
 	 * on the authority special header.
 	 */
-	if (host->len > 0) {
+	if (req->uri_host && req->uri_host->len > 0) {
 		/*
 		 * RFC 7230 5.4: Host header must
 		 * be ignored when URI is absolute.
 		 */
-		return host_val_eq(host, rule);
+		return host_val_eq(req->uri_host, rule);
```
krizhanovsky, line 243: This is the only place where `uri_host` is used - why do we need it here? Why can't we just use `req->host` here?

According to the TODO comment above, this is the main point for #1630. I'd say this is exactly the place where `req->host`, picked from all URI auth, Host and :authority, should be used. Most likely some tests should be fixed along with the documentation https://github.com/tempesta-tech/tempesta/issues/1630#issuecomment-1500325853 .

We also need (and this would be good to cover with tests if we don't have such tests yet) to be able to get different matching results for `host == "foo" -> ...`, `hdr Host == "foo" ->` and `hdr :authority == "foo" -> ...` . Please test the case from https://github.com/tempesta-tech/tempesta/issues/1667 - the issue also should be fixed with this PR.

fw/http.c
```diff
@@ -5758,6 +5733,28 @@ tfw_h1_req_process(TfwStream *stream, struct sk_buff *skb)
...
+			hid = TFW_HTTP_HDR_HOST;
+		__h2_msg_hdr_val(&hdrs[hid], &req->host);
+	} else {
+		if (TFW_STR_EMPTY(&req->host)) {
+			hid = TFW_HTTP_HDR_HOST;
+			tfw_http_msg_clnthdr_val(req, &hdrs[TFW_HTTP_HDR_HOST],
+						 TFW_HTTP_HDR_HOST,
+						 &req->host);
+		}
+	}
```
krizhanovsky, line 5755: So the authority part of URI takes priority over `Host` header in HTTP/1? Could you please cite RFC which describes this rule as well as for HTTP/2 precedence of `:authority` over the `Host` header.

We discussed that we should get rid of `http_host_required` frang limit and make the rule default. This could mean that we just should call `frang_http_host_check()` instead of this function and set `req->host` there.

We actually do not work without the limit: e.g. `tfw_h2_adjust_req()` just returns error on `(!auth && !host)` - this code can be eliminated. Probably there are other similar places.

Once we have mandatory Host and URI authority equal the task https://github.com/tempesta-tech/tempesta/issues/1667 should be fixed, so I assign it to you as well.

fw/http.h
```diff
@@ -470,6 +473,7 @@ struct tfw_http_req_t {
 	TfwHttpCond		cond;
 	TfwMsgParseIter		pit;
 	TfwStr			userinfo;
+	TfwStr			*uri_host;
```
krizhanovsky, line 476: Don't we have an alignment hole due to this member placement?

krizhanovsky (changes requested):
There are several questions and comments, which probably are good to be discussed

fw/http.h
dmpetroff: Nope. TfwStr starts itself with a pointer, so no alignment gaps here.

fw/http.c
dmpetroff: >We discussed that we should get rid of http_host_required frang limit and make the rule default

As I can see from the code, this rule also enforces equality of passed authorities which significantly changes tempesta-fw behavior. In particular we have tests that send different values in Host and forwarded headers and expect match on the Forwarded host value. Enforcing frang rule would break such tests.
At the same time new code doesn't introduce any pitfalls on mismatching absolute-uri/host/authority/forwarded headers, so maybe we should make a separate issue on that?
Also note that updated RFC explicitly mentioning that proxy which accepts absolute-uri MUST replace the Host headers if they don't match in the incoming request: ttps://www.rfc-editor.org/rfc/rfc7230#section-5.4
>When a proxy receives a request with an absolute-form of request-target, the proxy MUST ignore the received Host header field (if any) and instead replace it with the host information of the request-target.  A proxy that forwards such a request MUST generate a new Host field-value based on the received request-target rather than forward the received Host field-value.

fw/http.h
dmpetroff: Nope. TfwStr has 8 byte alignment

fw/http.c
```diff
@@ -5758,6 +5730,66 @@ tfw_h1_req_process(TfwStream *stream, struct sk_buff *skb)
...
+	 * a Host header field that identifies an entity that differs
+	 * from the entity in the ":authority" pseudo-header field.
+	 * */
+	TfwStr authority, host;
+	__h2_msg_hdr_val(&req->h_tbl->tbl[TFW_HTTP_HDR_H2_AUTHORITY], &authority);
+	__h2_msg_hdr_val(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST], &host);
+	return TFW_STR_EMPTY(&authority) || TFW_STR_EMPTY(&host)
+		|| tfw_strcmp(&authority, &host) == 0;
+}
+
```
krizhanovsky, line 5792: Let's do these validations in `extract_req_host()`:
1. `tfw_http_req_process()` is already too big
2. there is no sense to check the request version twice and spread the logic among many places (we already have the frang check)

fw/http.c
```diff
@@ -5758,6 +5730,66 @@ tfw_h1_req_process(TfwStream *stream, struct sk_buff *skb)
...
+	/* https://datatracker.ietf.org/doc/html/rfc9113#section-8.3.1
+	 * A server SHOULD treat a request as malformed if it contains
+	 * a Host header field that identifies an entity that differs
+	 * from the entity in the ":authority" pseudo-header field.
+	 * */
+	TfwStr authority, host;
+	__h2_msg_hdr_val(&req->h_tbl->tbl[TFW_HTTP_HDR_H2_AUTHORITY], &authority);
+	__h2_msg_hdr_val(&req->h_tbl->tbl[TFW_HTTP_HDR_HOST], &host);
+	return TFW_STR_EMPTY(&authority) || TFW_STR_EMPTY(&host)
+		|| tfw_strcmp(&authority, &host) == 0;
```
krizhanovsky, line 5790: We already have the check in `frang_assert_host_header()`, so we can eliminate it.

fw/http_match.c
```diff
@@ -228,32 +228,23 @@ match_host_forwarded(const TfwHttpReq *req, const TfwHttpMatchRule *rule)
...
 		return true;
-
-	if (match_host_forwarded(req, rule))
-		return true;
-
-	return false;
+	/* Detect if we had an absolute URI which MUST override any other
+	 * headers according to
+	 * RFC 7230 5.4: Host header must
+	 * be ignored when URI is absolute.
```
krizhanovsky, line 242:
```suggestion
	 * RFC 7230 5.4: Host header must be ignored when URI is absolute.
```

fw/http.h
krizhanovsky, line 476: This new member is used in boolean expressions only, so we should just use `flags` from `TFW_HTTP_MSG_COMMON` and save the 8 bytes

krizhanovsky (changes requested):
There are still things to fix and improve

const-t (changes requested):
Please fix using `TFW_HTTP_HDR_HOST` in cache.c

fw/http_limits.c
```diff
@@ -639,72 +639,55 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+			frang_msg("Request :authority differs from Host",
 				  &FRANG_ACC2CLI(ra)->addr, "\n");
-			return TFW_BLOCK;
-		}
-		break;
-	/*
-	 * In http/1.1 host header and authority defined in URI must always be
-	 * identical.
-	 */
+                        return TFW_BLOCK;
```
krizhanovsky, line 665: `if` has no braces, so we never reach `fallthrough`. I think it was a bad idea to missallign braces just to make `authority` and `host` variables local.

fw/http.c
krizhanovsky: This comment is still not addressed

krizhanovsky (changes requested):
There are still couple of not addressed comments and couple of a new comments to fix.

The logic looks good for me and I'd vote to fix it in the wiki (@const-t approve is required though).

fw/http.c
```diff
@@ -5897,6 +5927,12 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 		}
 	}
 
+	extract_req_host(req);
+	if (!check_authority_correctness(req))
+	{
+		tfw_http_req_parse_drop(req, 400, "Invalid authority");
```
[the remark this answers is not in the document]
dmpetroff: But haven't we agreed on some hard-coded rules that satisfy RFC "MUST" requirements? Also wording "treat request as malformed" assumes that response should be "400 bad request", not "403 forbidden"

fw/http.c
dmpetroff: This particular function transforms H2 into H1 and H2 cannot have absolute URI, so relying on authority/host is completely fine. And as these headers may change after `tfw_h2_req_set_loc_hdrs()` invocation, this isn't a bad idea to leave this code as is.

fw/http_limits.c
```diff
@@ -638,73 +638,56 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+		 * Also this MUST be removed after #1870 is complete*/
+		if (test_bit(TFW_HTTP_B_ABSOLUTE_URI, req->flags)) {
+			TfwStr host;
+			tfw_http_msg_clnthdr_val(req,
+						&req->h_tbl->tbl[TFW_HTTP_HDR_HOST],
+						TFW_HTTP_HDR_HOST, &host);
+			if (tfw_strcmp(&req->host, &host) != 0) {
+				frang_msg("Request host from absolute URI differs"
+					  " from Host header",
+					&FRANG_ACC2CLI(ra)->addr, "\n");
```
const-t, line 688:
```suggestion
				frang_msg("Request host from absolute URI differs"
					  " from Host header",
					  &FRANG_ACC2CLI(ra)->addr, "\n");
```

fw/http.c
krizhanovsky: I believe `tfw_h2_req_set_loc_hdrs()` doesn't impact the comment because it is about the code:
```C
bool auth = !TFW_STR_EMPTY(&ht->tbl[TFW_HTTP_HDR_H2_AUTHORITY]);
bool host = !TFW_STR_EMPTY(&ht->tbl[TFW_HTTP_HDR_HOST]);
....
if (!auth && !host) {
                T_WARN("Cant convert h2 request to http/1.1: no authority "                    
                       "found\n");                                                             
                return -EINVAL;                                                                
        }     
```
which is before `tfw_h2_req_set_loc_hdrs()`. Once we checked the headers in `extract_req_host()` there is no sense to check them again

krizhanovsky (commented):
Just couple of unresolved comments from the previous review.

fw/http_limits.c
```diff
@@ -638,73 +638,56 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
-		else if (frang_get_host_forwarded(req, &fwd_trim, &fwd_name)
-			 && frang_assert_host_header(&prim_trim, &fwd_trim)) {
-			frang_msg("Request authority in URI differs from forwarded",
-				  &FRANG_ACC2CLI(ra)->addr, "\n");
-			return TFW_BLOCK;
+		/* This is pure HTTP/1.1 check, that would never trigger for
+		 * HTTP/2 because it cannot have an absolute URI.
+		 * Also this MUST be removed after #1870 is complete*/
+		if (test_bit(TFW_HTTP_B_ABSOLUTE_URI, req->flags)) {
+			TfwStr host;
```
const-t, line 681: Please add new line after `host` declaration.

const-t (approved):
LGTM

krizhanovsky (commented):
Just a minor cleanup

krizhanovsky (approved):
LGTM

PR #1865, opened 2023-04-14
Cleanups and style fix

const-t:


fw/http_msg.c
```diff
@@ -1496,31 +1496,29 @@ static inline void
...
 	struct skb_shared_info *si = skb_shinfo(skb);
 
 	for (i = 0; i < frag_idx; i++) {
 		page = skb_frag_page(&si->frags[i]);
 		cleanup->pages[i] = compound_head(page);
 		cleanup->pages_sz++;
-		len = skb_frag_size(&si->frags[i]);
-		ss_skb_adjust_data_len(skb, -len);
-		si->nr_frags--;
+		len += skb_frag_size(&si->frags[i]);
```
keshonok, line 1507: With this change, the `len` variable should be initialized to zero either when it's declared or in the `for` loop initialization, the same way as `i` is initialized. The latter would seem more logical for `len` initialization.

krizhanovsky (approved):
LGTM after the Alexey's comment fix

PR #1875, opened 2023-05-05
Added restriction to specify DEBUG < 1

const-t:
DEBUG < 1 can cause problems. User can set DEBUG=0 thinking it will disable debugging, but in some places we have ifdef DEBUG that will be TRUE even for DEBUG=0 and project will be built with "partial" debugging, that lead to undefined behaviour.


krizhanovsky (changes requested):
For some reason this doesn't work for my Ubuntu 22 (GNU Make 4.3)
```
# make -j clean all
Makefile:21: *** DEBUG must be greater than 0.  Stop.
```

PR #1876, opened 2023-05-10
Fixed connection leak on pipelined upgrade request

const-t:
Because we don't have any clear requirements how to process pipelined requests during upgrade, decided deny all pipelined requests at such moment.

Fixed WS connection leak on closing client connection during upgrade request. Problem details: While backend processes WS upgrade request client could reset its connection that leads to leak WS server connection, when tempesta receive success upgrade response. This appeared because we `get` a reference to the connection during upgrade, but we didn't `put` the connection if client already disconnected at this time.

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -2591,6 +2591,20 @@ tfw_http_conn_msg_alloc(TfwConn *conn, TfwStream *stream)
 	if (unlikely(!hm))
 		return NULL;
 
+	if (TFW_FSM_TYPE(conn->proto.type) != TFW_FSM_H2 && type & Conn_Clnt) {
+		TfwHttpReq *req_prev;
+		TfwCliConn *cli_conn = (TfwCliConn *)conn;
+		struct list_head *seq_queue = &cli_conn->seq_queue;
+
+		req_prev = list_empty(seq_queue) ? NULL :
+			list_last_entry(seq_queue, TfwHttpReq, msg.seq_list);
+		if (req_prev && test_bit(TFW_HTTP_B_UPGRADE_WEBSOCKET,
+					 req_prev->flags))
+		{
+			goto clean;
+		}
+	}
+
 	hm->conn = conn;
 	tfw_connection_get(conn);
 	hm->stream = stream;
@@ -4143,6 +4157,12 @@ tfw_http_resp_fwd(TfwHttpResp *resp)
 		T_DBG2("%s: The client was disconnected, drop resp and req: "
 		       "conn=[%p]\n",
 			 __func__, cli_conn);
+		/*
+		 * Put the websocket server connection when client connection is
+		 * lost after successful upgrade request.
+		 */
+		if (unlikely(cli_conn->proto.type & TFW_FSM_WEBSOCKET))
+			tfw_connection_put(cli_conn->pair);
 		tfw_connection_close(req->conn, true);
 		tfw_http_resp_pair_free(req);
 		TFW_INC_STAT_BH(serv.msgs_otherr);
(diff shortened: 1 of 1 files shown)


fw/http.c
```diff
@@ -2591,6 +2591,20 @@ tfw_http_conn_msg_alloc(TfwConn *conn, TfwStream *stream)
 	if (unlikely(!hm))
 		return NULL;
 
+	if (TFW_FSM_TYPE(conn->proto.type) != TFW_FSM_H2 && type & Conn_Clnt) {
+		TfwHttpReq *req_prev;
+		TfwCliConn *cli_conn = (TfwCliConn *)conn;
+		struct list_head *seq_queue = &cli_conn->seq_queue;
+
+		req_prev = list_empty(seq_queue) ? NULL :
+			list_last_entry(seq_queue, TfwHttpReq, msg.seq_list);
```
const-t, line 2600: As alternative this can be done via introducing field `flags` in `TfwCliConn` struct;

fw/http.c
```diff
@@ -5777,6 +5805,9 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 	 */
 	if (parsed < skb->len) {
 		WARN_ON_ONCE(TFW_MSG_H2(req));
+		/* Don't pipeline anything after UPGRADE request. */
+		if (test_bit(TFW_HTTP_B_CONN_UPGRADE, req->flags))
```
krizhanovsky, line 5809: Let's print a warning for this case

fw/http.c
```diff
@@ -6501,6 +6532,9 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 	 * connection and its corresponding freeing.
 	 */
 	if (websocket) {
+		/* Don't upgrade if client connection has been closed. */
+		if (test_bit(TFW_HTTP_B_REQ_DROP, hmresp->req->flags))
+			return TFW_BAD;
```
krizhanovsky, line 6537: It seems a client can close its connection right after the check and before the `tfw_http_websocket_upgrade()` at the below, so the check seems unreliable. What happens if we just remove the check and upgrade the connection? Won't it be closed on the response forwarding?

The idea is to free all resources on a connection closing event through reference counting. In my understanding we won't find the client connection freed thanks to reference counting and we can just ignore the check...

fw/http.c
```diff
@@ -4847,9 +4867,17 @@ tfw_h1_resp_adjust_fwd(TfwHttpResp *resp)
...
+
 		T_DBG2("%s: resp=[%p] dropped: client disconnected\n",
 		       __func__, resp);
 		tfw_http_resp_pair_free(req);
+		/*
+		 * Put the websocket server connection when client connection is
+		 * lost after successful upgrade request.
+		 */
+		if (unlikely(cli_conn->proto.type & TFW_FSM_WEBSOCKET))
+			tfw_connection_put(cli_conn->pair);
```
krizhanovsky, line 4880: I'd move the conditional put() to a new function, just to make the code more readable

fw/http.c
krizhanovsky: I'm OK with the check, but I think `tfw_http_conn_msg_alloc()` isn't a good place for the websocket logic - we shouldn't call the function if we already have Upgrade request. The only path for the function is `tfw_http_req_process()` -> `tfw_h1_req_process()` -> `tfw_http_msg_create_sibling()` (the 2nd call of the function is from `tfw_http_resp_process()`, so it's our of our scope)  - in this path we already have the patch on line 5808 (we create sibling from a split skb). The other call `tfw_http_conn_msg_alloc()` is from `tfw_http_msg_process_generic()` and probably this is the place for this code (it already has websocket logic BTW).

krizhanovsky (changes requested):
It seems there are couple of issues with the patch, but maybe I was just wrong in the comments and we just need to discuss them

fw/http.c
const-t: No, the path `tfw_http_req_process() -> tfw_h1_req_process() -> tfw_http_msg_create_sibling()` it's not single path, it's only for pipelined requests that came in single SKB. And yes this path must contain check, but by mistake I moved the check to response. I will add check to this path, however we can't remove check from `tfw_http_conn_msg_alloc()`, because requests can come to tempesta in separated SKBs.

fw/http.c
const-t: Done

fw/http.c
const-t: Done

fw/http.c
krizhanovsky: Then shouldn't we remove it? Or maybe I didn't understand you...

krizhanovsky (approved):
LGTM except the only questionable check, which either should be removed of just discussed

fw/http.c
const-t: Yes, i have to remove it.

Issue #1688, opened 2022-09-02
Access to any vhost is possible with the valid SNI


b3b:
Access to vhosts does not controlled properly.
Its is possible to access any vhost, even not-TLS enabled, after the connection SNI was validated.

# Testing

Test to reproduce: [tls.test_tls_cert.TlsSNIwithHttpTable](https://github.com/tempesta-tech/tempesta-test/blob/78ddd8e0a4fd770c42a1f79fecc9d7daac69a02a/tls/test_tls_cert.py#L542)
Related issue: [#tempesta-test/212](https://github.com/tempesta-tech/tempesta-test/issues/212)

krizhanovsky:
It seems the bug was introduced by https://github.com/tempesta-tech/tempesta/pull/1609 , the fix for https://github.com/tempesta-tech/tempesta/issues/1608

b3b:
More related tests, interleaving multiple hosts in multiplexed connections: [TlsSNIwithHttpTableMulti](https://github.com/tempesta-tech/tempesta-test/blob/d3e2a2088c36e07aa1d18cedde3f61f81f95bff4/tls/test_tls_cert.py#L792)

b3b:
List of related issues after tests: https://github.com/tempesta-tech/tempesta-test/issues/212#issuecomment-1239118068

krizhanovsky:
The task is very confusing by many possible combinations of SNI, SAN, HTTP table rules, Host header value, authority and so on. There were many verbal and Slack communications with some examples. Please update https://github.com/tempesta-tech/tempesta/wiki/Tempesta-TLS#avoiding-virtual-host-confusion and https://github.com/tempesta-tech/tempesta/wiki/HTTP-security for `http_strict_host_checking` by description of how the matching (in `tfw_tls_sni()`) and verification work along with configuration examples.

This is required to finally finish with #1877 review

dmpetroff:
@krizhanovsky I've posted an example where HTTP tables do not work. As long as you can set exact server name in vhost directive, that solution could work more or less acceptable. Anyway, I offer you to look at #1878 where that solution had been implemented to see if it have any flaws.

krizhanovsky:
Closed in https://github.com/tempesta-tech/tempesta/pull/1878

PR #1877, opened 2023-05-10
Validate SNI against authority information from HTTP request

dmpetroff:
- server name is stored inside statically allocated buffer up to 64 bytes
- if server name is longer than 64 bytes, it is allocated dynamically with kmalloc
- if allocation fails, no validation is performed for requests in that connection. This opens up exploit for host names larger than 1 page (according to kernel manuals, kmalloc may fail if more than 1 page of memory is requested)

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -5624,31 +5624,9 @@ __extract_request_authority(TfwHttpReq *req)
 	}
 }
 
-static char*
-__get_sni_from_tls_ctx(TlsCtx *tctx)
-{
-	return likely(tctx->sni_len <= sizeof(tctx->sni_buf))
-		? tctx->sni_buf
-		: tctx->sni;
-}
-
 static bool
 __check_authority_correctness(TfwHttpReq *req)
 {
-	int conn_type = TFW_CONN_TYPE(req->conn)
-			& (TFW_FSM_HTTPS | TFW_FSM_WEBSOCKET);
-	if (conn_type == TFW_FSM_HTTPS) {
-		TlsCtx *tctx = tfw_tls_context(req->conn);
-
-		if (likely(tctx->sni_len != 0) &&
-		    unlikely(!tfw_str_eq_cstr(&req->host,
-		                              __get_sni_from_tls_ctx(tctx),
-		                              tctx->sni_len,
-		                              TFW_STR_EQ_CASEI))) {
-			return false;
-		}
-	}
-
 	switch (req->version) {
 	case TFW_HTTP_VER_11:
 		/* https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2.2
--- a/fw/http.c
+++ b/fw/http.c
@@ -5643,7 +5643,8 @@ __check_authority_correctness(TfwHttpReq *req)
 		if (likely(tctx->sni_len != 0) &&
 		    unlikely(!tfw_str_eq_cstr(&req->host,
 		                              __get_sni_from_tls_ctx(tctx),
-					      tctx->sni_len, 0))) {
+		                              tctx->sni_len,
+		                              TFW_STR_EQ_CASEI))) {
 			return false;
 		}
 	}
--- a/fw/http.c
+++ b/fw/http.c
@@ -5627,6 +5627,18 @@ __extract_request_authority(TfwHttpReq *req)
 static bool
 __check_authority_correctness(TfwHttpReq *req)
 {
+	int conn_type = TFW_CONN_TYPE(req->conn)
+			& (TFW_FSM_HTTPS | TFW_FSM_WEBSOCKET);
+	if (conn_type == TFW_FSM_HTTPS) {
+		TlsCtx *tctx = tfw_tls_context(req->conn);
+
+		if (likely(tctx->sni != NULL) &&
+		    unlikely(!tfw_str_eq_cstr(&req->host,
+		                              tctx->sni, tctx->sni_len, 0))) {
+			return false;
+		}
+	}
+
 	switch (req->version) {
 	case TFW_HTTP_VER_11:
 		/* https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2.2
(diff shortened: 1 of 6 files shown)


fw/tls.c
```diff
@@ -820,6 +822,22 @@ tfw_tls_sni(TlsCtx *ctx, const unsigned char *data, size_t len)
...
 			return -ENOENT;
 		}
+
+		/* Save server name for later validation with authority from requests */
+		if (likely(len <= sizeof(ctx->sni_buf))) {
+			ctx->sni_len = len;
+			ctx->sni = ctx->sni_buf;
+			memcpy_fast(ctx->sni_buf, data, len);
+		} else {
+			ctx->sni = kmalloc(len, GFP_KERNEL | GFP_NOWAIT);
```
krizhanovsky, line 832: This code is called from softirq, which can not sleep
```suggestion
			ctx->sni = kmalloc(len, GFP_ATOMIC);
```

fw/tls.c
```diff
@@ -820,6 +822,22 @@ tfw_tls_sni(TlsCtx *ctx, const unsigned char *data, size_t len)
...
+			if (unlikely(ctx->sni == NULL)) {
+				/* this is non-fatal, just warn about it */
+				SNI_WARN("Unable to allocate %zu bytes for server_name", len);
```
krizhanovsky, line 835: That's not practical to go with host names larger than 4KB, so let's return an error code and close the connection

fw/tls.c
```diff
@@ -820,6 +822,22 @@ tfw_tls_sni(TlsCtx *ctx, const unsigned char *data, size_t len)
 				 PR_TFW_STR(&srv_name));
 			return -ENOENT;
 		}
+
+		/* Save server name for later validation with authority from requests */
```
krizhanovsky, line 826: In the code above, if we have `vhost != NULL`, then the SNI strictly matches the host name or it is SAN and its suffix matches the host name. So why it's not enough just to reference `vhost` and maybe introduce a flag whether we had SAN request?

TBD if our configuration allows any SNI (e.g. we matched the default vhost), then should we enforce the SNI validation against HTTP authority?

fw/http.c
```diff
@@ -5627,6 +5627,18 @@ __extract_request_authority(TfwHttpReq *req)
 static bool
 __check_authority_correctness(TfwHttpReq *req)
 {
+	int conn_type = TFW_CONN_TYPE(req->conn)
+			& (TFW_FSM_HTTPS | TFW_FSM_WEBSOCKET);
+	if (conn_type == TFW_FSM_HTTPS) {
```
krizhanovsky, line 5632:
```suggestion
	if (TFW_CONN_TYPE(req->conn) & TFW_FSM_HTTPS) {
```
Isn't this the same?

tls/ttls.h
```diff
@@ -560,6 +565,8 @@ typedef struct ttls_context {
 	unsigned int		nb_zero;
 	int			client_auth;
 	char			*hostname;
+	char			*sni;
+	char			sni_buf[64];
```
krizhanovsky, line 569: These could be union at least (we can determine whether we need to free SNI by `sni_len`)

fw/tls.c
krizhanovsky: Nobody frees the `sni`

fw/http.c
```diff
@@ -5627,6 +5627,18 @@ __extract_request_authority(TfwHttpReq *req)
...
+		TlsCtx *tctx = tfw_tls_context(req->conn);
+
+		if (likely(tctx->sni != NULL) &&
+		    unlikely(!tfw_str_eq_cstr(&req->host,
+		                              tctx->sni, tctx->sni_len, 0))) {
```
krizhanovsky, line 5637: Probably just to add a pointer to vhost to `TlsCtx` (it could be `void *` to not to mess up the TLS layer with HTTP) and compare `vhost->name` with `req->host` here? Also we can replace this check just with comparison of vhost pointers after HTTP tables call in `tfw_http_req_process()`. However, the host comparison shouldn't be heavy though....

krizhanovsky (changes requested):
There are still plenty of things to be discussed, so I'd love to read review comments and proposals about the design and security implications from @const-t . For example, #1688 requires to deny accessing plain HTTP hosts from a TLS connection - why if the hosts aren't secure and plain HTTP might mean "any SNI"?

Also it seems other points from #1688 are to be implemented, e.g. case insensitive SNI matching.

fw/http.c
dmpetroff: Initial code requires websocket bit to be off. From what I know websocket connection starts from a regular HTTP request.

fw/tls.c
dmpetroff: It is freed from `tfw_tls_conn_dtor()`

fw/tls.c
dmpetroff: >So why it's not enough just to reference vhost and maybe introduce a flag whether we had SAN request?

Then we will need to validate authority against certificate CN/SAN for each subsequent request in that connection, which is quite ugly and more resource-hungry than just string comparison.

Also we could make a configuration option to prevent [domain fronting attack](https://www.comparitech.com/blog/information-security/what-is-a-domain-fronting-attack/) or accept it as we currently do.

tls/ttls.h
```diff
@@ -560,6 +569,10 @@ typedef struct ttls_context {
 	unsigned int		nb_zero;
 	int			client_auth;
 	char			*hostname;
+	union {
+		char		*sni;
+		char		sni_buf[4];
```
krizhanovsky, line 574: So small buffer will lead to `kmalloc()` call for almost all requests

fw/http.c
krizhanovsky: Yep, but we never call HTTP after an HTTP connection is upgraded to websocket in `tfw_http_resp_process()`, i.e. this code is never called fro websocket connections. You can use `BUG_ON()` to make sure that there is no websocket flag

PR #1878, opened 2023-05-16
Validate vhost picked by SNI vs vhost picked by authority

dmpetroff:
This is alternative implementation of #1688 that compares vhost pointers selected by `tfw_tls_sni()` and HTTP tables mechanism.

Related tempesta-test PR: https://github.com/tempesta-tech/tempesta-test/pull/456

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -5679,17 +5679,20 @@ tfw_http_extract_request_authority(TfwHttpReq *req)
 	TfwStr *hdrs = req->h_tbl->tbl;
 
 	if (TFW_MSG_H2(req)) {
-		/* RFC 9113, sec-8.3.1:
+		/*
+		 * RFC 9113, sec-8.3.1:
 		 * The recipient of an HTTP/2 request MUST NOT use
 		 * the Host header field to determine the target
-		 * URI if ":authority" is present.*/
+		 * URI if ":authority" is present.
+		 */
 		if (!TFW_STR_EMPTY(&hdrs[TFW_HTTP_HDR_H2_AUTHORITY]))
 			hid = TFW_HTTP_HDR_H2_AUTHORITY;
 		else
 			hid = TFW_HTTP_HDR_HOST;
 		__h2_msg_hdr_val(&hdrs[hid], &req->host);
 	} else {
-		/* req->host can be only filled by HTTP/1.x parser from
+		/*
+		 * req->host can be only filled by HTTP/1.x parser from
 		 * absoluteURI, so we act as described by RFC 9112, sec-3.2.2
 		 * (https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2.2):
 		 * When an origin server receives a request with an
@@ -5697,11 +5700,10 @@ tfw_http_extract_request_authority(TfwHttpReq *req)
 		 * MUST ignore the received Host header field (if any)
 		 * and instead use the host information of the request-target.
 		 */
-		if (TFW_STR_EMPTY(&req->host)) {
+		if (TFW_STR_EMPTY(&req->host))
 			tfw_http_msg_clnthdr_val(req, &hdrs[TFW_HTTP_HDR_HOST],
 						 TFW_HTTP_HDR_HOST,
 						 &req->host);
-		}
 	}
 }
 
@@ -5858,17 +5860,19 @@ next_msg:
 
 	case TFW_PASS:
 		/*
-		 * The request is fully parsed,
-		 * fall through and process it.
+		 * The request is fully parsed, fall through and process it.
 		 */
-
 		if (WARN_ON_ONCE(!test_bit(TFW_HTTP_B_CHUNKED, req->flags)
 				 && (req->content_length != req->body.len)))
-		{
 			return TFW_BLOCK;
-		}
 	}
 
+	/*
+	 * XXX __check_authority_correctness() is called after parsing a whole
+	 * request, including body. For example: if we have a request with
+	 * invalid host/authority it will be dropped only after full parsing
+	 * while it's enough to parse only headers.
+	 */
 	if (!__check_authority_correctness(req)) {
 		tfw_http_req_parse_drop(req, 400, "Invalid authority");
 		return TFW_BLOCK;
--- a/fw/http.c
+++ b/fw/http.c
@@ -5672,8 +5672,8 @@ tfw_h1_req_process(TfwStream *stream, struct sk_buff *skb)
 	return hmsib;
 }
 
-static void
-__extract_request_authority(TfwHttpReq *req)
+void
+tfw_http_extract_request_authority(TfwHttpReq *req)
 {
 	int hid = 0;
 	TfwStr *hdrs = req->h_tbl->tbl;
@@ -5723,19 +5723,6 @@ __check_authority_correctness(TfwHttpReq *req)
 		return !TFW_STR_EMPTY(&req->host);
 	}
 	return true;
-}
-
-static int
-tfw_http_req_process_authority_host(TfwHttpReq *req)
-{
-	__extract_request_authority(req);
-
-	if (!__check_authority_correctness(req)) {
-		tfw_http_req_parse_drop(req, 400, "Invalid authority");
-		return TFW_BLOCK;
-	}
-
-	return 0;
 }
 
 /**
@@ -5840,6 +5827,7 @@ next_msg:
 				}
 
 				__set_bit(TFW_HTTP_B_HEADERS_PARSED, req->flags);
+				tfw_http_extract_request_authority(req);
 			}
 
 			if (tfw_h2_strm_req_is_compl(req->stream)) {
@@ -5881,8 +5869,10 @@ next_msg:
 		}
 	}
 
-	if ((r = tfw_http_req_process_authority_host(req)))
-		return r;
+	if (!__check_authority_correctness(req)) {
+		tfw_http_req_parse_drop(req, 400, "Invalid authority");
+		return TFW_BLOCK;
+	}
 
 	/*
 	 * The message is fully parsed, the rest of the data in the
--- a/fw/http_limits.c
+++ b/fw/http_limits.c
@@ -603,6 +603,9 @@ frang_http_domain_fronting_check(const TfwHttpReq *req, FrangAcc *ra)
 	if ((TFW_CONN_TYPE(req->conn) & TFW_FSM_HTTPS) == 0)
 		return TFW_PASS;
 
+	if (tfw_tls_allow_any_sni)
+		return TFW_PASS;
+
 	tctx = tfw_tls_context(req->conn);
 
 	if (tctx->vhost != req->vhost) {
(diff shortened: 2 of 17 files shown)


fw/http_limits.c
```diff
@@ -738,6 +738,38 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+			return TFW_PASS;
+
+		tls_name = tctx->vhost ? tctx->vhost->name : null_name;
+		req_name = req->vhost ? req->vhost->name : null_name;
+		frang_msg("vhost by SNI doesn't match vhost"
+		          " by authority",
+		          &FRANG_ACC2CLI(ra)->addr,
+		          " ('%.*s' vs '%.*s')\n",
+		          PR_TFW_STR(&tls_name),
+		          PR_TFW_STR(&req_name));
```
krizhanovsky, line 767:
```suggestion
                frang_msg("vhost by SNI doesn't match vhost by authority",
                          &FRANG_ACC2CLI(ra)->addr, " ('%.*s' vs '%.*s')\n",
                          PR_TFW_STR(&tls_name), PR_TFW_STR(&req_name));
```

fw/http_limits.c
```diff
@@ -738,6 +738,38 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
 	return TFW_PASS;
```
krizhanovsky, line 738: please update the copyright year for the file

fw/http_limits.c
```diff
@@ -1099,6 +1131,12 @@ frang_http_req_process(FrangAcc *ra, TfwConn *conn, TfwFsmData *data,
 		{
 			T_FSM_EXIT();
 		}
+		/* Optional protection against domain fronting attacks */
+		if (!f_cfg->http_allow_domain_fronting
```
krizhanovsky, line 1135: Why do we need a new configuration option for this? I think we should just combine it with `http_strict_host_checking`. Especially that both the options are in permissive mode by default

fw/http_limits.c
```diff
@@ -738,6 +738,38 @@ frang_http_host_check(const TfwHttpReq *req, FrangAcc *ra)
...
+
+	tctx = tfw_tls_context(req->conn);
+
+	if (tctx->vhost != req->vhost) {
+		BasicStr tls_name, req_name;
+		static BasicStr null_name = {"NULL", 4};
+
+		/* Special case of default vhosts */
+		if (req->vhost == NULL
+		    && tfw_vhost_is_default(tctx->vhost))
```
krizhanovsky, line 757: `tctx->vhost` can be NULL and `tfw_vhost_is_default()` will crash. @pale-emperor please make sure in https://github.com/tempesta-tech/tempesta-test/issues/212 that there is a test, which requires SNI to match to authority host name, but an actual request doesn't lead to getting a `vhost` in the function.

tls/ttls.h
```diff
@@ -37,6 +37,7 @@
 #include "x509_crl.h"
 #include "dhm.h"
 #include "ecdh.h"
+#include "fw/http_types.h"
```
krizhanovsky, line 40: Tempesta TLS is going to be moved to upsream and it must not know about the HTTP layer.

tls/ttls.h
```diff
@@ -560,6 +562,7 @@ typedef struct ttls_context {
 	unsigned int		nb_zero;
 	int			client_auth;
 	char			*hostname;
+	TfwVhost		*vhost;
```
krizhanovsky, line 565: Please use `void *` here

krizhanovsky (changes requested):
It seems there are several minor issues, but more importantly it that it seems not whole https://github.com/tempesta-tech/tempesta/issues/1688 is addressed in the pull request. However, I think the vhost comparison way is the right one. @const-t 's review is required regarding the final security and usability of the solution

fw/http_limits.c
dmpetroff: These options are not related: http_strict_host_checking comes from "SHOULD" terms of HTTP RFCs. http_allow_domain_fronting that originates from the TLS world.

After all, I believe the more flexible tempesta is, the more use cases it may fit.

For example, domain fronting would make `cache.test_cache.TestCacheVhost` test completely useless.

fw/http_limits.c
dmpetroff: @krizhanovsky what is the semantic meaning of NULL tctx->vhost / req->vhost? I thought that tctx->vhost will always be non-NULL if `TFW_FSM_HTTPS` bit is set in the request connection type.

tls/ttls.h
dmpetroff: That's one more reason reason to compare SNI with the request authority as string: semantically this thing should not be an opaque pointer.

fw/http_limits.c
dmpetroff: In my understanding, any SNI would make it that tctx->vhost is at least set to the default vhost: https://github.com/tempesta-tech/tempesta/blob/master/fw/tls.c#L830

And I was unpleasantly surprised that TfwHttpReq doesn't follow the same scheme.

fw/http_limits.c
dmpetroff: Merged these options together

fw/http_limits.c
krizhanovsky: Probably I didn't get the question. I meant that since you set `ctx->vhost = NULL` in `tfw_tls_sni()`, you need to handle `tcts->vhost == NULL` or `tfw_vhost_is_default(tctx->vhost)` will lead to crash

dmpetroff:
Also fixed:
- case-insensitive SNI matching (converted to lowercase before all other operations)
- matching against 2nd level SAN wildcard certificates (like `*.local`)
- matching against non-wildcard SANs
- show exact SNI in "unknown domain" messages from `tfw_tls_sni` (done earlier)

Therefore this issue addresses 4 out of 5 problems mentioned in #1688.

@krizhanovsky I don't see much sense in restricting SNI to safe characters as this string is not passed to clients in any form and validation of authority from HTTP protocol works independently.

fw/vhost.c
```diff
@@ -1874,26 +1874,20 @@ tfw_vhost_add(TfwVhost *vhost)
...
 	if (!(svhm = kmalloc(n, GFP_KERNEL))) {
-		T_WARN("Cannot allocate mapping for SAN/CN %.*s -> %s\n",
-		       (int)cn->len, cn->data, hname);
+		T_WARN("Cannot allocate mapping for SAN/CN %.*s -> %.*s\n",
+		       (int)cn->len, cn->data,
+		       (int)vhost->name.len, vhost->name.data);
 		return;
 	}
 
-	svhm->vhost = __tfw_vhost_lookup(tfw_vhosts_reconfig, &ns,
```
dmpetroff, line 1890: @krizhanovsky @const-t , can I use vhost pointer instead of a name lookup? Name lookup fails to find vhost "default".

const-t: Why do we call `tfw_vhost_add_sni_map` for `default` vhost? As I understand `default` vhost must be used ONLY if user specified directive `allow_any_sni` and SNI doesn't match to any SAN or vhost name. If we call `tfw_vhost_add_sni_map` for `default` we assume that we can receive SNI = `default`(in this case `tfw_tls_sni` will block such handshake) or certificate's SAN can contain `default` name that strange.

fw/http_limits.c
```diff
@@ -627,6 +627,40 @@ frang_assert_host_header(const TfwStr *l, const TfwStr *r)
...
+
+	if (tctx->vhost != req->vhost) {
+		TfwVhost *tls_vhost = tctx->vhost;
+		BasicStr tls_name, req_name;
+		static BasicStr null_name = {"NULL", 4};
+
+		/* An exotic case where TLS connection hasn't assigned
+		 * any vhost to the TlsCtx */
+		if (unlikely(tls_vhost == NULL))
+			return TFW_PASS;
```
const-t, line 648: `tfw_tls_sni` has this lines:
```
/* No SNI => no vhost for later frang checks */
ctx->vhost = NULL;
```
but in this place we return `TFW_PASS` in case of `tctx->vhost` is `NULL`. This implies anyone can not send ServerName extension or send empty SNI and avoid any checks. @RomanBelozerov Do we have tests for such cases?

const-t (changes requested):

fw/http_limits.c
dmpetroff: >This implies anyone can not send ServerName extension or send empty SNI and avoid any checks.
I'd like to emphasize that SNI is not a part of server-side protection mechanism. It is required  in order to sent client certificate for SSL-enabled vhosts that share same IP.
For me such behavior is a perfect example of an idiom "all elements of an empty set are green because it has no elements". Missing SNI is always valid as long as client is ready to accept certificate from default vhost.

fw/tls.c
```diff
@@ -807,19 +810,34 @@ tfw_tls_sni(TlsCtx *ctx, const unsigned char *data, size_t len)
 			return -ENOENT;
 		}
 
+		/* Look for non-wildcard SAN */
+		if (!vhost && (vhost = tfw_vhost_lookup_sni(&srv_name)))
+		{
```
krizhanovsky, line 815:
```suggestion
		if (!vhost && (vhost = tfw_vhost_lookup_sni(&srv_name))) {
```

fw/tls.c
```diff
@@ -1014,7 +1025,7 @@ tfw_tls_cfgend(void)
 static int
 tfw_tls_start(void)
 {
-	tfw_tls.allow_any_sni = allow_any_sni_reconfig;
+	tfw_tls_allow_any_sni = allow_any_sni_reconfig;
```
const-t, line 1028: Now we have `tfw_tls_match_any_sni_to_dflt` that name do not match to implemented behaviour, I propose to rename it.

fw/tls.c
```diff
@@ -40,14 +40,12 @@
...
  *			  vhost.
  */
-static struct {
-	TlsCfg		cfg;
-	bool		allow_any_sni;
-} tfw_tls;
+/** Common tls configuration for all vhosts; */
+static TlsCfg		tfw_tls_cfg;
+/** If set, all the unknown SNI are matched to default vhost. */
+bool		    tfw_tls_allow_any_sni;
```
const-t, line 48: It's about style and up to you, but maybe we could introduce `getter` for `tfw_tls.allow_any_sni` and don't "globalize" the `allow_any_sni` var?

fw/http.c
```diff
@@ -5824,8 +5812,10 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 		}
 	}
 
-	if ((r = tfw_http_req_process_authority_host(req)))
-		return r;
+	if (!__check_authority_correctness(req)) {
```
const-t, line 5815: I must pay your attention at this place. `__check_authority_correctness` will be called after parsing whole request, including body. For example: If we have request with invalid host/authority it will be dropped only after full parsing, however enough to parse only headers.

const-t (approved):
LGTM. But I have few comments.

fw/tls.c
dmpetroff: I thought getters are not an accepted practice on project. If you don't mind, I'd keep things as it is.

krizhanovsky (approved):
LGTM

PR #1889, opened 2023-06-01
Fixed bug when headers and body after our modifications became shuffled

const-t:
This happened when we added data to SKB with linear data, but don't moved linear part to paged fragment. We must move linear data to paged, because we insert data only to paged fragments and in such cases inserted data will be placed after linear part, however we expect that data will be inserted before linear data.

For instance: HTTP2 response. After cutting headers we got SKB with body in linear data and paged data. When we try to add headers to the SKB we insert headers between body from linear data and body from paged data.


EvgeniiMekhanik (approved):

krizhanovsky (approved):
LGTM

Issue #1736, opened 2022-10-28
Forwarding http2 splitted cookie header for


const-t:
# Scope

We allow splitted cookie header for http2, but we MUST not send it to http1 server. See RFC 9113 8.2.3 and RFC 6265 5.4.

> If there are multiple Cookie header fields after decompression, these MUST be concatenated into a single octet string using the two-octet delimiter of 0x3b, 0x20 (the ASCII string "; ")

# Testing
Need to develop functional test which will verify that splitted cookie from HTTP2 client forwards to HTTP1 as single header.

dmpetroff:
Fixed in #1895

PR #1895, opened 2023-06-07
Merge multiple cookie headers coming from HTTP/2 client into a single Cookie header for HTTP/1.1 backend

dmpetroff:
Also fix "strange" error codes coming from bitwise OR'ing or Exx error codes.

This PR is passing `http2_general.test_h2_headers.TestSplitCookies.test_split_cookies` test. I've also checked that calculated HTTP/1.1 header size (`h1_hdrs_sz`) is equal to actual  header length intercepted by wireshark.

PR with enabling tests: https://github.com/tempesta-tech/tempesta-test/pull/469

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -3625,6 +3625,43 @@ tfw_h2_req_set_loc_hdrs(TfwHttpReq *req)
 	return 0;
 }
 
+/** Fuse multiple cookie headers into one.
+ * Works only with TFW_STR_DUP strings. */
+static int
+write_merged_cookie_headers(TfwStr *hdr, TfwMsgIter *it)
+{
+	int r = 0;
+	static const DEFINE_TFW_STR(h_cookie, "cookie" S_DLM);
+	static const DEFINE_TFW_STR(val_dlm, "; ");
+	static const DEFINE_TFW_STR(crlf, S_CRLF);
+	TfwStr *dup, *dup_end;
+	const TfwStr *cookie_dlm = &h_cookie;
+
+	TFW_STR_FOR_EACH_DUP(dup, hdr, dup_end) {
+		TfwStr *chunk, *chunk_end, hval = {};
+
+		if (unlikely(TFW_STR_PLAIN(dup))) {
+			return -EINVAL;
+		}
+
+		hval.chunks = dup->chunks;
+		hval.nchunks = dup->nchunks;
+		hval.len = dup->len;
+		TFW_STR_FOR_EACH_CHUNK(chunk, dup, chunk_end) {
+			if (chunk->flags & TFW_STR_HDR_VALUE)
+				break;
+			hval.chunks++;
+			hval.nchunks--;
+			hval.len -= chunk->len;
+		}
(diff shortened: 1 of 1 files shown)


const-t (approved):
LGTM with little cleanup.

fw/http.c
```diff
@@ -3624,6 +3624,43 @@ tfw_h2_req_set_loc_hdrs(TfwHttpReq *req)
...
+	static const DEFINE_TFW_STR(crlf, S_CRLF);
+	TfwStr *dup, *dup_end;
+	const TfwStr *cookie_dlm = &h_cookie;
+
+	TFW_STR_FOR_EACH_DUP(dup, hdr, dup_end) {
+		TfwStr *chunk, *chunk_end, hval = {};
+
+		if (unlikely(TFW_STR_PLAIN(dup))) {
+			return -EINVAL;
+		}
```
krizhanovsky, line 3644:
```suggestion
		if (unlikely(TFW_STR_PLAIN(dup)))
			return -EINVAL;
```
Please read the coding style guide

fw/http.c
```diff
@@ -3624,6 +3624,43 @@ tfw_h2_req_set_loc_hdrs(TfwHttpReq *req)
...
+static int
+write_merged_cookie_headers(TfwStr *hdr, TfwMsgIter *it)
+{
+	int r = 0;
+	static const DEFINE_TFW_STR(h_cookie, "cookie" S_DLM);
+	static const DEFINE_TFW_STR(val_dlm, "; ");
+	static const DEFINE_TFW_STR(crlf, S_CRLF);
+	TfwStr *dup, *dup_end;
+	const TfwStr *cookie_dlm = &h_cookie;
+
```
krizhanovsky, line 3638: Let's write `cookie` before the loop and we won't need change `cookie_dlm` value as well as the whole variable

fw/http.c
```diff
@@ -3839,6 +3882,12 @@ tfw_h2_adjust_req(TfwHttpReq *req)
 				continue;
 			}
 			break;
+		case TFW_HTTP_HDR_COOKIE:
+			if (TFW_STR_DUP(field)) {
+				r |= write_merged_cookie_headers(&ht->tbl[TFW_HTTP_HDR_COOKIE], &it);
+				continue;
+			}
+			break;
```
krizhanovsky, line 3890:
```suggestion
			if (!TFW_STR_DUP(field))
			        break;
			r |= write_merged_cookie_headers(&ht->tbl[TFW_HTTP_HDR_COOKIE], &it);
			continue;
```

krizhanovsky: The good point about line width limit is that it also requires to keep the code with smaller nesting (e.g. see also Code Complete book).

krizhanovsky (approved):
LGTM: I don't see any severe issues with the pull request - just please fix the coding style and let's merge

fw/http.c
dmpetroff: That would just make code a bit worse as you'll have to track if separator is needed to be written or not.

dmpetroff:
Closes #1736

PR #1907, opened 2023-06-16
Fixed kernel panic under heavy load with low mtu

const-t:
When `sk_rmem_alloc` above than `sk_rcvbuf` kernel tries to free some memory using `tcp_prune_queue()`. In this case `out_of_order_queue` is first target for collapsing, during collapsing this queue(see `tcp_collapse()`) new skb will be allocated. However, this skb doesn't have space for TCP headers, because in vanilla kernel at this stage skb can't be reused, thus skb doesn't need space for TCP headers. But in Tempesta FW we can reuse received skbs, this implies reused skb must have space for TCP headers, otherwise we will got kernel panic when `push`ing headers.


krizhanovsky (approved):
LGTM

PR #1909, opened 2023-06-21
Fix invalid return code in `tfw_tls_add_cn`

EvgeniiMekhanik:
Currently `tfw_tls_add_cn` always return T_BAD,
but it should return 0 if the wildcard CN matches
the host name.


const-t (commented):
LGTM

const-t (approved):

PR #1887, opened 2023-06-01
Fix invalid stream processing

EvgeniiMekhanik:
We should split stream processing into two
parts: in first part we pass `HTTP2_HEADERS`
type and `HTTP2_F_END_HEADERS` flag to
`STREAM_SEND_PROCESS` and later in `xmit`
callback we should pass `HTTP2_F_END_STREAM`
flag and appropriate type when we finish
send all data to corrsponding stream.

PR #1913, opened 2023-06-22
Validate stream_id from HEADERS frame

dmpetroff:
This validation existed in code but it was disabled due to setting `ctx->cur_stream` prior to actually validating stream_id for the HEADERS frame.

Related tempesta-test PR: https://github.com/tempesta-tech/tempesta-test/pull/479

### DIFF
--- a/fw/http_stream.c
+++ b/fw/http_stream.c
@@ -289,21 +289,37 @@ tfw_h2_stream_fsm(TfwStream *stream, unsigned char type, unsigned char flags,
 		if (type == HTTP2_PRIORITY)
 			break;
 
-		/* At this point we're likely being in a "half-closed (remote)" state
-		 * with response being sent to the client. This means that we need
-		 * to apply following RFC 9113 rules (sec 5.1)
-		 * 
-		 * If an endpoint receives additional frames, other than
-		 * WINDOW_UPDATE, PRIORITY, or RST_STREAM, for a stream that is
-		 * in this state, it MUST respond with a stream error (Section 5.4.2)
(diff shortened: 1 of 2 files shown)


const-t (approved):
LGTM

fw/http_stream.c
```diff
@@ -289,14 +289,26 @@ tfw_h2_stream_fsm(TfwStream *stream, unsigned char type, unsigned char flags,
...
+		 * in this state, it MUST respond with a stream error (Section 5.4.2)
+		 * of type STREAM_CLOSED. */
+		if (!send) {
+			switch (type) {
+			case HTTP2_WINDOW_UPDATE:
+			case HTTP2_RST_STREAM:
+				res = STREAM_FSM_RES_IGNORE;
+				break;
+			default:
+				res = STREAM_FSM_RES_TERM_CONN;
```
EvgeniiMekhanik: As i see here
```
half-closed (remote):
If an endpoint receives additional frames, other than WINDOW_UPDATE, PRIORITY, or
RST_STREAM, for a stream that is in this state, it respond with a
of type STREAM_CLOSED
```

EvgeniiMekhanik:
I think that we should remove commits which was reverted and (may be later not in this PR) discuss about HTTP2_STREAM_REM_CLOSED state because i don't find it in the RFC.

dmpetroff:
@EvgeniiMekhanik I'm merging with squash merge which produces single commit.

EvgeniiMekhanik (approved):

dmpetroff:
Implemented in #1887

PR #1920, opened 2023-07-06
Rewrite tfw_hpack_node_compare to make it clean & fast

dmpetroff:
The idea is simple: when adding header to hpack table, we use `tfw_http_hdr_split` to preprocess header line. Apply the same transformation when we look for hpack entry to make code much more clear.

Also order of hpack items is changed:
- first we compare header name lengths
- then header+value (excluding trailing whitespace) lengths
- then "optimistic" case-insensitive comparison of header names (optimistic = we just `| 0x20` to each character)
- then case-sensitive comparison of the value

At each step of string compare we try to process 4 chars at once (actually I'm thinking that 8 chars for value could be reasonable, but let's not optimize prematurely), falling back for 1-byte processing at the end. I've intentionally removed 2-char manipulation as I'm severely doubt it has noticeable effect.


fw/t/unit/test_hpack.c
```diff
@@ -1519,401 +1519,6 @@ TEST(hpack, enc_table_index)
 #undef HDR_VALUE_3
 }
 
-TEST(hpack, enc_table_rbtree)
```
const-t, line 1522: IIUC this test just verifies correctness of our rb-tree implementation. What problems do we have with this test?

const-t (approved):
In general PR looks good. I approve it, but need review by @krizhanovsky and after it need to discuss do we need to delete `enc_table_rbtree`?

fw/t/unit/test_hpack.c
dmpetroff: RBtree structure (ie child/parent/left/right) nodes are depends on comparison function.

fw/t/unit/test_hpack.c
```diff
@@ -139,6 +139,83 @@ test_h2_hdr_name(TfwStr *hdr, TfwStr *out_name)
 	}
 }
 
+/** Helper to validate rbtree invariants */
+static int
+validator_walk(TfwHPackETbl *tree, TfwHPackNode *n,
+               int black_path_acc, int *black_path_len,
+			   const char *loc)
```
krizhanovsky, line 146: lines misalignment

krizhanovsky (changes requested):
Looks good in general and the new comparison function looks simpler and faster, but an optimization and couple of cleanups are required

**Please also backport the fix to 0.7**

fw/t/unit/test_hpack.c
dmpetroff: @krizhanovsky I've spend whole morning trying to get rid of that boolean flag in `tfw_hpack_rbtree_find`, but it just do not work. This should be done in a separate task, because `tfw_hpack_encoder_index` calls both `tfw_hpack_rbtree_find` and `tfw_hpack_add_node` in such fashion that second call of `tfw_http_hdr_split` just cannot be avoided from there without changing huge number of other functions. Overall amount of changes goes far beyond scope of this task.
By rethinking that functions we definitely can make code significantly cleaner, and even remove redundant checks (for example, it's obvious that everything below `__tfw_hpack_encode` can never get NULL in `hdr` parameter. But unfortunately amount of changes is big and it is clearly not wise to apply them here.
IMO the root of the problem is TfwStr overusage which is almost just as "`void*`" here: you can represent dozens of invalid states with it (dup, not plain, empty) and boolean flag is just a hint about how that "`void*`" should be processed.

fw/hpack.c
```diff
@@ -3213,7 +3082,10 @@ tfw_hpack_encoder_index(TfwHPackETbl *__restrict tbl,
 	    && atomic64_read(&tbl->guard) < 0)
 		goto out;
 
-	res = tfw_hpack_rbtree_find(tbl, hdr, &node, &place);
+	tfw_http_hdr_split(hdr, &h_name, &h_val, spcolon);
```
krizhanovsky, line 3085: The patch still introduces the new `tfw_http_hdr_split()` call. My intention with https://github.com/tempesta-tech/tempesta/pull/1920/#discussion_r1271363420 comment was that we're OK with to leave too many calls of  `tfw_http_hdr_split()` for #1411, but the new code (this PR) should not make the situation worse. Comment https://github.com/tempesta-tech/tempesta/pull/1920/#discussion_r1263825337 proposed to move `tfw_http_hdr_split()` here (as it is now) AND remove the call from `tfw_http_hdr_split()`, so in terms of number of `tfw_http_hdr_split()` calls we won't make the code worse.

krizhanovsky (changes requested):
Looks better for now and I'm good with the changes, but let's remove `tfw_http_hdr_split()` from `tfw_http_hdr_split()` before merge.

fw/hpack.c
dmpetroff: Well, header name and value are needed for `tfw_hpack_rbtree_find` to work properly. Previously header parsing had been performed inside each call to [tfw_hpack_node_compare](https://github.com/tempesta-tech/tempesta/blob/5a8e13bbbbc2dc97f18a8178d268c5be09cfff7d/fw/hpack.c#L2285). In this PR among other things I've replaced that hand-written parsing **per each** rbtree node comparison with `tfw_hpack_rbtree_find` that is performed once per lookup (not once per comparison during lookup).
Now in other PR (probably in related to #1411) we need to avoid multiple calls to `tfw_hpack_rbtree_find` during same header processing. IMO this should be done once in [`tfw_hpack_encode`](https://github.com/tempesta-tech/tempesta/blob/5a8e13bbbbc2dc97f18a8178d268c5be09cfff7d/fw/hpack.c#L3760) and [`tfw_hpack_transform`](https://github.com/tempesta-tech/tempesta/blob/5a8e13bbbbc2dc97f18a8178d268c5be09cfff7d/fw/hpack.c#L3771C1-L3771C20).
If you'll trace down what's going on on execution path of these function you'll notice that there's [`tfw_hpack_encoder_index`](https://github.com/tempesta-tech/tempesta/blob/5a8e13bbbbc2dc97f18a8178d268c5be09cfff7d/fw/hpack.c#L3196) that is problematic to blindly transform to accept split headers only.

krizhanovsky (approved):
OK, I added comment https://github.com/tempesta-tech/tempesta/pull/1920#discussion_r1276817864 to #1411, so let's leave it unresolved. Please update the TODO comment and let's merge. 

**Please also make a PR for release-0.7 (now we have CI protection on the branch)**

dmpetroff:
@krizhanovsky 0.7 backport: #1950

Issue #1922, opened 2023-07-06
Systemd Tempesta start


krizhanovsky:
With the last [0.7 release](https://github.com/tempesta-tech/tempesta/releases/tag/ubuntu-22%2F0.7.0) we can't start Tempesta with systemd as specified in https://github.com/tempesta-tech/tempesta/wiki/Install-from-packages#running-tempesta-fw as well as via `/lib/tempesta/scripts/tempesta.sh --start`:
```
root@ubuntu:~# /lib/tempesta/scripts/tempesta.sh --start
Starting Tempesta...
/lib/tempesta/scripts/tempesta.sh: line 73: /lib/tempesta/etc/tempesta_fw.conf: No such file or directory
...compile html templates for JS challenge
cat: /lib/tempesta/etc/tempesta_fw.conf: No such file or directory
...start Tempesta FW
done
```

krizhanovsky:
Fixed in #1924

PR #1924, opened 2023-07-06
Update tempesta_installer.sh right way

pale-emperor:


krizhanovsky (approved):
LGTM. With the new installer script `systemd` correctly manages Tempesta FW, so I think #1922 can be closed now

PR #1925, opened 2023-07-07
Frang: tls_connection_burst does not work as expected

voodam:
- Fix tls_connection_burst
- Reset connection in kernel when Frang tcp_connection_* directives triggered

### DIFF
--- a/fw/filter.c
+++ b/fw/filter.c
@@ -72,8 +72,9 @@ tfw_ipv6_hash(const struct in6_addr *addr)
 }
 
 void
-tfw_filter_block_ip(const TfwAddr *addr)
+tfw_filter_block_ip(const TfwClient *cli)
 {
+	const TfwAddr *addr = &cli->addr;
 	TfwFRule rule = {
 		.addr	= addr->sin6_addr,
 		.action	= TFW_F_DROP,
@@ -82,6 +83,8 @@ tfw_filter_block_ip(const TfwAddr *addr)
 	size_t len = sizeof(rule);
 
 	T_DBG_ADDR("filter: block", addr, TFW_NO_PORT);
+
+	tfw_cli_conn_abort_all((void *)cli);
 
 	/* TODO create records on all NUMA nodes. */
 	if (!tdb_entry_create(ip_filter_db, key, &rule, &len)) {
--- a/fw/http.c
+++ b/fw/http.c
@@ -111,6 +111,7 @@
 #include "tls.h"
 #include "apm.h"
 #include "access_log.h"
+#include "vhost.h"
 #include "websocket.h"
 
 #include "sync_socket.h"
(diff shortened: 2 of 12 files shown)


EvgeniiMekhanik:
http://tempesta-vm.cloud.cherryservers.net:8080/job/Run%20tests/ TCP segmentation successful build

fw/http.c
```diff
@@ -6883,6 +6884,21 @@ static TfwConnHooks http_conn_hooks = {
...
+	TfwVhost *dflt_vh = tfw_vhost_lookup_default();
+
+	if (tfw_blk_flags & TFW_BLK_ATT_REPLY && dflt_vh
+	    && dflt_vh->frang_gconf->ip_block)
+		T_WARN_NL("Directive 'block action' can't be set to\n"
+			  "    'attack reply' if 'ip_block' from 'frang_limits' group is 'on'\n"
+			  "    (this is misconfiguration, see the wiki).\n");
+	tfw_vhost_put(dflt_vh);
+
+	return 0;
```
EvgeniiMekhanik, line 6899: We should return -1 in case when configuration is bad i think.

fw/http_limits.c
```diff
@@ -1404,6 +1407,15 @@ frang_tls_conn_limit(FrangAcc *ra, FrangGlobCfg *conf, int hs_state)
 	switch (hs_state) {
 	case TTLS_HS_CB_FINISHED_NEW:
 		ra->history[i].tls_sess_new++;
+
+		if (conf->tls_new_conn_burst
```
EvgeniiMekhanik, line 1411: Can you explain why we move this code here?

EvgeniiMekhanik:
I don't know why but t_frang.test_ip_block.FrangIpBlockMsg.test_on  (on your brunch in tempesta-test) is always fails now. Do you have the same problem?

fw/http_limits.c
voodam: Just not to create second counter ("i" reused in for loop, so we need "j")

voodam:
> I don't know why but t_frang.test_ip_block.FrangIpBlockMsg.test_on (on your brunch in tempesta-test) is always fails now. Do you have the same problem?

No, on both branches am-1716-tls-conn-burst test is passed.

fw/filter.c
```diff
@@ -83,6 +84,8 @@ tfw_filter_block_ip(const TfwAddr *addr)
 
 	T_DBG_ADDR("filter: block", addr, TFW_NO_PORT);
 
+	tfw_cli_conn_abort_all((TfwClient *)cli);
```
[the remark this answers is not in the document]
voodam: Fixed

fw/http.c
voodam: Fixed, also I'll fix all tests where this misconfiguration is used.

const-t (approved):

EvgeniiMekhanik:
LGTM, but i think that it is good to squash two last commits

voodam:
> LGTM, but i think that it is good to squash two last commits

Yes, sure. "Fixup! ..." commits are used for auto-squashing: https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html

PR #1797, opened 2023-01-25
Safe handling of websockets drop

krizhanovsky:
Fix for #1647 (see commit messages). Now
```
for i in `seq 200`; do ./run_tests.py ws.test_ws_ping.CacheTest.test_ping_websockets; done
```
passes for local tests.

PR #1930, opened 2023-07-10
Fix incorrect server connection reference conter

### LINKED ISSUE #1910: kernel BUG at fw/connection.h:529
kernel BUG at /home/evgeny/workdir/tempesta/fw/connection.h:529!
[  187.398989] invalid opcode: 0000 [#1] SMP NOPTI
[  187.399615] CPU: 0 PID: 20252 Comm: sysctl Tainted: G           OE     5.10.35+ #233
[  187.400344] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.15.0-1 04/01/2014
[  187.402588] RIP: 0010:tfw_sock_srv_del_conns+0x195/0x1c0 [tempesta_fw]
[  187.404825] Code: 0f 0b 48 c7 c7 c0 92 d0 c0 e8 04 e2 67 ce 0f 0b 48 c7 c7 b0 92 d0 c0 e8 f6 e1 67 ce 0f 0b 48 c7 c7 a0 92 d0 c0 e8 e8 e1 67 ce <0f> 0b 48 c7 c7 90 92 d0 c0 e8 da e1 67 ce 0f 0b 48 c7 c7 80 92 d0
[  187.409946] RSP: 0018:ffffbe6500eabc18 EFLAGS: 00010202
[  187.411335] RAX: 0000000040000001 RBX: ffff9670dfa16d08 RCX: ffff9670dfa150d0
[  187.412981] RDX: ffff9670a758cd10 RSI: 00000000fffffe01 RDI: ffffffffc0cbc1b7
[  187.414265] RBP: ffffbe6500eabc48 R08: 0000000000000001 R09: 0000000000000000
[  187.415341] R10: 0000000000000001 R11: 0000000000030600 R12: ffff9670dfa16cd8
[  187.416278] R13: ffff9670a758cd20 R14: ffff9670a758cd10 R15: ffff9670dfa150a0
[  187.417065] FS:  00007f5b77368740(0000) GS:ffff9671b7c00000(0000) knlGS:0000000000000000
[  187.417869] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[  187.418367] CR2: 00005646c7c2e6d8 CR3: 0000000163ff0004 CR4: 0000000000770ef0
[  187.418935] PKRU: 55555554
[  187.419150] Call Trace:
[  187.419318]  tfw_server_destroy+0x1f/0x90 [tempesta_fw]
[  187.419650]  __tfw_sg_del_srv+0xb5/0xc0 [tempesta_fw]
[  187.419971]  tfw_sg_release_all+0xd3/0x200 [tempesta_fw]
[  187.420308]  tfw_sock_srv_stop+0x158/0x1a0 [tempesta_fw]
[  187.420645]  tfw_mods_stop+0x38/0xc0 [tempesta_fw]
[  187.420950]  tfw_ctlfn_state_io+0x239/0x400 [tempesta_fw]
[  187.421291]  proc_sys_call_handler+0x12e/0x240
[  187.421570]  proc_sys_write+0x13/0x20
[  187.421803]  new_sync_write+0x114/0x1b0
[  187.422046]  vfs_write+0x1c0/0x270
[  187.422265]  ksys_write+0x67/0xe0
[  187.422476]  __x64_sys_write+0x19/0x20
[  187.422714]  do_syscall_64+0x38/0x90
[  187.422942]  entry_SYSCALL_64_after_hwframe+0x44/0xa9

while true
do
sudo ./run_tests.py ws/
done

5-10 minutes an we have crash in ws.test_ws_ping.RestartOnUpgrade.test

EvgeniiMekhanik:
We should put server connection reference counter
before call it's destructor

Closes #1910

### DIFF
--- a/fw/websocket.c
+++ b/fw/websocket.c
@@ -150,10 +150,17 @@ tfw_ws_srv_new_steal_sk(TfwSrvConn *srv_conn)
 	 */
 	conn->sk->sk_user_data = conn;
 
-	/* Connection destructor does failover for server connections */
 	srv_conn->sk = NULL;
-	if (srv_conn->destructor)
-		srv_conn->destructor(srv_conn);
+	/*
+	 * Connection destructor does failover for server connections.
+	 * There is no problem here that we have srv_conn reference counter
+	 * equal to zero, because srv->conn->sk is NULL. If connection will
+	 * be reestablished `tfw_sock_srv_connect_try` will be called and
+	 * we set new sk and not 0 reference counter to this connection,
+	 * ohterwise ss_close finished imidiatly and we never drop this
+	 * connection again.
+	 */
+	tfw_connection_put((TfwConn *)srv_conn);
 
 	return conn;
 }


fw/websocket.c
```diff
@@ -150,6 +150,7 @@ tfw_ws_srv_new_steal_sk(TfwSrvConn *srv_conn)
 	 */
 	conn->sk->sk_user_data = conn;
 
+	tfw_connection_put((TfwConn *)srv_conn);
```
const-t, line 153: This is not looks correct for me. You do `tfw_connection_put` right before calling `destructor` manually, that looks strange. Also if `tfw_connection_put` will call `destructor` we will have problems with socket, it implies `srv_conn->sk = NULL;` must be placed before calling `tfw_connection_put`.

const-t (changes requested):

fw/websocket.c
```diff
@@ -150,10 +150,9 @@ tfw_ws_srv_new_steal_sk(TfwSrvConn *srv_conn)
 	 */
 	conn->sk->sk_user_data = conn;
 
-	/* Connection destructor does failover for server connections */
 	srv_conn->sk = NULL;
-	if (srv_conn->destructor)
-		srv_conn->destructor(srv_conn);
+	/* Connection destructor does failover for server connections */
+	tfw_connection_put((TfwConn *)srv_conn);
```
krizhanovsky, line 155: We did not decrement the reference counter because the connection is still used: the function 'steals' the high level connection decriptor `TfwSrvCon`, i.e. it leave it alive, but replaces the underlying HTTP backend server connection with a websocket. I.e. technically we still should have non-zero reference counter here. I'd expect that with the change we might have double free issue or get something like https://github.com/tempesta-tech/tempesta/issues/1647 - trying to send data proxied from a client websocket to the dead server websocket connection

krizhanovsky (changes requested):
I might be wrong, but in this case we need a very accurate analyzing of reference counting, especially how the server websocket connection reference counter interacts with the reference counter for the client connection.

This particular piece of code isn't mine, but I verified it a lot during https://github.com/tempesta-tech/tempesta/pull/1797, so maybe #1797 would be a good reading to review this change again.

fw/websocket.c
const-t: Looks like the most safe way to do this steam is: replace `tfw_connection_put` with `tfw_connection_put_to_death`
this will set refcounter to `TFW_CONN_DEATHCNT + 1` and then call the destructor. If `tfw_connection_put` will be called from `tfw_sock_srv_connect_drop` before `tfw_sock_srv_connect_try` we will have `refcounter == TFW_CONN_DEATHCNT` that good for us. Also `TFW_CONN_DEATHCNT + 1` is safe for as while reconnecting, because refcounter will be setted to 1 when connection successfully established.

const-t: No, I wrong. `tfw_sock_srv_connect_drop` will not be called, because we don't have socket at this time. `tfw_ws_srv_new_steal_sk` will be only one place where we call `tfw_connection_put` for stolen connection.

const-t (approved):
LGTM. We can put connection's counter because during reconnecting refcounter will be updated.

fw/websocket.c
EvgeniiMekhanik: We can put connection's counter because during reconnecting refcounter will be updated. Also `ss_close` immediately return since conn->sk = NULL so we newer call connection_drop again until we restore conn->sk after successful reconnect.

krizhanovsky (approved):
LGTM and thank you for the good explanation of the reference counting mechanics in the new comment

PR #1932, opened 2023-07-10
Fallback to repo download if GitHub api Ratelimit Exceeded

pale-emperor:
I made fallback for GitHub.com API Rate Limit Exceeded case.
If error appear - we should download all artifacts from our nexus

Tested - works fine


krizhanovsky (approved):
LGTM

Issue #1935, opened 2023-07-12
blinking kernel panic in `pskb_expand_head`


RomanBelozerov:
# Scope
hash - [872f58a59de6f7c9140eca0321a955d38601f4e4](https://github.com/tempesta-tech/tempesta/commit/872f58a59de6f7c9140eca0321a955d38601f4e4)
My config (Tempesta on separate VM) :
```python3
[General]
ip = 192.168.122.1
ipv6 = ::1
verbose = 6
workdir = /tmp/host
duration = 10
concurrent_connections = 10
log_file = tests_log.log
stress_threads = 2
stress_large_content_length = 65536
stress_requests_count = 100
stress_mtu = 1500
long_body_size = 500
[Client]
ip = 192.168.122.1
ipv6 = ::1
hostname = localhost
ab = ab
wrk = wrk
h2load = h2load
tls-perf = tls-perf
workdir = /tmp/client
unavaliable_timeout = 300

[Tempesta]
ip = 192.168.122.116
ipv6 = ::1
hostname = 192.168.122.116
user = root
port = 22
srcdir = /root/tempesta
workdir = /tmp/host
config = tempesta.conf
tmp_config = tempesta_tmp.conf
unavaliable_timeout = 300

[Server]
ip = 192.168.122.1
ipv6 = ::1
hostname = localhost
user = root
port = 22
nginx = nginx
workdir = /tmp/nginx
resources = /var/www/html/
aliases_interface = virbr0
aliases_base_ip = 192.168.122.1
max_workers = 16
keepalive_timeout = 60
keepalive_requests = 100
unavaliable_timeout = 300
```

```python3
[ 2994.570738] BUG: kernel NULL pointer dereference, address: 0000000000000008
[ 2994.571210] #PF: supervisor read access in kernel mode
[ 2994.571526] #PF: error_code(0x0000) - not-present page
[ 2994.571846] PGD 0 P4D 0 
[ 2994.572009] Oops: 0000 [#1] SMP NOPTI
[ 2994.572245] CPU: 2 PID: 0 Comm: swapper/2 Tainted: G           OE     5.10.35+ #3
[ 2994.572727] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS 1.15.0-1 04/01/2014
[ 2994.573301] RIP: 0010:pskb_expand_head+0x217/0x3a0
[...]

PR #1942, opened 2023-07-17
Fix GPF when skb reused

const-t:
Remove tail part of skb because because it can be reused as `skb_head` of current response. If tail part will remain in skb we will have garbage in skb and `ss_skb_linear_transform()` will be applied twice.

### DIFF
--- a/fw/sock.c
+++ b/fw/sock.c
@@ -751,6 +751,12 @@ ss_tcp_process_skb(struct sock *sk, struct sk_buff *skb, int *processed)
 		WARN_ON_ONCE(skb_has_frag_list(skb));
 		WARN_ON_ONCE(skb->sk);
 
+		/*
+		 * Some SKBs may have dev, however tempesta uses dev to store
+		 * own flags, thus clear it.
+		 */
+		skb->dev = NULL;
+
 		if (unlikely(offset >= skb->len)) {
 			offset -= skb->len;
 			__kfree_skb(skb);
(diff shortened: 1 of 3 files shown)


EvgeniiMekhanik (approved):

fw/sock.c
```diff
@@ -751,6 +751,12 @@ ss_tcp_process_skb(struct sock *sk, struct sk_buff *skb, int *processed)
 		WARN_ON_ONCE(skb_has_frag_list(skb));
 		WARN_ON_ONCE(skb->sk);
 
+		/*
+		 * Some SKBs may have dev, however tempesta uses dev to store
+		 * own flags, thus clear it.
+		 */
+		skb->dev = NULL;
```
krizhanovsky, line 758: How this initialization correlates with the `skb_tfw_is_present()` check in `ss_skb_init_for_xmit()`. It seems we nullify `dev` for all ingress `skb`s, so does `ss_skb_init_for_xmit()` ever sees non-NULL `dev`?

Also why do we need to initialize `dev` here? Our code doesn't use except the case to store the flags on transmission path and this case is handled with `ss_skb_init_for_xmit()`.

krizhanovsky (approved):
The main fix looks good for me. I still have a question about seems additional `skb->dev` nullification.

I could not reproduce any of the GPF crashes on the case with `h2load` running from a host system against a VM with running Tempesta FW and a wordpress website inside a container (seems using virtualized interface as well).

Also I assign #1935 to you since you mentioned that it's probably fixed by the PR.

PR #1943, opened 2023-07-17
TDB: Fix BUG_ON in tdb_htrie_descend

const-t:
At this place we apply `TDB_DI2O()` for both data and index nodes that leads to incorrect offset for index nodes and cause `BUG_ON()`.


const-t:
Apparently #1935 also can be closed by this PR, but need to test it.

krizhanovsky (approved):
LGTM. **Please backport the fix to 0.7 as well**

Issue #1464, opened 2020-11-17
TDB torture triggers BUG_ON() at /tempesta_db/core/htrie.c:567


i-rinat:
It's possible to trigger `BUG_ON()` in `tdb_htrie_descend()` at `tempesta_db/core/htrie.c:567` by repeatedly requesting `index.html` with different query parameters. Here is a example that I used:
```
for i in $(seq 1000000); do curl --silent http://tempesta.local/?q=$i; done | pv > /dev/null
```
Required time depends on Tempesta's configured cache size and `index.html` file size on a backend server. In my case with `index.html` of `91692` bytes and 256MB cache (default value) it takes about a minute or two to trigger the bug.

-----
git: 1351c4376e23fbc7b5db5a94d812099c7ab69480

Tempesta config:
```
cache 1;
cache_fulfill * *;
listen 80;
server 192.168.122.1:80;
```

Backend is a default Apache installation at `192.168.122.1:80`, but with larger `index.html` file.

---

Here is dmesg log:
```
[   18.410895] [tempesta fw] Warning: Listening for HTTP/1.1 protocol is compatibility mode and may lack of performance.
[   18.415377] [tempesta fw] Configuration processing is completed.
[   18.434571] [tdb] Opened table /opt/tempesta/db/filter.tdb: size=16777216 rec_size=20 base=ffff88015a400000
[   18.612928] [tdb] Opened table /opt/tempesta/db/cache.tdb: size=268435456 rec_size=0 base=ffff88014a400000
[   18.624637] [tdb] Opened table /opt/tempesta/db/sessions.tdb: size=16777216 rec_size=312 base=ffff880149400000
[   18.636570] [tdb] Opened table /opt/tempesta/db/client.tdb: size=16777216 rec_size=616 base=ffff880148400000
[   18.640048] [tempesta fw] Open listen socket on: 0.0.0.0
[   18.650063] [tempesta fw] modules are started
[   18.651903] sysctl (1750) used greatest stack depth: 24624 bytes left
[   63.201545] ------------[ cut here ]------------
[   63.202977] kernel BUG at /root/tempesta/tempesta_db/core/htrie.c:567!
[   63.204720] invalid opcode: 0000 [#1] SMP KASAN PTI
[...]

PR #1952, opened 2023-08-01
Backported #1943(#1464)

const-t:
TDB: Fix BUG_ON in tdb_htrie_descend

At this place we apply `TDB_DI2O()` for both
data and index nodes that leads to incorrect
offset for index nodes and cause `BUG_ON()`.


krizhanovsky (approved):
LGTM

Issue #1971, opened 2023-09-05
Debug printing of kernel pointers


krizhanovsky:
# Motivation

At the moment we use many `%p` and `%pK` specifiers to print kernel addresses in debug builds, which leads with the default for Ubuntu `ernel.kptr_restrict=1` sysctl to hashed pointer addresses, which is extremely misleading during debugging.

I experimented with the [kptr_restrict](https://docs.kernel.org/admin-guide/sysctl/kernel.html#kptr-restrict), but did not find a good way to make the pointers appear in dmesg in raw format.

# Scope

Replace all `%p` and `%pK` with `%px` (see https://docs.kernel.org/core-api/printk-formats.html#unmodified-addresses), but **do this for debug messages only. The rest of the code, i.e. info, warning and error messages, must use `%pK`.**

krizhanovsky:
Example commit for the proposal https://github.com/tempesta-tech/tempesta/pull/1972/commits/096bbc6b967eed09c2b65d29548bd90d7cd14a61

Issue #1970, opened 2023-09-04
RCU stall on `tfw_cache_h2_copy_chunked_body()`


krizhanovsky:
`master` as of c43e7d44711a6445540b267410b95c338ab3170e with config
```
# cat /etc/tempesta/tempesta_fw.conf 
cache 2;
cache_fulfill * *;
cache_methods GET HEAD;
cache_ttl 3600;

listen 443 proto=https;
listen 80;

srv_group default {
        server 192.168.122.1:80;
}

#srv_group auth {
#        server 127.0.0.1:8040;
#}

tls_certificate /etc/tempesta/fullchain.pem;
tls_certificate_key /etc/tempesta/privkey.pem;
req_hdr_add X-Forwarded-Proto "https";
resp_hdr_set Strict-Transport-Security "max-age=31536000; includeSubDomains";

vhost default {
        tls_match_any_server_name;
        resp_hdr_set Cache-Control;
        nonidempotent GET * *;
        nonidempotent HEAD * *;
        nonidempotent POST * *;
        proxy_pass default;
}

#vhost auth {
#        proxy_pass auth;
#}

#vhost wp-auth {
#        req_hdr_add Location "wp-admin";
#        proxy_pass auth;
#}

frang_limits {
        request_rate 200;
        http_method_override_allowed true;
        http_methods post put get;
}

cache_purge;
cache_purge_acl 192.168.122.1 127.0.0.1;

access_log on;

block_action attack reply;
block_action error reply;

#http_chain admin {
#        mark == 23  -> default;
#        -> auth;
#}

http_chain {

        # Set cache_ttl for speciefic locations
        uri == "/blog/*" -> cache_ttl = 1;
        uri == "/knowledge-base/*" -> cache_ttl = 600;

        # Set Security
#        mark == 23 -> default;
#        uri == "*/google/*" -> auth;
#        uri == "*/wp-admin/*" -> admin;
#        uri == "*/wp-login.php*" -> admin;
#        uri == "*/phpmyadmin*" -> admin;
        -> default;
}
```
Tempesta FW is running in a VM and querying the new wordpress website from an outside LXD container.

Running (**100% reproducable**):
```
# /usr/bin/wget -q -O /dev/null http://127.0.0.1/index
```
spits this kernel call trace (not a panic though, but the command hungs):

```
[ 3135.113948] rcu: INFO: rcu_sched detected stalls on CPUs/tasks:
[...]

Issue #1934, opened 2023-07-12
BUG when caching chunked response


RomanBelozerov:
# Scope
My config (Tempesta on separate VM) :
```python3
[General]
ip = 192.168.122.1
ipv6 = ::1
verbose = 6
workdir = /tmp/host
duration = 10
concurrent_connections = 10
log_file = tests_log.log
stress_threads = 2
stress_large_content_length = 65536
stress_requests_count = 100
stress_mtu = 1500
long_body_size = 500

[Client]
ip = 192.168.122.1
ipv6 = ::1
hostname = localhost
ab = ab
wrk = wrk
h2load = h2load
tls-perf = tls-perf
workdir = /tmp/client
unavaliable_timeout = 300

[Tempesta]
ip = 192.168.122.116
ipv6 = ::1
hostname = 192.168.122.116
user = root
port = 22
srcdir = /root/tempesta
workdir = /tmp/host
config = tempesta.conf
tmp_config = tempesta_tmp.conf
unavaliable_timeout = 300

[Server]
ip = 192.168.122.1
ipv6 = ::1
hostname = localhost
user = root
port = 22
nginx = nginx
workdir = /tmp/nginx
resources = /var/www/html/
aliases_interface = virbr0
aliases_base_ip = 192.168.122.1
max_workers = 16
keepalive_timeout = 60
keepalive_requests = 100
unavaliable_timeout = 300


```
```python3
[  720.093923] watchdog: BUG: soft lockup - CPU#2 stuck for 22s! [swapper/2:0]
[...]

PR #1972, opened 2023-09-06
Fix HTTP body copying

krizhanovsky:
Fix for #1970 and probably #1934.

If `curr` isn't within the current frag, then do not go to repeat on the same frag, do this only if we copied something. Otherwise we just skip the current frag (typically response headers).

If we fully copied the current frag until it's end, then this case is different from the case when the frag ends with a chunk descriptor. The second case also have 2 subcases: the next frag starts from the chunk descriptor prefix or data. The 3 cases are processed with `curr` == NULL, and `curr` == `prev_end` and an additional check for `stop` == `begin` correspondingly.

Make `ss_skb_dump()` to print real addresses (#1971).

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -1252,7 +1252,7 @@ tfw_cache_h2_copy_chunked_body(unsigned int *acc_len, char **p, TdbVRec **trec,
 	long n;
 	TfwMsgIter it;
 	TfwStr *tmp;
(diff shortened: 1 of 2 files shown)


RomanBelozerov (approved):
LGTM. This works with remote setup on CI and for me. (I checked tests from #1934).

const-t (approved):
LGTM

PR #1974, opened 2023-09-07
Backported fix of infinite loop in tfw_cache_h2_copy_chunked_body()

krizhanovsky:
Backported fix of infinite loop in tfw_cache_h2_copy_chunked_body() (#1934, #1970).

Original pull request #1972.


const-t (approved):
LGTM

PR #1977, opened 2023-09-13
Add info about failed and successful TLS handshakes to perfstat (for …

voodam:
…server and client)


fw/tls.c
```diff
@@ -107,6 +115,14 @@ tfw_tls_connection_recv(TfwConn *conn, struct sk_buff *skb)
 		return T_OK;
 	case T_OK:
 		/* A complete TLS record is received. */
+		switch (tls->conf->endpoint) {
+		case TTLS_IS_CLIENT:
+			TFW_INC_STAT_BH(clnt.tls_hs_successful);
```
const-t, line 120: Looks like stat will be incremented for each skb, it implies for single handshake stat will be increased several times.

const-t (changes requested):

fw/tls.c
voodam: Yep, my bad, fixed. Also client stats seem to be useless, deleted.

fw/http_limits.c
```diff
@@ -1467,6 +1468,13 @@ frang_tls_handler(TlsCtx *tls, int state)
 	TfwVhost *dflt_vh = tfw_vhost_lookup_default();
 	int r;
 
+	switch (state) {
+	case TTLS_HS_CB_FINISHED_NEW:
+	case TTLS_HS_CB_FINISHED_RESUMED:
+		TFW_INC_STAT_BH(serv.tls_hs_successful);
+		break;
+	}
```
const-t, line 1476: It looks correct, but I suggest to move this stat incrementation to `T_FSM_STATE(TTLS_HANDSHAKE_OVER)` in tls_srv.c

fw/tls.c
```diff
@@ -94,6 +94,11 @@ tfw_tls_connection_recv(TfwConn *conn, struct sk_buff *skb)
 	default:
 	case T_BAD:
 		r = T_BAD;
+		switch (tls->conf->endpoint) {
+		case TTLS_IS_SERVER:
+			TFW_INC_STAT_BH(serv.tls_hs_failed);
+			break;
+		}
```
const-t, line 101: Maybe we can use `if` here, instead of switch with single case?

const-t (approved):
LGTM, but I have small suggestion.

fw/http_limits.c
voodam: From tls/tls_srv.c we haven't access to fw/procfs.h, because it's separate library, as I understood.

fw/tls.c
voodam: Fixed

fw/http_limits.c
const-t: I can suggest to add separated callback for this, but it's up to you.

voodam: I understand this code smell, but it seems acceptable for me. Any option will suit me.

voodam: Added separate callback.

fw/tls.c
```diff
@@ -865,6 +867,19 @@ tfw_tls_sni(TlsCtx *ctx, const unsigned char *data, size_t len)
...
 
+static inline int
+tfw_tls_over(TlsCtx *tls, int state)
+{
+	switch (state) {
+	case TTLS_HS_CB_FINISHED_NEW:
+	case TTLS_HS_CB_FINISHED_RESUMED:
+		TFW_INC_STAT_BH(serv.tls_hs_successful);
+		break;
+	}
```
const-t, line 878: Is there a sense to do `switch` here?

voodam: Probably no, fixed.

PR #1978, opened 2023-09-18
Backported fix of CCM tls encryption and decryption

EvgeniiMekhanik:
In CCM first byte of IV is L parameter, so we
should copy fixed IV starting from the second
byte and set first byte to predefined value.


krizhanovsky (approved):
LGTM

PR #1981, opened 2023-09-25
Cut port from authority when lookup vhost

voodam:
The code smells probably, but I'm not sure how to fix it: create separate fw/str function, cut port in different place or something else.


const-t (changes requested):

const-t (approved):
LGTM

PR #1983, opened 2023-09-27
Fix ulimited body parsing

const-t:
If the response doesn't have body tempesta parses the body until the connection is closed, in this case need to update overall body size and don't make chunks.

-Added error handling to `tfw_cache_h2_copy_body()`

### DIFF
--- a/fw/apm.c
+++ b/fw/apm.c
@@ -1241,11 +1241,26 @@ tfw_apm_hm_srv_rcount_update(TfwStr *uri_path, void *apmref)
 		atomic64_inc(&hmctl->rcount);
 }
 
+static inline u32
+__tfw_apm_crc32_calc(TfwMsgIter *it, TfwStr *chunk , struct sk_buff *skb_head,
+		     TfwStr *body)
+{
+	u32 crc = 0;
+
+	TFW_BODY_ITER_WALK(it, chunk)
+		crc = crc32(crc, chunk->data, chunk->len);
+
+	return crc;
+}
+
 bool
-tfw_apm_hm_srv_alive(int status, TfwStr *body, void *apmref)
+tfw_apm_hm_srv_alive(int status, TfwStr *body, struct sk_buff *skb_head,
+		     void *apmref)
 {
 	TfwApmHM *hm = READ_ONCE(((TfwApmData *)apmref)->hmctl.hm);
 	u32 crc32 = 0;
+	TfwMsgIter it;
+	TfwStr chunk = {0};
 
 	BUG_ON(!hm);
 	if (hm->codes && !test_bit(HTTP_CODE_BIT_NUM(status), hm->codes)) {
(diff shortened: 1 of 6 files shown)


const-t:
@krizhanovsky Is [this condition](https://github.com/tempesta-tech/tempesta/blob/master/fw/http.c#L1624) correct? If so `tfw_apm_hm_srv_alive()` will be called only for suspended server.

krizhanovsky (approved):
I have only one question and if there is no problem, then just comment why do we exepct body in the first skb and merge.

Regarding `tfw_http_hm_control()` - we return from the function if the server is alive (not suspended) since we do not need to mark it alive with `tfw_srv_mark_alive()`, so the code seems correct. This is an example why it's bad to write code without comments :)

fw/http.h
```diff
@@ -713,6 +706,78 @@ tfw_http_msg_header_table_size(void)
...
+		   struct sk_buff* start, struct sk_buff* end)
+{
+	int r;
+
+	it->skb_head = end;
+	it->skb = start;
+	it->frag = -1;
+
+	/* Set starting position. */
+	r = ss_skb_find_frag_by_offset(it->skb, body_start, &it->frag);
```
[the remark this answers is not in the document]
const-t: This is not first skb. `tfw_body_iter_init()` - Initializes iterator with data obtained during body parsing, we pass to this function pointer to start of the body and pointer to `sk_buff` that holds the body or first chunks of the body. We use these pointers as starting point. If skb `start` doesn't "contain" passed pointer `body_start` error will be handled below. `ss_skb_find_frag_by_offset()` finds number of the fragment that contains `body_start` pointer.

const-t:
> Regarding `tfw_http_hm_control()` - we return from the function if the server is alive (not suspended) since we do not need to mark it alive with `tfw_srv_mark_alive()`, so the code seems correct. This is an example why it's bad to write code without comments :)

This part is ok. However looks strange following condition if we take into account that `tfw_apm_hm_srv_alive()` must be called for **alive** server at [least once](https://github.com/tempesta-tech/tempesta/wiki/Health-monitor) to calculate valid crc32 in case when directive `resp_crc32` is `auto`:
```
if (!tfw_srv_suspended(srv) ||
	    !tfw_apm_hm_srv_alive(resp->status, &resp->body, srv->apmref))
```

fw/apm.c
```diff
@@ -1260,14 +1275,22 @@ tfw_apm_hm_srv_alive(int status, TfwStr *body, void *apmref)
...
+			  hm->name, status);
+		return false;
+	}
+
 	/*
 	 * Special case for 'auto' monitor: generate crc32
 	 * from body of first response and store it into monitor.
 	 */
 	if (!hm->crc32 && hm->auto_crc) {
-		hm->crc32 = tfw_str_crc32_calc(body);
```
EvgeniiMekhanik, line 1268: tfw_str_crc32_calc is now unused, we can remove it from our code and tests

EvgeniiMekhanik (approved):

fw/apm.c
const-t: I would prefer don't remove this function. Time to time we use crc32 for `TfwStr` even for debug purposes and it useful to have such function.

Issue #1968, opened 2023-08-31
Kernel BUG at ws.test_ws_ping.WsPipelining.test


EvgeniiMekhanik:
Start test: ws.test_ws_ping.WsPipelining.test
Aug 31 11:16:19 192.168.50.173 [ 4512.043479] [tdb] Start Tempesta DB
Aug 31 11:16:19 192.168.50.173 [ 4512.072036] [tempesta fw] Initializing Tempesta FW kernel module...
Aug 31 11:16:19 192.168.50.173 [ 4512.257778] [tempesta fw] Warning: Vhost default doesn't have certificate with matching SAN/CN.
[ 4512.257778]     Maybe that's fine, but it's worth checking the
[ 4512.257778]     config - if there is no relations between the
[ 4512.257778]     names, then host name confusion attack is possible.
Aug 31 11:16:19 192.168.50.173 [ 4512.259636] [tempesta fw] Configuration processing is completed.
Aug 31 11:16:19 192.168.50.173 [ 4512.347890] [tdb] Opened table /opt/tempesta/db/filter0.tdb: size=16777216 rec_size=20 base=0000000018b90a05
Aug 31 11:16:19 192.168.50.173 [ 4512.428598] [tdb] Opened table /opt/tempesta/db/cache0.tdb: size=268435456 rec_size=0 base=00000000dfb8d691
Aug 31 11:16:19 192.168.50.173 [ 4512.434621] [tdb] Opened table /opt/tempesta/db/sessions0.tdb: size=16777216 rec_size=312 base=000000001d9e20c0
Aug 31 11:16:19 192.168.50.173 [ 4512.440387] [tdb] Opened table /opt/tempesta/db/client0.tdb: size=16777216 rec_size=624 base=00000000391e0568
Aug 31 11:16:19 192.168.50.173 [ 4512.593704] [tempesta fw] Open listen socket on: 0.0.0.0:82
Aug 31 11:16:19 192.168.50.173 [ 4512.597764] [tempesta fw] Open listen socket on: 0.0.0.0:81
Aug 31 11:16:19 192.168.50.173 [ 4512.749895] [tempesta fw] Tempesta FW is ready
Aug 31 11:16:25 192.168.50.173 [ 4517.849057] [tempesta fw] Warning: Request dropped: Pipelined request received after UPGRADE request: 127.0.0.1
Aug 31 11:16:25 192.168.50.173 [ 4517.850458] ------------[ cut here ]------------
[...]

PR #1984, opened 2023-09-28
Remove unnecessary BUG_ON

const-t:
This BUG_ON can be removed, because `cli_conn->refcnt` can be below then two at this place. For instance in case when socket was stolen and after it client connection was aborted. Therefore server connection will be upgraded for invalid client connection. It's not a good case for us but not critical. We can add check that client connection is still alive in `tfw_ws_srv_new_steal_sk()`, but it will not fix the problem completely.

Details:
When the second request is received by Tempesta `tfw_http_msg_process_generic()` checks whether the client already sent upgrade request or not, if upgrade request was received as the first request and response to this request still not sent, Tempesta drops the connection with the client and decrements ref counter of client connection. However in parallel Tempesta processes response to the first upgrade request and at this moment might happen a race between connection closing in `tfw_http_msg_process_generic()` and `tfw_http_websocket_upgrade()`. In this case, decrement of reference counter may happen right after socket stealing `tfw_ws_srv_new_steal_sk()`, therefore ref counter of `cli_conn` in `tfw_http_websocket_upgrade()` will be equal to 1. Then Tempesta will try to forward response to closed connection, but in the function `tfw_h1_resp_adjust_fwd()` response will be dropped, because in this function we have condition that checks the state of the client connection.

### DIFF
--- a/fw/websocket.c
+++ b/fw/websocket.c
@@ -3,7 +3,7 @@
  *
  * Websocket proxy protocol implementation for Tempesta FW.
  *
- * Copyright (C) 2022-2023 Tempesta Technologies, Inc.
+ * Copyright (C) 2022-2024 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
@@ -197,7 +197,7 @@ tfw_http_websocket_upgrade(TfwSrvConn *srv_conn, TfwCliConn *cli_conn)
 	 */
 	cli_conn->pair = (TfwConn *)ws_conn;
 	tfw_connection_get(cli_conn->pair);
-	BUG_ON(atomic_read(&cli_conn->refcnt) < 2);
+	BUG_ON(atomic_read(&cli_conn->refcnt) < 1);
 
 	ws_conn->pair = (TfwConn *)cli_conn;
 	tfw_connection_get(ws_conn->pair);


fw/websocket.c
```diff
@@ -197,7 +197,6 @@ tfw_http_websocket_upgrade(TfwSrvConn *srv_conn, TfwCliConn *cli_conn)
 	 */
 	cli_conn->pair = (TfwConn *)ws_conn;
 	tfw_connection_get(cli_conn->pair);
-	BUG_ON(atomic_read(&cli_conn->refcnt) < 2);
```
krizhanovsky, line 200: I'm afraid that with this change we might observe referencing the dead client connection as in #1647 we tried to send a response on a dead client connection. This assertion was introduced in https://github.com/tempesta-tech/tempesta/pull/1797 and we can not drop a reference count on the client connection because we have not forwarded response on the connection (the comment above references `tfw_http_conn_drop()` for this logic).

Which function calls sequence leads to `cli_conn->refcnt < 2` and what's its value (I hope not 0)?

krizhanovsky (changes requested):
This is very subtle place and we might observe use-after-free crashes - probably something is completely wrong, something that wasn't considered in #1797. I think to fix the issue we need to consider #1968 crash along with #1647 and review the #1797 fixes - maybe some of the decisions were wrong. But just removing the assertion looks scary.

fw/websocket.c
const-t: I've added little bit details to commit message and PR description.

krizhanovsky (approved):
OK, thank you for the clarifications. Really, `atomic_read(&cli_conn->refcnt) < 2` is expected to be true after the pairing and getting the connection in lines 202-203. Most likely, this `BUG_ON()` should have been under the lines, but now it's also good.

LGTM

PR #1993, opened 2023-10-27
Rework stream state processing

### LINKED ISSUE #1823: h2spec: http2/4.3/3 and http2/6.2/2 - incorrect processing of header block in different streams
# Scope
Tempesta must not process header block (HEADERS, CONTINUATION, PUSH_PROMISE frames) in new streams until it receives END_HEADERS in previous stream.

1. Client send HEADERS frame with stream_id = 1 and END_HEADERS - False;
2. Client send HEADERS frame with stream_id = 3
3. Client send CONTINUATION frame with stream_id = 1 and END_HEADERS - True

ER: GOAWAY frame with PROTOCOL ERROR
AR: Tempesta returns HEADERS frame.

[RFC 9113 4.3](https://httpwg.org/specs/rfc9113.html#FieldBlock)

> Each field block is processed as a discrete unit. Field blocks MUST be transmitted as a contiguous sequence of frames, with no interleaved frames of any other type or from any other stream. The last frame in a sequence of HEADERS or CONTINUATION frames has the END_HEADERS flag set. The last frame in a sequence of PUSH_PROMISE or CONTINUATION frames has the END_HEADERS flag set. This allows a field block to be logically equivalent to a single frame.

# Testing

Tests to reproduce [1](https://github.com/tempesta-tech/tempesta-test/blob/c485d6c32ae365ca10756f2bd3fc14bb3c5eb21a/http2_general/test_h2_specs.py#L110) and [2](https://github.com/tempesta-tech/tempesta-test/blob/c485d6c32ae365ca10756f2bd3fc14bb3c5eb21a/http2_general/test_h2_specs.py#L120). Please enable tests after fix.

EvgeniiMekhanik:
Previously we iplement two states for streams (which are not described in the RFC) for two special cases:
- we receive headers without END_HEADERS flag and without END_STREAM flag.
- we receive headers without END_HEADERS flag but with END_STREAM flag. But this states don't cover all possible cases, so in this patch i remove this two states, and add to pointers to the connection context to save streams, for which we receive or send headers without
END_HEADERS flags, we save flag to indicate whether we receive or sent END_STREAM for current stream.
Later when we send or receive frame, we check if
appropriate stream pointer in the connection context is not NULL, new frame should belongs to the same
stream and should be continuation frame.

Closes #1823

### DIFF
--- a/fw/http_stream.c
+++ b/fw/http_stream.c
@@ -59,34 +59,64 @@ tfw_h2_stream_cache_destroy(void)
  * @return - Stream FSM exec status
  */
 TfwStreamFsmRes
-tfw_h2_stream_fsm(TfwStream *stream, unsigned char type, unsigned char flags,
-		  bool send, TfwH2Err *err)
+tfw_h2_stream_fsm(TfwH2Ctx *ctx, TfwStream *stream, unsigned char type,
+		  unsigned char flags, bool send, TfwH2Err *err)
 {
 	TfwStreamFsmRes res = STREAM_FSM_RES_OK;
+	TfwStreamState new_state;
+
+/*
+ * The next two macros checks RFC 9113 4.3:
+ * Each field block is processed as a discrete unit. Field blocks MUST be
+ * transmitted as a contiguous sequence of frames, with no interleaved
+ * frames of any other type or from any other stream. The last frame in a
+ * sequence of HEADERS or CONTINUATION frames has the END_HEADERS flag set.
+ * The last frame in a sequence of PUSH_PROMISE or CONTINUATION frames has
+ * the END_HEADERS flag set. This allows a field block to be logically
+ * equivalent to a single frame.
+ */
+#define TFW_H2_FSM_STREAM_CHECK(ctx, stream, op)			\
+do {									\
+	if (ctx->cur_##op##_headers					\
+	    && stream != ctx->cur_##op##_headers) { 			\
+		*err = HTTP2_ECODE_PROTO;				\
+		res = STREAM_FSM_RES_TERM_CONN;				\
+		goto finish;						\
+	}								\
+} while(0)
+
+#define TFW_H2_FSM_TYPE_CHECK(ctx, stream, op, type)			\
+do {									\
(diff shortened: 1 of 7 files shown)


fw/http_stream.c
```diff
@@ -120,43 +149,66 @@ tfw_h2_stream_fsm(TfwStream *stream, unsigned char type, unsigned char flags,
...
 			{
 			case HTTP2_F_END_HEADERS | HTTP2_F_END_STREAM:
 				stream->state = send
 					? HTTP2_STREAM_LOC_HALF_CLOSED
 					: HTTP2_STREAM_REM_HALF_CLOSED;
+
 				break;
 			case HTTP2_F_END_HEADERS:
 				/*
 				 * Headers is ended, next frame in the stream
```
krizhanovsky, line 173:
```suggestion
				 * Headers are ended, next frame in the stream
```

fw/http_stream.h
```diff
@@ -41,8 +41,6 @@ typedef enum {
 	HTTP2_STREAM_LOC_RESERVED,
```
krizhanovsky, line 41: Shouldn't we fix the comment above in reference to a newer RFC and don't the reduced set of states now is fully compliant with the RFC set?

krizhanovsky (approved):
LGTM, but I'm not so fluent with the code, so a good review from @const-t is required.

fw/http_stream.h
EvgeniiMekhanik: No we still have HTTP2_STREAM_LOC_CLOSED and HTTP2_STREAM_REM_CLOSED states, which are not described in the RFC.

fw/http_stream.h
```diff
@@ -162,6 +160,8 @@ struct tfw_http_stream_t {
 	long int		loc_wnd;
 	long int		rem_wnd;
 	unsigned short		weight;
+	bool			send_eos;
+	bool			recv_eos;
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Fixed

const-t (approved):

PR #2005, opened 2023-11-09
Fix handling proxy-revalidate directive

EvgeniiMekhanik:
According to RFC 9111 3.5, we should not cache
response to a request with an Authorization header field unless the response contains a Cache-Control field with a following response directives:
must-revalidate, public and s-maxage.
Proxy-revalidate is not in this list, so we should not cache such responses.


const-t (approved):
LGTM

krizhanovsky (approved):
LGTM.

Issue #1975, opened 2023-09-11
`WARNING` in `/root/tempesta/fw/tls.c:460`


RomanBelozerov:
# Scope

Tempesta - [`f5929f91b1fc8373d30d2f47678b1767d3e04ef0`](https://github.com/tempesta-tech/tempesta/tree/f5929f91b1fc8373d30d2f47678b1767d3e04ef0)

```
------------[ cut here ]------------
WARNING: CPU: 0 PID: 12 at /root/tempesta/fw/tls.c:460 tfw_tls_encrypt+0x873/0x8a0 [tempesta_fw]
Modules linked in: tempesta_fw(OE) tempesta_db(OE) tempesta_tls(OE) tempesta_lib(OE) iptable_mangle xt_mark tcp_diag inet_diag xt_nat xt_tcpudp veth sha256_ssse3 sha512_ssse3 xt_conntrack xt_MASQUERADE nf_conntrack_netlink xfrm_user xfrm_algo xt_addrtype iptable_filter iptable_nat bpfilter br_netfilter bridge stp llc md4 cmac nls_utf8 cifs libarc4 fscache libdes overlay nfnetlink_cttimeout nfnetlink openvswitch nsh nf_conncount nf_nat nf_conntrack nf_defrag_ipv6 nf_defrag_ipv4 dm_multipath scsi_dh_rdac scsi_dh_emc scsi_dh_alua kvm_amd ccp kvm binfmt_misc sch_fq_codel input_leds serio_raw joydev qemu_fw_cfg mac_hid ramoops reed_solomon msr efi_pstore ip_tables x_tables autofs4 btrfs blake2b_generic zstd_compress raid10 raid456 async_raid6_recov async_memcpy async_pq async_xor async_tx xor raid6_pq libcrc32c raid1 raid0 multipath linear hid_generic usbhid hid crct10dif_pclmul crc32_pclmul ghash_clmulni_intel bochs_drm drm_vram_helper drm_ttm_helper ttm drm_kms_helper aesni_intel
 syscopyarea sysfillrect crypto_simd cryptd glue_helper sysimgblt psmouse fb_sys_fops cec virtio_scsi drm pata_acpi e1000 i2c_piix4 floppy [last unloaded: tempesta_lib]
CPU: 0 PID: 12 Comm: ksoftirqd/0 Tainted: G           OE     5.10.35.tfw-4c9ba16 #1
[...]

PR #2007, opened 2023-11-10
Rework socket closing

EvgeniiMekhanik:
There was a problem in our closing procedure. When we close connection with TCP FIN we immediately mark socket as DEAD and do not wait until FIN and all remaining data will be sent. (Usual tcp_close() call sk_stream_wait_close() to wait until it will be nothing to send). If remote peer sends some data to us after we mark socket as DEAD TCP code answer with TCP RST, that could lead to loosing some unsend data.
To solve this problem clsing procedure was splitted into two part. In first part we send TCP FIN and return. Then we wait for ACK to our FIN from remote peer and finish closing.

### DIFF
--- a/fw/sock.c
+++ b/fw/sock.c
@@ -193,6 +193,8 @@ ss_active_guard_exit(unsigned long val)
 static void
 ss_conn_drop_guard_exit(struct sock *sk)
 {
+	if (sk->sk_user_data)
+		SS_CONN_TYPE(sk) &= ~Conn_Closing;
 	SS_CALL(connection_drop, sk);
 	ss_active_guard_exit(SS_V_ACT_LIVECONN);
 }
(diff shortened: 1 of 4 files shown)


const-t (approved):
LGTM

const-t (approved):

fw/sock.c
```diff
@@ -598,62 +668,26 @@ ss_do_close(struct sock *sk, int flags)
...
+adjudge_to_death:
+	ss_finish_closing(sk, flags);
+	return 0;
+}
+
+static void
+ss_full_close(struct sock *sk, int flags)
+{
+	if (ss_do_close(sk, flags))
+		ss_finish_closing(sk, flags);
```
krizhanovsky, line 690: We need this `ss_finish_closing()` call only if there is some data to send, i.e. we have just the same problem as before (e.g. with TLS alert not sent as in #1975)?

fw/sock_clnt.c
```diff
@@ -474,13 +475,16 @@ tfw_sk_write_xmit(struct sock *sk, struct sk_buff *skb, unsigned int mss_now,
...
-	if (unlikely(r) && r != -ENOMEM)
+	if (unlikely(r) && r != -ENOMEM) {
 		/*
 		 * We can not send unencrypted data and can not normally close the
 		 * socket with FIN since we're in progress on sending from the write
 		 * queue.
 		 */
-		ss_close(sk, SS_F_ABORT);
+		SS_CONN_TYPE(sk) |= Conn_Closing;
+		ss_close(sk, SS_F_ABORT_FORCE | __SS_F_FULL_CLOSE);
```
krizhanovsky, line 485: It seems I don't understand the main idea of the patch. I understood it as that we split the `ss_do_close()` to 2 parts and kill the socket only when we send all the data on it.

We have `ss_close()` calls also in tls.c and http.c along with sock.c itself - if there is no error (not a case for `ss_linkerror()`) and we have some data on the socket, then `ss_do_close()` is just called on `all other `ss_close()` and if we have some data on socket, then who kills the socket? I see only 3 calls of `ss_finish_closing()`: `ss_do_close()` for sockets with no data, `ss_linkerror()` and `ss_full_close()` which doesn't cover the cases for `ss_close()` calls.

fw/sock.c
```diff
@@ -1047,6 +1097,17 @@ ss_tcp_state_change(struct sock *sk)
 		 * done after all pending data has been sent.
 		 */
 		ss_close(sk, SS_F_SYNC);
+	} else if (sk->sk_state == TCP_FIN_WAIT2) {
```
const-t, line 1100: Looks like `TCP_FIN_WAIT2` is not enough, if closing has started simultaneously state will be changed to `TCP_CLOSING` that not handled in our case. Thus, `ss_linkerror` will not be called.

PR #2009, opened 2023-11-13
Implement WINDOW_UPDATE support on ingress traffic

EvgeniiMekhanik:
Change tempesta INITIAL_WINDOW_SIZE to 65536
(default value described in RFC).

### DIFF
--- a/fw/http_frame.c
+++ b/fw/http_frame.c
@@ -232,7 +232,7 @@ tfw_h2_context_init(TfwH2Ctx *ctx)
 	bzero_fast(ctx, sizeof(*ctx));
 
 	ctx->state = HTTP2_RECV_CLI_START_SEQ;
-	ctx->loc_wnd = MAX_WND_SIZE;
+	ctx->loc_wnd = DEF_WND_SIZE;
 	ctx->rem_wnd = DEF_WND_SIZE;
 
 	spin_lock_init(&ctx->lock);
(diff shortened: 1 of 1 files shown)


EvgeniiMekhanik:
Since we don't support big requests and responses now i have no idea how to test this PR

krizhanovsky (approved):
LGTM. We discussed today that there are tests employing WINDOW_UPDATE, so it seems we test the logic. Also since the default window now is smaller (64KB instead of 2GB), our tests for large bodies (like 2MB) will hit the logic. 

@RomanBelozerov could you please confirm that my conclusions are correct? In general, could you please make sure that we have tests for WINDOW_UPDATE.

const-t (approved):
LGTM

RomanBelozerov:
> LGTM. We discussed today that there are tests employing WINDOW_UPDATE, so it seems we test the logic. Also since the default window now is smaller (64KB instead of 2GB), our tests for large bodies (like 2MB) will hit the logic.
> 
> @RomanBelozerov could you please confirm that my conclusions are correct? In general, could you please make sure that we have tests for WINDOW_UPDATE.

Yes, you're right and we have tests to check WINDOW_UPDATE in [PR-550](https://github.com/tempesta-tech/tempesta-test/pull/550) and also all tests with request body greater than 64 kB check it (`t_long_body` directory in tempesta-test)

PR #2012, opened 2023-11-14
Increase TLS traceability.

voodam:
- Increase log level in error places.
- Make logs more verbose, replace some magic numbers to strings.


tls/tls_internal.h
```diff
@@ -417,4 +417,46 @@ ttls_rnd(void *buf, int len)
 }
 #endif
 
+static inline char *
+tls_state_to_str(unsigned int state) {
+	switch (state) {
```
const-t, line 422: Maybe we can convert this switch to table?

voodam: Something like this? Is it faster then switch?

```
if (state > MAX) {
    return "Unknown";
} else {
    return table[state];
}
```

voodam: Done. Is it good refactoring? Also, hstype_to_str is pretty big too, it can be converted in the same way.

tls/tls_cli.c
```diff
@@ -1946,7 +1947,7 @@ int ttls_handshake_client_step(TlsCtx *ssl)
 		break;
 
 	   default:
-		   T_DBG("invalid state %d\n", ssl->state);
+		   T_WARN("invalid state %d\n", tls_state_to_str(ssl->state));
 		   return(TTLS_ERR_BAD_INPUT_DATA);
    }
```
krizhanovsky, line 1953: The file isn't used at all at the moment, but I think the changes are still good

tls/tls_srv.c
```diff
@@ -455,7 +455,7 @@ ttls_pick_cert(TlsCtx *tls, const TlsCiphersuite *ci)
 	T_DBG("ciphersuite requires certificate\n");
 
 	if (!list) {
```
krizhanovsky, line 457:
```suggestion
	if (unlikely(!list)) {
```

tls/tls_internal.h
krizhanovsky: The switches are small and compiler and cpu are good in optimizing them and also the switches are only for error code paths.

krizhanovsky (commented):
LGTM. Let's merge it and test in our website

krizhanovsky (approved):
LGTM

PR #2013, opened 2023-11-15
Mekhanik evgenii/fix 1102

EvgeniiMekhanik:


fw/http_sess_conf.c
```diff
@@ -547,6 +592,21 @@ tfw_cfgop_sess_lifetime(TfwCfgSpec *cs, TfwCfgEntry *ce)
 	if (!cur_vhost)
 		defaults_override.lifetime_set = 1;
 
+set_max_age:
+	r = snprintf(max_age, sizeof(max_age), "; Max-Age=%u",
+		     sticky->sess_lifetime);
+	BUG_ON(!r || r >= sizeof(max_age));
+	max_age_len = strlen(max_age);
```
const-t, line 599: Should we use `strlen` here?

PR #2018, opened 2023-12-12
Fix PURGE + http_methods

voodam:
The use of the PURGE method is restricted when the 'http_methods' directive is enabled. PURGE is unsupported  by the directive and, therefore, cannot be used.


const-t (approved):
LGTM.

PR #2023, opened 2023-12-18
Tempesta & backend servers health statistics

voodam:
Pay attention to tfw_apm_global_data. Here, existing code is effectively reused,
but we use the same synchronization primitives as for the backend (which makes
sense as the data structure is the same). I'm uncertain about the performance
implications of this since the rest of the global Tempesta statistics work with
percpu variables.

Cache misses for Tempesta and backends differ: for Tempesta, a miss is
considered when a request doesn't hit the cache, while for the backend, it's
considered at the time of response delivery (uncached).

Is the use of for_each_possible_cpu correct?

Any ideas regarding code smells are appreciated.

There's an optional refactoring left after the commit rework, but I decided to
keep it, as it seems good on its own.

### DIFF
--- a/fw/apm.c
+++ b/fw/apm.c
@@ -1,7 +1,7 @@
 /*
  *		Tempesta FW
  *
- * Copyright (C) 2016-2021 Tempesta Technologies, Inc.
+ * Copyright (C) 2016-2024 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/fw/apm.c
+++ b/fw/apm.c
@@ -689,6 +689,13 @@ static int tfw_apm_jtmintrvl;		/* Time interval in jiffies. */
 static int tfw_apm_tmwscale;		/* Time window scale. */
 
 /*
+ * Global statistics for all requests passing through Tempesta (whether cached
+ * or not). Relies on the 'apm_stats' directive, same as the statistics for
+ * individual servers.
+ */
+static TfwApmData *tfw_apm_global_data;
+
+/*
  * Get the next bucket in the ring buffer entry that has a non-zero
  * hits count. Set the bucket's sequential number, the range number,
  * and the bucket number. Set the response time value for the bucket.
@@ -953,13 +960,10 @@ tfw_apm_calc(TfwApmData *data)
  * tfw_apm_stats() should be used for calls in kernel context.
  * tfw_apm_stats_bh() should be used for calls in user context.
  */
-#define __tfw_apm_stats_body(apmref, pstats, fn_lock, fn_unlock)	\
+#define __tfw_apm_stats_body(data, pstats, fn_lock, fn_unlock)		\
+({									\
 	unsigned int rdidx, seq = pstats->seq;				\
-	TfwApmData *data;						\
 	TfwApmSEnt *asent;						\
-									\
-	BUG_ON(!apmref);						\
-	data = &((TfwApmRef *)apmref)->data;				\
 									\
 	smp_mb__before_atomic();					\
 	rdidx = atomic_read(&data->stats.rdidx);			\
@@ -971,18 +975,33 @@ tfw_apm_calc(TfwApmData *data)
 	fn_unlock(&asent->rwlock);					\
 	pstats->seq = rdidx;						\
 									\
-	return (seq != rdidx);
+	seq != rdidx;							\
+})
 
 int
 tfw_apm_stats_bh(void *apmref, TfwPrcntlStats *pstats)
 {
-	__tfw_apm_stats_body(apmref, pstats, read_lock_bh, read_unlock_bh);
+	TfwApmData *data;
+	BUG_ON(!apmref);
+	data = &((TfwApmRef *)apmref)->data;
+	return __tfw_apm_stats_body(data, pstats, read_lock_bh, read_unlock_bh);
 }
 
 int
 tfw_apm_stats(void *apmref, TfwPrcntlStats *pstats)
 {
-	__tfw_apm_stats_body(apmref, pstats, read_lock, read_unlock);
+	TfwApmData *data;
+	BUG_ON(!apmref);
+	data = &((TfwApmRef *)apmref)->data;
+	return __tfw_apm_stats_body(data, pstats, read_lock, read_unlock);
+}
+
+int
+tfw_apm_stats_global(TfwPrcntlStats *pstats)
+{
+	BUG_ON(!tfw_apm_global_data);
+	return __tfw_apm_stats_body(tfw_apm_global_data, pstats, read_lock_bh,
+				    read_unlock_bh);
 }
 
 /*
@@ -1089,6 +1108,13 @@ tfw_apm_update(void *apmref, unsigned long jtstamp, unsigned long jrtt)
 	BUG_ON(!apmref);
 	__tfw_apm_update(&((TfwApmRef *)apmref)->data, jtstamp,
 			 jiffies_to_msecs(jrtt));
+}
+
+void
+tfw_apm_update_global(unsigned long jtstamp, unsigned long jrtt)
+{
+	BUG_ON(!tfw_apm_global_data);
+	__tfw_apm_update(tfw_apm_global_data, jtstamp, jiffies_to_msecs(jrtt));
 }
 
 static void
@@ -1224,29 +1250,61 @@ tfw_apm_ref_create(void)
 	return ref;
 }
 
+static void
+tfw_apm_data_destroy(TfwApmData *data)
+{
+	tfw_apm_free_ubuf(data);
+	kfree(data);
+}
+
+static TfwApmData *
+tfw_apm_data_create(void)
+{
+	int rbufsz = tfw_apm_tmwscale;
+	int size = sizeof(TfwApmRef)
+		+ rbufsz * sizeof(TfwApmRBEnt)
+		+ 2 * T_PSZ * sizeof(unsigned int);
+	TfwApmData *data;
+
+	if ((data = kzalloc(size, GFP_ATOMIC)) == NULL)
+		return NULL;
+
+	if (IS_ERR(tfw_apm_data_init(data))) {
+		tfw_apm_data_destroy(data);
+		return NULL;
+	}
+
+	return data;
+}
+
+/* Start the timer for the percentile calculation. */
 static inline void
-tfw_apm_hm_stop_timer(TfwApmHMCtl *hmctl) {
-	atomic_set(&hmctl->rearm, 0);
+tfw_apm_data_start_timer(TfwApmData *data)
+{
+	set_bit(TFW_APM_DATA_F_REARM, &data->flags);
+	timer_setup(&data->timer, tfw_apm_prcntl_tmfn, 0);
+	mod_timer(&data->timer, jiffies + TFW_APM_TIMER_INTVL);
+}
+
+/* Stop the timer and the percentile calculation. */
+static inline void
+tfw_apm_data_stop_timer(TfwApmData *data)
+{
+	clear_bit(TFW_APM_DATA_F_REARM, &data->flags);
 	smp_mb__after_atomic();
-	del_timer_sync(&hmctl->timer);
+	del_timer_sync(&data->timer);
 }
 
 int
 tfw_apm_add_srv(TfwServer *srv)
 {
 	TfwApmRef *ref;
-	TfwApmData *data;
 
 	BUG_ON(srv->apmref);
 	if (!(ref = tfw_apm_ref_create()))
 		return -ENOMEM;
 
-	data = &ref->data;
-	/* Start the timer for the percentile calculation. */
-	set_bit(TFW_APM_DATA_F_REARM, &data->flags);
-	timer_setup(&data->timer, tfw_apm_prcntl_tmfn, 0);
-	mod_timer(&data->timer, jiffies + TFW_APM_TIMER_INTVL);
-
+	tfw_apm_data_start_timer(&ref->data);
 	srv->apmref = ref;
 
 	return 0;
@@ -1264,11 +1322,7 @@ tfw_apm_del_srv(TfwServer *srv)
 	if (test_bit(TFW_SRV_B_HMONITOR, &srv->flags))
 		tfw_apm_hm_disable_srv(srv);
 
-	/* Stop the timer and the percentile calculation. */
-	clear_bit(TFW_APM_DATA_F_REARM, &ref->data.flags);
-	smp_mb__after_atomic();
-	del_timer_sync(&ref->data.timer);
-
+	tfw_apm_data_stop_timer(&ref->data);
 	tfw_apm_ref_destroy(ref);
 	srv->apmref = NULL;
 }
@@ -1445,6 +1499,13 @@ tfw_apm_hm_enable_srv(TfwServer *srv, const char *hm_name)
 
 	/* Activate server's health monitor. */
 	set_bit(TFW_SRV_B_HMONITOR, &srv->flags);
+}
+
+static inline void
+tfw_apm_hm_stop_timer(TfwApmHMCtl *hmctl) {
+	atomic_set(&hmctl->rearm, 0);
+	smp_mb__after_atomic();
+	del_timer_sync(&hmctl->timer);
 }
 
 void
@@ -1690,6 +1751,11 @@ tfw_apm_cfgend(void)
 	if (tfw_runstate_is_reconfig())
 		return 0;
 
+	tfw_apm_global_data = tfw_apm_data_create();
+	if (!tfw_apm_global_data)
+		return -EINVAL;
+	tfw_apm_data_start_timer(tfw_apm_global_data);
+
 	if ((r = tfw_apm_create_def_hm()))
 		return r;
 
@@ -1699,6 +1765,12 @@ tfw_apm_cfgend(void)
 static void
 tfw_apm_cfgclean(void)
 {
+	if (tfw_apm_global_data) {
+		tfw_apm_data_stop_timer(tfw_apm_global_data);
+		tfw_apm_data_destroy(tfw_apm_global_data);
+		tfw_apm_global_data = NULL;
+	}
+
 	/*
 	 * 'auto' health monitor may be created implicitly in cfgend(),
 	 * even if no `health_check` directive found.
--- a/fw/apm.c
+++ b/fw/apm.c
@@ -546,11 +546,11 @@ typedef struct {
  * Keeps the latest values of calculated percentiles.
  *
  * @pstats	- The percentile stats structure.
- * @rwlock	- Protect updates.
+ * @seqlock	- Protect updates.
  */
 typedef struct {
 	TfwPrcntlStats	pstats;
-	rwlock_t	rwlock;
+	seqlock_t	seqlock;
 } TfwApmSEnt;
 
 /*
@@ -563,7 +563,7 @@ typedef struct {
  * is used to decrease the lock contention. Readers read the stored values
  * at @asent[@rdidx % 2]. The writer writes the new percentile values to
  * @asent[(@rdidx + 1) % 2], and then increments @rdidx. The reading and
- * the writing are protected by a rwlock.
+ * the writing are protected by a seqlock.
  *
  * @asent	- The stats entries for reading/writing (flip-flop manner).
  * @rdidx	- The current index in @asent for readers.
@@ -941,11 +941,11 @@ tfw_apm_calc(TfwApmData *data)
 	tfw_apm_prnctl_calc(&data->rbuf, &data->rbctl, &pstats);
 
 	T_DBG3("%s: Percentile values may have changed.\n", __func__);
-	write_lock(&asent->rwlock);
+	write_seqlock(&asent->seqlock);
 	memcpy_fast(asent->pstats.val, pstats.val,
 		    T_PSZ * sizeof(asent->pstats.val[0]));
 	atomic_inc(&data->stats.rdidx);
-	write_unlock(&asent->rwlock);
+	write_sequnlock(&asent->seqlock);
 
 	return;
 }
@@ -953,39 +953,29 @@ tfw_apm_calc(TfwApmData *data)
 /*
  * Get the latest calculated percentiles.
  *
- * Return 0 if the percentile values didn't need recalculation.
- * Return 1 if potentially new percentile values were calculated.
- *
- * The two functions below differ only by the type of lock used.
- * tfw_apm_stats() should be used for calls in kernel context.
- * tfw_apm_stats_bh() should be used for calls in user context.
+ * Return false if the percentile values didn't need recalculation.
+ * Return true if potentially new percentile values were calculated.
  */
-#define __tfw_apm_stats_body(data, pstats, fn_lock, fn_unlock)		\
-({									\
-	unsigned int rdidx, seq = pstats->seq;				\
-	TfwApmSEnt *asent;						\
-									\
-	smp_mb__before_atomic();					\
-	rdidx = atomic_read(&data->stats.rdidx);			\
-	asent = &data->stats.asent[rdidx % 2];				\
-									\
-	pstats->ith = tfw_pstats_ith;					\
-	fn_lock(&asent->rwlock);					\
-	memcpy(pstats->val, asent->pstats.val,				\
-	       T_PSZ * sizeof(pstats->val[0]));				\
-	fn_unlock(&asent->rwlock);					\
-	pstats->seq = rdidx;						\
-									\
-	seq != rdidx;							\
-})
-
-int
-tfw_apm_stats_bh(void *apmref, TfwPrcntlStats *pstats)
+static bool
+__tfw_apm_stats(TfwApmData *data, TfwPrcntlStats *pstats)
 {
-	TfwApmData *data;
-	BUG_ON(!apmref);
-	data = &((TfwApmRef *)apmref)->data;
-	return __tfw_apm_stats_body(data, pstats, read_lock_bh, read_unlock_bh);
+	unsigned int rdidx, s, seq = pstats->seq;
+	TfwApmSEnt *asent;
+
+	smp_mb__before_atomic();
+	rdidx = atomic_read(&data->stats.rdidx);
+	asent = &data->stats.asent[rdidx % 2];
+
+	do {
+		s = read_seqbegin(&asent->seqlock);
+		memcpy(pstats->val, asent->pstats.val,
+		       T_PSZ * sizeof(pstats->val[0]));
+	} while (read_seqretry(&asent->seqlock, s));
+
+	pstats->ith = tfw_pstats_ith;
+	pstats->seq = rdidx;
+
+	return seq != rdidx;
 }
 
 int
@@ -994,15 +984,14 @@ tfw_apm_stats(void *apmref, TfwPrcntlStats *pstats)
 	TfwApmData *data;
 	BUG_ON(!apmref);
 	data = &((TfwApmRef *)apmref)->data;
-	return __tfw_apm_stats_body(data, pstats, read_lock, read_unlock);
+	return __tfw_apm_stats(data, pstats);
 }
 
 int
 tfw_apm_stats_global(TfwPrcntlStats *pstats)
 {
 	BUG_ON(!tfw_apm_global_data);
-	return __tfw_apm_stats_body(tfw_apm_global_data, pstats, read_lock_bh,
-				    read_unlock_bh);
+	return __tfw_apm_stats(tfw_apm_global_data, pstats);
 }
 
 /*
@@ -1174,8 +1163,8 @@ tfw_apm_data_init(TfwApmData *data)
 		tfw_apm_rbent_init(&rbent[i], 0);
 	spin_lock_init(&data->rbuf.slock);
 
-	rwlock_init(&data->stats.asent[0].rwlock);
-	rwlock_init(&data->stats.asent[1].rwlock);
+	seqlock_init(&data->stats.asent[0].seqlock);
+	seqlock_init(&data->stats.asent[1].seqlock);
 	atomic_set(&data->stats.rdidx, 0);
 
 	size = 2 * TFW_APM_UBUF_SZ * sizeof(TfwApmUBEnt);
--- a/fw/apm.c
+++ b/fw/apm.c
@@ -441,15 +441,17 @@ typedef struct {
 /*
  * Accounting entry for particular HTTP code.
  *
- * @history	- ring buffer of history records ;
+ * @history	- ring buffer of history records;
  * @hmcfg	- pointer to structure with settings for particular HTTP code;
  * @rsum	- current amount of responses from @history ring buffer;
+ * @total	- total amount of responses during all the time;
  * @lock	- spinlock for synchronized work with @history buffer;
  */
 typedef struct {
 	TfwApmHMHistory		history[HM_FREQ];
 	TfwApmHMCfg		*hmcfg;
 	unsigned int		rsum;
+	u64			total;
 	spinlock_t		lock;
 } TfwApmHMStats;
 
@@ -473,13 +475,11 @@ typedef struct {
 } TfwApmHMCtl;
 
 /* Entry for configuration of separate health monitors. */
-static TfwApmHM			*tfw_hm_entry;
+static TfwApmHM		*tfw_hm_entry;
 /* Entry for configuration of default 'auto' health monitor. */
-static TfwApmHM			*tfw_hm_default;
-/* Flag for explicit creation of default health monitor. */
-static bool			tfw_hm_expl_def;
+static TfwApmHM		*tfw_hm_default;
 /* Total count of monitored HTTP codes. */
-static unsigned int		tfw_hm_codes_cnt;
+static unsigned int	tfw_hm_codes_cnt;
 
 static LIST_HEAD(tfw_hm_list);
 static LIST_HEAD(tfw_hm_codes_list);
@@ -1313,7 +1313,9 @@ tfw_apm_hm_srv_limit(int status, void *apmref)
 	TfwApmHMCfg *cfg = NULL;
 	unsigned long ts;
 
-	BUG_ON(!hmstats);
+	if (!hmstats)
+		return false;
+
 	for (i = 0; i < tfw_hm_codes_cnt; ++i) {
 		if (tfw_http_status_eq(status, hmstats[i].hmcfg->code))
 		{
@@ -1326,38 +1328,61 @@ tfw_apm_hm_srv_limit(int status, void *apmref)
 
 	if (!history)
 		return false;
-
 	BUG_ON(!cfg);
+
 	spin_lock(&hmstats->lock);
 
-	ts = jiffies * HM_FREQ / (cfg->tframe * HZ);
-	i = ts % HM_FREQ;
-	if (history[i].ts != ts) {
-		history[i].ts = ts;
-		history[i].resp = 0;
+	/*
+	 * 'health_stat_server' directive reuses the same data structure as
+	 * 'server_failover_http' (TfwApmHMCfg) but without the set @tframe
+	 * and @limit (refer to tfw_cfgop_apm_health_stat_srv()). Thus, these
+	 * attributes are optional.
+	 */
+	if (cfg->tframe) {
+		ts = jiffies * HM_FREQ / (cfg->tframe * HZ);
+		i = ts % HM_FREQ;
+		if (history[i].ts != ts) {
+			history[i].ts = ts;
+			history[i].resp = 0;
+		}
+		++history[i].resp;
+		for (i = 0; i < HM_FREQ; ++i)
+			if (history[i].ts + HM_FREQ > ts)
+				sum += history[i].resp;
+		WRITE_ONCE(hmstats->rsum, sum);
 	}
-	++history[i].resp;
-	for (i = 0; i < HM_FREQ; ++i)
-		if (history[i].ts + HM_FREQ > ts)
-			sum += history[i].resp;
+	++hmstats->total;
 
-	WRITE_ONCE(hmstats->rsum, sum);
 	spin_unlock(&hmstats->lock);
 
-	if (sum > cfg->limit)
+	if (cfg->limit && sum > cfg->limit)
 		return true;
 
 	return false;
 }
 
+static TfwApmHM *
+tfw_apm_get_hm(const char *name)
+{
+	TfwApmHM *hm;
+
+	list_for_each_entry(hm, &tfw_hm_list, list) {
+		if (!strcasecmp(name, hm->name))
+			return hm;
+	}
+	return NULL;
+}
+
 void
-tfw_apm_hm_enable_srv(TfwServer *srv, void *hmref)
+tfw_apm_hm_enable_srv(TfwServer *srv, const char *hm_name)
 {
 	TfwApmHMCtl *hmctl;
 	unsigned long now;
-	TfwApmHM *hm = hmref;
+	TfwApmHM *hm;
 
 	BUG_ON(!srv->apmref);
+
+	hm = tfw_apm_get_hm(hm_name);
 	BUG_ON(!hm);
 	WARN_ON_ONCE(test_bit(TFW_SRV_B_HMONITOR, &srv->flags));
 	WARN_ON_ONCE(test_bit(TFW_SRV_B_SUSPEND, &srv->flags));
@@ -1410,12 +1435,13 @@ TfwHMStats *
 tfw_apm_hm_stats(void *apmref)
 {
 	int i;
-	long rtime;
+	long rtime = 0;
 	TfwHMStats *stats;
 	TfwApmHMCtl *hmctl = &((TfwApmData *)apmref)->hmctl;
 	TfwApmHM *hm = READ_ONCE(hmctl->hm);
 
-	BUG_ON(!hmctl->hmstats || !hm);
+	if (!hmctl->hmstats)
+		return NULL;
 
 	stats = kmalloc(tfw_hm_stats_size(tfw_hm_codes_cnt), GFP_KERNEL);
 	if (!stats)
@@ -1426,24 +1452,14 @@ tfw_apm_hm_stats(void *apmref)
 		BUG_ON(!hmctl->hmstats[i].hmcfg);
 		stats->rsums[i].code = hmctl->hmstats[i].hmcfg->code;
 		stats->rsums[i].sum = READ_ONCE(hmctl->hmstats[i].rsum);
+		stats->rsums[i].total = READ_ONCE(hmctl->hmstats[i].total);
 	}
 
-	rtime = (long)hm->tmt - (jiffies - READ_ONCE(hmctl->jtmstamp)) / HZ;
+	if (hm)
+		rtime = (long)hm->tmt - (jiffies - READ_ONCE(hmctl->jtmstamp)) / HZ;
 	stats->rtime = rtime < 0 ? 0 : rtime;
 
 	return stats;
-}
-
-void *
-tfw_apm_get_hm(const char *name)
-{
-	TfwApmHM *hm;
-
-	list_for_each_entry(hm, &tfw_hm_list, list) {
-		if (!strcasecmp(name, hm->name))
-			return hm;
-	}
-	return NULL;
 }
 
 bool
@@ -1455,7 +1471,7 @@ tfw_apm_check_hm(const char *name)
 		return false;
 	}
 
-	if (!strcasecmp(name, TFW_APM_HM_AUTO) || tfw_apm_get_hm(name))
+	if (tfw_apm_get_hm(name))
 		return true;
 
 	T_ERR_NL("health monitor with name '%s' does not exist\n", name);
@@ -1540,7 +1556,6 @@ __tfw_cfgop_cleanup_apm_hm(void)
 
 	tfw_hm_entry = NULL;
 	tfw_hm_default = NULL;
-	tfw_hm_expl_def = false;
 
 	list_for_each_entry_safe(hm, tmp, &tfw_hm_list, list) {
 		free_pages((unsigned long)hm->req, get_order(hm->reqsz));
@@ -1558,12 +1573,45 @@ tfw_cfgop_cleanup_apm_hm(TfwCfgSpec *cs)
 	__tfw_cfgop_cleanup_apm_hm();
 }
 
+
+/**
+ * Create 'auto' health monitor for default mode if explicit one have not been
+ * created during configuration parsing stage.
+ */
+static int
+tfw_apm_create_def_hm(void)
+{
+	int r;
+
+	if (tfw_apm_get_hm(TFW_APM_HM_AUTO))
+		return 0;
+
+	if ((r = tfw_cfgop_apm_add_hm(TFW_APM_HM_AUTO)))
+		return r;
+
+	if ((r = tfw_cfgop_apm_add_hm_req(TFW_APM_DFLT_REQ, tfw_hm_entry)))
+		return r;
+
+	if ((r = tfw_cfgop_apm_add_hm_url(TFW_APM_DFLT_URL, tfw_hm_entry)))
+		return r;
+
+	if ((r = tfw_cfgop_apm_alloc_hm_codes(tfw_hm_entry)))
+		return r;
+	/*
+	 * Default values for health response code is 200, for
+	 * crc32 is 'auto', and for monitor request timeout is 10s.
+	 */
+	__set_bit(HTTP_CODE_BIT_NUM(200), tfw_hm_entry->codes);
+	tfw_hm_entry->auto_crc = true;
+	tfw_hm_entry->tmt = 10;
+	tfw_hm_entry = NULL;
+	return 0;
+}
+
 static int
 tfw_apm_cfgend(void)
 {
 	unsigned int jtmwindow;
-	int r;
-
 	if (tfw_runstate_is_reconfig())
 		return 0;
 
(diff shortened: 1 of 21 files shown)


fw/apm.c
```diff
@@ -639,17 +628,42 @@ typedef struct {
...
+/*
+ * Order of the members is important for the memory allocation, @data must be
+ * at the end.
+ */
+typedef struct {
+	TfwApmHMCtl		hmctl;
+	atomic64_t		hits;
+	atomic64_t		misses;
+	TfwApmData		data;
+} TfwApmRef;
```
krizhanovsky, line 665: Why do we need the new data structure, why can't (shouldn't) we keep the cache misses and hits in `TfwApmData`?

fw/apm.c
```diff
@@ -930,52 +951,49 @@ tfw_apm_calc(TfwApmData *data)
...
 {
-	int i;
-
-	for (i = 0; i < T_PSZ; ++i)
-		if (pstats->ith[i] != tfw_pstats_ith[i])
-			return 1;
-	return 0;
+	BUG_ON(!tfw_apm_global_data);
+	return __tfw_apm_stats_body(tfw_apm_global_data, pstats, read_lock_bh,
+				    read_unlock_bh);
```
krizhanovsky, line 996: Speaking of the introduced additional synchronization, it seems only locking in this place is introduces (besides the atomic operations for the cache hits & misses). The locking is happen between the timer and procfs reading, so it's not a huge concurrency.

However, timers work in the same softirq as the main networking code, so a huge tail latencies may occur due to a procfs reading. Maybe we should use seqlocks for explicit reader starvation to let softirqs always make progress.

krizhanovsky (approved):
In general I'm good with the PR, but please consider moving to seqlock to guarantee softirq progress (@const-t @EvgeniiMekhanik please also pay attention to this).

Also please update the Wiki about different cache counters for per-server and overall Tempesta statistic.

P.S. very nice patch splitting and good refactoring!

fw/apm.c
```diff
@@ -33,7 +33,6 @@
 #include "cfg.h"
```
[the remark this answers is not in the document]
voodam: Done

fw/stat.h
```diff
@@ -0,0 +1,124 @@
+/**
+ *		Tempesta kernel library
+ *
+ * Copyright (C) 2019 Tempesta Technologies, Inc.
```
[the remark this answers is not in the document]
voodam: Done, but it's a mess in copyrights, probably I did it wrong, I need for the confirmation (on meeting, probably).

fw/apm.c
```diff
@@ -639,17 +628,42 @@ typedef struct {
...
-	TfwServer		*srv;
 	struct timer_list	timer;
 	unsigned long		flags;
-	TfwApmHMCtl		hmctl;
 } TfwApmData;
 
+/*
+ * Order of the members is important for the memory allocation, @data must be
+ * at the end.
+ */
```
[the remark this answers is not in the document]
voodam: Done

fw/apm.c
voodam: We need TfwApmData separately, without heath monitor, for the global statistics (`tfw_apm_global_data`). Hits and misses also don't required here.

fw/procfs.c
```diff
@@ -359,7 +407,54 @@ tfw_procfs_stop(void)
...
+}
+
+static void
+tfw_cfgop_cleanup_health_stat(TfwCfgSpec *cs)
+{
+	int cpu;
+
+	kfree(health_stat_codes);
+
+	for_each_possible_cpu(cpu) {
```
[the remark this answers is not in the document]
voodam: Got it, fixed. But why only work queue? It's also in tdb, tls and in `skb_init` in kernel patch (for `TfwSkbMemPool`.)

fw/apm.c
```diff
@@ -1329,47 +1417,62 @@ tfw_apm_hm_srv_limit(int status, void *apmref)
...
-
 	BUG_ON(!cfg);
 	spin_lock(&hmstats->lock);
 
-	ts = jiffies * HM_FREQ / (cfg->tframe * HZ);
-	i = ts % HM_FREQ;
-	if (history[i].ts != ts) {
-		history[i].ts = ts;
-		history[i].resp = 0;
+	if (cfg->tframe) {
```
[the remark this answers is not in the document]
voodam: It's used now for health_stat_server, which is utilized as structure for server_failover_http, but with empty tframe an limit (see `tfw_cfgop_apm_health_stat_srv`).

fw/stat.h
```diff
@@ -0,0 +1,124 @@
...
+	u64	misses;
+} TfwCacheStat;
+
+/*
+ * Structures for health monitoring statistics accountings in procfs.
+ */
+typedef struct {
+	int		code;
+	unsigned int	sum;
+	u64		total;	/* Another sum for total stats; can be unused. */
```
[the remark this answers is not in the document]
voodam: Fixed

fw/apm.c
```diff
@@ -665,6 +679,13 @@ static int tfw_apm_jtmwindow;		/* Time window in jiffies. */
 static int tfw_apm_jtmintrvl;		/* Time interval in jiffies. */
 static int tfw_apm_tmwscale;		/* Time window scale. */
 
+/*
+ * Global statistics for all requests passing through Tempesta (whether cached
+ * or not). Relies on the 'apm_stats' directive, same as the statistics for
+ * individual servers.
+ */
+static TfwApmData *tfw_apm_global_data;
```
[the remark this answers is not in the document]
voodam: But I did the commit to introduce this info in perfstat - sorrt, didn't understand what you mean.

fw/cache.c
```diff
@@ -308,7 +309,7 @@ __cache_method_test(tfw_http_meth_t method)
 	return cache_cfg.methods & (1 << method);
 }
 
-bool
+static inline bool
```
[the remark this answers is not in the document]
voodam: Got it, fixed in the all files. Also removing it works like detector for unused functions :)

fw/apm.c
```diff
@@ -643,12 +628,27 @@ typedef struct {
...
+ * of percentiles - the function that runs periodically on timer.
+ * If there are several different parties that do the calculation,
+ * then the data may need to be organized differently.
+ *
+ * @rbuf	- The ring buffer for the specified time window.
+ * @rbctl	- The control data helpful in taking optimizations.
+ * @stats	- The latest percentiles.
+ * @ubuf	- The buffer that holds data for updates, per CPU.
+ * @timer	- The periodic timer handle.
+ * @flags	- The atomic flags (see below).
```
keshonok, line 645: Looks like the flags are no longer below after the rearrangement of the code.

fw/apm.h
```diff
@@ -62,21 +62,43 @@ static const unsigned int tfw_pstats_ith[] = {
...
+/* Cache hits/misses statistics. */
 void tfw_apm_cache_inc_hits(void *apmref);
 void tfw_apm_cache_inc_misses(void *apmref);
 void tfw_apm_cache_stats(TfwCacheStat *stats, void *apmref);
+
+/*
+ * Although these procedures are used for health monitoring, they also collect
+ * some general server statistics. Therefore, they work even when health
+ * monitoring is turned off. In this case, server suspension and resumption
+ * will not work, and only overall response statistics for the entire uptime
```
keshonok, line 91: It's not particularly clear what you wanted to say. Perhaps, you can rephrase or elaborate?

fw/apm.c
```diff
@@ -995,23 +996,6 @@ tfw_apm_stats_global(TfwPrcntlStats *pstats)
...
 }
 
-/*
- * Verify that an APM Stats user using the same set of percentiles.
- *
- * Note: This module uses a single set of percentiles for all servers.
- * All APM Stats users must use the same set of percentiles.
- */
-int
-tfw_apm_pstats_verify(TfwPrcntlStats *pstats)
```
keshonok, line 971: Is there any particular reason why this function is removed? Just wondering. Thanks!

voodam: I don't understand why it's required (and no one on the current team). In the single place where it was used (in `tfw_procfs_cfgend`) it checked that `tfw_pstats_ith` equals `tfw_pstats_ith`, that always be true. It seems that `tfw_pstats_ith` is hardcoded in several places as the only valid index array (`TfwPrcntlStats.ith`), but it was supposed flexibility that wasn't fully implemented.

fw/apm.c
voodam: Fixed

fw/apm.c
```diff
@@ -962,18 +966,33 @@ tfw_apm_calc(TfwApmData *data)
...
+	TfwApmData *data;
+	BUG_ON(!apmref);
+	data = &((TfwApmRef *)apmref)->data;
+	return __tfw_apm_stats_body(data, pstats, read_lock, read_unlock);
+}
+
+int
+tfw_apm_stats_global(TfwPrcntlStats *pstats)
+{
+	BUG_ON(!tfw_apm_global_data);
```
keshonok, line 1005: Just to clarify -- If this is to imitate whatever is done in the macro, then `apmref` is verified for non-zero value because it's dereferenced later. The resulting `data` is then dereferenced as well, but it's presumed set at this stage.

fw/apm.c
voodam: Or probably it's misunderstanding of the design, what was initial idea behind that?

fw/apm.h
voodam: Done

fw/apm.c
keshonok: Yeah, that's what I thought. I don't fully recall what was the intended flexibility after all these years. As you mentioned, the percentiles were defined (hardcoded?) in several places in the code.

As far as I recall, this was the function to safeguard against changing the percentile arrangement in one place and not changing in another. Different people work on the code so it might be easy to miss or forget. As this function is only called once at startup, it's a cheap test that may prevent a lot of aggravation. So if that's no longer true and the percentile arrangement is defined in a single place as it should, then this function is no longer useful.

keshonok: Perhaps, then, this should go in the commit's comment.

fw/http.c
```diff
@@ -1600,12 +1601,12 @@ static bool
 tfw_http_hm_suspend(TfwHttpResp *resp, TfwServer *srv)
 {
 	unsigned long old_flags, flags = READ_ONCE(srv->flags);
+	bool hm = flags & TFW_SRV_F_HMONITOR;
 
-	if (!(flags & TFW_SRV_F_HMONITOR))
-		return true;
-
-	if (!tfw_apm_hm_srv_limit(resp->status, srv->apmref))
+	if (!tfw_apm_hm_srv_limit(resp->status, srv->apmref) && hm)
```
keshonok, line 1606: I'm not sure why this change is required. What is the difference? Why is there the need to check `hm` twice? Also, `hm` was introduced just for this purpose.

voodam: I introduced the total response statistics in the same data structure, that serves limit calculation (`TfwApmHMStats`). Also that's what I tried to describe in the comments in `fw/apm.h`. That was optimal from the point of view of data structures, but naming probably suffered, but I didn't refactor it.

voodam: So now we need to `tfw_apm_hm_srv_limit` to be called even if health monitoring is disabled, just to count statistics.

fw/http.c
```diff
@@ -4121,12 +4122,13 @@ __tfw_http_resp_fwd(TfwCliConn *cli_conn, struct list_head *ret_queue)
 			return;
 		}
 		list_del_init(&req->msg.seq_list);
-		tfw_http_resp_pair_free(req);
 		TFW_INC_STAT_BH(serv.msgs_forwarded);
+		tfw_inc_global_hm_stats(req->resp->status);
+		tfw_http_resp_pair_free(req);
```
keshonok, line 4127: Personally, I dislike that `tfw_http_resp_pair_free()` and `list_del_init()` are separated now by an aside logic. The are logically coupled and should stay together. That's why stats accounting is usually done after they are called in other places of the code. I can see that you need `req` alive in this particular place. Perhaps, stats accounting can be done before the call to `list_del_init()` in this specific case? I believe that was the code would be a little cleaner and easier to understand. It's not like it matters much from the functional point of view. Just an opinion. Thanks.

fw/http.c
```diff
@@ -6118,6 +6120,10 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb,
 }
 
 /**
+ * Process the cached response from tfw_cache_process() or the original
+ * response from a backend if it wasn't cached dueto certain circumstances
```
keshonok, line 6124: `dueto` -> `due to`

fw/http.c
keshonok: Thanks. What about an alternative that clearly shows the logic that the function is called unconditionally? Something along these lines:
```C
r = tfw_apm_hm_srv_limit(resp->status, srv->apmref);
if ((!(flags & TFW_SRV_F_HMONITOR))
    return true;
if (!r)
    return false.
```

keshonok: Also, I think the comment that it's called unconditionally because of this and that would do good here.

fw/cache.c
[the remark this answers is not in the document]
keshonok: It might be me but I think the removal of `inline` everywhere may have been overzealous.

fw/apm.c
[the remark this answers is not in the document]
keshonok: Perhaps, a comment to that would make it clear for code readers.

fw/procfs.c
```diff
@@ -176,17 +232,14 @@ tfw_srvstats_seq_show(struct seq_file *seq, void *off)
...
-	};
+	unsigned int val[T_PSZ] = { 0 };
+	TfwPrcntlStats pstats = {.val = val};
+	TfwHMStats *hm_stats;
 
 	if (!(qsize = kmalloc(sizeof(int) * srv->conn_n, GFP_KERNEL)))
 		return -ENOMEM;
 
-	tfw_apm_stats_bh(srv->apmref, &pstats);
+	tfw_apm_stats(srv->apmref, &pstats);
```
keshonok, line 242: Are you sure that you no longer need to use the `_bh` version?

fw/http.c
voodam: Done, thanks!

fw/http.c
voodam: I don't know, there are no much difference from my side. Anyway, fixed.

fw/http.c
voodam: Fixed

fw/cfg.h
```diff
@@ -191,6 +191,21 @@ typedef struct {
...
 	     (v) = (e)->vals[++(idx)])
 
+#define TFW_CFG_CHECK_NO_ATTRS(spec, entry)				\
+	if ((entry)->attr_n) {						\
+		T_ERR_NL("%s: Arguments may not have the '=' sign\n",	\
+			 (spec)->name);					\
+		return -EINVAL;						\
+	}
+
+#define TFW_CFG_CHECK_VAL_N(op, req, spec, entry)				\
```
EvgeniiMekhanik, line 201: I don't think we should implement this macros, we should use `tfw_cfg_check_val_n` in all places.

fw/cfg.h
```diff
@@ -191,6 +191,21 @@ typedef struct {
 	     (idx) < (e)->val_n;		\
 	     (v) = (e)->vals[++(idx)])
 
+#define TFW_CFG_CHECK_NO_ATTRS(spec, entry)				\
```
EvgeniiMekhanik, line 194: I think we should can implement `tfw_cfg_check_attr_n` function with same as in `tfw_cfg_check_val_n` error message.
```
T_ERR_NL("invalid number of attributes; expected: %d, got: %zu\n",
			 attr_n, ce->attr_n);
```

fw/apm.h
```diff
@@ -79,6 +81,91 @@ typedef struct {
...
+{
+	int i, code;
+	const char *val;
+	tfw_hm_stats_init(s, ce->val_n);
+
+	TFW_CFG_ENTRY_FOR_EACH_VAL(ce, i, val) {
+		BUG_ON(i >= s->ccnt);
+
+		if (tfw_cfgop_parse_http_status(val, &code)
+		    || (code > HTTP_STATUS_5XX
```
EvgeniiMekhanik: I don't see how can code be greater then HTTP_STATUS_5XX. As I see we check code correctness in `tfw_cfgop_parse_http_status`. If there is some error in this function, we should fix it, not check code hre.

fw/apm.h
```diff
@@ -79,6 +81,91 @@ typedef struct {
...
+			&& tfw_cfg_check_range(code, HTTP_CODE_MIN,
```
EvgeniiMekhanik: We also check range int the `tfw_cfgop_parse_http_status` why we should do it here?

fw/procfs.c
```diff
@@ -359,13 +401,54 @@ tfw_procfs_stop(void)
...
+{
+	int cpu;
+
+	TFW_CFG_CHECK_VAL_N(>, 0, cs, ce);
+	TFW_CFG_CHECK_NO_ATTRS(cs, ce);
+
+	health_stat_codes = kzalloc(tfw_hm_stats_size(ce->val_n),
+				    GFP_KERNEL);
+	if (!health_stat_codes)
+		return -EINVAL;
```
EvgeniiMekhanik: -ENOMEM

fw/procfs.c
```diff
@@ -359,13 +401,54 @@ tfw_procfs_stop(void)
...
+	}
+	return 0;
+}
+
+static void
+tfw_cfgop_cleanup_health_stat(TfwCfgSpec *cs)
+{
+	int cpu;
+
+	kfree(health_stat_codes);
```
EvgeniiMekhanik, line 438: We have a lot of problems in memory allocation/deallocation here! I use a simple garbage config
```
health_stat 1;
dsgfdsfdsfds
```
and catch kernel panic here.

If `tfw_hm_stats_init_from_cfg_entry` fails we free `health_stat_codes` but not zero this variable and have double free in
`tfw_cfgop_cleanup_health_stat` for example.

fw/procfs.c
```diff
@@ -359,13 +401,54 @@ tfw_procfs_stop(void)
...
+		kfree(health_stat_codes);
+		return -EINVAL;
+	}
+
+	for_each_possible_cpu(cpu) {
+		TfwPerfStat *pcp_stat = per_cpu_ptr(&tfw_perfstat, cpu);
+		pcp_stat->hm = kmalloc_node(tfw_hm_stats_size(ce->val_n),
+					    GFP_KERNEL, cpu_to_node(cpu));
+		if (!pcp_stat->hm)
+			return -EINVAL;
```
EvgeniiMekhanik: -ENOMEM

fw/apm.c
```diff
@@ -1652,6 +1713,12 @@ tfw_apm_cfgend(void)
 {
 	if (tfw_runstate_is_reconfig())
 		return 0;
+
+	tfw_apm_global_data = tfw_apm_data_create();
+	if (!tfw_apm_global_data)
+		return -EINVAL;
```
EvgeniiMekhanik, line 1720: -ENOMEM

fw/apm.c
```diff
@@ -944,14 +951,11 @@ tfw_apm_calc(TfwApmData *data)
  * tfw_apm_stats() should be used for calls in kernel context.
  * tfw_apm_stats_bh() should be used for calls in user context.
  */
-#define __tfw_apm_stats_body(apmref, pstats, fn_lock, fn_unlock)	\
+#define __tfw_apm_stats_body(data, pstats, fn_lock, fn_unlock)		\
```
EvgeniiMekhanik: May be change this macros to static inline function? I don't think that it is good to return  macros

fw/cache.c
```diff
@@ -2882,7 +2882,8 @@ cache_req_process_node(TfwHttpReq *req, tfw_http_cache_cb_t action)
 
 	resp = tfw_cache_build_resp(req, ce, lifetime, id);
 
-	tfw_apm_cache_inc_hits(((TfwServer *)resp->conn->peer)->apmref);
+T_WARN("!!!OLOLO %d %d %d", resp == NULL, resp && (resp->conn == NULL), resp && resp->conn && (resp->conn->peer == NULL));
```
EvgeniiMekhanik: This commit should be dropped

db/core/htrie.c
```diff
@@ -122,37 +116,25 @@ tdb_init_mapping(void *p, size_t db_size, unsigned int rec_len)
...
-static inline void
+static void
 tdb_free_index_blk(TdbHtrieNode *node)
 {
 	/* Just zero the block and leave it for garbage collector. */
 	bzero_fast(node, sizeof(*node));
 }
 
 /* TODO synchronize the bucket access. */
-static inline void
```
EvgeniiMekhanik, line 133: I am not sure that it is good idea to change all inline functions to not inline. Up to you

fw/cache.c
voodam: Deleted in further commit

fw/procfs.c
voodam: Fixed in further commit

fw/procfs.c
voodam: Fixed in further commit

fw/cfg.h
voodam: `tfw_cfg_check_val_n` is more verbose and less declarative, I would prefer macros with `return` inside for that reason. But I can replace it, of cource.

fw/apm.h
voodam: `HTTP_STATUS_5XX` is 5, it means that we provide code like 5* and this is correct code value that should be handled in `tfw_http_status_eq`

fw/apm.h
voodam: This is some mistake probably, it was in original code, I'll fix it.

fw/apm.c
voodam: Fixed in further commit, but here we can't simply do this because of fns (lambda seem to be overengineering from my side, but maybe I'm wrong).

fw/cache.c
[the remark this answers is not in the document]
voodam: What do you mean by overzealous, what exactly drawbacks are?

fw/stat.h
```diff
@@ -0,0 +1,117 @@
...
+{
+	int i, code;
+	const char *val;
+	tfw_hm_stats_init(s, ce->val_n);
+
+	TFW_CFG_ENTRY_FOR_EACH_VAL(ce, i, val) {
+		BUG_ON(i >= s->ccnt);
+
+		if (tfw_cfgop_parse_http_status(val, &code)
+		    || (code > HTTP_STATUS_5XX
```
EvgeniiMekhanik, line 103: Same as wrote previously we check code and range in `tfw_cfgop_parse_http_status`. If there is some mistake in this function we should fix it.

fw/sock_clnt.c
```diff
@@ -935,12 +935,9 @@ tfw_cfgop_listen(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
 tfw_cfgop_keepalive_timeout(TfwCfgSpec *cs, TfwCfgEntry *ce)
 {
-	int r;
-
-	if ((r = tfw_cfg_check_val_n(ce, 1)))
-		return -EINVAL;
+	TFW_CFG_CHECK_VAL_N(==, 1, cs, ce);
 
-	if ((r = tfw_cfg_parse_int(ce->vals[0], &tfw_cli_cfg_ka_timeout))) {
+	if (tfw_cfg_parse_int(ce->vals[0], &tfw_cli_cfg_ka_timeout)) {
```
EvgeniiMekhanik, line 940: It's better to leave `((r = tfw_cfg_parse_int(ce->vals[0], &tfw_cli_cfg_ka_timeout)))` as is and return `r` in case of error

fw/procfs.h
```diff
@@ -109,4 +116,27 @@ DECLARE_PER_CPU_ALIGNED(TfwPerfStat, tfw_perfstat);
...
 		this_cpu_add(tfw_perfstat.__VA_ARGS__, val)
 
+static inline void
+tfw_inc_global_hm_stats(int status)
+{
+	int i;
+	TfwPerfStat *this_stat = this_cpu_ptr(&tfw_perfstat);
+	TfwHMStats *hm = this_stat->hm;
+
+	if (!hm)
```
EvgeniiMekhanik, line 126: In what case we can have `hm == NULL` here?

fw/procfs.c
```diff
@@ -77,6 +86,18 @@ tfw_perfstat_collect(TfwPerfStat *stat)
 		SADD(serv.rx_bytes);
 		SADD(serv.tls_hs_successful);
 		SADD(serv.tls_hs_failed);
+
+		/*
+		 * Health statistics (differs from health monitor statistics):
+		 * total responses for tempesta
+		 */
+		if (stat->hm && pcp_stat->hm)
```
EvgeniiMekhanik, line 94: In what case stat->hm or pcp_stat->hm can be equal to zero? As I see we allocate it when Tempesta start and return error if allocation fails.

fw/http_sched_ratio.c
```diff
@@ -451,15 +451,12 @@ __tfw_sched_ratio_calc_dynamic(TfwRatio *ratio, TfwRatioData *rtodata,
...
  *    a partial recalculation of ratios?
  */
-static inline int
+static int
 __tfw_sched_ratio_get_rtt(size_t si, TfwRatio *ratio, TfwRatioData *rtodata)
 {
 	unsigned int recalc;
 	unsigned int val[T_PSZ] = { 0 };
-	TfwPrcntlStats pstats = {
-		.ith = tfw_pstats_ith,
```
EvgeniiMekhanik, line 460: Why do you change this?

fw/http_match.h
```diff
@@ -1,7 +1,7 @@
 /**
  *		Tempesta FW
  *
- * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
+ * Copyright (C) 2014-2023 NatSys Lab. (info@natsys-lab.com).
```
EvgeniiMekhanik, line 4: As I think we should leave as is Copyright year for NatSys Lab.

fw/http.c
```diff
@@ -1093,6 +1093,7 @@ tfw_h2_resp_fwd(TfwHttpResp *resp)
 	}
 	else {
 		TFW_INC_STAT_BH(serv.msgs_forwarded);
+		tfw_inc_global_hm_stats(resp->status);
```
EvgeniiMekhanik, line 1096: Why we call this function here and don't call it in other places where `TFW_INC_STAT_BH(serv.msgs_forwarded);` is called? For example `__tfw_http_resp_fwd`.

fw/apm.c
voodam: Fixed, but not only ENOMEM, see in further commit

fw/stat.h
voodam: Fixes

fw/sock_clnt.c
voodam: Fixed

fw/http_sched_ratio.c
voodam: We have tfw_pstats_ith to be hardcoded in several places, so the the only valid value for now can be only tfw_pstats_ith, so I moved it under the hood and removed verification function.

fw/stat.h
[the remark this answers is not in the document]
keshonok: I believe the copyright years should have been updated only in the files that had been changed in this PR.

fw/cache.c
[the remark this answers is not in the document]
keshonok: I meant that you had changed **all** inline functions in the code to regular. They have been inlined for a purpose, and I am not convinced the compiler will substitute all calls to former inline functions with their actual code (i.e. inline them) in all cases. It might do that for smaller functions called once or twice, and might not do that for functions called more frequently or for larger functions where the choice for code duplication (inlining) was made intentionally. All in all, I am not totally convinced by this global change. @krizhanovsky would know better on this subject. :-)

fw/procfs.c
voodam: Fixed

fw/procfs.h
voodam: When `health_stat` option is not set

fw/procfs.c
voodam: If `health_stat` option is not set (`health_stat_codes` == NULL). But we can check just one of two pointers (`stat` or `pcp_stat`), this is just a defensive programming here.

voodam:
Probably it will be better to squash commit (from time to time at least). It's pain to manage this (and probably adds more mess for reviewers).

fw/procfs.c
```diff
@@ -360,6 +360,12 @@ tfw_procfs_stop(void)
 }
 
 static TfwCfgSpec tfw_procfs_specs[] = {
+/*	{
+		.name       = "health_stat",
+		.handler    = tfw_cfgop_apm_stats,
+		.cleanup    = tfw_cfgop_cleanup_apm,
+	},
+*/
```
keshonok: I don't think this should be here.

fw/stat.h
```diff
@@ -0,0 +1,114 @@
...
+ */
+static inline int
+tfw_hm_stats_init_from_cfg_entry(TfwHMStats *s, TfwCfgEntry *ce)
+{
+	int i, code;
+	const char *val;
+	tfw_hm_stats_init(s, ce->val_n);
+
+	TFW_CFG_ENTRY_FOR_EACH_VAL(ce, i, val) {
+		BUG_ON(i >= s->ccnt);
```
keshonok, line 100: I think this is overly excessive.

fw/stat.h
```diff
@@ -0,0 +1,114 @@
...
+
+		if (tfw_cfgop_parse_http_status(val, &code))
+		{
```
keshonok, line 103: Shouldn't the curly brace be on the same line with `if` in this case?

fw/stat.h
voodam: Why?

fw/stat.h
voodam: Fixed

fw/stat.h
keshonok: a `BUN_ON()` is usually used to catch a probable bug that is a result of unforeseen interaction of various complex parts of code. However, here you explicitly assign `s->ccnt = ce->val_n`, and then execute `TFW_CFG_ENTRY_FOR_EACH_VAL()` macro that can't run `i` bigger or equal than `ce->val_n`. So, what is it exactly that this `BUG_ON()` is trying to catch?

I would also take a look at other uses of `TFW_CFG_ENTRY_FOR_EACH_VAL()` - just to be systematic.

voodam: Fixed

fw/cfg.h
keshonok: What is the reason that `tfw_cfg_check_val_n()` still exists and used in several places of Tempesta code? Or, the question rather should be: is there a clearly defined logic for whether the function or the new macro should be used? Right now it's inconsistent.

The new macro covers more cases. Yet, at the same time the macro is clearly used where the existing function would suffice. Yet more, like I said, the function is still used in several places and not replaced by the new macro.

What the new macro does well is that it makes checks short and concise in the code, and I like this refactoring a lot. It removes a lot of clutter. 

The other thing this refactoring does well is that it eliminates the disparity in return codes when checking the data. Not everything was covered by functions,  and where the functions were used `r` was returned, and in other places `-EINVAL` was returned explicitly. That created a visible disparity. It's less now that the return and the returned error are hidden inside the macro.

I am still ambivalent on the fact that it's a macro, but like I said it does provide for less clutter. Perhaps, other places with similar checks can be changed along the same lines? Not sure if that's a good idea but, perhaps, it's worth considering.

voodam: Yes, it was unfinished refactoring, I refactored a lot of places, but not all. I think we need to define best way to work with it on the future meeting (in 2024). Pros: more declarative, cons: less composable, we can't combine several such macros (the reason of tfw_cfg_check_val_n existence).

fw/cache.c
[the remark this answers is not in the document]
krizhanovsky: I meant that it just didn't make sense to add `inline` specifier in this particular pull request. There is no sense to remove all inline specifiers around the project.

fw/procfs.c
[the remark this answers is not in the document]
krizhanovsky: Good point. I created https://github.com/tempesta-tech/tempesta/issues/2033 to address this. This could be a point of severe memory lost.

krizhanovsky (approved):
LGTM

fw/http_match.h
krizhanovsky: Yeah, NatSys was replaced with Tempesta Technologies in 2015, so all copyrights including and after 2015 are for Tempesta Technoloigies. NatSys is defunct now.

fw/apm.c
```diff
@@ -1714,13 +1717,17 @@ static int
...
 	if (tfw_runstate_is_reconfig())
 		return 0;
 
-	tfw_apm_global_data = tfw_apm_data_create();
-	if (!tfw_apm_global_data)
-		return -EINVAL;
+	r2 = tfw_apm_data_create();
+	if (IS_ERR(r2))
+		return PTR_ERR(r2);
+	else
```
keshonok, line 1728: There's no need for `else`. Just continue the flow of the code.

fw/stat.h
```diff
@@ -0,0 +1,111 @@
...
+#include "http.h"
+
+/* Shared structures used for statistics. */
+
+/*
+ * Structures for health monitoring statistics accountings in procfs.
+ *
+ * @code	- HTTP code for which the statistics is collected.
+ * @sum		- Amount of responses for some limited period.
+ * @total	- Total amount of responses for the overall period.
```
keshonok, line 32: I know this isn't your doing, but why not make it a bit better? Below are thoughts on both better naming and better comments.

First, it's unclear that these are structures for backend health monitoring which account only for responses from backends.

`HTTP code` - this is called HTTP response code, or - better, in my mind - HTTP status code. It's often called just `status`, aka response status code. I believe `status` is more identifiable than just `code`. But `code` is quite just okay if it's aligned with other names on the structure.

`sum` and `total` look like synonyms, and thus difficult to remember which is which. Perhaps, `sum` can be `nresp` with the comment `Number of responses in a specified time frame`. Other names are welcome (like `rcnt`?), anything is better than inexpressive `sum`! Then, `total` looks fine, possible with a better comment `Total number of responses` (total is total, saying that it's _overall_ is meaningless).

Alternatives, as this is all about responses with specific status, could be: `status`, `count`, `total`. Or `code`, `count`, `total`. Or `rcode`, `rcount`, `rtotal` where `r` means `response`. When confined within `CodeStats{}` structure, these names have clear meaning.

fw/stat.h
```diff
@@ -0,0 +1,111 @@
...
+	unsigned int	sum;
+	u64		total;
+} TfwHMCodeStats;
+
+/*
+ * @rtime	- time until next server health checking (can be unused if not
+ *		  required);
+ * @ccnt	- count of @rsums elements;
+ * @rsums	- array of counters for separate HTTP codes;
+ */
```
keshonok, line 45: I know this isn't your doing, but why not make it a bit better? Below are thoughts on both better naming and better comments.

Here it's baffling what `rtime` stands for. It it's the `remaining time`, then it's definitely not telling at all. Perhaps, `tmcheck` - check time? `Time until next server health check`. It's unclear what it's meant that it can be unused if these stats are collected unconditionally.

Then, `rsums` could be just `stats` or `cstats` - `Array of stats per HTTP status code`, and `ccnt` -> `stats_sz`  (or `statsz`) - `Size of stats array` as per convention throughout the code.

Also, you can perhaps avoid multiplication to `HZ` every time and keep timers in jiffies as it's done in other parts of APM. Then, `tmcheck` would become `jtmcheck`.

fw/apm.c
```diff
@@ -440,15 +440,17 @@ typedef struct {
 /*
  * Accounting entry for particular HTTP code.
  *
- * @history	- ring buffer of history records ;
+ * @history	- ring buffer of history records;
  * @hmcfg	- pointer to structure with settings for particular HTTP code;
  * @rsum	- current amount of responses from @history ring buffer;
+ * @total	- total amount of responses during all the time;
```
keshonok, line 446: Does this include multiple reconfiguration?

fw/apm.h
```diff
@@ -59,41 +60,40 @@ static const unsigned int tfw_pstats_ith[] = {
...
-int tfw_apm_stats_bh(void *apmref, TfwPrcntlStats *pstats);
-int tfw_apm_pstats_verify(TfwPrcntlStats *pstats);
+/* Displayed in the perfstat, not in a backend statistics. */
+void tfw_apm_update_global(unsigned long jtstamp, unsigned long jrtime);
+int tfw_apm_stats_global(TfwPrcntlStats *pstats);
+
+/*
+ * Although these procedures are used for health monitoring, they also collect
+ * overall statistics on responses from each server. Therefore, if health
+ * monitoring is not active for a server, these procedures will still be
```
keshonok, line 84: `will still be executed` -> `are still executed`.

fw/procfs.c
```diff
@@ -359,7 +402,57 @@ tfw_procfs_stop(void)
...
+				    GFP_KERNEL);
+	if (!health_stat_codes)
+		return -ENOMEM;
+	if (tfw_hm_stats_init_from_cfg_entry(health_stat_codes, ce)) {
+		kfree(health_stat_codes);
+		health_stat_codes = NULL;
+		return -EINVAL;
+	}
+
+	for_each_online_cpu(cpu) {
```
keshonok, line 423: Not sure this is the correct approach. What is one of the initially online CPUs goes down, and another one comes up? What if an additional CPU is inserted (hotplug)?

The same goes for destruction of data allocated here.

fw/procfs.h
```diff
@@ -89,11 +90,17 @@ typedef struct {
...
+ * @hm	- Total responses for each HTTP code specified in 'health_stat'.
+ * 	  Accounts for all responses, whether served from the backend or the
+ *	  cache.
+ */
 typedef struct {
 	TfwSsStat	ss;
 	TfwClntStat	clnt;
 	TfwSrvStat	serv;
 	TfwCacheStat	cache;
+	TfwHMStats	*hm;
```
keshonok, line 103: Personally, I would make it a structure member here, not a pointer to structure, in unison with all other members of the structure. The per cpu allocation of the array for stats per each response code would be simpler and cleaner, in my opinion.

fw/procfs.c
```diff
@@ -214,21 +267,19 @@ tfw_srvstats_seq_show(struct seq_file *seq, void *off)
...
-						hm_stats->rsums[i].code,
-						hm_stats->rsums[i].sum);
-			kfree(hm_stats);
-		}
+	seq_printf(seq, "HTTP health monitor is enabled\t: %d\n",
+		   test_bit(TFW_SRV_B_HMONITOR, &srv->flags));
+	seq_printf(seq, "HTTP availability\t\t: %d\n",
+			!tfw_srv_suspended(srv));
+	if ((hm_stats = tfw_apm_hm_stats(srv->apmref))) {
+		seq_printf(seq, "\tTime until next health checking\t: %u\n",
```
keshonok, line 275: Time until next health check

fw/procfs.c
```diff
@@ -214,21 +267,19 @@ tfw_srvstats_seq_show(struct seq_file *seq, void *off)
...
+			   hm_stats->rtime);
+		for (i = 0; i < hm_stats->ccnt; ++i)
+			seq_printf(seq, "\tHTTP '%d' codes\t: %u (%llu total)"
```
keshonok, line 278: HTTP '%d' code (response code - clearer) (or HTTP status)

fw/procfs.c
```diff
@@ -77,6 +86,18 @@ tfw_perfstat_collect(TfwPerfStat *stat)
...
+		/*
+		 * Health statistics (differs from health monitor statistics):
+		 * total responses for tempesta
+		 */
+		if (stat->hm && pcp_stat->hm)
+			for (i = 0; i < stat->hm->ccnt; ++i) {
+				BUG_ON(stat->hm->rsums[i].code !=
+				       pcp_stat->hm->rsums[i].code);
+				stat->hm->rsums[i].total +=
+					pcp_stat->hm->rsums[i].total;
```
keshonok, line 99: `SADD(hm->rsums[i].total)` if `hm` is a structure inside `stat` and not a pointer, like suggested.

fw/procfs.c
```diff
@@ -77,6 +86,18 @@ tfw_perfstat_collect(TfwPerfStat *stat)
...
 		SADD(serv.tls_hs_successful);
 		SADD(serv.tls_hs_failed);
+
+		/*
+		 * Health statistics (differs from health monitor statistics):
+		 * total responses for tempesta
+		 */
+		if (stat->hm && pcp_stat->hm)
+			for (i = 0; i < stat->hm->ccnt; ++i) {
+				BUG_ON(stat->hm->rsums[i].code !=
```
keshonok, line 96: Is the `BUG_ON()` really needed here?

fw/procfs.c
```diff
@@ -85,17 +106,42 @@ static int
...
+	unsigned int val[T_PSZ] = { 0 };
+	TfwPrcntlStats pstats = {.val = val};
+
 	if (!ss_stat)
 		T_WARN("Cannot allocate sync sockets statistics\n");
 
 	memset(&stat, 0, sizeof(stat));
+
+	if (health_stat_codes) {
+		stat.hm = kmalloc(tfw_hm_stats_size(health_stat_codes->ccnt),
```
keshonok, line 126: I am not particularly fond of constant allocation and freeing of small chunks of memory here. Perhaps, Tempesta's memory pool would be better suitable for this?

fw/procfs.c
```diff
@@ -176,17 +232,14 @@ tfw_srvstats_seq_show(struct seq_file *seq, void *off)
...
-	unsigned int val[ARRAY_SIZE(tfw_pstats_ith)] = { 0 };
-	TfwPrcntlStats pstats = {
-		.ith = tfw_pstats_ith,
-		.val = val,
-	};
+	unsigned int val[T_PSZ] = { 0 };
+	TfwPrcntlStats pstats = {.val = val};
+	TfwHMStats *hm_stats;
 
 	if (!(qsize = kmalloc(sizeof(int) * srv->conn_n, GFP_KERNEL)))
```
keshonok, line 239: Same here, regarding the memory allocation - it's a very small chunk, it's short-lived, and it's happens quite regularly. Memory pool candidate?

fw/apm.c
```diff
@@ -1953,6 +2107,13 @@ static TfwCfgSpec tfw_apm_specs[] = {
 		.allow_repeat	= true,
 		.cleanup	= tfw_cfgop_apm_cleanup_server_failover,
 	},
+	{
+		.name		= "health_stat_server",
+		.deflt		= NULL,
+		.handler	= tfw_cfgop_apm_health_stat_srv,
+		.allow_none	= true,
+		.cleanup	= tfw_cfgop_apm_cleanup_server_failover,
```
keshonok, line 2115: Is this correct or is this a copy-paste error? Using the same cleanup function with such a specific name is confusing.

fw/apm.c
```diff
@@ -1953,6 +2107,13 @@ static TfwCfgSpec tfw_apm_specs[] = {
 		.allow_repeat	= true,
 		.cleanup	= tfw_cfgop_apm_cleanup_server_failover,
 	},
+	{
+		.name		= "health_stat_server",
```
keshonok, line 2111: Is there a description for this in `/etc/tempesta_fw.conf` in this PR?

fw/apm.c
```diff
@@ -1752,18 +1888,44 @@ tfw_cfgop_apm_cleanup_server_failover(TfwCfgSpec *cs)
...
+	const char *val;
+	TfwApmHMCfg *hm_entry;
+
+	TFW_CFG_CHECK_VAL_N(>, 0, cs, ce);
+	TFW_CFG_CHECK_NO_ATTRS(cs, ce);
+
+	TFW_CFG_ENTRY_FOR_EACH_VAL(ce, i, val) {
+		if (tfw_cfgop_parse_http_status(val, &code))
+		{
+			T_ERR_NL("Unable to parse http code value: '%s'\n",
```
keshonok, line 1905: Perhaps, `HTTP status value` would be clearer.

fw/apm.c
```diff
@@ -1752,18 +1888,44 @@ tfw_cfgop_apm_cleanup_server_failover(TfwCfgSpec *cs)
...
+{
+	int i, code;
+	const char *val;
+	TfwApmHMCfg *hm_entry;
+
+	TFW_CFG_CHECK_VAL_N(>, 0, cs, ce);
+	TFW_CFG_CHECK_NO_ATTRS(cs, ce);
+
+	TFW_CFG_ENTRY_FOR_EACH_VAL(ce, i, val) {
+		if (tfw_cfgop_parse_http_status(val, &code))
```
keshonok, line 1903: Just wanted to bring attention to this line. The function parses HTTP status, and then places the result in some "code" variable. I think that naming of a variable is important in understanding the code (no, I don't mean the `code` variable! :-)) when trying to follow the logic. Just `code` is very ambiguous and can be anything. On the other hand, for instance, `status` is quite specific when linked with `HTTP`. See a discussion in another comment.

fw/procfs.c
```diff
@@ -359,7 +402,57 @@ tfw_procfs_stop(void)
...
+	TFW_CFG_CHECK_NO_ATTRS(cs, ce);
+
+	health_stat_codes = kzalloc(tfw_hm_stats_size(ce->val_n),
+				    GFP_KERNEL);
+	if (!health_stat_codes)
+		return -ENOMEM;
+	if (tfw_hm_stats_init_from_cfg_entry(health_stat_codes, ce)) {
+		kfree(health_stat_codes);
+		health_stat_codes = NULL;
+		return -EINVAL;
```
EvgeniiMekhanik, line 420: We can just return -EINVAL, `health_stat_codes` will be freed in the `tfw_cfgop_cleanup_health_stat`

fw/procfs.c
EvgeniiMekhanik: Good point! I try to increment count of online CPU in qemu and catch a lot of problems with memory allocation/deallocation in all places where we use  `for_each_online_cpu`. (In sock.c also). @krizhanovsky we should fix it here or in other task?

EvgeniiMekhanik (approved):

fw/cache.c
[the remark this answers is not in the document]
voodam: It wasn't around the project but only in *.c files, but anyway, restored back.

fw/apm.c
voodam: fixed

fw/stat.h
voodam: I don't know... We have a lot of such things around the project. I just skipped them before to not inflate the PRs, considering it as minor refactoring.

In my opinion, sum, nresp, rcnt - it's all the same, no additional information. We should use tf_total (tf=time frame), for example, if we want to add some information. It's quite subjective thing I think.

voodam: I want to discuss it on meeting to form a common idea about that topic, I suppose it can be helpful.

fw/apm.c
voodam: No, reconfiguration isn't allowed for these directives (server_failover_http and health_stat_server).

fw/apm.h
voodam: Fixed

fw/procfs.c
voodam: Fixed

fw/procfs.c
voodam: Fixed

fw/procfs.c
voodam: Fixed

fw/procfs.c
voodam: I don't know, for me it looks more calming :)

fw/apm.c
voodam: No, like all other new directives, thanks! Fixed.

fw/procfs.c
```diff
@@ -85,17 +106,42 @@ static int
...
 
 	memset(&stat, 0, sizeof(stat));
+
+	if (health_stat_codes) {
+		stat.hm = kmalloc(tfw_hm_stats_size(health_stat_codes->ccnt),
+				  GFP_KERNEL);
+		if (stat.hm)
+			tfw_hm_stats_clone(stat.hm, health_stat_codes);
+	} else {
+		stat.hm = NULL;
```
const-t, line 131: Why do we need this explicit assigning of NULL? Above we `memset` to zero this structure. I believe we can get rid of explicit `memset` using `TfwPerfStat stat = {0}`

voodam: Hm, probably it's the bug from the time when this code was above the memset. I don't remember actually, but fixed.

fw/stat.h
keshonok: Oh, it's definitely subjective! I just presented my opinion which is based on my knowledge of English and the proper terminology. :-) Usually I try to give variables names that are meaningful in the context to anyone that knows the terminology, or just follow the common sense. I was difficult for me to follow this code as every time I saw these names I had to get back and read their description again. As for `tf_total`, a time frame is something partial by definition, and having the total within a partial feels strange. Nevertheless, as they say _the best is the enemy of the good_.

And of course it doesn't have to be a part of this PR.

fw/apm.c
```diff
@@ -1157,83 +1173,140 @@ tfw_apm_create(void)
...
+	TfwApmHMCfg *ent;
+
+	if ((ref = kzalloc(size, GFP_ATOMIC)) == NULL)
+		return ERR_PTR(-ENOMEM);
+
+	/*
+	 * End of the TfwApmData structure is the beginning of the TfwApmHMStats
+	 * of the TfwApmHMCtl.
+	 */
+	hmstats = tfw_apm_data_init(&ref->data);
```
keshonok, line 1209: It looks really weird getting the pointer to the end of the initialized data. This looks like an obfuscation. The original variant was more logical and correct, you have all data here to calculate the start of the area for stats.

fw/apm.c
```diff
@@ -1725,16 +1870,12 @@ tfw_cfgop_apm_server_failover(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
 		return -EINVAL;
 
-	hm_entry = kzalloc(sizeof(TfwApmHMCfg), GFP_KERNEL);
+	hm_entry = tfw_amp_create_hm_entry();
 	if (!hm_entry)
 		return -ENOMEM;
-
-	INIT_LIST_HEAD(&hm_entry->list);
-	hm_entry->code = code;
+	tfw_hm_entry_set_code(hm_entry, code);
```
keshonok, line 1876: In my opinion, this whole dance around `tfw_hm_cfg_200_created` flag looks overblown. Just assign `hm_entry->code` here, like you do with the other members. It will looks nice and simple. When all config is processed and it's time to create (or not) the default entry, just go over the list and look for the entry with 200 code. It's clean, simple and fast. And the resulting code will be easier to read and understand.

fw/apm.c
keshonok: Actually, it doesn't. `tfw_apm_data_init()` initializes the `TfwApmData{}` structure. The whole layout of the structure is defined here, and the memory for that layout is allocated here as well. So, whatever implementation it might be, it must follow the defined layout, and it doesn't go beyond that. Nor should it.
The layout defined here includes the memory for `TfwApmHMStats{}` at the end. You don't need the pointer to that area from `tfw_apm_data_init()`, that's very obscure. You have all data here to calculate where that area of memory is located. It's `ref + sizeof(TfwApmRef) + rbufsz * sizeof(TfwApmRBEnt) + 2 * T_PSZ * sizeof(unsigned int)` - which can be calculated once at the start.

fw/apm.c
keshonok: Below is an example of what I meant.

Here:
```C
hm_entry->code = code;
hm_entry->limit = limit;
hm_entry->tframe = tframe;
```
In `tfw_apm_create_def_health_stat_srv()`:
```C
list_for_each_entry(ent, &tfw_hm_codes_list, list) {
    if (ent->code == 200)
        break;
}
if (!list_entry_is_head(ent, &tfw_hm_codes_list, list))
    return 0;
if (!(ent = tfw_amp_create_hm_entry()))
    return -ENOMEM;
ent->code = 200;
```

Thus, no need for `tfw_hm_entry_set_code()` which obscures the initialization of the entry. The initialization is now very visible, simple, clean. Also, no need for globally defined `tfw_hm_cfg_200_created` that is only used for this specific purpose and nothing else.

PR #2029, opened 2024-01-03
Add pre-commit hook to autoupdate copyrights.

voodam:
Use it to enable hooks:
git config --local core.hooksPath .githooks/


const-t (approved):
LGTM

PR #2031, opened 2024-01-04
Fix `req_hdr_set` for Host header

const-t:
During request trasformation `__h2_req_hdrs()` modifies request's headers using rules specified in tempesta config. In this case there was an issue with `req_hdr_set` for Host header. Because `__h2_req_hdrs()` replaced only `Host` header and skipped `authority`, however `authority` must be replaced as well, because `tfw_h2_adjust_req()` may use "authority" instead of `Host`. That led to incorrect header modification. Fixed by manually assigning a new `Host` value to `authority` header once all transformations are finished.

Example:
`req_hdr_set host newvalue` - For http2 request with "authority" this had no effect, thus origin value of host was forwarded to backend.

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -3624,6 +3624,21 @@ tfw_h2_req_set_loc_hdrs(TfwHttpReq *req)
 		}
 	}
 
+	/*
+	 * When header modifications contains `req_hdr_set` rule for `Host`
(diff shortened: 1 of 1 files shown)


krizhanovsky (approved):
LGTM!

PR #2040, opened 2024-01-17
Fix error in matching custom uri brange

EvgeniiMekhanik:
We used incorrect token table, fix this error.
Also fix incorrect saving of eflags register
in debug macros.

### DIFF
--- a/fw/str_avx2.S
+++ b/fw/str_avx2.S
@@ -167,7 +167,7 @@ dbg_prefix_vec:
  */
 .extern __CUSTOM
 #define __CUST_TBL_0(t) $__CUSTOM+(t*0x40)
-#define __CUST_TBL_1(t) $__CUSTOM+(t*0x40+0x20)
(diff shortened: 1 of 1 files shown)


const-t (approved):

fw/str_avx2.S
```diff
@@ -195,13 +195,13 @@ dbg_prefix_vec:
 	push	%r9
 	push	%r10
 	push	%r11
-	push	%elfags
+	pushfq
```
krizhanovsky, line 198: According to the Intel manual vol 2, this is pushing `RFLAGS` with instruction `PUSHFQ`, why normal `PUSHF` isn't used?

krizhanovsky (approved):
Basically LGTM except the question about the `PUSHFQ` usage

fw/str_avx2.S
EvgeniiMekhanik: I use pushfq since we are in 64 bit mode, but since you right  PUSHF is enough. Fix.

PR #2041, opened 2024-01-23
Fix incorrect counting of `tp->copied_seq`

EvgeniiMekhanik:
We should continue to calculate `tp->copied_seq` in case of error during skb processing.


fw/sock.c
```diff
@@ -791,7 +796,7 @@ ss_tcp_process_skb(struct sock *sk, struct sk_buff *skb, int *processed)
 		 */
 		BUG_ON(conn == NULL);
 
-		if (SS_CONN_TYPE(sk) & Conn_Stop) {
+		if (SS_CONN_TYPE(sk) & Conn_Stop || r) {
```
const-t, line 799: `r` not reinitialized on beginning of iteration. skb on current iteration might be freed relies on result of previous iteration.

fw/sock.c
```diff
@@ -801,7 +806,6 @@ ss_tcp_process_skb(struct sock *sk, struct sk_buff *skb, int *processed)
 		if (r < 0) {
 			T_DBG2("[%d]: Processing error: sk=%pK r=%d\n",
 			       smp_processor_id(), sk, r);
-			goto out; /* connection must be dropped */
```
const-t, line 804: skb will leaks here if `connection_recv` < 0.

const-t (changes requested):

fw/sock.c
EvgeniiMekhanik: Yes and it is true. Previously in case of error we goto out and immediately return. But we can't do it because we should correctly update `tp->copied_seq`.  Now we continue to update  `tp->copied_seq` on each iteration after error but stop
processing any skb, because we catch error on previous iteration.

const-t (approved):
LGTM

fw/sock.c
```diff
@@ -812,12 +817,17 @@ ss_tcp_process_skb(struct sock *sk, struct sk_buff *skb, int *processed)
 		 * and the execution path should never reach here.
 		 */
 		BUG_ON(conn == NULL);
+
+		if (r) {
+			__kfree_skb(skb);
+			continue;
+		}
```
krizhanovsky, line 824: Basically this is just `ss_skb_queue_purge()` with `ADJUST_PROCESSED_SKB()` inside and I'd prefer to keep the code similar to the original version with `ss_skb_queue_purge()` to separate the main logic from the error path and make the main logic more clear and both the code paths fast (w.o. the additional checks in both the cases).

Please add a comment to the commit message why do we need to account `tp->copied_seq` on the error path, in other words what happens otherwise with the socket (it seems this leads to the hung socket on shutdown).

One more thing: there are other places where `ss_skb_queue_purge()` is called, so don't we need to `ADJUST_PROCESSED_SKB()` inside the function to fit all the cases? If not, then probably we can just copy and paste code from ``ss_skb_queue_purge()` to `ss_tcp_process_skb()` with the comment why do we need to account `tp->copied_seq`.

krizhanovsky (approved):
I just have a question and a code cleanup proposal, which are probably good to discuss or can be just fixed and merged

fw/tls.c
```diff
@@ -754,20 +761,20 @@ tfw_tls_conn_send(TfwConn *c, TfwMsg *msg)
...
 	/*
 	 * Only HTTP messages go this way, other (service) TLS records are sent
 	 * by tfw_tls_send().
 	 */
-	io->msgtype = TTLS_MSG_APPLICATION_DATA;
 	T_DBG("TLS %lu bytes (%u bytes, type %#x)"
 	      " are to be sent on conn=%pK/sk_write_xmit=%pK ready=%d\n",
-	      msg->len, io->msglen + TLS_HEADER_SIZE, io->msgtype, c,
-	      c->sk->sk_write_xmit, ttls_xfrm_ready(tls));
+	      msg->len, io->msglen + TLS_HEADER_SIZE, TTLS_MSG_APPLICATION_DATA,
```
krizhanovsky, line 771: This is debug messaging, but I believe there is no sense to log the constant - let's just remove it

fw/tls.c
```diff
@@ -754,20 +761,20 @@ tfw_tls_conn_send(TfwConn *c, TfwMsg *msg)
 {
 	int r;
 	TlsCtx *tls = tfw_tls_context(c);
-	TlsIOCtx *io = &tls->io_out;
 
 	/*
 	 * Only HTTP messages go this way, other (service) TLS records are sent
 	 * by tfw_tls_send().
 	 */
-	io->msgtype = TTLS_MSG_APPLICATION_DATA;
```
krizhanovsky, line 763: Why? I see that the TLS code does check `msgtype` against `TTLS_MSG_APPLICATION_DATA`, but now nobody sets the flag.

I do not understand what the commit fixes and and the fix seems buggy for me.

krizhanovsky (changes requested):
I request changes mostly because of https://github.com/tempesta-tech/tempesta/pull/2041/commits/1c96a3de66ccd537a33a4c15050f5e5449e8a763 - there were not bug described for the fix and the commit message is quite vague, so to review the patch I have to reveal the problem to understand the change:

- "We should set io->msgtype under tls->lock" - where the race appears and who corrupts the data consistency in the field? I could review the rest of the code in searching for the race condition, but I ask you describe such things in commit messages - it's easier for you since you already know the problem. Also it seems this is about `msg->ss_flags`, not `msgtype`...

- "in `tfw_tls_conn_send` we should not set it at all" - why? What's wrong with setting the flag in the function? Moreover, the function has comment about this "Only HTTP messages go this way, other (service) TLS records are sent by tfw_tls_send().", so it was done intntionally. If it was wrong then the comment should have been removed.

- "Also we should reset TTLS_F_ST_SHUTDOWN | TTLS_F_ST_CLOSE flags when `tfw_tls_send` fails." - again, why? What happens with the current code before the fix?

EvgeniiMekhanik:
My mistake, that I don't describe it more clear. There were two different BUGS. First of all
1. We should reset io->st_flags &= ~(TTLS_F_ST_SHUTDOWN | TTLS_F_ST_CLOSE); st_flags in case of error in `tfw_tls_send`. We use st_flags to specify what type of closing we choose. If this function fails we should reset st_flags.
2. We don't need to set io->msgtype in `tfw_tls_conn_send` it is used only to set ss_flags two lines below. There is a race between ttls_send_alert where we specify io->msgtype and `tfw_tls_conn_send`

fw/tls.c
EvgeniiMekhanik: fixed

fw/tls.c
```diff
@@ -593,6 +594,15 @@ tfw_tls_send(TlsCtx *tls, struct sg_table *sgt)
 	r = ss_send(conn->cli_conn.sk, &io->skb_list, flags);
 	WARN_ON_ONCE(!(flags & SS_F_KEEP_SKB) && io->skb_list);
 
+out:
+	/*
+	 * TTLS_F_ST_SHUTDOWN and TTLS_F_ST_CLOSE flags are used
+	 * to specify type of closing procedure. We save one of these
+	 * flags previosly in st_flags field, use it in this function
+	 * and reset them in st_flags field to be shure that we can
```
krizhanovsky, line 602:
```suggestion
	 * and reset them in st_flags field to be sure that we can
```

fw/tls.c
```diff
@@ -593,6 +594,15 @@ tfw_tls_send(TlsCtx *tls, struct sg_table *sgt)
...
+	 * use st_flags field for the same purpose.
+	 */
+	io->st_flags &= ~(TTLS_F_ST_SHUTDOWN | TTLS_F_ST_CLOSE);
```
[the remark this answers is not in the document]
EvgeniiMekhanik: These flags are used when we close client connection. They are used when we call `ttls_close_notify` or send  `ttls_send_alert`  with fatal message. For example we try to close connection with TTLS_F_ST_SHUTDOWN flag, because some request was incorrect. This function fails and we don't reset io->st_flags. Later when we handle error of this function we try to close connection immediately with TTLS_F_ST_CLOSE flag, so both flags in st_flags are set and we catch BUG_ON in `tfw_tls_close_msg_flags`.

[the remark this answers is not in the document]
krizhanovsky: OK, thank you. Please add this either to code comment or a commit message

krizhanovsky (approved):
LGTM

PR #2049, opened 2024-02-01
Increased size of `TfwHdrIndex`

const-t:
Size increased because we need to transfer more headers per message. From my point of view we must not
increase the size in future, maybe worth get rid of `TfwHttpHdrMap` using another approach. E.g iterating over headers table using bitmap.


krizhanovsky (approved):
LGTM

PR #2055, opened 2024-02-12
Fix handling extention frames.

### LINKED ISSUE #1824: h2spec: http2/5.5/2 - extension frames that appear in the middle of a field block
# Scope

1. Client send HEADERS frame with END_HEADERS - False;
2. Client send extensions frame.

ER: GOAWAY frame with PROTOCOLL_ERROR
AR: no response

See [h2spec](https://github.com/summerwind/h2spec/blob/af83a65f0b6273ef38bf778d400d98892e7653d8/http2/5_5_extending_http2.go#L41)

[RFC 9113 5.5](https://httpwg.org/specs/rfc9113.html#extensibility)

> Implementations MUST ignore unknown or unsupported values in all extensible protocol elements. Implementations MUST discard frames that have unknown or unsupported types. This means that any of these extension points can be safely used by extensions without prior arrangement or negotiation. However, extension frames that appear in the middle of a field block (Section 4.3) are not permitted; these MUST be treated as a connection error (Section 5.4.1) of type PROTOCOL_ERROR.

# Testing

[Test to reproduce](https://github.com/tempesta-tech/tempesta-test/blob/c485d6c32ae365ca10756f2bd3fc14bb3c5eb21a/http2_general/test_h2_specs.py#L118). Please enable test after fix.

EvgeniiMekhanik:
According RFC 9113 5.5:
Implementations MUST ignore unknown or unsupported values in all extensible protocol elements.
Implementations MUST discard frames that have
unknown or unsupported types. This means that any
of these extension points can be safely used by
extensions without prior arrangement or negotiation. However, extension frames that appear in the middle of a field block (Section 4.3) are not permitted;
these MUST be treated as a connection error
(Section 5.4.1) of type PROTOCOL_ERROR.
So we should check that extension frames don't appear in the middle of a field block, before ignore it.

Closes #1824

### DIFF
--- a/fw/http_frame.c
+++ b/fw/http_frame.c
@@ -1541,6 +1541,22 @@ tfw_h2_frame_type_process(TfwH2Ctx *ctx)
 		 */
 		T_DBG("HTTP/2: frame of unknown type '%u' received\n",
 		      hdr_type);
+		/*
+		 * According RFC 9113 5.5.
+		 * Implementations MUST ignore unknown or unsupported values
+		 * in all extensible protocol elements. Implementations MUST
+		 * discard frames that have unknown or unsupported types.
+		 * This means that any of these extension points can be safely
+		 * used by extensions without prior arrangement or negotiation.
+		 * However, extension frames that appear in the middle of a
+		 * field block (Section 4.3) are not permitted; these MUST be
+		 * treated as a connection error (Section 5.4.1) of type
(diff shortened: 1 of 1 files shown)


const-t (approved):
LGTM

krizhanovsky (approved):
LGTM

PR #2056, opened 2024-02-14
Fix setting mark for tempesta sockets.

### LINKED ISSUE #1987: Check that skb->mark is correctly forwarded to backend?
skb mark is not copied when skn is splitted in `tcp_write_xmit() -> tso_fragment()`. We should write tests for this case and check that all works fine.

EvgeniiMekhanik:
Kernel set skb->mark in `__ip_queue_xmit` using
sk->sk_mark, but for Tempesta sockets we can't
rely on this mechanizm. Tempesta can set skb->mark for some skbs. And moreover sk_mark is never set
for Tempesta sockets. Also we should copy skb->mark when we split skb in tso_fragment/tcp_fragment,
because skb can contain mark which was set by Tempesta.

Closes #1987

### DIFF
--- a/linux-5.10.35.patch
+++ b/linux-5.10.35.patch
@@ -740,7 +740,7 @@ index 000000000..80de50693
 + * Linux interface for Tempesta FW.
 + *
 + * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
-+ * Copyright (C) 2015-2024 Tempesta Technologies, Inc.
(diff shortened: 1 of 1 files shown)


krizhanovsky (commented):
Just see comment https://github.com/tempesta-tech/linux-5.10.35-tfw/pull/20#discussion_r1501877055

krizhanovsky (approved):
LGTM

Issue #1995, opened 2023-10-30
RFC compliance for caching of various request types


RomanBelozerov:
- [x] The response to a GET request is cacheable; a cache MAY use it to satisfy subsequent GET and HEAD requests. RFC 9110 9.3.1. Test to reproduce - `cache.test_cache.TestCacheMultipleMethodsHttp.test_HEAD_after_GET`
ER: Tempesta use cache for GET and HEAD requests
AR: Tempesta use cache only GET requests
- [x] Responses to POST requests are only cacheable when they include explicit freshness information and a Content-Location header field that has the same value as the POST's target URI. A cached POST response can be reused to satisfy a later GET or HEAD request. RFC 9110 9.3.3. 
Test to reproduce - `cache.test_cache.TestCacheMultipleMethodsHttp.test_cache_POST_request`
ER: Tempesta caches response to POST request and returns it for GET requests for same uri.
AR: Tempesta doesn't cache response to POST request.
- [x] If a successful PUT request passes through a cache that has one or more stored responses for the target URI, those stored responses will be invalidated. RFC 9110 9.3.4/9.3.5. 
Test to reproduce - `cache.test_cache.TestCacheMultipleMethodsHttp.test_update_cache_via_DELETE_request` and `cache.test_cache.TestCacheMultipleMethodsHttp.test_update_cache_via_PUT_request`
ER: Tempesta invalidate cache with same uri for GET and HEAD requests.
AR: Tempesta doesn't invalidate cache with same uri for GET and HEAD requests.
- [x] A 308 response is heuristically cacheable; i.e., unless otherwise indicated by the method definition or explicit cache controls. RFC 9110 15.4.9. 
Test to reproduce - `cache.test_cache.TestQueryParamsAndRedirectHttp.test_redirect_308_permanent_redirect`.
ER: Tempesta caches the 308 responses.
AR: Tempesta does not cache the 308 responses.
- [x] An origin server that evaluates an If-None-Match condition MUST NOT perform the requested method if the condition evaluates to false; instead, the origin server MUST respond with either a) the [304 (Not Modified)](https://www.rfc-editor.org/rfc/rfc9110.html#status.304) status code if the request method is GET or HEAD or b) the [412 (Precondition Failed)](https://www.rfc-editor.org/rfc/rfc9110.html#status.412) status code for all other request methods. RFC 9110 13.1.2. 
[...]

PR #2062, opened 2024-02-22
Use response to a GET request to satisfy HEAD request

EvgeniiMekhanik:
According to RFC 9110 9.3.1:
The response to a GET request is cacheable; a
cache MAY use it to satisfy subsequent GET and
HEAD requests unless otherwise indicated by the
Cache-Control header field.
According to RFC 9110 9.3.2:
The HEAD method is identical to GET except that
the server MUST NOT send content in the response.
So Tempesta should not send body when cached GET
response used to satisfy HEAD request.

Part of #1995


const-t (changes requested):
This PR has 2 issues:
1. cache.c:2839: `resp->content_length = ce->body_len;`. We must update `resp->content_length` depending on the method. For HEAD it will be 0. Otherwise access log will contain incorrect size of response.
2. When Tempesta responds to HEAD request we must not send `Content-Length` header or send `Content-Length: 0`. Now will be sent `Content-Length` that equals to size of the response. The same behaviour with cache and without it.

EvgeniiMekhanik:
> This PR has 2 issues:
> 
>     1. cache.c:2839: `resp->content_length = ce->body_len;`. We must update `resp->content_length` depending on the method. For HEAD it will be 0. Otherwise access log will contain incorrect size of response.
> 
>     2. When Tempesta responds to HEAD request we must not send `Content-Length` header or send `Content-Length: 0`. Now will be sent `Content-Length` that equals to size of the response. The same behaviour with cache and without it.

I don't agree with you here. https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD Content-Length might present and indicate size of body in case if it was GET request.

const-t (approved):

PR #2063, opened 2024-02-22
Fixed caching response to POST request.

EvgeniiMekhanik:
According to RFC 9110 9.3.3:
Responses to POST requests are only cacheable
when they include explicit freshness information
and a Content-Location header field that has the
same value as the POST's target URI. A cached POST response can be reused to satisfy a later GET or
HEAD request. In contrast, a POST request cannot
be satisfied by a cached POST response because
POST is potentially unsafe.

Part of #1995

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -53,10 +53,10 @@
 static const int tfw_cache_spec_headers_304[] = {
 	[0 ... TFW_HTTP_HDR_RAW] = 0,
 	[TFW_HTTP_HDR_ETAG] = 1,
+	[TFW_HTTP_HDR_CONTENT_LOCATION] = 1,
 };
 static const TfwStr tfw_cache_raw_headers_304[] = {
 	TFW_STR_STRING("cache-control:"),
-	TFW_STR_STRING("content-location:"),
 	TFW_STR_STRING("date:"),
 	TFW_STR_STRING("expires:"),
 	TFW_STR_STRING("last-modified:"),
@@ -476,9 +476,11 @@ tfw_cache_employ_resp(TfwHttpResp *resp)
 #define CC_RESP_DONTCACHE				\
 	(TFW_HTTP_CC_NO_STORE | TFW_HTTP_CC_PRIVATE	\
 	 | TFW_HTTP_CC_NO_CACHE)
-#define CC_RESP_CACHEIT					\
+#define CC_RESP_EXPLICIT_FRESH_INFO			\
 	(TFW_HTTP_CC_HDR_EXPIRES | TFW_HTTP_CC_MAX_AGE	\
-	 | TFW_HTTP_CC_S_MAXAGE | TFW_HTTP_CC_PUBLIC)
+	 | TFW_HTTP_CC_S_MAXAGE)
+#define CC_RESP_CACHEIT					\
+	(CC_RESP_EXPLICIT_FRESH_INFO | TFW_HTTP_CC_PUBLIC)
 /*
  * RFC 9111 3.5:
  * A shared cache MUST NOT use a cached response to a request with an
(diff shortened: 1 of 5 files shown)


fw/cache.c
```diff
@@ -511,9 +511,34 @@ tfw_cache_employ_resp(TfwHttpResp *resp)
...
+	 * they include a Content-Location header field that
+	 * has the same value as the POST's target URI.
+	 */
+	if (req->method == TFW_HTTP_METH_POST) {
+		TfwStr *h = &resp->h_tbl->tbl[TFW_HTTP_HDR_CONTENT_LOCATION];
+		TfwStr h_val;
+
+		tfw_http_msg_srvhdr_val(h, TFW_HTTP_HDR_CONTENT_LOCATION,
+					&h_val);
+		if (tfw_strcmp(&h_val, &req->uri_path))
```
const-t, line 538: This is sophisticated part and I'm not sure, that we must implement it strictly to RFC, but I must draw your attention. URI in this case might be in two forms absolute-URI and partial-URI, if request had partial-URI, but `content-location` in response has absolute-URI in current implementation Tempesta won't cache such response. However, rfc says that we should transform partial-URI to absolute before comparison:

> If Content-Location is included in a 2xx (Successful) response message and its value refers (after conversion to absolute form)  to a URI that is the same as the target URI

This conversion also not obvious to me, because absolute form of URI has scheme information and there is question what scheme we must use during this conversion? And another question what we must do if both URI have absolute form with different schemes(http vs https)?

fw/http_parser.c
```diff
@@ -12311,6 +12331,11 @@ tfw_http_parse_resp(void *resp_data, unsigned char *data, unsigned int len,
 				   __parse_content_length,
 				   TFW_HTTP_HDR_CONTENT_LENGTH);
 
+	/* 'Content-Location:*OWS' is read, process field-value. */
+	TFW_HTTP_PARSE_SPECHDR_VAL(Resp_HdrContent_LocationV, msg,
+				   __parse_content_location,
+				   TFW_HTTP_HDR_CONTENT_LOCATION);
```
const-t, line 12337: Please update `HP_ENTRY_NAME` for this header in hpack.c.

fw/cache.c
```diff
@@ -511,9 +511,34 @@ tfw_cache_employ_resp(TfwHttpResp *resp)
 	if ((req->cache_ctl.flags & TFW_HTTP_CC_HDR_AUTHORIZATION)
 	    && !(effective_resp_flags & CC_RESP_AUTHCAN))
 		return false;
-	if (!(effective_resp_flags & CC_RESP_CACHEIT)
-	    && !tfw_cache_status_bydef(resp))
-		return false;
+	if (!(effective_resp_flags & CC_RESP_CACHEIT)) {
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed

fw/http_parser.c
EvgeniiMekhanik: According to our private discussion we don't need this, because we don't parse this header for requests

fw/cache.c
EvgeniiMekhanik: req->uri_path contains relative URI. Yes, in this simple implementation we rely on the fact that server sends us response with Content-Location with appropriate relative format. If it is not so we don';t cache response. @krizhanovsky should I implement Content-Location parser and comparison as it described in RFC?

keshonok: Perhaps, there's the fourth solution: make this kind of behavior configurable to allow either one of these three strategies. The downside of this is that there may be too many configurable options already which adds to the complexity of Tempesta's use.

On this thought, it might be interesting to collect the configurations from real users/admins (anonymously, of course) to see what options are really used. That may lead to a less complex product. It's like simplicity and minimalism of Apple products versus the magnitude of options of Android systems. :-)

krizhanovsky: For some time ago we researched abosolute URIs in the wild and no legitimate clients actually produce such requests. IIRC mostly such requests are used for forward proxies, but here I could be wrong. In either case we already made some steps in not supporting absolute URLs. I thing we're good here with the current implementation.

I'd only add a good comment what RFC requires, what do we do and why.

fw/http.h
```diff
@@ -203,6 +203,7 @@ typedef enum {
 	TFW_HTTP_HDR_REGULAR,
 	TFW_HTTP_HDR_HOST = TFW_HTTP_HDR_REGULAR,
 	TFW_HTTP_HDR_CONTENT_LENGTH,
+	TFW_HTTP_HDR_CONTENT_LOCATION,
```
krizhanovsky, line 206: Please add comment that this is response specific - with `tfw_http_msg_header_table_size()` reworking we may have further confusions with the enum and maybe we'll need to go to two enums, for requests and responses.

krizhanovsky (approved):
LGTM

PR #2064, opened 2024-02-22
Add response with 308 status code to heuristically cacheable

EvgeniiMekhanik:
According to new RFC 9110 15.4.9:
A 308 response is heuristically cacheable; i.e., unless otherwise indicated by the method definition or explicit cache controls.

Part of #1995


const-t (approved):
LGTM

PR #2068, opened 2024-02-27
Fix deadlock and memory leak.

EvgeniiMekhanik:
`tfw_cfgop_update_srv` is called under `sg_sem`,
so we should not lock this semaphore when we lookup server.


krizhanovsky (approved):
LGTM

const-t (approved):
LGTM

PR #2072, opened 2024-03-04
Remove unnecessary warnings

EvgeniiMekhanik:
We should not print warning messages
in case when `ttls_ciphersuite_match`
fails, it is a normal situation, because
we iterate over all prossible ciphersuites
to choose appropriate one.


krizhanovsky (approved):
LGTM

Issue #500, opened 2016-05-23
[Cache] Reducing origin server requests


krizhanovsky:
Currently thundering herd problem is possible when many clients try to request the same resource, in this case Tempesta generates as many request to upstream as it receives from all the clients. Rather it should accurately track requests on the fly and made only one active request for upstream.

krizhanovsky:
The implementation must handle both the cases for absent and invalid cache entries. We need a generic protection against tundering herd, when many equal requests efficiently passes through the cache to a backend server. The only one request must go to a backend server, while others must be postponed.

Technically it could be special `in-progress` cache entry with a list of requests generating the entry. E.g. if there is no cache entry, then we just create a new one and add the current request to it's list and forward the request to a backend. If following requests find the cache entry, then they just add themelves to the cache entry list and a current routine returns (probably with TFW_POSTPONE or TFW_PASS code). When a response comes, we traverse the list and send the response for the each request (we should just increase page reference counters instead of real data copying).

Since we need to modify cache entries, the issue depends on #788

ikoveshnikov:
Addition to comment above:
Response may not come in time and an error response may be generated. This may happen if the request was evicted for some reason: request timeout, retry timeout or even client disconnection. In this case the error response must be send to client (if applied), other responses for the same cache entry must be evicted if stale, then the latest request must be forwarded to backend server. If all request was evicted for the cache entry, cache entry should be purged.

krizhanovsky:
Recently we observed not enough memory crash on TDB side due to this issue

PR #2074, opened 2024-03-08
TDB eviction and stale response processing

const-t:
Wiki branch related to this PR is `kt-use-stale-desc`

### DIFF
--- a/db/core/htrie.c
+++ b/db/core/htrie.c
@@ -34,6 +34,7 @@
 #include "lib/str.h"
 #include "htrie.h"
 
+#define TDB_MAX_PCP_SZ  (TDB_EXT_SZ / PAGE_SIZE) /* Store one extent per cpu. */
 #define TDB_MAGIC	0x434947414D424454UL /* "TDBMAGIC" */
 #define TDB_BLK_SZ	PAGE_SIZE
 #define TDB_BLK_MASK	(~(TDB_BLK_SZ - 1))
@@ -105,24 +106,45 @@ tdb_get_reftbl(TdbHdr *dbh, TdbExt *e)
 }
 
 /**
- * Add block to per-CPU freelist.
- *
- * NOTE Due to per-CPU nature of freelist, block can be allocated only on CPU
- * where it was freed. It can cause allocation failures in case when one
- * CPU frees more blocks then others and free blocks available only in per-CPU
- * freelists because main memory pool is exhausted.
+ * Add block to global freelist.
+ */
+static void
+tdb_ga_freelist_push(TdbHdr *dbh, unsigned long ptr, unsigned long *page_addr)
+{
+	spin_lock_bh(&dbh->gfl_lock);
+	if (!dbh->ga_freelist) {
+		dbh->ga_freelist = ptr;
+		*page_addr = 0;
+	} else {
+		*page_addr = dbh->ga_freelist;
+		dbh->ga_freelist = ptr;
+	}
+	spin_unlock_bh(&dbh->gfl_lock);
+}
+
+/**
+ * Add block to per-CPU or global freelist.
  */
 static void
 tdb_freelist_push(TdbHdr *dbh, unsigned long ptr)
 {
 	unsigned long *page_addr = TDB_PTR(dbh, ptr);
+	TdbPerCpu *pcpu = this_cpu_ptr(dbh->pcpu);
 
-	if (!this_cpu_ptr(dbh->pcpu)->freelist) {
-		this_cpu_ptr(dbh->pcpu)->freelist = ptr;
+	if (!pcpu->freelist) {
+		pcpu->freelist = ptr;
+		pcpu->fl_size = 1;
 		*page_addr = 0;
+		return;
+	}
+
+	/* If per-cpu list is full, push block to global list. */
+	if (pcpu->fl_size >= TDB_MAX_PCP_SZ) {
+		tdb_ga_freelist_push(dbh, ptr, page_addr);
 	} else {
-		*page_addr = this_cpu_ptr(dbh->pcpu)->freelist;
-		this_cpu_ptr(dbh->pcpu)->freelist = ptr;
+		*page_addr = pcpu->freelist;
+		pcpu->freelist = ptr;
+		pcpu->fl_size++;
 	}
 }
 
@@ -327,12 +349,63 @@ repeat:
 }
 
 static unsigned long
+tdb_alloc_blk_pcp_freelist(TdbHdr *dbh)
+{
+	unsigned long rptr, *next;
+	TdbPerCpu *pcpu = this_cpu_ptr(dbh->pcpu);
+
+	rptr = pcpu->freelist;
+	if (!rptr)
+		return 0;
+
+	next = TDB_PTR(dbh, rptr);
+	pcpu->freelist = *next;
+	pcpu->fl_size--;
+	tdb_get_blk(dbh, rptr);
+
+	return rptr;
+}
+
+static unsigned long
+tdb_alloc_blk_global_freelist(TdbHdr *dbh)
+{
+	unsigned long rptr = 0, *next;
+
+	spin_lock_bh(&dbh->gfl_lock);
+	rptr = dbh->ga_freelist;
+	if (!rptr) {
+		spin_unlock_bh(&dbh->gfl_lock);
+		return 0;
+	}
+
+	next = TDB_PTR(dbh, rptr);
+	dbh->ga_freelist = *next;
+	spin_unlock_bh(&dbh->gfl_lock);
+	tdb_get_blk(dbh, rptr);
+
+	return rptr;
+}
+
+static unsigned long
 tdb_alloc_blk(TdbHdr *dbh)
 {
 	TdbExt *e;
 	long g_nwb, rptr, next_blk;
 
+	rptr = tdb_alloc_blk_pcp_freelist(dbh);
+	if (rptr)
+		return rptr;
+
 retry:
+	if (dbh->oom) {
+		rptr = tdb_alloc_blk_global_freelist(dbh);
+		if (rptr)
+			return rptr;
+
+		TDB_ERR("out of free space\n");
+		return 0;
+	}
+
 	g_nwb = atomic64_read(&dbh->nwb);
 	e = tdb_ext(dbh, TDB_PTR(dbh, g_nwb));
 
@@ -365,8 +438,8 @@ retry:
 	if (unlikely(TDB_HTRIE_OFF(dbh, e) == dbh->dbsz)) {
 		/* We do this set because we skip the last page in extent. */
 		atomic64_set(&dbh->nwb, dbh->dbsz);
-		TDB_ERR("out of free space\n");
-		return 0;
+		dbh->oom = true;
+		goto retry;
 	}
 	BUG_ON(TDB_HTRIE_OFF(dbh, e) > dbh->dbsz);
 	set_bit(TDB_EXT_ID(TDB_EXT_BASE(dbh, e)), dbh->ext_bmp);
@@ -408,22 +481,6 @@ tdb_htrie_init_bucket(TdbBucket *b)
 #endif
 }
 
-static unsigned long
-tdb_alloc_blk_freelist(TdbHdr *dbh)
-{
-	unsigned long rptr, *next;
-
-	rptr = this_cpu_ptr(dbh->pcpu)->freelist;
-	if (!rptr)
-		return 0;
-
-	next = TDB_PTR(dbh, rptr);
-	this_cpu_ptr(dbh->pcpu)->freelist = *next;
-
-	tdb_get_blk(dbh, rptr);
-	return rptr;
-}
-
 /**
  * @return byte offset of the allocated data block and sets @len to actually
  * available room for writing if @len doesn't fit to block.
@@ -459,14 +516,7 @@ tdb_alloc_data(TdbHdr *dbh, size_t *len)
 		size_t max_data_len;
 
 		old_rptr = ((rptr - 1) & TDB_BLK_MASK);
-		if (this_cpu_ptr(dbh->pcpu)->freelist)
-			rptr = tdb_alloc_blk_freelist(dbh);
-		else
-			/*
-			 * Use a new page and/or extent for the data.
-			 * Less than a page can be allocated.
-			 */
-			rptr = tdb_alloc_blk(dbh);
+		rptr = tdb_alloc_blk(dbh);
 
 		if (!rptr)
 			goto out;
@@ -517,9 +567,6 @@ tdb_alloc_bucket(TdbHdr *dbh)
 
 	if (unlikely(!(rptr & ~TDB_BLK_MASK)
 		     || TDB_BLK_O(rptr + len - 1) > TDB_BLK_O(rptr))) {
-		if (this_cpu_ptr(dbh->pcpu)->freelist)
-			rptr = tdb_alloc_blk_freelist(dbh);
-		else
 			/* Use a new page and/or extent for local CPU. */
 			rptr = tdb_alloc_blk(dbh);
 		if (!rptr)
@@ -557,9 +604,6 @@ tdb_alloc_index(TdbHdr *dbh)
 		     || TDB_BLK_O(rptr + sizeof(TdbHtrieNode) - 1)
 			> TDB_BLK_O(rptr)))
 	{
-		if (this_cpu_ptr(dbh->pcpu)->freelist)
-			rptr = tdb_alloc_blk_freelist(dbh);
-		else
 			/* Use a new page and/or extent for local CPU. */
 			rptr = tdb_alloc_blk(dbh);
 		if (!rptr)
@@ -1358,6 +1402,8 @@ tdb_htrie_init(void *p, size_t db_size, unsigned int rec_len)
 			return NULL;
 		}
 	}
+
+	spin_lock_init(&hdr->gfl_lock);
 
 	/* Set per-CPU pointers. */
 	hdr->pcpu = alloc_percpu(TdbPerCpu);
--- a/db/core/htrie.c
+++ b/db/core/htrie.c
@@ -449,7 +449,7 @@ out:
 static unsigned long
 tdb_alloc_index(TdbHdr *dbh)
 {
-	unsigned long rptr = 0, old_rptr;
+	unsigned long rptr = 0;
 
 	local_bh_disable();
 
@@ -459,7 +459,6 @@ tdb_alloc_index(TdbHdr *dbh)
 		     || TDB_BLK_O(rptr + sizeof(TdbHtrieNode) - 1)
 			> TDB_BLK_O(rptr)))
 	{
-		old_rptr = ((rptr - 1) & TDB_BLK_MASK);
 		if (this_cpu_ptr(dbh->pcpu)->freelist)
 			rptr = tdb_alloc_blk_freelist(dbh);
 		else
--- a/db/core/htrie.c
+++ b/db/core/htrie.c
@@ -171,6 +171,9 @@ tdb_free_vsrec(TdbHdr *dbh, TdbVRec *rec)
 static void
 tdb_htrie_free_rec(TdbHdr *dbh, TdbRec *rec)
 {
+	if (dbh->before_free)
+		dbh->before_free(rec);
+
 	if (TDB_HTRIE_VARLENRECS(dbh)) {
 		TdbVRec *next, *curr = (TdbVRec *)rec;
 
@@ -932,7 +935,7 @@ tdb_bucket_remove_record(TdbHdr *dbh, TdbBucket *bckt, tdb_eq_cb_t *eq_cb,
 /* Must be called with locked bucket. */
 static TdbBucket *
 __tdb_htrie_remove(TdbHdr *dbh, TdbBucket *bckt, tdb_eq_cb_t *eq_cb, void *data,
-		   tdb_before_remove_cb_t *bf_remove_cb, bool force)
+		   bool force)
 {
 	TdbBucket *prev = bckt;
 
@@ -953,7 +956,7 @@ __tdb_htrie_remove(TdbHdr *dbh, TdbBucket *bckt, tdb_eq_cb_t *eq_cb, void *data,
 
 void
 tdb_htrie_remove(TdbHdr *dbh, unsigned long key, tdb_eq_cb_t *eq_cb, void *data,
-		 tdb_before_remove_cb_t *bf_remove_cb, bool force)
+		 bool force)
 {
 	int bits = 0;
 	unsigned long o;
@@ -970,7 +973,7 @@ tdb_htrie_remove(TdbHdr *dbh, unsigned long key, tdb_eq_cb_t *eq_cb, void *data,
 		force, bckt);
 
 	write_lock_bh(&bckt->lock);
-	bckt = __tdb_htrie_remove(dbh, bckt, eq_cb, data, bf_remove_cb, force);
+	bckt = __tdb_htrie_remove(dbh, bckt, eq_cb, data, force);
 	write_unlock_bh(&bckt->lock);
 }
 
@@ -1196,13 +1199,6 @@ retry:
 	write_unlock_bh(&bckt->lock);
 
 	goto retry;
-}
-
-static inline void
-tdb_eval_bf_remove_cb(tdb_before_remove_cb_t *bf_remove_cb, TdbRec *rec)
-{
-	if (bf_remove_cb)
-		bf_remove_cb(rec);
 }
 
 TdbBucket *
--- a/db/core/htrie.c
+++ b/db/core/htrie.c
@@ -979,9 +979,16 @@ retry:
 }
 
 static inline bool
-tdb_eval_eq_cb(bool (*eq_cb)(void *, void *), TdbRec *rec, void *data)
+tdb_eval_eq_cb(tdb_eq_cb_t *eq_cb, TdbRec *rec, void *data)
 {
 	return ((eq_cb && eq_cb(rec, data)) || !eq_cb);
+}
+
+static inline void
+tdb_eval_bf_remove_cb(tdb_before_remove_cb_t *bf_remove_cb, TdbRec *rec)
+{
+	if (bf_remove_cb)
+		bf_remove_cb(rec);
 }
 
 #define TDB_REMOVE_FOREACH_REC(body)					\
@@ -999,8 +1006,8 @@ tdb_eval_eq_cb(bool (*eq_cb)(void *, void *), TdbRec *rec, void *data)
 		 <= TDB_HTRIE_MINDREC)
 
 int
-tdb_htrie_remove(TdbHdr *dbh, unsigned long key, bool (*eq_cb)(void *, void *),
-		 void *data, bool force)
+tdb_htrie_remove(TdbHdr *dbh, unsigned long key, tdb_eq_cb_t *eq_cb, void *data,
+		 tdb_before_remove_cb_t *bf_remove_cb, bool force)
 {
 	int bits = 0;
 	TdbRec *rec;
@@ -1031,6 +1038,7 @@ tdb_htrie_remove(TdbHdr *dbh, unsigned long key, bool (*eq_cb)(void *, void *),
 			if (tdb_live_rec(dbh, rec)
 			    && tdb_eval_eq_cb(eq_cb, rec, data))
 			{
+				tdb_eval_bf_remove_cb(bf_remove_cb, rec);
 				tdb_htrie_free_rec(dbh, rec);
 				r = 0;
 			}
@@ -1053,6 +1061,8 @@ tdb_htrie_remove(TdbHdr *dbh, unsigned long key, bool (*eq_cb)(void *, void *),
 			TDB_REMOVE_FOREACH_REC({
 				if (tdb_live_rec(dbh, rec)) {
 					if (tdb_eval_eq_cb(eq_cb, rec, data)) {
+						tdb_eval_bf_remove_cb(bf_remove_cb,
+								      rec);
 						tdb_htrie_free_rec(dbh, rec);
 						r = 0;
 					} else {
--- a/db/core/htrie.c
+++ b/db/core/htrie.c
@@ -57,17 +57,20 @@ typedef struct {
 	unsigned int	shifts[TDB_HTRIE_FANOUT];
 } __attribute__((packed)) TdbHtrieNode;
 
+#define TDB_HTRIE_FOREACH_REC_SMALL(d, b, fr, r, t)			\
+	for (r = TDB_HTRIE_BCKT_1ST_REC(dbh, b), fr = r;		\
+	     r && (char *)r - (char *)fr + sizeof(t) <= TDB_HTRIE_MINDREC \
+	     && (char *)r - (char *)fr + TDB_HTRIE_RECLEN(d, r)		\
+	     <= TDB_HTRIE_MINDREC;					\
+	     r = (typeof(r))((char *)r + TDB_HTRIE_RECLEN(d, r)))
+
 /**
  * Unlocked and simplified version of tdb_htrie_bscan_for_rec() for
  * small records only.
  */
-#define TDB_HTRIE_FOREACH_REC_UNLOCKED(d, b, r, t)			\
+#define TDB_HTRIE_FOREACH_REC_UNLOCKED(d, b, fr, r, t)			\
 	for ( ; b; b = TDB_HTRIE_BUCKET_NEXT(d, b))			\
-		for (r = TDB_HTRIE_BCKT_1ST_REC(b);			\
-		     (char *)r - (char *)b + sizeof(t)	<= TDB_HTRIE_MINDREC \
-		     && (char *)r - (char *)b + TDB_HTRIE_RECLEN(d, r)	\
-			<= TDB_HTRIE_MINDREC;				\
-		     r = (typeof(r))((char *)r + TDB_HTRIE_RECLEN(d, r)))
+		TDB_HTRIE_FOREACH_REC_SMALL(d, b, fr, r, t)
 
 
 static inline TdbExt *
@@ -131,7 +134,7 @@ tdb_get_blk(TdbHdr *dbh, unsigned long ptr)
 	unsigned int blkoff = TDB_BLK_ID(ptr & TDB_BLK_MASK) >> PAGE_SHIFT;
 	atomic_t *refcnt = reftbl + blkoff;
 
-	BUG_ON((void*) refcnt > TDB_PTR(dbh,(TDB_EXT_BASE(dbh, e) + TDB_EXT_SZ)));
+	BUG_ON((void *) refcnt > TDB_PTR(dbh, (TDB_EXT_BASE(dbh, e) + TDB_EXT_SZ)));
 	atomic_inc(refcnt);
 }
 
@@ -149,6 +152,69 @@ tdb_put_blk(TdbHdr *dbh, unsigned long ptr)
 		tdb_freelist_push(dbh, ptr & TDB_BLK_MASK);
 
 	BUG_ON(pgref < 0);
+}
+
+/* Zero key for fixed size record indicates that record is freed. */
+static inline void
+tdb_free_fsrec(TdbHdr *dbh, TdbFRec *rec)
+{
+	rec->key = 0;
+	tdb_put_blk(dbh, TDB_OFF(dbh, rec));
+}
+
+static inline void
+tdb_free_vsrec(TdbHdr *dbh, TdbVRec *rec)
+{
+	tdb_put_blk(dbh, TDB_OFF(dbh, rec));
+}
+
+static void
+tdb_htrie_free_rec(TdbHdr *dbh, TdbRec *rec)
+{
+	if (TDB_HTRIE_VARLENRECS(dbh)) {
+		TdbVRec *next, *curr = (TdbVRec *)rec;
+
+		while (curr) {
+			next = curr->chunk_next
+				? TDB_PTR(dbh, TDB_DI2O(curr->chunk_next))
+				: 0;
+
+			tdb_free_vsrec(dbh, curr);
+			curr = next;
+		}
+	} else {
+		tdb_free_fsrec(dbh, (TdbFRec *)rec);
+	}
+}
+
+void
+tdb_htrie_get_rec(TdbRec *rec)
+{
+	atomic_inc(&rec->refcnt);
+}
+
+void
+tdb_htrie_put_rec(TdbHdr *dbh, TdbRec *rec)
+{
+	int refcnt = atomic_dec_return(&rec->refcnt);
+
+	BUG_ON(refcnt < 0);
+	if (!refcnt)
+		tdb_htrie_free_rec(dbh, rec);
+}
+
+/* Call only under lock. */
+static void
+tdb_rec_set_remove(TdbRec *rec)
+{
+	rec->flags |= TDB_HTRIE_REC_REMOVED_BIT;
+}
+
+/* Call only under lock. */
+static bool
+tdb_rec_is_removed(TdbRec *rec)
+{
+	return rec->flags & TDB_HTRIE_REC_REMOVED_BIT;
 }
 
 static TdbHdr *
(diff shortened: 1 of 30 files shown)


fw/cache.c
```diff
@@ -2431,7 +2446,11 @@ __cache_add_node(TDB *db, TfwHttpResp *resp, unsigned long key)
 	T_DBG3("%s: db=[%p] resp=[%p], req=[%p], key='%lu', data_len='%ld'\n",
 	       __func__, db, resp, resp->req, key, data_len);
 
-	/* TODO #522: revalidate existing entries before inserting a new one. */
+	/*
+	 * Remove each entry with current key, even fresh entries. We assume
+	 * that if we at current place, no valid entries are exists.
+	 */
+	tdb_entry_remove(db, key, &tfw_cache_rec_eq_req, resp->req, false);
```
krizhanovsky, line 2453: Isn't this exactly the invalidation (removal) of previously cached responses as discussed in https://github.com/tempesta-tech/tempesta-test/pull/602/files#r1628991786 ?

In particular, I'm wondering if we evict old entries on a test like
```sh
while :; do curl --http2 -ik -o /dev/null -D /dev/stdout -H 'Cache-Control: max-age=2' -H 'Host: tempesta-tech.com' https://tempesta-tech.com/blog/building-your-own-wordpress-staging-with-tempesta-fw/ ; sleep 3; done
```

To understand whether this (or any similar scenario, which we can imagine) scenario floods our cache could you please introduce a couple of counters for `/proc/tempesta/perfstat`:

- `Cache objects` - the number of cached objects, must be decreased on the removal
- `Cache bytes` - number of bytes at the moment stored in the cache

db/core/htrie.c
```diff
@@ -86,6 +86,71 @@ tdb_set_bit(unsigned long *bmp, unsigned int nr)
...
+		this_cpu_ptr(dbh->pcpu)->freelist = ptr;
+	}
+}
+
+static void
+tdb_get_blk(TdbHdr *dbh, unsigned long ptr)
+{
+	TdbExt *e = tdb_ext(dbh, TDB_PTR(dbh, ptr));
+	atomic_t *reftbl = tdb_get_reftbl(dbh, e);
+	unsigned int blkoff = TDB_BLK_ID(ptr & TDB_BLK_MASK) >> 12;
```
EvgeniiMekhanik, line 131: What is `12` here? Can you please implement define for this.

db/core/htrie.c
```diff
@@ -241,6 +315,7 @@ tdb_alloc_blk(TdbHdr *dbh)
 	/*
 	 * Align offsets of new blocks for data records.
 	 * This is only for first blocks in extents, so we lose only
+	 *
```
EvgeniiMekhanik, line 324: Extra space

fw/cache.c
```diff
@@ -1236,14 +1236,10 @@ tfw_cache_dbce_get(TDB *db, TdbIter *iter, TfwHttpReq *req)
...
-		 * (e.g. the bucket can not contain both the stale and fresh
-		 * entries) and rework the loop. */
-		if (tfw_cache_entry_key_eq(db, req, ce, false)) {
-			if (tfw_cache_entry_is_live(req, ce)
-			    && (!ce_best
-				|| tfw_cache_entry_age(ce)
-				   < tfw_cache_entry_age(ce_best)))
+		if (tfw_cache_entry_key_eq(db, req, ce)) {
+			if (!ce_best
+				   || tfw_cache_entry_age(ce)
```
EvgeniiMekhanik, line 1259: Invalid code shift

fw/vhost.c
```diff
@@ -904,6 +926,84 @@ tfw_mod_hdr_sort(TfwLocation *loc)
...
+
+static int
+tfw_cfgop_cache_use_stale(TfwCfgSpec *cs, TfwCfgEntry *ce, TfwLocation *loc)
+{
+	TfwCacheUseStale *cfg;
+	size_t i;
+	int n = 0;
+	bool mask5x = false, mask4x = false;
+
+	if (ce->attr_n) {
```
EvgeniiMekhanik, line 947: TFW_CFG_CHECK_NO_ATTRS, TFW_CFG_CHECK_VAL_N should be used for code consistance

fw/cache.c
```diff
@@ -3029,13 +3036,26 @@ cache_req_process_node(TfwHttpReq *req, tfw_http_cache_cb_t action)
 	}
 
 	resp = tfw_cache_build_resp(req, ce, lifetime);
+
+	if (resp && stale) {
+		req->resp = NULL;
+		req->stale_resp = resp;
```
EvgeniiMekhanik, line 3127: I don't agree with this commit. What purpose to send request to server and after receiving response send stale response?? It is very strange for me! If we receive new fresh response we should send it. If this directive is set we should send stale response to client without accessing backend. We should check status of response saved in cache and if it is in  `cache_use_stale` send it without accessing backend. For error cases we can send stale successful responses after receiving error from backend.
For example if I add  
`param(name="200", status=200, use_stale="200", r2_hdrs=[]),` to TestCacheUseStale in your branch. I receive stale 200 response after receiving successful response from server!

fw/cache.c
```diff
@@ -2442,13 +2460,52 @@ __cache_entry_size(TfwHttpResp *resp)
...
+	 * Remove only records that have same method, except when request is
+	 * POST or PURGE. Accordingly to RFC 9111 4.4. POST is used to
+	 * invalidate cache records.
+	 *
+	 * We don't check staleness at this stage, just to skip double check.
+	 * The first check happens when we trying to satisfy request from the
+	 * cache. Therefore if reached this point it means that record is stale
+	 * or not presented in the cache.
+	 */
+	if (req->method != TFW_HTTP_METH_POST &&
```
EvgeniiMekhanik, line 2479: We should also do it for DELETE and PUT

fw/cache.c
```diff
@@ -2472,7 +2529,12 @@ __cache_add_node(TDB *db, TfwHttpResp *resp, unsigned long key)
 	T_DBG3("%s: db=[%p] resp=[%p], req=[%p], key='%lu', data_len='%ld'\n",
 	       __func__, db, resp, resp->req, key, data_len);
 
-	/* TODO #522: revalidate existing entries before inserting a new one. */
+	/*
+	 * Remove each entry with current key, even fresh entries. We assume
+	 * that if we at current place, no valid entries are exists.
+	 */
+	tdb_entry_remove(db, key, &tfw_cache_rec_eq_req, resp->req,
```
EvgeniiMekhanik, line 2536: This function should be also called in `tfw_cache_process` instead of `tfw_cache_invalidate` as I think:
```
if (unlikely(req->method == TFW_HTTP_METH_PUT
			     || req->method == TFW_HTTP_METH_DELETE
			     || req->method == TFW_HTTP_METH_POST
			     || tfw_http_req_is_nip(req))
		    && cache_cfg.cache
		    && resp->status >= 200 && resp->status < 400)
			tfw_cache_invalidate(req);
```

EvgeniiMekhanik (changes requested):

fw/http_parser.c
```diff
@@ -7239,6 +7261,7 @@ do {									\
 	SET_HDR_CACHE_CONTROL_FIELD(max_age, TFW_HTTP_CC_MAX_AGE);
 	SET_HDR_CACHE_CONTROL_FIELD(max_stale, TFW_HTTP_CC_MAX_STALE);
 	SET_HDR_CACHE_CONTROL_FIELD(min_fresh, TFW_HTTP_CC_MIN_FRESH);
+	req->cache_ctl.stale_if_error = cstate->cache_ctl.stale_if_error;
```
EvgeniiMekhanik, line 7264: SET_HDR_CACHE_CONTROL_FIELD(stale_if_error, TFW_HTTP_CC_STALE_IF_ERROR);

fw/http_parser.c
```diff
@@ -11235,6 +11295,21 @@ do {										\
 		__FSM_I_MOVE_fixup(Resp_I_EoT, __fsm_n, 0);
 	}
 
+	__FSM_REQUIRE_FIRST_DIGIT(Resp_I_CC_StaleIfErrorVBeg,
+				  Resp_I_CC_StaleIfErrorV);
+
+	__FSM_STATE(Resp_I_CC_StaleIfErrorV) {
+		__fsm_sz = __data_remain(p);
```
EvgeniiMekhanik, line 11302: Should we don like for previos directives?

if (unlikely(resp->cache_ctl.flags & TFW_HTTP_CC_STALE_IF_ERROR))
	__FSM_EXIT(T_DROP);

db/core/htrie.c
```diff
@@ -792,22 +978,158 @@ tdb_htrie_insert(TdbHdr *dbh, unsigned long key, void *data, size_t *len)
...
+
+		if (!bckt->coll_next) {
+			write_unlock_bh(&bckt->lock);
+			write_unlock_bh(&prev->lock);
+			break;
+		}
+
+		b_tmp = bckt;
+		bckt = TDB_HTRIE_BUCKET_NEXT(dbh, bckt);
+		write_lock_bh(&bckt->lock);
```
EvgeniiMekhanik, line 1103: As I understand this function we should lock prev backet and current backet. We shoudl have not more then two locks.
But here we try to catch third lock. And why we don't unlock b_tmp?
As I think we should do
write_unlock_bh(&b_tmp->lock);
write_lock_bh(&bckt->lock);

Please explain for me logic here

fw/cache.c
const-t: You are wrong. You got stale response because you specified `cache_use_stale 200` in the config. It's desired behavior, however, maybe we should deny to use status codes except 4x and 5x.

This feature need in case when you for instance got 500 status code, but you have stale response and you reply with stale resp if it configured.

Error will be if you specify `cache_use_stale 500`, but you received 200 stale response.

db/core/htrie.c
const-t: Need to replace by PAGE_SHIFT.

fw/http.c
```diff
@@ -6782,15 +6875,21 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
...
-		r = tfw_http_req_block(bad_req, 502,
-				       "response blocked: filtered out",
-				       HTTP2_ECODE_PROTO);
-	else
-		r = tfw_http_req_drop(bad_req, 502,
-				      "response dropped: processing error",
-				      HTTP2_ECODE_PROTO);
+
+	if (tfw_http_resp_should_fwd_stale(bad_req, 502)) {
+		__tfw_http_resp_fwd_stale(hmresp);
```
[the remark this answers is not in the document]
const-t: I'm not sure, it's error case I think we need to drop the connection.

fw/vhost.c
```diff
@@ -904,6 +926,84 @@ tfw_mod_hdr_sort(TfwLocation *loc)
...
+
+	if (ce->val_n < 1) {
+		T_ERR_NL("%s too few arguments\n", cs->name);
+		return -EINVAL;
+	}
+
+	cfg = kzalloc(sizeof(TfwCacheUseStale), GFP_KERNEL);
+	if (!cfg)
+		return -ENOMEM;
+	loc->cache_use_stale = cfg;
```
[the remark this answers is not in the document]
const-t: This overriding only for default vhost, I don't see a reason to fix it, because default vhost will be removed soon. Now I just I forbade this directive for default vhost. 

About inheritance: it's has same logic with `req_hdr_set` and we also don't have inheritance for these directives. I think we should fix it together in another task, we have such task for inheritance.

fw/http_parser.c
const-t: Why? We can have repeating headers with different values

fw/vhost.c
```diff
@@ -904,6 +904,84 @@ tfw_mod_hdr_sort(TfwLocation *loc)
...
+
+	for (i = 0; i < ce->val_n; i++) {
+		if (!mask4x && !strcasecmp(ce->vals[i], "4*")) {
+			mask4x = true;
+			__set_bit_range(400, 499, cfg->codes);
+		}
+		else if (!mask5x && !strcasecmp(ce->vals[i], "5*")) {
```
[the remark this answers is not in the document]
const-t: Now it's forbidden to specify in config.

fw/vhost.h
```diff
@@ -120,6 +120,22 @@ enum {
...
+ * @on_error	- Whether or not stale cached response to be used when request
+ * 		  unable to be forwarded to backend.(e.g no alive servers);
+ * @on_timeout	- Whether or not stale cached response to be used when request
+ * 		  is timedout during forwarding to backend;
+ */
+typedef struct {
+	DECLARE_BITMAP(codes, 512);
+	bool		on_error;
+	bool		on_timeout;
+} TfwCacheUseStale;
```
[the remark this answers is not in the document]
const-t: Added, for wiki I have another branch in wiki repo: `kt-use-stale-desc`

fw/cache.c
const-t: Fixed

fw/cache.c
const-t: Fixed

db/core/htrie.c
const-t: We do it in this sequence because we can't free the bucket than now hold the lock and just relies on the lock from previous bucket. However, maybe it's overengineering, for simplicity we can lock/unlock only head bucket, to lock whole list.

fw/vhost.c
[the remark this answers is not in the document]
EvgeniiMekhanik: No error is still present.
vhost xxx {
	proxy_pass default;
	tls_certificate /home/evgeny/workdir/certs/tempesta-tech.test.crt;
	tls_certificate_key /home/evgeny/workdir/certs/tempesta-tech.test.key;
	cache_use_stale 400;
	cache_use_stale 500;
}

And we have a memory leak. Because we don't check dublicate values.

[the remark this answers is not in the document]
EvgeniiMekhanik: For `http_resp_code_block`  I catch duplicate entry: 'http_resp_code_block', for such config. Moreover
```
if (cfg->http_resp_code_block) {
		if (ce->dflt_value)
			return 0;
		kfree(cfg->http_resp_code_block);
		cfg->http_resp_code_block = NULL;
	}
```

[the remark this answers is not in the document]
EvgeniiMekhanik: We need also check how we handle this directive when it is present in vhost and location inside vhost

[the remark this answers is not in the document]
EvgeniiMekhanik: It seems we should set .allow_repeat = false, to this option

fw/vhost.c
```diff
@@ -904,6 +926,84 @@ tfw_mod_hdr_sort(TfwLocation *loc)
...
+		return -EINVAL;
+	}
+
+	cfg = kzalloc(sizeof(TfwCacheUseStale), GFP_KERNEL);
+	if (!cfg)
+		return -ENOMEM;
+	loc->cache_use_stale = cfg;
+
+	for (i = 0; i < ce->val_n; i++) {
+		if (!mask4x && !strcasecmp(ce->vals[i], "4*")) {
```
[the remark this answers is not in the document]
EvgeniiMekhanik: We have TFW_CFG_CHECK_VAL_DUP on master branch, please use it to avoid dublicates

fw/http.c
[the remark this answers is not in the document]
EvgeniiMekhanik: @krizhanovsky pay attention here. As I think if we fail to send stail response we should got to `else` directive and call `tfw_http_req_block` or `tfw_http_req_drop`. Also I am not sure that we can send stale response if we blocked by frang.

fw/cache.c
EvgeniiMekhanik: After discussion we decide that administrator can't set 2** or 3** in cache_use_stale

fw/http_parser.c
EvgeniiMekhanik: Are you sure? I check that without this check we can set cache_control: stale-if-error=20, stale-if-error=25. Is it ok?
For other cache-control directives it is prohibited.

db/core/htrie.c
```diff
@@ -602,6 +718,11 @@ tdb_htrie_create_rec(TdbHdr *dbh, unsigned long off, unsigned long key,
 	if (data)
 		memcpy_fast(ptr, data, len);
 
+	bckt = (TdbBucket *)((unsigned long)r & TDB_HTRIE_DMASK);
+	bckt->flags |= (TDB_HTRIE_COMPLETE_BIT * complete);
```
krizhanovsky, line 722: This is for variable length records only, right? This should be under `if (TDB_HTRIE_VARLENRECS(dbh))`.

Why this is for the whole bucket? If we have fully resolved key, we may have collisions in the bucket, so even for variable length records we can have many `TdbVRec`s in a bucket. I'd expect that the bit should be set for `TdbVRec`.

db/core/htrie.c
```diff
@@ -792,22 +977,158 @@ tdb_htrie_insert(TdbHdr *dbh, unsigned long key, void *data, size_t *len)
...
 	unsigned long o;
+	TdbBucket *b;
 	TdbHtrieNode *root = TDB_HTRIE_ROOT(dbh);
 
 	o = tdb_htrie_descend(dbh, &root, key, &bits);
 	if (!o)
 		return NULL;
+	b = TDB_PTR(dbh, o);
+	read_lock_bh(&(b)->lock);
+	if (!tdb_bucket_is_complete(b)) {
```
krizhanovsky, line 1123: I don't think we need to implement #500 right now, but there is a comment https://github.com/tempesta-tech/tempesta/issues/500#issuecomment-377051964 , which is aligned with `incomplete` buckets (why not records?), so it'd be good to write a good `TODO #500` comment and align the current implementation with the upcoming task.

db/core/main.c
```diff
@@ -288,7 +334,12 @@ tdb_rec_get_alloc(TDB *db, unsigned long key, TdbGetAllocCtx *ctx)
 	}
 	ctx->is_new = true;
 	r = tdb_entry_alloc(db, key, &ctx->len);
+	if (!r) {
+		spin_unlock(&db->ga_lock);
+		return r;
```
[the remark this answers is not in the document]
const-t: @krizhanovsky Why do we need? At this place we can have 2 possible situations:
First:

```
      while (!TDB_ITER_BAD(iter)) {
		if (ctx->eq_rec(iter.rec,  ctx->ctx)) {
			spin_unlock(&db->ga_lock);
			return iter.rec;
		}
		tdb_rec_next(db, &iter);
	}
```
scanned whole bucket and didn't find any suitable record, on the last iteration bucket was put by `tdb_rec_next()`. However, if record found the caller of `tdb_rec_get_alloc()` puts the bucket. 

Second: We don't have the bucket with such key, just try to allocate new, but on fail we just return with spin_unlock. At this moment we don't have the record so there is not to put.

fw/cache.c
```diff
@@ -2986,14 +3121,32 @@ cache_req_process_node(TfwHttpReq *req, tfw_http_cache_cb_t action)
 		}
 	}
 
-	resp = tfw_cache_build_resp(req, ce, lifetime);
+	resp = tfw_cache_build_resp(req, ce, age);
```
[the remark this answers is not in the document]
const-t: From my point of view, in this function we even can remove stale entries from the cache. However it will require architectural changes of htrie. `cache_req_process_node` knows whether entry alive or not by calling `tfw_cache_entry_is_live`, if entry is not alive we can remove it from the cache here in case when ability to response with stale responses is disabled.  At this moment we have two problems to solve:
1. We can't remove arbitrary element from the collision chain, because we need to lock previous element or whole chain, we can't lock previous element because our chain is single linked list. When we will turn it into double we need to verify all remove/insert logic, that we don't have deadlocks.
2. For stale responses we need refcounter or something else that give an ability to hold the record without locking and then remove this record if it was **successfully** revalidated.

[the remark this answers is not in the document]
krizhanovsky: >  in this function we even can remove stale entries from the cache.

Yeah, but I'd move the eviction/GC to TDB: TDB is small while cache.c is already pretty big, so it'd be good to reduce code complexity in cache. Also client.c needs eviction as well.

Meantime, *revalidation* is a tricky thing - TDB can't just remove old entries, it must have some understanding on what's fine to evict and what's not. I think we can pass a callback to TDB, which will implement an eviction policy.

> We can't remove arbitrary element from the collision chain, because we need to lock previous element or whole chain, we can't lock previous element because our chain is single linked list. 

Can't we use some lock free approach and do additional spiining/checking on the list traversal?

db/core/htrie.c
```diff
@@ -792,22 +977,158 @@ tdb_htrie_insert(TdbHdr *dbh, unsigned long key, void *data, size_t *len)
...
+static inline void
+tdb_eval_bf_remove_cb(tdb_before_remove_cb_t *bf_remove_cb, TdbRec *rec)
+{
+	if (bf_remove_cb)
+		bf_remove_cb(rec);
+}
+
+#define TDB_REMOVE_FOREACH_REC(body)					\
+	do {								\
+		size_t rlen = sizeof(*rec) +				\
```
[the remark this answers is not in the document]
krizhanovsky: In my understanding `TDB_HTRIE_RBODYLEN` is about the record _body_ length, not about the difference between `sizeof(TdbVRec)` and `sizeof(TdbFRec)`.

I think we have the bug. See that `tdb_htrie_create_rec()` assigns the data length to `rec->len`, so `TDB_HTRIE_RBODYLEN` for `TdbVRec` is the data length, carried by the record. Maybe the bug doesn't appear (frequently) since it affects only chained records, there is only 8 byte difference (which can be even smaller due to `TDB_HTRIE_RALIGN` - we simply may observe incorrect record instead of accessing invalid memory location). @EvgeniiMekhanik please also have a look - this bug may appear on the current pilot!

db/core/main.c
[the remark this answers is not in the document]
krizhanovsky: Yeah, you're right.

fw/cache.c
```diff
@@ -1236,14 +1252,11 @@ tfw_cache_dbce_get(TDB *db, TdbIter *iter, TfwHttpReq *req)
...
-		if (tfw_cache_entry_key_eq(db, req, ce, false)) {
-			if (tfw_cache_entry_is_live(req, ce)
-			    && (!ce_best
-				|| tfw_cache_entry_age(ce)
-				   < tfw_cache_entry_age(ce_best)))
+		if (tfw_cache_should_satisfy(req, ce) &&
+		    tfw_cache_entry_key_eq(db, req, ce))
+		{
+			if (!ce_best || tfw_cache_entry_age(ce)
+				      < tfw_cache_entry_age(ce_best))
```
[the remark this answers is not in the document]
EvgeniiMekhanik: The key is URI, so for one uri we can have HEAD and GET records which are both valid

fw/vhost.c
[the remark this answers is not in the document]
const-t: Fixed

db/core/htrie.c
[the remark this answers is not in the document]
const-t: I see, you are right we take the size of `TdbRec` even for `TdbVRec`. I fixed it throughout htrie code.

etc/tempesta_fw.conf
```diff
@@ -637,6 +637,18 @@
 #   None.
 #
 
+# TAG: cache_use_stale
+#
+# Specifies the conditions under which a stale cached response is allowed
+# for use.
+#
+# Syntax:
+#   cache_use_stale <code|4*|5*>;
```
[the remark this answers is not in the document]
const-t: Yes, It's handled by 504 code. There is test for this case `cache.test_h2_cache_control.TestH2CacheUseStaleTimeout` or its http version.

db/core/htrie.c
```diff
@@ -86,6 +86,71 @@ tdb_set_bit(unsigned long *bmp, unsigned int nr)
...
+		*page_addr = this_cpu_ptr(dbh->pcpu)->freelist;
+		this_cpu_ptr(dbh->pcpu)->freelist = ptr;
+	}
+}
+
+static void
+tdb_get_blk(TdbHdr *dbh, unsigned long ptr)
+{
+	TdbExt *e = tdb_ext(dbh, TDB_PTR(dbh, ptr));
+	atomic_t *reftbl = tdb_get_reftbl(dbh, e);
```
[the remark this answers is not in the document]
const-t: There is two reasons:
1. Because we want to store reference counter into tdb, that gives an ability to save/restore tdb on/from disk.
2. During system boot, Tempesta reserves memory for TDB. If default allocation failed we rollback to vmalloc. Using refcounter with such pages requires additional logic that we don't have now.

fw/cache.c
const-t: Done

fw/cache.c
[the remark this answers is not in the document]
krizhanovsky: This comment is still not addressed

> Also I think we need to at least document the current design or maybe fix it. We construct a response from a stale cache entry and most likely the response will be just freed once we get a fresh response from the upstream, i.e. we do the construction with no reason

we need at least a new issue for this

fw/client.c
```diff
@@ -324,6 +408,15 @@ static TfwCfgSpec tfw_client_specs[] = {
...
 		}
 	},
+	{
+		.name = "client_lru_size",
+		.deflt = "8000",
+		.handler = tfw_cfg_set_int,
+		.dest = &client_cfg.lru_size,
+		.spec_ext = &(TfwCfgSpecInt) {
+			.range = { 1, UINT_MAX },
+		},
```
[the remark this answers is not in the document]
const-t: I also forget to add this directive to config, I'll do it.

db/core/htrie.c
```diff
@@ -86,6 +86,71 @@ tdb_set_bit(unsigned long *bmp, unsigned int nr)
...
+	return TDB_PTR(dbh, (TDB_EXT_BASE(dbh, e) + TDB_EXT_SZ) - TDB_BLK_SZ);
+}
+
+/**
+ * Add block to per-CPU freelist.
+ *
+ * NOTE Due to per-CPU nature of freelist, block can be allocated only on CPU
+ * where it was freed. It can cause allocation failures in case when one
+ * CPU frees more blocks then others and free blocks available only in per-CPU
+ * freelists because main memory pool is exhausted.
```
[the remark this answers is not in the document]
krizhanovsky: This comment isn't addressed: is it long to introduce a global spinlock-protected free list and limit per-cpu freelist with some number, like 100 blocks?

fw/cache.c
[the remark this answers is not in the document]
const-t: Reworked to store only one version per key.

fw/cache.c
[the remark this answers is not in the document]
const-t: Implemented in new commit [Enhancement for cache_use_stale](https://github.com/tempesta-tech/tempesta/pull/2074/commits/85fa6c67c3089b967d41f698413d1ad1b5408af7)

krizhanovsky (approved):
There are several notes, but LGTM in general

ktest/linux/slab.h
```diff
@@ -35,6 +35,7 @@
 /* asm/page.h */
 #define PAGE_SIZE	4096UL
 #define PAGE_MASK	(~(PAGE_SIZE-1))
+#define PAGE_SHIFT	12
```
EvgeniiMekhanik, line 38: Why you don't use all this values from linux kenrel header?

fw/client.c
```diff
@@ -292,7 +374,7 @@ tfw_client_start(void)
 	if (!client_db)
 		return -EINVAL;
 
-	return 0;
+	return tfw_client_init_lru();
```
EvgeniiMekhanik, line 377: There is a leak here! If `tfw_client_init_lru` fails `tfw_client_stop` will never called (just try to return -ENOMEM in `tfw_client_init_lru` instead of memory allocation). And tdb_close(client_db); will never called

fw/client.c
```diff
@@ -66,22 +67,87 @@ typedef struct {
...
+	TdbFRec *rec = ((TdbFRec *)cli) - 1;
 
 	T_DBG2("put client %p, users=%d\n",
 	       cli, ent->users);
 
 	spin_lock(&ent->lock);
 
 	--ent->users;
+	if (ent->users > 0)
+		tdb_rec_put(client_db, rec);
```
EvgeniiMekhanik, line 150: May be move it just after `if (likely(ent->users)) { ` on string 153. Why we check  `if (ent->users > 0)` two times?

fw/http_parser.c
```diff
@@ -7249,6 +7271,7 @@ do {									\
 	SET_HDR_CACHE_CONTROL_FIELD(max_age, TFW_HTTP_CC_MAX_AGE);
```
EvgeniiMekhanik, line 7271: We should add (req->cache_ctl.flags & TFW_HTTP_CC_STALE_IF_ERROR) for consistency

fw/cache.c
```diff
@@ -3018,6 +3018,38 @@ tfw_cache_build_resp(TfwHttpReq *req, TfwCacheEntry *ce, long age)
...
 }
 
+TfwHttpResp *
+tfw_cache_build_resp_stale(TfwHttpReq *req)
+{
+	TDB *db = node_db();
+	TfwCacheEntry *ce = req->stale_ce;
+	TfwHttpResp *resp = tfw_cache_build_resp(req, ce, req->stale_ce_age);
+
+#if defined(DUBEG)
```
EvgeniiMekhanik, line 3033: DUBEG? May be DEBUG?

EvgeniiMekhanik: And It seems we don't need this ifdef because we already use T_DBG

fw/cache.c
```diff
@@ -3093,10 +3125,12 @@ cache_req_process_node(TfwHttpReq *req, tfw_http_cache_cb_t action)
...
+		req->stale_ce = ce;
+		req->stale_ce_age = age;
 	}
 
-	T_DBG("Cache: service request [%p] w/ key=%lx, ce=%p",
-	      req, ce->trec.key, ce);
+	T_DBG("Cache: service request [%p] w/ key=%lx, ce=%p", req,
+	      ce->trec.key, ce);
 
 	TFW_INC_STAT_BH(cache.hits);
```
EvgeniiMekhanik, line 3148: I seems that there us an extra record put in case when `tfw_handle_validation_req` returns false and `stale` is true. We put record during sending 304 response (req->stale_ce is not NULL) and after put.

EvgeniiMekhanik: I add test to your branch

fw/http.c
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes `__tfw_http_resp_fwd_stale` can returns false and we should handle this case. For example here if `__tfw_http_resp_fwd_stale` returns false we should call `tfw_http_req_drop`

fw/http.c
```diff
@@ -6543,6 +6707,46 @@ tfw_http_resp_terminate(TfwHttpMsg *hm)
...
+	data.req = (TfwMsg *)req;
+	data.resp = (TfwMsg *)hmresp;
+	if (tfw_gfsm_move(&hmresp->conn->state, TFW_HTTP_FSM_RESP_MSG_FWD,
+			  &data))
+	{
+		/* The response is freed by tfw_http_req_block(). */
+		tfw_http_req_block(req, 403, "response blocked: filtered out",
+				   HTTP2_ECODE_PROTO);
+		TFW_INC_STAT_BH(serv.msgs_filtout);
+		return T_BLOCK;
```
EvgeniiMekhanik, line 6741: return `tfw_http_req_block` we should not return T_BLOCK to lower layer

fw/http.c
```diff
@@ -1483,13 +1483,140 @@ tfw_http_nip_req_resched_err(TfwSrvConn *srv_conn, TfwHttpReq *req,
...
+}
+
+/**
+ * The same as @__tfw_http_resp_fwd_stale(), but used in case when we don't
+ * have response from upstream.
+ */
+static bool
+__tfw_http_resp_fwd_stale_noresp(TfwHttpReq *req)
+{
+	if (!tfw_cache_build_resp_stale(req))
```
EvgeniiMekhanik, line 1551: If it fails we don't send anything to client. For example when we can't send error responses to client we call `tfw_http_resp_build_error`. We need error injection tests to check Tempesta FW behavior if `tfw_cache_build_resp_stale` and etc. functions fails

fw/http.c
```diff
@@ -1483,13 +1483,140 @@ tfw_http_nip_req_resched_err(TfwSrvConn *srv_conn, TfwHttpReq *req,
...
+	return req->cache_ctl.flags & TFW_HTTP_CC_STALE_IF_ERROR &&
+		tfw_http_use_stale_if_error(status);
+}
+
+static inline void
+tfw_http_send_err_resp_nolog(TfwHttpReq *req, int status)
+{
+	/* Response must be freed before calling tfw_http_send_err_resp_nolog(). */
+	if (tfw_http_resp_should_fwd_stale(req, status)) {
+		__tfw_http_resp_fwd_stale_noresp(req);
```
EvgeniiMekhanik, line 1613: It seems should be like
```
if (!tfw_http_resp_should_fwd_stale(req, status) || !__tfw_http_resp_fwd_stale_noresp(req)) {
		if (TFW_MSG_H2(req))
			tfw_h2_send_err_resp(req, status, false);
		else
			tfw_h1_send_err_resp(req, status);
}
```

fw/http.c
```diff
@@ -6543,6 +6707,46 @@ tfw_http_resp_terminate(TfwHttpMsg *hm)
...
+			  &data))
+	{
+		/* The response is freed by tfw_http_req_block(). */
+		tfw_http_req_block(req, 403, "response blocked: filtered out",
+				   HTTP2_ECODE_PROTO);
+		TFW_INC_STAT_BH(serv.msgs_filtout);
+		return T_BLOCK;
+	}
+
+	if (!__tfw_http_resp_fwd_stale(hmresp))
```
EvgeniiMekhanik, line 6744: It seems should be
```
if (!__tfw_http_resp_fwd_stale(hmresp)) {
 return tfw_http_req_drop(...)
}
```

db/core/htrie.c
```diff
@@ -365,8 +438,8 @@ tdb_alloc_blk(TdbHdr *dbh)
 	if (unlikely(TDB_HTRIE_OFF(dbh, e) == dbh->dbsz)) {
 		/* We do this set because we skip the last page in extent. */
 		atomic64_set(&dbh->nwb, dbh->dbsz);
-		TDB_ERR("out of free space\n");
-		return 0;
+		dbh->oom = true;
```
EvgeniiMekhanik, line 441: It seems that we never set dbh->oom to false.  So ig it was set one time we change our allocation strategy for all time. Is it ok?

fw/vhost.c
```diff
@@ -3494,6 +3648,14 @@ static TfwCfgSpec tfw_vhost_internal_specs[] = {
 		.allow_repeat = true,
 		.allow_reconfig = true,
 	},
+	{
+		.name = "cache_use_stale",
+		.deflt = NULL,
+		.handler = tfw_cfgop_in_cache_use_stale,
+		.allow_none = true,
+		.allow_repeat = true,
```
EvgeniiMekhanik, line 3656: allow_repeat = true? Is it ok to set?
```
cache_use_stale 4*;
cache_use_stale 500;
```

fw/vhost.c
```diff
@@ -3336,6 +3482,14 @@ static TfwCfgSpec tfw_vhost_location_specs[] = {
 		.allow_repeat = false,
 		.allow_reconfig = true,
 	},
+	{
+		.name = "cache_use_stale",
+		.deflt = NULL,
+		.handler = tfw_cfgop_loc_cache_use_stale,
+		.allow_none = true,
+		.allow_repeat = true,
```
EvgeniiMekhanik, line 3490: allow_repeat = true?

fw/vhost.c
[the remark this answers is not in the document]
EvgeniiMekhanik: if allow_repeat is true there is a memory leak here.

fw/http.c
```diff
@@ -4002,8 +4038,7 @@ tfw_h2_adjust_req(TfwHttpReq *req)
 		return r;
 
 	/* First line. */
-	__h2_msg_hdr_val(&ht->tbl[TFW_HTTP_HDR_H2_METHOD], &meth);
-	r = tfw_msg_write(&it, &meth);
+	r = __h2_write_method(req, &it);
```
EvgeniiMekhanik, line 4090: It is not a problem of this patch, but it seems that it is better to rewrite all this code to
```
if (unlikely((r = ...))
    return r;
```

fw/http2.c
```diff
@@ -484,14 +484,26 @@ tfw_h2_stream_xmit_prepare_resp(TfwStream *stream)
...
-	 * on HEAD request.
-	 */
-	if (test_bit(TFW_HTTP_B_CHUNKED, resp->flags)
-	    && !test_bit(TFW_HTTP_B_VOID_BODY, resp->flags))
-		r = tfw_http_msg_cutoff_body_chunks(resp);
+
+	if (test_bit(TFW_HTTP_B_REQ_HEAD_TO_GET, resp->req->flags)
+	    && !TFW_STR_EMPTY(&resp->body)) {
+		/* Send only headers for HEAD method. */
+		r = ss_skb_list_chop_head_tail(&resp->msg.skb_head, 0,
```
EvgeniiMekhanik, line 491: It seems that we delete body in `tfw_http_adjust_resp` isn't it?

fw/http2.c
```diff
@@ -484,14 +484,26 @@ tfw_h2_stream_xmit_prepare_resp(TfwStream *stream)
...
+					       tfw_str_total_len(&resp->body)
+					       + resp->trailers_len);
+		if (unlikely(r))
+			goto finish;
+		resp->msg.len = 0;
```
EvgeniiMekhanik, line 496: Can you explain please why we need this?

ai-tmpst (approved):
Found 3 minor issues, otherwise LGTM.

fw/http.c
const-t: Added to https://github.com/tempesta-tech/tempesta/issues/1411

ktest/linux/slab.h
const-t: It's library to override all these values from linux kernel and use it in userspace.

fw/http2.c
const-t: `tfw_http_adjust_resp()` deletes body only for http, here we do it http2

fw/http2.c
const-t: Typo, fixed

fw/http.c
```diff
@@ -1540,6 +1540,7 @@ __tfw_http_resp_fwd_stale(TfwHttpMsg *hmresp)
 
 	return sent;
 }
+ALLOW_ERROR_INJECTION(__tfw_http_resp_fwd_stale, NULL);
```
EvgeniiMekhanik, line 1543: I am not sure that it is ok to return NULL for error injection if this function really returns bool value

fw/http.c
```diff
@@ -1558,6 +1559,7 @@ __tfw_http_resp_fwd_stale_noresp(TfwHttpReq *req)
 
 	return true;
 }
+ALLOW_ERROR_INJECTION(__tfw_http_resp_fwd_stale_noresp, NULL);
```
EvgeniiMekhanik, line 1562: I am not sure that it is ok to return NULL for error injection if this function really returns bool value, may be it is better to make it return int value and return -ENOMEM or any other error on fail if we want to use error injection tests.

fw/cache.c
```diff
@@ -3108,99 +3106,163 @@ tfw_cache_can_use_stale(TfwHttpReq *req, TfwCacheEntry *ce, long age)
...
+	age = tfw_cache_entry_age(ce);
+
+	if (!(lifetime = tfw_cache_entry_is_live(req, ce, age))) {
+		bool success;
+
+		T_DBG3("%s: db=[%p] req=[%p] ce=[%p] CE is not alive\n",
+		       __func__, db, req, ce);
+
+		if ((success = tfw_cache_can_use_stale(req, ce, age)))
+			cache_do_service_request_stale(req, action, ce, age);
```
EvgeniiMekhanik, line 3230: Who drop `ce` after `cache_do_service_request_stale`?

const-t: On next line there is a comment. It will be freed during stale forwarding or in req destructor.

fw/cache.c
```diff
@@ -3108,99 +3106,163 @@ tfw_cache_can_use_stale(TfwHttpReq *req, TfwCacheEntry *ce, long age)
...
-			goto put;
+			return;
 		}
 	}
 
-	if (!stale)
-		resp = tfw_cache_build_resp(req, ce, age);
+	resp = tfw_cache_build_resp(req, ce, age);
+	if (unlikely(!resp)) {
+		tfw_http_conn_msg_free((TfwHttpMsg *)req);
```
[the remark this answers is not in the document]
const-t: Fixed

PR #2076, opened 2024-03-11
Fix socket cpu migration

### LINKED ISSUE #2075: Socket cpu migration
Currently there are a lot of cases when we can observe socket CPU migration. This situation can leads to two problems: performance degradation upon capture socket lock and response reordering (which leads to violation of the HTTP1 RFC).

EvgeniiMekhanik:
Socket cpu migration can lead to two problems:
performance degradation and response reordering,
which leads to broken HTTP1.
Previously we use RSS and RPS to prevent it, but
there were several problems in our scripts:
- we exclude loopback interfaces from setup, because
  we don't take into account response reordering
  problem.
- we don't take into account that some interfaces
  have some suffix lile @if14, and we should remove
  it from device name in our scripts.
- we don't try to setup combined RSS queues, only
  RX queues, but there are a lot of cases when network
  interface has only combined queues.
- we don't take into account overflow when we calculate
  1 << x, when x is greater or equal then 64.
- When we setup RPS we use cpu_mask, which is calculated
  as $(perl -le 'printf("%x", (1 << '$CPUS_N') - 1)').
  But in this case we use all cpus, not only one.
- we don't setup RPS for network interface if, RSS setup
  fails.
This patch fix all this problems.

Closes #2075

### DIFF
--- a/scripts/tempesta.sh
+++ b/scripts/tempesta.sh
@@ -3,7 +3,7 @@
 # Tempesta FW service script.
 #
 # Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
-# Copyright (C) 2015-2022 Tempesta Technologies, Inc.
+# Copyright (C) 2015-2024 Tempesta Technologies, Inc.
 #
 # This program is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


EvgeniiMekhanik:
When we use all existing CPUs, when we set RPS it is ok. CPU will be changed only one time and only for loopback, becase for loopback we set skb_hash in tcp_make_synack and this skb will return to us! It is not a problem, because later CPU will be calulated correctly in netif_rx_skb_internal for all other skbs.

scripts/tfw_lib.sh
```diff
@@ -92,19 +91,83 @@ irqbalance_ban_irqs()
 	systemctl restart irqbalance.service >/dev/null
 }
 
+# This function prepares cpu mask for RSS and RPS.
+# It takes into account that we can't calculate
+# value, which is greater when (1 << 63) and cam't
```
const-t, line 96: `cam't` typo

const-t (commented):
`functionfor` typo

EvgeniiMekhanik:
Do not review the second commit. I move it in separate PR.

const-t (approved):
LGTM, please don't forget to remove second commit.

scripts/tfw_lib.sh
```diff
@@ -92,19 +91,83 @@ irqbalance_ban_irqs()
...
+	val=0
+	cond=0
+
+	echo `make_cpu_mask $delta $mask32 $val $cond`
+}
+
+make_cpu_rps_mask()
+{
+	delta=$1
+	# enable all CPUs, which is less the (1 << delta) - 1
```
krizhanovsky, line 149:
```suggestion
	# enable all CPUs, which is less than (1 << delta) - 1
```

krizhanovsky (approved):
LGTM

scripts/tfw_lib.sh
EvgeniiMekhanik: Fixed

scripts/tempesta.sh
```diff
@@ -46,12 +46,12 @@ tls_mod=tempesta_tls
 tdb_mod=tempesta_db
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Fixed

Issue #2033, opened 2024-01-05
Get rid of the `possible`/`online` CPUs mess


krizhanovsky:
# Motivation

The problem is addressed in https://github.com/tempesta-tech/tempesta/pull/2023#discussion_r1434538816 : at the moment our code uses `for_each_possible_cpu()` in several places and even `NR_CPUS`, but it was never tested against (and even designed!) hot-plug or even switch on/off (e.g. with `sysctl`) CPUs. Some other places use `for_each_online_cpus()`. We have no idea whether Tempesta works with dynamically switched on/off or added/removed CPUs.

So in some cases we allocate extra memory for CPUs, which are never used. E.g. imagine how bad it is on basic Ubuntu 22:
```
$ grep 'CONFIG_NR_CPUS=\|CONFIG_HOTPLUG_CPU=' /boot/config-5.15.0-91-generic 
CONFIG_NR_CPUS=8192
CONFIG_HOTPLUG_CPU=y
```

# Scope

Need to move to `for_each_online_cpu()` and remove `NR_CPUS` from `fw/cache.c`.

# Testing

There is no need any specific testing, our regular CI is enough

# Documentation

Probably it makes sense to update https://tempesta-tech.com/knowledge-base/Requirements/ if switching off `CONFIG_HOTPLUG_CPU` could provide any performance improvements due to simplified macros around the kernel.

krizhanovsky:
Done with #2077 and https://github.com/tempesta-tech/linux-5.10.35-tfw/pull/22

PR #2077, opened 2024-03-12
En/fix 2033

enuribekov-tempesta:
According to issue 2033
1. `for_each_possible_cpu()` changed to `each_online_cpu()`
2. Static array with size `NR_CPUS` changed to dynamic one with size `num_online_cpu()`
3. Updated kernel patch with the same changes.

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -341,6 +341,9 @@ tfw_release_node_cpus(void)
 {
 	int node;
 
+	if(!c_nodes)
+		return;
+
 	for(node = 0; node < nr_online_nodes; node++) {
 		if(c_nodes[node].cpu)
 			kfree(c_nodes[node].cpu);
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -339,14 +339,11 @@ tfw_cache_key_node(unsigned long key)
 static void
 tfw_release_node_cpus(void)
 {
-	int i;
+	int node;
 
-	for (i = 0; i < MAX_NUMNODES; i++) {
-		if(!c_nodes[i].cpu) {
-			T_WARN("%s c_node deallocation interrupted on node %i\n", __func__, i);
-			break;
-		}
-		kfree(c_nodes[i].cpu);
+	for_each_online_node(node) {
+		if(c_nodes[node].nr_cpus > 0)
+			kfree(c_nodes[node].cpu);
 	}
 }
 
@@ -354,26 +351,29 @@ tfw_release_node_cpus(void)
  * Just choose any CPU for each node to use queue_work_on() for
  * nodes scheduling. Reserve 0th CPU for other tasks.
  */
-static void
+static int
 tfw_init_node_cpus(void)
 {
-	int i, nr_cpus, cpu, node;
+	int nr_cpus, cpu, node;
 
 	nr_cpus = num_online_cpus();
 
-	for (i = 0; i < MAX_NUMNODES; i++) {
-		c_nodes[i].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL);
-		if(!c_nodes[i].cpu) {
-			T_ERR("%s Failed to allocate c_nodes[%i] cpu\n", __func__, i);
-			tfw_release_node_cpus();
-			return;
-		}
-	}
-
 	for_each_online_cpu(cpu) {
+
+
 		node = cpu_to_node(cpu);
+
+		pr_err("node.nr_cpus %u",c_nodes[node].nr_cpus);
+
+		c_nodes[node].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL);
+		if(!c_nodes[node].cpu) {
+			T_ERR("%s Failed to allocate c_nodes[%i] cpu\n", __func__, node);
+			return -ENOMEM;
+		}
+
 		c_nodes[node].cpu[c_nodes[node].nr_cpus++] = cpu;
 	}
+	return 0;
 }
 
 
@@ -3215,7 +3215,8 @@ tfw_cache_start(void)
 		goto close_db;
 	}
 #endif
-	tfw_init_node_cpus();
+	if(tfw_init_node_cpus() == -ENOMEM) 
+		goto close_db;
 
 	TFW_WQ_CHECKSZ(TfwCWork);
 	for_each_online_cpu(i) {
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -234,7 +234,7 @@ enum {
 };
 
 typedef struct {
-	int* cpu;
+	int 		*cpu;
 	atomic_t	cpu_idx;
 	unsigned int	nr_cpus;
 	TDB		*db;
@@ -341,10 +341,9 @@ tfw_release_node_cpus(void)
 {
 	int i;
 
-	for (i = 0; i < MAX_NUMNODES; i++)
-	{
+	for (i = 0; i < MAX_NUMNODES; i++) {
 		if(!c_nodes[i].cpu) {
-			T_ERR("%s c_node deallocation interrupted on node %i\n", __func__, i);
+			T_WARN("%s c_node deallocation interrupted on node %i\n", __func__, i);
 			break;
 		}
 		kfree(c_nodes[i].cpu);
@@ -363,7 +362,7 @@ tfw_init_node_cpus(void)
 	nr_cpus = num_online_cpus();
 
 	for (i = 0; i < MAX_NUMNODES; i++) {
-		c_nodes[i].cpu = kmalloc(nr_cpus*sizeof(int), GFP_KERNEL);
+		c_nodes[i].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL);
 		if(!c_nodes[i].cpu) {
 			T_ERR("%s Failed to allocate c_nodes[%i] cpu\n", __func__, i);
 			tfw_release_node_cpus();
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -358,20 +358,16 @@ tfw_init_node_cpus(void)
 {
 	int nr_cpus, cpu, node;
 
-	T_ERR("nodes: %i", nr_online_nodes);
-
 	c_nodes = kzalloc(nr_online_nodes * sizeof(CaNode), GFP_KERNEL);
 		if (!c_nodes) {
 			T_ERR( "Failed to allocate nodes map for cache work scheduler");
 			return -ENOMEM;
 	}
 
-	return 0;
-
 	for_each_online_cpu(cpu) {
 		node = cpu_to_node(cpu);
 		nr_cpus = nr_cpus_node(node);
-		T_ERR("node - nr_cpus %i - %i",node, nr_cpus);
+		T_DBG("node - nr_cpus %i - %i",node, nr_cpus);
 		c_nodes[node].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL);
 		if (!c_nodes[node].cpu) {
 			T_ERR( "Failed to allocate a CPU %i for cache work scheduler", cpu);
@@ -3209,7 +3205,10 @@ tfw_cache_start(void)
 	if (!(cache_cfg.cache || g_vhost->cache_purge))
 		return 0;
 
-	for_each_node_with_cpus(i) {
+	if (!tfw_init_node_cpus())
+		goto close_db;
+
+	for(i = 0; i < nr_online_nodes; i++) {
 		c_nodes[i].db = tdb_open(cache_cfg.db_path,
 					 cache_cfg.db_size, 0, i);
 		if (!c_nodes[i].db)
@@ -3223,8 +3222,6 @@ tfw_cache_start(void)
 		goto close_db;
 	}
 #endif
-	if (!tfw_init_node_cpus())
-		goto close_db;
 
 	TFW_WQ_CHECKSZ(TfwCWork);
 	for_each_online_cpu(i) {
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -333,8 +333,7 @@ tfw_cache_key_node(unsigned long key)
 }
 
 /**
- * Release memory dynamically allocated
- * for every node cpus
+ * Release memory dynamically allocated for every node cpus
  */
 static void
 tfw_release_node_cpus(void)
@@ -359,18 +358,14 @@ tfw_init_node_cpus(void)
 	nr_cpus = num_online_cpus();
 
 	for_each_online_cpu(cpu) {
-
-
 		node = cpu_to_node(cpu);
-
-		pr_err("node.nr_cpus %u",c_nodes[node].nr_cpus);
-
+		T_DBG("node.nr_cpus %u",c_nodes[node].nr_cpus);
 		c_nodes[node].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL);
-		if(!c_nodes[node].cpu) {
-			T_ERR("%s Failed to allocate c_nodes[%i] cpu\n", __func__, node);
+		if (!c_nodes[node].cpu) {
+			//T_ERR("Failed to allocate c_nodes[%i] cpu\n", __func__, node);
+			T_ERR( "Failed to allocate a CPU %i for cache work scheduler", cpu);
 			return -ENOMEM;
 		}
-
 		c_nodes[node].cpu[c_nodes[node].nr_cpus++] = cpu;
 	}
 	return 0;
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -363,8 +363,7 @@ tfw_init_node_cpus(void)
 			T_ERR( "Failed to allocate nodes map for cache work scheduler");
 			return -ENOMEM;
 	}
-
-	for_each_online_node(node) {
+	for_each_node_with_cpus(node) {
 		nr_cpus = nr_cpus_node(node);
 		c_nodes[node].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL);
 		if (!c_nodes[node].cpu) {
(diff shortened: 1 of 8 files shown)


fw/cache.c
```diff
@@ -328,21 +328,52 @@ tfw_cache_key_node(unsigned long key)
...
+ * Release memory dynamically allocated
+ * for every node cpus
+ */
+static void
+tfw_release_node_cpus(void)
+{
+	int i;
+
+	for (i = 0; i < MAX_NUMNODES; i++)
+	{
```
krizhanovsky, line 341:
```suggestion
	for (i = 0; i < MAX_NUMNODES; i++) {
```

fw/cache.c
```diff
@@ -328,21 +328,52 @@ tfw_cache_key_node(unsigned long key)
...
 static void
 tfw_init_node_cpus(void)
 {
-	int cpu, node;
+	int i, nr_cpus, cpu, node;
+
+	nr_cpus = num_online_cpus();
+
+	for (i = 0; i < MAX_NUMNODES; i++) {
+		c_nodes[i].cpu = kmalloc(nr_cpus*sizeof(int), GFP_KERNEL);
```
krizhanovsky, line 362:
```suggestion
		c_nodes[i].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL);
```

fw/cache.c
```diff
@@ -328,21 +328,52 @@ tfw_cache_key_node(unsigned long key)
...
  */
 static void
 tfw_init_node_cpus(void)
 {
-	int cpu, node;
+	int i, nr_cpus, cpu, node;
+
+	nr_cpus = num_online_cpus();
+
+	for (i = 0; i < MAX_NUMNODES; i++) {
```
krizhanovsky, line 361: Why do we need the loops over all NUMA nodes here and in `tfw_release_node_cpus()`? Each CPU belongs to only one node, so with the nested loops we multiply CPUs by `MAX_NUMNODES`, which also comes from the Linux config and can be quite large, so we back to the same problem as in the original #2033 task

tls/mpool.c
```diff
@@ -397,7 +397,7 @@ ttls_mpool_exit(void)
 	int i;
```
krizhanovsky, line 397: Please update the copyright year in this and all the modified files or enable a github hook, which does it automatically (it should be described on the wiki)

krizhanovsky (changes requested):
It seems the NUMA nodes processing in cache.c can be done in more optimal way and there are a couple of minor comments to cleanup the code

fw/cache.c
```diff
@@ -234,7 +234,7 @@ enum {
 };
 
 typedef struct {
-	int		cpu[NR_CPUS];
+	int* cpu;
```
EvgeniiMekhanik, line 237: int *cpu

fw/cache.c
```diff
@@ -328,21 +328,52 @@ tfw_cache_key_node(unsigned long key)
...
+		kfree(c_nodes[i].cpu);
+	}
+}
+
 /**
  * Just choose any CPU for each node to use queue_work_on() for
  * nodes scheduling. Reserve 0th CPU for other tasks.
  */
 static void
 tfw_init_node_cpus(void)
```
EvgeniiMekhanik, line 355: This function should return int. And in case if allocation failed you should return -ENOMEM; I catch kernel panic when c_nodes[i].cpu = kmalloc return NULL, because  `tfw_release_node_cpus` was called twice.

EvgeniiMekhanik (changes requested):
Need fixes

enuribekov-tempesta:
Added update, not  yet tested properly, just as proposal.

And by the way. 
At the moment we allocate `c_nodes` as a static array with a size of `MAX_NUMNODES`.
What about dynamical allocating of this array with a size of `num_online_nodes` ?

fw/cache.c
```diff
@@ -3159,7 +3189,8 @@ tfw_cache_start(void)
 		goto close_db;
 	}
 #endif
-	tfw_init_node_cpus();
+	if(tfw_init_node_cpus() == -ENOMEM)
```
krizhanovsky, line 3192:
```suggestion
	if (tfw_init_node_cpus() == -ENOMEM) 
```

krizhanovsky: Why only on `-ENOMEM`? If we adjust `tfw_init_node_cpus()`, then we'll get a bug here

fw/cache.c
```diff
@@ -328,21 +328,51 @@ tfw_cache_key_node(unsigned long key)
...
+	nr_cpus = num_online_cpus();
 
 	for_each_online_cpu(cpu) {
+
+
 		node = cpu_to_node(cpu);
+
+		pr_err("node.nr_cpus %u",c_nodes[node].nr_cpus);
+
+		c_nodes[node].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL);
```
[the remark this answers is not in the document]
krizhanovsky: This is very wrong! Say you have 64 CPUs in 2 nodes, each node (CPU processor package) has 32 CPU cores (maybe with hyperthreads). In this loop you make 64 allocations and constantly rewrite one of the two `cpu` pointers, leaking the memory.

The initial loop is essentially maps global cpu ID (as you can observe as `processor` in `/proc/cpuinfo`) to a relative CPU id within a node. To make the code simple I'd propose you to rewrite it in 2 loops: one for each node to allocate `cpu` and the second one for each online CPU to write the mappings.

Also I suggest to add an assertion like `BUG_ON(c_nodes[node].nr_cpus > nr_cpus_node(node))`

fw/cache.c
```diff
@@ -3159,7 +3189,8 @@ tfw_cache_start(void)
 		goto close_db;
 	}
 #endif
-	tfw_init_node_cpus();
+	if(tfw_init_node_cpus() == -ENOMEM) 
+		goto close_db;
```
krizhanovsky, line 3193: In case of error you're leaking memory. You need to add a new label to free the allocated per-node CPU maps.

krizhanovsky (changes requested):
Many problems with memory initialization, freeing and leaking

fw/cache.c
```diff
@@ -328,21 +328,51 @@ tfw_cache_key_node(unsigned long key)
...
 tfw_init_node_cpus(void)
 {
-	int cpu, node;
+	int nr_cpus, cpu, node;
+
+	nr_cpus = num_online_cpus();
 
 	for_each_online_cpu(cpu) {
+
+
```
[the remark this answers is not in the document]
enuribekov-tempesta: Fixed

fw/cache.c
```diff
@@ -328,21 +328,51 @@ tfw_cache_key_node(unsigned long key)
...
 		node = cpu_to_node(cpu);
+
+		pr_err("node.nr_cpus %u",c_nodes[node].nr_cpus);
+
+		c_nodes[node].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL);
+		if(!c_nodes[node].cpu) {
+			T_ERR("%s Failed to allocate c_nodes[%i] cpu\n", __func__, node);
```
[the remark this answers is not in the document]
enuribekov-tempesta: Fixed

fw/cache.c
enuribekov-tempesta: What do you mean by the term 'adjust'?

fw/cache.c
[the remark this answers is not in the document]
enuribekov-tempesta: No. We allocate CPU array for every node, than fill it by CPU IDs. So  in case of 2 Node * 32 CPU we'll perform 2 allocations.
But indeed, in current code arrays are oversized. I'll change it size to `nr_cpus_node`, as well as will add `BUG_ON`.

fw/cache.c
```diff
@@ -328,21 +328,46 @@ tfw_cache_key_node(unsigned long key)
...
+ * Release memory dynamically allocated for every node cpus
+ */
+static void
+tfw_release_node_cpus(void)
+{
+	int node;
+
+	for_each_online_node(node) {
+		if(c_nodes[node].nr_cpus > 0)
+			kfree(c_nodes[node].cpu);
```
keshonok, line 341: Is this a protection against garbage in `c_nodes[node].cpu`? Otherwise, it's safe to `kfree()` any pointer, even a null pointer, without an extra check.

Also, after freeing it's highly desirable that you keep both `.nr_cpu` and `.cpu` consistent with the new state of being unused (i.e. zero). Again, if this state is maintained properly, then you don't need that extra check above.

An aside question: does this (and other) code handle CPUs that can be put online or offline, or CPU migration, or when the node that was online is now offline? Also, perhaps, you should `kfree()` whatever you have allocated and not just rely on these `_online_node()`, etc, functions?

enuribekov-tempesta: 1. Yes, but this code will be removed in the next version. 
2. Yes again, but  `c_nodes` also will be dynamically allocated, so we don't need it. But need to check `c_nodes`.
3. Currently assumed no CPU manipulation, just decreasing memory consumption to use it only for online nodes/CPUs.

fw/cache.c
keshonok: Perhaps, you should react on any error. That way you don't depend on possible future changes in the underlying function.

fw/cache.c
keshonok: Thanks. At first glance, I didn't find any stipulated requirements regarding CPU manipulation in Tempesta Wiki. If there is, perhaps, it should be made more obvious?

enuribekov-tempesta: > Thanks. At first glance, I didn't find any stipulated requirements regarding CPU manipulation in Tempesta Wiki. If there is, perhaps, it should be made more obvious?

I don't know. I definitely saw such information, but don't remember where is.

fw/cache.c
```diff
@@ -3145,7 +3179,10 @@ tfw_cache_start(void)
 	if (!(cache_cfg.cache || g_vhost->cache_purge))
 		return 0;
 
-	for_each_node_with_cpus(i) {
+	if (!tfw_init_node_cpus())
```
EvgeniiMekhanik, line 3182: You close db in case of SUCCESS!, more over you should not go to close_db because db is not open! (I catch three different kernel bugs here!)

EvgeniiMekhanik: if ((r = tfw_init_node_cpus())) is good

fw/cache.c
```diff
@@ -3145,7 +3179,10 @@ tfw_cache_start(void)
...
+		goto close_db;
+
+	for(i = 0; i < nr_online_nodes; i++) {
 		c_nodes[i].db = tdb_open(cache_cfg.db_path,
 					 cache_cfg.db_size, 0, i);
 		if (!c_nodes[i].db)
```
EvgeniiMekhanik, line 3188: If allocation fails please set r = -ENOMEM;

fw/cache.c
```diff
@@ -329,20 +330,53 @@ tfw_cache_key_node(unsigned long key)
...
+	}
 
 	for_each_online_cpu(cpu) {
 		node = cpu_to_node(cpu);
+		nr_cpus = nr_cpus_node(node);
+		T_DBG("node - nr_cpus %i - %i",node, nr_cpus);
+		c_nodes[node].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL);
+		if (!c_nodes[node].cpu) {
+			T_ERR( "Failed to allocate a CPU %i for cache work scheduler", cpu);
+			tfw_release_node_cpus();
```
EvgeniiMekhanik, line 370: You should call it at the end of `tfw_cache_start` in case of error, otherwise it will never called if `tdb_open` failed for example

fw/cache.c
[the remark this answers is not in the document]
EvgeniiMekhanik: @krizhanovsky is right here! Just print node is always 0 here. c_nodes[node].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL); is always rewrite pointer!

EvgeniiMekhanik (changes requested):
A lot of mistakes. Please fix and also test new code. Simulate situations when allocations fails to be sure that we don't catch kernel panic and there is no memory leaks also

fw/cache.c
```diff
@@ -329,18 +330,53 @@ tfw_cache_key_node(unsigned long key)
...
+			return -ENOMEM;
+		}
+	}
 
 	for_each_online_cpu(cpu) {
 		node = cpu_to_node(cpu);
+		T_DBG("node - nr_cpus %i - %i",node, nr_cpus);
```
const-t, line 375:
```suggestion
		T_DBG2("node - nr_cpus %i - %i", node, nr_cpus);
```

fw/cache.c
```diff
@@ -3145,11 +3181,16 @@ tfw_cache_start(void)
 	if (!(cache_cfg.cache || g_vhost->cache_purge))
 		return 0;
 
-	for_each_node_with_cpus(i) {
+	if (r = tfw_init_node_cpus())
```
const-t, line 3184:
```suggestion
	if ((r = tfw_init_node_cpus()))
```
Did you try to compile your code?

fw/cache.c
```diff
@@ -3145,11 +3181,16 @@ tfw_cache_start(void)
...
+		goto close_db;
```
const-t, line 3185: As I see in `tfw_init_node_cpus()` allocation of `c_nodes` may fail, thus `c_nodes` will be NULL. Then you do
```
close_db:
	for_each_node_with_cpus(i)
		tdb_close(c_nodes[i].db);
```
that cause null ptr dereference.

Why we even try to close db if we didn't open any dbs so far?

const-t (changes requested):

fw/cache.c
enuribekov-tempesta: It looks like the last commit for some reasons haven't pushed.

fw/cache.c
EvgeniiMekhanik: This comment is not resolved! Add debug print to `tfw_release_node_cpus`! Then simulate situation when `tdb_open ` fails, you can see that `tfw_release_node_cpus` is not called! There is a memory leak. You should add `tfw_release_node_cpus` after `node_cpus_alloc_err` label  not here!

fw/cache.c
```diff
@@ -329,18 +330,55 @@ tfw_cache_key_node(unsigned long key)
...
+		if (!c_nodes) {
+			T_ERR( "Failed to allocate nodes map for cache work scheduler");
+			return -ENOMEM;
+	}
+	for_each_node_with_cpus(node) {
+		nr_cpus = nr_cpus_node(node);
+		T_DBG2("node: %d  nr_cpus: %d",node, nr_cpus);
+		c_nodes[node].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL);
+		if (!c_nodes[node].cpu) {
+			T_ERR( "Failed to allocate a CPU %i for cache work scheduler", cpu);
```
EvgeniiMekhanik, line 369: `cpu ` is not changed in this loop, here should be
`Failed to allocate CPU array for node %d for cache work scheduler`

fw/cache.c
```diff
@@ -329,18 +330,55 @@ tfw_cache_key_node(unsigned long key)
...
+			tfw_release_node_cpus();
```
EvgeniiMekhanik, line 370: This comment is not resolved! Add debug print to tfw_release_node_cpus! Then simulate situation when tdb_open fails, you can see that tfw_release_node_cpus is not called! There is a memory leak. You should add tfw_release_node_cpus after node_cpus_alloc_err label not here!

fw/cache.c
```diff
@@ -329,18 +330,54 @@ tfw_cache_key_node(unsigned long key)
...
+		if (!c_nodes) {
+			T_ERR( "Failed to allocate nodes map for cache work scheduler");
+			return -ENOMEM;
+	}
+	for_each_node_with_cpus(node) {
+		nr_cpus = nr_cpus_node(node);
+		T_DBG2("node: %d  nr_cpus: %d",node, nr_cpus);
+		c_nodes[node].cpu = kmalloc(nr_cpus * sizeof(int), GFP_KERNEL);
+		if (!c_nodes[node].cpu) {
+			T_ERR("Failed to allocate CPU array for node %d for cache work scheduler", node);
```
EvgeniiMekhanik, line 369: To long line

krizhanovsky (approved):
LGTM

fw/cache.c
```diff
@@ -329,18 +330,53 @@ tfw_cache_key_node(unsigned long key)
...
+ * At the moment we doesn't support CPU hotplug, so enumerate only online CPUs.
+ */
+static int
 tfw_init_node_cpus(void)
 {
-	int cpu, node;
+	int nr_cpus, cpu, node;
+
+	c_nodes = kzalloc(nr_online_nodes * sizeof(CaNode), GFP_KERNEL);
+		if (!c_nodes) {
```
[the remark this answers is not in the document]
const-t: @enuribekov-tempesta This comment is not addressed. Pleas fix it.

fw/cache.c
```diff
@@ -329,18 +330,55 @@ tfw_cache_key_node(unsigned long key)
...
- * nodes scheduling. Reserve 0th CPU for other tasks.
+ * Release node-cpu map.
  */
 static void
+tfw_release_node_cpus(void)
+{
+	int node;
+
+	for(node = 0; node < nr_online_nodes; node++) {
+		if(c_nodes[node].cpu)
```
const-t, line 341: You doing:
`c_nodes = kzalloc(nr_online_nodes * sizeof(CaNode), GFP_KERNEL);` and in case of allocation error you invoking `tfw_release_node_cpus()`, it implies you must check `c_nodes` here, otherwise you will got null ptr dereference. Also you can use `kfree(c_nodes[node].cpu);` without pre-check as @keshonok mentioned. I don't see any places where garbage can be moved to `c_nodes`.

const-t (approved):
LGTM

Clean commit history before merge, please

Issue #2058, opened 2024-02-20
Prohibit DELETE method in default configuration


const-t:
# Scope
RFC defines [DELETE](https://www.rfc-editor.org/rfc/rfc9110#name-delete) as rm for resource that must invalidate the cache on successful response. However, Tempesta MUST NOT invalidate the cache on every DELETE request, it leads to uncontrolled invalidating of arbitrary resources. Therefore, need to prohibit DELETE method if it not explicitly allowed in configuration or prohibit cache invalidation using DELETE method.

PR #2065, opened 2024-02-23
Mekhanik evgenii/fix 1995 5

EvgeniiMekhanik:

PR #2078, opened 2024-03-13
Prohibit DELETE method in default configuration #2058

biathlon3:
My first attempt, I am not sure that it is the best solution, but it seems to work.


fw/http_limits.c
```diff
@@ -1139,17 +1139,22 @@ frang_http_req_process(FrangAcc *ra, TfwConn *conn, TfwFsmData *data,
...
-			if (req->method == _TFW_HTTP_METH_NONE ||
-			    req->method == _TFW_HTTP_METH_INCOMPLETE) {
-				T_FSM_EXIT();
-			}
-			r = frang_http_methods(req, ra,
-					       f_cfg->http_methods_mask);
-		}
-		__FRANG_FSM_MOVE(Frang_Req_Hdr_UriLen);
-	}
+    T_FSM_STATE(Frang_Req_Hdr_Method) {
```
EvgeniiMekhanik, line 1142: Invalid tabulation.

fw/http_limits.c
```diff
@@ -1139,17 +1139,22 @@ frang_http_req_process(FrangAcc *ra, TfwConn *conn, TfwFsmData *data,
...
+        if (req->method == _TFW_HTTP_METH_NONE ||
+            req->method == _TFW_HTTP_METH_INCOMPLETE) {
+            T_FSM_EXIT();
+        }
+
+        if (!f_cfg->http_methods_mask) {
+            /* If we have a default value, we set all bits to 1 except DELETE */
+            f_cfg->http_methods_mask = ~(1 << TFW_HTTP_METH_DELETE);
```
EvgeniiMekhanik, line 1150: I think we should also prohibit PUT method here

EvgeniiMekhanik:
You should also fix tests, where DELETE method is used by adding appropriate DELETE method to frang config

krizhanovsky (changes requested):
Definitely need more work and it seems the issue isn't so straightforward as I thought, so we need a good discussion on this.

Also please make pull requests from our repo, not your own - this way we can just switch between branches during review instead of cloning repos of each developers.

Finally, please don't forget to update wiki about allowed methods by default. Probably it also makes sense to describe in the wiki how DELETE behaves with the cache.

fw/http_limits.c
```diff
@@ -1140,14 +1140,22 @@ frang_http_req_process(FrangAcc *ra, TfwConn *conn, TfwFsmData *data,
...
-					       f_cfg->http_methods_mask);
+		if (req->method == _TFW_HTTP_METH_NONE ||
+		    req->method == _TFW_HTTP_METH_INCOMPLETE)
+		{
+			T_FSM_EXIT();
 		}
+
+		if (!f_cfg->http_methods_mask) {
+			/* If we have a default value,
+			 * we set all bits to 1 except DELETE and PUT*/
```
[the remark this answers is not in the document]
biathlon3: Yes, I agree, this is important.

fw/http_limits.c
```diff
@@ -1140,14 +1140,22 @@ frang_http_req_process(FrangAcc *ra, TfwConn *conn, TfwFsmData *data,
...
+			f_cfg->http_methods_mask = ~((1 << TFW_HTTP_METH_DELETE) |
+						     (1 << TFW_HTTP_METH_PUT));
```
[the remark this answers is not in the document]
biathlon3: 1. I suggest doing it in tfw_frang_clean(FrangVhostCfg *cfg). We set it as default value and if we have something in the config file, it will be changed.

2. What do I have to do with this config?
    frang_limits {
        http_methods ;
    }
Because it also allows everything

3. What about OPTIONS in default?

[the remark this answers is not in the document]
krizhanovsky: It seems `tfw_vhost_cfgstart()` is a better function to store default values.

Also please check configuration reload (i.e. we had some values for `http_methods` and after that going to use defaults) on-the fly

fw/http_limits.c
```diff
@@ -1140,14 +1140,22 @@ frang_http_req_process(FrangAcc *ra, TfwConn *conn, TfwFsmData *data,
...
+		}
+
+		r = frang_http_methods(req, ra,
+				       f_cfg->http_methods_mask);
+
 		__FRANG_FSM_MOVE(Frang_Req_Hdr_UriLen);
 	}
```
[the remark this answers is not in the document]
krizhanovsky: There is related commit https://github.com/tempesta-tech/tempesta/pull/2065/commits/176cb6f9bf23be50fb3b093d75b1676e6c627272

fw/vhost.c
```diff
@@ -132,6 +132,11 @@ static const TfwCfgEnum tfw_method_enum[] = {
  */
 #define TFW_CAPUACL_ARRAY_SZ	(32)
 
+
+#define TFW_FRANG_HTTP_METHODS_MASK_DEFAULT ((1 << TFW_HTTP_METH_GET) | \
+		(1 << TFW_HTTP_METH_HEAD) | \
+		(1 << TFW_HTTP_METH_POST))
+
```
krizhanovsky, line 139:
```suggestion
#define TFW_FRANG_HTTP_METHODS_MASK_DEFAULT ((1 << TFW_HTTP_METH_GET) | \
                                             (1 << TFW_HTTP_METH_HEAD) | \
                                             (1 << TFW_HTTP_METH_POST))   

```
We almost never use 2 empty lines in C/C++ code and we use alignments (tabs plus spaces)

krizhanovsky (approved):
LGTM with just a small coding style cleanup

const-t (changes requested):
I would like to suggest another solution. In `TfwCfgSpec` we can specify default values for each directive.
Example:
```
{
		.name = "http_methods",
		.deflt = "get post head",
		.handler = tfw_cfgop_frang_http_methods,
		.allow_reconfig = true,
	},
```
Field `deflt` it's just value, same as in config, that will be parsed and assigned to directive. Please, change `TfwSfgSpec` for **each context**.

fw/vhost.c
```diff
@@ -1904,6 +1908,19 @@ __tfw_cfgop_frang_http_methods(TfwCfgSpec *cs, TfwCfgEntry *ce,
 		methods_mask |= (1UL << method_id);
 	}
 
+	if (!ce->val_n) {		
+		if (!ce->dflt_value) {
+			T_ERR_NL("frang: Empty http_methods.");
+			return -EINVAL;
+		}
+
+		T_ERR_NL("frang: http_methods should contain at least one "
```
const-t, line 1917: Looks like dead code and we never reach this place. If we have `ce->dflt_value == false` it means value was specified in config and now we know that this value is empty. Therefore, we will always return in `if (!ce->dflt_value)`. Is there a case when `ce->dflt_value == true` but `ce->val_n == false`?

biathlon3: This situation is possible  when we have  " http_methods  ;" in config

fw/vhost.c
```diff
@@ -3220,7 +3237,7 @@ static TfwCfgSpec tfw_vhost_internal_specs[] = {
 		.handler = tfw_cfg_handle_children,
 		.cleanup = tfw_cfgop_frang_cleanup,
 		.dest = tfw_vhost_frang_specs,
-		.allow_none = true,
+		.allow_none = false,
```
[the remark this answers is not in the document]
biathlon3: If we leave allow_none as true, the default configuration won't be read in spec_finish_handling(TfwCfgSpec specs[]).
It was the reason why I wrote my previous version.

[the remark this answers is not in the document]
keshonok: This seems wrong. Are you sure? There are plenty of places in the code with both `.allow_none = true` and `.deflt` set to some value.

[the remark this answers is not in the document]
biathlon3: Yes, guys, I was wrong.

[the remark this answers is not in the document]
biathlon3: I tested it more careful and found that with .allow_none = true, in this position, it stops using default configuration when vhost is configured in tempesta_fw.cfg.
Looks like I hastened to agree with you. And we need to use .allow_none = false or change behavior of spec_finish_handling(TfwCfgSpec specs[]) or change somthing else.

const-t:
Please remove todo from `http_limits.c`, when you will rebase.
`TODO: Remove when PR #2078 merged`

fw/vhost.c
const-t: In my understanding  when we have " http_methods ;" then we have `ce->dflt_value == false` and `ce->val_n == false` in other cases we will have `ce->val_n > 0`. For example: http_methods ""; or http_methods " "; will have `ce->val_n > 0` and will disallowed in `TFW_CFG_ENTRY_FOR_EACH_VAL`. I can't find another value for http_methods where `ce->val_n == 0` except " http_methods ;". Do you know such value for http_methods?

biathlon3: Yes, when we have 
		.name = "http_methods",
		.deflt = "",
in the code

const-t: Now we doesn't, and I believe we must not maintain this. If you warring about this, it would be better to write a comment in appropriate place.

fw/http_limits.c
```diff
@@ -1142,14 +1140,13 @@ frang_http_req_process(FrangAcc *ra, TfwConn *conn, TfwFsmData *data,
 
 	/* Ensure that HTTP request method is one of those defined by a user. */
 	T_FSM_STATE(Frang_Req_Hdr_Method) {
-		if (f_cfg->http_methods_mask) {
```
biathlon3, line 1145: I removed checking f_cfg->http_methods_mask  because this checking breaks new logic.

krizhanovsky:
Please implement a workaround for #2119 with a TODO comment

biathlon3:
> Please implement a workaround for #2119 with a TODO comment

I manage to do it in normal way.

const-t:
A want to ask you to check following(and others similar conditions) part in `tfw_http_req_process()`:

> if (res.type == TFW_HTTP_RES_REDIR) {
> 	tfw_http_req_redir(req, res.redir.resp_code, &res.redir);
> 	return T_OK;
> }

In this case we don't freed split skb as well. To check it just use redirects.
```
http_chain redirects {
    # Absolute path
    uri == "/company.html" -> 301 = https://tempesta-tech.com/company;

    # Use variables
    uri == "/blog/" -> 308 = https://$host/$request_uri/;

    # Relative path
    uri == "*/services.html" -> 303 = /services;

    # Temporal redirection for the default index page only to a temporal landing page.
    uri == "/" -> 307 = /c++-services;
}
```

fw/tls.c
```diff
@@ -1128,6 +1128,11 @@ tfw_tls_start(void)
 	return 0;
 }
 
+bool tfw_tls_get_allow_any_sni_reconfig(void)
```
krizhanovsky, line 1131:
```suggestion
bool
tfw_tls_get_allow_any_sni_reconfig(void)
```

const-t (approved):
In general looks good, but could you add a little bit description to the last commit.
> Chenged frang inheritance for locatians in default vhost
Why it's changed, how it's changed.
Also looks like you fixed live-reconfig logic for frang, describe it also in the commit message.

EvgeniiMekhanik (changes requested):
Change fixing memory leak

etc/tempesta_fw.conf
```diff
@@ -1071,7 +1075,31 @@
...
+#       tls_connection_burst 0;
+#       tls_incomplete_connection_rate 0;
+#       client_header_timeout 0;
+#       client_body_timeout 0;
+#       http_header_chunk_cnt 0;
+#       http_body_chunk_cnt 0;
+#       http_uri_len 0;
+#       http_body_len 1073741824;
+#       http_hdr_len 0;
+#       http_header_cnt 0;
```
EvgeniiMekhanik, line 1095: 50 is default value for this setting. Please also check all other settings

fw/vhost.c
```diff
@@ -2141,7 +2197,7 @@ tfw_cfgop_frang_strict_host_checking(TfwCfgSpec *cs, TfwCfgEntry *ce)
 	int r;
 	FrangVhostCfg *cfg = tfw_cfgop_frang_get_cfg();
 
-	if (ce->dflt_value && cfg->http_strict_host_checking)
+	if (ce->dflt_value && !cfg->http_strict_host_checking)
```
EvgeniiMekhanik, line 2200: We need to discuss this place. If I understand it correctly we have the same problem for `http_body_len` also and may be some other settings.

fw/vhost.c
```diff
@@ -1903,6 +1911,11 @@ __tfw_cfgop_frang_http_methods(TfwCfgSpec *cs, TfwCfgEntry *ce,
 		methods_mask |= (1UL << method_id);
 	}
 
+	if (!ce->val_n && !ce->dflt_value) {
```
EvgeniiMekhanik, line 1914: You don't need to check !ce->dflt_value here, because if it is dflt_value vam_n is always 3 (get post head).  This check is better to do before `TFW_CFG_ENTRY_FOR_EACH_VAL` loop

EvgeniiMekhanik: TFW_CFG_CHECK_VAL_N(>, 0, cs, ce);
TFW_CFG_CHECK_NO_ATTRS(cs, ce);

is right way. Tempesta interpret get=hhhsdsd as attribute and silently ignore such invalid configuration

fw/vhost.c
```diff
@@ -2733,56 +2801,66 @@ static TfwCfgSpec tfw_global_frang_specs[] = {
...
-		.handler = tfw_cfgop_frang_uri_len,
+		.handler = tfw_cfgop_frang_glob_set_int,
+		.dest = &tfw_frang_vhost_reconfig.http_uri_len,
 		.allow_reconfig = true,
 	},
 	{
 		.name = "http_body_len",
 		.deflt = "1073741824", /* 1 Gb. */
-		.handler = tfw_cfgop_frang_body_len,
+		.handler = tfw_cfgop_frang_glob_set_long,
```
EvgeniiMekhanik, line 2811: It is incorrect to use `tfw_cfgop_frang_glob_set_long` here. If we have global frang limits like below:
frang_limits {
    http_body_len 0;
 }
 frang_limits {
 }

We should not override `http_body_len 0;` in the second config as i understand the idea in all other places

fw/vhost.c
```diff
@@ -2733,56 +2801,66 @@ static TfwCfgSpec tfw_global_frang_specs[] = {
...
+		.dest = &tfw_frang_vhost_reconfig.http_body_len,
 		.allow_reconfig = true,
 	},
 	{
 		.name = "http_strict_host_checking",
 		.deflt = "true",
-		.handler = tfw_cfgop_frang_strict_host_checking,
+		.handler = tfw_cfgop_frang_glob_set_bool,
```
EvgeniiMekhanik, line 2818: We can't use tfw_cfgop_frang_glob_set_bool here because default value is true.
frang_limits {
    http_strict_host_checking false;
 }
 frang_limits {
 }

We have http_strict_host_checking true, but should have false

fw/vhost.c
```diff
@@ -1345,8 +1347,16 @@ tfw_location_new(TfwVhost *vhost, tfw_match_t op, const char *arg, size_t len)
 		return NULL;
 	vhost->loc_sz++;
 
-	if (tfw_frang_cfg_inherit(loc->frang_cfg, vhost->loc_dflt->frang_cfg))
-		return NULL;
+	if (strncasecmp(vhost->name.data, TFW_VH_DFT_NAME, vhost->name.len)) {
+		if (tfw_frang_cfg_inherit(loc->frang_cfg,
```
EvgeniiMekhanik, line 1351: Please look at this function `tfw_frang_cfg_inherit` we have BUG in it.
if (!curr->http_ct_vals)
 we only set r = -ENOMEM and do not return. Then we memcpy(curr->http_ct_vals, from->http_ct_vals, sz);

fw/vhost.c
```diff
@@ -1345,8 +1347,16 @@ tfw_location_new(TfwVhost *vhost, tfw_match_t op, const char *arg, size_t len)
 		return NULL;
 	vhost->loc_sz++;
 
-	if (tfw_frang_cfg_inherit(loc->frang_cfg, vhost->loc_dflt->frang_cfg))
-		return NULL;
+	if (strncasecmp(vhost->name.data, TFW_VH_DFT_NAME, vhost->name.len)) {
```
EvgeniiMekhanik, line 1350: It seems that it is totally wrong or I don't understand the idea.
frang_limits {
     http_body_len 30;
}
 frang_limits {
      http_body_len 90;
 }

 location prefix "/static/" {
     frang_limits {
          http_body_len 270;
     }
 }

What http_body_len we should have here? I think that 270. If there is no frang_limits in location we should have 90, but I print that we have default value 1073741824. Please add test how it works.

fw/http_limits.c
```diff
@@ -684,13 +684,6 @@ frang_http_domain_fronting_check(const TfwHttpReq *req, FrangAcc *ra)
 	if (!(TFW_CONN_TYPE(req->conn) & TFW_FSM_HTTPS))
 		return T_OK;
 
-	/*
-	 * Do not perform any SNI<=>authority validation if
-	 * tls_match_any_server_name is set.
-	 */
-	if (tfw_tls_allow_any_sni)
```
EvgeniiMekhanik, line 693: Why do you delete this check? Previously when tls_match_any_server_name is true, we return T_OK. Why this behavior is changed?

fw/vhost.c
```diff
@@ -2056,6 +2069,52 @@ tfw_cfgop_frang_glob_set_int(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+		kfree(cfg->http_ct_vals);
+		cfg->http_ct_vals = NULL;
+	}
+	return __tfw_cfgop_frang_http_ct_vals(cs, ce, cfg);
+}
+
+static int
+tfw_cfgop_frang_glob_rsp_code_block(TfwCfgSpec *cs, TfwCfgEntry *ce)
+{
+	FrangVhostCfg *cfg = &tfw_frang_vhost_reconfig;
```
EvgeniiMekhanik, line 2107: Why we can't use `tfw_cfgop_frang_rsp_code_block`. The only difference that here we directly set FrangVhostCfg *cfg = &tfw_frang_vhost_reconfig; instead of FrangVhostCfg *cfg = tfw_cfgop_frang_get_cfg(); ?

fw/vhost.c
```diff
@@ -2056,6 +2069,52 @@ tfw_cfgop_frang_glob_set_int(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+
+	if (cfg->http_resp_code_block) {
+		if (ce->dflt_value)
+			return 0;
+		kfree(cfg->http_resp_code_block);
+		cfg->http_resp_code_block = NULL;
+	}
+	return __tfw_cfgop_frang_rsp_code_block(cs, ce, cfg);
```
EvgeniiMekhanik, line 2115: Can you also check `__tfw_cfgop_frang_rsp_code_block`. I see that we return -EINVAL after memory allocation without
freeing previously allocated cb = kzalloc(sizeof(FrangHttpRespCodeBlock), GFP_KERNEL); Is there memory leak or we will free it later?

fw/http_limits.c
const-t: When `tfw_tls_allow_any_sni` is true, it means connection with empty sni or sni not listed in our vhost is "routed" to default vhost. In other words `tfw_tls_allow_any_sni` is the way to enable default vhost and it's not related to other vhosts. But when we have this check here and we want to check SNI vs authority for some vhosts and don't check for other we can't do this, because `tfw_tls_allow_any_sni` will override it.

Now I would prefer to introduce new knob only for domain fronting, that must be disabled by default. It will depends on `tfw_tls_allow_any_sni` also, but in the future we will remove `default` vhost and `tfw_tls_allow_any_sni` and will check SNI only when vhost is found by http tables.

fw/vhost.c
biathlon3: It assigns to conf->http_resp_code_block and will be freed by __tfw_frang_clean() in case of error.

fw/http.c
```diff
@@ -6090,9 +6090,14 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb,
 	/* Don't accept any following requests from the peer. */
 	if (r == T_BLOCK) {
 		TFW_INC_STAT_BH(clnt.msgs_filtout);
-		return tfw_http_req_parse_block(req, 403,
+		r = tfw_http_req_parse_block(req, 403,
 				"parsed request has been filtered out",
 				HTTP2_ECODE_PROTO);
+		if (skb) {
```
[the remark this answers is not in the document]
biathlon3: Code reworked. I tried to make it universal, but  if you think it is not the best way, I will rework it for special cases only.

etc/tempesta_fw.conf
biathlon3: Fixed

fw/vhost.c
biathlon3: I returned the old behavior, only explicit parameters should be overridden.

fw/vhost.c
biathlon3: Reworked.

fw/vhost.c
biathlon3: I returned the old behavior, only explicit parameters should be overridden.

fw/vhost.c
biathlon3: Add else to avoid using curr->http_ct_vals after error.

fw/vhost.c
biathlon3: It was attempt to solve strange location inheritance for default vhost. I returned it back now because my code had changed overriding.

biathlon3:
This PR was replaced by PR #2139

PR #2079, opened 2024-03-14
Mekhanik evgenii/fix part of 2059

EvgeniiMekhanik:


fw/hpack.c
```diff
@@ -1233,7 +1235,12 @@ tfw_hpack_hdr_set(TfwHPack *__restrict hp, TfwHttpReq *__restrict req,
 	 */
 	if (hp->index <= HPACK_STATIC_ENTRIES) {
 		WARN_ON_ONCE(s_hdr->nchunks > 2);
-		if (s_hdr->nchunks != 2)
+		/*
+		 * According to RFC 7230 3.2:
+		 * header name must be at least 1 byte, and the value can be 0
+		 * or more characters.
+		 */
+		if (s_hdr->nchunks < 1 || s_hdr->nchunks > 2)
```
const-t, line 1243: Why do we have these conditions? Looks like we runtime check predefined values.

const-t (approved):
LGTM

EvgeniiMekhanik:
Yes, I can't reproduce BUG_ON(!tcp_skb_pcount(skb)) and has nothing to do with two other problems in this issue

Issue #1695, opened 2022-09-19
Compilation issues using GCC12


s0nx:
# Motivation

There are some compiling issues when using GCC12.

# Scope
Fedora 36 with GCC 12.2.1 was used to build Tempesta.

Current issues:
- [ ] (-Werror=address issues #1689 )

enuribekov-tempesta:
@krizhanovsky 
See no issues with gcc-12.2.0 on Ubuntu.
Is it make sense to check it on Fedora?

enuribekov-tempesta:
It looks like such behavior is correct and correctness of pointer should be checked as direct comparation with NULL

`-Waddress`
Warn about suspicious uses of address expressions. These include comparing the address of a function or a declared object to the null pointer constant such as in
```
void f (void);
void g (void)
{
  if (!f)   // warning: expression evaluates to false
    abort ();
}

```
Such uses typically indicate a programmer error: the address of most functions and objects necessarily evaluates to true (the exception are weak symbols), so their use in a conditional might indicate missing parentheses in a function call or a missing dereference in an array expression.

[https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)

PR #2082, opened 2024-03-22
FIXME removed

enuribekov-tempesta:
Because gcc 12/13 requires explicit comparation pointer with NULL, current solution taken as permanent.
See folowing details in #1695


const-t (approved):
LGTM. Looks like the last case from [this](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Waddress) doc page. Here we must use explicit equation with `NULL`.
```
void f (const int *p, int i)
{
  return p + i == NULL;
}
```

Issue #2006, opened 2023-11-09
Handle HTTPS/1 and HTTP/2 on the same port


krizhanovsky:
# Motivation

For now we see that there are HTTP/1 clients attempting to access our HTTP/2-only website. https://w3techs.com/technologies/details/ce-http2 reports that there are only 40% of Internet clients are HTTP/2 and the rest 60% are definitely not HTTP/3. https://blog.cloudflare.com/http3-usage-one-year-on/ also reports that there are a lot on API clients using HTTP/1.

# Scope

Extend [listen](https://github.com/tempesta-tech/tempesta/wiki/Handling-clients#listening-address) directive to make it allow HTTPS along HTTP/2 on the same port, like
```
listen 443 proto=h2,https
```
The protocol, and HTTP/1 or HTTP/2 parsers, must be defined on ALPN.

# Testing

- [ ] Both the protocols are defined and both actually work for http/1 and http/2 requests
- [ ] http/2 defined and http/1 request fails on APLN with an appropriate warning
- [ ] http/1 defined and http/2 request fails on APLN with an appropriate warning

# Documentation

Update https://github.com/tempesta-tech/tempesta/wiki/Handling-clients#listening-address

krizhanovsky:
@ykargin could you please develop tests for the feature

Issue #1422, opened 2020-06-16
h2 regression for HTTP/1 cache


krizhanovsky:
## Scope

The original report is https://github.com/tempesta-tech/tempesta/issues/1415#issuecomment-641326456 : there are perf profile symbols for hpack in pure HTTP/1 traffic. We sacrificed HTTP/1 performance to make HTTP/2 faster, but still need to explore the the profile and improve all the things around.

## Testing

Just make a performance test with profiling. Probably, keep-alive connections case will show the problem better as it emphasises the cache functionality and reduces VM overhead.

krizhanovsky:
`$ ./wrk --latency -c 1000 -d 1000 -t 2 https://debian:443` and Tempesta FW caching 2-bytes file produce following profile:

```
     3.01%  [tempesta_fw]   [k] tfw_hpack_cache_decode_expand
     2.83%  [tempesta_fw]   [k] tfw_http_msg_expand_data
```

kingluo:
@krizhanovsky 
`tfw_http_msg_expand_data` is a required function that copies data, even for HTTP/2, so we only care about `tfw_hpack_cache_decode_expand`.

**My idea is to trade extra space for time efficiency: don't use a state machine to parse the hpack data, but directly locate the name and value data in the cache records (or static index table) and write to skbs.**

```
struct TfwHdrPtr {
	long idx; // if negative, means static table index, otherwise `ce->trec` index
	unsigned long len; // only used for `ce->trec`, may cross multiple records
} __attribute__((packed));


typedef struct {
	TdbVRec		trec;
	TfwHdrPtr*	ptrs; // data pointers used to build HTTP/1 response
	...
} TfwCacheEntry;


<---------------`ptrs` layout--------------->
---------------------------------------------
| n_idx | n_len | v_idx | v_len | ... | ... |
---------------------------------------------
<-----8bytes---->
<------16 bytes, one header----->

```

In `tfw_cache_h2_copy_hdr`, we have enough information to record the pointers:

```c
if (st_index) {
	if (tfw_cache_h2_copy_int(&ce->hdr_len, st_index, 0xF,  // <---- name static index
				  p, trec, tot_len))
		return -ENOMEM;
}
else {
	if (tfw_cache_h2_copy_int(&ce->hdr_len, 0, 0xF, p, trec,
				  tot_len)
		|| tfw_cache_h2_copy_int(&ce->hdr_len, s_nm.len,
					 0x7f, p, trec, tot_len)
[...]

PR #2083, opened 2024-03-26
The first attempt to solve #2006

biathlon3:
It works, but I may have missed releasing some resources.

### DIFF
--- a/fw/gfsm.h
+++ b/fw/gfsm.h
@@ -104,8 +104,19 @@ enum {
 	/* Security rules enforcement. */
 	TFW_FSM_FRANG_REQ,
 	TFW_FSM_FRANG_RESP,
+	TFW_FSM_H2_HTTPS	= 11,
 
 	TFW_FSM_NUM /* Must be <= TFW_GFSM_FSM_N */
+};
+
+/**
+ * Option HTTP/1.1 or HTTP2 for current client
+ * when the interface is configured to accept both protocols
+ * (TFW_FSM_H2_HTTPS).
+ */
+enum {
+	TFW_FSM_PROT_H1		= 0,
+	TFW_FSM_PROT_H2		= 1
 };
 
 #define TFW_FSM_TYPE(t)		((t) & TFW_GFSM_FSM_MASK)
--- a/fw/sock_clnt.c
+++ b/fw/sock_clnt.c
@@ -44,6 +44,7 @@
 static struct kmem_cache *tfw_h1_conn_cache;
 static struct kmem_cache *tfw_https_conn_cache;
 static struct kmem_cache *tfw_h2_conn_cache;
+static struct kmem_cache *tfw_h2_https_conn_cache;
 static int tfw_cli_cfg_ka_timeout = -1;
 
 static inline struct kmem_cache *
@@ -58,6 +59,8 @@ tfw_cli_cache(int type)
 	case TFW_FSM_HTTP:
 	case TFW_FSM_WS:
 		return tfw_h1_conn_cache;
+	case TFW_FSM_H2_HTTPS:
+		return tfw_h2_https_conn_cache;
 	default:
 		BUG();
 	}
@@ -429,7 +432,9 @@ tfw_sk_prepare_xmit(struct sock *sk, struct sk_buff *skb, unsigned int mss_now,
 	}
 
 	*nskbs = UINT_MAX;
-	h2_mode = TFW_CONN_PROTO(conn) == TFW_FSM_H2;
+	h2_mode = TFW_CONN_PROTO(conn) == TFW_FSM_H2 ||
+		  (TFW_CONN_PROTO(conn) == TFW_FSM_H2_HTTPS &&
+		   conn->proto.option == TFW_FSM_PROT_H2);
 	if (h2_mode) {
 		r = tfw_h2_sk_prepare_xmit(sk, skb, mss_now, limit, nskbs);
 		if (unlikely(r && r != -ENOMEM))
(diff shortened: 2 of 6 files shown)


fw/tls.c
```diff
@@ -664,10 +665,10 @@ tfw_tls_conn_init(TfwConn *c)
 		goto err_cleanup;
 	}
 
-	if (TFW_FSM_TYPE(c->proto.type) == TFW_FSM_H2)
+	if (TFW_FSM_TYPE(c->proto.type) == TFW_FSM_H2 ||
+	    TFW_FSM_TYPE(c->proto.type) == TFW_FSM_H2_HTTPS)
```
const-t, line 669: Another overhead for HTTP1, we initialize h2 context(including hpack) even for http1. Would be better to do it, when we know protocol.

@krizhanovsky Why do we init higher layer protocol context(http and http2) before finishing lower layer connection establishing, I mean TLS. Is there a reason or just implemented this way?

const-t (changes requested):

fw/tls.c
krizhanovsky: Yes, this is kind of reversion of https://github.com/tempesta-tech/tempesta/commit/3c97cb23d - we should initialize HTTP/2 context only for HTTP/2 connections. Note also https://github.com/tempesta-tech/tempesta/issues/1422 - another HTTP/2 -related regression in HTTP/1.

@const-t Frankly, I don't remember why do we initialize HTTP/2 context before finishing TLS handshake. It makes sense to allocate memory necessary for HTTP/2 early to improve spacial locality and reduce number of allocator calls. But for now for me it seems that we can initialize the context later... Need to check commits history - I suppose this point should have been discussed at some point.

fw/sync_socket.h
```diff
@@ -32,6 +32,7 @@
 typedef struct ss_proto_t {
 	const struct ss_hooks	*hooks;
 	int			type;
+	int			option;
```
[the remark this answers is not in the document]
biathlon3: > Can we override `type` when we know protocol? By default(no alpn) we will use HTTP2 in case when HTTP2 is specified to socket. Also, we always must prefer HTTP2 over HTTP1.

I rewrite code without "option"

fw/tls.c
biathlon3: > Another overhead for HTTP1, we initialize h2 context(including hpack) even for http1. Would be better to do it, when we know protocol.

I replace initialization h2 context for TFW_FSM_H2_HTTPS to tfw_http_msg_process

fw/sock_clnt.c
```diff
@@ -1215,6 +1227,15 @@ tfw_sock_clnt_init(void)
 		return -ENOMEM;
 	}
 
+	tfw_h2_https_conn_cache = kmem_cache_create("tfw_h2_https_conn_cache",
```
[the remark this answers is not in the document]
biathlon3: I used tfw_h2_conn_cache instead of tfw_h2_https_conn_cache.

fw/gfsm.h
```diff
@@ -104,6 +104,7 @@ enum {
 	/* Security rules enforcement. */
 	TFW_FSM_FRANG_REQ,
 	TFW_FSM_FRANG_RESP,
+	TFW_FSM_H2_HTTPS	= 11,
```
const-t, line 107: I suggesting another solution, please consider it:
1. Postpone all initialization of http2/http1 proto after handshake completeness.
2. Add to `ss_proto_t` flag `negotiable` or similar.
3. By default always use `TFW_FSM_H2` or `TFW_FSM_HTTP/S ` proto type.
4. If `negotiable` is true, at ALPN matching stage you just change the proto, to specified by ALPN.

With this approach we don't need `TFW_FSM_H2_HTTPS` dummy state.

fw/tls.c
```diff
@@ -615,6 +615,9 @@ tfw_tls_conn_dtor(void *c)
 	if (TFW_FSM_TYPE(((TfwConn *)c)->proto.type) == TFW_FSM_H2)
 		tfw_h2_context_clear(tfw_h2_context(c));
 
+	if (((TfwConn *)c)->proto.type & Conn_Negotiable)
+		((TfwConn *)c)->proto.type = TFW_FSM_H2;
```
const-t, line 619: Why do we need this?

fw/sock_clnt.c
```diff
@@ -616,13 +616,15 @@ static const SsProto tfw_sock_listen_protos[] = {
...
+
+	{ &tfw_sock_tls_clnt_ss_hooks,	TFW_FSM_H2 | Conn_Negotiable},
+	{ &tfw_sock_tls_clnt_ss_hooks,	Conn_H2Clnt | Conn_Negotiable},
 };
 
 static const SsProto *
 tfw_sock_clnt_proto(int type)
 {
 	int i;
-
```
const-t, line 625: Return this blank line please.

fw/sock_clnt.c
```diff
@@ -120,7 +120,10 @@ tfw_cli_conn_free(TfwCliConn *cli_conn)
 	tfw_connection_validate_cleanup((TfwConn *)cli_conn);
 	BUG_ON(!list_empty(&cli_conn->seq_queue));
 
-	kmem_cache_free(tfw_cli_cache(TFW_CONN_TYPE(cli_conn)), cli_conn);
+	if (!(TFW_CONN_TYPE(cli_conn) & Conn_Negotiable))
```
const-t, line 123: Please add comment to the code that describes why we have this condition.

const-t (approved):
LGTM. Please squash unnecessary commits.

krizhanovsky (approved):
LGTM

PR #2086, opened 2024-04-01
[Installer] Obsolete OS distributions have been removed

ykargin:


const-t (approved):
I can approve relying on our [wiki](https://github.com/tempesta-tech/tempesta/wiki/Install-from-packages).

RomanBelozerov (approved):

PR #2089, opened 2024-04-02
Checking if conn was be freed earlier

biathlon3:
We need to check client connection before sending an error and just free http connection (tfw_http_conn_msg_free) if the client connection was closed.


keshonok:
I may be mistaken, but generally speaking, this feels wrong. As long as there are users of a connection, the `refcnt` of the connection should hold a value which would not allow it to be freed. That's why there was no need to check for non-zero connection pointer before. I tend think the bug is somewhere else.

biathlon3:
> I may be mistaken, but generally speaking, this feels wrong. As long as there are users of a connection, the `refcnt` of the connection should hold a value which would not allow it to be freed. That's why there was no need to check for non-zero connection pointer before. I tend think the bug is somewhere else.

Unfortunately I was unable to reproduce this error and I had to use only my imagination. But I am sure we try to use NULL pointer in this place. If you have any idea about what really happened, welcome to help.

EvgeniiMekhanik:
I agree with @keshonok. I think that we should reproduce this BUG and only after understanding the problem fix it. If one of these pointers is equal to zero there is a broken logic in some other place

biathlon3:
Hi guys, is this not explanation why we need to check conn before using?

* NOTE: @hm->conn may be NULL if @hm is the response that was served
 * from cache.
 */
void
tfw_http_conn_msg_free(TfwHttpMsg *hm)

keshonok:
> Hi guys, is this not explanation why we need to check conn before using?
> 
>     * NOTE: @hm->conn may be NULL if @hm is the response that was served
> 
>     * from cache.
>       */
>       void
>       tfw_http_conn_msg_free(TfwHttpMsg *hm)

The comment is about a response. Aren't you trying to check the `conn` pointer in a request?

biathlon3:
This situation happens when a server connection failed or was closed for too long, and requests in a queue for this connection stay there longer than max_jqage. In this situation, we make error messages for ordinary requests and put them to the queue for sending to clients. But health monitor requests dont have information about client and we don't check for it before creating an error message. It is the cause of this problem.
I added a chek for this and now app just remove these requests from the queue for server connection and free them.
It works, but I am not sure that it is the best scenario. Maybe It is better  to just leave them in the queue?

EvgeniiMekhanik (approved):

fw/http.c
```diff
@@ -1694,9 +1704,9 @@ tfw_http_req_zap_error(struct list_head *eq)
 
 	list_for_each_entry_safe(req, tmp, eq, fwd_list) {
 		list_del_init(&req->fwd_list);
-		if ((TFW_MSG_H2(req) && req->stream)
-		    || (!TFW_MSG_H2(req)
-			&& !test_bit(TFW_HTTP_B_REQ_DROP, req->flags)))
+		if ((TFW_MSG_H2(req) && req->stream) ||
```
EvgeniiMekhanik, line 1707: Unnecessary changes, please revert it and LGTM. I also implement test which reproduce this bug on master branch https://github.com/tempesta-tech/tempesta-test/pull/610

const-t (approved):

krizhanovsky:
I think it makes sense to squash all the commits into one and write a good comment with good description of the problem and the fix

PR #2090, opened 2024-04-02
Change WARNING with dump to error message.

enuribekov-tempesta:
Because non-positive value of shift in short tables is a normal behaviour during decoding corrupted Huffman sequence we don't need detailed dump, short error message will be enough

### DIFF
--- a/fw/hpack.c
+++ b/fw/hpack.c
@@ -702,11 +702,7 @@ ht_small:
 				hp->curr -= shift;
 			}
 			else {
-				/*
-				 * @shift for short tables must be greater
-				 * than zero.
-				 */
-				WARN_ON_ONCE(1);
+				T_ERR("Decoding error. Short table shift less than zero.");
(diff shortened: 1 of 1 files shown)


fw/hpack.c
```diff
@@ -702,11 +702,7 @@ tfw_huffman_decode(TfwHPack *__restrict hp, TfwHttpReq *__restrict req,
 				hp->curr -= shift;
 			}
 			else {
-				/*
-				 * @shift for short tables must be greater
-				 * than zero.
-				 */
-				WARN_ON_ONCE(1);
+				T_ERR("Decoding error. Short table shift less than zero.");
```
const-t, line 705: Usually we use `T_ERR` when happened something critical, like OOM. Please use `T_WARN` instead.

const-t (commented):
LGTM

const-t (approved):

PR #2093, opened 2024-04-04
Decrement connection counter if limit exceeded.

enuribekov-tempesta:
I'm not sure that it is the best place to decrement counter, but placing it closer to connection drop logic may led to side effects and will require additional testing.

### DIFF
--- a/fw/http_limits.c
+++ b/fw/http_limits.c
@@ -193,7 +193,7 @@ frang_acc_history_init(FrangAcc *ra, unsigned long ts)
 	/*
 	 * Increment connection counters even when we return T_BLOCK.
 	 * Linux will call sk_free() from inet_csk_clone_lock(), so our
-	 * frang_conn_close() is also called. @conn_curr is decremented
+	 * tfw_classify_conn_close() is also called. @conn_curr is decremented
 	 * there, but @conn_new is not changed. We count both failed
 	 * connection attempts and connections that were successfully
 	 * established.
@@ -333,7 +333,7 @@ frang_conn_new(struct sock *sk, struct sk_buff *skb)
 	r = frang_conn_limit(ra, dflt_vh->frang_gconf);
 
 	/*
-	 * In case of closing connection by FIN-ACK, frang_conn_close()
+	 * In case of closing connection by FIN-ACK, tfw_classify_conn_close()
 	 * is responsible for resources releasing.
 	 * In case of dropping connection by RST, we should release
 	 * resources here.
@@ -343,8 +343,8 @@ frang_conn_new(struct sock *sk, struct sk_buff *skb)
 			tfw_filter_block_ip(cli);
 			tfw_client_put(cli);
 		} else {
-			spin_lock(&ra->lock);
 			BUG_ON(!ra->conn_curr);
+			spin_lock(&ra->lock);
 			ra->conn_curr--;
 			spin_unlock(&ra->lock);
 			sk->sk_security = NULL;
--- a/fw/http_limits.c
+++ b/fw/http_limits.c
@@ -217,6 +217,7 @@ frang_conn_limit(FrangAcc *ra, FrangGlobCfg *conf)
 	if (unlikely(ra->conn_curr > conn_max)) {
 		frang_limmsg("connections max num.", ra->conn_curr,
 			     conn_max, &FRANG_ACC2CLI(ra)->addr);
+    ra->conn_curr--;
 		spin_unlock(&ra->lock);
 		return T_BLOCK;
 	}
@@ -226,6 +227,8 @@ frang_conn_limit(FrangAcc *ra, FrangGlobCfg *conf)
 	{
 		frang_limmsg("new connections burst", ra->history[i].conn_new,
 			     conf->conn_burst, &FRANG_ACC2CLI(ra)->addr);
+		BUG_ON(!ra->conn_curr);
+		ra->conn_curr--;
 		spin_unlock(&ra->lock);
 		return T_BLOCK;
 	}
@@ -241,6 +244,8 @@ frang_conn_limit(FrangAcc *ra, FrangGlobCfg *conf)
 			frang_limmsg("new connections rate",
 				     csum, conf->conn_rate,
 				     &FRANG_ACC2CLI(ra)->addr);
+			BUG_ON(!ra->conn_curr);
+			ra->conn_curr--;
 			spin_unlock(&ra->lock);
 			return T_BLOCK;
 		}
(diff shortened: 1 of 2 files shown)


const-t (changes requested):
I think we should make another solution, and it's not a good place to decrement `conn_curr`. We have `tfw_classify_conn_close()` where we must decrement `conn_curr`, but looks like in some circumstances it not calls, I suggest to find out why it works this way.

@EvgeniiMekhanik You changed this place, maybe you have to add something?

enuribekov-tempesta:
> I think we should make another solution, and it's not a good place to decrement `conn_curr`. We have `tfw_classify_conn_close()` where we must decrement `conn_curr`, but looks like in some circumstances it not calls, I suggest to find out why it works this way.

We call `tfw_classify_conn_close()`,  but with wrong number on input.
'''
 [173693.331967] [tempesta fw] Warning: frang: connections max num. exceeded for 127.0.0.1: 9 (lim=2)
[173693.342409] frang_conn_new
[173693.342420] [tempesta fw] Warning: frang: connections max num. exceeded for 127.0.0.1: 10 (lim=2)
[173693.862733] classify_conn_close 10
[173693.862738] 9
[173693.931757] classify_conn_close 9
[173693.933752] 8
[173694.008727] frang_conn_new
[173694.010662] [tempesta fw] Warning: frang: connections max num. exceeded for 127.0.0.1: 9 (lim=2)
[173694.018024] frang_conn_new
[173694.018034] [tempesta fw] Warning: frang: connections max num. exceeded for 127.0.0.1: 10 (lim=2)
[173694.026732] frang_conn_new
'''
But OK, I'll try to find other place.

EvgeniiMekhanik:
As i wrote in private discussion, I think that the best place to call `tfw_classify_conn_close` is ss_linkerror`. When `tcp_done` is called there are two ways, if socket is dead `sk_destroy_cb = ss_conn_drop_guard_exit->tfw_classify_conn_close` is called.  If socket is not dead `ss_linkerror` is called, when socket state become TCP_CLOSE. If there is a connection for this socket `ss_conn_drop_guard_exit` is also called and everything is ok, but if frang_conn_limit fails there is no connection. So we should add:
```
else if (sk->security)
    tfw_classify_conn_close(sk);
```
in `ss_linkerror`

enuribekov-tempesta:
> As i wrote in private discussion, I think that the best place to call `tfw_classify_conn_close` is ss_linkerror`. When `tcp_done`is called there are two ways, if socket is dead`sk_destroy_cb = ss_conn_drop_guard_exit->tfw_classify_conn_close`is called. If socket is not dead`ss_linkerror`is called, when socket state become TCP_CLOSE. If there is a connection for this socket`ss_conn_drop_guard_exit` is also called and everything is ok, but if frang_conn_limit fails there is no connection. So we should add:
> 
> ```
> else if (sk->security)
>     tfw_classify_conn_close(sk);
> ```
> 
> in `ss_linkerror`

`ss_linkerror` is not called during connection reset.  But we need release resources in the case of a reset as well as in case of normal connection close.

enuribekov-tempesta:
New version is added.
1. I tracked the call stack and found that `tfw_classify_conn_close()` is not a part of FIN-drop scenery. For current architecture at least. So previous approach looks right for me.
2. Resources release moved to the  single point after `frang_conn_limit()` call.
3. Added clearing of `sk->security`

fw/http_limits.c
```diff
@@ -217,6 +217,7 @@ frang_conn_limit(FrangAcc *ra, FrangGlobCfg *conf)
 	if (unlikely(ra->conn_curr > conn_max)) {
 		frang_limmsg("connections max num.", ra->conn_curr,
 			     conn_max, &FRANG_ACC2CLI(ra)->addr);
+    ra->conn_curr--;
```
krizhanovsky, line 220: alignment issue

fw/http_limits.c
```diff
@@ -241,6 +244,8 @@ frang_conn_limit(FrangAcc *ra, FrangGlobCfg *conf)
 			frang_limmsg("new connections rate",
 				     csum, conf->conn_rate,
 				     &FRANG_ACC2CLI(ra)->addr);
+			BUG_ON(!ra->conn_curr);
```
krizhanovsky, line 247: Here and at the below I think it's better to use `WARN_ON()` or even `WARN_ON_ONCE()`: it seems we do not crash if `!ra->conn_curr` and we'll notice the problem in dmesg any way, but this `BUG_ON()` may lead to a crash.

fw/http_limits.c
```diff
@@ -331,11 +336,25 @@ frang_conn_new(struct sock *sk, struct sk_buff *skb)
...
-	}
 
+	/*
+	 * In case of closing connection by FIN-ACK, tfw_classify_conn_close()
+	 * is responsible for resources releasing.
+	 * In case of dropping connection by RST, we should release
+	 * resources here.
+	 */
+	if (unlikely(r == T_BLOCK)) {
+		if(dflt_vh->frang_gconf->ip_block) {
```
krizhanovsky, line 347:
```suggestion
		if (dflt_vh->frang_gconf->ip_block) {
```
Why this code is not in `frang_conn_limit()` to not to acquire the spin lock twice?

fw/http_limits.c
```diff
@@ -331,11 +336,25 @@ frang_conn_new(struct sock *sk, struct sk_buff *skb)
...
+			tfw_filter_block_ip(cli);
+			tfw_client_put(cli);
+		} else {
+			BUG_ON(!ra->conn_curr);
+			spin_lock(&ra->lock);
+			ra->conn_curr--;
```
krizhanovsky, line 353: I made only a quick review and @const-t  and @EvgeniiMekhanik  will provide more feedback, but for me the counter decrements and `tfw_client_put()` should happen where we actually drop a client connection.

fw/http_limits.c
enuribekov-tempesta: Done

fw/http_limits.c
enuribekov-tempesta: Done

fw/http_limits.c
enuribekov-tempesta: Short answer: because it was before, I just added else clause.
Long answer: because `frang_conn_limit()` has 4 return points and in this way we catch all possible results.
Sure, not a big problem to rework  `frang_conn_limit()` to make single return point.
Also it located in unlikely clause.
BTW it lead to another very interesting question. `T_BLOCK` considered as less probable limitation result. 
But in case of (D)DoS number of the blocked connections may on several orders of magnitude higher than enabled.
So what clause is really "likely" ?

enuribekov-tempesta:
Issue resolved in PR#2107

PR #2095, opened 2024-04-05
fix(1902): correct the skb len

### LINKED ISSUE #1902: Correct handling on trailer headers in h1 -> h2 proxying
# Scope
For response:
```python3
HTTP/1.1 200 OK
Content-type: text/html
Last-Modified: Wed, 14 Jun 2023 10:18:29 GMT
Date: Wed, 14 Jun 2023 10:18:29 GMT
Server: Deproxy Server
Transfer-Encoding: chunked
Trailer: X-Token

10
abcdefghijklmnop
10
qrstuvwxyzABCDEF
10
GHIJKLMNOPQRSTUV
4
WXYZ
0
X-Token: value
```
Tempesta forwarded h2 response:

```python3
:status: 200
content-type: text/html
last-modified: Wed, 14 Jun 2023 10:19:11 GMT
date: Wed, 14 Jun 2023 10:19:11 GMT
trailer: X-Token
x-token: value
via: 2.0 tempesta_fw (Tempesta FW pre-0.7.0)
content-length: 52
server: Tempesta FW/pre-0.7.0

abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ

```

and garbage as separate TCP frame:
```python3
b'Token: value\r\n\r\n'
```

# Testing

encoding.test_encoding.TestH2ChunkedWithTrailer

kingluo:
Fix #1902

When we contruct frames for HTTP/2 response and
reuse the same skbs from the backend (HTTP/1), we forget to adjust the skb size according to the trailer headers (if exist) size, then it comes with extra partial HTTP/1 response to the client. Then, the client will send a `GOAWAY` to temepesta due to wrong frame size, and tempesta sends back the `TCP RESET`.

### DIFF
--- a/fw/http.h
+++ b/fw/http.h
@@ -301,8 +301,7 @@ typedef struct {
 	TfwConn		*conn;						\
 	void (*destructor)(void *msg);					\
 	TfwStr		crlf;						\
-	TfwStr		body;						\
-	int		trailers_len;
+	TfwStr		body;
 
 
 static inline void
@@ -485,6 +484,7 @@ struct tfw_http_resp_t {
 	char			*body_start_data;
 	struct sk_buff		*body_start_skb;
 	TfwStr			cut;
+	int			trailers_len;
 };
 
 /**
--- a/fw/http.h
+++ b/fw/http.h
@@ -301,7 +301,8 @@ typedef struct {
 	TfwConn		*conn;						\
 	void (*destructor)(void *msg);					\
 	TfwStr		crlf;						\
-	TfwStr		body;
+	TfwStr		body;						\
+	int		trailers_len;
 
 
 static inline void
--- a/fw/http_msg.c
+++ b/fw/http_msg.c
@@ -568,6 +568,7 @@ tfw_http_msg_hdr_close(TfwHttpMsg *hm)
 	/* Cumulate the trailer headers length */
 	if (parser->hdr.flags & TFW_STR_TRAILER) {
 		TfwHttpResp* resp = (TfwHttpResp*) hm;
+
 		resp->trailers_len += parser->hdr.len +
 			tfw_str_eolen(&parser->hdr);
 	}
--- a/fw/http_msg.c
+++ b/fw/http_msg.c
@@ -565,6 +565,11 @@ tfw_http_msg_hdr_close(TfwHttpMsg *hm)
 	/* Close just parsed header. */
 	parser->hdr.flags |= TFW_STR_COMPLETE;
 
+	/* Cumulate the trailer headers length */
+	if (parser->hdr.flags & TFW_STR_TRAILER)
+		hm->trailers_len += parser->hdr.len +
+			tfw_str_eolen(&parser->hdr);
+
 	/*
 	 * We make this frang check here, because it is the earliest
 	 * place where we can determine that new added header is violating
@@ -1050,7 +1055,8 @@ tfw_http_msg_cutoff_body_chunks(TfwHttpResp *resp)
 	int r;
 
 	r = ss_skb_cutoff_data(resp->body.skb, &resp->cut, 0,
-			       tfw_str_eolen(&resp->body));
+			       tfw_str_eolen(&resp->body) +
+			       ((TfwHttpMsg*)resp)->trailers_len);
 	if (unlikely(r))
 		return r;
 
--- a/fw/http_msg.c
+++ b/fw/http_msg.c
@@ -566,9 +566,11 @@ tfw_http_msg_hdr_close(TfwHttpMsg *hm)
 	parser->hdr.flags |= TFW_STR_COMPLETE;
 
 	/* Cumulate the trailer headers length */
-	if (parser->hdr.flags & TFW_STR_TRAILER)
-		hm->trailers_len += parser->hdr.len +
+	if (parser->hdr.flags & TFW_STR_TRAILER) {
+		TfwHttpResp* resp = (TfwHttpResp*) hm;
+		resp->trailers_len += parser->hdr.len +
 			tfw_str_eolen(&parser->hdr);
+	}
 
 	/*
 	 * We make this frang check here, because it is the earliest
@@ -1056,7 +1058,7 @@ tfw_http_msg_cutoff_body_chunks(TfwHttpResp *resp)
 
 	r = ss_skb_cutoff_data(resp->body.skb, &resp->cut, 0,
 			       tfw_str_eolen(&resp->body) +
-			       ((TfwHttpMsg*)resp)->trailers_len);
+			       resp->trailers_len);
 	if (unlikely(r))
 		return r;
 


fw/http_parser.c
```diff
@@ -973,6 +973,8 @@ process_trailer_hdr(TfwHttpMsg *hm, TfwStr *hdr, unsigned int id)
 	hdr->flags |= TFW_STR_TRAILER;
 	__set_bit(TFW_HTTP_B_CHUNKED_TRAILER, hm->flags);
 
+	hm->trailers_len += tfw_str_len(hdr) + 2;
```
const-t, line 976: Seems `tfw_str_len` don't need here, you can simply do `hdr->len`, because `len` of "root" TfwStr always must contain length of all "child" chunks. Duplicates is not allowed here because you use currently parsing `hdr`.

And also, could you please replace magic number `2` here. As I understand it's length of CRLF, but IIRC in some cases we can have only one \n or \r, maybe `tfw_str_eolen` is more applicable here?

fw/http.h
```diff
@@ -299,7 +299,8 @@ typedef struct {
 	TfwConn		*conn;						\
 	void (*destructor)(void *msg);					\
 	TfwStr		crlf;						\
-	TfwStr		body;
+	TfwStr		body;						\
+	int		trailers_len;
```
const-t, line 303: We use it only with responses, maybe we should move `trailers_len` to `tfw_http_resp_t`? `cut` also located there.

const-t (changes requested):
You forgot to remove http.c.bak.

fw/http_parser.c
kingluo: > Seems tfw_str_len don't need here, you can simply do hdr->len, because len of "root" TfwStr always must contain length of all "child" chunks. Duplicates is not allowed here because you use currently parsing hdr.

okay, yes, I forgot this fact.

> could you please replace magic number 2 here. As I understand it's length of CRLF, but IIRC in some cases we can have only one \n or \r, maybe tfw_str_eolen is more applicable here?

No, for HTTP/1, the trailer chars are always `\r\n`, so the `len` is always `2`. I add some comment.

fw/http.h
kingluo: Yes, I agree. However, `process_trailer_hdr()` gets `TfwHttpMsg*` as argument, but it is called for both req an resp, so it seems no way to distinguish them here?

fw/http_parser.c
biathlon3: > No, for HTTP/1, the trailer chars are always `\r\n`, so the `len` is always `2`. I add some comment.

You can use SLEN(S_CRLF) instead and it will be understandable without comments. 
And It's better to use a consistent comment style in one file.

kingluo:
FYI, test H2 trailer headers in go:

```go
// trailers.go
package main

import (
        "io"
        "log"
        "net/http"
)

func main() {
        mux := http.NewServeMux()
        mux.HandleFunc("/trailers", func(w http.ResponseWriter, req *http.Request) {
                w.Header().Set("Trailer", "foobar")
                w.Header().Set("Content-Type", "text/html")
                w.Header().Set("Server", "Deproxy Server")
                w.WriteHeader(http.StatusOK)
                io.WriteString(w, "GHIJKLMNOPQRSTUV")
                w.Header()["Trailer:foobar"] = []string{"foobar"}
        })
        log.Fatal(http.ListenAndServe(":8080", mux))
}
```

```bash
# run demo backend
go run trailers.go

# check trailer headers
curl --raw -k --http2 -v \
    --resolve "server.test.com:443:127.0.0.1" https://server.test.com/trailers
```

```
< HTTP/2 200
< content-type: text/html
< trailer: foobar
< date: Wed, 10 Apr 2024 10:12:52 GMT
< via: 2.0 tempesta_fw (Tempesta FW 0.8.0)
< content-length: 20
< server: Tempesta FW/0.8.0
<
* TLSv1.2 (IN), TLS header, Supplemental data (23):
< foobar: foobar
* Connection #0 to host server.test.com left intact
GHIJKLMNOPQRSTUVWXYZ
```

fw/hpack.c
```diff
@@ -3486,6 +3486,52 @@ tfw_hpack_hdr_expand(TfwHttpResp *__restrict resp, TfwStr *__restrict hdr,
 	if (!hdr)
 		return -EINVAL;
 
+	if (hdr->flags & TFW_STR_TRAILER) {
```
EvgeniiMekhanik, line 3489: I think we have `tfw_http_hdr_split` instead of all code below.

fw/tempesta_fw.h
```diff
@@ -38,6 +38,10 @@
 
 #define DEF_MAX_PORTS		8
 
+enum {
+	SS_F_HTTT2_FRAME_TRAILER_HEADERS = 0x40,
```
EvgeniiMekhanik, line 42: All skb flags are located in skbuff.h

kingluo: yes, but that involves the linux-5.10.35-tfw repo, so I place it here temporarily as a placeholder.

fw/hpack.c
kingluo: okay, I'll have a check.

fw/http.c
```diff
@@ -4720,6 +4720,44 @@ tfw_h2_hdr_sub(unsigned short hid, const TfwStr *hdr, const TfwHdrMods *h_mods)
...
+tfw_h2_hpack_encode_trailer_headers(TfwHttpResp *resp, const TfwHdrMods *h_mods,
+			    unsigned int stream_id)
+{
+	int r;
+	unsigned int i;
+	TfwHttpTransIter *mit = &resp->mit;
+	TfwHttpHdrMap *map = mit->map;
+	TfwHttpHdrTbl *ht = resp->h_tbl;
+
+	for (i = 0; i < map->count; ++i) {
```
EvgeniiMekhanik, line 4733: It seems that its better to iterate `for (i = TFW_HTTP_HDR_RAW; i < ht->off; ++i)`

fw/http.c
```diff
@@ -5347,6 +5388,39 @@ tfw_h2_resp_adjust_fwd(TfwHttpResp *resp)
...
+		r = tfw_h2_hpack_encode_trailer_headers(resp, h_mods, stream_id);
+
+		resp->req->stream->xmit.t_len = resp->mit.acc_len - acc;
+		nskb_head = resp->msg.skb_head;
+		resp->msg.skb_head = resp->mit.iter.skb = skb;
+
+		if (unlikely(r))
+			goto clean;
+
+		nskb = nskb_head;
```
EvgeniiMekhanik, line 5413: We should set appropriate skb flags in __tfw_hpack_encode, not implement additional loop here.

fw/http.c
```diff
@@ -5347,6 +5388,39 @@ tfw_h2_resp_adjust_fwd(TfwHttpResp *resp)
 	if (unlikely(r))
 		goto clean;
 
+	if (resp->trailers_len > 0) {
+		// MUST _NOT_ use resp pool to encode trailer headers
```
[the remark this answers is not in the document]
kingluo: as I said in the comment, the pool-related functions assume the body must be after the headers and no trailer headers, so when the body is written to the skb, you have no chance to append new headers between the original headers (be fixed in length then) and the body anymore, if you do, it will overwrite the data between them and cause memory violation, which I have tested and verified. Of course, we can rewrite those functions, but the cost is very high, so why not use simple wrapper code to tackle it?

fw/http.c
kingluo: No, here we no only need to handle the flags but also the skb list (for trailer headers). `__tfw_hpack_encode` has no information here. Since the new skb list is wrapped on the original code path, so we need to iterate all skbs and attach them to the message's skb head.

fw/tempesta_fw.h
EvgeniiMekhanik: You should make appropriate changes in  linux-5.10.35-tfw and change linux-5.10.35.patch according them

fw/http_frame.c
```diff
@@ -2123,6 +2126,11 @@ tfw_h2_make_frames(struct sock *sk, struct sk_buff *skb, TfwH2Ctx *ctx,
 	TfwFrameHdr frame_hdr = {};
 	unsigned long *len = (type == HTTP2_DATA ?
 		&stream->xmit.b_len : &stream->xmit.h_len);
+	TfwFrameType otype = type;
+	if (type == HTTP2_HEADERS && stream->xmit.b_len == 0 && stream->xmit.h_len == 0) {
```
EvgeniiMekhanik, line 2130: Instead of this part of code you should implement `tfw_h2_make_trailer_frames` same as it done for header and data frames. (see tfw_h2_make_headers_frames` and `tfw_h2_make_data_frames`)

fw/http_frame.c
```diff
@@ -2146,7 +2154,7 @@ tfw_h2_make_frames(struct sock *sk, struct sk_buff *skb, TfwH2Ctx *ctx,
 	limit -= FRAME_HEADER_SIZE;
 
 	frame_hdr.stream_id = stream->id;
-	frame_hdr.type = type;
+	frame_hdr.type = type == HTTP2_TRAILER_HEADERS ? otype : type;
```
EvgeniiMekhanik, line 2157: In case if you implement `tfw_h2_make_trailer_frames` toy don't need HTTP2_TRAILER_HEADERS and this part of code

fw/http_stream.c
```diff
@@ -692,6 +692,12 @@ tfw_h2_stream_send_process(TfwH2Ctx *ctx, TfwStream *stream, unsigned char type)
 	if (!stream->xmit.b_len)
```
EvgeniiMekhanik, line 692: You should check that there is no stream->xmit.t_len == 0 here also

fw/sock_clnt.c
```diff
@@ -342,12 +346,42 @@ tfw_h2_sk_prepare_xmit(struct sock *sk, struct sk_buff *skb,
 		 * We clear this flag to prevent it's copying
 		 * during skb splitting.
 		 */
-		if (!stream->xmit.b_len) {
+		if (!stream->xmit.t_len && !stream->xmit.b_len) {
```
EvgeniiMekhanik, line 349: Why? If body is finished we should process stream and clear SS_F_HTTT2_FRAME_DATA. It is not depend on trailers

fw/sock_clnt.c
```diff
@@ -342,12 +346,42 @@ tfw_h2_sk_prepare_xmit(struct sock *sk, struct sk_buff *skb,
...
+
+		/*
+		 * We clear this flag to prevent it's copying
+		 * during skb splitting.
+		 */
+		if (!stream->xmit.t_len) {
+			skb_clear_tfw_flag(skb, SS_F_HTTT2_FRAME_TRAILER_HEADERS);
+			if (!stream->xmit.b_len) {
+				skb_clear_tfw_flag(skb, SS_F_HTTT2_FRAME_DATA);
+				TFW_H2_STREAM_SEND_PROCESS(h2, stream, HTTP2_TRAILER_HEADERS);
```
EvgeniiMekhanik, line 380: Here we should process sending headers frames (because trailer frames ia s header frames)

fw/http.c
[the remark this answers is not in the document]
EvgeniiMekhanik: @krizhanovsky what do you think?

fw/cache.c
```diff
@@ -2824,6 +2857,20 @@ tfw_cache_build_resp(TfwHttpReq *req, TfwCacheEntry *ce, long lifetime,
 	}
 	resp->content_length = ce->body_len;
 
+	if (has_trailers) {
```
EvgeniiMekhanik, line 2860: As I see most of this code is necessary only because we don't use pool for making trailer headers. I am right? @krizhanovsky how do you think may be we should rework pool related functions and use them for making trailers?

EvgeniiMekhanik (changes requested):
Need fixes

fw/sock_clnt.c
kingluo: No, here it's used to avoid stream state machine processing, which assumes DATA is the last step, but if there is trailers, the last step is trailers, so it should be postponed.

fw/sock_clnt.c
```diff
@@ -342,12 +346,42 @@ tfw_h2_sk_prepare_xmit(struct sock *sk, struct sk_buff *skb,
...
 	}
 
+	if (FRAME_TRAILER_SHOULD_BE_MADE(flags)) {
+		r = tfw_h2_make_headers_frames(sk, skb, h2, stream, mss_now,
+					    *limit - stream->xmit.processed,
+					    &tmp_truesize);
+		if (unlikely(r)) {
+			T_WARN("%s: failed to make data frames (%d)",
+			       __func__, r);
+			if (r == -ENOMEM && headers_was_done) {
```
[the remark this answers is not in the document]
kingluo: Not necessary, because the skbs of trailers are separated from the DATA, so only after the skbs of HEADERS and DATA have been processed, it starts to do this part.

fw/sock_clnt.c
kingluo: yes, my mistake.
but it should be `TFW_H2_STREAM_SEND_PROCESS(h2, stream, HTTP2_DATA);`, because it just pretends no trailers and the DATA has been done.

fw/cache.c
kingluo: There are two paths to generate responses.
1. backend --> client, this path uses a pool
2. cache --> client, this path has no pool, but just bundles all headers into a skb list

both paths have different logic and functions to construct responses, and both assume no trailers, moreover, in the prepare_xmit hook, it also assumes such layout of the skb list.
I checked the code of both paths before, and I think it's very very complicated to do refactoring (too many details are involved), I think it is not worthing to pay such a big cost for a relatively trial RFC fix. To be honest, even regarding my current hacking way, it took me a week.

fw/tempesta_fw.h
kingluo: yes, I know.

fw/sock_clnt.c
EvgeniiMekhanik: Stream state machine should assumes that DATA s not a last step. We should rework it to  make it possible to handle trailers as I think (when we receive trailers it is already done). @krizhanovsky what do you think?

fw/hpack.c
kingluo: `tfw_http_hdr_split` did a different thing. It just splits the chunks of a TfwStr into two parts and excludes `: `, but my code merges sparse chunks into two chunks, which meets the requirement of `tfw_hpack_hdr_expand`. So it cannot be replaced by `tfw_http_hdr_split`. It's tested.

fw/http.c
kingluo: No, it will cause a dead loop!
This code snippet is copied from [`tfw_h2_hpack_encode_headers`](https://github.com/tempesta-tech/tempesta/blob/9bc155571c17c69b0bda9f09f9c2f501c5c0d24d/fw/http.c#L4771), the only modification is for trailers.

fw/http_frame.c
kingluo: No, `tfw_h2_make_headers_frames` and `tfw_h2_make_data_frames` do nothing but provide `type` to `tfw_h2_make_frames`. However, here the `HTTP2_TRAILER_HEADERS` is actually a subclass of `HTTP2_HEADERS`, that is, it needs to be treated as header in that function too except some trivial changes specific to trailers. I just deliberately do this to reuse maximum code.

fw/http_stream.c
kingluo: if there are trailers, then the state machine input of `HTTP2_DATA` will be replaced by `HTTP2_TRAILER_HEADERS`, that is, this branch is not executable, instead, it will execute the following branch, and the latter one is the superset of the previous one.

fw/sock_clnt.c
kingluo: sorry, I checked again, it's not a mistake.
see https://github.com/tempesta-tech/tempesta/pull/2095#discussion_r1577524887

fw/cache.c
kingluo: @EvgeniiMekhanik @krizhanovsky @const-t 
If I understand correctly after checking the code `__tfw_http_msg_expand_from_pool`, the so-called "pool" is just used to linearize the regular headers and the response body. That is, if use pool, the memory contains headers and body is linear, otherwise, it's scattered into multiple skbs. But that's okay for trailers, and more precisely, it's better for trailers:
1. the trailers are encoded separately from regular headers and body
2. the trailers' skbs have different flags which must not be mixed with regular headers and body, in the latter case, small headers list and body may share one skb, if mixed with trailers too, that's complicated.

So, IMO, KISS (keep it simple and stupid) is better.

krizhanovsky:
@kingluo pelase don't forget to squash at least several commits (like printk removal) before the merge

fw/http_parser.c
krizhanovsky: > No, for HTTP/1, the trailer chars are always `\r\n`, so the `len` is always `2`. I add some comment.

What is we actually receive `\n` only? It's prohibited by the RFC, but we saw clients and maybe servers, who actually send `\n` only instead of `\r\n`. Note also that a "server" could be just a some very home brew microservice.

fw/hpack.c
krizhanovsky: @kingluo I didn't check the code yet, but typically if a reviewer has a question on code, then any other reader is likely have the same question, so it worth commenting the code. In this case it'd be good to add a comment like you wrote above

kingluo: okay, done.

fw/http_parser.c
const-t: It fixed in a wrong way, `SLEN(S_CRLF)` always 2 bytes, but in some circumstances we can have only `\n`, this is allowed by our parser. You need to use `tfw_str_eolen()`. However you can't do it, in `process_trailer_hdr()`, because at this stage header is not fully parsed. I suggest to move this part to `tfw_http_msg_hdr_close()`.

fw/cache.c
```diff
@@ -2737,12 +2744,38 @@ tfw_cache_build_resp(TfwHttpReq *req, TfwCacheEntry *ce, long lifetime,
...
+	nmit.start_off = 0;
+	nmit.iter.skb = nmit.iter.skb_head = NULL;
+	nmit.iter.frag = -1;
+
 	for (h = TFW_HTTP_HDR_REGULAR; h < ce->hdr_num; ++h) {
 		bool skip = !TFW_MSG_H2(req) && (h >= ce->hdr_h2_off);
+		bool is_trailer_hdr = ((TfwCStr *)p)->flags & TFW_CSTR_TRAILER;
+		int r;
+
+		if (is_trailer_hdr && TFW_MSG_H2(req)) {
```
const-t, line 2758: Looks hacky for me. I would prefer to change cache entry layout this way:
```
 Cache entry {
 [HTTP1/HTTP2 headers]
 [HTT2 only HEADERS]
 [DATA]
 [Trailer headers]
 } 
```

Count of ` [HTTP1/HTTP2 headers]` +  `[HTT2 only HEADERS]` stored in `TfwCacheEntry::hdr_num`, for `[Trailer headers]` we can introduce new filed in `TfwCacheEntry`. To do this we can break `TfwCacheEntry::hdr_num` into two unsigned shorts. `unsigned short TfwCacheEntry::hdr_num` and `unsigned short `TfwCacheEntry::trailer_hdr_num`. Before building the body we iterate over headers until `TfwCacheEntry::hdr_num`. After building the body you need just iterate over the `TfwCacheEntry::trailer_hdr_num`.

While writing data to cache, we can try to adopt `TfwHttpTransIter::map`, for working
with trailers. For instance just add offset where trailer headers starts(Seems this should work, because trailers always at the end).

const-t (changes requested):
We must postpone this issue until #1973 will be merged. In this patch flags for skbs will be eliminated and HTTP2 framing code refactored as well. There is no sense to fix issue with trailers now, and then rework it after #1973.

Another one, after #1973 non-cache path MUST use pool for trailers transformation, there is no any reasons why not. Framing of trailers must be done in framing state machine. Cache path must be implemented in another way as well(Commented above).

kingluo:
Which case uses only "\n"?



| |
***@***.***
|
|
***@***.***
|




---- Replied Message ----
| From | ***@***.***> |
| Date | 04/30/2024 22:33 |
| To | ***@***.***> |
| Cc | jinhua ***@***.***>***@***.***> |
| Subject | Re: [tempesta-tech/tempesta] fix(1902): correct the skb len (PR #2095) |

@const-t commented on this pull request.

In fw/http_parser.c:

> @@ -973,6 +973,8 @@ process_trailer_hdr(TfwHttpMsg *hm, TfwStr *hdr, unsigned int id)
 	hdr->flags |= TFW_STR_TRAILER;
 	__set_bit(TFW_HTTP_B_CHUNKED_TRAILER, hm->flags);
 
+	hm->trailers_len += tfw_str_len(hdr) + 2;


It fixed in a wrong way, SLEN(S_CRLF) always 2 bytes, but in some circumstances we can have only \n, this is allowed by our parser. You need to use tfw_str_eolen(). However you can't do it, in process_trailer_hdr(), because at this stage header is not fully parsed. I suggest to move this part to tfw_http_msg_hdr_close().

kingluo:
Yes, I know, we can do it in normal way. But as I said, it takes much more work to do because it's more complicated than the hacking way. It also involves more parts of the framework, e.g. currently the parser hardcore the hdr table.



| |
***@***.***
|
|
***@***.***
|




---- Replied Message ----
| From | ***@***.***> |
| Date | 04/30/2024 23:12 |
| To | ***@***.***> |
| Cc | jinhua ***@***.***>***@***.***> |
| Subject | Re: [tempesta-tech/tempesta] fix(1902): correct the skb len (PR #2095) |

@const-t requested changes on this pull request.

We must postpone this issue until #1973 will be merged. In this patch flags for skbs will be eliminated and HTTP2 framing code refactored as well. There is no sense to fix issue with trailers now, and then rework it after #1973.

Another one, after #1973 non-cache path MUST use pool for trailers transformation, there is no any reasons why not. Framing of trailers must be done in framing state machine. Cache path must be implemented in another way as well(Commented above).

In fw/cache.c:

>  	for (h = TFW_HTTP_HDR_REGULAR; h < ce->hdr_num; ++h) {
 		bool skip = !TFW_MSG_H2(req) && (h >= ce->hdr_h2_off);
+		bool is_trailer_hdr = ((TfwCStr *)p)->flags & TFW_CSTR_TRAILER;
+		int r;
+
+		if (is_trailer_hdr && TFW_MSG_H2(req)) {


Looks hacky for me. I would prefer to change cache entry layout this way:

 Cache entry {
 [HTTP1/HTTP2 headers]
 [HTT2 only HEADERS]
 [DATA]
 [Trailer headers]
 } 


Count of [HTTP1/HTTP2 headers] + [HTT2 only HEADERS] stored in TfwCacheEntry::hdr_num, for [Trailer headers] we can introduce new filed in TfwCacheEntry. To do this we can break TfwCacheEntry::hdr_num into two unsigned shorts. unsigned short TfwCacheEntry::hdr_num and unsigned short TfwCacheEntry::trailer_hdr_num. Before building the body we iterate over headers until TfwCacheEntry::hdr_num. After building the body you need just iterate over the TfwCacheEntry::trailer_hdr_num`.

While writing data to cache, we can try to adopt TfwHttpTransIter::map, for working
with trailers. For instance just add offset where trailer headers starts(Seems this should work, because trailers always at the end).

fw/cache.c
kingluo: @EvgeniiMekhanik @const-t @krizhanovsky 
Yes, of course, I know what you're talking about, we can do it in a normal way, which was also my option.

But as I said before, in my assessment, this requires more work as it is more complex than the hacking way and is also error-prone as it involves more parts of the framework and I need to modify and add more code, such as the current parser hardcore hdr table.

Why did I choose the hacking way? Because it's KISS. I wonder if it's worth doing a massive refactor just for the trailer issue. Anyway, I won't reject any options, I will only code for your final decision.

So anyway, I think we need a decision here, rather than the tech detail.

krizhanovsky (changes requested):
This PR contradicts our nearest targets to solve the performance issues and implement #498, so I believe we need to solve in an appropriate way to encode the trailer headers in a trailing HTTP HEADERS frame, this should be done on the fly, w/o additional loops.

Also the PR leaves a problem, not clearly stated in #1902: if we move the trailer header to normal headers in response
```python3
:status: 200
content-type: text/html
last-modified: Wed, 14 Jun 2023 10:19:11 GMT
date: Wed, 14 Jun 2023 10:19:11 GMT
trailer: X-Token
x-token: value
via: 2.0 tempesta_fw (Tempesta FW pre-0.7.0)
content-length: 52
server: Tempesta FW/pre-0.7.0

abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ
```
then we also must remove `trailer: X-Token` or we confuse a client that we have a trailer, but we actually don't. In this context the PR doesn't solve this since it doesn't remove the header.

fw/hpack.c
```diff
@@ -3486,6 +3486,66 @@ tfw_hpack_hdr_expand(TfwHttpResp *__restrict resp, TfwStr *__restrict hdr,
...
+	// ":"
+	// " "
+	// "b"
+	// "a"
+	// into
+	// "fo"
+	// ": "
+	// "ba"
+	// it's different from `tfw_http_hdr_split()`, which only splits
+	// name chunks and value chunks into two strs, no compact.
```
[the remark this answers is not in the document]
kingluo: yes, I looked over the coding style. Here I just emphasize each line is an item of the array, if list them in one line, it's not so obvious.

kingluo:
> This PR contradicts our nearest targets to solve the performance issues and implement #498, so I believe we need to solve in an appropriate way to encode the trailer headers in a trailing HTTP HEADERS frame, this should be done on the fly, w/o additional loops.
> 
> Also the PR leaves a problem, not clearly stated in #1902: if we move the trailer header to normal headers in response
> 
> ```python
> :status: 200
> content-type: text/html
> last-modified: Wed, 14 Jun 2023 10:19:11 GMT
> date: Wed, 14 Jun 2023 10:19:11 GMT
> trailer: X-Token
> x-token: value
> via: 2.0 tempesta_fw (Tempesta FW pre-0.7.0)
> content-length: 52
> server: Tempesta FW/pre-0.7.0
> 
> abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ
> ```
> 
> then we also must remove `trailer: X-Token` or we confuse a client that we have a trailer, but we actually don't. In this context the PR doesn't solve this since it doesn't remove the header.

Sorry, I don't get it. The PR is used to avoid merging the trailers into regular headers, so `trailer: X-Token` must be at the regular headers to indicate there is a trailer.

fw/cache.c
```diff
@@ -2824,6 +2857,20 @@ tfw_cache_build_resp(TfwHttpReq *req, TfwCacheEntry *ce, long lifetime,
...
+		resp->req->stream->xmit.t_len = t_len;
+		skb_set_tfw_flags(nskb, SS_F_HTTP2_FRAME_START);
+		do {
+			skb_set_tfw_cb(nskb, stream_id);
+			skb_set_tfw_flags(nskb, SS_F_HTTT2_FRAME_TRAILER_HEADERS);
+			skb = nskb->next;
+			nskb->next = nskb->prev = NULL;
+			ss_skb_queue_tail(&resp->msg.skb_head, nskb);
+			nskb = skb;
+		} while (nskb != nskb_head);
```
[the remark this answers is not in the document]
kingluo: yes, I agree we should handle it in a normal way instead of hacking. As I said, that was one of my two options when I was working on it.
but I need to emphasize that under the current code framework, this hacking way is already the most efficient way to achieve the goal. **The extra conditions exist because the framework requires to loop over the whole header list including trailers so that it gets the correct pointer to the body, if you remove the inner conditions, then you need two loops, which is worse because each loop needs to iterate the whole header list (the cache entry has one header list, which didn't distinguish the regular headers and trailers).**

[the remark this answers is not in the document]
kingluo: Let me repeat my opinion. The hacking way works, no matter what disadvantage you said (I also admit). But why did I choose it? It's just one week of development work, but for the normal way, maybe several weeks and more extra time to ensure it does not bring in new bugs because the amount of code will be much larger and involves more aspects, which is normally more error-prone. Not to mention, as @const-t said, it depends on https://github.com/tempesta-tech/tempesta/pull/1973 too, I need to cooperate with that PR including the framing redesign.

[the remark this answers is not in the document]
kingluo: All in all, shall I start the refactoring later? or after https://github.com/tempesta-tech/tempesta/pull/1973?

biathlon3 (approved):
Looks good enough.

EvgeniiMekhanik (approved):
LGTM, but please move `trailers_len` to response as @const-t  described

fw/http_msg.c
```diff
@@ -565,6 +565,13 @@ tfw_http_msg_hdr_close(TfwHttpMsg *hm)
 	/* Close just parsed header. */
 	parser->hdr.flags |= TFW_STR_COMPLETE;
 
+	/* Cumulate the trailer headers length */
+	if (parser->hdr.flags & TFW_STR_TRAILER) {
+		TfwHttpResp* resp = (TfwHttpResp*) hm;
```
const-t, line 570: Usually we add blank line after declaration.

const-t (approved):
LGTM

krizhanovsky (approved):
LGTM

PR #2097, opened 2024-04-08
Drop request sock in case when frang block new connection

### LINKED ISSUE #2096: Extra calls of `tfw_classify_conn_estab`
There is a problem with extra calls of `tfw_classify_conn_estab` in case when this function return T_BLOCK. To understand this problem look at the tcp handshake: client sends SYN (and switched to TCP_SYN_SENT state) to server, after receiving SYN server (which is in TCP_LISTEN state) sends SYN_ACK to client. After receiving SYN_ACK client send last ACK to server and switched to ESTABLISHED state. In this state client can send data to server. When server receives last ACK in `tcp_v4_rcv` sk = __inet_lookup_skb(&tcp_hashinfo, skb, __tcp_hdrlen(th), th->source, th->dest, sdif, &refcounted); is returned in TCP_NEW_SYN_RECV state (it is previously created request sock), then if tcp_check_req->syn_recv_sock->tempesta_new_clntsk fails we send TCP_RST to client. But if client send some just after sending last ACK, we again try to establish connection tcp_check_req->syn_recv_sock->tempesta_new_clntsk. Also we can cat again try to establish connection because client can try to retransmit last ACK of TCP handshake before he receive our RST. Eventually client will be blocked, but count of calls of `tfw_classify_conn_estab` can be  (1 - 3).

EvgeniiMekhanik:
To understand this problem look at the tcp handshake: client sends SYN (and switched to TCP_SYN_SENT state) to server, after receiving SYN server (which is in TCP_LISTEN state) sends SYN_ACK to client. After
receiving SYN_ACK client send last ACK to server and switched to ESTABLISHED state. In this state client can send data to server. When server receives last ACK it use previously created request sock, and call tcp_check_req->syn_recv_sock->tempesta_new_clntsk. If we don't drop request sock after `tempesta_new_clntsk` fails, this request sock can be used later if client sends some data just after sending last ACK, or if client tries to retransmit last ACK.

Closes #2096

### DIFF
--- a/linux-5.10.35.patch
+++ b/linux-5.10.35.patch
@@ -790,6 +790,23 @@ index 000000000..80de50693
 +
 +#endif /* __TEMPESTA_H__ */
 +
+diff --git a/include/net/inet_sock.h b/include/net/inet_sock.h
+index 89163ef8c..49ad1ddc9 100644
+--- a/include/net/inet_sock.h
++++ b/include/net/inet_sock.h
+@@ -87,7 +87,12 @@ struct inet_request_sock {
+ 				ecn_ok	   : 1,
+ 				acked	   : 1,
+ 				no_srccheck: 1,
++#ifdef CONFIG_SECURITY_TEMPESTA
++				smc_ok	   : 1,
++				aborted	   : 1;
++#else
+ 				smc_ok	   : 1;
++#endif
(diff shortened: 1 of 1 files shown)


krizhanovsky (approved):
LGTM

Issue #1775, opened 2022-12-26
empty last DATA frame for h2 request is not handled as expected


RomanBelozerov:
make correct h2 request with empty last data frame:

AR: client does not receive response.
ER: client receive 200 response.

RFC 9113 6.9.1.

[Test to reproduce.](https://github.com/tempesta-tech/tempesta-test/blob/79876a950547dff367052de1c4c8378fafb57693/http2_general/test_h2_frame.py#L65)

Issue #1175, opened 2019-02-01
Race condition during stopping of server connection.


aleksostapenko:
The crash occurred during long tests on 9011 server (dump `201812220406`):
```
[133672.417707] [tempesta] Warning: request dropped: processing error, status 502: 192.168.0.2
[133672.417719] [tempesta] Warning: request dropped: processing error, status 502: 192.168.0.2
[133672.417723] [tempesta] Warning: request dropped: processing error, status 502: 192.168.0.2
[133672.417727] [tempesta] Warning: request dropped: processing error, status 502: 192.168.0.2
[133672.417728] [tempesta] Warning: request dropped: processing error, status 502: 192.168.0.2
[133672.417731] [tempesta] Warning: request dropped: processing error, status 502: 192.168.0.2
[133672.417745] [tempesta] Warning: request dropped: processing error, status 502: 192.168.0.2
[133672.417746] [tempesta] Warning: request dropped: processing error, status 502: 192.168.0.2
[133672.417749] [tempesta] Warning: request dropped: processing error, status 502: 192.168.0.2
[133676.171920] ------------[ cut here ]------------
[133676.176810] kernel BUG at /root/builds/tempesta/source/tempesta_fw/connection.h:449!
[133676.184846] invalid opcode: 0000 [#1] SMP PTI
[133676.189501] Modules linked in: tempesta_fw(O) tempesta_db(O) tempesta_tls(O) tempesta_lib(O) iptable_mangle xt_mark tcp_diag inet_diag fuse 8021q garp mrp stp mlx5_core llc mlx4_ib ib_core mlx4_en intel_rapl x86_pkg_temp_thermal intel_powerclamp coretemp kvm_intel kvm ast irqbypass ttm crct10dif_pclmul crc32_pclmul drm_kms_helper iTCO_wdt ghash_clmulni_intel iTCO_vendor_support mei_me intel_cstate intel_uncore igb ipmi_si drm dca pcspkr joydev intel_rapl_perf i2c_algo_bit mei i2c_i801 intel_pch_thermal ipmi_devintf sg ie31200_edac ipmi_msghandler video button mlx4_core devlink ptp pps_core ip_tables x_tables autofs4 ext4 crc16 mbcache jbd2 crc32c_generic fscrypto ecb sd_mod evdev hid_generic usbhid hid crc32c_intel mpt3sas xhci_pci raid_class xhci_hcd scsi_transport_sas aesni_intel aes_x86_64 crypto_simd
[...]

Issue #1881, opened 2023-05-23
trailers in h2 request


RomanBelozerov:
# Scope
[RFC 9113 8.1](https://www.rfc-editor.org/rfc/rfc9113.html#name-http-message-framing)

Cases:
> Trailers MUST NOT include pseudo-header fields (Section 8.3). An endpoint that receives pseudo-header fields in trailers MUST treat the request or response as malformed (Section 8.1.1).

**Fixed**:
> ~~An endpoint that receives a HEADERS frame without the END_STREAM flag set after receiving the HEADERS frame that opens a request or after receiving a final (non-informational) status code MUST treat the corresponding request or response as malformed (Section 8.1.1).~~

> ~~Trailer fields are carried in a field block that also terminates the stream. That is, trailer fields comprise a sequence starting with a HEADERS frame, followed by zero or more CONTINUATION frames, where the HEADERS frame bears an END_STREAM flag.~~

# Testing
Tests to reproduce;
- http2_general.test_h2_headers.TestTrailers.test_trailers_with_pseudo_headers_in_request
- ~~http2_general.test_h2_headers.TestTrailers.test_trailers_without_end_stream_in_request~~
- ~~http2_general.test_h2_headers.TestTrailers.test_trailers_with_continuation_frame_in_request~~

krizhanovsky:
Fixed in https://github.com/tempesta-tech/tempesta/pull/2125

PR #2103, opened 2024-04-12
Fix HTTP2 empty frames processing

const-t:
The main goal of current commit, it's prevent flooding by empty(zero payload length) frames. To achieve this was fixed handling of following empty frames: DATA, HEADERS, CONTINUATION.

All of these headers can be empty, but with exceptions:
 - Only first HEADERS frame can be empty(RFC doesn't have any direct prohibitions) and must be followed by CONTINUATION frame.
 - DATA frame MUST have END_STREAM flag.
 - CONTINUATION frame MUST have END_HEADERS.

Other empty frames are prohibited, because they are not described in protocol specification and there is not proper use cases. Handling of such frames leads to unreasonable CPU utilization thus potentially DOS.

Useage of allowed empty frames is described by protocol specification and receiving or sending them multiple times is prohibited by protocol. Thus we don't need any additional tracking.

Also added IDLE state to streams state machine. We should do it to follow specification and handle
streams states correctly.

### DIFF
--- a/fw/http_frame.c
+++ b/fw/http_frame.c
@@ -147,7 +147,7 @@ do {									\
  */
 #define FRAME_FSM_READ_PYLD()						\
 do {									\
-	WARN_ON_ONCE(ctx->rlen >= ctx->to_read);			\
+	WARN_ON_ONCE(ctx->rlen > ctx->to_read);				\
 	n = min_t(int, ctx->to_read - ctx->rlen, buf + len - p);	\
 	p += n;								\
 	ctx->rlen += n;							\
@@ -1308,7 +1308,15 @@ do {									\
 		if (hdr->flags & HTTP2_F_PADDED)
 			return tfw_h2_recv_padded(ctx);
 
-		SET_TO_READ_VERIFY(ctx, HTTP2_RECV_DATA);
+		/* TODO: #1196 Rework this part. */
+		if (tfw_h2_get_stream_state(ctx->cur_stream) >=
+		    HTTP2_STREAM_LOC_CLOSED)
+			ctx->state = HTTP2_IGNORE_FRAME_DATA;
+		else
+			ctx->state = HTTP2_RECV_DATA;
+
+		SET_TO_READ(ctx);
+
 		return 0;
 
 	case HTTP2_HEADERS:
@@ -1344,7 +1352,17 @@ do {									\
 		if (hdr->flags & HTTP2_F_PRIORITY)
 			return tfw_h2_recv_priority(ctx);
 
-		SET_TO_READ_VERIFY(ctx, HTTP2_RECV_HEADER);
+		/* TODO: #1196 Rework this part. */
+		if (ctx->cur_stream &&
+		    tfw_h2_get_stream_state(ctx->cur_stream) >=
+		    HTTP2_STREAM_LOC_CLOSED)
+		{
+			ctx->state = HTTP2_IGNORE_FRAME_DATA;
+		} else {
+			ctx->state = HTTP2_RECV_HEADER;
+		}
+
+		SET_TO_READ(ctx);
 		return 0;
 
 	case HTTP2_PRIORITY:
@@ -1359,10 +1377,11 @@ do {									\
 		if (hdr->length != FRAME_PRIORITY_SIZE)
 			goto conn_term;
 
-		if (ctx->cur_stream) {
-			STREAM_RECV_PROCESS(ctx, hdr);
-			ctx->state = HTTP2_RECV_FRAME_PRIORITY;
-		} else {
+		/* TODO: #1196 Rework this part. */
+		if (!ctx->cur_stream ||
+		    tfw_h2_get_stream_state(ctx->cur_stream) >=
+		    HTTP2_STREAM_LOC_CLOSED)
+		{
 			/*
 			 * According to RFC 9113 section 5.1:
 			 * PRIORITY frames are allowed in the `closed` state,
@@ -1370,7 +1389,11 @@ do {									\
 			 * already removed from memory, just ignore this frame.
 			 */
 			ctx->state = HTTP2_IGNORE_FRAME_DATA;
+		} else {
+			STREAM_RECV_PROCESS(ctx, hdr);
+			ctx->state = HTTP2_RECV_FRAME_PRIORITY;
 		}
+
 		SET_TO_READ(ctx);
 		return 0;
 
@@ -1540,7 +1563,8 @@ do {									\
 		ctx->data_off = FRAME_HEADER_SIZE;
 		ctx->plen = ctx->hdr.length;
 
-		SET_TO_READ_VERIFY(ctx, HTTP2_RECV_CONT);
+		SET_TO_READ(ctx);
+		ctx->state = HTTP2_RECV_CONT;
 		return 0;
 
 	default:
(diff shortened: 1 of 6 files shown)


krizhanovsky:
Does this fix #1775 and several points from #1881?

const-t:
> Does this fix #1775 and several points from #1881?

Yes, this fixes #1175, because now we handling empty frames and this fixed following points from #1881

http2_general.test_h2_headers.TestTrailers.test_trailers_without_end_stream_in_request
http2_general.test_h2_headers.TestTrailers.test_trailers_with_continuation_frame_in_request

fw/http_stream.c
```diff
@@ -355,6 +355,65 @@ do {									\
...
+
+		if (type == HTTP2_HEADERS) {
+			switch (flags
+				& (HTTP2_F_END_HEADERS | HTTP2_F_END_STREAM))
+			{
+			case HTTP2_F_END_HEADERS | HTTP2_F_END_STREAM:
+				SET_STATE(HTTP2_STREAM_REM_HALF_CLOSED);
+				break;
+			case HTTP2_F_END_HEADERS:
+				ctx->cur_recv_headers = NULL;
```
EvgeniiMekhanik, line 384: This is unnecessary, because cur_recv_headers is always equal to zero here.

fw/http_frame.c
```diff
@@ -1308,7 +1307,8 @@ do {									\
 		if (hdr->flags & HTTP2_F_PADDED)
 			return tfw_h2_recv_padded(ctx);
 
-		SET_TO_READ_VERIFY(ctx, HTTP2_RECV_DATA);
+		SET_TO_READ(ctx);
```
EvgeniiMekhanik, line 1310: We need to ignore frames for LOC_CLOSED state

fw/http_stream.c
```diff
@@ -409,34 +477,85 @@ do {									\
 						SET_STATE(new_state);
 					}
 				}
+			} else {
```
EvgeniiMekhanik, line 480: You do not process a case when we receive EN_OF_STREAM without end of headers here. Please also add appropriate test for this case

fw/http_stream.c
```diff
@@ -409,34 +477,85 @@ do {									\
...
+					ctx->cur_send_headers = stream;
+				else
+					ctx->cur_recv_headers = stream;
+			}
+			break;
+		} else if (type == HTTP2_HEADERS) {
+			TfwStream *snd_hdrs = send ? ctx->cur_send_headers
+						   : ctx->cur_recv_headers;
+
+			if (snd_hdrs) {
```
EvgeniiMekhanik, line 491: We use TFW_H2_FSM_TYPE_CHECK early. Is it not enough?

fw/http_stream.c
const-t: This case is handled in parser, because it's valid case for trailers. Moving it here doesn't have sense.

fw/http_frame.c
```diff
@@ -1951,7 +1994,37 @@ tfw_h2_frame_process(TfwConn *c, struct sk_buff *skb, struct sk_buff **next)
 			kfree_skb(nskb);
 			goto out;
 		}
-	} else {
+	}
+	else if (unlikely(tfw_h2_allowed_empty_frame(h2))) {
```
EvgeniiMekhanik, line 1998: I have a question here, i am not sure. Can we use previous path of code for our purpose? and pass NULL skb in tfw_http_msg_process_generic, when we deal with empty data/header/continuation frame?

fw/http_frame.c
```diff
@@ -1951,7 +1994,37 @@ tfw_h2_frame_process(TfwConn *c, struct sk_buff *skb, struct sk_buff **next)
...
+		/*
+		 * Process empty frames.
+		 */
+		struct sk_buff *pskb, *end = h2->skb_head->prev;
+
+		/*
+		 * Free all SKBs in queue except the last. The last one
+		 * will be passed to message processing function.
+		 */
+		while (unlikely(h2->skb_head != end)) {
```
EvgeniiMekhanik, line 2008: I am not sure here. I think that the loop from previous path of code should be here. Can we have some skbs which are not relate to this stream here? Is here a chance that we drop them?

fw/http_stream.c
EvgeniiMekhanik: I wrote test and check error.

fw/http_frame.c
```diff
@@ -1630,8 +1632,20 @@ tfw_h2_frame_recv(void *data, unsigned char *buf, unsigned int len,
...
-		if (ctx->to_read)
+		if (ctx->to_read) {
 			FRAME_FSM_NEXT();
+		} else if (ctx->state != HTTP2_IGNORE_FRAME_DATA &&
+			   (ctx->hdr.type == HTTP2_HEADERS ||
+			    ctx->hdr.type == HTTP2_CONTINUATION ||
+			    ctx->hdr.type == HTTP2_DATA))
+		{
+			/*
+			 * HEADERS, CONTINUATION and DATA is allowed to have
```
EvgeniiMekhanik, line 1643: are allowed

fw/http_stream.c
EvgeniiMekhanik: We also need check case when client send us two or more END_STREAM flags, and return error

fw/http_frame.c
const-t: Good point.

fw/http_frame.c
const-t: No, if we have skbs not related to current stream during processing frame it's protocol violation.

fw/http_stream.c
const-t: Oh I see, you mean CONTINUATION frames. By the protocol [CONTINUATION](https://www.rfc-editor.org/rfc/rfc9113.html#name-continuation) frame doesn't have END_STREAM flag. In previous versions of Tempesta we handled such frames correctly, but it's incorrect we must treat such cases as protocol error.

fw/http_stream.c
```diff
@@ -409,34 +477,85 @@ do {									\
...
-			default:
+				SET_STATE(new_state);
+			}
+			else if (flags & HTTP2_F_END_STREAM) {
+				/* Next exptected CONTINUATION in trailers. */
 				if (send) {
 					ctx->cur_send_headers = stream;
+					new_state =
+						HTTP2_STREAM_LOC_HALF_CLOSED;
+					SET_STATE(new_state);
```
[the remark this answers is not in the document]
const-t: Yes, for consistency would be better to change it and I'll do it. But now it won't work, Tempesta can't send trailers for h2.

fw/http_stream.c
const-t: I believe we can rely on `TFW_H2_FSM_TYPE_CHECK` in this case.

fw/http_frame.c
```diff
@@ -147,7 +147,6 @@ do {									\
  */
 #define FRAME_FSM_READ_PYLD()						\
 do {									\
-	WARN_ON_ONCE(ctx->rlen >= ctx->to_read);			\
```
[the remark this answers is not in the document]
const-t: Yes, we should remove or edit. And I decided to edit this warn for catching invalid cases(It seems possible only if developer made a mistake), but not remove.

fw/http_frame.c
```diff
@@ -1359,18 +1375,22 @@ do {									\
 		if (hdr->length != FRAME_PRIORITY_SIZE)
 			goto conn_term;
 
-		if (ctx->cur_stream) {
-			STREAM_RECV_PROCESS(ctx, hdr);
-			ctx->state = HTTP2_RECV_FRAME_PRIORITY;
-		} else {
+		if (!ctx->cur_stream ||
```
EvgeniiMekhanik, line 1378: It will be changed in 1196, please add comment about it.

fw/http_frame.c
```diff
@@ -1344,7 +1351,16 @@ do {									\
 		if (hdr->flags & HTTP2_F_PRIORITY)
 			return tfw_h2_recv_priority(ctx);
 
-		SET_TO_READ_VERIFY(ctx, HTTP2_RECV_HEADER);
+		if (ctx->cur_stream &&
```
EvgeniiMekhanik, line 1354: It will be changed in 1196, please add comment about it.

fw/http_frame.c
```diff
@@ -1308,7 +1308,14 @@ do {									\
 		if (hdr->flags & HTTP2_F_PADDED)
 			return tfw_h2_recv_padded(ctx);
 
-		SET_TO_READ_VERIFY(ctx, HTTP2_RECV_DATA);
+		if (tfw_h2_get_stream_state(ctx->cur_stream) >=
```
EvgeniiMekhanik, line 1311: It will be changed in 1196, please add comment about it.

EvgeniiMekhanik (approved):

fw/http_frame.c
EvgeniiMekhanik: Or you can alloc skb with zero data ss_skb_alloc(0) in case when if (!h2->skb_head)  and tfw_h2_allowed_empty_frame

krizhanovsky (approved):
I don't have anything specific to comment on this PR, so LGTM

Issue #2098, opened 2024-04-08
[RESEARCH] High memory consumption after frang configuration


ykargin:
# Motivation

After frang configuration in [PR 598](https://github.com/tempesta-tech/tempesta-test/pull/598) tests started to fail with

```
 [ 6570.228871] ksoftirqd/0: page allocation failure: order:9, mode:0x40a20(GFP_ATOMIC|__GFP_COMP), nodemask=(null),cpuset=/,mems_allowed=0
 [ 6570.229960] CPU: 0 PID: 12 Comm: ksoftirqd/0 Tainted: G           OE     5.10.35.tfw-04d37a1 #1
 [ 6570.230476] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS rel-1.15.0-0-g2dd4b9b3f840-prebuilt.qemu.org 04/01/2014
 [ 6570.231459] Call Trace:
 [ 6570.231964]  dump_stack+0x74/0x92
 [ 6570.232458]  warn_alloc.cold+0x7b/0xdf
```
Research needed

Test to reproduce - `tls/test_tls_integrity.ManyClients` and `tls/test_tls_integrity.ManyClientsH2` with `-T 1` option  and body > 16KB (8GB memory)

krizhanovsky:
Looks like not enough memory.

RomanBelozerov:
I receive meamleak for these tests and tempesta commit - 10b38e071bad93e87d5fab6d0213246d3b1b5c84. I used remote setup (Tempesta on a separate VM) and cmd `./run_tests.py -T 1 tls/test_tls_integrity.ManyClientsH2` with MTU 80. So I run this test with 16KB, 64KB and 200KB body and I see the usage of all available memory (6GB for my VM for Tempesta) and meamleak after test ~1GB

look like it is fixed in #2105. I cannot get meamleak for this PR, but I see the usage of all available memory. I think Tempesta uses an unexpected lot of memory in these tests. 10 clients with 64KB response/request body, python uses ~400MB, but Tempesta ~5GB, why?

biathlon3:
Here we have the next situation for 64KB test.

In this test, Tempesta FW receives 65536 bytes of data request from 10 clients, routes them to a server, gets responds from a server and sends them to the clients.
With option -T 1, each request and respond are split by byte.
The key point is that if Tempesta FW receives only one byte, it uses a full skb (about 900 bytes).

[...]

PR #2105, opened 2024-04-19
Free postponed SKBs on connection close

const-t:
@TODO: Description


const-t:
In this [comment](https://github.com/tempesta-tech/tempesta/issues/2098#issuecomment-2097793114) described how to reproduce the leak.

fw/pool.c
```diff
@@ -190,37 +190,52 @@ tfw_pool_free(TfwPool *p, void *ptr, size_t n)
...
+void
+tfw_pool_clean_single(TfwPool *pool, void *ptr)
+{
+	TfwPoolChunk *c, *next, *prev = pool->curr;
+
+	BUG_ON(!pool);
+	BUG_ON(!ptr);
+
+	for (c = pool->curr->next; c; c = next) {
+		if (!(next = c->next))
```
EvgeniiMekhanik, line 205: Why we can't delete the last chunk if ptr belong to it?

fw/hpack.c
```diff
@@ -833,7 +833,8 @@ tfw_hpack_add_index(TfwHPackDTbl *__restrict tbl,
 				T_DBG3("%s: dropped index: %u\n", __func__,
 				       early);
 				if (cp->last)
-					tfw_pool_clean(tbl->h_pool, cp->hdr);
+					tfw_pool_clean_single(tbl->h_pool,
```
EvgeniiMekhanik, line 836: Previously the idea was to delete all chunks from ptr to the last one. Now we delete only single chunk. Can you please explain this change in the commit message

EvgeniiMekhanik (approved):

fw/pool.c
const-t: No, we must not delete current and must not delete head of the pool, that located in the last chunk. I have a lot of comments about it in tests.

enuribekov-tempesta (approved):
LGTM

Issue #2084, opened 2024-03-29
frang: `concurrent_tcp_connections` - the connection counter works incorrectly


RomanBelozerov:
Tempesta - 66e8acf486041cd6ebcc228816a911491739d2d1

For frang config:

```
frang_limits {
    concurrent_connections 2;
}
```

1. Establish 10 client connections from same IP (number of warning in dmesg - 8).
2. Close connections.
3. Wait (optional)
4. Establish new 10 client connection from that IP.

ER: Tempesta allows 2 connections and closes 8 connections. The last dmesg warning - `frang: connections max num. exceeded for 127.0.0.1: 8 (lim=2)`
AR: Tempesta closes 10 connections. The last dmesg warning - `frang: connections max num. exceeded for 127.0.0.1: 18 (lim=2)`

# Testing
`t_frang.test_concurrent_connections.ConcurrentConnections.test_clear_client_connection_stats_greater`

PR #2107, opened 2024-04-23
Add sk_free callback handler.

enuribekov-tempesta:
In this way tfw_classify_conn_close() function called when connection closed on the socket level.
In this particular case we fix the situation when tfw_classify_conn_close() skipped when connection dropped by connections limiter (Issue #2084).

### DIFF
--- a/fw/http_limits.c
+++ b/fw/http_limits.c
@@ -352,6 +352,9 @@ tfw_classify_conn_close(struct sock *sk)
 
 	spin_lock(&ra->lock);
 
+	if(ra == NULL)
+		return;
+
 	BUG_ON(!ra->conn_curr);
 	ra->conn_curr--;
 
--- a/fw/http_limits.c
+++ b/fw/http_limits.c
@@ -1809,6 +1809,7 @@ tfw_http_limits_hooks_register(void)
 
 static TempestaOps tempesta_ops = {
 	.sk_alloc	= tfw_classify_conn_estab,
+	.sk_free        = tfw_classify_conn_close,
 	.sock_tcp_rcv	= tfw_classify_tcp,
 };
 
--- a/fw/http_limits.c
+++ b/fw/http_limits.c
@@ -331,10 +331,8 @@ frang_conn_new(struct sock *sk, struct sk_buff *skb)
 	 * for a single client is absolutely straight-forward.
 	 */
 	r = frang_conn_limit(ra, dflt_vh->frang_gconf);
-	if (unlikely(r == T_BLOCK) && dflt_vh->frang_gconf->ip_block) {
+	if (unlikely(r == T_BLOCK) && dflt_vh->frang_gconf->ip_block)
 		tfw_filter_block_ip(cli);
-		tfw_client_put(cli);
-	}
 
 finish:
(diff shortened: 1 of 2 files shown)


EvgeniiMekhanik (approved):

EvgeniiMekhanik:
I think we should remove tfw_vhost_put in case when ip_block is set. See if (unlikely(r == T_BLOCK) && dflt_vh->frang_gconf->ip_block). Please check it.

enuribekov-tempesta:
> I think we should remove tfw_vhost_put in case when ip_block is set. See if (unlikely(r == T_BLOCK) && dflt_vh->frang_gconf->ip_block). Please check it. Otherwise we drop put it two times

Good idea, I'll check it.

enuribekov-tempesta:
@EvgeniiMekhanik 
Yes. When result of `frang_conn_limit()` is `T_BLOCK` and `ip_block` is enabled  `tfw_client_put()` called twice and led to crash.
But simple removing of  `tfw_vhost_put()` doesn't impact on it.
Looking for solution.

const-t (approved):
LGTM

fw/http_limits.c
```diff
@@ -350,8 +348,6 @@ tfw_classify_conn_close(struct sock *sk)
 {
 	FrangAcc *ra = frang_acc_from_sk(sk);
 
-	BUG_ON(!ra);
```
EvgeniiMekhanik, line 353: You should return here if ra == NULL. If cli = tfw_client_obtain fails. ra will be null here and we should immediately return here! otherwise we catch bug

fw/http_limits.c
```diff
@@ -361,7 +357,8 @@ tfw_classify_conn_close(struct sock *sk)
 
 	spin_unlock(&ra->lock);
 
-	sk->sk_security = NULL;
+	if(sk)
```
EvgeniiMekhanik, line 360: sk is always not zero here!

enuribekov-tempesta: https://github.com/tempesta-tech/linux-5.10.35-tfw/pull/25/files#r1582808517

fw/http_limits.c
```diff
@@ -350,18 +348,20 @@ tfw_classify_conn_close(struct sock *sk)
 {
 	FrangAcc *ra = frang_acc_from_sk(sk);
 
-	BUG_ON(!ra);
-
 	assert_spin_locked(&sk->sk_lock.slock);
 
 	spin_lock(&ra->lock);
 
+	if(ra == NULL)
```
EvgeniiMekhanik, line 355: Before spin_lock(&ra->lock);

fw/http_limits.c
EvgeniiMekhanik: I think he means BUG_ON(!ra);  ra is saved in sk_security

EvgeniiMekhanik (approved):

fw/http_limits.c
const-t: > You should return here if ra == NULL. If cli = tfw_client_obtain fails. ra will be null here and we should immediately return here! otherwise we catch bug

Exactly. I meant this in my comment to kernel patch.

PR #2110, opened 2024-05-02
Fix incorrect msg->len calculation during redirect preparation.

EvgeniiMekhanik:
msg len calculation was totally wrong: we should adjust status line len instead of length of S_301, because we can use different status codes, also we should adjust S_F_SET_COOKIE length.

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -533,7 +533,7 @@ tfw_http_prep_date(char *buf)
 	tfw_http_prep_date_from(buf, tfw_current_timestamp());
 }
 
-static inline char *
+char *
 tfw_http_resp_status_line(int status, size_t *len)
 {
 	switch(status) {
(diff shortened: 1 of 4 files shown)


fw/http.c
```diff
@@ -808,16 +809,18 @@ do { 								\
...
 		body_val = p;
 		body_len = tfw_str_to_cstr(body, body_val, remaining);
 	}
 
+	status_line_len = strlen(status_line);
 	{
 		TfwStr msg = {
 			.chunks = (TfwStr []){
 				{ .data = status_line,
-				  .len = strlen(status_line) },
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed

fw/http.c
```diff
@@ -815,7 +813,7 @@ do { 								\
 		body_len = tfw_str_to_cstr(body, body_val, remaining);
 	}
 
-	status_line_len = strlen(status_line);
+	status_line_len = sizeof(status_line);
```
const-t, line 816: `sizeof` return size of pointer in this case.

const-t (changes requested):

fw/http.c
EvgeniiMekhanik: Yes I make this change automatically according our discussion(

krizhanovsky (approved):
LGTM

const-t (approved):

PR #2116, opened 2024-05-07
Decrement SS_V_ACT_LIVECONN if connection_new failed

EvgeniiMekhanik:
`ss_linkerror` decrements SS_V_ACT_LIVECONN only if sk_user_data is not NULL, so we should decrement
it explicitly if sk_user_data is not set.


const-t (approved):

krizhanovsky (commented):
LGTM , but please update https://tempesta-tech.com/knowledge-base/Debugging-and-troubleshooting/#error-injection and provide there an example (or reference it in the tests code) in the wiki

PR #2125, opened 2024-05-24
Implement trailer checking for HTTP/2

biathlon3:


enuribekov-tempesta (approved):
LGTM

fw/hpack.c
```diff
@@ -1645,6 +1679,12 @@ tfw_hpack_decode(TfwHPack *__restrict hp, unsigned char *__restrict src,
 			if ((r = frang_http_hdr_limit(req, parser->hdr.len)))
 				goto out;
 
+			if (test_bit(TFW_HTTP_B_HEADERS_PARSED, req->flags))
```
const-t, line 1682: `frang_http_req_trailer_check()` has a comment:

> NOTE: we check only regular headers (without HTTP/2-specific pseudo-header fields), since pseudo-headers must not appear in trailers (RFC 7540 section 8.1.2.1), and during parsing stage, in @H2_MSG_VERIFY(), we have already verified that."

`H2_MSG_VERIFY()` renamed to `__h2_msg_verify()` and seems this function doesn't do any trailer checks. Need need find out why `__h2_msg_verify()` these checks was removed from the function. And I guess move your checks to this function. Looks like `__h2_msg_verify()` more suitable for it, because we invoke it, before parsing value.

fw/hpack.c
```diff
@@ -1388,6 +1388,40 @@ tfw_hpack_hdr_set(TfwHPack *__restrict hp, TfwHttpReq *__restrict req,
...
+	 *
+	 * Trailers MUST NOT include pseudo-header fields.
+	 */
+	switch (tag) {
+	case TFW_TAG_HDR_H2_STATUS:
+	case TFW_TAG_HDR_H2_METHOD:
+	case TFW_TAG_HDR_H2_SCHEME:
+	case TFW_TAG_HDR_H2_AUTHORITY:
+	case TFW_TAG_HDR_H2_PATH:
+	case TFW_TAG_HDR_ACCEPT:
```
EvgeniiMekhanik, line 1405: I see only ":method", ":scheme", ":authority" , ":path"  pseudo-headers in [8.3.1. ](https://www.rfc-editor.org/rfc/rfc9113#section-8.3.1)[Request Pseudo-Header Fields](https://www.rfc-editor.org/rfc/rfc9113#name-request-pseudo-header-field) and [8.1.2.3.](https://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2.3) [Request Pseudo-Header Fields](https://httpwg.org/specs/rfc7540.html#HttpRequest)

fw/hpack.c
EvgeniiMekhanik: I am not sure that we can do it in `__h2_msg_verify`. This function (__h2_msg_verify) is not called when we deal with indexed header (it was added to hpack table and not parsed).

EvgeniiMekhanik (changes requested):
Need to fix some places

EvgeniiMekhanik:
You also need to enabled tests blocked by this issue in appropriate branch in tempesta-test

fw/hpack.c
biathlon3: I added a comment why this list of headers is used.

fw/hpack.c
biathlon3: I added comment comment why I wrote checks here. It is possible to move one check to __h2_msg_verify() , but in this case we will have two checks in different places of the code.

EvgeniiMekhanik (approved):
Please also rewrite comment in `frang_http_req_trailer_check`. We don't verify that pseudo-headers must not appear in trailers in H2_MSG_VERIFY

const-t (approved):
LGTM

PR #2129, opened 2024-05-30
En/fix 1988

enuribekov-tempesta:


fw/apm.h
```diff
@@ -88,8 +88,10 @@ TfwHMStats *tfw_apm_hm_stats(void *apmref);
 
 /* Health monitor procedures ('health_check' directive). */
 void tfw_apm_hm_srv_rcount_update(TfwStr *uri_path, void *apmref);
-bool tfw_apm_hm_srv_alive(int status, TfwStr *body, struct sk_buff *skb_head,
-			  void *apmref);
+//bool tfw_apm_hm_srv_alive(int status, TfwStr *body, struct sk_buff *skb_head,
+		//	  void *apmref);
```
krizhanovsky, line 92: Please don't leave commented out code

enuribekov-tempesta: Automatic patching is an evil. Fixed.

fw/apm.c
```diff
@@ -1327,30 +1326,45 @@ __tfw_apm_crc32_calc(TfwMsgIter *it, TfwStr *chunk , struct sk_buff *skb_head,
 }
 
 bool
-tfw_apm_hm_srv_alive(int status, TfwStr *body, struct sk_buff *skb_head,
-		     void *apmref)
+tfw_apm_hm_srv_alive(TfwHttpResp *resp, TfwServer *srv)
```
keshonok, line 1329: I don't really have anything against this shortening of the list of arguments, just wanted to comment that the logic as I understand it was that all arguments on the original long list were already available in the calling function, hence there was no need to derive each of them again.

fw/apm.c
```diff
@@ -1327,30 +1326,45 @@ __tfw_apm_crc32_calc(TfwMsgIter *it, TfwStr *chunk , struct sk_buff *skb_head,
...
 	u32 crc32 = 0;
 	TfwMsgIter it;
 	TfwStr chunk = {0};
 
 	BUG_ON(!hm);
+       /*
+       * If hm->codes exists
+       * and 'status' has not in hm->codes.
+       * return NOT ALIVE
+       */
```
keshonok, line 1345: I believe this is an unnecessary comment as it simply repeats what is already clearly expressed in the code.

fw/apm.c
```diff
@@ -1327,30 +1326,45 @@ __tfw_apm_crc32_calc(TfwMsgIter *it, TfwStr *chunk , struct sk_buff *skb_head,
...
 		return false;
 	}
-
-	if (!body->len) {
+	/*
+	 * Check CRC for zero body
+	 * and return ALIVE if CRC also zero,
+	 * go to process CRC error if not
+	 * and continue processing if body exists.
+	 */
```
keshonok, line 1356: I believe this is an unnecessary comment as it simply repeats what is already clearly expressed in the code.

fw/apm.c
```diff
@@ -1327,30 +1326,45 @@ __tfw_apm_crc32_calc(TfwMsgIter *it, TfwStr *chunk , struct sk_buff *skb_head,
...
 			goto crc_err;
 		return true;
 	}
 
-	if (unlikely(tfw_body_iter_init(&it, &chunk, body->data, body->skb,
-					skb_head)))
-	{
+	/*
+	 * If body is invalid return NOT ALIVE
+	 */
```
keshonok, line 1365: I believe this is an unnecessary comment as it simply repeats what is already clearly expressed in the code.

fw/apm.c
```diff
@@ -1360,16 +1374,19 @@ tfw_apm_hm_srv_alive(int status, TfwStr *body, struct sk_buff *skb_head,
 	 * Special case for 'auto' monitor: generate crc32
 	 * from body of first response and store it into monitor.
 	 */
-	if (!hm->crc32 && hm->auto_crc) {
-		hm->crc32 = __tfw_apm_crc32_calc(&it, &chunk, skb_head, body);
+	if (hm->crc32 == 0 && hm->auto_crc) {
```
keshonok, line 1377: Does the original variant leads to some sort of UB? Just asking.

fw/apm.c
enuribekov-tempesta: "Regarding tfw_http_hm_control() - we return from the function if the server is alive (not suspended) since we do not need to mark it alive with tfw_srv_mark_alive(), so the code seems correct. This is an example why it's bad to write code without comments :) "
https://github.com/tempesta-tech/tempesta/pull/1983#pullrequestreview-1663079775

But OK, commit removed.

fw/apm.c
enuribekov-tempesta: `skb_head` is unused argument. And it presence was a motivation to review and change `tfw_apm_hm_srv_alive()`
 signature. 
 The other is just a pinch of readability.

fw/apm.c
enuribekov-tempesta: This is a not simple shortening.
Here we see what entities function uses - response and server, not a bunch of heterogenous parameters, where we can loose unused ones.
As a model for `tfw_apm_hm_srv_alive()` was used `tfw_http_hm_suspend()`.

fw/apm.c
keshonok: Oh, I meant `!hm->crc32` as the original variant versus `hm->crc32 == 0`. Just didn't see the sense in changing it unless there's UB that I don't know of. :-) Usually these kinds of micro-changes are not encouraged and the original variant is a typical idiomatic C language construction.

fw/apm.c
keshonok: Well, then this should be your comment. Describe the actual meaning of what's done in the code, not just rephrase the code with words.

fw/apm.c
enuribekov-tempesta: Idiomatic :)
Classic idiomas are changing.
https://github.com/tempesta-tech/tempesta/issues/1695#issuecomment-2014672322
Agree, it is not this case, but I've already changed these piece of code and direct comparation  show that is numeric value, not pointer or boolean.

const-t: There is no sense to make these changes. [#1695 (comment)](https://github.com/tempesta-tech/tempesta/issues/1695#issuecomment-2014672322) is about address comparison, it's not related to primitive types.

Even for pointers we still can use `(!p)`, `-Waddress` intended to catch such cases: `NULL + 1 == NULL`

enuribekov-tempesta: OK

fw/http.c
```diff
@@ -1658,14 +1658,10 @@ tfw_http_hm_control(TfwHttpResp *resp)
...
 		return;
 
-	if (!tfw_srv_suspended(srv) ||
-	    !tfw_apm_hm_srv_alive(resp->status, &resp->body, resp->msg.skb_head,
-				  srv->apmref))
-	{
-		return;
-	}
-
+	if (tfw_srv_suspended(srv) && tfw_apm_hm_srv_alive(resp, srv)) {
```
const-t, line 1661: You've rewritten this condition, but are sure that logic is correct? The main go of the issue is find out whether should we do it this way or not.

Let's see, now you calling `tfw_apm_hm_srv_alive()` only for suspended servers(the same as before), but `tfw_http_hm_control()` calls only when response received, it means we can receive response from suspended server. Why?

fw/apm.c
```diff
@@ -1327,10 +1326,13 @@ __tfw_apm_crc32_calc(TfwMsgIter *it, TfwStr *chunk , struct sk_buff *skb_head,
...
-tfw_apm_hm_srv_alive(int status, TfwStr *body, struct sk_buff *skb_head,
-		     void *apmref)
+tfw_apm_hm_srv_alive(TfwHttpResp *resp, TfwServer *srv)
 {
-	TfwApmHM *hm = READ_ONCE(((TfwApmRef *)apmref)->hmctl.hm);
+	int status = resp->status;
+	TfwStr *body = &resp->body;
+
+	TfwApmHM *hm = READ_ONCE(((TfwApmRef *)(srv->apmref))->hmctl.hm);
+
```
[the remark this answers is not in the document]
enuribekov-tempesta: Done

fw/http.c
```diff
@@ -1658,14 +1658,10 @@ tfw_http_hm_control(TfwHttpResp *resp)
...
 	tfw_srv_mark_alive(srv);
+	}
+	return;
 }
```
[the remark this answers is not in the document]
enuribekov-tempesta: Done

fw/apm.c
```diff
@@ -1348,9 +1350,8 @@ tfw_apm_hm_srv_alive(int status, TfwStr *body, struct sk_buff *skb_head,
 		return true;
 	}
 
-	if (unlikely(tfw_body_iter_init(&it, &chunk, body->data, body->skb,
-					skb_head)))
-	{
+	if (unlikely(tfw_body_iter_init(&it, &chunk, body->data, body->skb, 
+					resp->msg.skb_head ))) {
```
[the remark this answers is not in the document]
enuribekov-tempesta: Done

fw/http.c
enuribekov-tempesta: !a || !b = !(a && b)

enuribekov-tempesta:
Created new:
https://github.com/tempesta-tech/tempesta/pull/2151

PR #2131, opened 2024-06-07
Implementation of the ability to send js challenge according with custom rules.

biathlon3:


enuribekov-tempesta (approved):
LGTM

fw/http_tbl.c
```diff
@@ -499,6 +503,10 @@ tfw_cfgop_http_rule(TfwCfgSpec *cs, TfwCfgEntry *e)
 		rule->act.type = TFW_HTTP_MATCH_ACT_CACHE_TTL;
 		rule->act.cache_ttl = act_val_parsed;
 	}
+	else if (!strcasecmp(action, "jsch")) {
+		is_jsch_global = false;
```
EvgeniiMekhanik, line 507: Here is inconsistent behavior - you change global variable only if one jsch is set.
For example:
config = """
http_chain {
    host == "aaa" -> jsch;
    host == "bbb" ->  good;
} 

If host == "aaa" ->jsch is set we don't support js challenge for "bbb". But if it is not set we support it! Why? How   host == "aaa" -> jsch; deal with host "bbb"?

fw/http_sess.c
```diff
@@ -588,17 +588,19 @@ tfw_http_sess_resp_process(TfwHttpResp *resp, bool cache)
...
-	/*
-	 * RFC 6265 4.1.1 and 4.1.2 says that we should not set session cookie
-	 * if it's not necessary. Since client didn't send up the cookie and
-	 * it seems that we don't enforce them, we can just set the cookie in
-	 * each response forwarded to the client.
-	 */
-	if (test_bit(TFW_HTTP_B_HAS_STICKY, req->flags))
-		return 0;
-	return tfw_http_sticky_add(resp, cache);
+	if (req->need_jsch) {
```
EvgeniiMekhanik, line 591: This function is called for all types of cookies which are set by tempesta. It is not only about js_challenge, so this check is wrong

fw/http.c
```diff
@@ -6134,37 +6142,39 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb,
 	 * to GET. We should send js challenge to the client because the real
 	 * method, expected by the client is GET.
 	 */
-	switch (tfw_http_sess_obtain(req)) {
-	case TFW_HTTP_SESS_SUCCESS:
-		break;
+	if (req->need_jsch) {
```
EvgeniiMekhanik, line 6145: This check here is wrong. tfw_http_sess_obtain called not only because of js challenge! see `cookie learned`. I add new tests for it.

fw/http_sess.c
```diff
@@ -843,20 +847,22 @@ tfw_http_sess_obtain(TfwHttpReq *req)
...
-		return r;
-	default:
-		/*
-		 * Any js challenge processing error: cookie not found
-		 * or invalid or request comes not in time. We increment
-		 * max_misses and restart js challenge.
-		 */
-		BUG_ON(r < __TFW_HTTP_SESS_PUB_CODE_MAX);
-		return tfw_http_sticky_challenge_start(req);
+	if (req->need_jsch) {
```
EvgeniiMekhanik, line 850: This check is wrong here. We get cookies for cookie learn also.

fw/http_sess.c
```diff
@@ -666,6 +668,8 @@ tfw_http_sess_check_jsch(StickyVal *sv, TfwHttpReq* req)
 {
 	unsigned long min_time;
 	TfwCfgJsCh *js_ch = req->vhost->cookie->js_challenge;
+	if (!req->need_jsch)
```
EvgeniiMekhanik, line 671: Please combine it with the nex check in one if

fw/http_tbl.c
```diff
@@ -156,6 +156,10 @@ tfw_http_tbl_scan(TfwMsg *msg, TfwHttpTable *table, TfwHttpActionResult *action)
 			return 0;
 		case TFW_HTTP_MATCH_ACT_BLOCK:
 			return -1;
+		case TFW_HTTP_MATCH_ACT_JSCH:
+			action->type = TFW_HTTP_RES_JSCH;
+			action->vhost = tfw_vhost_lookup_default();
```
EvgeniiMekhanik, line 161: Its wrong. I add test to check it.
http_chain {{
            uri == "*1.html" -> jsch;
            host == "good" -> good;
            -> default;
        }}

For such configuration if host is good and uri == "*1.html" it should go to good vhost  not default!

EvgeniiMekhanik (changes requested):
Need to rework. We also need changes in etc/tempesta_fw.conf - descriptions and examples

PR #2133, opened 2024-06-07
Fix cleaning not initialized http2 context.

EvgeniiMekhanik:
We initialize http2 context in `tfw_tls_over`, but
if this function was not called, because tls handshake
fails we clean uninitialized context, that can lead to
kernel BUG.

### DIFF
--- a/fw/tls.c
+++ b/fw/tls.c
@@ -611,9 +611,11 @@ tfw_tls_conn_dtor(void *c)
 {
 	struct sk_buff *skb;
 	TlsCtx *tls = tfw_tls_context(c);
+	bool hs_was_done = false;
 
-	if (TFW_FSM_TYPE(((TfwConn *)c)->proto.type) == TFW_FSM_H2)
-		tfw_h2_context_clear(tfw_h2_context(c));
+	if (TFW_CONN_PROTO((TfwConn *)c) == TFW_FSM_H2
+	    && (hs_was_done = ttls_hs_done(tls)))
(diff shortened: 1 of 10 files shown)


EvgeniiMekhanik:
Also we can revert moving `tfw_h2_context_init` from last commit in master instead of this patch

EvgeniiMekhanik:
We also can move `tfw_h2_context_init` back to `tfw_tls_conn_init`

const-t (approved):

fw/tls.c
```diff
@@ -611,9 +611,11 @@ tfw_tls_conn_dtor(void *c)
 {
 	struct sk_buff *skb;
 	TlsCtx *tls = tfw_tls_context(c);
+	bool hs_was_done = false;
 
-	if (TFW_FSM_TYPE(((TfwConn *)c)->proto.type) == TFW_FSM_H2)
-		tfw_h2_context_clear(tfw_h2_context(c));
+	if (TFW_CONN_PROTO((TfwConn *)c) == TFW_FSM_H2
+	    && (hs_was_done = ttls_hs_done(tls)))
```
const-t, line 617: Why do we need `hs_was_done`?

EvgeniiMekhanik: Yes we can remove this variable. First of all i plan to pass it to `tfw_cli_conn_release`, to clear &ctx->skb_head only for intialized context. But I understand, that we cant do it in a good way. Better to clear &ctx->skb_head in context cleanup function

PR #2139, opened 2024-06-14
2058 prohibit delete method in default configuration

biathlon3:


fw/vhost.c
```diff
@@ -2572,6 +2615,188 @@ tfw_vhost_cfgclean(void)
...
+}
+
+
+static void
+tfw_cfgop_location_print(TfwLocation *loc)
+{
+        int i;
+         TfwHdrMods *h_mods = &loc->mod_hdrs[TFW_VHOST_HDRMOD_RESP];
+
+        if (loc->arg) {
```
krizhanovsky, line 2684:
```suggestion
        if (!loc->arg)
                return;
```
Please do your best to reduce nesting in the code

fw/vhost.c
```diff
@@ -2572,6 +2615,188 @@ tfw_vhost_cfgclean(void)
 	tfw_global.hdr_via = s_hdr_via_dflt;
 }
 
+#if defined(PRINT_CONFIG)
+void
+printFrang(const char *tab, const FrangVhostCfg *frang)
+{
+	int i;
+
+	if (frang) {
```
krizhanovsky, line 2624:
```suggestion
	if (!frang)
	        return;
```

fw/vhost.c
```diff
@@ -2572,6 +2615,188 @@ tfw_vhost_cfgclean(void)
...
+
+	memset(str, 0, sizeof(str));
+
+	tfw_cfgop_frang_global_print(&tfw_frang_glob_reconfig);
+
+	T_LOG_NL("tls_match_any_server_name=%s\n",
+		 tfw_tls_get_allow_any_sni_reconfig() ? "true" : "false");
+
+	hash_for_each(vhosts->vh_hash, i, vhost, hlist) {
+
```
krizhanovsky, line 2785: extra empty line

fw/vhost.c
```diff
@@ -2577,6 +2615,188 @@ tfw_vhost_cfgclean(void)
...
+			 frang->http_trailer_split ? "true" : "false");
+		T_LOG_NL("%s   http_method_override=%s\n", tab,
+			 frang->http_method_override ? "true" : "false");
+
+		T_LOG_NL("%s}\n", tab);
+	}
+}
+
+
+void printBasicStr(BasicStr *str)
```
const-t: Usually we don't use camel case for function's name in C code.

fw/vhost.c
```diff
@@ -2154,8 +2163,12 @@ tfw_cfgop_frang_body_len(TfwCfgSpec *cs, TfwCfgEntry *ce)
 {
 	int r;
 	FrangVhostCfg *cfg = tfw_cfgop_frang_get_cfg();
+	long deflt;
 
-	if (ce->dflt_value && cfg->http_body_len)
+	if ((r = tfw_cfg_parse_long(cs->deflt, &deflt)))
+		return r;
+
+	if (ce->dflt_value && cfg->http_body_len != deflt)
```
const-t, line 2171: Here we trying to not override values that not specified in new `frang_limits` by default values. Please describe it in commit message. And also, looks like if `ce->dflt_value` is false we can skip default value parsing. 

@krizhanovsky FYI it looks controversial to have two different `deflt` values for single directive in different scopes.

const-t (approved):
LGTM, but need to make little corrections.

fw/connection.c
```diff
@@ -174,7 +174,9 @@ tfw_connection_recv(TfwConn *conn, struct sk_buff *skb)
 				r = tfw_ws_msg_process(conn, skb);
 			else
 				r = tfw_http_msg_process(conn, skb, &splitted);
-			if (r == T_DROP && splitted) {
+
+			if (splitted && (r == T_DROP || r == T_BLOCK_WITH_RST
```
EvgeniiMekhanik, line 178: r can be also -ERRNO  here, I think that we should not check `r` here, but `splitted` should be zero if r == T_OK or r == T_POSTPONED. Please check it.

fw/http.c
```diff
@@ -5987,8 +5987,9 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb,
 			TFW_INC_STAT_BH(clnt.msgs_otherr);
 			return tfw_http_req_parse_drop(req, 500,
 					"Can't split pipelined requests",
-					HTTP2_ECODE_PROTO);
+					HTTP2_ECODE_PROTO);	
 		}
+		*splitted = skb;
```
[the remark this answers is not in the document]
EvgeniiMekhanik: We need to add test for this case

fw/http.c
```diff
@@ -5987,8 +5987,9 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb,
 			TFW_INC_STAT_BH(clnt.msgs_otherr);
 			return tfw_http_req_parse_drop(req, 500,
 					"Can't split pipelined requests",
-					HTTP2_ECODE_PROTO);
+					HTTP2_ECODE_PROTO);
```
EvgeniiMekhanik, line 5990: Extra space after HTTP2_ECODE_PROTO);

fw/vhost.c
```diff
@@ -1253,14 +1263,17 @@ tfw_frang_cfg_inherit(FrangVhostCfg *curr, const FrangVhostCfg *from)
...
 		}
-		delta = (void *)from->http_ct_vals - (void *)curr->http_ct_vals;
-		memcpy(curr->http_ct_vals, from->http_ct_vals, sz);
-		curr->http_ct_vals->vals = (void *)curr->http_ct_vals
-				+ sizeof(FrangCtVals);
-		curr->http_ct_vals->data -= delta;
-		/* Restore data pointers. */
-		for (val = curr->http_ct_vals->vals; val->str; ++ val)
-			val->str -= delta;
+		else {
```
EvgeniiMekhanik, line 1266: We don't need this `else`, because we return on `if`

fw/vhost.c
EvgeniiMekhanik: Here he try to not override values that not specified in new frang_limits (default values) by values specified in top level frang config

fw/vhost.c
```diff
@@ -2729,60 +2965,77 @@ static TfwCfgSpec tfw_global_frang_specs[] = {
...
-		.handler = tfw_cfgop_frang_uri_len,
+		.handler = tfw_cfgop_frang_glob_set_int,
+		.dest = &tfw_frang_vhost_reconfig.http_uri_len,
 		.allow_reconfig = true,
 	},
 	{
 		.name = "http_body_len",
 		.deflt = "1073741824", /* 1 Gb. */
-		.handler = tfw_cfgop_frang_body_len,
+		.handler = tfw_cfgop_frang_glob_set_long,
```
EvgeniiMekhanik, line 2986: We have inconsistency behavior here.
```
frang_limits {
    http_body_len 0;
}
 frang_limits {
 }
 ```
I have `http_body_len=1073741824` (I think that we should 0), because default value should not override value which was set previously. For
```
frang_limits {
    http_ct_required true;
}
 frang_limits {
}
```
I have  http_ct_required=true. 

I think we have the same problem for `concurrent_tcp_connections`, `http_header_cnt`, `http_body_len`, `http_strict_host_checking`, `http_methods`. For example in vhost section you check  if (ce->dflt_value && cfg->http_body_len != deflt) for `http_body_len` but in global section if (ce->dflt_value && *(unsigned int *)(cs->dest). But ce->dest can be 0, which is specified by administrator!

File: fw/vhost.c [hunk 11]
```diff
@@ -2572,6 +2644,170 @@ tfw_vhost_cfgclean(void)
...
+		return;
+
+	memset(str, 0, sizeof(str));
+	tfw_cfgop_frang_global_print(&tfw_frang_glob_reconfig);
+
+	T_LOG_NL("tls_match_any_server_name=%s\n",
+		 tfw_tls_get_allow_any_sni_reconfig() ? "true" : "false");
+	hash_for_each(vhosts->vh_hash, i, vhost, hlist) {
+                len = vhost->name.len < 128 ? vhost->name.len : 127;
+                memcpy(str, vhost->name.data, len);
```
EvgeniiMekhanik, line 2800: Why we can't use `strncpy` here and do not call  memset(str, 0, len); every time?

fw/vhost.c
```diff
@@ -2572,6 +2644,170 @@ tfw_vhost_cfgclean(void)
...
+                T_LOG_NL("vhost %s {", str);
+                memset(str, 0, len);
+                tfw_cfgop_location_print(vhost->loc_dflt);
+                for (j = 0; j < vhost->loc_sz; ++j) {
```
EvgeniiMekhanik, line 2804: Extra `{`

fw/vhost.c
```diff
@@ -2572,6 +2644,170 @@ tfw_vhost_cfgclean(void)
 	tfw_global.hdr_via = s_hdr_via_dflt;
 }
 
+#if defined(PRINT_CONFIG)
+static void
+printFrang(const char *tab, const FrangVhostCfg *frang)
```
EvgeniiMekhanik, line 2649: we don't use camelCase for function names

fw/vhost.c
```diff
@@ -21,6 +21,9 @@
 #include <linux/slab.h>
 #include <linux/sort.h>
 
+#if defined(DEBUG)
+#define PRINT_CONFIG
```
EvgeniiMekhanik, line 25: I think we should not need extra PRINT_CONFIG. If DBG_VHOST is specified we can print it

fw/vhost.c
```diff
@@ -2729,60 +2965,77 @@ static TfwCfgSpec tfw_global_frang_specs[] = {
 		},
 		.allow_reconfig = true,
 	},
-	/* Option can be redefined per vhost|location. */
+	/* Option can be redefined per vhost|location.
+	 *
+	 *  All handler are changed to tfw_cfgop_frang_glob_...
+	 *  because here we need to compare the default value
+	 *  with 0 after kzalloc, and for non global configuration
```
EvgeniiMekhanik, line 2972: I don't think that it is ok. For `http_body_len` we have different behavior then for `http_uri_len` for example.

EvgeniiMekhanik (changes requested):
Need fixes

fw/vhost.c
biathlon3: I agree memset is redundant here, but, seems, memcpy and strncpy are very similar in this case.

fw/http2.c
```diff
@@ -283,16 +283,21 @@ tfw_h2_remove_idle_streams(TfwH2Ctx *ctx, unsigned int id)
...
-	list_for_each_entry_safe_reverse(stream, tmp, &ctx->idle_streams.list,
-					 hcl_node)
-	{
-		if (id <= stream->id)
-			break;
+	/*
+	 * I add this because we get BUG: kernel NULL pointer dereference,
+	 * in case ERROR: cannot establish a new h2 connection
+	 */
+	if (ctx->idle_streams.num)
```
EvgeniiMekhanik, line 290: This is unnecessary, as I think it should not be needed on the current master. If not please implement tests when it is necessary

fw/vhost.c
```diff
@@ -2043,26 +2064,44 @@ tfw_cfgop_frang_glob_set_bool(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+	if (ce->dflt_value && cs->__called_cfg)
 		return 0;
 	return tfw_cfg_set_bool(cs, ce);
 }
 
 static int
 tfw_cfgop_frang_glob_set_int(TfwCfgSpec *cs, TfwCfgEntry *ce)
 {
-	if (ce->dflt_value && *(unsigned int *)(cs->dest))
+	if (ce->dflt_value && cs->__called_cfg)
```
EvgeniiMekhanik, line 2075: It seems we can't use `__called_cfg` for this purpose. This value is not zeroed after fail reload. https://github.com/tempesta-tech/tempesta-test/pull/598/commits/3683828fff2a921e2f6ee6c2666192dcf78614fc look last test It doesn't pass

fw/http2.c
biathlon3: Yes. I agree, I wrote it before changing on the master.

fw/vhost.c
biathlon3: Initially, I used a separate variable, but found that it almost always changes in parallel with __called_cfg.
I fixed this error.

EvgeniiMekhanik (approved):

EvgeniiMekhanik:
I remove @KonsKo approve according our discussion.

KonsKo:
> I remove @KonsKo approve according our discussion.

I suppose wrong person was mentioned.

fw/vhost.c
const-t: Seems we still have error here for following case:
RC1:
```
http_limits {
directive-correct;
directive-incorrect;
}
```
RC2:
```
http_limits {
another-new-directive;
}
```
If during the first reconfiguration using RC1 config, we fail on `directive-incorrect` after calling handler for `directive-correct`, then on the second reconfiguration using RC2, default value for `directive-correct` will not be set.

const-t (approved):
LGTM

PR #2140, opened 2024-06-16
Fix memory leak

EvgeniiMekhanik:
We don not call `tfw_sync_socket_exit` and
`tfw_cache_stop` in case when `tfw_sync_socket_init` or `tfw_cache_start`fails, so we don't clear work queue array that leads to memory leak.

### DIFF
--- a/fw/sock.c
+++ b/fw/sock.c
@@ -1805,6 +1805,27 @@ ss_active(void)
 	return READ_ONCE(__ss_active);
 }
 
+static inline int __init
+tfw_sync_socket_wq_init(int cpu)
+{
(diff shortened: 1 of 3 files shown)


const-t (approved):

fw/sock.c
```diff
@@ -1805,6 +1805,27 @@ ss_active(void)
 	return READ_ONCE(__ss_active);
 }
 
+static inline int
```
krizhanovsky, line 1808:
```suggestion
static inline int __init
```

krizhanovsky (approved):
LGTM

PR #2144, opened 2024-06-19
feat(2142): integrate checkpatch.pl to pre-commit git hook

### LINKED ISSUE #2142: Integrate `checkpatch.pl` to our build/commit pipeline
# Motivation

It's too time consuming to comment coding style violations in each pull request.

# Scope

The Linux kernel provides [checkpatch.pl](https://docs.kernel.org/dev-tools/checkpatch.html) script, which verifies patches along regular C files. We should either integrate it into git commit hook or github actions to be run on each pull request, that an author can fix all the violation w/o reviewer.

An example:
```bash
$ ./scripts/checkpatch.pl --no-tree -f ../tempesta/fw/sync_socket.h 
WARNING: Missing or malformed SPDX-License-Identifier tag in line 1
#1: FILE: ../tempesta/fw/sync_socket.h:1:
+/**

WARNING: do not add new typedefs
#32: FILE: ../tempesta/fw/sync_socket.h:32:
+typedef struct ss_proto_t {

WARNING: do not add new typedefs
#66: FILE: ../tempesta/fw/sync_socket.h:66:
+typedef struct tfw_conn_t TfwConn;

WARNING: do not add new typedefs
#69: FILE: ../tempesta/fw/sync_socket.h:69:
+typedef struct ss_hooks {

WARNING: do not add new typedefs
#95: FILE: ../tempesta/fw/sync_socket.h:95:
+typedef struct {

total: 0 errors, 5 warnings, 188 lines checked

NOTE: For some of the reported defects, checkpatch may be able to
      mechanically convert to the typical style using --fix or --fix-inplace.

../tempesta/fw/sync_socket.h has style problems, please review.

NOTE: If any of the errors are false positives, please report
      them to the maintainer, see CHECKPATCH in MAINTAINERS.
```

# Adjustments

We still have small changes from the kernel Coding style, which must be changed in the script:
- [x] if-else-if conditions on a new line https://github.com/tempesta-tech/tempesta/blob/master/CodingStyle#L155
- [x] CamelCase usage for types https://github.com/tempesta-tech/tempesta/blob/master/CodingStyle#L293
- [x] `typedef` usage for types https://github.com/tempesta-tech/tempesta/blob/master/CodingStyle#L293
- [x] function type is on a separate line https://github.com/tempesta-tech/tempesta/blob/master/CodingStyle#L346

kingluo:
close #2142


scripts/checkpatch.pl
```diff
@@ -0,0 +1,7820 @@
...
+		if ($sline =~ /^.\s*\}\s*else\s*$/ ||
+		    $sline =~ /^.\s*else\s*\{\s*$/) {
+			CHK("BRACES", "Unbalanced braces around else statement\n" . $herecurr);
+		}
+
+# check for if-else-if conditions on a new line
+		if ($sline =~ /^.\s*\}\s*else\s*/ ||
+		    $sline =~ /^.\s*\}\s*else if\s*/) {
+			WARN("BRACES", "`else` or `else if` should on a new line\n" . $herecurr);
+		}
```
krizhanovsky, line 6247: We're OK with
```C
if (foo) {
        // ...
} else {
        // ...
}
```
But not OK with
```C
if (foo) {
        // ...
} else if (bar) {
        // ...
}
```

scripts/checkpatch.pl
```diff
@@ -0,0 +1,7820 @@
...
+					#print "APW: ALLOWED: chunk-1 block<$block>\n";
+					$allowed = 1;
+				}
+			}
+			if ($level == 0 && $block =~ /^\s*\{/ && !$allowed) {
+				my $cnt = statement_rawlines($block);
+				my $herectx = get_stat_here($linenr, $cnt, $here);
+
+				WARN("BRACES",
+				     "braces {} are not necessary for single statement blocks\n" . $herectx);
```
krizhanovsky, line 6233: I tested the script against my last patch https://github.com/tempesta-tech/tempesta/pull/2157/files and it shows warning 
```
WARNING: braces {} are not necessary for single statement blocks
#1404: FILE: tls/x509.c:303:
+	else if (ret != TTLS_ERR_ASN1_UNEXPECTED_TAG) {
+		return TTLS_ERR_X509_INVALID_ALG + ret;
 	}
```
which contradicts the coding style https://github.com/tempesta-tech/linux-6.8.9-tfw/blob/main/Documentation/process/coding-style.rst Chapter 3 . Seems the upstream issue, but probably still worth fixing (maybe also to send the fix to LKML?)

.githooks/pre-commit
```diff
@@ -1,10 +1,14 @@
...
-
-# pre-commit hook to update copyrights
+set -euo pipefail
 
 year=$(date +%Y)
 files=$(git diff --cached --name-only)
 
+# check code style
+git diff --cached --diff-filter=ACM -- '*.c' '*.h' | \
+	./scripts/checkpatch.pl --no-tree --ignore ELSE_AFTER_BRACE,SPDX_LICENSE_TAG -
```
krizhanovsky, line 9: The script returns warnings and errors, so maybe block commit if there are any error? I just don't want to review pull requests with many style violations. (BTW #2157 also has many violations). @const-t I'd love to know your feedback on this.

krizhanovsky (changes requested):
It seems the CamelCase from 2142 is missed and there are also a couple of places to fix. @const-t I also assign the PR to you - I'd love to know your opinion at least on the integration part and what do we require mandatory and what's we can live with.

.githooks/pre-commit
kingluo: Yes, so that's why I set the bash options at the head of the script. It will fail out in case of any error.

https://github.com/tempesta-tech/tempesta/blob/80436fb92f38da6958cdcc5286c3c9729b82f422/.githooks/pre-commit#L2

const-t: Looks good to me. I agree, we must not allow to commit patch that has warnings/error. In exceptional cases when `checkpatch.pl` gives false positive result committer can use `no-verify`, to bypass commit hook.

const-t (approved):

scripts/checkpatch.pl
kingluo: Yes, this is a bug in `checkpatch.pl`, but it only applies to patch files, not to source files.

scripts/checkpatch.pl
kingluo: fixed.

scripts/checkpatch.pl
krizhanovsky: OK, why not to fix this? We, as the upstream kernel, allow:
```C
if (foo)
     bar;
```
and do not appreciate
```C
if (foo) {
     bar;
}
```
Also (like for the native Linux coding style) we do not allow things like:
```C
if (foo) {
     bar1;
} else
     bar2;

if (foo) {
     bar1;
}
else  if (foo2)
     bar2;
```
and require:
```C
if (foo) {
     bar1;
} else {
     bar2;
}

if (foo) {
     bar1;
}
else  if (foo2) {
     bar2;
}
```

krizhanovsky (changes requested):
There are still 2 things to fix from the previous review:
1. https://github.com/tempesta-tech/tempesta/pull/2144#discussion_r1666040181
2. CamelCase structures naming

kingluo:
> 2. CamelCase structures naming

The script allows CamelCase by default. So you mean we should have a CamelCase whitelist?

scripts/checkpatch.pl
kingluo: It seems a bit complicated to fix. I'll try it anyway.

krizhanovsky:
I mean that we need to enforce that structures are named with CamelCase, i'd imagine something like:
1. require the first letter is Capital
2. no `_` in the name

krizhanovsky (approved):
LGTM

PR #2146, opened 2024-06-20
Fix BUG_ON in tls.c

EvgeniiMekhanik:
ttls_rcv should not return T_BLOCK or T_DROP
because T_BLOCK s error code for high level
modules (like frang), and T_DROP is error
code, which is returned when connection
should be alive, but if we can't decrypt
request, we should close the connection.


fw/tls.c
```diff
@@ -957,15 +957,16 @@ tfw_tls_over(TlsCtx *tls, int state)
...
 	if (state == TTLS_HS_CB_FINISHED_NEW
 	    || state == TTLS_HS_CB_FINISHED_RESUMED)
 		TFW_INC_STAT_BH(serv.tls_hs_successful);
 
 	if (TFW_FSM_TYPE(sk_proto) == TFW_FSM_H2 &&
-	    tfw_h2_context_init(tfw_h2_context_unsafe(conn))) {
+	    ((r = tfw_h2_context_init(tfw_h2_context_unsafe(conn))))) {
 		    T_ERR("cannot establish a new h2 connection\n");
-		    return T_DROP;
+		    return r;
```
const-t, line 969: Just wandering why we can't do simply `return T_BAD`, it anyway will be "casted" to T_BAD in `tfw_tls_connection_recv()`?

const-t (approved):
LGTM

fw/tls.c
EvgeniiMekhanik: It seems that it's better to return DROP_WIRH_FIN for all types of internal tls errors if (tls is noit finished). There is no sense to send tcp_shutdown if there is no pending error responses.

Issue #1988, opened 2023-10-18
Health monitor server status


const-t:
# Scope

Looks strange following condition in `tfw_http_hm_control()` if we take into account that `tfw_apm_hm_srv_alive()` must be called for alive server at [least once](https://github.com/tempesta-tech/tempesta/wiki/Health-monitor) to calculate valid crc32 in case when directive `resp_crc32` is `auto`:

```
if (!tfw_srv_suspended(srv) ||
	    !tfw_apm_hm_srv_alive(resp->status, &resp->body, srv->apmref))
```

[Discussion in PR](https://github.com/tempesta-tech/tempesta/pull/1983#issuecomment-1738980934)

Need to find out is there a bug or not and develop functional test for this case.

enuribekov-tempesta:
It looks code itself is correct, but number of issues was found around this code:
typos, bugs, unused function arguments etc.

enuribekov-tempesta:
Created issue for bug found during testing HM.

https://github.com/tempesta-tech/tempesta/issues/2128

enuribekov-tempesta:
Sometime could't receive correct answer from monitored server.
The reason  was https://github.com/tempesta-tech/tempesta-test/issues/625
Reworked test for using nginx and got correct answer.

enuribekov-tempesta:
`tfw_http_hm_suspend()` worked incorrectly, reworked.

enuribekov-tempesta:
In a function `tfw_apm_hm_srv_alive()` `hm->codes bitmap` is empty, so always return error during comparation with any real status code.

enuribekov-tempesta:
It looks like instead of saving `server_failover_http`  configuration we got memory leak.
https://github.com/tempesta-tech/tempesta/issues/2130

krizhanovsky:
fixed in #2151

PR #2151, opened 2024-06-26
En/fix 1988 1

enuribekov-tempesta:
New pull request for #1988 made from the scratch.

### DIFF
--- a/fw/apm.c
+++ b/fw/apm.c
@@ -1385,12 +1385,6 @@ tfw_apm_hm_srv_alive(TfwHttpResp *resp, TfwServer *srv)
 	 */
 	if (!hm->crc32 && hm->auto_crc) {
 		hm->crc32 = __tfw_apm_crc32_calc(&it, &chunk);
-		/*This warinig is used by tests to check crc generation
-		 * '\n' is neseccary for flushing message to logger,
-		 * without it tests will always fail.  
-		 */
-		T_WARN("Auto CRC generated\n");
-
 	} else if (hm->crc32) {
 		crc32 = __tfw_apm_crc32_calc(&it, &chunk);
 		if (hm->crc32 != crc32)
--- a/fw/apm.c
+++ b/fw/apm.c
@@ -1346,7 +1346,7 @@ __tfw_apm_crc32_calc(TfwMsgIter *it, TfwStr *chunk)
 * - Integrity of the response body
 * - CRC
 * Successful passing all of the checks considered
-* as a sign that server alive.
+* as a sign that the server is alive.
 */
 bool
 tfw_apm_hm_srv_alive(TfwHttpResp *resp, TfwServer *srv)
--- a/fw/apm.c
+++ b/fw/apm.c
@@ -1376,6 +1376,8 @@ tfw_apm_hm_srv_alive(int status, TfwStr *body, struct sk_buff *skb_head,
 	 */
 	if (!hm->crc32 && hm->auto_crc) {
 		hm->crc32 = __tfw_apm_crc32_calc(&it, &chunk, skb_head, body);
+		T_WARN("Auto CRC generated");
+
 	} else if (hm->crc32) {
 		crc32 = __tfw_apm_crc32_calc(&it, &chunk, skb_head, body);
 		if (hm->crc32 != crc32)
@@ -1704,7 +1706,7 @@ tfw_apm_create_def_hm(void)
 }
 
 static TfwApmHMCfg *
-tfw_amp_create_hm_entry(void)
+tfw_apm_create_hm_entry(void)
 {
 	TfwApmHMCfg *hm_entry = kzalloc(sizeof(TfwApmHMCfg), GFP_KERNEL);
 	if (!hm_entry)
@@ -1725,7 +1727,7 @@ tfw_apm_create_def_health_stat_srv(void)
 	if (tfw_hm_cfg_200_created)
 		return 0;
 
-	hm_entry = tfw_amp_create_hm_entry();
+	hm_entry = tfw_apm_create_hm_entry();
 	if (!hm_entry)
 		return -ENOMEM;
 
@@ -1901,7 +1903,7 @@ tfw_cfgop_apm_server_failover(TfwCfgSpec *cs, TfwCfgEntry *ce)
 	if (tfw_cfg_check_range(tframe, 1, USHRT_MAX))
 		return -EINVAL;
 
-	hm_entry = tfw_amp_create_hm_entry();
+	hm_entry = tfw_apm_create_hm_entry();
 	if (!hm_entry)
 		return -ENOMEM;
 	tfw_hm_entry_set_code(hm_entry, code);
@@ -1928,7 +1930,7 @@ tfw_cfgop_apm_health_stat_srv(TfwCfgSpec *cs, TfwCfgEntry *ce)
 				 val);
 			return -EINVAL;
 		}
-		hm_entry = tfw_amp_create_hm_entry();
+		hm_entry = tfw_apm_create_hm_entry();
 		if (!hm_entry)
 			return -ENOMEM;
 		tfw_hm_entry_set_code(hm_entry, code);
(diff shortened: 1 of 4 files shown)


fw/apm.c
```diff
@@ -1375,9 +1384,15 @@ tfw_apm_hm_srv_alive(int status, TfwStr *body, struct sk_buff *skb_head,
 	 * from body of first response and store it into monitor.
 	 */
 	if (!hm->crc32 && hm->auto_crc) {
-		hm->crc32 = __tfw_apm_crc32_calc(&it, &chunk, skb_head, body);
+		hm->crc32 = __tfw_apm_crc32_calc(&it, &chunk);
+		/*This warinig is used by tests to check crc generation
+		 * '\n' is neseccary for flushing message to logger,
+		 * without it tests will always fail.  
+		 */
+		T_WARN("Auto CRC generated\n");
```
const-t, line 1392: Is it necessary to have this WARN only for testing? Can we verify correctness of the code relying on state of the backend? For example: If backend was dead and then it became alive, it means all works. Then do negative tests with wrong crc where expected result is dead server.

enuribekov-tempesta: May be it is not the best way to test it, but overhead looks as reasonable. We get the single warning message at the moment when HM start working. The message explicitly indicate CRC generation.
In case of initially dead backend we:
1. Add delay before test (@RomanBelozerov  asked me to minimize test execution time, as it possible)
2. Change single warning to backend on/off  warnings. 
3. Indicate proper works of subsystem at all, but not CRC issues. Using framework we can check only backend CRC, which is not our responsibility.

enuribekov-tempesta: @const-t
I though a lot about it and maybe make sense to introduce separate logging macros which will be active only for CI builds and print out information only for test framework?

const-t: >  I though a lot about it and maybe make sense to introduce separate logging macros which will be active only for CI builds and print out information only for test framework?

I would like to suggest to use [dynamic debug](https://www.kernel.org/doc/html/v4.19/admin-guide/dynamic-debug-howto.html) for such debug prints. I've used it and it works perfect. @krizhanovsky  @RomanBelozerov Guys, what do you think?

krizhanovsky: I'm strongly against the warning used for testing only. The most bad thing is that this is *warning* for absolutely normal cast to compute CRC, even not an informational message, - it will flood the user logs and the user won't be able to figure out what's wrong. This is absolute not go.

Probably, it's it's to hard to test in automatic way, we can just test it manually. But I don't see why it's impossible to implement the test in Python. @RomanBelozerov seems some help from your side is required here

fw/apm.c
```diff
@@ -1340,11 +1339,21 @@ __tfw_apm_crc32_calc(TfwMsgIter *it, TfwStr *chunk , struct sk_buff *skb_head,
...
 }
 
+/**
+* Validate response from the given server.
+* Check:
+* - Is response status code belongs to monitored set
+* - Integrity of the response body
+* - CRC
+* Successful passing all of the checks considered
+* as a sign that server alive.
```
krizhanovsky, line 1349:
```suggestion
* as a sign that the server is alive.
```

fw/http.c
```diff
@@ -1615,54 +1615,62 @@ do {									\
...
 		return;
 	}
 
-	tfw_srv_mark_alive(srv);
+	if(lim_exceeded) {
+		T_WARN("Server error limit exceeded");
+		tfw_http_hm_try_suspend(resp, srv);
+	}
+
+	if(tfw_apm_hm_srv_alive(resp, srv)) {
```
krizhanovsky, line 1670:
```suggestion
	if (tfw_apm_hm_srv_alive(resp, srv)) {
```

fw/http.c
```diff
@@ -1615,54 +1615,62 @@ do {									\
...
-	    !tfw_apm_hm_srv_alive(resp->status, &resp->body, resp->msg.skb_head,
-				  srv->apmref))
-	{
+	if (tfw_srv_suspended(srv)) {
+		T_DBG2("Server suspended");
 		return;
 	}
 
-	tfw_srv_mark_alive(srv);
+	if(lim_exceeded) {
```
krizhanovsky, line 1665:
```suggestion
	if (lim_exceeded) {
```

fw/http.c
```diff
@@ -1615,54 +1615,62 @@ do {									\
...
 
-	return true;
+		flags = old_flags;
+	}
 }
 
+/**
+* The main function of Health Monioring.
+* Getting response from the server, it updates responses statistics,
+* checks HM limits and makes solution about server health.
```
krizhanovsky, line 1647:
```suggestion
* checks HM limits and makes a decision about server health.
```
BTW ChatGPT amazingly proofreads text and explains the fixes, so you can write better text and learn why

fw/http.c
```diff
@@ -1615,54 +1615,62 @@ do {									\
...
-				  srv->apmref))
-	{
+	if (tfw_srv_suspended(srv)) {
+		T_DBG2("Server suspended");
 		return;
 	}
 
-	tfw_srv_mark_alive(srv);
+	if(lim_exceeded) {
+		T_WARN("Server error limit exceeded");
```
krizhanovsky, line 1666: I appreciate this message - it provides important information for a system administrator about the status of the backend servers. However, could you please add the server IP address since it could be hard to identify which server fails (a typical CDN proxy services thousands of upstream servers).

fw/http.c
```diff
@@ -1615,54 +1615,62 @@ do {									\
...
+		tfw_http_hm_try_suspend(resp, srv);
+	}
+
+	if(tfw_apm_hm_srv_alive(resp, srv)) {
+		T_DBG2("Mark server alive");
+		tfw_srv_mark_alive(srv);
```
krizhanovsky, line 1672: The function must be renamed or commented here: it's very confusing to see a function call with name 'mark_alive' right after thec all 'try_suspend'

krizhanovsky (changes requested):
Nothing serious, but bunch of cleanups is required. Probably it's more serious to fix the tests in https://github.com/tempesta-tech/tempesta-test/pull/637 w/o using the warning

fw/http.c
enuribekov-tempesta: I prefer Reverso

fw/http.c
enuribekov-tempesta: `tfw_http_hm_try_suspend()`  is the conditional function and has following comment:
/*
 * Try to mark server as suspended.
 * In case of HM is active do it, otherwise left unchanged.
 */
`tfw_srv_mark_alive()` is  just inline for clearing flag

fw/http.c
```diff
@@ -1615,54 +1615,66 @@ do {									\
...
-	if (!(flags & TFW_SRV_F_HMONITOR))
-		return true;
-	if (!lim_exceeded)
-		return false;
+	while (flags & TFW_SRV_F_HMONITOR) {
 
-	do {
 		old_flags = cmpxchg(&srv->flags, flags,
-				    flags | TFW_SRV_F_SUSPEND);
+				flags | TFW_SRV_F_SUSPEND);
```
krizhanovsky, line 1630: Previous alignment was good

krizhanovsky (approved):
LGTM

fw/http.c
const-t: The logic is not clear for me. We trying to suspend server using `tfw_http_hm_try_suspend` even in case if server was suspended in previous call we checking is it alive `tfw_apm_hm_srv_alive`. Why?

const-t (approved):

fw/http.c
```diff
@@ -1615,54 +1615,66 @@ do {									\
...
 {
 	TfwServer *srv = (TfwServer *)resp->conn->peer;
 
-	if (tfw_http_hm_suspend(resp, srv))
-		return;
+	/*
+	* Total response statistics is counted permanently regardless
+	* of the state of the health monitor.
+	*/
+	bool lim_exceeded = tfw_apm_hm_srv_limit(resp->status, srv->apmref);
```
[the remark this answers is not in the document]
enuribekov-tempesta: Yes, because these counters used for statistics too.
Perhaps this can be done much more elegantly, but I have kept the existing logic because statistics go beyond that.

PR #2153, opened 2024-06-28
Fix finding place to insertion frame header

EvgeniiMekhanik:
Fix `ss_skb_data_ptr_by_offset` function, return
pointer to data if offset is equal to skb->len.
Also free empty skbs before insertion frame header, not before pushing to socket write queue.


krizhanovsky (approved):
LGTM

const-t (approved):
LGTM

PR #2158, opened 2024-07-01
Fix crash in `tls_on_send_alert`.

EvgeniiMekhanik:
We should not set `tls_on_send_alert` callback for HTTP1.


krizhanovsky (approved):
LGTM. I also tested the branch with `curl` and the testing framework and can not reproduce the crash

const-t (approved):
LGTM. The same, can't reproduce.

PR #2159, opened 2024-07-01
Add check for dublicate options in config

EvgeniiMekhanik:
Check that admin can't set dublicate options
in config (max_miss, cookie_name and so on).

### DIFF
--- a/fw/cfg.h
+++ b/fw/cfg.h
@@ -201,13 +201,20 @@ typedef struct {
 		return -EINVAL;						\
 	}
 
-#define TFW_CFG_CHECK_VAL_N(op, req, spec, entry)				\
+#define TFW_CFG_CHECK_VAL_N(op, req, spec, entry)			\
 	if (! ((entry)->val_n op (req)) ) {				\
 		T_ERR_NL("%s: Invalid number of arguments: %zu, must "	\
 			 "be %s %d\n", (spec)->name, (entry)->val_n,	\
 			 #op, (req));					\
 		return -EINVAL;						\
 	}
+
+#define TFW_CFG_CHECK_VAL_DUP(name, val_was_set, code)		        \
+	if (val_was_set) {						\
+		T_ERR_NL("Duplicate argument: '%s'\n", key);		\
+		code;							\
+	}								\
+	val_was_set = true;
 
 /**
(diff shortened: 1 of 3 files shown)


fw/http_sess_conf.c
```diff
@@ -371,10 +380,22 @@ tfw_cfgop_cookie_options_set(TfwCfgSpec *cs, TfwCfgEntry *ce)
 
 	TFW_CFG_ENTRY_FOR_EACH_ATTR(ce, i, key, val) {
 		if (!strcasecmp(key, "Path")) {
+			if (was_path) {
+				T_ERR_NL("Duplicate argument: '%s'\n", key);
+				return -EINVAL;
+			}
```
keshonok, line 386: This repetitive code pattern clearly asks for a macro that includes the check **and** the assignment to `true`. There are places where the assignment is done after additional piece of code such as parsing of the value. But if that parsing fails it won't matter anyway, so the assignment to `true` can be done earlier in the macro.

biathlon3 (approved):
LGTM

fw/http_sess_conf.c
EvgeniiMekhanik: Yes good point! Fixed

fw/cfg.h
```diff
@@ -201,14 +201,21 @@ typedef struct {
...
-#define TFW_CFG_CHECK_VAL_N(op, req, spec, entry)				\
+#define TFW_CFG_CHECK_VAL_N(op, req, spec, entry)			\
 	if (! ((entry)->val_n op (req)) ) {				\
 		T_ERR_NL("%s: Invalid number of arguments: %zu, must "	\
 			 "be %s %d\n", (spec)->name, (entry)->val_n,	\
 			 #op, (req));					\
 		return -EINVAL;						\
 	}
 
+#define TFW_CFG_CHECK_VAL_DUBLICATE(name, val_was_set, code)		\
```
keshonok, line 212: While spelling `dublicate` exists, it's very rare. `Duplicate` is the right word. Also, perhaps, the standard shortcut would suffice here for a shorter name? `TFW_CFG_CHECK_VAL_DUP`.

EvgeniiMekhanik: Yes fixed

PR #2160, opened 2024-07-03
Rework allocation for stream schedulers.

EvgeniiMekhanik:
According to our discussion the best way is to allocate two pages for each connection and save streams schedulers after the end of `struct TfwH2Conn`. When we create new stream, we get scheduler for such stream from the list of preallocated schedulers. If limit of such schedulers is exceeded we allocate new page and use it to initialize new schedulers.

### DIFF
--- a/fw/http2.c
+++ b/fw/http2.c
@@ -204,6 +204,85 @@ tfw_h2_cleanup(void)
 	tfw_h2_stream_cache_destroy();
 }
 
+static inline void
+tfw_h2_context_init_sched_enry_storage(TfwH2Ctx *ctx)
+{
+	TfwH2Conn *conn = container_of(ctx, TfwH2Conn, h2);
+	unsigned int def_sched_entry_count =
+		(2 * PAGE_SIZE - sizeof(TfwH2Conn)) / sizeof(TfwStreamSchedEntry);
+	void *base = (void *)((char *)conn + sizeof(TfwH2Conn));
+
+	/*
+	 * We allocate 2* PAGE_SIZE bytes for each TfwH2Conn, and
+	 * use free memory after TfwH2Conn structure for sched
+	 * entries.
+	 */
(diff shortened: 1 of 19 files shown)


EvgeniiMekhanik:
I think about add space for the whole TfwStream structure at the end of http2 context but it has 912B size, so we need to much memory for this purpose.

fw/http2.c
```diff
@@ -204,14 +204,110 @@ tfw_h2_cleanup(void)
...
+	tfw_h2_stream_sched_init_entry_storage(&ctx->sched, base,
+					       def_sched_entry_count);
+}
+
+TfwH2Ctx *
+tfw_h2_context_alloc(void)
+{
+	struct page *pg;
+
+	pg = alloc_pages(GFP_ATOMIC, 1);
```
const-t, line 228: Is this not too much just for H2 context?

I've a doubts about page allocation at this place. I would prefer to use `TfwPool` or memcache.

fw/http_stream_sched.h
```diff
@@ -27,18 +27,22 @@
...
  */ 
 typedef struct tfw_stream_sched_entry_t {
 	u64				total_weight;
 	long int			active_cnt;
-	struct tfw_stream_sched_entry_t	*parent;
+	TfwStream 			*owner;
+	struct tfw_stream_sched_entry_t *parent;
 	struct eb_root			active;
 	struct eb_root			blocked;
+	struct list_head		in_empty_list;
```
const-t, line 45: Could we get rid off `in_empty_list` if we will store the pointer to the next entry in the current entry? I mean that while `TfwStreamSchedEntry` in the freelist it casted to `void *` that points to the next `TfwStreamSchedEntry` when we remove it from the list, we zeroes the pointer and cast to `TfwStreamSched` again. In other words when `TfwStreamSchedEntry` in empty list it's casted to linked list.

const-t (changes requested):

PR #2161, opened 2024-07-05
Implementation multi pattern regular expressions

### LINKED ISSUE #496: Multi-pattern regular expressions
TL (#102) must implement multi-pattern regular expressions to efficiently handle HTTP matching rules for filtering and configuration (see for example #471 and #495). [Intel HyperScan](https://github.com/01org/hyperscan) can be used as reference or foundation for the feature.

biathlon3:
To compile regexp **required** installation of hscollider. For simplicity it built as debian package.

Add repository
```
sudo install -m 0755 -d /etc/apt/keyrings
sudo wget -O /etc/apt/keyrings/tempesta.asc https://tempesta-tech.com:8081/repository/public-keys/keys/public.key
sudo chmod a+r /etc/apt/keyrings/tempesta.asc
echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/tempesta.asc] \
https://tempesta-tech.com:8081/repository/tempesta/ focal main" | sudo tee /etc/apt/sources.list.d/tempesta.list
sudo apt update && sudo apt upgrade
```

Install the package
```
apt install linux-regex
```


biathlon3:
For deployment see file install.txt. Description will be expanded.

biathlon3:
This PR works on Kernel 6.8.9

Description of using hscollider - http://intel.github.io/hyperscan/dev-reference/tools.html

Description of regular expression syntax - https://perldoc.perl.org/perlre
It is supported in Hyperscan not fully, constraints described here - http://intel.github.io/hyperscan/dev-reference/compilation.html

Now regex implemented in tempasta config as fallows: 
for locations, added keyword "regex" and regex started with "^";
for httptables, regex started with "^".

for example:
```
location regex "^/new/" {
    frang_limits {
	http_body_len 5;
	http_strict_host_checking true;
    }
}

http_chain {
  uri == "^/html/" -> default;
  -> default;
}
```

biathlon3:
I squashed it into one commit so it can be reviewed.

const-t:
Description of using hscollider - http://intel.github.io/hyperscan/dev-reference/tools.html

Description of regular expression syntax - https://perldoc.perl.org/perlre
It is supported in Hyperscan not fully, constraints described here - http://intel.github.io/hyperscan/dev-reference/compilation.html

Use `~` for regexp in location and http chain:
```
location ~ "/^/show/id/[0-9]+/i" {
    frang_limits {
	http_body_len 5;
	http_strict_host_checking true;
    }
}

http_chain {
  uri ~  "^/html/" -> default;
  -> default;
}
```

EvgeniiMekhanik:
I catch memory leak in case when configuraion fails

kmemleak_alloc+0x4a/0x90
    __kmalloc_noprof+0x3a5/0x520
    tfw_kzalloc+0x14/0x20 [tempesta_lib]
    rex_make_item+0x27/0x100 [xdp_rex]
    configfs_mkdir+0x735/0xd30
    vfs_mkdir+0x45c/0x7a0
    do_mkdirat+0x292/0x320
    __x64_sys_mkdir+0xd7/0x130
    x64_sys_call+0x10cf/0x25f0
    do_syscall_64+0x7e/0x170
    entry_SYSCALL_64_after_hwframe+0x76/0x7e

config:

listen 80;
listen 443 proto=h2;
        
tls_certificate /home/evgeny/workdir/certs/tempesta-tech.test.crt;
tls_certificate_key /home/evgeny/workdir/certs/tempesta-tech.test.key;
tls_match_any_server_name;

access_log off;

srv_group grp1 {
    server 127.0.0.1:8000;
}

vhost test {
    proxy_pass grp1;
    location ~ "/^/article/id/[0-9]+/" {
        proxy_pass grp1;
    }
    location ~ "/shop/" {
        proxy_pass grp1;
    }
    location ~ "/wiki/" {
        proxy_pass grp1;
    }
}

vhost work {
    proxy_pass grp1;
    location ~ "/shop/i" {
        proxy_pass grp1;
    }
    location ~ "/wiki/" {
        proxy_pass grp1;
    }
}

http_chain {
    host ~ "/test/i" -> test;
    host ~ "/work/i" -> work;
    host ~ "/(a+)+$$/" -> work;
    -> block;
}

fw/vhost.c
```diff
@@ -76,6 +77,10 @@ static const TfwCfgEnum tfw_match_enum[] = {
 	{ "eq",		TFW_HTTP_MATCH_O_EQ },
 	{ "prefix",	TFW_HTTP_MATCH_O_PREFIX },
 	{ "suffix",	TFW_HTTP_MATCH_O_SUFFIX },
+        /*regex case sensitive*/
+        { "regex",	TFW_HTTP_MATCH_O_REGEX },
```
EvgeniiMekhanik: broken aligment

fw/vhost.c
```diff
@@ -192,6 +205,8 @@ static const __tfw_match_fn __tfw_match_fn_tbl[] = {
 	[TFW_HTTP_MATCH_O_EQ]		= __tfw_match_eq,
 	[TFW_HTTP_MATCH_O_PREFIX]	= __tfw_match_prefix,
 	[TFW_HTTP_MATCH_O_SUFFIX]	= __tfw_match_suffix,
+        [TFW_HTTP_MATCH_O_REGEX]	= __tfw_match_regex,
```
EvgeniiMekhanik: broken alignment

regex/Makefile
```diff
@@ -0,0 +1,79 @@
...
+
+CC_FLAGS_HSRUNTIME	:= -isystem $(shell $(CC) -print-file-name=include)
+CC_FLAGS_REMOVE_SIMD	:= -mno-80387 -mno-fp-ret-in-387 -mno-sse -mno-mmx -mno-sse2 -mno-3dnow -mno-avx
+CC_FLAGS_HSRUNTIME	+= -DHAVE_SSE2
+CC_FLAGS_SIMD		:= -msse4.2 -msse4.1
+CC_FLAGS_HSRUNTIME	+= -DHAVE_SSE41 -DHAVE_SSE42
+CC_FLAGS_SIMD		+= -mavx -mavx2
+CC_FLAGS_HSRUNTIME	+= -DHAVE_AVX -DHAVE_AVX2
+CC_FLAGS_SIMD		+= -mavx512f -mavx512cd -mavx512bw -mavx512vl -mavx512vnni
+CC_FLAGS_HSRUNTIME	+= -DHAVE_AVX512
```
EvgeniiMekhanik, line 30: We should check that AVX512 is present. I catch BUG when I compile it with this flag, but without real avx512 support

scripts/tempesta.sh
```diff
@@ -113,7 +113,7 @@ templater()
 	mkdir $TFW_ROOT/etc 2>/dev/null
 	while IFS= read -r raw_line
 	do
-		line=$(echo "$raw_line" | sed -e '/request /s/\\r\\n/\x0d\x0a/g')
+		line=$(echo "$raw_line" | sed -e '/request /s/\\r\\n/\x0d\x0a/g' | sed -e 's/^[[:space:]]*//g')
 		if [[ ${line:0:1} = \# ]]; then
 			:
 		elif [[ $line =~ '!include' ]]; then
```
EvgeniiMekhanik: I think that also it's better to use `pattern='^!include[[:space:]]+[^[:space:]]+'`
and then
elif [[ $line =~ $pattern' ]]; then

to skip patterns like ~inlcudeqqqq

fw/vhost.c
```diff
@@ -1528,23 +1532,20 @@ tfw_location_init(TfwLocation *loc, tfw_match_t op, const char *arg,
...
 		loc->mod_hdrs[TFW_VHOST_HDRMOD_REQ].hdrs + TFW_USRHDRS_ARRAY_SZ;
 
-	switch (op) {
-	case TFW_HTTP_MATCH_O_REGEX:
+	if (op == TFW_HTTP_MATCH_O_REGEX) {
 		int r;
-		unsigned short regex_idx;
 
-		if ((r = tfw_write_regex(arg, &regex_idx))) {
+		r = tfw_write_regex(arg, &regex_idx);
```
EvgeniiMekhanik, line 1538: Please look comment in `tfw_write_regex`. It seems that comment before `if (len < sizeof(number_of_regex))` is incorrect. Because here we pass `arg`, not `argmem` here

fw/http_match.c
```diff
@@ -827,6 +827,11 @@ tfw_http_arg_adjust(const char *arg, tfw_http_match_fld_t field,
 	if (wc_arg && !raw_hdr_name)
 		return NULL;
 
+	if (wc_arg && regex) {
```
EvgeniiMekhanik: host ~ "*" -> test -> block still work here, because raw_hdr_name is NULL, is it ok?

Makefile
```diff
@@ -184,7 +184,10 @@ clean:
 	find . \( -name \*~ -o -name \*.orig -o -name \*.symvers \) \
 		-exec rm -f {} \;
 
+install:
```
EvgeniiMekhanik: I am not sure, that it is a good idea. We don't install tempesta_fw.conf to etc directory for example. I think it's better to pass path to this script using bash variable. And by default i think it's better to set default value in ./scripts folder

scripts/regex_setup.sh
```diff
@@ -0,0 +1,76 @@
...
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 59
+# Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+
+regex_dir_path=${REGEX_DIR_PATH:="/opt/tempesta/regex"}
```
EvgeniiMekhanik, line 20: It seems that you set `regex_dir_path` in regex.c without any chance to change it.

fw/regex.c
```diff
@@ -0,0 +1,213 @@
...
+		NULL };
+
+	snprintf(regex_path, sizeof(regex_path), "%s%s", regex_dir_var_name,
+		 regex_dir_path);
+
+	envp[3] = regex_path;
+
+	T_LOG_NL("Compiling regex\n");
+	/* Execute helper script. */
+	ret = call_usermodehelper("/bin/bash", argv, envp, UMH_WAIT_PROC);
```
EvgeniiMekhanik, line 201: Why it is better then call this script after Tempesta FW will be load? I am not sure that it is better then call it directly from tempesta.sh

fw/http_match.c
```diff
@@ -883,14 +883,37 @@ write_regex(const char *arg, int regex)
 	return r;
 }
 
+static void
+find_spaces(const char *s, size_t len, size_t *begin_spaces_out,
```
EvgeniiMekhanik: strange return values for this function. For string like `ddddddd` :
begin_spaces_out == 0 and
end_spaces_out == 0

regex/kmod/rex.c
```diff
@@ -243,64 +190,46 @@ int bpf_scan_tfwstr(const TfwStr *str,
...
-{
-	u32 size = xdp->data_end - xdp->data;
-	void *addr = xdp->data;
-
-	if (unlikely(offset > 0xffff || len > 0xffff))
-		return ERR_PTR(-EFAULT);
-
-	if (offset + len > size)
-		return ERR_PTR(-EINVAL);
+EXPORT_SYMBOL(bpf_scan_bytes);
```
EvgeniiMekhanik, line 183: This function never used in our code. We're just returning the code here to its original state, like in the repository we got it from, right?

fw/http_msg.h
```diff
@@ -25,6 +25,7 @@
 
 #define S_CRLF			"\r\n"
 #define S_DLM			": "
+#define S_COLON			":"
```
EvgeniiMekhanik, line 28: Can you explain please the changes with S_COLON and find spaces. How it worked before and why we change this behavior?

fw/http_match.c
```diff
@@ -835,7 +905,13 @@ tfw_http_arg_adjust(const char *arg, tfw_http_match_fld_t field,
 
 	if (raw_hdr_name && field != TFW_HTTP_MATCH_F_COOKIE) {
 		name_len = strlen(raw_hdr_name);
-		full_name_len = name_len + SLEN(S_DLM);
+		full_name_len = name_len + SLEN(S_COLON);
```
EvgeniiMekhanik, line 852: Why `full_name_len` is not initialized for TFW_HTTP_MATCH_F_COOKIE

EvgeniiMekhanik (changes requested):

EvgeniiMekhanik:
I also can't understand
if I add
hdr cookie ~ "/fff*/" -> block; in http_chain should request with  ("cookie", "fffoooo=bbb"), blocked?

fw/http_match.c
const-t: I'm not sure, but I think it is legacy code. This condition `field != TFW_HTTP_MATCH_F_COOKIE` actually always true, cookie is special header.

fw/http_match.c
const-t: No, it is not ok. Such regex also not allowed for special headers. Fixed

fw/vhost.c
const-t: Fixed comment and changed condition

fw/http_match.c
const-t: This is correct, but I've changed arguments name and added comment.

fw/http_match.c
const-t: I read it carefully and this condition is the part of the code that do matching using `cookie "name" == "value"` where name and value are parameters of cookie. In this case `raw_hdr_name` contains cookie parameter name instead of header name.

regex/Makefile
const-t: I've added the check

scripts/tempesta.sh
const-t: fixed

Makefile
const-t: Agreed, install target isn't good. Now I pass it via module parameter.

scripts/regex_setup.sh
const-t: REGEX_DIR_PATH is env var, tempesta.sh passes it to kernel and then kernel sets is when calls helper

fw/regex.c
const-t: When we calling `setup_script.sh` from tempesta.sh we do this only when all modules loaded and Tempesta ready to receive connection, but there is interval between tempesta is ready and regex is compiled. If Tempesta receives a request during this interval, we can’t apply the regex because it hasn’t been compiled yet.

regex/kmod/rex.c
const-t: Yes, you are right

fw/http_msg.h
const-t: Previously we stored a patter for header match with space after colon, after the patch we save the pattern without spaces. We do that to not skip spaces in pattern during header comparison, especially for http2, where spaces are forbidden.

scripts/tempesta.sh
```diff
@@ -114,7 +118,7 @@ templater()
 		line=$(echo "$raw_line" | sed -e '/request /s/\\r\\n/\x0d\x0a/g')
 		if [[ ${line:0:1} = \# ]]; then
 			:
-		elif [[ $line =~ '!include' ]]; then
+		elif [[ $line =~ ^[[:space:]]*!include[[:space:]]+.+$ ]]; then
```
EvgeniiMekhanik, line 122: Is it ok that we can have any count of space before include?

fw/t/unit/regex.c
```diff
@@ -0,0 +1 @@
+../../regex.c
```
EvgeniiMekhanik, line 1: No empty line at the end of the file?

EvgeniiMekhanik (approved):
Please also add three stress tests with reconfiguration with regexp:
- With success reconfiguration regexp config - regexp config
- Fail reconfig
- Success was regexp now not regexp

fw/http_match.c
```diff
@@ -84,6 +85,7 @@ map_op_to_str_eq_flags(tfw_http_match_op_t op)
 		[TFW_HTTP_MATCH_O_EQ]		= TFW_STR_EQ_DEFAULT,
 		[TFW_HTTP_MATCH_O_PREFIX]	= TFW_STR_EQ_PREFIX,
 		[TFW_HTTP_MATCH_O_SUFFIX]	= TFW_STR_EQ_DEFAULT,
+		[TFW_HTTP_MATCH_O_REGEX]	= TFW_STR_EQ_REGEX
```
krizhanovsky, line 88: Please add some info to the HTTP tables wiki, make good examples

fw/http_tbl.c
krizhanovsky, line 365: Add `~` to the operation description in the comment

fw/http_match.c
```diff
@@ -819,6 +819,11 @@ tfw_http_arg_adjust(const char *arg, tfw_http_match_fld_t field,
 
 	*type_out = tfw_http_tbl_arg_type(field);
 
+	if (wc_arg && regex) {
+		T_ERR_NL("http_match: use simple wildcard argument: hdr == *. Instead regex.\n");
+		return ERR_PTR(-EINVAL);
```
krizhanovsky, line 824:
```suggestion
		T_ERR_NL("http_match: use simple wildcard argument `hdr == *`"
		         " instead of regex.\n");
```
By the way

fw/vhost.c
```diff
@@ -1421,6 +1421,11 @@ tfw_cfgop_out_cache_control_ignore(TfwCfgSpec *cs, TfwCfgEntry *ce)
 /*
  * Find a location directive entry. The entry is looked up
  * in the array that holds all location directives.
+ *
+ * NOTE: This works only for non-regex locations. Regex locations may have the
+ * same pattern. It seems not worth trying to detect duplicate patterns,
+ * because two similar but not identical patterns can still resolve to the same
+ * location.
```
krizhanovsky, line 1428: Please add regular expressions for `location` to wiki

regex/kmod/rex.c
krizhanovsky, line 76: Please add our copyrights to all changed files like regex/kmod/rex.c

fw/t/unit/test_http_match.c
```diff
@@ -913,7 +1010,9 @@ TEST_SUITE(http_match)
 	TEST_RUN(http_match, hdr_host_prefix);
 	TEST_RUN(http_match, hdr_host_suffix);
 	TEST_RUN(http_match, raw_header_eq);
+	TEST_RUN(http_match, raw_header_eq_chunked);
 	TEST_RUN(http_match, raw_header_eq_ws);
+	TEST_RUN(http_match, raw_header_eq_nows);
 	TEST_RUN(http_match, method_eq);
 	TEST_RUN(http_match, cookie);
 	TEST_RUN(http_match, choose_host);
```
krizhanovsky, line 1018: Could you please add unit tests for regular expressions

install.txt
```diff
@@ -0,0 +1,42 @@
...
+
+cmake ./
+make
+
+after compilation
+copy hscollider from /linux-regex-module/bin/ to /tempesta/scripts/ or default app directory???
+
+git clone https://github.com/tempesta-tech/tempesta.git
+cd tempesta
+git checkout ag_Multi-pattern-regular-expressions
```
krizhanovsky, line 42: I think this should be README.md in the `regex` directory and reference the original https://github.com/G-Core/linux-regex-module to give credits to the team. Probably this wiki https://github.com/tempesta-tech/tempesta/wiki/Installation-tempesta-supported-regex should be in the README - it's useful for us, but not for users (can we remove the wiki page after that?)

scripts/tempesta.sh
```diff
@@ -58,6 +61,7 @@ lib_mod=tempesta_lib
 tls_mod=tempesta_tls
 tdb_mod=tempesta_db
 tfw_mod=tempesta_fw
+rgx_mod=xdp_rex
```
krizhanovsky, line 64: That project was for the XDP module and technically could be useful in https://github.com/tempesta-tech/xFWas , but anyway the regexps, especially for Tempesta FW, have nothing to do about XDP. Let's rename the module to something like `tempesta_regex` that an administrator can easily see all our modules

fw/http_match.c
```diff
@@ -767,7 +767,7 @@ write_regex(const char *arg, int regex)
 	}
 
 	++number_of_db_regex;
-	sprintf(file_name, "/tmp/tempesta/%u.txt", number_of_db_regex);
+	sprintf(file_name, "/opt/tempesta/regex/%u.txt", number_of_db_regex);
```
krizhanovsky, line 770: We should get the path from the script: this path is likely to change, IIRC @symstu recently made a change in Escudo that all the products are installed in `/opt/tempesta` and have their own subdirs, `fw`, `xfw`, `webshield`. We can do this with sysctl or module parameter, but this should be a generic home/work directory specification

regex/kmod/rex.c
```diff
@@ -91,12 +90,12 @@ static int rex_scan_cb(unsigned int expression, unsigned long long from,
 	return (features & REX_SINGLE_SHOT) ? 1 : 0;
 }
 
-int bpf_scan_bytes(const void *buf, __u32 buf__sz, struct rex_scan_attr *attr)
+int rex_scan_tfwstr(const TfwStr *str, struct rex_scan_attr *attr)
```
krizhanovsky, line 93: Now the code is in our tree, so please adjust coding style of all new and changed functions and definitions

regex/kmod/rex.c
```diff
@@ -119,60 +118,11 @@ int bpf_scan_bytes(const void *buf, __u32 buf__sz, struct rex_scan_attr *attr)
...
-	case HS_INVALID:
-	case HS_UNKNOWN_ERROR:
-	default:
-		WARN(1, "hs_scan() failed with code %d\n", (int)err);
-		return -EFAULT;
-	}
-}
-EXPORT_SYMBOL(bpf_scan_bytes);
-
-int bpf_scan_vector(const char *const *buf,
```
krizhanovsky, line 143: I think there is no reason to remove the BPF-purpose routines - they might be useful in xFW, but for sure we don't need code for old kernel versions

fw/regex.c
```diff
@@ -0,0 +1,240 @@
...
+ * 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ * Prototype for fast percentiles calculation.
+ */
+
+#include "regex.h"
+#include "log.h"
+#include "regex/kmod/rex.h"
+#include "tempesta_fw.h"
+
+static char *regex_setup_script_path = "/lib/tempesta/scripts/regex_setup.sh";
```
krizhanovsky, line 27: This path is also installation-dependent, we should get it from the outside: sysctl, module param etc

fw/http_match.c
```diff
@@ -242,174 +242,243 @@ match_host(const TfwHttpReq *req, const TfwHttpMatchRule *rule)
...
+	}
 
-	for (i = TFW_HTTP_HDR_RAW; i < req->h_tbl->off; ++i) {
-		bool col_found;
-		const TfwStr *hdr, *dup, *end, *chunk;
-		const char *c, *cend, *p, *pend;
-		char prev;
-		short cnum;
+	return 0;
+}
```
krizhanovsky, line 259: Isn't it the same as `tfw_cstricmp_2lc()`?

krizhanovsky (approved):
A few improvements are possible, but looks good to merge in general.

I appreciate `match_hdr_raw()` refactoring!

fw/http_match.c
const-t: The main difference is `tfw_cstricmp_2lc()` always calls `tolower()` for both arguments `string` and `pattern` but `__cmp_hdr_raw_name()` doesn't call `tolower()` for h2 at all and do `memcpy_fast()`. For h1 `tolower()` will be called only for `string` not for a substring.

regex/kmod/rex.c
const-t: I removed this this function because it has been added by us in this commit https://github.com/tempesta-tech/tempesta/pull/2161/changes/4d5e3d48d0d74d0f7cef00b4c6d9febe96d05865 and this code is not test covered and even not being used, I believe this is bad approach to leave the code which we are not sure about.

fw/regex.c
const-t: Maybe you commented old version, currently this path is passed to module using parameters.

fw/http_match.c
const-t: Maybe you commented old version, currently this path is passed to module using parameters.

fw/vhost.c
const-t: I've added. https://github.com/tempesta-tech/tempesta/wiki/Virtual-hosts-and-locations

fw/http_tbl.c
const-t: added

fw/http_match.c
const-t: added https://github.com/tempesta-tech/tempesta/wiki/HTTP-tables

const-t:
@krizhanovsky PR is updated. Please consider all review comments addressed.

PR #2164, opened 2024-07-10
Mekhanik evgenii/1411 rework allocation streams

EvgeniiMekhanik:


fw/http2.c
```diff
@@ -192,26 +192,147 @@ tfw_h2_apply_new_settings(TfwH2Ctx *ctx)
...
+	free_pages((unsigned long)ctx, TFW_H2_PAGE_ORDER);
+}
+
+static inline void
+tfw_h2_context_init_stream_storage_impl(TfwH2Ctx *ctx, void *base,
+					unsigned long *next_block)
+{
+	TfwStream *stream = (TfwStream *)base;
+
+	while ((char *)stream <= (char *)next_block - sizeof(TfwStream)) {
```
const-t, line 224: It seems here is memory corruption because of `<=`, in case when quantity is a multiple of size, the store will be filled without hole in the end that need for storing pointer to next block.

const-t (changes requested):

fw/http2.c
```diff
@@ -192,26 +192,147 @@ tfw_h2_apply_new_settings(TfwH2Ctx *ctx)
...
+	(char *)base + PAGE_SIZE * (1 << TFW_H2_PAGE_ORDER);
+#define TFW_H2_NEXT_BLOCK(boundary)				\
+	(unsigned long *)(boundary - sizeof(unsigned long));
+
+TfwH2Ctx *
+tfw_h2_context_alloc(void)
+{
+	struct page *pg;
+
+	pg = alloc_pages(GFP_ATOMIC, TFW_H2_PAGE_ORDER);
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed. We decide to allocate 2 pages.

fw/http2.c
EvgeniiMekhanik: Why if stream == next - sizeof(TfwStream) we have one place for struct stream. I am not right?

PR #2165, opened 2024-07-11
Catch malformed HM response.

enuribekov-tempesta:
In case of parsing err of backend response TFW returns error co client or blocks connection.
In case of this response is from HM there are not any clients to receive server error and no reasons to block own HM traffic. So the solution is just log error and drop response.

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -6790,6 +6790,18 @@ bad_msg:
 	 */
 	bad_req = hmresp->req;
 	tfw_http_popreq(hmresp, false);
+
+	/*
+	 * Special case: malformed response to a Health Monitor request.
+	 * There's no client involved, so just log-n-drop the response now
+	 * without further processing.
+	 */
+	if(unlikely(test_bit(TFW_HTTP_B_HMONITOR, bad_req->flags))) {
+		T_WARN("Health Monitor response malformed");
+		tfw_http_resp_pair_free(bad_req);
+		return T_OK;
+	}
+
 	/* The response is freed by tfw_http_req_parse_block/drop(). */
 	if (filtout)
 		r = tfw_http_req_block(bad_req, 502,


fw/http.c
```diff
@@ -6782,6 +6782,18 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 	 */
 	bad_req = hmresp->req;
 	tfw_http_popreq(hmresp, false);
+
+	/*
+	 * The special case - malformed response on Health Monitoring request.
+	 * In this case client is not involved so we just log-n-drop
+	 * this response without following processing.
+	 */
```
keshonok, line 6790: Perhaps, something like this?
_Special case: malformed response to a Health Monitor request. There's no client involved, so just log-n-drop the response now without further processing._

fw/http.c
```diff
@@ -6782,6 +6782,18 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
...
+	if(test_bit(TFW_HTTP_B_HMONITOR, bad_req->flags)) {
```
keshonok, line 6791: Does `likely()`/`unlikely()` make sense here?

enuribekov-tempesta: Definitely

fw/http.c
```diff
@@ -6782,6 +6782,18 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
...
+
+	/*
+	 * Special case: malformed response to a Health Monitor request.
+	 * There's no client involved, so just log-n-drop the response now 
+	 * without further processing.
+	 */
+	if(unlikely(test_bit(TFW_HTTP_B_HMONITOR, bad_req->flags))) {
+		T_WARN("Health Monitor response malformed");
+		tfw_http_resp_pair_free(bad_req);
+		return T_OK;
```
const-t, line 6794: I've doubts about the case when we got malformed response and part of legit response in the single skb. Now you just drop the response and have potential leak of splitted skb. I think this case is possible if client sent the request that we forwarded and then health monitor timer sent the request. Am I right?

const-t (commented):
A want to ask you to do little improvement, in case of response to health monitoring we should stop response processing after `tfw_http_hm_control((TfwHttpResp *)hmresp);` call. But keep in mind that we must process second response if it presented in the skb.

fw/http.c
EvgeniiMekhanik: It seems that if we can't parse health monitor response, we can't process second response if it is present in the same skb. But it seems we really have a memory leak if spliiting occurs. Also it seems that we can return T_BAD here. not T_OK

EvgeniiMekhanik: I fix memory leak here https://github.com/tempesta-tech/tempesta/pull/2212

EvgeniiMekhanik (approved):

fw/http.c
EvgeniiMekhanik: LGTM, but I think we should return T_BAD not T_OK. @const-t how do you think?

const-t (approved):
LGTM. Mem leak fixed in other PR, but we still have problem with responding to client in case of pipelined responses to different clients from single connection. However @EvgeniiMekhanik will create an issue for that.

Issue #2147, opened 2024-06-20
Crash on malformed server response.


enuribekov-tempesta:
# Motivation

In case of space symbol in any place of any header in the server response parse error occurred than kernel crashes.

# Scope
At the moment such behavior could not be reproduced in any environment different from tempesta-test framework.
Log messages sometimes is absent, sometimes provide some information on the same test with different header strings.

For leading space:
```
[  903.232566] [tempesta fw] Warning: Parser error: state=RGen_HdrOtherN input(-8)=0x20('200 OK
                Content-Length: 0
               
               ') data_len=39 off=17
[  903.237226] Response parsed: len=39 parsed=0 msg_len=0 ver=2 res=-4090
[  903.240292] BUG: kernel NULL pointer dereference, address: 00000000000003b0

```
For space in the middle: 

```
[ 1378.240079] [tempesta fw] Warning: Parser error: state=RGen_HdrOtherN input(-8)=0x6f('00 OK
               Con tent-Length: 0
               
               ') data_len=39 off=18
[ 1378.247868] Response parsed: len=39 parsed=0 msg_len=0 ver=2 res=-4090

```

Wireshark/tcpdump doesn't show the last message due crash.
 
 # Testing
Code for reproducing available is here 
https://github.com/tempesta-tech/tempesta-test/compare/deproxy_response_fail

Change line 439 `response content` to `"HTTP/1.1 200 OK\r\n Content-Length: 0\r\n\r\n"`

Run `./run_tests.py health_monitoring.test_health_monitor.TestHealthMonitorCrc32.test`

krizhanovsky:
Fixed in https://github.com/tempesta-tech/tempesta/pull/2165

PR #2167, opened 2024-07-11
Add CRLF filter.

enuribekov-tempesta:
Strings in configuration file containing '\r' and '\n' passed to code as is which led to malforming of HTTP requests.
To fix this added two sed filters which substitute escaped symbols with hex codes. One for main file and second for includes. WARNING: This requires no additional escape symbols in the config file.

I took this task by myself, because #2147 was very dependent on it and task itself take no more then one hour.

### DIFF
--- a/scripts/tempesta.sh
+++ b/scripts/tempesta.sh
@@ -74,8 +74,9 @@ templater()
 	# Replace !include dircetive with file contents
 	> $tfw_cfg_temp
 	mkdir $TFW_ROOT/etc 2>/dev/null
-	while IFS= read -r line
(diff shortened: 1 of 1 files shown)


const-t (approved):

const-t:
I approved, but I don't like that we do this for whole config. It means we can insert `\r\n` from config file to response or request header. It would be better if we will do it only for health monitoring.

krizhanovsky:
@enuribekov-tempesta can you do as @const-t requested?

enuribekov-tempesta:
@krizhanovsky Exactly.

PR #2168, opened 2024-07-11
Mekhanik evgenii/tfw tls encrypt

EvgeniiMekhanik:


fw/tls.c
```diff
@@ -262,7 +263,18 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int mss_now,
 	struct scatterlist sg[AUTO_SEGS_N], out_sg[AUTO_SEGS_N];
 	struct page **pages = NULL, **pages_end, **p;
 	struct page *auto_pages[AUTO_SEGS_N];
+	bool skb_was_added = false;
 
+	/*
+	 * We should add sk overhead before adding new data to skb which
+	 * already inserted in socket write queue. Because we we fail to
```
krizhanovsky, line 270:
```suggestion
	 * already inserted in socket write queue. Because if we fail to
```

fw/tls.c
```diff
@@ -333,20 +342,23 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int mss_now,
 	t_sz = skb_tail->truesize;
```
krizhanovsky, line 342: It seems `tfw_tls_encrypt()` also make a side effect in `tfw_tcp_propagate_dseq()` at the above.

In general: do we need the new complexity, isn't it easier to not to call `sk_write_xmit()` in the kernel and just reset the TCP connection? If we failed with the current skb encryption, what we're expecting to get with the next call? In my understanding, we'll just either get corrupted application data or terminate the connection anyway.

fw/tls.c
```diff
@@ -262,7 +263,18 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int mss_now,
 	struct scatterlist sg[AUTO_SEGS_N], out_sg[AUTO_SEGS_N];
 	struct page **pages = NULL, **pages_end, **p;
 	struct page *auto_pages[AUTO_SEGS_N];
+	bool skb_was_added = false;
 
+	/*
+	 * We should add sk overhead before adding new data to skb which
+	 * already inserted in socket write queue. Because we we fail to
+	 * add overhead after inserting new data to such skb we catch
+	 * warings or bugs in kernel code later.
```
krizhanovsky, line 272: I guess this happens because we modify skb and don't return its initial size on the function failure. In this context playing with the overhead looks like a dirty workaround: we don't have any real data in the connection, but large overhead.

fw/tls.c
```diff
@@ -398,16 +412,8 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int mss_now,
...
-	/*
-	 * TLS record header is always allocated from the reserved skb headroom.
-	 * The room for the tag may also be allocated from the reserved tailroom
-	 * or in a new page fragment in skb_tail or next, probably new, skb.
-	 * So to adjust the socket write memory we have to check the both skbs
-	 * and only for TTLS_TAG_LEN.
-	 */
-	if (ss_add_overhead(sk, t_sz))
-		return -ENOMEM;
+	ss_del_overhead(sk, TLS_MAX_OVERHEAD - t_sz);
```
krizhanovsky, line 415: So we call the socket memory allocation routines twice instead of once and actually the routines aren't so small - there is a lot of things happening inside them.

fw/tls.c
```diff
@@ -478,18 +484,58 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int mss_now,
...
+			ss_skb_chop_head_tail_nofail(NULL, skb, head_sz, 0);
+			/*
+			 * New skb with TLS tag was added to socket write queue,
+			 * remove it.
+			 */
+			__skb_unlink(skb_tail, &sk->sk_write_queue);
+			skb_tail = skb_tail_save;
+		}
+		tfw_tcp_propagate_dseq(sk, skb);
+		tfw_tcp_propagate_dseq(sk, skb_tail);
```
krizhanovsky, line 530: Why don't we reduce the skb's `end_seq` by the TLS header size to address this addition?
```C
/* TLS header is always allocated from the skb headroom. */
tcb->end_seq += head_sz;
```

krizhanovsky (changes requested):
Maybe there are no bugs in the PR, but I commented several suspicious places. What I don't like about this approach is introduced complexity in already very fragile and sophisticated code and some overhead.

I'd love to have a good discussion how can we reset the TCP connection on the kernel side, it seems we need to do this anyway. I think we should not call the function if we encounter an error and should just reset the connection.

const-t (commented):
> Maybe there are no bugs in the PR, but I commented several suspicious places. What I don't like about this approach is introduced complexity in already very fragile and sophisticated code and some overhead.

Strongly agreed with Alexandr, I would prefer to close the connection in this case.

fw/tls.c
```diff
@@ -334,19 +331,25 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int mss_now,
 	WARN_ON_ONCE(next == skb);
 	if (skb_tail == skb) {
 		r = ss_skb_expand_head_tail(skb->next, skb, head_sz, TTLS_TAG_LEN);
-		if (r < 0)
+		if (r < 0) {
+			tcb->end_seq -= head_sz;
```
const-t, line 335: Just question, maybe we can do `tcb->end_seq += head_sz` after success work with skb?

const-t (approved):

fw/http_limits.c
```diff
@@ -1298,18 +1299,20 @@ frang_resp_process(TfwHttpResp *resp)
 			cli_addr = &req->conn->peer->addr;
 	}
 
+	if (unlikely(resp->content_length > body_len))
+		exceeded = resp->content_length;
+	else if (unlikely(resp->body.len > body_len))
+		exceeded = resp->body.len;
```
krizhanovsky, line 1305: Not a big deal, but according to our and the Linux kernel coding style there should be braces. Did you enable the github hook for checkpatch.pl?

fw/tls.c
```diff
@@ -282,11 +282,8 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int mss_now,
 	head_sz = ttls_payload_off(xfrm);
 	len = skb->len;
 	type = skb_tfw_tls_type(skb);
-	if (!type) {
-		T_WARN("%s: bad skb type %u\n", __func__, type);
-		r = -EINVAL;
-		goto err_epilogue;
-	}
+	/* Checked early before call this function. */
+	BUG_ON(!type);
```
krizhanovsky, line 286: Is a crash inevitable in this case? If not and we still can not to crash, then let's just put WARN_ON_ONCE(). By the way, checkpatch.pl also claims unnecessary BUG statements

fw/tls.c
krizhanovsky: It seems we need correct `tcb->end_seq` for `tfw_tcp_propagate_dseq()`.

fw/tls.c
```diff
@@ -446,15 +449,15 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int mss_now,
...
+			goto free_pages;
 		frags += r;
 
 		r = ss_skb_to_sgvec_with_new_pages(next,
 		                                   out_sgt.sgl + out_frags,
 		                                   &pages_end);
-		if (r <= 0)
-			goto out;
+		if (r < 0)
+			goto free_pages;
```
krizhanovsky, line 460: Cool! One more memory leak is fixed!

krizhanovsky (approved):
LGTM, but it'd be good to have a couple of improvements

fw/tls.c
EvgeniiMekhanik: Yes it is necessary for tfw_tcp_propagate_dseq

fw/tls.c
```diff
@@ -446,15 +449,15 @@ tfw_tls_encrypt(struct sock *sk, struct sk_buff *skb, unsigned int mss_now,
 		       skb_shinfo(next)->nr_frags + !!skb_headlen(next),
 		       frags, r);
 
-		if (r <= 0)
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed

PR #2173, opened 2024-07-17
fix(1902): correct trailer headers placement

### LINKED ISSUE #1902: Correct handling on trailer headers in h1 -> h2 proxying

kingluo:
close #1902

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -4,7 +4,7 @@
  * HTTP cache (RFC 7234).
  *
  * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
- * Copyright (C) 2015-2024 Tempesta Technologies, Inc.
+ * Copyright (C) 2015-2025 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -203,13 +203,16 @@ __tfw_dbg_dump_ce(const TfwCacheEntry *ce)
 #define TFW_CSTR_SPEC_IDX	0x2
 /*
  * TfwCStr contains header, which should be skipped for http2 responses
- * constructed from cache.
+ * constructed from cache. We skip transfer encoding header, when we
+ * build response from cache for HTTP2 request.
+ * RFC 9113 8.1:
+ * HTTP/2 uses DATA frames to carry message content. The chunked
+ * transfer encoding defined in Section 7.1 of [HTTP/1.1] cannot be
+ * used in HTTP/2; see Section 8.2.2.
  */
 #define TFW_CSTR_SKIP_FOR_H2	0x4
 
-/*
- * TfwCStr contains trailer.
- */
+/* TfwCStr contains trailer. */
 #define TFW_CSTR_TRAILER	0x8
 
 /**
@@ -1065,8 +1068,8 @@ do {									\
 
 		*p += TFW_CSTR_HDRLEN;
 		r = write_actor(db, trec, resp, p, s->len, &dc_iter);
-			if (unlikely(r))
-				break;
+		if (unlikely(r))
+			break;
 
 		dc_iter.skip = skip;
 	}
@@ -2753,7 +2756,8 @@ tfw_cache_purge_method(TfwHttpReq *req)
 
 	/* Deny PURGE requests by default. */
 	if (!(cache_cfg.cache && g_vhost->cache_purge
-	    && g_vhost->cache_purge_acl)) {
+	    && g_vhost->cache_purge_acl))
+	{
 		tfw_http_send_err_resp(req, 403, "purge: not configured");
 		return -EINVAL;
 	}
@@ -2851,14 +2855,6 @@ tfw_cache_build_resp_body(TDB *db, TdbVRec *trec, TfwMsgIter *it, char *p,
 	 * Finish chunked body encoding. Add 0\r\n
 	 * after chunked body.
 	 */
-	TfwStr b_len = {
-		.chunks = (TfwStr []){
-			{.data = S_ZERO, .len = SLEN(S_ZERO)},
-			{.data = S_CRLF, .len = SLEN(S_CRLF)}
-		},
-		.len = SLEN(S_ZERO S_CRLF),
-		.nchunks = 2
-	};
 	bool sh_frag = h2 ? false : true;
 	int r;
 
@@ -2941,6 +2937,11 @@ tfw_cache_build_resp_body(TDB *db, TdbVRec *trec, TfwMsgIter *it, char *p,
 	}
 
 	if (chunked_body) {
+		static TfwStr b_len = {
+			.data = S_ZERO S_CRLF,
+			.len = SLEN(S_ZERO S_CRLF)
+		};
+
 		r = tfw_http_msg_expand_data(it, &it->skb_head,
 					     &g_crlf, NULL);
 		if (unlikely(r))
@@ -3057,7 +3058,6 @@ tfw_cache_build_resp(TfwHttpReq *req, TfwCacheEntry *ce, long age)
 	if (!(resp = tfw_http_msg_alloc_resp(req)))
 		goto out;
 
-
 	/* Copy version information and flags */
 	resp->version = ce->version;
 	tfw_http_copy_flags(resp->flags, ce->hmflags);
@@ -3171,7 +3171,8 @@ write_body:
 	/* Fill skb with body from cache for HTTP/2 or HTTP/1.1 response. */
 	BUG_ON(p != TDB_PTR(db->hdr, ce->body));
 	if ((ce->body_len || chunked_body)
-	    && req->method != TFW_HTTP_METH_HEAD) {
+	    && req->method != TFW_HTTP_METH_HEAD)
+	{
 		if (tfw_cache_build_resp_body(db, trec, it, p, ce->body_len,
 					      h2_mode, chunked_body))
 			goto free;
@@ -3179,7 +3180,8 @@ write_body:
 	resp->content_length = ce->body_len;
 
 	if (unlikely(ce->trailer_off < ce->hdr_num)
-	    && req->method != TFW_HTTP_METH_HEAD) {
+	    && req->method != TFW_HTTP_METH_HEAD)
+	{
 		unsigned long t_len = 0;
 
 		if (h2_mode)
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -72,6 +72,7 @@ static const TfwStr tfw_cache_raw_headers_304[] = {
 /* Flags stored in a Cache Entry. */
 #define TFW_CE_MUST_REVAL	0x0001		/* MUST revalidate if stale. */
 #define TFW_CE_STALE_IF_ERROR	0x0002
+#define TFW_CE_CHUNKED_BODY	0x0004		/* CE entry contains chunked body. */
 
 /*
  * @trec	- Database record descriptor;
@@ -81,6 +82,7 @@ static const TfwStr tfw_cache_raw_headers_304[] = {
  * @hdr_num	- number of headers;
  * @hdr_len	- length of whole headers data;
  * @hdr_h2_off	- start of http/2-only headers in the headers list;
+ * @trailer_off	- start of trailer headers;
  * @body_len	- length of the response body;
  * @method	- request method, part of the key;
  * @flags	- various cache entry flags;
@@ -110,6 +112,7 @@ typedef struct {
 	unsigned int	rph_len;
 	unsigned int	hdr_num;
 	unsigned int	hdr_h2_off;
+	unsigned int	trailer_off;
 	unsigned int	hdr_len;
 	unsigned int	body_len;
 	unsigned int	method: 4;
@@ -123,6 +126,7 @@ typedef struct {
 	long		key;
 	long		status;
 	long		hdrs;
+	long		trailers;
 	long		body;
 	long		hdrs_304[TFW_CACHE_304_HDRS_NUM];
 	long		stale_if_error;
@@ -162,24 +166,25 @@ __tfw_dbg_dump_ce(const TfwCacheEntry *ce)
 			      "TdbVRec", ce->trec.key, ce->trec.chunk_next,
 			      ce->trec.len, ce_total_size(ce));
 	len += CE_DUMP_MEMBER("%d", "key_len",		ce->key_len);
-	len += CE_DUMP_MEMBER("%d", "status_len",     	ce->status_len);
-	len += CE_DUMP_MEMBER("%d", "rph_len",        	ce->rph_len);
-	len += CE_DUMP_MEMBER("%d", "hdr_num",        	ce->hdr_num);
-	len += CE_DUMP_MEMBER("%d", "hdr_h2_off",     	ce->hdr_h2_off);
-	len += CE_DUMP_MEMBER("%d", "hdr_len",        	ce->hdr_len);
-	len += CE_DUMP_MEMBER("%d", "body_len",       	ce->body_len);
-	len += CE_DUMP_MEMBER("%d", "method",	      	ce->method);
-	len += CE_DUMP_MEMBER("%d", "flags",	      	ce->flags);
+	len += CE_DUMP_MEMBER("%d", "status_len",	ce->status_len);
+	len += CE_DUMP_MEMBER("%d", "rph_len",		ce->rph_len);
+	len += CE_DUMP_MEMBER("%d", "hdr_num",		ce->hdr_num);
+	len += CE_DUMP_MEMBER("%d", "hdr_h2_off",	ce->hdr_h2_off);
+	len += CE_DUMP_MEMBER("%d", "trailer_off",	ce->trailer_off);
+	len += CE_DUMP_MEMBER("%d", "hdr_len",		ce->hdr_len);
+	len += CE_DUMP_MEMBER("%d", "body_len",		ce->body_len);
+	len += CE_DUMP_MEMBER("%d", "method",		ce->method);
+	len += CE_DUMP_MEMBER("%d", "flags",		ce->flags);
 
 	len += CE_DUMP_MEMBER("%lu", "age",		ce->age);
-	len += CE_DUMP_MEMBER("%lu", "date",	      	ce->date);
+	len += CE_DUMP_MEMBER("%lu", "date",		ce->date);
 	len += CE_DUMP_MEMBER("%lu", "req_time",	ce->req_time);
-	len += CE_DUMP_MEMBER("%lu", "resp_time",   	ce->resp_time);
-	len += CE_DUMP_MEMBER("%lu", "last_modified",   ce->last_modified);
+	len += CE_DUMP_MEMBER("%lu", "resp_time",	ce->resp_time);
+	len += CE_DUMP_MEMBER("%lu", "last_modified",	ce->last_modified);
 	len += CE_DUMP_MEMBER("%lx", "key off",		ce->key);
-	len += CE_DUMP_MEMBER("%lx", "status off",  	ce->status);
-	len += CE_DUMP_MEMBER("%lx", "hdrs off",    	ce->hdrs);
-	len += CE_DUMP_MEMBER("%lx", "body off",    	ce->body);
+	len += CE_DUMP_MEMBER("%lx", "status off",	ce->status);
+	len += CE_DUMP_MEMBER("%lx", "hdrs off",	ce->hdrs);
+	len += CE_DUMP_MEMBER("%lx", "body off",	ce->body);
 	/* @hmflags occupies a single ulong */
 	len += CE_DUMP_MEMBER("%lx [%*pbl]", "hmflags", *ce->hmflags,
 			      _TFW_HTTP_FLAGS_NUM, ce->hmflags);
@@ -196,6 +201,16 @@ __tfw_dbg_dump_ce(const TfwCacheEntry *ce)
 #define TFW_CSTR_DUPLICATE	TFW_STR_DUPLICATE
 /* TfwCStr contains special header and its id. */
 #define TFW_CSTR_SPEC_IDX	0x2
+/*
+ * TfwCStr contains header, which should be skipped for http2 responses
+ * constructed from cache.
+ */
+#define TFW_CSTR_SKIP_FOR_H2	0x4
+
+/*
+ * TfwCStr contains trailer.
+ */
+#define TFW_CSTR_TRAILER	0x8
 
 /**
  * String header for cache entries used for TfwStr serialization.
@@ -232,11 +247,16 @@ typedef struct {
 
 /* Cache modes. */
 typedef enum {
-        TFW_CACHE_UNDEFINED = -1,
-        TFW_CACHE_NONE = 0,
-        TFW_CACHE_SHARD,
-        TFW_CACHE_REPLICA,
+	TFW_CACHE_UNDEFINED = -1,
+	TFW_CACHE_NONE = 0,
+	TFW_CACHE_SHARD,
+	TFW_CACHE_REPLICA,
 } TfwCacheMode;
+
+typedef enum {
+	TFW_CACHE_HDR,
+	TFW_CACHE_TRAILER,
+} TfwCacheHdrType;
 
 static struct {
 	int cache;
@@ -919,7 +939,8 @@ tfw_cache_set_status(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp,
 }
 
 static bool
-tfw_cache_skip_hdr(const TfwCStr *str, char *p, const TfwHdrMods *h_mods)
+tfw_cache_skip_hdr(const TfwCStr *str, char *p, const TfwHdrMods *h_mods,
+		   bool h2_mode)
 {
 	unsigned int i;
 	const TfwHdrModsDesc *desc;
@@ -929,6 +950,12 @@ tfw_cache_skip_hdr(const TfwCStr *str, char *p, const TfwHdrMods *h_mods)
 	 */
 	TfwStr hdr = { .data = p + str->name_len_sz + 1,
 		       .len  = str->name_len };
+
+	if (h2_mode && str->flags & TFW_CSTR_SKIP_FOR_H2)
+		return true;
+
+	if (!h_mods)
+		return false;
 
 	/* Fast path for special headers */
 	if (str->flags & TFW_CSTR_SPEC_IDX) {
@@ -970,13 +997,14 @@ tfw_cache_skip_hdr(const TfwCStr *str, char *p, const TfwHdrMods *h_mods)
 static int
 tfw_cache_build_resp_hdr(TDB *db, TfwHttpResp *resp, TfwHdrMods *hmods,
 			 TdbVRec **trec, char **p, unsigned long *acc_len,
-			 bool skip)
+			 bool skip, TfwCacheHdrType type)
 {
 	tfw_cache_write_actor_t *write_actor;
 	TfwCStr *s;
 	TfwHttpReq *req = resp->req;
 	TfwDecodeCacheIter dc_iter = { .h_mods = hmods, .skip = skip };
 	int d, dn, r = 0;
+	bool h2_mode = TFW_MSG_H2(req);
 
 /* Go to the next chunk if we at the end of current. */
 #define NEXT_CHUNK()							\
@@ -989,8 +1017,8 @@ do {									\
 
 	BUG_ON(!req);
 
-	write_actor = !TFW_MSG_H2(req) ? tfw_cache_h2_decode_write
-				       : tfw_cache_h2_write;
+	write_actor = !h2_mode ? tfw_cache_h2_decode_write
+			       : tfw_cache_h2_write;
 
 	NEXT_CHUNK();
 	s = (TfwCStr *)*p;
@@ -998,17 +1026,19 @@ do {									\
 	BUG_ON(*p > (*trec)->data + (*trec)->len);
 
 	if (likely(!(s->flags & TFW_CSTR_DUPLICATE))) {
-		if (!skip && dc_iter.h_mods)
-			dc_iter.skip = tfw_cache_skip_hdr(s, *p, hmods);
+		if (!skip)
+			dc_iter.skip = tfw_cache_skip_hdr(s, *p, hmods,
+							  h2_mode);
 		r = write_actor(db, trec, resp, p, s->len, &dc_iter);
 		if (likely(!r))
 			*acc_len += dc_iter.acc_len;
 		return r;
 	}
 
-	if (!skip && dc_iter.h_mods)
+	if (!skip)
 		dc_iter.skip = tfw_cache_skip_hdr((TfwCStr *)*p,
-						  *p + TFW_CSTR_HDRLEN, hmods);
+						  *p + TFW_CSTR_HDRLEN, hmods,
+						  h2_mode);
 
 	/* Process duplicated headers. */
 	dn = s->len;
@@ -1016,10 +1046,29 @@ do {									\
 		NEXT_CHUNK();
 		s = (TfwCStr *)*p;
 		BUG_ON(s->flags & TFW_CSTR_DUPLICATE);
+
+		skip = dc_iter.skip;
+		if (!dc_iter.skip) {
+			/*
+			 * If header is present in headers and in trailers
+			 * we save this header in cache two times. When we
+			 * build response from cache we check what chunks
+			 * of duplicate header belong to headers and what
+			 * to trailers and write it to response depends on
+			 * header or trailer we restore now.
+			 */
+			dc_iter.skip = !(((type == TFW_CACHE_HDR &&
+					 !(s->flags & TFW_CSTR_TRAILER))
+					|| (type == TFW_CACHE_TRAILER
+					    && (s->flags & TFW_CSTR_TRAILER))));
+		}
+
 		*p += TFW_CSTR_HDRLEN;
 		r = write_actor(db, trec, resp, p, s->len, &dc_iter);
-		if (unlikely(r))
-			break;
+			if (unlikely(r))
+				break;
+
+		dc_iter.skip = skip;
 	}
 	*acc_len += dc_iter.acc_len;
 
@@ -1093,7 +1142,7 @@ tfw_cache_send_304(TfwHttpReq *req, TfwCacheEntry *ce)
 		BUG_ON(!trec);
 
 		if (tfw_cache_build_resp_hdr(db, resp, NULL, &trec, &p, &h_len,
-					     false))
+					     false, TFW_CACHE_HDR))
 		{
 			goto err_setup;
 		}
@@ -1670,6 +1719,9 @@ tfw_cache_h2_copy_hdr(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, int hid,
 	const bool need_extra_quotes =
 		test_bit(TFW_HTTP_B_HDR_ETAG_HAS_NO_QOUTES, resp->flags)
 		&& (hid == TFW_HTTP_HDR_ETAG);
+	unsigned int extra_flags = (hid == TFW_HTTP_HDR_TRANSFER_ENCODING ?
+		TFW_CSTR_SKIP_FOR_H2 : 0);
+	unsigned int dupl_extra_flags = 0;
 
 	T_DBG3("%s: ce=[%p] p=[%p], trec=[%p], tot_len='%zu'\n", __func__, ce,
 	       *p, *trec, *tot_len);
@@ -1701,11 +1753,11 @@ tfw_cache_h2_copy_hdr(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, int hid,
 	if (TFW_STR_DUP(hdr)) {
 		CSTR_MOVE_HDR();
 		if (hid >= TFW_HTTP_HDR_REGULAR  && hid < TFW_HTTP_HDR_RAW)
-			CSTR_WRITE_HDR(TFW_CSTR_SPEC_IDX | TFW_CSTR_DUPLICATE,
-				       hdr->nchunks, 0, hid);
+			CSTR_WRITE_HDR(TFW_CSTR_SPEC_IDX | TFW_CSTR_DUPLICATE |
+				       extra_flags, hdr->nchunks, 0, hid);
 		else
-			CSTR_WRITE_HDR(TFW_CSTR_DUPLICATE, hdr->nchunks, 0,
-				       st_index);
+			CSTR_WRITE_HDR(TFW_CSTR_DUPLICATE | extra_flags,
+				       hdr->nchunks, 0, st_index);
 	}
 
 	TFW_STR_FOR_EACH_DUP(dup, hdr, dup_end) {
@@ -1749,12 +1801,15 @@ tfw_cache_h2_copy_hdr(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, int hid,
 							    need_extra_quotes))
 			return -ENOMEM;
 
+		dupl_extra_flags = extra_flags | (dup->flags & TFW_STR_TRAILER ?
+			TFW_CSTR_TRAILER : 0);
+
 		if (hid >= TFW_HTTP_HDR_REGULAR && hid < TFW_HTTP_HDR_RAW)
-			CSTR_WRITE_HDR(TFW_CSTR_SPEC_IDX,
+			CSTR_WRITE_HDR(TFW_CSTR_SPEC_IDX | dupl_extra_flags,
 				       ce->hdr_len - prev_len, 0, hid);
 		else
-			CSTR_WRITE_HDR(0, ce->hdr_len - prev_len, s_nm.len,
-				       st_index);
+			CSTR_WRITE_HDR(dupl_extra_flags, ce->hdr_len - prev_len,
+				       s_nm.len, st_index);
 	}
 
 	T_DBG3("%s: p=[%p], trec=[%p], ce->hdr_len='%u', tot_len='%zu'\n",
@@ -1765,7 +1820,8 @@ tfw_cache_h2_copy_hdr(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, int hid,
 
 static long
 tfw_cache_h2_add_hdr(TDB *db, TfwCacheEntry *ce, char **p, TdbVRec **trec,
-		     unsigned short st_idx, TfwStr *val, size_t *tot_len)
+		     unsigned short st_idx, TfwStr *val, size_t *tot_len,
+		     unsigned int flags)
 {
 	TfwCStr *cs;
 	unsigned long len;
@@ -1806,7 +1862,8 @@ tfw_cache_h2_add_hdr(TDB *db, TfwCacheEntry *ce, char **p, TdbVRec **trec,
 	if (tfw_cache_h2_copy_str(db, &ce->hdr_len, p, trec, val, tot_len))
 		return -ENOMEM;
 
-	CSTR_WRITE_HDR(0, ce->hdr_len - prev_len - TFW_CSTR_HDRLEN, 0, st_idx);
+	CSTR_WRITE_HDR(flags, ce->hdr_len - prev_len - TFW_CSTR_HDRLEN, 0,
+		       st_idx);
 
 	return ce->hdr_len - prev_len;
 }
@@ -1829,33 +1886,7 @@ tfw_cache_add_hdr_cenc(TDB *db, TfwHttpResp *resp, TfwCacheEntry *ce, char **p,
 		return r;
 
 	val_ce.len = chunk.len;
-	return tfw_cache_h2_add_hdr(db, ce, p, trec, 26, &val_ce, tot_len);
-}
-
-/**
- * Add 'Content-Length' header to the cache record. The length will be taken
- * from parsed response.
- */
-static long
-tfw_cache_add_hdr_clen(TDB *db, TfwHttpResp *resp, TfwCacheEntry *ce, char **p,
-		       TdbVRec **trec, size_t *tot_len)
-{
-	TfwStr val = {
-		.chunks = (TfwStr []){
-			{ .data = *this_cpu_ptr(&g_c_buf), .len = 0 }
-		},
-		.nchunks = 1
-	};
-	unsigned long body_len = TFW_HTTP_RESP_CUT_BODY_SZ(resp);
-
-	val.chunks->len = tfw_ultoa(body_len, val.chunks->data,
-				    TFW_ULTOA_BUF_SIZ);
-
-	val.len = val.chunks->len;
-	if (!val.len)
-		return -EINVAL;
-
-	return tfw_cache_h2_add_hdr(db, ce, p, trec, 28, &val, tot_len);
+	return tfw_cache_h2_add_hdr(db, ce, p, trec, 26, &val_ce, tot_len, 0);
 }
 
 /**
@@ -1972,7 +2003,7 @@ __set_etag(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, long h_off,
 /**
  * Check if the header @hdr must be present in 304 response. If yes save its
  * offset in cache entry @ce for fast creation of 304 response in
- * tfw_cahe_send_304().
+ * tfw_cache_send_304().
  */
 static bool
 __save_hdr_304_off(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *hdr, long off)
@@ -2146,14 +2177,21 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 		 * version of this header.
 		 */
 		if (TFW_STR_EMPTY(field)
-		    || (field->flags & (TFW_STR_HBH_HDR | TFW_STR_NOCCPY_HDR))
+		    || (field->flags &
+			(TFW_STR_HBH_HDR | TFW_STR_NOCCPY_HDR
+			 | TFW_STR_TRAILER_HDR))
 		    || hid == TFW_HTTP_HDR_SERVER)
 		{
 			--ce->hdr_num;
 			continue;
 		}
 
-		if (hid == TFW_HTTP_HDR_TRANSFER_ENCODING) {
+		/*
+		 * Save header if it contains chunks which belongs both
+		 * for headers and trailers two times here and later.
+		 */
+		if ((field->flags & TFW_STR_TRAILER
+		    && !(field->flags & TFW_STR_TRAILER_AND_HDR))) {
 			--ce->hdr_num;
 			continue;
 		}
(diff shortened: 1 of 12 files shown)


fw/http_frame.c
```diff
@@ -1967,23 +1973,30 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
 	length = tfw_h2_calc_frame_length(ctx, stream, type, len,
 					  max_len - FRAME_HEADER_SIZE);
 	if (type == HTTP2_DATA) {
-		ctx->rem_wnd -= length;
-		stream->rem_wnd -= length;
+		if (!stream->xmit.t_len) {
```
EvgeniiMekhanik, line 1976: This is mistake, we should adjust window in case when response contains trailers.
I implement new test for this case.
https://github.com/tempesta-tech/tempesta-test/pull/651
It passes on master but fails on your branch

fw/http2.c
```diff
@@ -489,6 +489,38 @@ tfw_h2_stream_xmit_prepare_resp(TfwStream *stream)
...
+		// and append them to the resp->msg.skb_head.
+		struct sk_buff *skb_head, *skb, *nskb_head, *nskb;
+		unsigned long acc;
+
+		skb_head = resp->msg.skb_head;
+		skb = resp->mit.iter.skb;
+		resp->msg.skb_head = resp->mit.iter.skb = NULL;
+		acc = resp->mit.acc_len;
+
+		r = tfw_h2_hpack_encode_trailer_headers(resp);
```
EvgeniiMekhanik, line 505: I don't agree with this solution. We should rework our standard functions to add trailers. Here we have extra loop over skbs and it looks bad for me. @krizhanovsky @const-t please look he and write your opinion.

fw/cache.c
```diff
@@ -2838,14 +2861,48 @@ tfw_cache_build_resp(TfwHttpReq *req, TfwCacheEntry *ce, long lifetime)
...
+		TfwHttpTransIter pmit, nmit;
+		struct sk_buff *pskb_head = resp->msg.skb_head;
+
+		has_trailers = true;
+		pmit = nmit = resp->mit;
+		nmit.start_off = 0;
+		nmit.iter.skb = nmit.iter.skb_head = NULL;
+		nmit.iter.frag = -1;
+
+		resp->msg.skb_head = NULL;
```
EvgeniiMekhanik, line 2884: I don't agree with this solution. We should rework our standard functions to add trailers. Here we have extra loop over skbs and it looks bad for me. @krizhanovsky @const-t please look he and write your opinion.

fw/http_frame.c
```diff
@@ -1967,23 +1973,30 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
...
+			ctx->rem_wnd -= length;
+			stream->rem_wnd -= length;
+		}
 		stream->xmit.b_len -= length;
-	} else {
+	} else if (stream->xmit.h_len) {
 		stream->xmit.h_len -= length;
+	} else if (stream->xmit.t_len) {
+		ctx->rem_wnd -= length;
```
EvgeniiMekhanik, line 1984: Why we decrement window here? trailers are the headers frames.

fw/http_frame.c
kingluo: I thought we could adjust the wnd stuff until trailers handling, but yes, you're right, maybe just revert them.

fw/http_frame.c
kingluo: ditto.

fw/http2.c
```diff
@@ -489,6 +489,38 @@ tfw_h2_stream_xmit_prepare_resp(TfwStream *stream)
 	if (test_bit(TFW_HTTP_B_CHUNKED, resp->flags))
 		r = tfw_http_msg_cutoff_body_chunks(resp);
```
EvgeniiMekhanik, line 490: we should check r here, and do not process trailers in case of error.

fw/http2.c
EvgeniiMekhanik: I don't see any problem here. I  implement such part of code
```
if (resp->trailers_len > 0) {
		{
			TfwHttpTransIter *mit = &resp->mit;
			TfwMsgIter *it = &mit->iter;
			
			it->frag = 4; // Setup it->frag or Iter->skb and frag which points to the end of the body. I set here 4, to check that test works fine.
 		}

		{
			unsigned long acc = resp->mit.acc_len;
			r = tfw_h2_hpack_encode_trailer_headers(resp);
			if (unlikely(r))
				goto finish;
			stream->xmit.t_len = resp->mit.acc_len - acc;
		}
```
And all works fine. So we don't need all previous code. We have two solutions:
1. Bad solution, not better performance that current solution, but less code. Use `tfw_http_iter_set_at
` or `tfw_msg_iter_move` to set at the end of the body.
2. Save pointer to the body and use it, may found body end during `tfw_http_msg_cutoff_body_chunks`

fw/hpack.c
```diff
@@ -3488,6 +3488,59 @@ tfw_hpack_hdr_expand(TfwHttpResp *__restrict resp, TfwStr *__restrict hdr,
...
 		return -EINVAL;
 
+	// compact str with sparse chunks in place
+	// e.g. changes
+	// "f" "o" ":" " " "b" "a"
+	// into
+	// "fo" ": " "ba"
+	// it's different from `tfw_http_hdr_split()`, which only splits
+	// name chunks and value chunks into two strs, no compact.
+	if (hdr->flags & TFW_STR_TRAILER) {
```
EvgeniiMekhanik, line 3498: Why we should do it? Is this transformation necessary according RFC? If yes please add comment. And we don't need this part of code. I think that we can do it during parsing trailer headers in http_parser.c. @const-t how do you think?

kingluo: This conversion only makes sense for our implementation to satisfy the needs of the header encoding below.

fw/http_frame.c
```diff
@@ -2116,6 +2149,13 @@ do {									\
 						       &stream->xmit.postponed);
```
EvgeniiMekhanik, line 2149: As I think trailers works same as headers frames. So we can't send postponed skbs if we start but not finished sending trailer frames. We also need test this rarely case. We should implement test with a very long trailer. And simulate a situation when client sends us new settings (for example) after receiving first trailer frame but before receiving all trailers.

kingluo: The role of trailers is equivalent to the body. In fact, the only application of HTTP/2 trailers so far is GRPC, which uses the trailers to represent the call result, and the trailers are expected to follow the body from the perspective of GRPC client.

fw/http_frame.c
```diff
@@ -2095,7 +2104,31 @@ do {									\
...
+					       snd_wnd, stream->xmit.t_len);
+		if (unlikely(r)) {
+			T_WARN("Failed to make trail headers frame %d", r);
+			return r;
+		}
+
+		T_FSM_JMP(HTTP2_SEND_FRAMES);
+	}
+
+	T_FSM_STATE(HTTP2_MAKE_TRAILER_CONTINUATION_FRAMES) {
```
EvgeniiMekhanik, line 2122: I add test for long trailers https://github.com/tempesta-tech/tempesta-test/pull/651/commits/ec625cefffced6fd6e0334d5cebf08c545e59ba3 and it doesn't pass

fw/http_frame.c
EvgeniiMekhanik: I mean that if we have long trailer about 30000 bytes for example we can't send it without continuation frame. If we send some other frame after we send first trailer (in fact headers) frame but before we send continuation trailer (continuation headers) frame we catch connection error of type protocol error. For this purpose we have postponed skbs.

fw/http2.c
kingluo: If we really need to eliminate the "ugly" skb_head and mit manipulation, we can simply do `it->skb = NULL`, which triggers the new skb allocation to contain encoded trailers.
https://github.com/tempesta-tech/tempesta/blob/2a8d72d25aa638e58bdd2d0359d71c64f259ef9b/fw/http_msg.c#L1303

EvgeniiMekhanik: We don't need to allocate extra skb if there is enough place in current skb. For short answers with trailers (which could be placed in one skb, you double the number of sending skbs, which is very inefficient.

fw/http_frame.c
kingluo: I get it. I'll check it later.
A quick question, shouldn't `stream->xmit.frame_length` be non-zero if any continuations? If so, why it will come into this branch?

EvgeniiMekhanik: We come to this branch after all continuation frames will be sent

fw/http_stream.h
```diff
@@ -121,23 +123,27 @@ typedef enum {
...
 	TfwHttpResp 		*resp;
 	struct sk_buff		*skb_head;
 	struct sk_buff		*postponed;
 	unsigned int		h_len;
+	unsigned int		t_len;
 	unsigned int		frame_length;
 	u64			b_len : 60;
 	u64			is_blocked : 1;
-	u64			state : 3;
+	u64			state : 4;
```
EvgeniiMekhanik, line 145: This field moved to another u64 bitfield. The idea was that b_len + is_blocked + state == 64 bits. I think you can reduce b_len to 58 and nove t_len after this bitfield. Now there is a hole between frame_length and b_len

fw/http_frame.c
kingluo: Yes, I know. I mean the code should use `*_len` as conditional. The `frame_length` here looks rather obscure. I'll tune it later.

fw/http_frame.c
```diff
@@ -1894,21 +1894,25 @@ tfw_h2_calc_frame_length(TfwH2Ctx *ctx, TfwStream *stream, TfwFrameType type,
...
-		return stream->xmit.h_len ? 0 : HTTP2_F_END_HEADERS;
-	case HTTP2_DATA:
-		return stream->xmit.b_len ? 0 : HTTP2_F_END_STREAM;
-	default:
-		BUG();
-	};
+	char flags = 0;
 
-	return 0;
+	if (stream->xmit.h_len && !stream->xmit.b_len
```
EvgeniiMekhanik, line 1899: According to RFC 9113 8.1:

Trailer fields are carried in a field block that also terminates the stream. That is, trailer fields comprise a sequence starting with a [HEADERS](https://www.rfc-editor.org/rfc/rfc9113.html#HEADERS) frame, followed by zero or more [CONTINUATION](https://www.rfc-editor.org/rfc/rfc9113.html#CONTINUATION) frames, where the [HEADERS](https://www.rfc-editor.org/rfc/rfc9113.html#HEADERS) frame bears an END_STREAM flag. Trailers MUST NOT include [pseudo-header fields](https://www.rfc-editor.org/rfc/rfc9113.html#PseudoHeaderFields) ([Section 8.3](https://www.rfc-editor.org/rfc/rfc9113.html#PseudoHeaderFields)). An endpoint that receives pseudo-header fields in trailers MUST treat the request or response as [malformed](https://www.rfc-editor.org/rfc/rfc9113.html#malformed) ([Section 8.1.1](https://www.rfc-editor.org/rfc/rfc9113.html#malformed)).

So if trailer exists we should set HTTP2_F_END_STREAM after sending HEADERS frame for trailers.

fw/http_frame.c
```diff
@@ -1894,21 +1894,25 @@ tfw_h2_calc_frame_length(TfwH2Ctx *ctx, TfwStream *stream, TfwFrameType type,
 static inline char
 tf2_h2_calc_frame_flags(TfwStream *stream, TfwFrameType type)
```
EvgeniiMekhanik, line 1895: please rename this functiin to `tf2_h2_calc_frame_flags` it was my mistake, I miss it in my PR

kingluo: what do you mean? It's already named as `tf2_h2_calc_frame_flags`.

EvgeniiMekhanik: tfw_h2_calc_frame_flags sorry again my mistake)

fw/http2.c
kingluo: You can't use the response body's skb to hpack the trailer, because the skb is allocated from `pg_skb_alloc`, but `tfw_http_msg_expand_data` used by hpack assumes that you can use a whole page with one frag, which is conflicted to each other, and may cause the trailer to be truncated. Your successful case of reusing the frag in the test was nothing but a lucky case where the frag lounge just fit the trailer size. If you try to rerun the test with `-T 1`, you will see an error (frag=1, but the frag cannot actually accept the trailer). So all in all, a new skb seems to be the only feasible way. Note that we also can't use `resp->pool`, as mentioned repeatedly before, that way assumes that the response body follows the headers, i.e., that is a dedicated method for allocating normal headers and body, which does not work with trailers.

fw/cache.c
kingluo: Can't find a way to eliminate this `skb_head` operation, because the response built from cache requires the headers to be built first, so there's no way to avoid creating a new skb before building the body and appending it after the body.

fw/http2.c
EvgeniiMekhanik: It works! You should setup iterator->skb to the last body skb and iterator->frag to the last frag (which contains body!) of the last body skb. I write code below and it works! It is a dirty code to show that if you setup iterator all works fine. You should find a way to find the end of the body (skb and its last used fragment) and setup itertor!
```
{
			TfwHttpTransIter *mit = &resp->mit;
			TfwMsgIter *it = &mit->iter;
			
			struct sk_buff *skb = resp->msg.skb_head;
			do {
				if (skb->next == resp->msg.skb_head)
					break;
				skb = skb->next;
			} while(skb != resp->msg.skb_head);

			it->skb = skb;
			it->frag = skb_shinfo(it->skb)->nr_frags - 1;
		}

		{
			unsigned long acc = resp->mit.acc_len;
			TfwHttpTransIter *mit = &resp->mit;
			TfwMsgIter *it = &mit->iter;
			
			r = tfw_h2_hpack_encode_trailer_headers(resp);
			if (unlikely(r))
				goto finish;
			stream->xmit.t_len = resp->mit.acc_len - acc;
		}
```

fw/http_stream.h
```diff
@@ -121,23 +123,27 @@ typedef enum {
...
  * @state		- current stream xmit state (what type of
  * 			  frame should be made for this stream);
+ * @is_trailer_cont	- need to send trailer CONTINUATION frames;
  */
 typedef struct {
 	TfwHttpResp 		*resp;
 	struct sk_buff		*skb_head;
 	struct sk_buff		*postponed;
 	unsigned int		h_len;
+	unsigned int		t_len;
```
EvgeniiMekhanik, line 141: This structure now has 5 bytes hole. please move `t_len` to avoid this
 `pahole -C TfwHttpXmit fw/tempesta_fw.ko`

fw/http_frame.c
EvgeniiMekhanik: This comment is not resolved. If trailers are present we should not set HTTP2_F_END_STREAM here according RFC

fw/http2.c
EvgeniiMekhanik: I also change increase body size in test and check it also for segmentation. If you setup iterator all works fine!

kingluo: I don't understand.

resp body skbs come from `pg_skb_alloc` used by the kernel from the backend socket receive path, once the frag is allocated, it cannot be extended in place, but a new chunk is allocated as a new frag.

```c
static int
__skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
	       int len, TfwStr *it, int *fragn)
{
...
	/* See if the split starts in the page fragments data. */
	for (i = 0; i < si->nr_frags; ++i) {
		const skb_frag_t *frag = &si->frags[i];
		d_size = skb_frag_size(frag);

		/*
		 * @pspt can be the end of a frag, but it can also be a start of
		 * another fragment. Not necessary, that both frags are
		 * neighbours.
		 */
		if (offset >= 0 && offset < d_size) {
			len = max_t(long, len, offset - d_size);
			ret = __split_pgfrag(skb_head, skb, i, offset, len, it, fragn);
			goto done;
		}
		/*
		 * In case another fragment does not exist, remember the
		 * fragment number, after which you need to insert/delete data.
		 */
		if (offset == d_size)
			j = i;
	}
```
__skb_fragmentkb_fragment
```

In contrast, "tfw_http_msg_add_data" determines the fragment lounge, assuming it can use the entire page where the existing fragment resides.

```c
int
tfw_http_msg_add_data(TfwMsgIter *it, TfwHttpMsg *hm, TfwStr *field,
		      const TfwStr *data)
{
	const TfwStr *c, *end;

	BUG_ON(TFW_STR_DUP(data));
	if (WARN_ON_ONCE(it->frag >= skb_shinfo(it->skb)->nr_frags))
		return -E2BIG;

	TFW_STR_FOR_EACH_CHUNK(c, data, end) {
		char *p;
		unsigned int c_off = 0, c_size, f_room, n_copy;
this_chunk:
		c_size = c->len - c_off;
		if (it->frag >= 0) {
			unsigned int f_size;
			skb_frag_t *frag = &skb_shinfo(it->skb)->frags[it->frag];

			f_size = skb_frag_size(frag);
			f_room = PAGE_SIZE - frag->bv_offset - f_size;
			p = (char *)skb_frag_address(frag) + f_size;
			n_copy = min(c_size, f_room);
			skb_frag_size_add(frag, n_copy);
			ss_skb_adjust_data_len(it->skb, n_copy);
		} else {
			f_room = skb_tailroom(it->skb);
			n_copy = min(c_size, f_room);
			p = skb_put(it->skb, n_copy);
		}
```

We can see that the latter frag extend will overwrite the unallocated area of ​​the previous frag extend. Isn't it? So is it really safe to reuse skb between two different frag extend strategies?

fw/http_frame.c
kingluo: What do you mean? Please see this [commit](https://github.com/tempesta-tech/tempesta/pull/2173/commits/372dbb0e17b1172a8e9e22af2d01008de52be96b).

If there are trailers, the body would not be empty. So the code you refer to cannot be true: `!stream->xmit.b_len`.

fw/http2.c
EvgeniiMekhanik: And also I check that you can use pool. You just should force allocate new frag in `tfw_http_msg_alloc_from_pool` and setup iterator.

```
static int
tfw_http_msg_alloc_from_pool(TfwHttpTransIter *mit, TfwPool* pool, size_t size, bool forse)
{
	if (!forse) {
		addr = tfw_pool_alloc_not_align_np(pool, size, &np);
	}
	else {
		addr = __tfw_pool_alloc_page(pool, size, /* align */ true);
		np = true;
	}
```

Also little rework of `__tfw_http_msg_expand_from_pool` is necessary in `if (unlikely(skb_room == 0 || nr_frags == MAX_SKB_FRAGS))` part of code.

EvgeniiMekhanik: Here is a branch with example of code https://github.com/tempesta-tech/tempesta/tree/MekhanikEvgenii/fix-1902-dirty

kingluo: **As I raised my doubt above, the response has three different frag allocation strategies:**
`headers (per-cpu page cache) --> body (pg_skb_alloc) --> trailers (kernel pages)`.

**So no matter use pool or not, and no matter reuse the last skb of the body to hpack trailers or not, we cannot reuse the last frag of the body.**

I make a small change to consider both skb reuse and frag distinguish, please check this [commit](https://github.com/tempesta-tech/tempesta/pull/2173/commits/3d06edbbd2a58f319e06e8851ad8ea2e08c4e330). Hopefully it satisfy your need.

---

_P.S. no need to rework the pool functions to adapt trailers, because it has the assumption of the header-body order, if we rework it, the functions will be too much fan in, which seems not a good coding practice. And no offense, but your example code requires much more work to get it to work, where I also tried before._

fw/cache.c
kingluo: Remove the skb_head manipulation loop: [commit](https://github.com/tempesta-tech/tempesta/pull/2173/commits/6bcd9e49065c028403a1c93b5722e3cbaea366d3).

fw/http2.c
EvgeniiMekhanik: I don't see problems in different allocation strategies. @const-t please pay attention on this place. Moreover in my example of usage pool  we don't reuse the last frag of the body, but allocate new one. And in my example we don't need a big part of code in `tfw_hpack_hdr_expand`. Much more work is not a problem as I think.

fw/hpack.c
EvgeniiMekhanik: We don't need this part of code if we use pool.

fw/cache.c
EvgeniiMekhanik: I try to rewrite cache functions to make them able to work with trailers in the proper way next Friday (review day).

fw/http_frame.c
kingluo: It's not due to these statements, but `ss_skb_cutoff_data`.

```
[10845.573649] Call Trace:
[10845.575340]  <IRQ>
[10845.577160]  dump_stack+0x96/0xc4
[10845.578844]  skb_fragment.cold+0x5/0x31 [tempesta_fw]
[10845.580546]  ss_skb_cutoff_data+0x696/0x6e0 [tempesta_fw]
[10845.582240]  ? ss_skb_list_chop_head_tail+0x1d0/0x1d0 [tempesta_fw]
[10845.584021]  ? __asan_storeN+0x12/0x20
[10845.585705]  ? eb_delete+0x203/0x210 [tempesta_lib]
[10845.587407]  tfw_http_msg_cutoff_body_chunks+0x6d/0xc0 [tempesta_fw]
[10845.589112]  tfw_h2_stream_xmit_prepare_resp+0x3d8/0x410 [tempesta_fw]
[10845.590800]  tfw_h2_make_frames+0x7dc/0x9d0 [tempesta_fw]
[10845.592491]  ? tcp_established_options+0x1ab/0x330
[10845.594184]  ? tfw_h2_frame_process+0xc80/0xc80 [tempesta_fw]
[10845.595847]  ? bpf_skops_hdr_opt_len+0x210/0x210
[10845.597516]  tfw_sk_fill_write_queue+0x323/0x3e0 [tempesta_fw]
[10845.599158]  ? tfw_cli_conn_release+0xc0/0xc0 [tempesta_fw]
[10845.600788]  ? tfw_h2_sched_activate_stream+0xc8/0x190 [tempesta_fw]
[10845.602427]  ? ss_skb_tcp_entail_list+0x25/0x1a0 [tempesta_fw]
[10845.604115]  ss_tx_action+0xacf/0xc90 [tempesta_fw]
[10845.605703]  ? ss_skb_tcp_entail_list+0x1a0/0x1a0 [tempesta_fw]
[10845.607384]  ? __kernel_fpu_begin_mask+0x9a/0x150
[10845.609071]  net_tx_action+0xfa/0x3c0
[10845.610741]  __do_softirq+0xfe/0x383
[10845.612321]  asm_call_irq_on_stack+0x12/0x20
[10845.613910]  </IRQ>

```

```c
static int
skb_fragment(struct sk_buff *skb_head, struct sk_buff *skb, char *pspt,
             int len, TfwStr *it, int *fragn)
{
        if (unlikely(abs(len) > PAGE_SIZE)) {
                T_WARN("Attempt to add or delete too much data: %u\n", len);
                return -EINVAL;
        }
...
```

fw/cache.c
kingluo: Ok, looking forward to your suggested commits which I can then push directly, thanks.

fw/http2.c
kingluo: [done](https://github.com/tempesta-tech/tempesta/pull/2173/commits/bb1589a70478b69aff2a7a455c72da865a0ec4bd).

fw/hpack.c
kingluo: [ditto](https://github.com/tempesta-tech/tempesta/pull/2173/commits/bb1589a70478b69aff2a7a455c72da865a0ec4bd).

fw/http_frame.c
kingluo: No matter whether one trailer header is too long or too much trailer headers, as long as the size of the body plus trailers exceeds a `PAGE_SIZE`, it will fail at `tfw_http_msg_cutoff_body_chunks`, so this is not the bug of this PR, instead, we need to fix it individually.

I hacked the trailer `HEADER_FRAME` size to 100 bytes so the other trailer data could get packed into `CONTINUATION_FRAME`. It passes the test. So the trailer continuation framing logic has no problem.

Please resolve this conversation.

```diff
diff --git a/fw/http_frame.c b/fw/http_frame.c
index 94ccee5e..b6a71d5f 100644
--- a/fw/http_frame.c
+++ b/fw/http_frame.c
@@ -2114,7 +2114,7 @@ do {                                                                      \
        T_FSM_STATE(HTTP2_MAKE_TRAILER_FRAMES) {
                CALC_SND_WND_AND_SET_FRAME_TYPE(HTTP2_HEADERS);
                r = tfw_h2_insert_frame_header(sk, ctx, stream, frame_type,
-                                              snd_wnd, stream->xmit.t_len);
+                                              snd_wnd, 100);
                if (unlikely(r)) {
                        T_WARN("Failed to make trail headers frame %d", r);
                        return r;

```

```diff
diff --git a/encoding/test_encoding.py b/encoding/test_encoding.py
index 5bf2662..bda4e0b 100644
--- a/encoding/test_encoding.py
+++ b/encoding/test_encoding.py
@@ -526,8 +526,12 @@ class TestH2ChunkedWithTrailer(tester.TempestaTest, CommonUtils):

     def setUp(self):
         self.backends = copy.deepcopy(self.backends_template)
+        trailers = f"X-Token: {self.token}\r\n"
+        for i in range(100):
+            trailers += f"foobar-{i}: {'a'*10}\r\n"
+        trailers += "\r\n"
         self.backends[0]["response_content"] += (
-            self.encode_chunked(BODY_PAYLOAD, CHUNK_SIZE)[:-2] + f"X-Token: {self.token}\r\n\r\n"
+            self.encode_chunked(BODY_PAYLOAD, CHUNK_SIZE)[:-2] + f"{trailers}"
         )
         super().setUp()

```

```bash
$ t encoding.test_encoding.TestH2ChunkedWithTrailer
Root privileges are required: need access to module loading on localhost.

Running functional tests...

test (encoding.test_encoding.TestH2ChunkedWithTrailer) ... ok

Ran 1 test in 2.308s

OK

```

```
[73739.183457] [tempesta fw] Tempesta FW is ready
[73739.235065] ---> HTTP2_MAKE_TRAILER_CONTINUATION_FRAMES
```

fw/cache.c
kingluo: @EvgeniiMekhanik Done. https://github.com/tempesta-tech/tempesta/pull/2173/commits/4932ecbd306c5cdc3e80d8198ab6d2de869a9389

fw/http_frame.c
```diff
@@ -1931,13 +1939,14 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
...
 	int r;
+	char flags;
 
 
 	/*
 	 * Very unlikely case, when skb_head and one or more next skbs
 	 * are empty because of transformation during making HEADERS.
 	 */
-	if (type == HTTP2_CONTINUATION || type == HTTP2_DATA) {
+	if (type == HTTP2_CONTINUATION || type == HTTP2_DATA || !stream->xmit.h_len) {
```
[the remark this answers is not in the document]
kingluo: yes, but that script was added after this PR.

fw/cache.c
```diff
@@ -2838,13 +2860,16 @@ tfw_cache_build_resp(TfwHttpReq *req, TfwCacheEntry *ce, long lifetime)
 	if (tfw_cache_set_status(db, ce, resp, &trec, &p, &h_len))
 		goto free;
 
-	for (h = TFW_HTTP_HDR_REGULAR; h < ce->hdr_num; ++h) {
-		bool skip = !TFW_MSG_H2(req) && (h >= ce->hdr_h2_off);
+	trailer_off = TFW_MSG_H2(req) ? ce->trailer_off : ce->hdr_num;
+	for (h = TFW_HTTP_HDR_REGULAR; h < trailer_off; ++h) {
+		bool skip = !TFW_MSG_H2(req) && (h >= ce->hdr_h2_off)
+			     && (h < ce->trailer_off);
```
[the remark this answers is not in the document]
kingluo: because `trailer_off` can be `ce->hdr_num`.

fw/cache.c
```diff
@@ -2155,6 +2161,21 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 	ce->hdr_h2_off = ce->hdr_num + 1;
 	ce->hdr_num += 2;
 
+	ce->trailer_off = ce->hdr_num;
+	ce->trailers = TDB_OFF(db->hdr, p);
+	FOR_EACH_HDR_FIELD_FROM(field, end1, resp, TFW_HTTP_HDR_REGULAR) {
```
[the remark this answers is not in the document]
kingluo: I tried one loop before, but the previous comment said it brings in too many conditions.

fw/cache.c
```diff
@@ -2155,6 +2161,21 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
...
+		int hid = field - resp->h_tbl->tbl;
+
+		if (!(field->flags & TFW_STR_TRAILER))
```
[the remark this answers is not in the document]
kingluo: These conditions are non-sense to trailers, so I removed them.

fw/http_frame.c
kingluo: done: https://github.com/tempesta-tech/tempesta/pull/2173/commits/7c4107e41c246a346d7edb965a039e6d0e247285

fw/http_frame.c
EvgeniiMekhanik: It is not correct you should send postponed frames after sending headers frames, when h_len == 0 (before you send body) and after you send trailer frames (t_len == 0). Otherwise you can block all other frames until body and trailers will be send (body can be very long)

fw/http2.c
kingluo: Can you re-think if it's ok to reuse the skb/frag? They use different allocations, if we mix them, we may override something. Please check my comments above. Anyway, if you're all 100% sure, I'll not insist, then please go ahead with one of your solutions.
`headers (pool, i.e. per-cpu page cache) --> body (pg_skb_alloc) --> trailers (kernel pages)`

fw/cache.c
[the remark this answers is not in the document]
kingluo: Can you give some suggestions? And it's the same with your comment above?

fw/http_frame.c
kingluo: I don't get your words. Can you make a suggestion change directly? Thanks.

fw/cache.c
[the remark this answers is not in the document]
const-t: Why are they non-sense? Can you guarantee that hop-by-hop headers from trailer will not be saved into the cache?

Should `cache_resp_hdr_del`  be applicable to trailer headers?

fw/cache.c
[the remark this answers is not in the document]
const-t: I'm suggesting:
```
if (resp->trailers_len > 0) {
    FOR_EACH_HDR_FIELD_FROM(field, end1, resp, TFW_HTTP_HDR_REGULAR) {
    ...
    }
}
```

fw/http2.c
const-t: I've seen this patch. However, from my point of view, would be better to do it another way, without changing `tfw_http_msg_expand_data()` and without code duplication in `tfw_hpack_hdr_expand()`. We must using pool and use `tfw_hpack_transform()`. There are few reasons why it's better:
1. We don't change common purpose function `tfw_http_msg_expand_data()` only for trailers case.
2. We don't add [this code](https://github.com/tempesta-tech/tempesta/blob/bfe4c9cab5075d8ced311571299e308cf780b660/fw/hpack.c#L3492).
3. We could use dynamic table, because now we can't do it.
4. Preparation of pool will be located in same place where we do other work with trailers.

fw/cache.c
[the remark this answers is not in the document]
kingluo: I can't refer to any standard specification, but in practice, trailers are user data just like the body. The most typical usage is in scenarios like GRPC (for HTTP/1.1, they are converted one-to-one from GRPC). Also, they should not be modified by proxies, just like the body should be transmitted as is.

But yes, I can add back the conditions if needed.

fw/http2.c
kingluo: Fine with me. Submit your commit directly or provide suggested changes please, thank you.

fw/cache.c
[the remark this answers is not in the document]
kingluo: done: https://github.com/tempesta-tech/tempesta/pull/2173/commits/7290143ded06a2f8175773dc2e34901fabea2388

fw/http2.c
EvgeniiMekhanik: I fix this part of code. @const-t can you look on it?

const-t (approved):
LGTM, but need minor fixes. I have to add, that `resp_hdr_set`(delete/replace header) will to work with trailers. Could you please describe it on the wiki page of `resp_hdr_set`.

fw/http2.c
```diff
@@ -485,8 +485,27 @@ tfw_h2_stream_xmit_prepare_resp(TfwStream *stream)
...
+		if (unlikely(r)) {
+			T_WARN("Failed to encode body");
+			goto finish;
+		}
+	}
+
+	if (resp->trailers_len > 0) {
+		unsigned long acc = resp->mit.acc_len;
+		resp->mit.iter.skb = resp->msg.skb_head->prev;
+		resp->mit.iter.frag = skb_shinfo(resp->msg.skb_head->prev)->nr_frags - 1;
```
krizhanovsky, line 499: Too long line. No copyright updates.

Please enable git hooks with the coding style checks and automatic copyrights!

fw/http_msg.h
```diff
@@ -2,7 +2,7 @@
  *		Tempesta FW
  *
  * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
- * Copyright (C) 2015-2023 Tempesta Technologies, Inc.
+ * Copyright (C) 2015-2024 Tempesta Technologies, Inc.
```
krizhanovsky, line 5: The only file with update copyright is the file, which wasn't updated :)

krizhanovsky (changes requested):
With https://github.com/tempesta-tech/tempesta/issues/1902#issuecomment-2091286028 I meant to *remove* the current logic of a trailer header movement to normal headers, even for HTTP/1, but with this PR I see a lot of new code, but not code removals.

The code complexity grows making the performance worse and maintenance harder.

Also now we still have the problem that `Trailer: X-Token` remains in headers.

Lastly, please resolve all comments which are actually resolved.

const-t (changes requested):

fw/http.c
```diff
@@ -673,7 +674,7 @@ tfw_h2_prep_resp(TfwHttpResp *resp, unsigned short status, TfwStr *msg)
 
 	body = TFW_STR_BODY_CH(msg);
 
-	r = tfw_h2_frame_local_resp(resp, hdrs_len, body);
+	r = tfw_h2_frame_local_resp(resp, hdrs_len, body, h2_mode);
```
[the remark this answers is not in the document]
EvgeniiMekhanik: `tfw_h2_frame_local_resp` is called from `tfw_cache_build_resp`, so we should check h2_mode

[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed

fw/cache.c
```diff
@@ -2690,8 +2692,11 @@ tfw_cache_add_body_page(TfwMsgIter *it, char *p, int sz, bool h2)
  */
 static int
 tfw_cache_build_resp_body(TDB *db, TdbVRec *trec, TfwMsgIter *it, char *p,
-			  unsigned long body_sz, bool h2)
+			  unsigned long body_sz, bool h2, bool chunked_body)
 {
+#define S_CRLF "\r\n"
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed

fw/cache.c
```diff
@@ -2107,7 +2107,8 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 			continue;
 		}
 
-		if (hid == TFW_HTTP_HDR_TRANSFER_ENCODING) {
+		if (hid == TFW_HTTP_HDR_TRANSFER_ENCODING
+		    && TFW_MSG_H2(req)) {
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed implement new Cstr flags

fw/cache.c
```diff
@@ -2099,6 +2099,7 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 		 */
 		if ((field->flags & (TFW_STR_HBH_HDR | TFW_STR_NOCCPY_HDR |
 				     TFW_STR_TRAILER))
+		    || (field->flags & TFW_STR_TRAILER_HDR && TFW_MSG_H2(req))
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed implement new Cstr flags

const-t (approved):
LGTM

fw/cache.c
```diff
@@ -2155,6 +2160,24 @@ tfw_cache_copy_resp(TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 	ce->hdr_h2_off = ce->hdr_num + 1;
 	ce->hdr_num += 2;
 
+	ce->trailer_off = ce->hdr_num;
+	ce->trailers = TDB_OFF(db->hdr, p);
+	if (resp->trailers_len > 0) {
+		FOR_EACH_HDR_FIELD_FROM(field, end1, resp, TFW_HTTP_HDR_REGULAR) {
+			int hid = field - resp->h_tbl->tbl;
+
+			if (!(field->flags & TFW_STR_TRAILER))
```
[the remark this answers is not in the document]
EvgeniiMekhanik: @krizhanovsky  how do you think? I think that trailers are additional headers which should not be checked. Please answer here.

fw/http_parser.c
```diff
@@ -2492,6 +2492,133 @@ __req_parse_transfer_encoding(TfwHttpMsg *hm, unsigned char *data, size_t len)
 	return __parse_transfer_encoding(hm, data, len, true, false);
 }
 
+static int
+__resp_parse_trailer(TfwHttpMsg *hm, unsigned char *data, size_t len)
```
const-t, line 2496: From my point of view, too much efforts to support rarely used feature. I would suggest to drop `Trailer` header at all even without parsing for both directions, downstream and upstream. Accordingly to [RFC](https://datatracker.ietf.org/doc/html/rfc9110#section-6.6.2) `Trailer` is just a hint, that doesn't give any guaranties and implementation shouldn't rely on it.


> A sender that intends to generate one or more trailer fields in a message **SHOULD** generate a [Trailer](https://datatracker.ietf.org/doc/html/rfc9110#field.trailer) header field in the header section of that message to indicate which fields might be present in the trailers

> though there is no guarantee that a sender of Trailer will always follow through by sending the named fields

const-t (changes requested):
Needed small changes and discussion

fw/http_parser.c
const-t, line 2496: Seems we allow duplicated `Trailer` header, though RFC doesn't contain direct prohibitions this looks like confusion and we must not allow duplicated `Trailer` header.

fw/http_frame.c
```diff
@@ -1883,23 +1883,28 @@ tfw_h2_calc_frame_length(TfwH2Ctx *ctx, TfwStream *stream, TfwFrameType type,
...
-			(stream->xmit.b_len ? HTTP2_F_END_HEADERS :
-			 HTTP2_F_END_HEADERS | HTTP2_F_END_STREAM);
-	case HTTP2_CONTINUATION:
-		return stream->xmit.h_len ? 0 : HTTP2_F_END_HEADERS;
-	case HTTP2_DATA:
-		return stream->xmit.b_len ? 0 : HTTP2_F_END_STREAM;
-	default:
-		BUG();
-	};
+	char flags = 0;
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed

fw/http_frame.c
```diff
@@ -1919,16 +1924,18 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
 	unsigned int mark = stream->xmit.skb_head->mark;
 	unsigned int max_len = (*snd_wnd > TLS_MAX_PAYLOAD_SIZE + TLS_MAX_OVERHEAD) ?
 		TLS_MAX_PAYLOAD_SIZE : *snd_wnd - TLS_MAX_OVERHEAD;
+	bool trailers = false;
 	unsigned int length;
 	char *data;
 	int r = 0;
-
+	char flags;
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed

fw/http_stream.c
```diff
@@ -320,7 +321,7 @@ tfw_h2_stream_add_closed(TfwH2Ctx *ctx, TfwStream *stream)
  */
 TfwStreamFsmRes
 tfw_h2_stream_fsm(TfwH2Ctx *ctx, TfwStream *stream, unsigned char type,
-		  unsigned char flags, bool send, TfwH2Err *err)
+		  char flags, bool send, TfwH2Err *err)
```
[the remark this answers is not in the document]
EvgeniiMekhanik: yes fixed

fw/http_msg.c
```diff
@@ -1098,16 +1132,48 @@ tfw_http_msg_del_hbh_hdrs(TfwHttpMsg *hm)
...
 }
 
+int
+tfw_http_msg_adjust_trailer_hdr(TfwHttpMsg *hm)
+{
+	TfwHttpHdrTbl *ht = hm->h_tbl;
+	int r = 0;
+
+	if (unlikely(hm->trailer_hid)) {
+		TfwStr *hdr = &ht->tbl[hm->trailer_hid];
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Remove this part of code

fw/http_parser.c
EvgeniiMekhanik: Remove this part of code

fw/http_parser.c
```diff
@@ -5283,6 +5327,15 @@ tfw_http_parse_req(void *req_data, unsigned char *data, unsigned int len,
 				p += 17;
 				__FSM_MOVE_hdr_fixup(RGen_LWS, 1);
 			}
+			if (likely(__data_available(p, 8))
+				   && C8_INT_LCM(p, 't', 'r', 'a', 'i',
+						    'l', 'e', 'r', ':'))
+			{
+				__msg_hdr_chunk_fixup(data, __data_off(p + 8));
+				parser->_i_st = &&Req_HdrTrailerV;
+				p += 8;
```
const-t, line 5336: For HTTP1 if there is no whitespaces we parse headers into 3 chunks: [name][:][value], some functions may rely on that. But here we will have [name:][v][alue]. Or if there is whitespace we will have: [name:][ ][value] and in this case we lose flag for whitespace. Let's do this consistent to other headers, `p += 7` and let `__FSM_MOVE_hdr_fixup(RGen_LWS, 1);` to fixup whitespace.

fw/http_parser.c
```diff
@@ -11172,6 +11245,14 @@ __resp_parse_cache_control(TfwHttpResp *resp, unsigned char *data, size_t len)
 
 	__FSM_START(parser->_i_st);
 
+	/*
+	 * According RFC 7230 4.1.2:
+	 * A sender MUST NOT generate a trailer that contains response
+	 * control data (e.g., see Section 7.1 of [RFC7231]).
+	 */
+	if (unlikely(test_bit(TFW_HTTP_B_HEADERS_PARSED, resp->flags)))
```
const-t, line 11253: Let's move this condition to `Resp_I_CC_start` just to not walk through it on each function call, on segmentation. In other functions as well, where it is possible.

const-t (approved):
LGTM, but need minor changes.

fw/http_parser.c
EvgeniiMekhanik: Yes fixed

fw/http_parser.c
EvgeniiMekhanik: Yes fixed

fw/cache.c
```diff
@@ -2818,16 +2876,50 @@ tfw_cache_build_resp_body(TDB *db, TdbVRec *trec, TfwMsgIter *it, char *p,
...
+	if (unlikely(!body_sz))
+		goto add_zero_chunk;
+
+	if (chunked_body) {
+		char cstr_blen[TFW_ULTOA_BUF_SIZ] = {0};
+		/*
+		 * Encode body in one single chunk.
+		 * blen\r\nbody
+		 */
+		TfwStr b_len = {
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed

fw/http2.c
```diff
@@ -513,14 +513,36 @@ tfw_h2_stream_xmit_prepare_resp(TfwStream *stream)
...
+		}
+
 		stream->xmit.b_len = TFW_HTTP_RESP_CUT_BODY_SZ(resp);
 		/*
 		 * Response is chunked encoded, but it is not a response
 		 * on HEAD request.
 		 */
 		if (test_bit(TFW_HTTP_B_CHUNKED, resp->flags)
-		    && !test_bit(TFW_HTTP_B_VOID_BODY, resp->flags))
+		    && !test_bit(TFW_HTTP_B_VOID_BODY, resp->flags)) {
```
krizhanovsky, line 539:
```suggestion
		    && !test_bit(TFW_HTTP_B_VOID_BODY, resp->flags))
		{
```

fw/http_frame.c
```diff
@@ -1883,23 +1883,28 @@ tfw_h2_calc_frame_length(TfwH2Ctx *ctx, TfwStream *stream, TfwFrameType type,
...
+	if (!stream->xmit.b_len && !stream->xmit.t_len
+	    && (type == HTTP2_HEADERS || type == HTTP2_DATA)
+	    && !tfw_h2_stream_is_eos_sent(stream))
+		flags |= HTTP2_F_END_STREAM;
+
+	if (!stream->xmit.b_len && stream->xmit.t_len
+	    && type == HTTP2_HEADERS && trailers)
+		flags |= HTTP2_F_END_STREAM;
+
+
```
krizhanovsky, line 1900: extra empty line

fw/http_parser.c
```diff
@@ -467,6 +467,7 @@ __FSM_STATE(st, cold) {							\
 	__FSM_JMP(RGen_HdrOtherN);					\
 }
 
+
```
krizhanovsky, line 470: extra empty line

krizhanovsky (approved):
LGTM, I didn't see any issues, just some cleanups are required

fw/cache.c
```diff
@@ -196,6 +201,16 @@ __tfw_dbg_dump_ce(const TfwCacheEntry *ce)
...
 #define TFW_CSTR_SPEC_IDX	0x2
+/*
+ * TfwCStr contains header, which should be skipped for http2 responses
+ * constructed from cache.
+ */
+#define TFW_CSTR_SKIP_FOR_H2	0x4
+
+/*
+ * TfwCStr contains trailer.
+ */
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed

fw/cache.c
```diff
@@ -989,37 +1017,58 @@ do {									\
...
+					|| (type == TFW_CACHE_TRAILER
+					    && (s->flags & TFW_CSTR_TRAILER))));
+		}
+
 		*p += TFW_CSTR_HDRLEN;
 		r = write_actor(db, trec, resp, p, s->len, &dc_iter);
-		if (unlikely(r))
-			break;
+			if (unlikely(r))
+				break;
```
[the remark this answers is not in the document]
EvgeniiMekhanik: yes fixed

fw/http_parser.c
EvgeniiMekhanik: Yes fixed

fw/http_frame.c
EvgeniiMekhanik: Yes fixed

fw/http2.c
EvgeniiMekhanik: Yes fixed

fw/http.h
```diff
@@ -764,6 +766,7 @@ int tfw_h2_resp_add_loc_hdrs(TfwHttpResp *resp, const TfwHdrMods *h_mods,
 int tfw_h2_resp_status_write(TfwHttpResp *resp, unsigned short status,
 			     bool use_pool, bool cache);
 int tfw_h2_resp_encode_headers(TfwHttpResp *resp);
+int tfw_h2_hpack_encode_trailer_headers(TfwHttpResp *resp);
```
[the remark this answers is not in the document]
EvgeniiMekhanik: yes fixed

fw/cache.c
```diff
@@ -2946,6 +3056,7 @@ tfw_cache_build_resp(TfwHttpReq *req, TfwCacheEntry *ce, long age)
 	if (!(resp = tfw_http_msg_alloc_resp(req)))
 		goto out;
 
+
```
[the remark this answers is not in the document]
EvgeniiMekhanik: yes fixed

fw/http.c
```diff
@@ -5099,6 +5080,47 @@ tfw_h2_hpack_encode_headers(TfwHttpResp *resp, const TfwHdrMods *h_mods)
...
+		if (hid == TFW_HTTP_HDR_SERVER)
+			continue;
+
+		r = tfw_hpack_transform(resp, tgt);
+		if (unlikely(r))
+			return r;
+	}
+
+	return 0;
+}
```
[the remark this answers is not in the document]
EvgeniiMekhanik: yes fixed

EvgeniiMekhanik (approved):

PR #2174, opened 2024-07-18
Do not remove/insert exclusive streams

EvgeniiMekhanik:
- There is no sence to remove/insert streams which are exclusive on its level of dependency
tree, because they don't share badwidth with
any stream.
- Stop and delete exclusive stream after sending all pending data immediately.


EvgeniiMekhanik:
Small PR, implemented one point from 1411

fw/http_stream_sched.c
```diff
@@ -447,11 +450,18 @@ tfw_h2_remove_stream_dep(TfwStreamSched *sched, TfwStream *stream)
 		 * weights and add them to the scheduler of the parent of
 		 * the removed stream.
 		 */
-		new_weight = child->weight *
-			stream->weight / total_weight;
-		child->weight = new_weight > 0 ? new_weight : 1;
-		deficit = !parent_has_children ?
-			child->sched_node.key : stream->sched_node.key;
+		if (parent_has_children) {
+			new_weight = child->weight *
```
const-t, line 454: We doing the same calculations in two places, let's extract it into static function.

fw/http_frame.c
```diff
@@ -2187,9 +2204,23 @@ tfw_h2_make_frames(struct sock *sk, TfwH2Ctx *ctx, unsigned long snd_wnd,
...
+			tfw_h2_sched_stream_enqueue(sched, stream, parent,
+						    deficit);
+		} else if (!tfw_h2_stream_is_active(stream)) {
+			tfw_h2_sched_deactivate_stream(sched, stream);
+			/*
+			 * Remove exclusive stream after sending all pending
+			 * data.
+			 */
+			if (!stream->xmit.skb_head)
+				tfw_h2_stream_clean(ctx, stream);
```
const-t, line 2222: In `tfw_h2_sched_deactivate_stream` we add stream into `blocked` queue and here we clean the stream, looks suspicious.

fw/http_stream_sched.c
EvgeniiMekhanik: Yes fixed

fw/http_frame.c
EvgeniiMekhanik: We should do it. We should deactivate stream for correct counting of active streams in scheduler. and we should clean stream for memory freeing.

const-t: Oh, I see we do `tfw_h2_stream_sched_remove()` before freeing the stream.

const-t (approved):
LGTM

fw/http_frame.c
```diff
@@ -2169,9 +2168,14 @@ do {									\
 		 */
 		if (unlikely(stream->xmit.skb_head))
 			ss_skb_tcp_entail_list(sk, &stream->xmit.skb_head);
-		tfw_h2_stream_add_closed(ctx, stream);
 		if (stream == ctx->error)
 			ctx->error = NULL;
+		/*
+		 * Don't put exclusive streams in closed queue
+		 * remove it later immediately.
```
krizhanovsky: Could you please fix the comment - the second line is absolutely unclear

fw/http_frame.c
```diff
@@ -2208,18 +2211,32 @@ tfw_h2_make_frames(struct sock *sk, TfwH2Ctx *ctx, unsigned long snd_wnd,
...
 
-	while (tfw_h2_stream_sched_is_active(&sched->root)
+#define SCHED_REMOVE_EXCLUSIVE_STREAM(sched, stream)			\
+do {									\
+	if (!tfw_h2_stream_is_exclusive(stream)) {			\
+		parent = stream->sched.parent;				\
+		tfw_h2_stream_sched_remove(sched, stream);		\
+	} else {							\
+		parent = NULL;						\
+	}								\
```
krizhanovsky, line 2221: The macro is named `SCHED_REMOVE_EXCLUSIVE_STREAM`, but `tfw_h2_stream_sched_remove()` is called for `!tfw_h2_stream_is_exclusive(stream)`

fw/http_frame.c
```diff
@@ -2230,9 +2247,23 @@ tfw_h2_make_frames(struct sock *sk, TfwH2Ctx *ctx, unsigned long snd_wnd,
...
-		tfw_h2_sched_stream_enqueue(sched, stream, parent, deficit);
+		r = tfw_h2_stream_xmit_process(sk, ctx, stream, !parent,
+					       &snd_wnd);
+
+		/* We don't recalculate deficits of exclusive streams. */
+		if (parent) {
+			deficit = tfw_h2_stream_recalc_deficit(stream);
+			tfw_h2_sched_stream_enqueue(sched, stream, parent,
+						    deficit);
+		} else if (!tfw_h2_stream_is_active(stream)) {
```
krizhanovsky, line 2258: Could you please place `else if ()` on separate line to make the conditions separate from the code above and more readable?

krizhanovsky (approved):
LGTM, just a couple of cleanups are good to do before the merge

fw/http_frame.c
EvgeniiMekhanik: Yes fixed

fw/http_frame.c
EvgeniiMekhanik: Yes fixed

fw/http_frame.c
EvgeniiMekhanik: In all other places we place `else if` on the same line. Are you sure we should do it here?

PR #2176, opened 2024-07-19
db: File mapping fix

const-t:
`kernel_read()` is able to read at most MAX_RW_COUNT bytes. Break the reading of a large file into a few steps.

### DIFF
--- a/db/core/file.c
+++ b/db/core/file.c
@@ -4,7 +4,7 @@
  * File mapping and IO.
  *
  * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
- * Copyright (C) 2015-2021 Tempesta Technologies.
+ * Copyright (C) 2015-2024 Tempesta Technologies.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
@@ -166,10 +166,11 @@ ma_free(unsigned long addr, int node)
  * Map file to reserved set of unswappable pages.
  */
 static unsigned long
-tempesta_map_file(struct file *file, unsigned long len, int node)
+tempesta_map_file(struct file *file, size_t len, int node)
 {
 	MArea *ma;
 	loff_t off = 0;
+	size_t copied = 0;
 	unsigned long addr = -ENOMEM;
 
 	BUG_ON(len & ~TDB_EXT_MASK);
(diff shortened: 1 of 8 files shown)


const-t:
Unfortunately I forgot to implement `M/G` prefixes for tempesta's config. It will be on my TODO list.

const-t:
Related kernel fix https://github.com/tempesta-tech/linux-5.10.35-tfw/pull/28

db/core/file.c
```diff
@@ -194,14 +195,27 @@ tempesta_map_file(struct file *file, unsigned long len, int node)
...
+		if (bytes < 0) {
+			addr = bytes;
+			TDB_ERR("Cannot read %li bytes to addr %p, ret = %ld\n",
+				to_read, (void *)ma->start + copied, addr);
+			fput(file);
+			__ma_free(ma);
+			goto err;
+		}
+
+		if (bytes == 0)
```
EvgeniiMekhanik, line 216: It seems wrong. If we bytes == 0 before we read all (copied < len) we should return error

EvgeniiMekhanik (changes requested):
Small fix

EvgeniiMekhanik (approved):

krizhanovsky (approved):
LGTM

fw/cfg.c
```diff
@@ -1230,6 +1230,44 @@ tfw_cfg_get_attr(const TfwCfgEntry *e, const char *attr_key,
...
+ */
+int
+tfw_cfg_mem_check_range(const char *mem, const char *mem_min,
+			const char *mem_max)
+{
+	unsigned long value = memparse(mem, NULL);
+	unsigned long min = memparse(mem_min, NULL);
+	unsigned long max = memparse(mem_max, NULL);
+
+	if (min != max && (value < min || value > max)) {
```
EvgeniiMekhanik, line 1245: Why we check that min != max ? If min == max I think we should also check value

fw/cfg.c
```diff
@@ -1771,6 +1809,39 @@ tfw_cfg_set_str(TfwCfgSpec *cs, TfwCfgEntry *e)
...
+
+	if (tfw_cfg_check_single_val(e))
+		return -EINVAL;
+
+	/* Check value restrictions if we have any in the spec extension. */
+	cse = cs->spec_ext;
+	if (cse) {
+		int r;
+
+		r = tfw_cfg_mem_check_multiple_of(e->vals[0],
```
EvgeniiMekhanik, line 1829: as for me better
if ((r =  tfw_cfg_mem_check_multiple_of(..))
    || (r =  tfw_cfg_mem_check_range()))
     return r;

EvgeniiMekhanik: And also may be better to call val = memparse(e->vals[0], NULL); before and then pass val to tfw_cfg_mem_check_multiple_of? Because you call memparse in checking functions two extra times

fw/cfg.c
const-t: Fixed

fw/cfg.c
const-t: > And also may be better to call val = memparse(e->vals[0], NULL); before and then pass val to tfw_cfg_mem_check_multiple_of? Because you call memparse in checking functions two extra times

No there is no sense, because in `tfw_cfg_mem_check_multiple_of()` we'll have to do "reverse" `memparse` from 
`val` to find correct prefix for error message.

> as for me better
> if ((r = tfw_cfg_mem_check_multiple_of(..))
> || (r = tfw_cfg_mem_check_range()))
> return r;

If you wish, I'll change it, for me the same.

fw/cfg.c
```diff
@@ -1771,6 +1809,35 @@ tfw_cfg_set_str(TfwCfgSpec *cs, TfwCfgEntry *e)
...
+int
+tfw_cfg_set_mem(TfwCfgSpec *cs, TfwCfgEntry *e)
+{
+	long *dest_long;
+	unsigned long val;
+	TfwCfgSpecMem *cse;
+
+	BUG_ON(!cs->dest);
+
+	if (tfw_cfg_check_single_val(e))
```
EvgeniiMekhanik, line 1821: we decide TFW_CFG_CHECK_VAL_N every time

EvgeniiMekhanik (approved):

fw/cfg.c
const-t: In this case `tfw_cfg_check_single_val()` is suitable, because does few checks.

Issue #2178, opened 2024-07-23
BUG in sock.c


EvgeniiMekhanik:
11483.898309] [tempesta fw] Warning: frang: http_resp_code_block limit exceeded for 127.0.0.2: 2 (lim=1)
[11483.900427] [tempesta fw] Warning: frang: http_resp_code_block limit exceeded for 127.0.0.2: 3 (lim=1)
[11483.900431] [tempesta fw] Warning: response blocked: filtered out: 127.0.0.2
[11483.901084] [tempesta fw] Warning: response blocked: filtered out: 127.0.0.2
[11483.902606] [tempesta tls] Warning: Cannot send TLS alert 0:1, -9
[11483.903024] [tempesta fw] Warning: Close TCP socket w/o sending alert to the peer: 127.0.0.2
[11483.903667] ------------[ cut here ]------------
[11483.903994] kernel BUG at /home/evgeny/workdir/tempesta/fw/sock.c:517!
[11483.904472] invalid opcode: 0000 [#1] SMP NOPTI
[11483.904843] CPU: 7 PID: 0 Comm: swapper/7 Tainted: G           OE     5.10.35+ #298
[11483.905717] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.16.2-debian-1.16.2-1 04/01/2014
[11483.906479] RIP: 0010:ss_send+0x2c4/0x320 [tempesta_fw]

I catch this BUG during testing https://github.com/tempesta-tech/tempesta/pull/2139, but it seems that it is not problem in this branch. Test from https://github.com/tempesta-tech/tempesta-test/pull/598/commits/1f7290ebab8fcfc7701583bad7c90706bb9345ee http_resp_code_block_2 (reproduced only once)

EvgeniiMekhanik:
I reproduce this BUG on current master. Just run this test in the loop

EvgeniiMekhanik:
It seems we have two problems here:
1. When we process response `tfw_http_resp_process` and catch error we call `tfw_http_req_block` we close client connection in this function. We can also receive FIN from the client during calling  `tfw_http_req_block` and totally destroy connection. If we processing another response for this connection on other CPU we can catch a lot of bugs in different places.
2. Another problem that after calling ss_conn_drop_guard_exit we call tfw_classify_conn_close immediatly even if connection still have reference counter, so if we call some frang functions for this connection on other cpu we catch a bug

PR #2182, opened 2024-07-24
Fix bug with accessing already released connection

EvgeniiMekhanik:
There was a problem with accessing already released connection when we send error response to client,
after error in processing http response. Reponse
processing can occurs on different cpus (moreover
we can process several responses in parallel on
several cpus). Problem we fail to process two or
more responses in parallel.
We call tfw_h2_error_resp->tfw_h2_send_err_resp
and free request inside `tfw_h2_send_err_resp`
and also decrement connection reference counter
because of it. Also we call `tfw_h2_conn_terminate_close` and send FIN to the client. If FIN from the client comes immediatly `ss_conn_drop_guard_exit` will
be called on another cpu in parallel. For one
response everything is ok, we drop connection
in `ss_conn_drop_guard_exit` and release it, but
if we also call `tfw_h2_error_resp` for second response we release connection inside `tfw_h2_send_err_resp` after sending response and access already deleted connection in `tfw_h2_conn_terminate_close`!
- To fix this problem we should increment connection reference counter before `tfw_h2_error_resp` and
release it after `ss_conn_drop_guard_exit`.
- Also we have a problem with `sk->security` - we release this pointer in `ss_conn_drop_guard_exit` but this function can be called when connection stil has active processed responses and we can pass this responses to frang. So we should release `sk->security` when we totally release connection.

### DIFF
--- a/fw/connection.c
+++ b/fw/connection.c
@@ -219,3 +219,22 @@ tfw_connection_hooks_unregister(int type)
 {
 	conn_hooks[TFW_CONN_TYPE2IDX(type)] = NULL;
 }
+
+/*
+ * Do an opposite to what tfw_connection_link_to_sk() does. Tempesta
+ * is unlinked from Sync Sockets layer, so that no data can be sent
+ * anymore on a connection. The previously held socket is released.
+ * Note that clearing of conn->sk is necessary. In case of failover
+ * on a server connection an indicator is needed to remove a hold
+ * on the socket. A zeroed conn->sk is that indicator.
+ */
+void
+tfw_connection_unlink_to_sk(TfwConn *conn)
+{
+	struct sock *sk = conn->sk;
+
+	if (sk->sk_security)
+		tfw_classify_conn_close(sk);
+	conn->sk = NULL;
+	ss_sock_put(sk);
+}
--- a/fw/connection.h
+++ b/fw/connection.h
@@ -555,23 +555,6 @@ tfw_connection_unlink_from_sk(struct sock *sk)
 	sk->sk_user_data = NULL;
 }
 
-/*
- * Do an opposite to what tfw_connection_link_to_sk() does. Tempesta
- * is unlinked from Sync Sockets layer, so that no data can be sent
- * anymore on a connection. The previously held socket is released.
- * Note that clearing of conn->sk is necessary. In case of failover
- * on a server connection an indicator is needed to remove a hold
- * on the socket. A zeroed conn->sk is that indicator.
- */
-static inline void
-tfw_connection_unlink_to_sk(TfwConn *conn)
-{
-	struct sock *sk = conn->sk;
-
-	conn->sk = NULL;
-	ss_sock_put(sk);
-}
-
 static inline void
 tfw_connection_unlink_from_peer(TfwConn *conn)
 {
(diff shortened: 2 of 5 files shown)


EvgeniiMekhanik:
This PR fixes https://github.com/tempesta-tech/tempesta/issues/2178 and possible fixes https://github.com/tempesta-tech/tempesta/issues/2169 (need check)

krizhanovsky (commented):
The fix looks good for me, but why do we face a case when the same upstream connection is used by many CPUs concurrently? Didn't we recently fix the problem with TCP connections CPU locality , even for the loopback interface?

EvgeniiMekhanik:
Thanks to @kingluo he pay my attention on our private discussion, that we also have a problem with `tfw_cli_conn_send`. I implement new commit to fix it. @kingluo please look my PR and if you see some other problem places please write me about it here!

kingluo (approved):
LGTM. Thanks.

fw/connection.c
```diff
@@ -214,3 +214,22 @@ tfw_connection_hooks_unregister(int type)
...
+/*
+ * Do an opposite to what tfw_connection_link_to_sk() does. Tempesta
+ * is unlinked from Sync Sockets layer, so that no data can be sent
+ * anymore on a connection. The previously held socket is released.
+ * Note that clearing of conn->sk is necessary. In case of failover
+ * on a server connection an indicator is needed to remove a hold
+ * on the socket. A zeroed conn->sk is that indicator.
+ */
+void
+tfw_connection_unlink_to_sk(TfwConn *conn)
```
keshonok, line 227: I understand the reason for the move, yet I wonder if there's a way to leave this function where it was, together with the opposite function. If there was it would be preferable.

fw/sock_clnt.c
```diff
@@ -152,6 +152,18 @@ tfw_cli_conn_send(TfwCliConn *cli_conn, TfwMsg *msg)
...
+	/*
+	 * msg can be freed in ss_do_send callback and
+	 * if it is the last msg for this connection
+	 * connection reference counter became equal to 1.
+	 * But if client close conenction and send FIN to us
+	 * connection reference counter will be decremented
+	 * and connection will be freed, so we should increment
+	 * connection reference counter here to access `cli_conn`
+	 * after sending data in this function.
+	 */
```
keshonok, line 164: Please allow me to rephrase a little bit. :-) You are free to use this any way you want.

_msg can be freed in ss_do_send() callback. If it's the last msg for this connection the ref counter becomes equal to 1. However if the client closes the connection and sends FIN, the connection ref counter is decremented to zero and the connection is freed. To avoid that, increment the connection ref counter here to guarantee access to cli_conn after sending data._

fw/connection.c
EvgeniiMekhanik: I think about it, but we need to tfw_classify_conn_close, which defined in http_limits.h. http_limits.h include connection.h, so I don't see any way to stay it in header except forward tfw_classify_conn_close declaration. I prefer move this function to connection.c

fw/sock_clnt.c
EvgeniiMekhanik: Yes fixed.

EvgeniiMekhanik:
> The fix looks good for me, but why do we face a case when the same upstream connection is used by many CPUs concurrently? Didn't we recently fix the problem with TCP connections CPU locality , even for the loopback interface?

Here we access client connection when we process response! Response processing can occurs and occurs on different CPU, because server socket is not a client socket!

EvgeniiMekhanik:
It seems my second commit was wrong. We call `tfw_cli_conn_send` from two places:
`tfw_h2_resp_fwd` and we get connection before call tfw_cli_conn_send in this function!
`__tfw_http_resp_fwd` which is called for http1 only where we have no skb destructor and where we never destroy message on other cpu

const-t (approved):
LGTM

PR #2200, opened 2024-08-06
Rework checking Tempesta state on load

EvgeniiMekhanik:
Previoucly we check dmesg to be sure that
Tempesta FW is running, but it is not good
because when access_log is on and dmesg buffer
is small we can't check it correctly, because
dmesg buffer can be owerriten. To fix this problem:
- Implement new Tempesta FW state - started (failed reconfig) to check situation when Tempesta still
running after incorrect live reconfiguration.
- Check only Tempesta FW state during Tempesta loading. If state is "started" or "started (failed reconfig)" continue to work, otherwise unload
Tempesta modules with error.

### DIFF
--- a/scripts/tempesta.sh
+++ b/scripts/tempesta.sh
@@ -237,24 +237,24 @@ prepare_db_directory()
 	rm -f /opt/tempesta/db/*.tdb;
 }
 
-start_tempesta_and_check()
+start_tempesta_and_check_state()
 {
-	# If 'net.tempesta.state' entry exists but Tempesta start process has
-	# been failed for some reason (e.g. configuration parsing error), then
-	# 'sysctl' does not indicate any problems and exits with zero code;
-	# so, stderr may return 'sysctl: setting key "net.tempesta.state"'
-	# followed by some error message or just empty string.
-	# The most reliable way to check Tempesta status is to check
-	# net.tempesta.state and if it is "start" check dmesg.
-	err=$(sysctl -w net.tempesta.state=start 2>&1 1>/dev/null)
+	local _err
+
+	_err=$(sysctl -w net.tempesta.state=start 2>&1 1>/dev/null)
 	TFW_STATE=$(sysctl net.tempesta.state 2> /dev/null)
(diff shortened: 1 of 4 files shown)


scripts/tempesta.sh
```diff
@@ -234,24 +234,24 @@ prepare_db_directory()
...
+	local _err
+
+	_err=$(sysctl -w net.tempesta.state=start 2>&1 1>/dev/null)
 	TFW_STATE=$(sysctl net.tempesta.state 2> /dev/null)
-	TFW_STATE=${TFW_STATE##* }
+	TFW_STATE=${TFW_STATE##net.tempesta.state = }
 
-	if [[ ${TFW_STATE} != "start" ]] || [[ -z "`check_dmesg 'Tempesta FW is ready'`" ]]; then
-		echo $err
+	remove_tmp_conf
```
EvgeniiMekhanik, line 245: Pay attention on this place now i remove tmp conf before check state (so it will be removed in case of error also

krizhanovsky: In my understanding we're safe here: `tfw_start()` is called in the process context of the sysctl write, so the file was read at this moment.

scripts/tempesta.sh
```diff
@@ -234,24 +234,24 @@ prepare_db_directory()
...
+	if [[ ${TFW_STATE} != "start" && ${TFW_STATE} != "start (failed reconfig)" ]]; then
+		unload_modules
+		error "cannot start Tempesta FW (sysctl message: ${_err##*: }, please check dmesg)"
+	else
+		if [[ $TFW_STATE == "start (failed reconfig)" ]]; then
+			error "Tempesta FW reconfiguration fails (sysctl message: ${_err##*: }, please check dmesg)."`
+				`" Tempesta FW still running on old configuration."
```
krizhanovsky, line 252:
```suggestion
				`" Tempesta FW is still running with old configuration."
```

krizhanovsky (approved):
LGTM

PR #2208, opened 2024-08-12
FIX

EvgeniiMekhanik:


fw/hpack.c
```diff
@@ -902,15 +902,15 @@ tfw_hpack_add_index(TfwHPackDTbl *__restrict tbl,
 				curr = count;
 			}
 			else if (entries == previous) {
-				memcpy_fast(entries + new_block - tail,
-					    entries + wrap, tail);
+				memcpy_fast((char *)(entries + new_block) - tail,
```
EvgeniiMekhanik, line 905: Pay attention to this place. I can't reproduce it in test

fw/hpack.c
```diff
@@ -902,15 +902,15 @@ tfw_hpack_add_index(TfwHPackDTbl *__restrict tbl,
 				curr = count;
 			}
 			else if (entries == previous) {
-				memcpy_fast(entries + new_block - tail,
-					    entries + wrap, tail);
+				memcpy_fast((char *)entries + new_block - tail,
```
EvgeniiMekhanik, line 905: pay attention on this place I can't reproduce it in test

kingluo (approved):

const-t (approved):
LGTM

fw/hpack.c
const-t: Seems it's impossible to reproduce. To reach this branch we must preallocate in the pool significantly more space, than size of `length * sizeof(TfwHPackEntry) << 1`.

`sizeof(TfwHPackEntry) = 112`

Example: Initially we have one page in the pool, on first realloc will be requested `length * sizeof(TfwHPackEntry) << 1 = 32 * 112 << 1 = 7168 bytes`, it can't be satisfied from current page and new pages will be allocated. On the next realloc we will have the same situation.

```
pool = 4096, new_length=(64 * 112) = 7168, n_pool = 8192
pool = 8192, new_length=(128 * 112) = 14336, n_pool = 16384
pool = 16384, new_length=(256 * 112) = 28672, n_pool = 32768
```
If can reach this branch only if we will pre-allocate 2 pages instead of 1, however now we [have only 1 preallocated](https://github.com/tempesta-tech/tempesta/blob/master/fw/hpack.h#L31) page.

PR #2212, opened 2024-08-13
Fix memory leak.

EvgeniiMekhanik:
When we have pipelined responses in one skb we
split such skb and save pointer to the next skb
in local variable. If processing of first response fails we don't free splitted skb. This patch fixes this behavior.

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -6446,7 +6446,7 @@ tfw_http_resp_cache(TfwHttpMsg *hmresp)
 	 */
 	if (test_bit(TFW_HTTP_B_HMONITOR, req->flags)) {
 		tfw_http_hm_drop_resp((TfwHttpResp *)hmresp);
-		return T_BAD;
+		return T_OK;
 	}
 	/*
 	 * This hook isn't in tfw_http_resp_fwd() because responses from the
@@ -6771,8 +6771,11 @@ next_msg:
 	 * @hmsib is not attached to the connection yet.
 	 */
 	r = tfw_http_resp_cache(hmresp);
-	if (unlikely(r != T_OK))
+	if (unlikely(r != T_OK)) {
+		if (hmsib)
+			tfw_http_conn_msg_free(hmsib);
 		return r;
+	}
 
 next_resp:
 	*splitted = NULL;
--- a/fw/http.c
+++ b/fw/http.c
@@ -6787,7 +6787,7 @@ next_msg:
 		return T_BAD;
 	}
 
-	return r;
+	return T_OK;
 bad_msg:
 	/*
 	 * Response can't be parsed or processed. This is abnormal situation,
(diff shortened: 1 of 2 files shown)


fw/http.c
```diff
@@ -6710,6 +6712,7 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 	 * protocol data.
 	 */
 	if (skb && !websocket) {
+		*splitted = NULL;
 		hmsib = tfw_http_msg_create_sibling(hmresp, skb);
```
EvgeniiMekhanik, line 6716: pay attention here. If we successfully create `hmsib` but fails later. Who frees `hmsib`?

EvgeniiMekhanik:
We need https://github.com/tempesta-tech/tempesta-test/issues/661 to check this PR

fw/http.c
```diff
@@ -6711,6 +6713,7 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 	 * protocol data.
 	 */
 	if (skb && !websocket) {
+		*splitted = NULL;
```
krizhanovsky, line 6716: In this and the above cases we explicitly set `splitted` to `NULL`, but there are return paths in the function, which do not touch `splitted` at all. I'd propose to leave only `*splitted = skb`.

BTW split is irregular verb with the same form, split split split. Could you please also fix this in `tfw_http_req_process()`?

krizhanovsky (approved):
LGTM

fw/http.c
```diff
@@ -6756,10 +6759,14 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 	 * @hmsib is not attached to the connection yet.
 	 */
 	r = tfw_http_resp_cache(hmresp);
-	if (unlikely(r != T_OK))
+	if (unlikely(r != T_OK)) {
+		if (hmsib)
```
const-t, line 6763: In case of error where we don't drop the connection we must not freed `hmsib`

fw/http.c
```diff
@@ -6756,10 +6759,14 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
 	 * @hmsib is not attached to the connection yet.
 	 */
 	r = tfw_http_resp_cache(hmresp);
-	if (unlikely(r != T_OK))
```
const-t, line 6759: Seems in line `6740:goto next_resp;` we must call `tfw_http_popreq` to dequeue req from fwd queue.

fw/http.c
```diff
@@ -6756,10 +6759,14 @@ tfw_http_resp_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb)
...
+	if (unlikely(r != T_OK)) {
+		if (hmsib)
+			tfw_http_conn_msg_free(hmsib);
```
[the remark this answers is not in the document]
EvgeniiMekhanik: No, if `websocket` we don't create `hmsib` also if no `websocket` we should use this `hmsib` when we go to next resp

fw/http.c
EvgeniiMekhanik: This function always return error after which we drop the connection

fw/http.c
EvgeniiMekhanik: I totally remove this code since frang always return error if we should drop connection

fw/http.c
EvgeniiMekhanik: We should zeroed `splitted` because if it will not be zeroed we will free it in `tfw_connection_recv` later

const-t (approved):
LGTM

PR #2213, opened 2024-08-15
Remove lower limit of reconnection attempts.

enuribekov-tempesta:
The first 6 attempts of reconnection have increasing time-out defined in the constant array. Previous logic, set 6 as a minimum attempts value. Actually, increasing time-out does not require passing all of the sequence and it can be interrupted at the any step.


fw/sock_srv.c
```diff
@@ -1280,12 +1280,9 @@ static int
 tfw_cfgop_conn_retries(TfwCfgSpec *cs, TfwCfgEntry *ce, unsigned int *recns)
 {
 	int r;
-
 	if((r = tfw_cfgop_intval(cs, ce, recns)))
 		return r;
-	*recns = *recns ? max_t(int, *recns, ARRAY_SIZE(tfw_srv_tmo_vals))
-			: UINT_MAX;
-
+	*recns = *recns ? *recns : UINT_MAX;
```
const-t, line 1285: I'm not sure about this fix. Our re-connection logic consist of 2 parts. 

The first is going over `tfw_srv_tmo_vals` without `tfw_connection_repair()`, Tempesta not tries to re-schedule or clear fwd queue, just fast reconnect. 

The second is simple reconnect with 1 sec interval that calls `tfw_connection_repair()` which evict or re-schedule requests to another connections. However, if number of re-connections reached maximum number of re-connections `tfw_connection_repair()` will clear fwd queue without rescheduling.

Let's consider example:
We set `server_connect_retries 3`, Tempesta does 3 attempts with intervals 1, 10, 100 and here max re-connections is reached, it means when Tempesta will move to the next part that includes `tfw_connection_repair()` and 1 sec intervals, it will not re-schedule requests from fwd queue and just try to evict them by timeout, but requests in the queue still fresh and could be re-scheduled to another connection. If timeout is not come for request, such request will hang in the fwd queue and wait for eviction or successful re-connection.

Another one thing, Tempesta will try to reconnect infinitely with 1 sec interval even after this patch. Need to discuss that behavior, if this is expected we should depict that clearly in our wiki.

I would prefer to do `tfw_srv_tmo_vals` always and don't increment `recns` for this tries and after move to re-connection with repairing. But it requires discussion.

enuribekov-tempesta: Fully agreed.
This fix likely point for discussion, than solution which I found right.

ttaym (approved):

fw/sock_srv.c
```diff
@@ -167,23 +155,22 @@ tfw_sock_srv_connect_try_later(TfwSrvConn *srv_conn)
 {
 	unsigned long timeout;
 
-	if (srv_conn->recns < ARRAY_SIZE(tfw_srv_tmo_vals)) {
+	if (srv_conn->recns < tfw_srv_tmo_nr) {
 		if (srv_conn->recns)
-			T_DBG_ADDR("Cannot establish connection",
+		T_DBG_ADDR("Cannot establish connection",
```
const-t, line 160: Broken identation

fw/sock_srv.c
```diff
@@ -167,23 +155,22 @@ tfw_sock_srv_connect_try_later(TfwSrvConn *srv_conn)
...
 	} else {
-		if (srv_conn->recns == ARRAY_SIZE(tfw_srv_tmo_vals)
-		    || !(srv_conn->recns % 60))
+		if (srv_conn->recns == tfw_srv_tmo_nr || !(srv_conn->recns % 60))
 		{
 			srv_warn("cannot establish connection",
 				 &srv_conn->peer->addr,
 				 ": %u tries, keep trying...\n",
-				 srv_conn->recns);
+		         srv_conn->recns - tfw_srv_tmo_nr + 1);
```
const-t, line 169: Broken identation

const-t (approved):

krizhanovsky:
@const-t @ttaym @RomanBelozerov  please update the wiki page to make `server_connect_retries` clear and avoid further confusions

EvgeniiMekhanik (approved):

PR #2219, opened 2024-08-21
Add tls certificate date validation

const-t:
Check certificate date and print warning, do not forbid to run Tempesta with expired certificate, only notice.

### DIFF
--- a/fw/tls_conf.c
+++ b/fw/tls_conf.c
@@ -176,6 +176,7 @@ tfw_tls_set_cert(TfwVhost *vhost, TfwCfgSpec *cs, TfwCfgEntry *ce)
 	unsigned char *crt_data;
 	size_t crt_size;
 	TlsCertConf *conf;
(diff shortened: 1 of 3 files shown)


fw/tls_conf.c
```diff
@@ -201,6 +202,18 @@ tfw_tls_set_cert(TfwVhost *vhost, TfwCfgSpec *cs, TfwCfgEntry *ce)
 		goto err;
 	}
 
+	if ((flags = ttls_x509_check_cert_validity(&conf->crt))) {
```
krizhanovsky, line 205: Could you please add a comment why we check only dates, and don't do the full check like with `ttls_x509_crt_verify_with_profile()`, which checks also algorithms and so on.

krizhanovsky (approved):
LGTM

EvgeniiMekhanik (approved):

PR #2223, opened 2024-08-26
Fix memory leak in `ttls_update_checksum`

EvgeniiMekhanik:
We allocate `tfm` when we call `ttls_sha256_init_start` and save pointer in `ctx->desc`, where ctx is
`ttls_sha256_context`. But in `ttls_update_checksum` we sometimes pass `tmp_sha256` into this function and will never free it. This patch fix this.


const-t (approved):
LGTM

krizhanovsky (approved):
LGTM, good catch!

PR #2224, opened 2024-08-26
Fix memory leak in apm.c

EvgeniiMekhanik:
We allocate `hm_entry` in `tfw_apm_create_def_health_stat_srv` and add it to `tfw_hm_codes_list`, so we should destroy it when we cleanup health monitor data structures.


krizhanovsky (approved):
LGTM!

const-t (approved):
LGTM

PR #2225, opened 2024-08-27
Fix memory leak during reconfiguration

EvgeniiMekhanik:
When we reload Tempesta FW and one of the old
listening address is equal to the listening
address from the new config, we forget to allocate `TfwListenSock` structure to this address but
forget to delete it. This patch fix it.


krizhanovsky (approved):
LGTM

const-t (approved):
LGTM

PR #2226, opened 2024-08-27
Fix sock leak

EvgeniiMekhanik:
When we call `tfw_sock_srv_connect_try` we first of all set socket refcounter to 1, during socket creation and then increment it in `tfw_connection_link_to_sk`. If `ss_connect` fails we should first put socket refcounter and then call connection drop (where we put it again) to deallocate sock.


krizhanovsky (approved):
LGTM

const-t (approved):
LGTM

PR #2233, opened 2024-09-05
Updating tempesta.sh, proposition.

KonsKo:
We have a problem with multiple starting the Tempesta many times (i.e. running test suite) in case where there are many networking devices. The problem is there is a part in script `tempesta.sh` that goes over networking devices and does some stuff (enabling RSS). This part may take about 3 seconds for 13 networking devices, and for about 1800 tests where we start the Tempesta for each of them it may take a lot of time.

Proposal:
- I updated the script `tempesta.sh` to have opportunity to `--start` it without running `tfw_set_net_queues`
- I created a new script for only running `tfw_set_net_queues`
- For our test suite, we may run the script `enable_rss.sh` one in the beginning, and we may start the Tempesta like `tempesta.sh --start --no-rss`
- Possibly, it may save us 5400 seconds, or 90 minutes in case of 13 networking devices and 1800 tests


krizhanovsky:
I guess there are several virtual interfaces among these 13, so probably we should not try to assign the queues on *all* interfaces and require to specify, which interfaces are used by Tempesta. Alternatively, we can assign the queues only for `lo` and determine the actual interfaces from the config file by parsing listen and server addresses.

KonsKo:
A command argument for the networking devices was already existed. I have added a new env var for the same purposes. I have updated the script logic a little.

RomanBelozerov (approved):
Looks good, probably, you should create a PR in tempesta-test for these changes?

krizhanovsky:
Please update https://tempesta-tech.com/knowledge-base/Run-and-stop/

KonsKo:
@const-t please review

const-t (changes requested):
PR looks good, but I would like to ask you to make an ability to use `-d` with `--reload` and `--restart` actions. Directive `listen` is live reconfigurable, thus user may want to add a new listener with a new device into config and do `--reload -d old_device, new_device`.

Now you changed order of arguments from `[options] {action}` to `{action} [options]` if after changes order will stay changed please update the line `echo -e "\nUsage: ${TFW_NAME} [options] {action}\n"`

KonsKo:
> PR looks good, but I would like to ask you to make an ability to use `-d` with `--reload` and `--restart` actions. Directive `listen` is live reconfigurable, thus user may want to add a new listener with a new device into config and do `--reload -d old_device, new_device`.
> 
> Now you changed order of arguments from `[options] {action}` to `{action} [options]` if after changes order will stay changed please update the line `echo -e "\nUsage: ${TFW_NAME} [options] {action}\n"`

I have updated the script.

I have changed the order of elements because the option `-d` is applicable to not all `actions`.

I have added some extra validations to notify an user, if you do not like it, I will be able to update it in the future.

KonsKo:
I have updated the Wiki page https://github.com/tempesta-tech/tempesta/wiki/Run-and-stop

const-t (approved):

KonsKo:
@krizhanovsky Could you merge the PR please? I suppose the CI does not work.

PR #2238, opened 2024-09-11
Mekhanik evgenii/fix 2175

EvgeniiMekhanik:


fw/cache.c
```diff
@@ -270,13 +272,20 @@ static TfwStr g_crlf = { .data = S_CRLF, .len = SLEN(S_CRLF) };
 		u_end = (uri_path)->chunks + (uri_path)->nchunks;	\
 	}								\
 	if (!TFW_STR_EMPTY(host)) {					\
```
const-t, line 274: `if (!TFW_STR_EMPTY(host))` must always be true, otherwise we have wrong logic. We even can replace it with `BUG_ON(TFW_STR_EMPTY(host));` because key always must consist of two parts uri and host.

const-t (changes requested):

fw/cache.c
```diff
@@ -261,22 +261,27 @@ static DEFINE_PER_CPU(char[RESP_BUF_LEN], g_c_buf);
...
 		h_start = (host)->chunks;				\
 		h_end = (host)->chunks + (host)->nchunks;		\
 	} else {							\
-		h_start = h_end = u_end;				\
+		h_start = host;						\
+		h_end = (host) + 1;					\
 	}								\
-	for ( ; c != h_end; ++c, c = (c == u_end) ? h_start : c)
+	for ( ; !u_fin && !h_fin;					\
+	     ++c, c = (c == u_end) ? h_start : c, u_fin = (c == u_end),	\
```
const-t, line 283: It seems we can get rid off `u_fin` and `h_fin`, because in this version we always have `h_start` and `h_end`.

const-t (approved):
LGTM

fw/cache.c
```diff
@@ -261,22 +261,27 @@ static DEFINE_PER_CPU(char[RESP_BUF_LEN], g_c_buf);
 static TfwStr g_crlf = { .data = S_CRLF, .len = SLEN(S_CRLF) };
 
 /* Iterate over request URI and Host header to process request key. */
-#define TFW_CACHE_REQ_KEYITER(c, uri_path, host, u_end, h_start, h_end)	\
+#define TFW_CACHE_REQ_KEYITER(c, uri_path, host, u_end, h_start,	\
+			      h_end, u_fin, h_fin)			\
+	u_fin = h_fin = false;						\
+	BUG_ON(TFW_STR_EMPTY(host));					\
```
krizhanovsky, line 267: We can have empty Host, i.e. HTTP/1 requires Host header, but a client can just sent `Host:\n`. Anyway, there is no reason to crash kernel: in worst case just WARN() and return `false`. But in general, I see no reason why not to iterate over empty `Host`. @const-t why we have a broken logic in this case?

fw/http_parser.c
```diff
@@ -5898,6 +5898,16 @@ Req_Method_1CharStep: __attribute__((cold))
...
 
+	/*
+	 * This state is necessary to drop requests with empty port like
+	 * GET http://tempesta-tech.com:
+	 */
+	__FSM_STATE(Req_UriPortBegin) {
+		if (likely(isdigit(c)))
+			__FSM_MOVE_f(Req_UriPort, &req->host);
+		TFW_PARSER_DROP(Req_UriPortBegin);
+	}
```
krizhanovsky, line 5909: Please adjust some functional tests and definitely the parser unit tests.

krizhanovsky (changes requested):
I still have a couple of notes, the thing about empty host seems crucial

fw/cache.c
ai-tmpst: Agree, it looks we can check only if c is not equal h_end:
`for ( ; c != h_end; ++c, c = (c == u_end) ? h_start : c)`

fw/cache.c
EvgeniiMekhanik: I remove the BUG_ON but not add WARN with returning false, because this macro is used in different places one of it need false and other need -EINVAL return code. We don't check that URI is not empty and it seems that for host we also can not check it. Host cant be empty because we always check it in `__check_authority_correctness` in http.c. I also add such test for Host:\n and cache

fw/cache.c
EvgeniiMekhanik: No we can't do it because sometimes `c = uri_path;` == h_end. For `GET http://user@-x/ HTTP/1.1\r\nHost: bad.com\r\n\r\n` for example

fw/cache.c
```diff
@@ -261,22 +261,26 @@ static DEFINE_PER_CPU(char[RESP_BUF_LEN], g_c_buf);
 static TfwStr g_crlf = { .data = S_CRLF, .len = SLEN(S_CRLF) };
 
 /* Iterate over request URI and Host header to process request key. */
-#define TFW_CACHE_REQ_KEYITER(c, uri_path, host, u_end, h_start, h_end)	\
+#define TFW_CACHE_REQ_KEYITER(c, uri_path, host, u_end, h_start,	\
+			      h_end, u_fin, h_fin)			\
+	u_fin = h_fin = false;						\
```
krizhanovsky, line 266: Shouldn't we initialize the variables like:
```C
u_fin = TFW_STR_EMPTY(uri_path);
h_fin = TFW_STR_EMPTY(host);
```

The current contract for the macro is unclear: whether we rely on the fact that `host` is never empty or are we OK to process empty host?

krizhanovsky (approved):
I don't think there is sense to do more review, so I approve and ask @ai-tmpst could you please pay attention to `TFW_CACHE_REQ_KEYITER` during your review?

fw/cache.c
ai-tmpst: Now I understand the problem: `uri_path` is placed immediately after `host` field in `TfwHttpReq`, and after `h_end = (host) + 1;` h_end points to uri_path. Therefore `c == uri_path == h_end` and the loop is not executed.

ai-tmpst: Maybe in such a way we can get rid of u_fin and h_fin?
`for ( ; c != u_end || ((u_end == h_end) ? 0 : (c = h_start, u_end = h_end, 1)); ++c)`

PR #2239, opened 2024-09-11
Fix releasing not dead sock

EvgeniiMekhanik:
When `ss_connect` fails we put socket refcounter
and release sock (when it becames equal to zero),
but socket was not closed. That leads to error
message in dmesg and some socket resources could
be not freed. This patch fix this behaviour, now
we close socket before free it.

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -218,21 +218,27 @@ typedef struct {
 	TfwRBQueue		wq;
 } TfwWorkTasklet;
 
+/* Cache modes. */
+typedef enum {
+        TFW_CACHE_UNDEFINED = -1,
+        TFW_CACHE_NONE = 0,
+        TFW_CACHE_SHARD,
+        TFW_CACHE_REPLICA,
+} TfwCacheMode;
+
 static struct {
 	int cache;
 	unsigned int methods;
 	unsigned long db_size;
 	const char *db_path;
-} cache_cfg __read_mostly;
+} cache_cfg __read_mostly = {
+	.cache = TFW_CACHE_UNDEFINED,
+	.methods = 0,
+	.db_size = 0,
+	.db_path = NULL
+};
 
 unsigned int cache_default_ttl;
-
-/* Cache modes. */
-enum {
-	TFW_CACHE_NONE = 0,
-	TFW_CACHE_SHARD,
-	TFW_CACHE_REPLICA,
-};
 
 typedef struct {
 	int 		*cpu;
@@ -3222,11 +3228,13 @@ static int
 tfw_cache_start(void)
 {
 	int i, r = 1;
-	TfwGlobal *g_vhost = tfw_vhost_get_global();
+
+	if (WARN_ON_ONCE(cache_cfg.cache == TFW_CACHE_UNDEFINED))
+		return -EINVAL;
 
 	if (tfw_runstate_is_reconfig())
 		return 0;
-	if (!(cache_cfg.cache || g_vhost->cache_purge))
+	if (!cache_cfg.cache)
 		return 0;
 
 	if ((r = tfw_init_node_cpus()))
@@ -3292,12 +3300,13 @@ node_cpus_alloc_err:
 static void
 tfw_cache_stop(void)
 {
-	TfwGlobal *g_vhost = tfw_vhost_get_global();
 	int i;
+
+	BUG_ON(cache_cfg.cache == TFW_CACHE_UNDEFINED);
 
 	if (tfw_runstate_is_reconfig())
 		return;
-	if (!(cache_cfg.cache || g_vhost->cache_purge))
+	if (!cache_cfg.cache)
 		return;
 
 	for_each_online_cpu(i)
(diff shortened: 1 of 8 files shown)


fw/cache.c
```diff
@@ -3292,11 +3292,12 @@ tfw_cache_start(void)
 static void
 tfw_cache_stop(void)
 {
+	TfwGlobal *g_vhost = tfw_vhost_get_global();
 	int i;
 
 	if (tfw_runstate_is_reconfig())
 		return;
-	if (!cache_cfg.cache)
+	if (!(cache_cfg.cache || g_vhost->cache_purge))
```
const-t, line 3300: This condition in `tfw_cache_start()` looks strange, why do we need to initialize cache if cache disabled, but purging is enabled? For instance in this case we initialize the cache: `cache 0; cache_purge invalidate;`, looks wrong.

const-t (approved):
In general LGTM, except cache initialization.

ai-tmpst (approved):

fw/cache.c
krizhanovsky: Yes, agree - we should not allow configuration of cache purging if we don't have cache at all! @EvgeniiMekhanik could you please fix this in `tfw_cache_start()` and add a test that this configuration isn't allowed?

fw/sock_srv.c
```diff
@@ -277,6 +277,9 @@ tfw_sock_srv_connect_try(TfwSrvConn *srv_conn)
 		if (r != -ESHUTDOWN)
 			T_ERR("Unable to initiate a connect to server: %d\n",
 				r);
+		bh_lock_sock(sk);
+		ss_do_close(sk, 0);
+		bh_unlock_sock(sk);
 		sock_put(sk);
 		SS_CALL(connection_drop, sk);
```
krizhanovsky, line 284: Why not `ss_close()`? It calls also `sock_put()` and `connection_drop`. It's error prone and breaks the abstraction layers w/o a clear reason

krizhanovsky (changes requested):
It seems a rework required in both the patches

fw/sock_srv.c
EvgeniiMekhanik: Because here socket state is SS_CLOSE, but we should not call ss_do_close for socket in this state in all other places except here. We can add special flag for such socket, but I think that this solution is better

fw/cache.h
```diff
@@ -23,7 +23,16 @@
 
 #include "http.h"
 
+/* Cache modes. */
+typedef enum {
+        TFW_CACHE_UNDEFINED = -1,
+        TFW_CACHE_NONE = 0,
+        TFW_CACHE_SHARD,
+        TFW_CACHE_REPLICA,
+} TfwCacheMode;
```
krizhanovsky, line 32: There is no sense to expose the particular caching strategy to the rest of the code. Let's move the enum to cache.c and provide `tfw_cache_enabled()` instead of `tfw_cache_get_mode()`

fw/cache.c
```diff
@@ -3222,11 +3220,12 @@ static int
 tfw_cache_start(void)
 {
 	int i, r = 1;
-	TfwGlobal *g_vhost = tfw_vhost_get_global();
+
+	BUG_ON(cache_cfg.cache == TFW_CACHE_UNDEFINED);
```
krizhanovsky, line 3224:
```suggestion
	if (WARN_ON_ONCE(cache_cfg.cache == TFW_CACHE_UNDEFINED))
	    return some_error_code;
```

Let's not crash server without a good reason!

krizhanovsky (approved):
LGTM, but please do the fixes before the merge!

fw/sock_srv.c
krizhanovsky: I don't understand
> Because here socket state is SS_CLOSE, but we should not call ss_do_close for socket in this state in all other places except here.

Yes, we discuss this particular place in `tfw_sock_srv_connect_try()`. 

My objection that here are 5 lines of code
```C
		bh_lock_sock(sk);
		ss_do_close(sk, 0);
		bh_unlock_sock(sk);
		sock_put(sk);
		SS_CALL(connection_drop, sk);
```
dealing with low-level socket API - this is subject for sock.c, no sock_src.c, which should care about connection management and work with connection abstractions.

What would be wrong if we replace these 5 lines with `ss_close()`?

EvgeniiMekhanik: look at the `ss_tx_action` case `SS_CLOSE`. we check that:
```
if (!((1 << sk->sk_state)
			      & (TCPF_ESTABLISHED | TCPF_SYN_SENT
				 | TCPF_CLOSE_WAIT)))
```
and if this if is true don't call `ss_do_close`. It is necessary to prevent multiple socket closing. But socket in `tfw_sock_srv_connect_try` has CLOSE state and this if is true. So if we call ss_close from `tfw_sock_srv_connect_try` instead of this five lines of code we never drop connection and close socket! We can pass special flag in ss_tx_action but I prefer this way. If you don't think so ok I will change it

PR #2243, opened 2024-09-17
Do not add Content-Length header for HTTP2 responses (#2209)

### LINKED ISSUE #2209: Content-Length header addition
We have some not optimal behavior. In case when `Content-Length` not listed in response headers and upstream server indicates message finishing by closing connection in function `tfw_http_resp_terminate()` we add `Content-Length` header.
Upd: The first part of this task about not optimal behavior was done in PR2243.

But for http2 protocol we call `tfw_h2_msg_cutoff_headers()` and after HPack encode headers, it implies we must not add the header in `tfw_http_resp_terminate()` but add this header during http2 message adjusting.

ai-tmpst:
The Content-Length header should not be added for HTTP2 responses in tfw_http_resp_terminate(), as it is ultimately removed by the tfw_h2_msg_cutoff_headers() function. Skip the Content-Length header addition for HTTP2 clients to avoid unnecessary operations.

Fix #2209

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -6665,7 +6665,6 @@ static void
 tfw_http_resp_terminate(TfwHttpMsg *hm)
 {
 	TfwFsmData data;
-	int r = 0;
 
 	/*
 	 * Response to HTTP2 client which has flag TFW_HTTP_B_CHUNKED_APPLIED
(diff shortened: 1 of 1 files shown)


const-t (approved):
LGTM

krizhanovsky (approved):
LGTM

Issue #2227, opened 2024-08-28
Duplicated error report


const-t:
During working with configuration faced strange behavior, Tempesta prints error reports twice:
```
[13925.858895] [tempesta fw] Warning: Configuration parsing has failed. Clean up...
[13925.963633] [tempesta fw] exiting...
[13926.116951] [tdb] Shutdown Tempesta DB
[13946.807356] [tdb] Start Tempesta DB
[13946.828263] [tempesta fw] Initializing Tempesta FW kernel module...
[13947.015465] [tempesta fw] ERROR: don't know how to handle: 'cache_use_stale'
[13947.016620] [tempesta fw] ERROR: configuration parsing error:
                  5: cache_use_stale 4* 4*;
                     ^^^^^^^^^^^^^^^^^^^^^^
[13947.019105] [tempesta fw] Warning: Configuration parsing has failed. Clean up...
[13947.139416] [tempesta fw] ERROR: don't know how to handle: 'cache_use_stale'
[13947.140538] [tempesta fw] ERROR: configuration parsing error:
                  5: cache_use_stale 4* 4*;
                     ^^^^^^^^^^^^^^^^^^^^^^
[13947.142926] [tempesta fw] Warning: Configuration parsing has failed. Clean up...
[13947.251560] [tempesta fw] exiting...
[13947.405052] [tdb] Shutdown Tempesta DB

```

After some investigation I found that `fprintf` that used by `sysctl` utility calls `write` twice in case of `EINVAL` error.

```
openat(AT_FDCWD, "/proc/sys/net/tempesta/state", O_WRONLY|O_TRUNC) = 3
write(3, "start\n", 6)                  = -1 EINVAL (Invalid argument)
write(3, "start\n", 6)                  = -1 EINVAL (Invalid argument)
close(3)                                = 0
```

`echo "start" > /proc/sys/net/tempesta/state` has expected behavior.
```
newfstatat(1, "", {st_mode=S_IFREG|0644, st_size=0, ...}, AT_EMPTY_PATH) = 0
write(1, "start\n", 6)                  = -1 EINVAL (Invalid argument)
close(1)                                = 0
```

It's definitely not a problem of `sysctl`, it calls `fprintf` only once as expected, I checked this.

ai-tmpst:
sysctl built from the procps repo doesn't produce duplicated errors, but I have not found any fundamental difference in the code. 
Think it's better to not use sysctl and write directly to /proc/sys/net/tempesta/state.

[...]

PR #2244, opened 2024-09-18
Fix duplicate error reporting (#2227)

ai-tmpst:
If sysctl handler returns EINVAL error, fprintf() in 'sysctl' command calls write() one more time. It leads to duplication of errors in dmesg when 'sysctl -w net.tempesta.state=start' is executed.

Use 'echo' to write directly to /proc/sys/net/tempesta/state instead of 'sysctl' execution. This change ensures that errors are reported only once, providing clearer feedback.

### DIFF
--- a/scripts/tempesta.sh
+++ b/scripts/tempesta.sh
@@ -242,7 +242,7 @@ start_tempesta_and_check_state()
 {
 	local _err
 
-	_err=$(sysctl -w net.tempesta.state=start 2>&1 1>/dev/null)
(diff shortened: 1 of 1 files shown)


const-t (approved):
Changes is reasonable for me. We can return EINVAL on error and there is no any specs describes such behavior of `fprintf` that could force us to not return EINVAL in such cases.

krizhanovsky (approved):
LGTM

Issue #2245, opened 2024-09-18
Redundant min(src->len, dst->len) in tfw_strcpy()


ai-tmpst:
From tfw_strcpy() function in fw/str.c:702:
```
	if (unlikely(src->len > dst->len))
		return -E2BIG;

	switch (mode) {
	case 3: /* The both are plain. */
		memcpy_fast(dst->data, src->data, min(src->len, dst->len));
		break;
```

If the condition (src->len > dst->len) is true, it returns an error. Since the condition guarantees that src->len <= dst->len, min(src->len, dst->len) can be simplified to src->len.

krizhanovsky:
Done in #2247

PR #2247, opened 2024-09-18
Use src->len directly for copy length in tfw_strcpy() (#2245)

ai-tmpst:
There was used 'min(src->len, dst->len)' to determine the number of bytes to copy. However, this check is redundant because the condition 'if (unlikely(src->len > dst->len))' above already ensures that 'src->len' will not exceed 'dst->len'.

This patch simplifies the memcpy operation by using 'src->len' directly as the copy length, making the code more straightforward and slightly more efficient.


const-t (approved):
LGTM

PR #2252, opened 2024-09-24
ALPN default protocol

const-t:
In case when ALPN not used by client, Tempesta must choose HTTP1 protocol instead of HTTP2. Likely such client doesn't support HTTP2.


krizhanovsky (approved):
LGTM

Issue #2117, opened 2024-05-09
Tls errors under ping flood


const-t:
# Scope
Faced weird behavior. I wrote little script for ping flooding, when I running it to test Tempesta, I'm getting a lot of warnings in dmesg about invalid tls processing. However when I added delays between ping frames, all warnings are gone. I tested this script on other implementations. Nginx just blocks me, because it has flood protection. Golang server processes all frames without errors.
 
 commit: 10b38e071bad93e87d5fab6d0213246d3b1b5c84
 
 Looks like we don't disconnect client on error in tls protocol.
 
**stats**:
```
Server successful TLS handshakes	: 10
Server failed TLS handshakes		: 33825
```

**log**:
```
[10137.585955] [tempesta tls] Warning: bad TLS version 3:23
[10137.587562] [tempesta tls] Warning: [::ffff:192.168.122.1] Bad TLS record (err -0xFFFFF004)
[10153.840855] net_ratelimit: 14992 callbacks suppressed
[10153.840859] [tempesta tls] Warning: [::ffff:192.168.122.1] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[10153.843922] [tempesta tls] Warning: [::ffff:192.168.122.1] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[10153.940461] [tempesta tls] Warning: [::ffff:192.168.122.1] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[10153.940473] [tempesta tls] Warning: [::ffff:192.168.122.1] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[10153.945246] [tempesta tls] Warning: [::ffff:192.168.122.1] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[10153.947550] [tempesta tls] Warning: [::ffff:192.168.122.1] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[10153.949924] [tempesta tls] Warning: [::ffff:192.168.122.1] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[10153.952290] [tempesta tls] Warning: [::ffff:192.168.122.1] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[...]

PR #2257, opened 2024-10-08
Fix memory under ping flood

### LINKED ISSUE #2149: Memory leak found in ping flood
# Motivation

Fix memory leak caused by TLS errors in ping floods https://github.com/tempesta-tech/tempesta/issues/2117.

# Scope

You can observe memory leaks by running a golang [h2 test program](https://github.com/tempesta-tech/tempesta/files/15262513/main.zip) and waiting for error cases. In my environment (4 CPU, 4 GB memory, two VMs), after the test the memory increased from 20% to 30%, which means a 10% memory leak.

```bash
./ping_flood -address 192.168.2.1:443 -threads 1 -connections 10 -debug 1
```

log:

```
[ 3824.299249] net_ratelimit: 1459 callbacks suppressed
[ 3824.299252] [tempesta tls] Warning: [::ffff:192.168.2.2] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[ 3824.304515] [tempesta tls] Warning: [::ffff:192.168.2.2] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[ 3824.310823] [tempesta tls] Warning: [::ffff:192.168.2.2] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[ 3824.385908] [tempesta tls] Warning: [::ffff:192.168.2.2] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[ 3824.389382] [tempesta tls] Warning: [::ffff:192.168.2.2] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[ 3824.393002] [tempesta tls] Warning: [::ffff:192.168.2.2] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[ 3824.396539] [tempesta tls] Warning: [::ffff:192.168.2.2] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[ 3824.400946] [tempesta tls] Warning: [::ffff:192.168.2.2] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[ 3824.405525] [tempesta tls] Warning: [::ffff:192.168.2.2] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[ 3824.409623] [tempesta tls] Warning: [::ffff:192.168.2.2] TLS cannot decrypt msg on state Handshake Over, ret=-74(bad ciphertext)
[ 3837.658748] net_ratelimit: 871 callbacks suppressed
[ 3837.659059] [tempesta tls] Warning: bad TLS version 12:23
[ 3837.659200] [tempesta tls] Warning: [::ffff:192.168.2.2] Bad TLS record (err -0xFFFFF004)
[ 3837.660977] [tempesta tls] Warning: bad TLS version 12:23
[ 3837.661152] [tempesta tls] Warning: [::ffff:192.168.2.2] Bad TLS record (err -0xFFFFF004)
[ 3837.661401] [tempesta tls] Warning: bad TLS version 12:23
[ 3837.661569] [tempesta tls] Warning: [::ffff:192.168.2.2] Bad TLS record (err -0xFFFFF004)
[ 3837.662657] [tempesta tls] Warning: bad TLS version 12:23
[ 3837.662846] [tempesta tls] Warning: [::ffff:192.168.2.2] Bad TLS record (err -0xFFFFF004)
[ 3837.663054] [tempesta tls] Warning: bad TLS version 12:23
[ 3837.663241] [tempesta tls] Warning: [::ffff:192.168.2.2] Bad TLS record (err -0xFFFFF004)

```

EvgeniiMekhanik:
There was a memory leak if `tfw_connection_recv`
fails with T_BAD in `tfw_tls_connection_recv -
because we continue to process skbs in this case
but don't call `tfw_connection_recv` and do not
free skbs. But if error occurs we should decrypt
skbs and pass it to `tfw_connection_recv` to process WINDOW_UPDATE frames (it is necessary to send
custom error responses with long body).
Also fix some replated problems:
- we should check Conn_Shutdown bit in all places where we check Conn_Stop bit to prevent processing any other frames.
- we should continue to process frames if `tfw_http_msg_process_generic` fails same as we do it in tls.c.
- add ALLOW_ERROR_INJECTION marso to check fixes.

Closes #2149 #2117

### DIFF
--- a/fw/connection.h
+++ b/fw/connection.h
@@ -455,6 +455,13 @@ __tfw_connection_get_if_live(TfwConn *conn)
 	__tfw_connection_get_if_live((TfwConn *)(c))
 
 static inline bool
+tfw_connection_was_stopped(TfwConn *conn)
+{
+	return (TFW_CONN_TYPE(conn) & Conn_Stop)
+		|| (TFW_CONN_TYPE(conn) & Conn_Shutdown);
+}
+
+static inline bool
 tfw_connection_stop_rcv(TfwConn *conn)
 {
 	/*
@@ -462,8 +469,8 @@ tfw_connection_stop_rcv(TfwConn *conn)
 	 * for HTTP2 client connections, to ensure delivery of
 	 * response with a body larger than HTTP window size.
 	 */
-	return (TFW_CONN_TYPE(conn) & Conn_Stop) &&
-		(!((TFW_CONN_TYPE(conn) & Conn_H2Clnt) == Conn_H2Clnt));
+	return (tfw_connection_was_stopped(conn)
+		&& (!((TFW_CONN_TYPE(conn) & Conn_H2Clnt) == Conn_H2Clnt)));
 }
 
 static inline void
@@ -611,6 +618,35 @@ tfw_peer_for_each_conn(TfwPeer *p, int (*cb)(TfwConn *))
 
 	return r;
 }
+
+#define TFW_CONN_PROCESS_RESULT(r, was_stopped, extra_cond_to_out, nskb) \
+do {									\
+	/*								\
+	 * If error occurs second time (save_error_code is not zero or	\
+	 * was_stopped is true) close connection immediately with RST.	\
+	 */								\
+	r = (r != T_OK && (save_err_code != T_OK || was_stopped)) ?	\
+		SS_BLOCK_WITH_RST : r;					\
+	if (r == T_BLOCK_WITH_FIN || r == T_BLOCK_WITH_RST) {		\
+		kfree_skb(nskb);					\
+		goto out;						\
+	} else if (r && r != T_DROP) {					\
+		if (extra_cond_to_out) {				\
+			kfree_skb(nskb);				\
+			goto out;					\
+		}		 					\
+	/*								\
+	 * In case of T_BAD or system errors we close connection	\
+	 * with tcp_shutdown() and gracefully send all pending		\
+	 * responses to client. We should continue to process		\
+	 * WINDOW_UPDATE frames so, we should decrypt all skbs,		\
+	 * not drop them.						\
+	 */								\
+		save_err_code = T_BAD;					\
+	} else if (save_err_code != T_OK) {				\
+		r = save_err_code;					\
+	}								\
+} while(0)
 
 extern unsigned int tfw_cli_max_concurrent_streams;
 
--- a/fw/connection.h
+++ b/fw/connection.h
@@ -619,35 +619,6 @@ tfw_peer_for_each_conn(TfwPeer *p, int (*cb)(TfwConn *))
 	return r;
 }
 
-#define TFW_CONN_PROCESS_RESULT(r, was_stopped, extra_cond_to_out, nskb) \
-do {									\
-	/*								\
-	 * If error occurs second time (save_error_code is not zero or	\
-	 * was_stopped is true) close connection immediately with RST.	\
-	 */								\
-	r = (r != T_OK && (save_err_code != T_OK || was_stopped)) ?	\
-		SS_BLOCK_WITH_RST : r;					\
-	if (r == T_BLOCK_WITH_FIN || r == T_BLOCK_WITH_RST) {		\
-		kfree_skb(nskb);					\
-		goto out;						\
-	} else if (r && r != T_DROP) {					\
-		if (extra_cond_to_out) {				\
-			kfree_skb(nskb);				\
-			goto out;					\
-		}		 					\
-	/*								\
-	 * In case of T_BAD or system errors we close connection	\
-	 * with tcp_shutdown() and gracefully send all pending		\
-	 * responses to client. We should continue to process		\
-	 * WINDOW_UPDATE frames so, we should decrypt all skbs,		\
-	 * not drop them.						\
-	 */								\
-		save_err_code = T_BAD;					\
-	} else if (save_err_code != T_OK) {				\
-		r = save_err_code;					\
-	}								\
-} while(0)
-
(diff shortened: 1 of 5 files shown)


const-t (approved):
Logic looks correct, but I must notice `TFW_CONN_PROCESS_RESULT` is confusing and does implicit return. Maybe we would rewrite it?

krizhanovsky:
It seems the only test we have for this is https://github.com/tempesta-tech/tempesta-test/pull/615 , which tests other HTTP/2 floods. So it seems we're going to merge this with no any tests. @RomanBelozerov @EvgeniiMekhanik can we enable at least part of the test PR to test this memory leak?

fw/connection.h
```diff
@@ -612,6 +619,35 @@ tfw_peer_for_each_conn(TfwPeer *p, int (*cb)(TfwConn *))
...
+			kfree_skb(nskb);				\
+			goto out;					\
+		}		 					\
+	/*								\
+	 * In case of T_BAD or system errors we close connection	\
+	 * with tcp_shutdown() and gracefully send all pending		\
+	 * responses to client. We should continue to process		\
+	 * WINDOW_UPDATE frames so, we should decrypt all skbs,		\
+	 * not drop them.						\
+	 */								\
```
krizhanovsky, line 644: the comment is misaligned

fw/connection.h
```diff
@@ -612,6 +619,35 @@ tfw_peer_for_each_conn(TfwPeer *p, int (*cb)(TfwConn *))
...
+	r = (r != T_OK && (save_err_code != T_OK || was_stopped)) ?	\
+		SS_BLOCK_WITH_RST : r;					\
+	if (r == T_BLOCK_WITH_FIN || r == T_BLOCK_WITH_RST) {		\
+		kfree_skb(nskb);					\
+		goto out;						\
+	} else if (r && r != T_DROP) {					\
+		if (extra_cond_to_out) {				\
+			kfree_skb(nskb);				\
+			goto out;					\
+		}		 					\
```
krizhanovsky, line 637: This condition is used only in `http_frame.c`

fw/http_frame.c
```diff
@@ -1848,18 +1855,19 @@ tfw_h2_frame_process(TfwConn *c, struct sk_buff *skb, struct sk_buff **next)
...
-			kfree_skb(nskb);
-			goto out;
-		}
+		/*
+		 * If Conn_Stop bit is not set we don't send error response
+		 * to client, so we don't need to process WINDOW_UPDATE frames
+		 * and can immediately close connection.
+		 */ 
+		TFW_CONN_PROCESS_RESULT(r, was_stopped,
+					!(TFW_CONN_TYPE(c) & Conn_Stop), nskb);
```
krizhanovsky, line 1864: The `tfw_http_msg_process_generic()` and `TFW_CONN_PROCESS_RESULT()` are identical, so instead of the macro we can just introduce a function to call  `tfw_http_msg_process_generic()` and process the result

fw/connection.h
```diff
@@ -612,6 +619,35 @@ tfw_peer_for_each_conn(TfwPeer *p, int (*cb)(TfwConn *))
...
+	 * with tcp_shutdown() and gracefully send all pending		\
+	 * responses to client. We should continue to process		\
+	 * WINDOW_UPDATE frames so, we should decrypt all skbs,		\
+	 * not drop them.						\
+	 */								\
+		save_err_code = T_BAD;					\
+	} else if (save_err_code != T_OK) {				\
+		r = save_err_code;					\
+	}								\
+} while(0)
```
krizhanovsky, line 649: The macro isn't really generic and treats result of different function in different way. We can reduce copy & paste in `http_frame.c` and I'm not sure about keeping such a macro just for 2, still bit different, code places. Personally I'd prefer just to reference the each other code in comment.

krizhanovsky (approved):
In general, looks good, but I ask @ai-tmpst to review this as well since the code isn't trivial and still could have non-obvious problems

fw/connection.h
```diff
@@ -462,8 +469,8 @@ tfw_connection_stop_rcv(TfwConn *conn)
 	 * for HTTP2 client connections, to ensure delivery of
 	 * response with a body larger than HTTP window size.
 	 */
-	return (TFW_CONN_TYPE(conn) & Conn_Stop) &&
-		(!((TFW_CONN_TYPE(conn) & Conn_H2Clnt) == Conn_H2Clnt));
+	return (tfw_connection_was_stopped(conn)
+		&& (!((TFW_CONN_TYPE(conn) & Conn_H2Clnt) == Conn_H2Clnt)));
```
ai-tmpst, line 473: `!((TFW_CONN_TYPE(conn) & Conn_H2Clnt) == Conn_H2Clnt)` 
Couldn't it be reduced to `!((TFW_CONN_TYPE(conn) & Conn_H2Clnt)`?

ai-tmpst (approved):

fw/connection.h
EvgeniiMekhanik: No because Conn_H2Clnt = 2 | TFW_FSM_HTTPS, so !((TFW_CONN_TYPE(conn) & Conn_H2Clnt) is false for H2 connections but it also false for https connection. But we should not process anything if error occurs in https connection.  For http2 connection we should do it because we should process window_update frames from the client to have ability to send error response with body

ai-tmpst: Oh, I see, it's not a one-bit flag. Sorry, I should look at it by myself.

PR #2259, opened 2024-10-10
Add ring buffers mapped to userspace

ai-tmpst:


fw/ringbuffer.c
```diff
@@ -0,0 +1,248 @@
...
+				memcpy_fast(rb->data + mhead, data, cur_len);
+				data += cur_len;
+				head += cur_len;
+				len -= cur_len;
+			}
+		}
+	}
+
+	smp_store_release(&rb->head, head);
+	smp_mb();
```
const-t, line 89: Why do you use mem barrier with per-cpu data?

ai-tmpst: You are right, it always executes in softirq context, and preemption disabling is excessive. Will get rid of it.

fw/ringbuffer.c
```diff
@@ -0,0 +1,248 @@
...
+	TfwRingbufferArea *rba = this_cpu_ptr(&rb_area);
+	TfwRingbuffer *rb = rba->rb;
+	unsigned int i;
+	u64 head, tail;
+	u32 full_size = 0;
+
+	if (atomic_read(&rba->unmapped))
+		return -EAGAIN;
+
+	preempt_disable();
```
EvgeniiMekhanik, line 54: We don't need `preempt_disable` as we discussed in private discussion since this function is executed in softirq

fw/ringbuffer.c
```diff
@@ -0,0 +1,248 @@
...
+} TfwRingbufferArea;
+
+#define AREA_SIZE(size) (size + TFW_RINGBUFFER_DATA_OFFSET)
+#define RBA_AREA_SIZE(rba) AREA_SIZE((rba)->rb->size)
+
+static DEFINE_PER_CPU(TfwRingbufferArea, rb_area);
+static bool proc_file_is_open;
+
+int
+tfw_ringbuffer_write(TfwStr **strs, unsigned int count)
```
EvgeniiMekhanik, line 43: I am not sure but it seems that we can use just
```
int
tfw_ringbuffer_write(TfwStr *str)
```
because TfwStr can contains a lot of TfwStr as chunks.

fw/ringbuffer.c
```diff
@@ -0,0 +1,248 @@
...
+
+	if (atomic_read(&rba->unmapped))
+		return -EAGAIN;
+
+	preempt_disable();
+
+	head = rb->head;
+	tail = smp_load_acquire(&rb->tail);
+
+	for (i = 0; i < count; ++i)
```
EvgeniiMekhanik, line 59: If we pass single TfwStr with a lot of chunks, we can skip this loop.

fw/ringbuffer.c
```diff
@@ -0,0 +1,248 @@
...
+		return -EAGAIN;
+
+	if (size == area_size)
+		atomic_set(&rba->unmapped, 0);
+
+	return 0;
+}
+
+void
+tfw_ringbuffer_test_set_unmapped(int unmapped)
```
EvgeniiMekhanik, line 148: It seems that this function is used only for test purpose, maybe move it to unit tests?

ai-tmpst: In this function we change a field of an internal object.
OK, maybe I can move this field to TfwRingbufer structure.

fw/ringbuffer.c
```diff
@@ -0,0 +1,248 @@
...
+			tail += cur_size;
+			size_for_cpu -= cur_size;
+		}
+
+		smp_store_release(&rb->tail, tail);
+		smp_mb();
+	}
+
+	return size;
+}
```
krizhanovsky, line 191: We don't keep testing-specific code in the main code, designated for the real workload. I'd move the function to `test_ringbuffer.c` and just include `ringbuffer.c` to `test_ringbuffer.c`

fw/ringbuffer.c
krizhanovsky: I think the point is that atomics imply full barriers on x86-64

fw/ringbuffer.h
```diff
@@ -0,0 +1,90 @@
...
+#include <stdint.h>
+
+#define u32 uint32_t
+#define u64 uint64_t
+
+#endif /* __KERNEL__ */
+
+#define TFW_RINGBUFFER_DATA_OFFSET 4096
+#define TFW_RINGBUFFER_MIN_SIZE    4096
+#define TFW_RINGBUFFER_MAX_SIZE    (4096 * 4096)
```
krizhanovsky, line 64: It makes sense to express the constants using `PAGE_SIZE`

krizhanovsky (approved):
LGTM after cleanups mentioned by @const-t and @EvgeniiMekhanik . But please also have a look onto other RB implementations - maybe we can get ready to use implementation. For now the implementation lacks the file operations, but later we might face a bigger problem to let user-space threads sleep on no events and make it low overhead

fw/ringbuffer.h
```diff
@@ -0,0 +1,90 @@
...
+typedef struct {
+	u64 head;
+	u64 tail;
+	u32 mask;
+	u32 size;
+	u32 cpu;
+	u32 max_cpu_num;
+	char userspace_data[256];
+	char __aligned(TFW_RINGBUFFER_DATA_OFFSET) data[];
+} TfwRingbuffer;
```
krizhanovsky, line 75: Could you please align the data structure members and comment them, in both the new structures. Not everything is clear from a quick look

fw/ringbuffer.c
```diff
@@ -0,0 +1,248 @@
...
+	pfn = page_to_pfn(rba->pg);
+
+	if (remap_pfn_range(vma, vma->vm_start, pfn, size, vma->vm_page_prot))
+		return -EAGAIN;
+
+	if (size == area_size)
+		atomic_set(&rba->unmapped, 0);
+
+	return 0;
+}
```
krizhanovsky, line 145: So we're going to have per-cpu file, right? Could you please comment the function about this consideration? Basically having a minimal documentation in the source, like I referened in the linux kernel, would be useful

fw/ringbuffer.c
```diff
@@ -0,0 +1,248 @@
...
+
+	for_each_online_cpu(cpu) {
+		TfwRingbufferArea *rba = per_cpu_ptr(&rb_area, cpu);
+		atomic_set(&rba->unmapped, 1);
+		if (rba->pg) {
+			__free_pages(rba->pg, get_order(RBA_AREA_SIZE(rba)));
+			rba->pg = NULL;
+		}
+	}
+}
```
krizhanovsky, line 248: What happens if we unload the module while the user space still didn't close and/or unmap the file?

krizhanovsky (commented):
There are some logic absent - please address it in a next PR

fw/ringbuffer.c
```diff
@@ -0,0 +1,248 @@
+/**
```
krizhanovsky, line 1: Technically this is the second lock-free RB in Tempesta FW and it's not generic. I'd rename the file to something pointing out that this is a kernel/user-space transport, not just an abstract data RB

fw/ringbuffer.c
ai-tmpst: Oops, I answered to the comment about preempt_disable.

So we can do without smp_mb, don't we?

fw/ringbuffer.h
```diff
@@ -0,0 +1,90 @@
...
+ * The communication between the kernel and user space is lockless. The kernel
+ * manages the write pointer (head), while user space manages the read pointer
+ * (tail). Each side only modifies its own pointer, preventing race conditions
+ * and eliminating the need for locking mechanisms.
+ *
+ * Since the ring buffers are memory-mapped into user space, data does not need
+ * to be copied between the kernel and user space. Instead, user-space threads
+ * can directly access the data in the kernel’s memory, greatly improving
+ * performance by avoiding the overhead of traditional system calls and memory
+ * copying.
```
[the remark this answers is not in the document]
ai-tmpst: Actually my current version also require data length knowledge before the call. 
Maybe we get rid of it if we allow incomplete data blocks. As you mentioned early, we could send fields as TLV and mark the end of the data block with a special "end" TLV. So we could get the size of the room between tail and head before writing, and on every field addition we have to make sure if there is enough place for this field and ending TLV.

Will add the motivation to the comments in the code.

fw/ringbuffer.h
ai-tmpst: I wanted to define this for userspace too.
OK, will find another way to do it.

fw/ringbuffer.c
ai-tmpst: No, just one file with multiple mapped per-cpu areas.

There is no appropriate munmap field in `struct file_operations`. I could set vma->vm_ops and catch unmap event, but I think it would be better to move this field to the Ringbuffer structure, so a client could change it when it prepared to read.

fw/ringbuffer.c
ai-tmpst: The memory remains available to user space.
I thought about it and wanted to notify the user space with a special event. But maybe it would be more correctly to handle this situation on the RB level. Will think.

fw/ringbuffer.c
const-t: > So we can do without smp_mb, don't we?

Yes, we don't need `smp_mb`

fw/ringbuffer.h
krizhanovsky:
```
$ getconf -a|grep 4096
PATH_MAX                           4096
_POSIX_PATH_MAX                    4096
PIPE_BUF                           4096
_POSIX_PIPE_BUF                    4096
NL_ARGMAX                          4096
PAGESIZE                           4096
PAGE_SIZE                          4096
POSIX_ALLOC_SIZE_MIN               4096
POSIX_REC_MIN_XFER_SIZE            4096
POSIX_REC_XFER_ALIGN               4096
```

fw/ringbuffer.c
krizhanovsky: We agreed on the call that this API introduces unwanted overhead on forming an array, so probably we should just provide a raw memory and a commit call, which moves the `tail`

fw/t/unit/test_mmap_buffer.c
```diff
@@ -0,0 +1,186 @@
...
+			p1 = p2; \
+		} \
+	} while (0)
+
+	tfw_mmap_buffer_get_room(holder, &p1, &s1, &p2, &s2);
+	WALK_BUFFER(fill_buf);
+	EXPECT_EQ(r, expect_wr);
+	if (r)
+		return;
+	tfw_mmap_buffer_commit(holder, size);
```
krizhanovsky, line 124: The API looks good and I didn't get from https://github.com/tempesta-tech/tempesta/pull/2259#discussion_r1799440292

> Actually my current version also require data length knowledge before the call.

It seems now we get as much memory as RB has and can write as much data as we need (and can), so there is no need to compute the size of all written data upfront.

fw/mmap_buffer.c
```diff
@@ -0,0 +1,305 @@
...
+}
+
+static int
+dev_file_open(struct inode *ino, struct file *filp)
+{
+	TfwMmapBufferHolder *holder;
+	int i;
+
+	for (i = 0; i < holders_cnt; ++i) {
+		if (strcmp(holders[i]->dev_name, (char *)filp->f_path.dentry->d_iname) == 0) {
```
krizhanovsky, line 105: too wide line: we that's OK to slightly exceed 80 characters per line, but not so much

fw/mmap_buffer.c
```diff
@@ -0,0 +1,305 @@
...
+		if (holder->dev_major < 0) {
+			T_WARN("Registering char device failed for %s\n", filename);
+			goto err;
+		}
+
+		holder->dev_class = class_create(THIS_MODULE, filename);
+		device_create(holder->dev_class, NULL,
+					  MKDEV(holder->dev_major, 0), NULL, filename);
+		strscpy(holder->dev_name, filename, sizeof(holder->dev_name));
+		holders[holders_cnt++] = holder;
```
krizhanovsky, line 255: `holders_cnt` can exceed `MAX_HOLDERS`

fw/mmap_buffer.h
```diff
@@ -0,0 +1,175 @@
...
+	char dev_name[TFW_MMAP_BUFFER_MAX_NAME_LEN];
+	atomic_t dev_is_opened;
+	/* is_freeing indicates that freeing process started. It's necessary to
+	 * exclude repeated file opening.
+	 */
+	atomic_t is_freeing;
+	int dev_major;
+	struct class *dev_class;
+	struct page *pg[];
+} TfwMmapBufferHolder;
```
krizhanovsky, line 149: Coding style: could you please align the data structure members and adjust comments like for example in https://github.com/tempesta-tech/tempesta/blob/master/fw/cache.c#L129 ?

fw/mmap_buffer.c
```diff
@@ -0,0 +1,305 @@
...
+		|| size > TFW_MMAP_BUFFER_MAX_SIZE
+		|| !is_power_of_2(size))
+		return NULL;
+
+	if (filename && strlen(filename) >= TFW_MMAP_BUFFER_MAX_NAME_LEN - 1)
+		return NULL;
+
+	holder = kmalloc(sizeof(TfwMmapBufferHolder) +
+					 sizeof(struct page *) * num_online_cpus(),
+					 GFP_KERNEL);
```
krizhanovsky, line 215: It seems you need to zeroize the memory (or GFP_ZERO) since you check `holder->pg[cpu] ` for `NULL` in `tfw_mmap_buffer_free()`

fw/mmap_buffer.c
```diff
@@ -0,0 +1,305 @@
...
+		return -EINVAL;
+
+	buf = *per_cpu_ptr(holder->buf, cpu_id);
+	pfn = page_to_pfn(virt_to_page(buf));
+
+	if (remap_pfn_range(vma, vma->vm_start, pfn, size, vma->vm_page_prot))
+		return -EAGAIN;
+
+	vma->vm_ops = &dev_vm_ops;
+	(void)dev_vm_ops;
```
krizhanovsky, line 180: I din't get the sense of `(void)dev_vm_ops` statement...

fw/mmap_buffer.c
```diff
@@ -0,0 +1,305 @@
...
+
+	if (size == buf_size)
+		atomic_set(&buf->is_ready, 1);
+
+	return 0;
+
+#undef NTH_ONLINE_CPU
+}
+
+static void dev_file_vm_close(struct vm_area_struct *vma)
```
krizhanovsky, line 190:
```suggestion
static void
dev_file_vm_close(struct vm_area_struct *vma)
```

krizhanovsky (approved):
I commented some issues, but there are also pending reviews from @const-t and @EvgeniiMekhanik , so I just approve since there is no any design issues, and leave the rest for your fixes and reviews.

fw/mmap_buffer.c
```diff
@@ -0,0 +1,305 @@
...
+
+	*size2 = 0;
+
+	if (!atomic_read(&buf->is_ready)) {
+		*size1 = 0;
+		return;
+	}
+
+	head = buf->head % buf->size;
+	tail = smp_load_acquire(&buf->tail) % buf->size;
```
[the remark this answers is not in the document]
ai-tmpst: The problem is - we allocate the buffer by `alloc_pages_node(cpu_to_node(cpu), GFP_KERNEL, order)`, and there is two bad way: we use division or we lose almost a half of the buffer size (minus header size).
Maybe we can use another function for allocation which doesn't restrict memory size by a power of 2. But I haven't found appropriate one.
I don't know... Maybe we can use __alloc_percpu and get struct page pointers walking through every allocated page. I considered this way but it looked a little bit ugly for me. Is there another ways?

fw/t/unit/test_mmap_buffer.c
ai-tmpst: Yes, in this variant we don't need to compute whole event size before writing.

fw/mmap_buffer.c
ai-tmpst: Yes, I saw it and replaced by kzalloc.

fw/mmap_buffer.c
ai-tmpst: Didn't notice, left after debugging.

krizhanovsky:
Should we just close it in favor of https://github.com/tempesta-tech/tempesta/pull/2273 ?

fw/mmap_buffer.c
```diff
@@ -0,0 +1,303 @@
...
+	if (size < TFW_MMAP_BUFFER_MIN_SIZE
+		|| size > TFW_MMAP_BUFFER_MAX_SIZE
+		|| !is_power_of_2(size))
+		return NULL;
+
+	if (filename && strlen(filename) >= TFW_MMAP_BUFFER_MAX_NAME_LEN - 1)
+		return NULL;
+
+	holder = kmalloc(sizeof(TfwMmapBufferHolder) +
+					 sizeof(struct page *) * num_online_cpus(),
```
EvgeniiMekhanik, line 214: Broken aligment

fw/mmap_buffer.c
```diff
@@ -0,0 +1,303 @@
...
+
+TfwMmapBufferHolder *
+tfw_mmap_buffer_create(const char *filename, unsigned int size)
+{
+	TfwMmapBufferHolder *holder;
+	unsigned int order;
+	int cpu;
+
+	if (size < TFW_MMAP_BUFFER_MIN_SIZE
+		|| size > TFW_MMAP_BUFFER_MAX_SIZE
```
EvgeniiMekhanik, line 206: Broken aligment

fw/mmap_buffer.c
```diff
@@ -0,0 +1,303 @@
...
+					 sizeof(struct page *) * num_online_cpus(),
+					 GFP_KERNEL);
+	if (!holder)
+		return NULL;
+
+	order = get_order(size);
+
+	holder->dev_major = -1;
+	holder->buf = __alloc_percpu_gfp(sizeof(TfwMmapBuffer *),
+									 sizeof(u64), GFP_KERNEL);
```
EvgeniiMekhanik: broken aligment

fw/mmap_buffer.c
```diff
@@ -0,0 +1,303 @@
...
+	holder = kmalloc(sizeof(TfwMmapBufferHolder) +
+					 sizeof(struct page *) * num_online_cpus(),
+					 GFP_KERNEL);
+	if (!holder)
+		return NULL;
+
+	order = get_order(size);
+
+	holder->dev_major = -1;
+	holder->buf = __alloc_percpu_gfp(sizeof(TfwMmapBuffer *),
```
EvgeniiMekhanik: It seems that `__alloc_percpu_gfp` can return NULL, so we should handle this case

fw/mmap_buffer.c
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	if (filename) { /* do not create the file in unit tests */
+		holder->dev_major = register_chrdev(0, filename, &dev_fops);
+		if (holder->dev_major < 0) {
+			T_WARN("Registering char device failed for %s\n", filename);
+			goto err;
+		}
+
+		holder->dev_class = class_create(THIS_MODULE, filename);
```
EvgeniiMekhanik, line 251: class_create can fails, so we should check dev_class

fw/mmap_buffer.c
```diff
@@ -0,0 +1,303 @@
...
+		device_create(holder->dev_class, NULL,
```
EvgeniiMekhanik, line 252: device_create can also fails.so we should handle this case

fw/t/unit/test_mmap_buffer.c
```diff
@@ -0,0 +1,186 @@
...
+
+	/* Check all the possible head and tail offsets */
+	for (i = 0; i < MAX_SIZE + 1; ++i)
+		test_write_read(MAX_SIZE, 0, 0);
+
+	tfw_mmap_buffer_free(holder);
+
+#undef MAX_SIZE
+}
+
```
EvgeniiMekhanik, line 180: We should add simple test to check device createion.
I add
```
TEST(tfw_mmap_buffer, create_dev)
{
	holder = tfw_mmap_buffer_create("test", TFW_MMAP_BUFFER_MIN_SIZE);
	EXPECT_NOT_NULL(holder);
	EXPECT_NULL(tfw_mmap_buffer_create("test", TFW_MMAP_BUFFER_MIN_SIZE));
	tfw_mmap_buffer_free(holder);
}
```
And kernel crashes because we don't handle invalid device creation

fw/mmap_buffer.c
```diff
@@ -0,0 +1,303 @@
...
+
+		if (holder->pg[cpu]) {
+			__free_pages(holder->pg[cpu],
+						 get_order(TFW_MMAP_BUFFER_FULL_SIZE(buf->size)));
+			holder->pg[cpu] = NULL;
+		}
+	}
+
+	if (holder->dev_major > 0) {
+		device_destroy(holder->dev_class, MKDEV(holder->dev_major, 0));
```
EvgeniiMekhanik, line 297: It seems that deinitialization should be more complicated, because device can be NULL and class can be NULL

fw/mmap_buffer.c
```diff
@@ -0,0 +1,305 @@
...
+	holder = kmalloc(sizeof(TfwMmapBufferHolder) +
+					 sizeof(struct page *) * num_online_cpus(),
+					 GFP_KERNEL);
+	if (!holder)
+		return NULL;
+
+	order = get_order(size);
+
+	holder->dev_major = -1;
+	holder->buf = (TfwMmapBuffer **)alloc_percpu_gfp(sizeof(TfwMmapBuffer *),
```
EvgeniiMekhanik, line 222: `alloc_percpu_gfp` can fails we should handle this case

EvgeniiMekhanik (changes requested):
Need fixes

fw/mmap_buffer.c
```diff
@@ -0,0 +1,305 @@
...
+};
+
+static const struct vm_operations_struct dev_vm_ops = {
+	.close = dev_file_vm_close
+};
+
+void
+tfw_mmap_buffer_get_room(TfwMmapBufferHolder *holder,
+						 char **part1, unsigned int *size1,
+						 char **part2, unsigned int *size2)
```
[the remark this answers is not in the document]
ai-tmpst: DONE

fw/mmap_buffer.c
ai-tmpst: I used 4-spaces tabs. Fixed.

fw/mmap_buffer.c
ai-tmpst: I used 4-spaces tabs. Fixed.

fw/mmap_buffer.c
ai-tmpst: Agree, will fix.

fw/mmap_buffer.c
ai-tmpst: Fixed.

fw/mmap_buffer.c
ai-tmpst: Fixed.

fw/mmap_buffer.c
```diff
@@ -0,0 +1,305 @@
...
+{
+	int cpu;
+
+	if (!holder)
+		return;
+
+	atomic_set(&holder->is_freeing, 1);
+
+	for_each_online_cpu(cpu) {
+		TfwMmapBuffer *buf = *per_cpu_ptr(holder->buf, cpu);
```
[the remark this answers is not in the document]
ai-tmpst: Fixed.

fw/mmap_buffer.c
ai-tmpst: Fixed.

fw/mmap_buffer.h
ai-tmpst: Fixed.

fw/t/unit/test_mmap_buffer.c
ai-tmpst: Done.

fw/mmap_buffer.c
ai-tmpst: Fixed.

fw/mmap_buffer.c
ai-tmpst: Fixed.

fw/ringbuffer.c
ai-tmpst: Fixed.

ai-tmpst:
Looks through all the comments. All issues are fixed.
This PR can be closed. Let's continue work in PR [2273](https://github.com/tempesta-tech/tempesta/pull/2273).

PR #2263, opened 2024-10-15
Remove warnings in case of error `tfw_http_req_cache_cb`

EvgeniiMekhanik:
There are to may warnings in dmesg if there is no appropriate backend server, so print only debug message.

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -1483,6 +1483,15 @@ tfw_http_nip_req_resched_err(TfwSrvConn *srv_conn, TfwHttpReq *req,
 			  " re-forwarded or re-scheduled");
 }
 
+static inline void
+tfw_http_send_err_resp_nolog(TfwHttpReq *req, int status)
+{
+	if (TFW_MSG_H2(req))
+		tfw_h2_send_err_resp(req, status, false);
+	else
+		tfw_h1_send_err_resp(req, status);
+}
+
 /* Common interface for sending error responses. */
 void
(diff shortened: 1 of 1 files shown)


fw/http.c
```diff
@@ -5497,11 +5497,13 @@ tfw_http_req_cache_cb(TfwHttpMsg *msg)
...
 
 send_502:
-	tfw_http_send_err_resp(req, 502, "request dropped: processing error");
+	T_DBG("request dropped: processing error, status 502");
+	tfw_http_send_err_resp(req, 502, NULL);
 	TFW_INC_STAT_BH(clnt.msgs_otherr);
 	return;
 send_500:
-	tfw_http_send_err_resp(req, 500, "request dropped: processing error");
+	T_DBG("request dropped: processing error, status 500");
```
krizhanovsky, line 5505: I'm OK with this version, we already have a `tfw_http_send_err_resp()` with NULL `reason`. However, I'd prefer a way proposed by @const-t to introduce a function like `tfw_http_send_err_resp_nolog()`, which don't accept `reason` and which is called by `tfw_http_send_err_resp()` for the actual part of the sending

krizhanovsky (approved):
LGTM

fw/http.c
EvgeniiMekhanik: Yes fixed

const-t (approved):
LGTM

PR #2274, opened 2024-10-31
TMP

EvgeniiMekhanik:
Should be merged only after #2272


ai-tmpst (approved):
LGTM

fw/sock.c
```diff
@@ -884,20 +888,30 @@ do {									\
...
 
-		r = SS_CALL(connection_recv, conn, skb);
-
-		if (r < 0) {
-			T_DBG2("[%d]: Processing error: sk=%pK r=%d\n",
-			       smp_processor_id(), sk, r);
-			goto out; /* connection must be dropped */
+		r = SS_CALL(connection_recv, conn, skb, &tmp_save_err_code);
+		WARN_ON(r && tmp_save_err_code);
+		r = tfw_handle_error(r, &save_err_code, was_stopped);
```
const-t, line 893: We do `tfw_handle_error()` on tls layer and http layer why do we need it here? In my understanding, tls and http layer are responsible for error handling and here we should have only result from upper layers.

fw/sock.c
```diff
@@ -820,6 +820,9 @@ ss_tcp_process_skb(struct sock *sk, struct sk_buff *skb, int *processed)
 	void *conn;
 	struct sk_buff *skb_head = NULL;
 	struct tcp_sock *tp = tcp_sk(sk);
+	int save_err_code = SS_OK, tmp_save_err_code;
```
const-t, line 823: I must notice that `save_err_code` has life time the same `ss_tcp_process_skb()` it means, client can send multiple bad packets with some interval, but `save_err_code` will be `SS_OK`.

fw/tls.c
```diff
@@ -187,7 +185,8 @@ tfw_tls_connection_recv(TfwConn *conn, struct sk_buff *skb)
 			tfw_tls_purge_io_ctx(&tls->io_in);
 			kfree_skb(nskb);
 			spin_unlock(&tls->lock);
-			return T_BAD;
+			return tfw_handle_error(T_BAD, save_err_code,
```
const-t, line 188: Why we calling `tfw_handle_error()` instead of return `SS_BLOCK_WITH_RST`? It looks confusing, we must drop connection in case of such errors.

EvgeniiMekhanik:
@krizhanovsky  I have a question. If something wen wrong and we close connection with TCP RST, we call ss_close -> ss_do_close. But ss_do_close we be called later in ss_tx_action. So if some new skbs from client come before ss_do_close will be called there will be some TLS warnings in dmesg (because we can't decrypt them). Is it ok or we should implement special flag to drop such skbs.

fw/sock.c
EvgeniiMekhanik: In this case we use Conn_Stop bit. If error occurs on previus call of this function Conn_Stop bit was set.

keshonok:
I've got just a cursory glance, so I might be mistaken. The logic looks a bit complex with these fine-tune flags and details. Does this work for HTTP2 only? I far as I understand, thus should not happen for HTTP.  It's not obvious, not from the commit message, and not from comments in the code. Perhaps, this should be made obvious both in the code comments in appropriate places, and in the commit message. Thanks!

fw/tls.c
```diff
@@ -114,23 +113,21 @@ tfw_tls_connection_recv(TfwConn *conn, struct sk_buff *skb)
...
+		return tfw_handle_error(r, save_err_code, was_stopped);
 	case T_POSTPONE:
 		/* No complete TLS record seen yet. */
 		spin_unlock(&tls->lock);
 		/*
 		 * save_error_code is T_OK or T_BAD if error occurs
 		 * on one of the previous steps.
 		 */
-		return save_err_code;
+		return T_OK;
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed

lib/log.h
```diff
@@ -64,6 +64,22 @@ enum {
 	T_OK		 = 0,
 };
 
+static inline int
+tfw_handle_error(int r, int *save_err_code, bool was_stopped)
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes implement special file to this case

fw/connection.c
```diff
@@ -176,10 +176,12 @@ tfw_connection_recv(TfwConn *conn, struct sk_buff *skb)
...
 			if (unlikely(TFW_CONN_PROTO(conn) == TFW_FSM_WS
-				     || TFW_CONN_PROTO(conn) == TFW_FSM_WSS))
+				     || TFW_CONN_PROTO(conn) == TFW_FSM_WSS)) {
 				r = tfw_ws_msg_process(conn, skb);
-			else
-				r = tfw_http_msg_process(conn, skb, &split);
+			} else {
+				r = tfw_http_msg_process(conn, skb, &split,
+							 save_err_code);
+			}
```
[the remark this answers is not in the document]
EvgeniiMekhanik: fixed

fw/tls.c
EvgeniiMekhanik: This function returns T_BAD if it first error (in this case we send tcp_shutdown as it was before) or BLOCK_WITH_RST if it is a seconf error

PR #2277, opened 2024-11-07
Move streams closed by RST to closed queue

EvgeniiMekhanik:
According RFC 9113 :
An endpoint that sends a RST_STREAM frame on a stream that is in the "open" or "half-closed (local)" state could receive any type of frame. The peer might have sent or enqueued for sending these frames before
processing the RST_STREAM frame.
For this purpose Tempesta FW implements special not described in RFC state - HTTP2_STREAM_LOC_CLOSED
state. When Tempesta FW sends RST stream stream moved to HTTP2_STREAM_LOC_CLOSED state and silently ignores new received frames. But such stream is not moved to closed queue, so it will be never deleted until
connection will be closed. If count of such streams exceeded max concurrent streams client can't open
new streams. This patch fixes this problem.

### DIFF
--- a/fw/http2.c
+++ b/fw/http2.c
@@ -334,6 +334,29 @@ tfw_h2_current_stream_remove(TfwH2Ctx *ctx)
 }
 
 /*
+ * Send RST stream and move stream to the queue of closed streams.
+ * When count of closed streams exceeded TFW_MAX_CLOSED_STREAMS,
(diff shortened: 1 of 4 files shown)


const-t (approved):
LGTM

ai-tmpst (approved):
LGTM

PR #2289, opened 2024-11-21
Vk 2052 introduce ja5t calc

Kutumov:


tls/tls_srv.c
```diff
@@ -883,6 +894,11 @@ ttls_parse_client_hello(TlsCtx *tls, unsigned char *buf, size_t len,
 		}
 		io->hslen -= 2;
 		tls->hs->cs_cur_len += 2;
+
+		/* ja5t must be initialized with zeros */
+		tls->sess.ja5t.cipher_suite_hash *= TTLS_JA5_HASH_CALC_PRIME;
+		tls->sess.ja5t.cipher_suite_hash += cs;;
```
krizhanovsky, line 900:
```suggestion
		tls->sess.ja5t.cipher_suite_hash += cs;
```

These 2 lines implement our hashing, which could change -what about to make a macro to compute a hash for any field? Like
```c
#define TTLS_COMPUTE_JA5_ACCHASH(hash, field)                                   \              
do {                                                                            \              
        (hash) *= TTLS_JA5_HASH_CALC_PRIME;                                     \              
        (hash) += (field);                                                      \              
} while (0)                   
```

tls/tls_srv.c
```diff
@@ -1061,20 +1081,21 @@ ttls_parse_client_hello(TlsCtx *tls, unsigned char *buf, size_t len,
...
-                        if ((ret = ttls_parse_extension(tls, tmp, ext_sz, ext_type)))
-                                return ret;
-                }
+
+        T_DBG3("ClientHello: read %u bytes for ext %u\n", io->rlen + n, ext_type);
+        if (ext_supported) {
+        	if ((ret = ttls_parse_extension(tls, tmp, ext_sz, ext_type)))
+            	return ret;
+        }
+
```
krizhanovsky, line 1098: Senseless change with broken identation

tls/tls_srv.c
```diff
@@ -1121,6 +1142,10 @@ ttls_parse_client_hello(TlsCtx *tls, unsigned char *buf, size_t len,
 	 */
 	ttls_process_session_ticket(tls);
 
+	/* JA5t computation */
+	tls->sess.ja5t.is_abbreviated = tls->hs->resume;
+	tls->sess.ja5t.alpn = tls->alpn_chosen->id;;
```
krizhanovsky: Not exactly: this is choosen ALPN, but we need ALPN, sent by a client, so it seems `ttls_parse_alpn_ext()` is a better place. E.g. if we end up with HTTP/2, we still need to fingerprint `http/1.1,h2`. Ideally, if we can handle the difference between `http/1.1,h2` and `h2,http/1.1`, but not necessary - if it causes overhead, we'll be just fine with setting bits for h2 or http/1.1

krizhanovsky (changes requested):
It seems the only issue is with ALPN computation and the rest is just coding style, probably introduced with inappropriate dev  setup

tls/tls_srv.c
```diff
@@ -1061,20 +1081,21 @@ ttls_parse_client_hello(TlsCtx *tls, unsigned char *buf, size_t len,
 		 */
 		BUG_ON(io->rlen > ext_sz);
 		n = min_t(int, ext_sz - io->rlen, buf + len - p);
-                if (ext_supported) {
+        if (ext_supported) {
 			if (unlikely(ext_type == TTLS_TLS_EXT_SESSION_TICKET))
                               tmp = tls->hs->ticket_ctx.ticket;
 		        memcpy_fast(tmp + io->rlen, p, n);
-                }
+        }
```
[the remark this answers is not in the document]
Kutumov: I use tabs.
Why 3? This brace fits to `if (ext_supported) {` which has 2 breaks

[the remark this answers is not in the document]
Kutumov: Guess i'm using space, sorry

tls/tls_srv.c
Kutumov: fixed

tls/ttls.h
```diff
@@ -269,6 +269,34 @@ struct ttls_alpn_proto {
 	int		id;
 };
```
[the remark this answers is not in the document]
Kutumov: done

tls/ecp.h
```diff
@@ -99,9 +99,9 @@ extern ttls_ecp_group_id ttls_preset_curves[];
 /**
  * Curve information for use by other modules.
  *
- * @grp_id		- nternal identifier;
+ * @grp_id		- internal identifier;
  * @tls_id		- TLS NamedCurve identifier;
- * @bit_size		- urve size in bits;
+ * @bit_size	- curve size in bits;
```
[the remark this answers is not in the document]
Kutumov: done

tls/tls_srv.c
Kutumov: done

tls/tls_srv.c
Kutumov: done

tls/tls_srv.c
```diff
@@ -720,6 +723,7 @@ ttls_parse_client_hello(TlsCtx *tls, unsigned char *buf, size_t len,
 	TlsIOCtx *io = &tls->io_in;
 	T_FSM_INIT(tls->state, "TLS ClientHello");
 
+
```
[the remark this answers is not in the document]
Kutumov: done

tls/tls_srv.c
```diff
@@ -1061,20 +1088,21 @@ ttls_parse_client_hello(TlsCtx *tls, unsigned char *buf, size_t len,
...
-                       ext_type);
-                if (ext_supported) {
-                        if ((ret = ttls_parse_extension(tls, tmp, ext_sz, ext_type)))
-                                return ret;
-                }
+		T_DBG3("ClientHello: read %u bytes for ext %u\n", io->rlen + n,
+			ext_type);
+		if (ext_supported) {
+			ret = ttls_parse_extension(tls, tmp, ext_sz, ext_type);
+			if (ret)
```
const-t, line 1103:
```suggestion
			if (unlikely(ret))
```

const-t (approved):
LGTM

tls/tls_srv.c
```diff
@@ -391,20 +394,29 @@ ttls_parse_alpn_ext(TlsCtx *tls, const unsigned char *buf, size_t len)
...
+		for (i = 0; i < TTLS_ALPN_PROTOS && alpn_list[i].name; ++i) {
+			our = &alpn_list[i];
+			if (ttls_alpn_ext_eq(our, theirs, cur_len)) {
+				/* Calculate JA5t */
+				tls->sess.ja5t.alpn <<= 2;
+				tls->sess.ja5t.alpn += our->id;
+				/* Use our order of preference */
+				if (alpn_chosen_idx > i && ttls_alpn_match_cb(tls, our)) {
+					alpn_chosen_idx = i;
+					tls->alpn_chosen = our;
```
krizhanovsky, line 410: It seems there are issues in the logic:

1. we save the last matching ALPN instead of 1st because we don't break loop now on first match
2. if we have `http/1.1,h2` from a client and say `h2,http/1.1` in Tempesta, then we account `http/1.1,h2` twice in `tls->sess.ja5t.alpn`
3. other ALPNs aren't accounted at all, but if a client sends us something non-standard, then it could be quite a signature

krizhanovsky (changes requested):
There are still ALPN fingerprinting issues

Kutumov:
Add ja5t calculation even if client hello has incorrect extensions

tls/tls_srv.c
Kutumov: 1. We still choose the 1st matching alpn from our list by saving the least matching index in alpn_chosen_idx, i.e. our order of preference is used.
2. In outer loop we have http/1.1, then iterating in the inner loop we find http/1.1 in our list and we add http/1.1 to ja5t. In the second iteration of outer loop we have h2 and find it in our list and add to ja5t. We use `ttls_alpn_ext_eq` to decide whether to use current alpn in ja5t calculation, so only one calculation takes place for every alpn value

Kutumov: 3. Added has_unknown_alpn flag in ja5t

tls/tls_srv.c
```diff
@@ -391,20 +394,29 @@ ttls_parse_alpn_ext(TlsCtx *tls, const unsigned char *buf, size_t len)
...
-	for (i = 0; i < TTLS_ALPN_PROTOS && alpn_list[i].name; ++i) {
-		our = &alpn_list[i];
-		for (theirs = start; theirs != end; theirs += cur_len) {
-			cur_len = *theirs++;
-			if (ttls_alpn_ext_eq(our, theirs, cur_len)
-			    && ttls_alpn_match_cb(tls, our))
-			{
-				tls->alpn_chosen = our;
-				return 0;
+
```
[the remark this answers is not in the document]
Kutumov: done

tls/ttls.h
```diff
@@ -284,13 +316,14 @@ struct ttls_alpn_proto {
  *			  from TLS ticket);
  */
 typedef struct {
-	TlsX509Crt	*peer_cert;
-	long		start;
-	int		etm;
+	TlsX509Crt		*peer_cert;
+	long			start;
+	int				etm;
```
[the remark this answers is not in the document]
Kutumov: done

tls/ecp.h
```diff
@@ -99,9 +99,9 @@ extern ttls_ecp_group_id ttls_preset_curves[];
 /**
```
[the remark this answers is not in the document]
Kutumov: done

tls/ttls.h
```diff
@@ -269,6 +269,38 @@ struct ttls_alpn_proto {
...
+/**
+ * Just a small prime, relatively far from a power of 2
+ */
+#define TTLS_JA5_HASH_CALC_PRIME 11
+
+#define TTLS_COMPUTE_JA5_ACCHASH(hash, field)	\
+do {											\
+		(hash) *= TTLS_JA5_HASH_CALC_PRIME;		\
+		(hash) += (field);						\
+} while (0)
```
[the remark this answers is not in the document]
Kutumov: I had 4 character tabs in my IDE. Fixed now

tls/tls_srv.c
Kutumov: done

tls/tls_srv.c
```diff
@@ -391,20 +395,32 @@ ttls_parse_alpn_ext(TlsCtx *tls, const unsigned char *buf, size_t len)
...
+		has_unknown_alpn = true;
+		for (i = 0; i < TTLS_ALPN_PROTOS && alpn_list[i].name; ++i) {
+			our = &alpn_list[i];
+			if (ttls_alpn_ext_eq(our, theirs, cur_len)) {
+				/* Calculate JA5t */
+				tls->sess.ja5t.alpn <<= 2;
+				tls->sess.ja5t.alpn += our->id;
+				/* Use our order of preference */
+				if (alpn_chosen_idx > i &&
+					ttls_alpn_match_cb(tls, our)) {
```
krizhanovsky, line 410:
```suggestion
                                if (alpn_chosen_idx > i &&
                                    ttls_alpn_match_cb(tls, our))
                                {
```

krizhanovsky (approved):
LGTM, thank you for the changes!

PR #2291, opened 2024-11-27
Fix race in accessing `g_tmp_mpool`

EvgeniiMekhanik:
`g_tmp_mpool` is per cpu vatiable, but it can be accessed by softirq context (when we handle new tls connection) and process context (when we read/parse cetificate and key). For example we allocate memory using this pool in process context `ttls_rsa_import_raw`-> `__rsa_setup_ctx` -> `ttls_mpi_inv_mod` -> `ttls_mpool_alloc_stack` and free pool from softirq `ttls_recv` -> `ttls_mpi_pool_cleanup_ctx`. So we should disable softirq when we use pool in process context.


krizhanovsky (approved):
LGTM. Just please update copyrights in the files (no github actions?)

const-t (approved):
LGTM

PR #1955, opened 2023-08-16
Rework HTTP message forwarning

const-t:
HTTP part of #1103.

PR #2294, opened 2024-11-28
Update request's `host` during heders modification

const-t:
We must update `TfwHttpReq::host` during headers modification to prevent inconsistency between `TfwHttpReq::host` and new host header.

One case where we got problem because of this, it's cache where we calculate size of the cache entry using `req->host.len`(total len not changed, we didn't update `req->host`), but when copy headers to cache we iterating over each chunk, however some chunks already has another length, because they are changed in `__hdr_sub()`. Chunks are changed, because `req->host` stores pointers to chunks, but not values of chunks, however `req->host.len` is value.


const-t:
Moved to https://github.com/tempesta-tech/tempesta/pull/1955

PR #2297, opened 2024-11-28
Mekhanik evgenii/fix 2181

EvgeniiMekhanik:


const-t (approved):
LGTM

const-t:
> choose another stream to send and broke RFC, that leads
> to issue descirbed in https://github.com/tempesta-tech/tempesta/pull/2281.

I would like to ask you, to not use links to issue on github in commit message, it would be better to add some short description in commit message. Even if github repo will be removed for some reasons we need to have clear messages in history.

krizhanovsky (approved):
LGTM

PR #2302, opened 2024-12-04
Fix NUMA for the cache

const-t:


fw/cache.c
```diff
@@ -2628,7 +2623,7 @@ tfw_cache_add(TfwHttpResp *resp, tfw_http_cache_cb_t action)
 		 * rather than in softirq...
 		 */
 		for_each_node_with_cpus(nid)
-			__cache_add_node(c_nodes[nid].db, resp, key);
+			__cache_add_node(get_db_for_node(nid), resp, key);
```
const-t, line 2626: Here we doing `__cache_add_node()`, however more proper way to do it is to call only `tfw_cache_copy_resp()` for each node, instead of calling whole `__cache_add_node()`. I would like to do it in another patch if we not decide to move NUMA to TDB.

const-t:
Typo in branch, actually it must be `kt-2295-panic`

fw/cache.c
```diff
@@ -376,7 +376,7 @@ tfw_release_node_cpus(void)
 	if(!c_nodes)
 		return;
 
-	for(node = 0; node < nr_online_nodes; node++) {
+	for_each_node_with_cpus(node) {
```
EvgeniiMekhanik, line 379: It seems that during kernel loading we can also use `for_each_node_with_cpus`

EvgeniiMekhanik:
We have a problem with function `tfw_cache_key_node` It can return numa node without any active cpu and this leads to BUG.
But I can't suggest any effective way to avoid this problem, because we can have active 0 and 2 and inactive 1 numa node for example. I suggest one way below but not effective. 
```
	static DEFINE_PER_CPU(int, node);
	int *nid = per_cpu_ptr(&node, smp_processor_id());
	unsigned short rc;

	while (1) {
		if (nr_cpus_node(*nid))
			break;
		else if ((*nid = next_online_node(*nid)) >= MAX_NUMNODES)
			*nid = first_online_node;
	}

	rc = *nid;
	*nid = next_online_node(*nid);
	return rc;
```

EvgeniiMekhanik (approved):
According our private discussion we decide to not take into account cases when there is a numa node without any cpu. Please change `for_each_node_with_cpu` to `for_each_online_node`

fw/cache.c
const-t: During kernel loading we can't use `for_each_node_with_cpus`, we do memory reserve too early, at this stage we can't get `nr_cpus` by node, it always be zero. It must be implemented in other way.

const-t:
> According our private discussion we decide to not take into account cases when there is a numa node without any cpu. Please change for_each_node_with_cpu to for_each_online_node

There is no practical sense to replace `for_each_node_with_cpu` with `for_each_online_node`. Replica mode still can be used in setup where not each node has cpu, in this case we must ignore such nodes and not allocate entries on node without cpu.

fw/cache.c
```diff
@@ -2650,17 +2623,64 @@ tfw_cache_add(TfwHttpResp *resp, tfw_http_cache_cb_t action)
...
+tfw_cache_purge_invoke_impl(TfwHttpReq *req,
+			    int purge_impl(TDB *db, TfwHttpReq *req))
+{
+	int ret;
+
+	if (cache_cfg.cache == TFW_CACHE_REPLICA) {
+		int nid;
+
+		for_each_node_with_cpus(nid)
+			ret = purge_impl(c_nodes[nid].db, req);
```
krizhanovsky: Shouldn't we "remember" the first non-zero value? At the moment if a call fails, but the next one succeed, we return 0

krizhanovsky (approved):
LGTM

fw/cache.c
const-t: We could, but as I see there is no difference for caller. In situation where we have 4 nodes and one fails to purge, but three nodes purged successfully might be treated as success of as fail it depends on the rest code. But for now it's the same.

PR #2303, opened 2024-12-05
Implement ja5 hash calculation.

EvgeniiMekhanik:


fw/http_msg.c
```diff
@@ -545,6 +545,71 @@ tfw_http_msg_hdr_open(TfwHttpMsg *hm, unsigned char *hdr_start)
 	       hdr_start, *hdr_start, hdr->skb);
 }
 
+#define CALC_JA5H_SUMM(req, hash)		\
+do {						\
+	req->ja5h.summ *= 11;			\
```
const-t, line 550: @EvgeniiMekhanik @Kutumov Guys, maybe we need to extract ja5 related stuff to the header file? Seems you are using same constants but with different names or even unnamed. Also it looks confusing when hashing functions are spread around.

fw/http.h
```diff
@@ -336,6 +336,45 @@ typedef struct {
...
+ */
+typedef struct {
+	unsigned int	version:1;
+	unsigned int	method:5;
+	unsigned int	cookie_num:5;
+	unsigned int	headers_num:6;
+	unsigned int	has_refer:1;
+	unsigned int	padding:6;
+	unsigned int	summ;
+} __packed HttpJa5h;
```
Kutumov, line 372: Packed or not - in any case size is 2 ints

fw/http_msg.c
```diff
@@ -545,6 +545,71 @@ tfw_http_msg_hdr_open(TfwHttpMsg *hm, unsigned char *hdr_start)
...
+
+static void
+tfw_http_req_calc_ja5h_summ_for_raw_hdr(TfwHttpReq *req, TfwStr *hdr)
+{
+	const TfwStr *dup, *dup_end, *c, *chunk_end;
+	size_t len = 0;
+	unsigned int summ = 0;
+
+#define TFW_CHAR4_INT(data)						\
+	((data[3] << 24) | (data[2] << 16) | (data[1] << 8) | data[0])
```
Kutumov, line 562: Maybe just cast to int?

fw/http.h
EvgeniiMekhanik: No, you can print sizeof  and see 7 bytes

fw/http.h
```diff
@@ -336,6 +336,45 @@ typedef struct {
...
+ *		  and more headers);
+ * @has_refer	- has Refer;
+ * @summ	- sum * 11 + header, where header is an 'id'
+ *		  if 'id' < TFW_HTTP_HDR_RAW or 4 or less bytes
+ *		  (depending on how many bytes available) value
+ *		  (header name or header name + value if name is
+ *		   less then 4 bytes);
+ */
+typedef struct {
+	unsigned int	version:1;
```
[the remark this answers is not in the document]
EvgeniiMekhanik: Yes fixed

fw/http_msg.c
EvgeniiMekhanik: Yes fixed

fw/http_msg.c
```diff
@@ -545,6 +545,67 @@ tfw_http_msg_hdr_open(TfwHttpMsg *hm, unsigned char *hdr_start)
...
+static void
+tfw_http_req_calc_ja5h_summ_for_raw_hdr(TfwHttpReq *req, TfwStr *hdr)
+{
+	const TfwStr *dup, *dup_end, *c, *chunk_end;
+	size_t len = 0;
+	unsigned int summ = 0;
+
+#define TFW_CHAR4_INT(data)	*((unsigned int *)(data))
+#define TFW_CHAR3_INT(data)	((data[2] << 16) | (data[1] << 8) | data[0])
+#define TFW_CHAR2_INT(data)	((data[1] << 8) | data[0])
```
Kutumov, line 563: This calculation can also be replace by cast to `short`

fw/http_msg.c
```diff
@@ -545,6 +545,67 @@ tfw_http_msg_hdr_open(TfwHttpMsg *hm, unsigned char *hdr_start)
...
+
+static void
+tfw_http_req_calc_ja5h_summ_for_raw_hdr(TfwHttpReq *req, TfwStr *hdr)
+{
+	const TfwStr *dup, *dup_end, *c, *chunk_end;
+	size_t len = 0;
+	unsigned int summ = 0;
+
+#define TFW_CHAR4_INT(data)	*((unsigned int *)(data))
+#define TFW_CHAR3_INT(data)	((data[2] << 16) | (data[1] << 8) | data[0])
```
Kutumov, line 562: Part of this calculation can also be replaced by cast to `short`

fw/http.h
Kutumov: Okay. I didnt know that unused bitfields area is also shrinked by this attribute.

fw/hpack.c
```diff
@@ -1324,6 +1324,7 @@ tfw_hpack_hdr_set(TfwHPack *__restrict hp, TfwHttpReq *__restrict req,
 		break;
 	case TFW_TAG_HDR_COOKIE:
 		parser->_hdr_tag = TFW_HTTP_HDR_COOKIE;
+		h2_set_hdr_cookie(req, &entry->cstate);
```
const-t, line 1327: This is subtle place. If I right here, might be inaccuracy in some cases. Let's assume that in the first request we received 1 cookie header, then in the second request we received 3 new headers and plus one indexed header from the last request. We calculate `cookie_num` for 3 headers and after override it with old one `cookie_num` that was stored in hpack entry. I'm not sure that we must fix it, but if it works so we definitely must write a comment.

fw/http_msg.c
```diff
@@ -734,10 +724,11 @@ tfw_http_msg_hdr_close(TfwHttpMsg *hm)
 		req->header_list_sz += h->len + TFW_HTTP_MSG_HDR_OVERHEAD(hm);
 		req->headers_cnt++;
 		HTTP_JA5H_CALC_NUM(req, headers, TFW_HTTP_JA5H_HEADERS_MAX);
-		if (likely(id < TFW_HTTP_HDR_RAW))
-			CALC_JA5H_SUMM(req, id);
-		else
+		if (likely(id < TFW_HTTP_HDR_RAW)){
```
const-t, line 727: Unnecessary braces.

fw/hpack.c
EvgeniiMekhanik: Good point, I will add test and fix

fw/http_msg.c
EvgeniiMekhanik: Fixed

fw/http_parser.c
```diff
@@ -8013,10 +8016,20 @@ do {									\
 }
 STACK_FRAME_NON_STANDARD(__h2_req_parse_content_type);
 
+void
+h2_set_hdr_cookie(TfwHttpReq *req, const TfwCachedHeaderState *cstate)
+{
+	if (cstate->is_set) {
```
const-t, line 8022: Unnecessary braces

fw/http_parser.c
```diff
@@ -8029,6 +8042,14 @@ __h2_req_parse_cookie(TfwHttpMsg *hm, unsigned char *data, size_t len, bool fin)
 	__FSM_START(parser->_i_st);
 
 	__FSM_STATE(Req_I_CookieStart) {
+		HTTP_JA5H_REQ_CALC_NUM(((TfwHttpReq *)hm), cookie,
+				       TFW_HTTP_JA5H_COOKIE_MAX, 1);
+		cookie_num =
```
const-t, line 8047: Agreed with solution, bit in case when chunked cookie received, this will not work, you use local variable, that will be set to zero on each cookie chunk, therefore `cstate` will have wrong value. It would be better to use `parser->cstate.cookie.cookie_num` directly or for local var initialization(In this case you can declare `cookie_num` inside `__FSM_STATE(Req_I_CookieStart)` state's scope).

const-t (changes requested):

const-t (approved):
LGTM

Kutumov (approved):

PR #2305, opened 2024-12-06
Remove memory corruption during trailer processing

EvgeniiMekhanik:
We pass `TfwHtppMsg` which can be (request or response) in `tfw_http_msg_hdr_close`, so we should check that it is a server connection before cast it to `TfwHttpResp`, when we set response trailer length.

### DIFF
--- a/fw/http_msg.c
+++ b/fw/http_msg.c
@@ -566,7 +566,7 @@ tfw_http_msg_hdr_close(TfwHttpMsg *hm)
 	parser->hdr.flags |= TFW_STR_COMPLETE;
 
 	/* Cumulate the trailer headers length */
-	if (parser->hdr.flags & TFW_STR_TRAILER) {
+	if (is_srv_conn && parser->hdr.flags & TFW_STR_TRAILER) {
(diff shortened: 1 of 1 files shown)


fw/http_msg.c
```diff
@@ -566,7 +566,7 @@ tfw_http_msg_hdr_close(TfwHttpMsg *hm)
 	parser->hdr.flags |= TFW_STR_COMPLETE;
 
 	/* Cumulate the trailer headers length */
-	if (parser->hdr.flags & TFW_STR_TRAILER) {
+	if (is_srv_conn && parser->hdr.flags & TFW_STR_TRAILER) {
```
krizhanovsky, line 569: Maybe to aggregate this `if` with the one at the below?
```C
if (is_srv_conn) {
    if (parser->hdr.flags & TFW_STR_TRAILER) {
       ....
     if (!TFW_MSG_H2(hm) && ....
 ```

krizhanovsky (approved):
LGTM!

fw/http_msg.c
EvgeniiMekhanik: The if below is for `!is_srv_conn`

PR #2307, opened 2024-12-12
Remove `BUG_ON(v_min == USHRT_MAX)` in apm.c

EvgeniiMekhanik:
There was a commit (903cddae857658239cdc9b898970b03988cd2d13) where this BUG_ON was added. But there is a still small chance of race between calling `tfw_apm_prnctl_calc` and `__tfw_apm_update -> tfw_apm_rbent_checkreset`. If this race occurs `tot_cnt` and the buckets `cnt[][]` will be zeroed during there usage for calculation in `tfw_apm_prnctl_calc`, that leads to BUG_ON.
We can use lock to prevent race condition, but more effective way is revert (903cddae) and break the loop if race occurs and recalculate statistics on the next call.

### DIFF
--- a/fw/apm.c
+++ b/fw/apm.c
@@ -969,10 +969,10 @@ tfw_apm_rbctl_update(TfwApmData *data, bool recalc)
 /*
  * Calculate the latest percentiles if necessary.
  *
- * Return 0 if the calculation is successful.
- * Return > 0 and < @prcntlsz if the calculation is incomplete.
+ * Return true if the calculation is successful.
+ * Return false if the calculation is incomplete.
  */
-static int
+static bool
 tfw_apm_calc(TfwApmData *data)
 {
 	int nfilled, recalc;
@@ -989,7 +989,7 @@ tfw_apm_calc(TfwApmData *data)
 
 	recalc = test_and_clear_bit(TFW_APM_DATA_F_RECALC, &data->flags);
 	if (!tfw_apm_rbctl_update(data, recalc))
-		return 0;
+		return true;
 
 	nfilled = tfw_apm_prnctl_calc(&data->rbuf, &data->rbctl, &pstats);
 	if (nfilled < T_PSZ) {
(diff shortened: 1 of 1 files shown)


fw/apm.c
```diff
@@ -950,18 +987,24 @@ tfw_apm_calc(TfwApmData *data)
...
+		T_DBG3("%s: Percentile values may have changed.\n", __func__);
+		write_seqlock(&asent->seqlock);
+		memcpy_fast(asent->pstats.val, pstats.val,
+			    T_PSZ * sizeof(asent->pstats.val[0]));
+		atomic_inc(&data->stats.rdidx);
+		write_sequnlock(&asent->seqlock);
+	}
 
-	return;
+	return nfilled % T_PSZ;
```
krizhanovsky, line 1007:
```suggestion
	return nfilled < T_PSZ;
```
and make the function to return `bool`? In my understanding this is no way to get `nfilled > T_PSZ`

krizhanovsky (approved):
LGTM

keshonok:
> `tfw_apm_update->__tfw_apm_update->tfw_apm_rbent_checkreset->__tfw_apm_rbent_reset`:
> 
> ```
> memset(crbent->pcntrng.__reset_from, 0,
> 	       offsetof(TfwPcntRanges, __reset_till)
> 	       - offsetof(TfwPcntRanges, __reset_from));
> ```
> 
> It seems that it was in `903cddae857658239cdc9b898970b03988cd2d13`

Thanks, found it. I would greatly appreciate if you could describe the scenario for how exactly that BUG_ON() could be triggered? It might be obvious, but I can't see one now.

EvgeniiMekhanik:
@keshonok  look:
In `tfw_apm_prnctl_calc` we calculate `cnt = pcntrng->cnt[st[i].r][st[i].b]` on each loop iteration. But  `pcntrng->cnt` array can be zeroed on separate cpu in `__tfw_apm_rbent_reset`. In this case `pval[p] <= cnt` will be always false (we don't increment cnt!) and p is also not incremented. In this case on one of the loop iteration BUG_ON occurs

const-t (approved):

PR #2309, opened 2024-12-21
Vk 2052 add ja5 filter

Kutumov:


tls/ttls.c
```diff
@@ -2874,6 +2886,10 @@ ttls_init(void)
 	if (!ttls_hs_cache)
 		goto err_free;
 
+	if (tls_get_ja5_storage_size())
```
const-t, line 2889: In this PR we breaking project architecture introducing new cross-module dependencies: `tls` depends on `fw` and `tdb`, `lib` depends on `tdb`. I must point that modules loads in the order and these dependencies breaks the order. In `ja5_filter.c` we intermix helper functions and code that works with `tdb`, therefore each module who uses `lib` will depends on `tdb`, but `lib` must be independent.

I'm not sure that we must spread ja5 logic across the modules, looks like we can place this logic into `fw` module, then just configure `tls` as we do it now. I mean we must calculate rates and hit db relies on "events". We even can make new submodule `ja5` in `fw`(like `client` or `cache`). In lib we can place only function helpers that may be used from each module without new dependencies and completely stateless(now it's not).

Yes, I see that `ja5t` hash calculation is spread between `fw` and `tls`, but it's not looks so dangerous at least now, because we relies only on stateless, independent(means it not uses other modules) `lib` module here.

const-t (changes requested):

fw/ja5_conf.c
```diff
@@ -0,0 +1,243 @@
...
+ja5_cfgop_handle_hash_entry(TfwCfgSpec *cs, TfwCfgEntry *ce)
+{
+	TlsJa5t hash;
+	u32 conns_per_sec;
+	u32 recs_per_sec;
+	TlsJa5HashEntry *he;
+	u64 key;
+
+	BUILD_BUG_ON(sizeof(hash) > sizeof(u64));
+	TFW_CFG_CHECK_VAL_EQ_N(3, cs, ce);
```
EvgeniiMekhanik, line 135: We still have memory leak.
Try:

ja5t max_entries=200 {
    hash 123 0 10;
}

then

ja5t max_entries=200 {
    hash sdfdsfds 0 10;
}

First successfully allocated config will never freed. You can check it if you build kernel with kmemleak or just print allocated|deallocated addresses

fw/ja5_conf.c
```diff
@@ -0,0 +1,243 @@
...
+
+static void
+free_cfg(TlsJa5FilterCfg *cfg)
+{
+	u32 bkt_i;
+	TlsJa5HashEntry *entry;
+
+	BUG_ON(!cfg);
+
+	hash_for_each(cfg->hashes, bkt_i, entry, hlist)
```
EvgeniiMekhanik, line 211: Should we use [hash_for_each_safe](https://elixir.bootlin.com/linux/v6.12.4/C/ident/hash_for_each_safe) ?
As read in documentation hash_for_each_safe - iterate over a hashtable safe against removal of hash entry

fw/ja5_conf.c
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	if (!(he = kmalloc(sizeof(TlsJa5HashEntry), GFP_KERNEL)))
+		return -ENOMEM;
+
+	he->ja5_hash = hash;
+	he->conns_per_sec = conns_per_sec;
+	he->records_per_sec = recs_per_sec;
+	INIT_HLIST_NODE(&he->hlist);
+	he->refcnt.counter = 1;
```
EvgeniiMekhanik, line 163: `atomic_set` should be used

fw/ja5_conf.c
```diff
@@ -0,0 +1,243 @@
...
+
+int
+ja5_cfgop_begin(TfwCfgSpec *cs, TfwCfgEntry *ce)
+{
+	BUG_ON(tls_filter_cfg_reconfig);
+	TFW_CFG_CHECK_VAL_EQ_N(0, cs, ce);
+	TFW_CFG_CHECK_ATTR_LE_N(1, cs, ce);
+
+	if (!(tls_filter_cfg_reconfig =
+		kzalloc(sizeof(TlsJa5FilterCfg), GFP_KERNEL)))
```
EvgeniiMekhanik, line 179: Invalid aligment

fw/ja5_conf.c
```diff
@@ -0,0 +1,243 @@
...
+	}
+	rcu_read_unlock_bh();
+
+	return entry;
+}
+
+static void
+tls_put_ja5_hash_entry(TlsJa5HashEntry *entry)
+{
+	if (entry && !atomic64_dec_return(&entry->refcnt))
```
EvgeniiMekhanik, line 79: It's better to check that `atomic64_dec_return` doesn't return negative value (previously we have a lot of bugs in such code).
```
r = atomic64_dec_return(&entry->refcnt);
BUG_ON(r < 0)
if (!r)
   kfree(entry)
```

fw/ja5_conf.c
```diff
@@ -0,0 +1,243 @@
...
+	DECLARE_HASHTABLE(hashes, TLS_JA5_HASHTABLE_BITS);
+} TlsJa5FilterCfg;
+
+static TlsJa5FilterCfg __rcu	*tls_filter_cfg;
+static TlsJa5FilterCfg		*tls_filter_cfg_reconfig;
+
+static TlsJa5HashEntry*
+tls_get_ja5_hash_entry(TlsJa5t fingerprint)
+{
+	u64 key = hash_calc((char *)&fingerprint, sizeof(fingerprint));
```
EvgeniiMekhanik, line 55: May be it's better to calculate hash after checking that `tls_filter_cfg` is not NULL to avoid extra calculation.

fw/ja5_conf.c
```diff
@@ -0,0 +1,243 @@
...
+	TlsJa5FilterCfg *cfg;
+
+	if (!tls_filter_cfg)
+		return NULL;
+
+	rcu_read_lock_bh();
+	cfg = rcu_dereference_bh(tls_filter_cfg);
+	hash_for_each_possible(cfg->hashes, entry, hlist, key) {
+		if (!memcmp(&fingerprint, &entry->ja5_hash, 
+						sizeof(fingerprint))) {
```
EvgeniiMekhanik, line 66: Invalid aligment

fw/ja5_conf.c
```diff
@@ -0,0 +1,243 @@
...
+
+/* Define default size as multiple of TDB extent size */
+#define TLS_JA5_DEFAULT_STORAGE_SIZE ((1 << 21) * 25)
+#define TLS_JA5_HASHTABLE_BITS 10
+
+typedef struct {
+	struct hlist_node	hlist;
+	atomic64_t		refcnt;
+	/* TODO: make an unified macro for different types of ja5 hashes */
+	TlsJa5t			ja5_hash;
```
EvgeniiMekhanik, line 39: `TlsJa5t` is not packed need to discuss on mitting

fw/ja5_conf.c
```diff
@@ -0,0 +1,243 @@
...
+tls_put_ja5_hash_entry(TlsJa5HashEntry *entry)
+{
+	if (entry && !atomic64_dec_return(&entry->refcnt))
+		kfree(entry);
+}
+
+u64
+tls_get_ja5_conns_limit(TlsJa5t fingerprint)
+{
+	u64 res = -1;
```
EvgeniiMekhanik, line 86: U64_MAX

fw/ja5_conf.c
```diff
@@ -0,0 +1,243 @@
...
+		tls_put_ja5_hash_entry(e);
+	}
+
+	return res;
+}
+
+u64
+tls_get_ja5_recs_limit(TlsJa5t fingerprint)
+{
+	u64 res = -1;
```
EvgeniiMekhanik, line 100: U64_MAX

lib/ja5_filter.h
```diff
@@ -0,0 +1,208 @@
...
+		/* Allow connection if DB is full */
+		return 0;
+
+	spin_lock(&rates->conns_lock);
+	res = ja5_calc_rate(rates->conns);
+	spin_unlock(&rates->conns_lock);
+
+	put_fingerprint_rates(rates);
+
+	T_LOG_NL("JA5 Fingerprint %08llx: connections/sec %d",
```
EvgeniiMekhanik, line 179: T_DBG

lib/ja5_filter.h
```diff
@@ -0,0 +1,208 @@
...
+		/* Allow record if DB is full */
+		return 0;
+
+	spin_lock(&rates->recs_lock);
+	res = ja5_calc_rate(rates->recs);
+	spin_unlock(&rates->recs_lock);
+
+	put_fingerprint_rates(rates);
+
+	T_LOG_NL("JA5 Fingerprint %08llx: records/sec %d",
```
EvgeniiMekhanik, line 204: T_DBG

fw/tls.c
```diff
@@ -1160,6 +1180,11 @@ tfw_tls_start(void)
 {
 	tfw_tls_allow_any_sni = allow_any_sni_reconfig;
 
+	if (tls_get_ja5_storage_size()) {
+		if (!ja5t_init_filter(tls_get_ja5_storage_size()))
```
EvgeniiMekhanik, line 1184: Save `tls_get_ja5_storage_size` result to avoid second call of this function

lib/ja5_filter.h
```diff
@@ -0,0 +1,208 @@
...
+		list_add_tail(&rates->list_node, &storage.lru_list);
+	spin_unlock(&storage.lru_list_lock);
+
+	return rates;
+}
+
+static bool
+init_filter(size_t max_storage_size)
+{
+	if (storage.tdb)
```
EvgeniiMekhanik, line 127: I catch that `storage.tdb` is not NULL under heavy load and during Tempesta FW reload. It seems that we can't reload Tempesta FW under heavy load with ja5 setting in config

lib/ja5_filter.h
```diff
@@ -0,0 +1,208 @@
...
+}
+
+static void
+put_fingerprint_rates(Rates *rates)
+{
+	tdb_rec_put(storage.tdb, (char *)rates - sizeof(TdbRec));
+}
+
+static void
+get_alloc_ctx_init_rec(TdbRec *rec, void *)
```
EvgeniiMekhanik, line 63: It seems that we should zeroed `rates` here. Please check is `rates` timeslots is zeroed here?

lib/ja5_filter.h
```diff
@@ -0,0 +1,208 @@
...
+static struct {
+	TDB			*tdb;
+	struct list_head	lru_list;
+	spinlock_t		lru_list_lock;
+} storage;
+
+static bool
+get_alloc_ctx_eq_rec(TdbRec *, void *)
+{
+	return true;
```
EvgeniiMekhanik, line 53: It seems that since we return always true here, we always return first entry from tdb. I mean that in `get_fingerprint_rates` we always catch first entry isn't it?

EvgeniiMekhanik (changes requested):
Need fixes

lib/ja5_filter.h
Kutumov: record are always zeroed by TDB during allocation

lib/ja5_filter.h
Kutumov: Yes we always catch the first entry, and it's alright since we use fingerprints as keys, not hashes.

fw/ja5_conf.c
Kutumov: I decided not to pack fingerprint in order to easily use it as a u64 key in hash tables.
Packing it seems low profitable because if you want to save 1 byte in it you always have to pack all structures that include it

fw/ja5_conf.c
```diff
@@ -0,0 +1,203 @@
...
+		return -EINVAL;
+	}
+
+	if (tfw_cfg_parse_uint(ce->vals[2], &recs_per_sec)) {
+		T_ERR_NL("Failed to parse hash entry in ja5 section: "
+			"invalid records per second value %s", ce->vals[2]);
+		return -EINVAL;
+	}
+
+	if (kstrtou64(ce->vals[0], 16, (u64 *)&hash)) {
```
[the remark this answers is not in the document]
Kutumov: tfw_cfg_parse_<type> functions try to detect numbers base by their prefixes such as 'b' or '0x'. Hashes dont use prefixed as by the task. So tfw_cfg_parse_ulong is inapplicable here

fw/ja5_conf.c
Kutumov: Could you show the correct alignment?

fw/ja5_conf.c
Kutumov: First successfully allocated config `tls_filter_cfg `  doesnot have to freed if reconfiguration fails. It remains used by tempesta.

fw/ja5_conf.c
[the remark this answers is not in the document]
Kutumov: Applied tfw_cfg_parse_ where possible

lib/ja5_filter.h
Kutumov: Now true is returned here if tdb is already initialized. We decided not to reconfigure filters db size on the fly because it's hard to reallocate it

fw/ja5_conf.c
```diff
@@ -0,0 +1,203 @@
...
+		return -ENOMEM;
+
+	if (ce->attr_n == 1) {
+		if (strcasecmp(ce->attrs[0].key, "storage_size")) {
+			T_ERR_NL("Failed to parse ja5 section: "
+			"invalid attribute %s", ce->attrs[0].key);
+			return -EINVAL;
+		}
+
+		if (kstrtoul(ce->attrs[0].val, 10,
```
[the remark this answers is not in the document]
Kutumov: done

tls/ttls.h
```diff
@@ -31,6 +31,7 @@
 #include <net/tls.h>
 
 #include "lib/str.h"
+#include "lib/ja5.h"
```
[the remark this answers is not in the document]
Kutumov: `TlsJa5t` from lib/ja5.h is used in this file

tls/tls_srv.c
```diff
@@ -2243,6 +2245,17 @@ ttls_handshake_server_step(TlsCtx *tls, unsigned char *buf, size_t len,
 		r = ttls_parse_client_hello(tls, buf, len, hh_len, read);
 		if (r)
 			return r;
+		else {
```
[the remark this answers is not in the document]
Kutumov: done

tls/ttls.c
Kutumov: done

fw/ja5_conf.c
Kutumov: done

fw/ja5_conf.c
Kutumov: done

fw/ja5_conf.c
Kutumov: done

fw/tls.c
Kutumov: done

lib/ja5_filter.h
Kutumov: done

lib/ja5_filter.h
Kutumov: done

fw/ja5_conf.c
Kutumov: done

fw/ja5_conf.c
Kutumov: done

fw/ja5_conf.c
Kutumov: done

fw/ja5_conf.c
Kutumov: done

lib/ja5_filter.h
Kutumov: But i think it's better to switch to hashed as keys

fw/ja5_conf.c
Kutumov: I added freeing of `tls_filter_cfg ` in `ja5_cfgop_cleanup`

lib/ja5_filter.h
EvgeniiMekhanik: It seems that we zeroed record, not record data. Isn't it?

EvgeniiMekhanik:
I write tests https://github.com/tempesta-tech/tempesta-test/pull/746 and always catch deadlock on the stress tests

lib/ja5_filter.h
Kutumov: U r right. Fixed

lib/ja5_filter.h
```diff
@@ -0,0 +1,214 @@
...
+		spin_lock(&storage.lru_list_lock);
+		list_for_each_prev_safe(pos, tmp, &storage.lru_list) {
+			list_move(pos, &tail_to_delete);
+			if (!--cnt)
+				break;
+		}
+		spin_unlock(&storage.lru_list_lock);
+
+		list_for_each_safe(pos, tmp, &tail_to_delete)
+			put_fingerprint_rates((Rates *)pos);
```
const-t, line 105: Here you must call `tdb_entry_remove`

const-t (changes requested):

lib/ja5_filter.h
```diff
@@ -0,0 +1,214 @@
...
+	while (!(rec = tdb_rec_get_alloc(storage.tdb, key, &get_alloc_ctx))) {
+		struct list_head *pos, *tmp, tail_to_delete;
+		u32 cnt = DB_RECS_TO_FREE_CNT;
+
+		INIT_LIST_HEAD(&tail_to_delete);
+
+		/* Cut off DB_RECS_TO_FREE_CNT entries from the LRU list */
+		spin_lock(&storage.lru_list_lock);
+		list_for_each_prev_safe(pos, tmp, &storage.lru_list) {
+			list_move(pos, &tail_to_delete);
```
const-t, line 98: Why do we need `list_move`? Why not simply `list_del` and after remove from tdb?

fw/tls.c
```diff
@@ -1010,6 +1012,24 @@ tfw_tls_alpn_match(const TlsCtx *tls, const ttls_alpn_proto *alpn)
 	return false;
 }
 
+static bool
+tfw_ja5t_limit_conn(TlsJa5t fingerprint)
+{
+	u64 limit = tls_get_ja5_conns_limit(fingerprint);
```
const-t, line 1018: Is there a sense to even try to calculate conns rate if `limit == U64_MAX`? `ja5t_get_conns_rate` doesn't look cheap.

Kutumov: There's no sense

lib/ja5_filter.h
Kutumov: I made cutting tail first and then removing from tdb in order to reduce locking time of `storage.lru_list_lock`

lib/ja5_filter.h
```diff
@@ -0,0 +1,214 @@
...
+ja5_get_records_rate(u64 fingerprint)
+{
+	u32 res;
+	Rates *rates;
+
+	if (!storage.tdb)
+		return 0;
+
+	if (!(rates = get_fingerprint_rates(fingerprint)))
+		/* Allow record if DB is full */
```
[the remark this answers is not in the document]
Kutumov: It's a discussable question. I think i depends on product politics

lib/ja5_filter.h
```diff
@@ -0,0 +1,214 @@
...
+
+typedef struct {
+	/** Keep  @list_node first for easy type casts */
+	struct list_head	list_node;
+	TimeSlot		conns[JA5_FILTER_TIME_SLOTS_CNT];
+	spinlock_t		conns_lock;
+	TimeSlot		recs[JA5_FILTER_TIME_SLOTS_CNT];
+	spinlock_t		recs_lock;
+} Rates;
+
```
krizhanovsky, line 43: extra empty line

lib/ja5_filter.h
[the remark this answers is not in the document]
const-t: After discussion we decided to save current behavior.

lib/ja5_filter.h
```diff
@@ -0,0 +1,214 @@
...
+	}
+	slot->counter++;
+
+	for (; slot->ts > end_ts;
+		slot_num = (slot->ts - 1) % JA5_FILTER_TIME_SLOTS_CNT,
+		slot = &slots[slot_num])
+		sum += slot->counter;
+
+	return sum;
+}
```
krizhanovsky, line 164: The computation of the rate limit under spin lock doesn't look too advances. However, at the moment we use somewhat similar rate limit in the XDP volumetric DDoS mitigation module and it easily process more than 100Gbps on adequate CPU power.

Scalable and efficient rate limits are hard. There is [Hierarchical Heavy Hitters with the Space Saving Algorithm](https://arxiv.org/pdf/1102.5540.pdf) proposes probabilistic space efficient approach, which is said is easy to be parallelized and merged. We also discussed per-cpu hierarchical approach.

I think we firstly need to solve the problem on the XDP  (and more performance crucial) level and next in L7 (Tempesta FW).

krizhanovsky (approved):
There is nothing good about the spinlocks, but I think we good at this stage and will move to scalable rate limiters later

fw/tls.c
Kutumov: Actually there's a sense, because a call of `ja5t_get_conns_rate` is required to increment the counter

fw/ja5t_filter.c
```diff
@@ -0,0 +1,44 @@
...
+	return ja5_get_conns_rate(*(u64 *)&fingerprint);
+}
+
+u32
+ja5t_get_records_rate(TlsJa5t fingerprint)
+{
+	BUILD_BUG_ON(sizeof(fingerprint) != sizeof(u64));
+
+	return ja5_get_records_rate(*(u64 *)&fingerprint);
+}
```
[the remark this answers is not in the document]
Kutumov: `ja5_get_records_rate`, `ja5_get_conns_rate` and `init_filter` must be called from a separate file because `ja5_filter.h` hold static data operated by these functions

lib/ja5_filter.h
```diff
@@ -155,10 +154,9 @@ ja5_calc_rate(TimeSlot slots[])
 	}
 	slot->counter++;
 
-	for (; slot->ts > end_ts;
-		slot_num = (slot->ts - 1) % JA5_FILTER_TIME_SLOTS_CNT,
-		slot = &slots[slot_num])
-		sum += slot->counter;
+	for (slot_num = 0; slot_num < JA5_FILTER_TIME_SLOTS_CNT; slot_num++)
+		if (slot[slot_num].ts + JA5_FILTER_TIME_SLOTS_CNT >= ts)
```
Kutumov, line 158: Not `slot[slot_num]` but `slot**s**[slot_num]`

lib/ja5_filter.h
```diff
@@ -0,0 +1,214 @@
+/**
```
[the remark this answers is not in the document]
Kutumov: moved

lib/ja5_filter.h
Kutumov: done

lib/ja5_filter.h
Kutumov: done

Makefile
```diff
@@ -1,7 +1,7 @@
 #		Tempesta FW
 #
 # Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
-# Copyright (C) 2015-2022 Tempesta Technologies, Inc.
+# Copyright (C) 2015-2024 Tempesta Technologies, Inc.
```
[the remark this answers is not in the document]
Kutumov: new changes appeared

etc/tempesta_fw.conf
```diff
@@ -1421,3 +1421,40 @@
 # Example:
 #   access_log dmesg mmap mmap_host=localhost mmap_log=access.log;
 #
+
+# TAG: ja5t
+#
+# Specifies TLS filtering behaviour: which Ja5 hashes and how to filter.
+#
+# Syntax:
+#   ja5t storage_size=<STORAGE_SIZE> {
```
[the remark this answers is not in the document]
Kutumov: Added docs for ja5h. But I dont understand what you mean under `ja5`

fw/cfg.h
```diff
@@ -194,26 +194,71 @@ typedef struct {
...
+#define TFW_CFG_CHECK_NO_ATTRS(spec, entry)			\
+	do {							\
+		if ((entry)->attr_n) {				\
+			T_ERR_NL("%s: Arguments may not have "	\
+				"the '=' sign\n", (spec)->name);\
+			return -EINVAL;				\
+		}						\
+	} while (0)
+
+#define TFW_CFG_CHECK_ATTR_N(op, req, spec, entry)			\
```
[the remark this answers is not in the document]
Kutumov: done

fw/cfg.h
```diff
@@ -194,26 +194,71 @@ typedef struct {
...
-		code;							\
-	}								\
+	do {								\
+		if (!((entry)->val_n op (req))) {			\
+			T_ERR_NL("%s: Invalid number of arguments: "	\
+			"%zu, must be %s %d\n",				\
+			(spec)->name, (entry)->val_n, #op, (req));	\
+			return -EINVAL;					\
+		}							\
+	} while (0)
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.c
```diff
@@ -0,0 +1,267 @@
...
+ */
+
+#include <linux/list.h>
+#include <linux/hashtable.h>
+#include <linux/slab.h>
+
+#include "tempesta_fw.h"
+#include "log.h"
+#include "ja5_conf.h"
+#include "hash.h"
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.c
```diff
@@ -0,0 +1,267 @@
...
+
+/* Define default size as multiple of TDB extent size */
+#define TLS_JA5_DEFAULT_STORAGE_SIZE ((1 << 21) * 25)
+#define TLS_JA5_HASHTABLE_BITS 10
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.c
```diff
@@ -0,0 +1,267 @@
...
+
+typedef struct {
+	struct hlist_node	hlist;
+	atomic_t		refcnt;
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.c
```diff
@@ -0,0 +1,267 @@
...
+
+	key = hash_calc((char *)&fingerprint, sizeof(fingerprint));
+
+	rcu_read_lock_bh();
+	cfg = rcu_dereference_bh(tls_filter_cfg);
+	hash_for_each_possible(cfg->hashes, entry, hlist, key) {
+		if (!memcmp(&fingerprint, &entry->ja5_hash, 
+			    sizeof(fingerprint)))
+		{
+			atomic_inc(&entry->refcnt);
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.h
```diff
@@ -0,0 +1,38 @@
...
+
+u64 tls_get_ja5_recs_limit(TlsJa5t fingerprint);
+
+int ja5_cfgop_handle_hash_entry(TfwCfgSpec *cs, TfwCfgEntry *ce);
+
+int ja5_cfgop_begin(TfwCfgSpec *cs, TfwCfgEntry *ce);
+
+int ja5_cfgop_finish(TfwCfgSpec *cs);
+
+void ja5_cfgop_cleanup(TfwCfgSpec *cs);
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5t_filter.c
```diff
@@ -0,0 +1,44 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 59
+ * Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ */
+
```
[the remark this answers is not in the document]
Kutumov: done

lib/ja5_filter.h
```diff
@@ -0,0 +1,214 @@
...
+}
+
+static u32
+ja5_calc_rate(TimeSlot slots[])
+{
+	u32 sum = 0;
+	u64 ts = jiffies * JA5_FILTER_TIME_SLOTS_CNT / HZ;
+	u64 end_ts = ts - JA5_FILTER_TIME_SLOTS_CNT;
+	u8 slot_num = ts % JA5_FILTER_TIME_SLOTS_CNT;
+	TimeSlot *slot = &slots[slot_num];
```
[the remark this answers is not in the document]
Kutumov: done

etc/tempesta_fw.conf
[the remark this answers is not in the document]
krizhanovsky: With `ja5` I meant the whole hash, including HTTP, TLS and TCP/IP in future

fw/ja5t_filter.c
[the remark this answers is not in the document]
krizhanovsky: @Kutumov I didn't get 

> because ja5_filter.h hold static data operated by these functions

I don't see any static variable in fw/ja5_filter.h . Which static data do you mean?

fw/ja5_conf.c
```diff
@@ -0,0 +1,249 @@
...
+		return -ENOMEM;
+
+	he->ja5_hash = hash;
+	he->conns_per_sec = conns_per_sec;
+	he->records_per_sec = recs_per_sec;
+	INIT_HLIST_NODE(&he->hlist);
+	refcount_set(&he->refcnt, 1);
+
+	key = hash_calc((char *)&hash, sizeof(hash));
+	hash_add(tls_filter_cfg_reconfig->hashes, &he->hlist, key);
```
EvgeniiMekhanik, line 169: It seems that we add several entries with same hash to the table one by one  and then use only first one (the last added). It seems that we can delete previously added entries.

Kutumov: I will fix it in PR 2310

fw/ja5_filter.h
```diff
@@ -0,0 +1,257 @@
...
+		spin_lock(&storage.lru_list_lock);
+		list_for_each_prev_safe(pos, tmp, &storage.lru_list) {
+			list_move(pos, &tail_to_delete);
+			if (!--cnt)
+				break;
+		}
+		spin_unlock(&storage.lru_list_lock);
+
+		list_for_each_safe(pos, tmp, &tail_to_delete)
+			tdb_entry_remove(storage.tdb, key, NULL, NULL, true);
```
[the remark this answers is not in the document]
Kutumov: fixed

fw/ja5_filter.h
```diff
@@ -0,0 +1,260 @@
...
+		list_for_each_prev_safe(pos, tmp, &storage.lru_list) {
+			list_move(pos, &tail_to_delete);
+			if (!--cnt)
+				break;
+		}
+		spin_unlock(&storage.lru_list_lock);
+
+		list_for_each_safe(pos, tmp, &tail_to_delete) {
+			u64 key = ((TdbRec *)pos - 1)->key;
+			// TODO: remove directly by record bypassing search by key
```
const-t, line 133:
```suggestion
			/* TODO: remove directly by record bypassing search by key */
```

const-t: Even better: /* TODO #515: remove directly by record bypassing search by key */

PR #2310, opened 2024-12-25
Add ja5 http filter and config

Kutumov:


fw/ja5_conf.h
```diff
@@ -20,19 +20,242 @@
...
+get_ja5_hash_entry(JA5_STRUCT_T fingerprint)
+{
+	u64 key;
+	Ja5HashEntry *entry = NULL;
+	Ja5FilterCfg *cfg;
+
+	if (!filter_cfg)
+		return NULL;
+
+	key = hash_calc((char *)&fingerprint, sizeof(fingerprint));
```
const-t, line 69: Hash from hash that introduce collision, seems radix more applicable here, however it's just note, don't change anything.

Maybe we can use ja5 as key for hashtable without crc hashing?

const-t (changes requested):

fw/ja5_conf.h
```diff
@@ -20,19 +20,242 @@
  * Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  */
 
+#pragma once
+
```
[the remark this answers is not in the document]
krizhanovsky: I agree. Just wanted to add that typically we keep some functions implemented in .h to make them inline in all .c files (e.g. Linux kernel linked list routines). But in this case, for example `get_ja5_conns_limit()` is used on conf files only, so  it's not performance critical, so it's better to implement it in a .c file.

Also at the moment there are too many ja5 files. There is no reason to have spearate conf files for http and tls. I'd suggest to have 2 .c and 2 .h files at most - main logic and configuration. But probably even configuration can be merged with the main logic. Please note that we can't organize the linux kernel module code in subfolders. Having much larger subsustems in same files, it's unnatural for the project to keep configuration for ja5 http and tls separately.

fw/ja5_conf.h
krizhanovsky: Yeah, maybe it makes sense just to leave a TODO comment here

fw/ja5_conf_http.c
```diff
@@ -0,0 +1,67 @@
...
+http_ja5_cfgop_finish(TfwCfgSpec *cs)
+{
+	return ja5_cfgop_finish(cs);
+}
+
+void
+http_ja5_cfgop_cleanup(TfwCfgSpec *cs)
+{
+	ja5_cfgop_cleanup(cs);
+}
```
krizhanovsky, line 67: Do we really need the bunch of single-call wrapping functions passing just the same arguments? This is a simple aliasing and since this is in C file, the compiler can't optimize out the wrappers and we have two copies of the same binary code.

fw/ja5_conf_tls.c
```diff
@@ -0,0 +1,67 @@
...
+tls_ja5_cfgop_finish(TfwCfgSpec *cs)
+{
+	return ja5_cfgop_finish(cs);
+}
+
+void
+tls_ja5_cfgop_cleanup(TfwCfgSpec *cs)
+{
+	ja5_cfgop_cleanup(cs);
+}
```
krizhanovsky, line 67: Let's remove the wrappers. They might have sense, but only if we have any plans to extend the API, which isn't the case for now. If we leave the code, then most likely we'll just live with the wrappers forever

krizhanovsky (changes requested):
The most of the pull request is just a bunch of wrapping code with no any logic. I don't see a reason for wrappers without any concrete plans in extending the API, which can make real logic from the simple wrappers

fw/ja5_conf.h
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.h
```diff
@@ -20,19 +20,242 @@
...
+	key = hash_calc((char *)&fingerprint, sizeof(fingerprint));
+
+	rcu_read_lock_bh();
+	cfg = rcu_dereference_bh(filter_cfg);
+	hash_for_each_possible(cfg->hashes, entry, hlist, key) {
+		if (!memcmp(&fingerprint, &entry->ja5_hash,
+			sizeof(fingerprint))) {
+			atomic_inc(&entry->refcnt);
+			break;
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.h
```diff
@@ -20,19 +20,242 @@
...
+	Ja5HashEntry *he;
+	u64 key;
+
+	BUILD_BUG_ON(sizeof(hash) > sizeof(u64));
+	TFW_CFG_CHECK_VAL_EQ_N(3, cs, ce);
+	TFW_CFG_CHECK_NO_ATTRS(cs, ce);
+
+	if (tfw_cfg_parse_uint(ce->vals[1], &conns_per_sec)) {
+		T_ERR_NL("Failed to parse hash entry in ja5 section: "
+			"invalid connections per second value %s", ce->vals[1]);
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.h
```diff
@@ -20,19 +20,242 @@
...
+		return -EINVAL;
+	}
+
+	if (tfw_cfg_parse_uint(ce->vals[2], &recs_per_sec)) {
+		T_ERR_NL("Failed to parse hash entry in ja5 section: "
+			"invalid records per second value %s", ce->vals[2]);
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.h
```diff
@@ -20,19 +20,242 @@
...
+		return -EINVAL;
+	}
+
+	if (kstrtou64(ce->vals[0], 16, (u64 *)&hash)) {
+		T_ERR_NL("Failed to parse hash entry in ja5 section: "
+			"invalid hash value %s", ce->vals[0]);
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.h
```diff
@@ -20,19 +20,242 @@
...
+	TFW_CFG_CHECK_ATTR_LE_N(1, cs, ce);
+
+	if (!(filter_cfg_reconfig =
+		kzalloc(sizeof(Ja5FilterCfg), GFP_KERNEL)))
+		return -ENOMEM;
+
+	if (ce->attr_n == 1) {
+		if (strcasecmp(ce->attrs[0].key, "storage_size")) {
+			T_ERR_NL("Failed to parse ja5 section: "
+			"invalid attribute %s", ce->attrs[0].key);
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.h
```diff
@@ -20,19 +20,242 @@
...
+			return -EINVAL;
+		}
+
+		if (tfw_cfg_parse_ulonglong(ce->attrs[0].val,
+			&filter_cfg_reconfig->storage_size)) {
+			T_ERR_NL("Failed to parse ja5 section: "
+				"invalid storage_size value");
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.h
```diff
@@ -20,19 +20,242 @@
...
+			return -EINVAL;
+		}
+	} else {
+		filter_cfg_reconfig->storage_size =
+			JA5_DEFAULT_STORAGE_SIZE;
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.h
Kutumov: added a comment

fw/ja5_conf.h
```diff
@@ -20,19 +20,242 @@
...
+#include "hash.h"
+#include "tempesta_fw.h"
+
+#ifndef JA5_STRUCT_T
+#error "JA5_STRUCT_T macro is not defined before include of ja5_conf.h"
+#endif
+
+/* Define default size as multiple of TDB extent size */
+#define JA5_DEFAULT_STORAGE_SIZE ((1 << 21) * 25)
+#define JA5_HASHTABLE_BITS 10
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf.h
```diff
@@ -20,19 +20,242 @@
...
+
+typedef struct {
+	struct hlist_node	hlist;
+	atomic_t		refcnt;
+	JA5_STRUCT_T		ja5_hash;
```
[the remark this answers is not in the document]
Kutumov: done

fw/ja5_conf_tls.c
Kutumov: API will be extended when we will add TCP and mixed fingerprints

krizhanovsky (approved):
I don't have any more comments, but a careful reviews from @EvgeniiMekhanik and @const-t are required

fw/http.c
```diff
@@ -7870,6 +7876,18 @@ tfw_cfgop_cleanup_max_header_list_size(TfwCfgSpec *cs)
 	max_header_list_size = 0;
 }
 
+static TfwCfgSpec tfw_tls_hash_specs[] = {
```
const-t, line 7879: Can we move all `ja5` configuration related code to `ja5_conf.c` to have all in one place? Like we do for vhost.

fw/http.c
```diff
@@ -7957,6 +7975,20 @@ static TfwCfgSpec tfw_http_specs[] = {
 		.allow_none = true,
 		.cleanup = tfw_cfgop_cleanup_max_header_list_size,
 	},
+	{
+		.name = "ja5h",
+		.deflt = NULL,
+		.handler = tfw_cfg_handle_children,
+		.cleanup = http_ja5_cfgop_cleanup,
+		.dest = tfw_tls_hash_specs,
```
const-t, line 7983:
```suggestion
		.dest = tfw_http_hash_specs,
```

fw/http.c
const-t, line 7879:
```suggestion
static TfwCfgSpec tfw_http_hash_specs[] = {
```

const-t (approved):
LGTM. Except comments that overlaps with 2309.

fw/http.c
Kutumov: done

fw/http.c
Kutumov: done

fw/http.c
Kutumov: done

PR #2312, opened 2024-12-30
Mekhanik evgenii/2262 dbg

EvgeniiMekhanik:


keshonok:
Perhaps, the real question here is: Is the use of tfw_pool functions correct and justified in the vhost case? I have a feeling that the intention for tfw_pool was different.

fw/pool.c
```diff
@@ -74,19 +74,19 @@ tfw_pool_alloc_pages(unsigned int order)
 	unsigned long pg_res;
```
krizhanovsky, line 74: Please update the copyright year

krizhanovsky (approved):
Good catch!

keshonok:
Perhaps, my comment was misunderstood, so here's my attempt to assert the thought it expressed. If that's not the case, then please ignore it.

The TFW pool was designed as a pure _per-CPU pool_. _"Since Tempesta handles a message only on one CPU, then the pool should not bother about concurrency."_ - that's what's said in the comment in the source code. This is a very specific design which conforms to Tempesta's design of uninterruptible message processing in one go.

However, in vhost code one of the developers didn't quite realize that and used TFW pool as a regular memory pool for allocating chunks of memory to store configuration values. And yes, that code can be and is called from the user context.  I think that's the only place where something like that is done.

I believe that it's vhost code that should be fixed. It's not an incorrect synchronization in pool.c, it's an incorrect use of TFW pool. Otherwise TFW pool's description in the source code should be reworked.

const-t (approved):

EvgeniiMekhanik:
> Perhaps, my comment was misunderstood, so here's my attempt to assert the thought it expressed. If that's not the case, then please ignore it.
> 
> The TFW pool was designed as a pure _per-CPU pool_. _"Since Tempesta handles a message only on one CPU, then the pool should not bother about concurrency."_ - that's what's said in the comment in the source code. This is a very specific design which conforms to Tempesta's design of uninterruptible message processing in one go.
> 
> However, in vhost code one of the developers didn't quite realize that and used TFW pool as a regular memory pool for allocating chunks of memory to store configuration values. And yes, that code can be and is called from the user context. I think that's the only place where something like that is done.
> 
> I believe that it's vhost code that should be fixed. It's not an incorrect synchronization in pool.c, it's an incorrect use of TFW pool. Otherwise TFW pool's description in the source code should be reworked.

During our discussion we decide to don't remove using pool in `vhost.c`. It seems that it don't not affect performance.

keshonok:
> During our discussion we decide to don't remove using pool in `vhost.c`. It seems that it don't not affect performance.

Thanks for the info, that's an excellent point! Perhaps, the description in pool.c should be changed to reflect that and suggest a yet another use pattern - as opposed to the use of regular in-kernel memory allocation at configuration processing stage for the configuration data. This gives the data a certain locality, even though I am not sure it's going to be all that useful.

PR #2315, opened 2025-01-08
Add ja5t and ja5h fingerprints to dmesg and mmap access log

Kutumov:


fw/access_log.c
```diff
@@ -289,6 +293,8 @@ do_access_log_req_mmap(TfwHttpReq *req, u16 resp_status,
 	char *data, *p;
 	struct timespec64 ts;
 	u16 len;
+	TlsJa5t *tls_ja5t = TFW_MSG_H2(req) ? &tfw_tls_context(req->conn)->sess.ja5t : NULL;
```
krizhanovsky, line 296: Why only for HTTP/2? What about HTTP/1 over TLS?

const-t: We can use `TFW_CONN_TLS()` macro here to check the connection type, checking version of HTTP protocol is not needed here.

Kutumov: It's a mistake. TLS check is reqiured. Fixed in PR https://github.com/tempesta-tech/tempesta/pull/2321

PR #2321, opened 2025-01-09
Add ja5h and ja5t access log

Kutumov:


fw/access_log.c
```diff
@@ -375,6 +382,10 @@ do_access_log_req_mmap(TfwHttpReq *req, u16 resp_status,
 		goto drop;
 	}
 
+	if (tls_ja5t)
+		WRITE_FIELD(*tls_ja5t);
+	WRITE_FIELD(req->ja5h);
```
ai-tmpst, line 387: We can't just add these fields here.
As it implemented at this time, every field has a corresponding bit in TfwBinLogEvent::fields. By default is expected that all the fields are present. We reset bits if we need to skip the field:
```
	if (*dropped) {
		WRITE_FIELD(*dropped);
		*dropped = 0;
	} else {
		TFW_MMAP_LOG_FIELD_RESET(event, TFW_MMAP_LOG_DROPPED);
	}
```

You need to add appropriate elements to TfwBinLogFields and TfwBinLogFieldsLens (order matters!).
Also you need to add support of new fields to utils/tfw_logger.cc:
- add column descriptions to tfw_fields[];
- add fields reading to read_access_log_event().

ai-tmpst (changes requested):

fw/access_log.c
ai-tmpst: Almost forgot.
Also you need to add new fields to `table_creation_querry` in utils/clickhouse.cc.
Maybe it would be great to construct this query basing on `tfw_fields` array in utils/tfw/logger.cc, but it quit optional.

ai-tmpst (approved):
LGTM

fw/access_log.c
```diff
@@ -71,7 +71,7 @@
 
 #define MMAP_LOG_PATH "tempesta_mmap_log"
 
-static int access_log_type = ACCESS_LOG_OFF;
+static int access_log_type = ACCESS_LOG_MMAP;
```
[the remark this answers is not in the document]
Kutumov: Forgot to revert it after tests. Foxed

EvgeniiMekhanik (approved):
I have only one question. Can we save hashes in %llx format in clickhouse? Currently I see it in %llu format, so I need to convert it to %llx manually to use it in Tempesta config

Issue #2323, opened 2025-01-13
BUG: Kasan use after free


EvgeniiMekhanik:
**To Reproduce**
Build Kernel with KASAN.
sudo ./run_tests.py t_sites.test_tempesta_tech.TestTempestaTechSite

**Configuration file**
Your configuration file.

**Version or commit hash**
Tempesta: bc9127622ac881661ac629f85433062fac823380
Tempesta-test: ac2ce38c258f0fb91d2dab626a2b6e82e831cd8a

**Stacktrace or debug log**

```
[  417.694361] Start test: t_sites.test_tempesta_tech.TestTempestaTechSite.test_get_resource
[  417.982771] lxdbr0: port 1(vethffb690fb) entered blocking state
[  417.983286] lxdbr0: port 1(vethffb690fb) entered disabled state
[  417.986225] device vethffb690fb entered promiscuous mode
[  417.986764] lxdbr0: port 1(vethffb690fb) entered blocking state
[  417.987236] lxdbr0: port 1(vethffb690fb) entered forwarding state
[  417.987969] lxdbr0: port 1(vethffb690fb) entered disabled state
[  418.273316] phystYtVUP: renamed from veth93c9315a
[  418.319298] eth0: renamed from phystYtVUP
[  418.357414] lxdbr0: port 1(vethffb690fb) entered blocking state
[  418.358364] lxdbr0: port 1(vethffb690fb) entered forwarding state
[  418.588677] Not activating Mandatory Access Control as /sbin/tomoyo-init does not exist.
[  420.343961] [tdb] Start Tempesta DB
[  420.406000] [tempesta fw] Initializing Tempesta FW kernel module...
[  422.043078] [tempesta fw] Warning: Vhost default doesn't have certificate with matching SAN/CN.
[  422.043078]     Maybe that's fine, but it's worth checking the
[  422.043078]     config - if there is no relations between the
[  422.043078]     names, then host name confusion attack is possible.
[  422.274473] [tempesta fw] Configuration processing is completed.
[  422.322792] [tdb] Opened table /opt/tempesta/db/filter0.tdb: size=16777216 rec_size=20 base=00000000e9dab221
[  422.542972] [tdb] Opened table /opt/tempesta/db/cache0.tdb: size=268435456 rec_size=0 base=00000000255553f0
[  422.557751] [tdb] Opened table /opt/tempesta/db/sessions0.tdb: size=16777216 rec_size=312 base=00000000d4bad4f0
[...]

PR #2324, opened 2025-01-14
Fix accessing already closed client db.

EvgeniiMekhanik:
During establishing new TCP connection `tempesta_new_clntsk` callback is called. If Tempesta FW is not loaded it immediately returns 0 (success), otherwise `tfw_classify_conn_estab` is called inside Tempesta FW module. In this function we check if the source port of new created socket is added to to the special bitmap and if not just return success otherwise Tempesta creates new client (We pass to this callback all new created sockets regardless of whether they are related to Еempesta or not! For example when curl connect to nginx new created socket will be passed to this callback also! But we add only Tempesta FW listen sockets prots to bitmap, so for all other cases this callback just return success). The problem occurs when Tempesta unloaded - First of all we stop all Tempesta FW modules and Tempesta FW listen sockets, then unload Tempesta. If some programm (for example nginx) start listen on the same port as Tempesta and curl connect to it new created socket will be passed to Tempesta FW callback and new client will be created (has nothing to do with Tempesta!), moreover this client will be created from already closed db, that leads to BUG. We should remove Tempesta listen socket from bitmap during it's release.

### DIFF
--- a/fw/client.c
+++ b/fw/client.c
@@ -392,8 +392,10 @@ tfw_client_stop(void)
 {
 	if (tfw_runstate_is_reconfig())
 		return;
-	if (client_db)
+	if (client_db) {
 		tdb_close(client_db);
+		client_db = NULL;
+	}
 
 	tfw_client_free_lru();
 }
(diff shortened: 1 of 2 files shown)


krizhanovsky (approved):
Looks good for me, but a careful review of the #2323 crash scenario and its fix is required from @const-t

const-t (approved):
LGTM

PR #2325, opened 2025-01-16
Mekhanik evgenii/bug fixes

EvgeniiMekhanik:


fw/http_frame.c
```diff
@@ -2139,6 +2140,12 @@ do {									\
 		r = tfw_h2_entail_stream_skb(sk, ctx, stream,
 					     &stream->xmit.frame_length,
 					     true);
+		if (unlikely(r)) {
+			T_WARN("Failed to send frame %d", r);
+			return r;
+		}
+		if (unlikely(stream->xmit.postponed) && !ctx->cur_send_headers)
```
const-t, line 2147: Is it really rare case? False prediction it worse than no predication at all.

const-t (approved):
LGTM

fw/ss_skb.c
```diff
@@ -1010,18 +1032,27 @@ __ss_skb_cutoff(struct sk_buff *skb_head, struct sk_buff *skb, char *ptr,
...
+	int _, to_delete;
 
 	while (len) {
 		bzero_fast(&it, sizeof(TfwStr));
-		r = skb_fragment(skb_head, skb, ptr, -len, &it, &_);
+		to_delete = unlikely(abs(len) > PAGE_SIZE) ?
+			PAGE_SIZE : len;
+		next = skb->next;
+
+		r = skb_fragment(skb_head, skb, ptr, -to_delete, &it, &_);
```
krizhanovsky, line 1045: When and who deletes more than 4KB data? IIRC we considered this case, but ended up not to spend resources for this case and maybe we even have warnings/bugons for too large deletions.

krizhanovsky (approved):
LGTM, but just have a small question

fw/http_frame.c
EvgeniiMekhanik: Good point! Under the load it's not a rare case.

fw/ss_skb.c
EvgeniiMekhanik: For a long trailer headers

PR #2326, opened 2025-01-18
Fix accessing already freed memory.

EvgeniiMekhanik:
When Tempesta FW receives HEAD request and cache is enabled Tempesta FW sends GET request to backend
and later deletes body and trailers (if exists) for response. Also Tempesta FW deletes all hop-by-hop
headers. We should skip deleteting hop-by-hop
headers for trailers if they were already deleted
to prevent accessing already freed memory.

### DIFF
--- a/fw/http_msg.c
+++ b/fw/http_msg.c
@@ -868,13 +868,14 @@ __hdr_expand(TfwHttpMsg *hm, TfwStr *orig_hdr, const TfwStr *hdr, bool append)
 	return tfw_strcpy(append ? &it : orig_hdr, hdr);
 }
 
+/* Remove header from header table. */
 static void
 __hdr_del_from_tbl(TfwHttpHdrTbl *ht, unsigned int hid)
 {
 	if (hid < TFW_HTTP_HDR_RAW) {
 		TFW_STR_INIT(&ht->tbl[hid]);
 	} else {
-		if (hid < ht->off - 1)
+		if (hid + 1 < ht->off)
 			memmove(&ht->tbl[hid], &ht->tbl[hid + 1],
 				(ht->off - hid - 1) * sizeof(TfwStr));
 		--ht->off;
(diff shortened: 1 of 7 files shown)


fw/http_msg.c
```diff
@@ -868,6 +868,19 @@ __hdr_expand(TfwHttpMsg *hm, TfwStr *orig_hdr, const TfwStr *hdr, bool append)
 	return tfw_strcpy(append ? &it : orig_hdr, hdr);
 }
 
+static void
+__hdr_del_from_tbl(TfwHttpHdrTbl *ht, unsigned int hid)
+{
+	if (hid < TFW_HTTP_HDR_RAW) {
+		TFW_STR_INIT(&ht->tbl[hid]);
```
const-t, line 875: Seems we can't use this function for regular header removing and for trailer header removing in the same time.
For trailer we must delete only the header that located in the trailer, however this function will remove the header for both sections headers and trailer. The case is duplicated header in trailer and in headers section `TFW_STR_INIT(&ht->tbl[hid]);` will NULLify both of them, but must set to NULL only the header in trailers. We must do `TFW_STR_FOR_EACH_DUP` for such header and remove duplicate only with  `TFW_STR_TRAILER` flag.

For raw headers the same, we must check duplication before memmoving or decrementing `off`.

const-t (changes requested):

const-t (approved):
LGTM

fw/http_msg.c
```diff
@@ -868,6 +868,19 @@ __hdr_expand(TfwHttpMsg *hm, TfwStr *orig_hdr, const TfwStr *hdr, bool append)
...
+	} else {
+		if (hid < ht->off - 1)
```
krizhanovsky, line 877:
```suggestion
		if (hid + 1 < ht->off)
```
Or is it guaranteed here that `ht->off` is never 0? `hid` is always lower than 2^32, so this is safer

fw/http_msg.c
krizhanovsky: @const-t @EvgeniiMekhanik it seems this worth commenting in code?

krizhanovsky (approved):
LGTM

PR #2331, opened 2025-01-24
Fix NULL dereference in `ss_conn_drop_guard_exit`

EvgeniiMekhanik:
There are two cases when `sk_>sk_user_data` could
be equal to zero in `ss_conn_drop_guard_exit`:
- when `ss_do_close` is called for TCP_FIN_WAIT2 tcp_done() is called from tcp_time_wait() and connection is dropped inside `ss_do_close`.
- when `tcp_tfw_sk_write_xmit` which is called from `__sk_close_locked->tcp_send_fin->tcp_write_xmit` failed. In this case `tcp_done` is also called and connection is dropped.
So we should check if `sk_>sk_user_data` is equal to zero and immediately return.

### DIFF
--- a/fw/sock.c
+++ b/fw/sock.c
@@ -189,6 +189,20 @@ ss_active_guard_exit(unsigned long val)
 static void
 ss_conn_drop_guard_exit(struct sock *sk)
 {
+	/*
+	 * There are two cases when `sk_>sk_user_data` could be
+	 * equal to zero:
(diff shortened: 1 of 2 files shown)


const-t (approved):
LGTM

krizhanovsky (approved):
LGTM on the Tempesta side, but there is a question with related https://github.com/tempesta-tech/linux-5.10.35-tfw/pull/35

PR #2332, opened 2025-01-27
Fix use after free.

EvgeniiMekhanik:
We can't use 'req' after `tfw_http_resp_process`
fails, because there are some case when 'req' was
already freed. For example:
tfw_http_resp_process->tfw_http_resp_cache->
tfw_cache_process, `tfw_cache_process` fails and
tfw_http_send_err_resp->tfw_http_send_err_resp_nolog-> tfw_h2_send_err_resp->tfw_h2_send_resp->tfw_h2_prep_resp, ->tfw_h2_prep_resp, `tfw_h2_prep_resp` fails and we clear request in `tfw_http_resp_build_error`.

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -7137,6 +7137,7 @@ tfw_http_msg_process_generic(TfwConn *conn, TfwStream *stream,
 {
 	int r = T_BAD;
 	TfwHttpMsg *req;
+	bool websocket;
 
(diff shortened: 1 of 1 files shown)


const-t (approved):

krizhanovsky (approved):
LGTM

PR #2333, opened 2025-01-27
Mekhanik evgenii/fix 2328 2

EvgeniiMekhanik:


tls/ttls.c
```diff
@@ -2292,7 +2310,10 @@ ttls_recv(void *tls_data, unsigned char *buf, unsigned int len, unsigned int *re
 			BUG_ON(!tls->hs && tls->state != TTLS_HANDSHAKE_OVER);
 			if (ttls_hs_checksumable(tls)) {
 				size_t n = *read - (int)parsed + hh_len;
-				ttls_update_checksum(tls, buf - hh_len, n);
+				r = ttls_update_checksum(tls, buf - hh_len, n);
+				if (unlikely(r))
+					return r;
+				r = T_POSTPONE;
```
const-t, line 2317: Maybe just `return r;` without `if` and assignment?

const-t (approved):
LGTM

tls/mpool.c
```diff
@@ -337,7 +337,7 @@ __mpi_profile_clone(TlsCtx *tls, int ec)
 	}
 
 	ptr = (char *)__get_free_pages(GFP_ATOMIC, __MPOOL_HS_ORDER);
-	if (WARN_ON_ONCE(!ptr))
+	if (!ptr)
```
krizhanovsky, line 340: At least let's make it `unlikely()`. IIRC `BUG` and `WARN` statements imply `unlikely` hints

krizhanovsky (approved):
LGTM

tls/ttls.c
EvgeniiMekhanik: We can't do it. If ttls_update_checksum is successful we should  change 0 to T_POSTPONED

tls/mpool.c
EvgeniiMekhanik: yes fixed

PR #2334, opened 2025-01-30
Fix double free in case of error in `tfw_h2_msg_cutoff_headers`

EvgeniiMekhanik:
Previously we immediately return from `tfw_h2_msg_cutoff_headers` if `ss_skb_linear_transform` fails, without updating `resp->msg.skb_head`. Later in `tfw_h2_resp_encode_headers` we free all skbs from `cleanup` structure in `__tfw_h2_resp_cleanup`. But in this case resp->msg.skb_head points to the already freed skbs (from cleanup structure!). Later when we free response we again free skbs from `resp->msg.skb_head` that leads to double free and BUG. This patch fixes this problem, we update resp->msg.skb_head if error occurs.


const-t (approved):
LGTM

krizhanovsky (approved):
LGTM

PR #2336, opened 2025-02-04
Fix memory leak in `tfw_mmap_buffer_create_dev` test.

EvgeniiMekhanik:
There was a memory leak in unit test `tfw_mmap_buffer_create_dev` During investigation was found that problem in bad deallocation in `__class_register` function from linux kernel code.


const-t (approved):
LGTM. Please add to commit message notice that we must drop this patch during migration.

PR #2338, opened 2025-02-05
Fix WARNING in case of error in `tfw_h2_prep_resp`.

EvgeniiMekhanik:
There can be a kernel warning if `tfw_h2_frame_local_resp` fails and `ctx->cur_send_headers` is set (Tempesta FW sends response with a long headers). In this case `tfw_h2_req_unlink_stream_with_rst` -> `tfw_h2_stream_fsm_ignore_err` fails and WARNING occurs. This patch fix this behaviour:
- First of all we should not process stream state in `tfw_h2_req_unlink_stream_with_rst`, just set stream state to CLOSE state, because this function is called only when we don't send any response to client and just drop request.
- We don't need to call `tfw_h2_req_unlink_stream_with_rst` from `tfw_h2_prep_resp` if it fails because it will be called later from `tfw_h2_send_resp`.
- If `tfw_h2_prep_resp` is called from `tfw_http_prep_redir` and fails, we also don't need to call `tfw_h2_req_unlink_stream_with_rst` because we need to send error response later.

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -678,9 +678,13 @@ tfw_h2_prep_resp(TfwHttpResp *resp, unsigned short status, TfwStr *msg)
 	r = tfw_h2_frame_local_resp(resp, hdrs_len, body);
 
 out:
-	if (r)
-		tfw_h2_req_unlink_stream_with_rst(req);
-	else
+	/*
+	 * In case of error stream will be unlinked later in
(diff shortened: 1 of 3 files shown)


const-t (approved):
LGTM

krizhanovsky (approved):
LGTM

PR #2340, opened 2025-02-10
Fix bug during call `tcp_done`.

EvgeniiMekhanik:
When Tempesta FW closes socket from `ss_do_close`
function, Tempesta FW sends TCP FIN using `tcp_send_fin` or TCP RST using `tcp_send_active_reset`. Both this functions use `tcp_write_xmit` to push FIN/RST to network. If error occurs in `tcp_write_xmit->tcp_tfw_sk_write_xmit` we call `tcp_tfw_handle_error->tcp_done`. In this case `ss_do_close` called recursively from `tcp_done` function, that leads to extra `inet_csk_destroy_sock` call and extra socket put. This patch fixes this behaviour -  Tempesta FW set new special flag at the beginning of `ss_do_close` and if this flag is set we don't call `sk_state_change` / `inet_csk_destroy_sock` from `tcp_done`.

### DIFF
--- a/fw/sock.c
+++ b/fw/sock.c
@@ -2,7 +2,7 @@
  *		Synchronous Socket API.
  *
  * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
- * Copyright (C) 2015-2024 Tempesta Technologies, Inc.
+ * Copyright (C) 2015-2025 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
@@ -1498,7 +1498,10 @@ static inline void
 ss_do_shutdown(struct sock *sk)
 {
 	tcp_shutdown(sk, SEND_SHUTDOWN);
-	SS_CONN_TYPE(sk) |= Conn_Shutdown;
+	if (unlikely(sk->sk_state == TCP_CLOSE))
+		ss_linkerror(sk, 0);
+	else
+		SS_CONN_TYPE(sk) |= Conn_Shutdown;
 }
 
 static inline bool
--- a/linux-5.10.35.patch
+++ b/linux-5.10.35.patch
@@ -503,51 +503,6 @@ index b4dae640d..1b6d4a669 100644
  struct crypto_sync_skcipher *crypto_alloc_sync_skcipher(
  				const char *alg_name, u32 type, u32 mask)
  {
-diff --git a/drivers/base/class.c b/drivers/base/class.c
-index c34514811..aa551e4de 100644
---- a/drivers/base/class.c
-+++ b/drivers/base/class.c
-@@ -187,12 +187,40 @@ int __class_register(struct class *cls, struct lock_class_key *key)
- 
- 	error = kset_register(&cp->subsys);
- 	if (error) {
-+	/*
-+	 * Free memory, which was allocated in `kobject_set_name`
-+	 * This patch should be dropped during porting on
-+	 * linux kernel 6.12.12.
-+	 */
-+#ifdef CONFIG_SECURITY_TEMPESTA
-+		kfree_const(cp->subsys.kobj.name);
-+		goto err_out;
-+#else
- 		kfree(cp);
- 		return error;
-+#endif
- 	}
- 	error = class_add_groups(class_get(cls), cls->class_groups);
- 	class_put(cls);
-+	/*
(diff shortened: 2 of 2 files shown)


fw/sock.c
```diff
@@ -645,6 +645,14 @@ ss_do_close(struct sock *sk, int flags)
 	/* The below is mostly copy-paste from tcp_close(), 5.10.35. */
 	sk->sk_shutdown = SHUTDOWN_MASK;
 
+	/*
+	 * We set this flag at the beginning of this function and reset it at
+	 * the end of this function to prevent recursive call of this function,
+	 * when `tcp_done` is called because of error in `tcp_send_fin` /
+	 * `tcp_send_active_reset`->  `tcp_write_xmit->tcp_tfw_handle_error`.
+	 */
+	sock_set_flag(sk, SOCK_TIMESTA_CLOSING);
```
const-t, line 654: I have proposal for this patch.

First of all I can't see cases how `tcp_write_xmit()` could be called from `tcp_send_active_reset()`, the last avoids using of `write_queue` and calls  `tcp_transmit_skb()` directly. Therefore `tcp_tfw_handle_error()` can't be called from `tcp_send_active_reset()`.

The second. Looking into `ss_do_close()` I see single call of `tcp_send_fin()` [here](https://github.com/tempesta-tech/tempesta/blob/cc8e43c6f38642401e55c6703b21006357475233/fw/sock.c#L672), and only inside this function `ss_do_close()` might be called recursively. Let's breakdown this call chain:
```
(#1) ss_do_close()
          \/ 
tcp_send_fin() #Before the call set state to TCP_FIN_WAIT1
          \/
tcp_write_xmit() #Try to send skb with FIN.
          \/
handle_error() #Error occurred during sending FIN. 
          \/
tcp_done() #SET state to TCP_CLOSE.
          \/
(#2) ss_do_close() -> goto adjudge_to_death -> set socket as DEAD -> call inet_csk_destroy_sock() the first time.
          \/
return from tcp_done() to ss_do_close()
          \/
(#1) ss_do_close() -> fall to adjudge_to_death -> set socket as DEAD again -> call inet_csk_destroy_sock() second time, because state is TCP_CLOSE.
``` 
As we can see, we should not call `tcp_done()` in case of error during sending FIN, maybe instead of adding new flag we can rely on socket's state and do something like this:
```
tcp_tfw_handle_error(struct sock *sk, int error)
{
	tcp_send_active_reset(sk, GFP_ATOMIC);
	sk->sk_err = error;
	sk->sk_error_report(sk);
	tcp_write_queue_purge(sk);
	if (sk->sk_state == TCP_FIN_WAIT1)
		tcp_set_state(sk, TCP_CLOSE);
	else
		tcp_done(sk);
}
```

EvgeniiMekhanik: it seems that we can't rely on the state here. For example when we call tcp_shutdown->tcp_close_state->tcp_send_fin and if tcp_send_fin fails we should call tcp_done

const-t: Maybe we can check the state in `tcp_shutdown` as well?
```
static inline void
ss_do_shutdown(struct sock *sk)
{
	tcp_shutdown(sk, SEND_SHUTDOWN);
	SS_CONN_TYPE(sk) |= Conn_Shutdown;
       if (sk->state == TCP_CLOSE)
           ss_do_close(sk, 0);
}
```

krizhanovsky: Yeah, it'd be wonderful if @const-t 's proposal can be implemented: it's not so beautiful to introduce a new flag in the kernel patch just for one condition.

fw/sock.c
```diff
@@ -1498,7 +1498,11 @@ static inline void
 ss_do_shutdown(struct sock *sk)
 {
 	tcp_shutdown(sk, SEND_SHUTDOWN);
-	SS_CONN_TYPE(sk) |= Conn_Shutdown;
+	if (unlikely(sk->sk_state == TCP_CLOSE)) {
```
const-t, line 1501: Extra braces

const-t (approved):
LGTM

fw/sock.c
```diff
@@ -1498,7 +1498,11 @@ static inline void
 ss_do_shutdown(struct sock *sk)
```
krizhanovsky, line 1498: Please update the copy right year - check you .git configuration, it should be done in automatic way on commit

krizhanovsky (approved):
LGTM

Issue #2322, opened 2025-01-13
TempestaFW blocks GET/HEAD requests with `Content-Type` or `Content-Length` headers


RomanBelozerov:
# Scope

I receive `Warning` in dmesg using TempestaFW as proxy for Jenkins with Blue Ocean plugin:

> Warning: Content-Length or Content-Type not allowed to be used with such (overridden) method

A lot of apps can use these headers by default and TemepstaFW will be block them. This can be a problem.

 My proposal:  

- remove the check or;
- add a new directive for disabling the check (global/vhost/location);

# Testing

- [x] add new tests or create a new issue in tempesta-test (probably we need to update DeproxyAutoParser)

# Documentation

- [x] update wiki

krizhanovsky:
Required for S3 compatibility. We need a configuration option to switch on/off this feature

PR #2341, opened 2025-02-11
Content-Lenth and Content-Type in bodyless message

const-t:
RFC 9110 8.6:
A server MUST NOT send a Content-Length header field in any response with a status code of 1xx (Informational) or 204 (No Content).

Now for responses 1xx and 204 Tempesta FW treats `Content-Length: 0` as the absence of a Content-Length header. Some implementations send `Content-Length: 0` within 204 (No Content) response, to be able to process such messages the rule from RFC 9110 8.6 has been relaxed.

For requests with bodyless methods such as HEAD, GET, etc. Tempesta also treats `Content-Length: 0` as empty body and considers such requests as valid.

Added directive `http_allow_empty_body_content_type` that allows Tempesta FW to process requests with bodyless methods. By default Tempesta FW drops such requests.

### DIFF
--- a/etc/tempesta_fw.conf
+++ b/etc/tempesta_fw.conf
@@ -1189,6 +1189,18 @@
 #   http_max_header_list_size 4096;
 #
 
+# TAG: http_allow_empty_body_content_type
+#
+# This directive controls whether HTTP requests are allowed to include
+# a Content-Type header even when the message body is empty.
+#
+# Syntax:
+#   http_allow_empty_body_content_type true|false
+#
+# Example:
+#   http_allow_empty_body_content_type true;
+#
+
 #
 # Health monitoring configuration.
 #
--- a/fw/http_parser.c
+++ b/fw/http_parser.c
@@ -1969,18 +1969,6 @@ __parse_content_length(TfwHttpMsg *hm, unsigned char *data, size_t len)
 
 	__FSM_STATE(I_ContLen) {
 		/*
-		 * A server MUST NOT send a Content-Length header field in any response
-		 * with a status code of 1xx (Informational) or 204 (No Content).
-		 * TODO: server MUST NOT send a Content-Length header field in any 2xx
-		 * (Successful) response to a CONNECT request
-		 */
-		if (TFW_CONN_TYPE(msg->conn) & Conn_Srv) {
-			TfwHttpResp *resp = (TfwHttpResp *)msg;
-			if (resp->status - 100U < 100U || resp->status == 204)
-				return CSTR_NEQ;
-		}
-
-		/*
 		 * According to RFC 7230 3.3.2, in cases of multiple Content-Length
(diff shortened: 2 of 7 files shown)


etc/tempesta_fw.conf
```diff
@@ -1189,6 +1189,18 @@
...
+# TAG: allow_empty_body_content_type
+#
+# This directive controls whether HTTP requests are allowed to include
+# a Content-Type header even when the message body is empty.
+#
+# Syntax:
+#   allow_empty_body_content_type true|false
+#
+# Example:
+#   allow_empty_body_content_type true;
```
symstu-tempesta, line 1201: looks like you have defined `http_allow_empty_body_content_type` for the config variable

const-t: Yes, it's typo. Thank you

etc/tempesta_fw.conf
```diff
@@ -1189,6 +1189,18 @@
...
+#
+# This directive controls whether HTTP requests are allowed to include
+# a Content-Type header even when the message body is empty.
+#
+# Syntax:
+#   http_allow_empty_body_content_type true|false
+#
+# Example:
+#   http_allow_empty_body_content_type true;
+#
```
krizhanovsky, line 1202: Could you please also describe this in wiki? Also with good story from https://github.com/tempesta-tech/tempesta/issues/2322 and https://github.com/tempesta-tech/tempesta/pull/2341#issue-2845726528 : make examples for the implementations, cite the RFC and how to we violate it (e.g. with `Content-Length: 0`).

fw/http_parser.c
```diff
@@ -1996,6 +1984,25 @@ __parse_content_length(TfwHttpMsg *hm, unsigned char *data, size_t len)
...
+		 * TODO: server MUST NOT send a Content-Length header field in any 2xx
+		 * (Successful) response to a CONNECT request
+		 *
+		 * Treat `Content-Length: 0` as the absence of a Content-Length
+		 * header. Some implementations send `Content-Length: 0` within
+		 * 204 (No Content) or 1xx responses, to be able to process such
+		 * messages the rule from RFC has been relaxed.
+		 */
+		if (TFW_CONN_TYPE(msg->conn) & Conn_Srv
+		    && msg->content_length > 0) {
```
krizhanovsky, line 1999:
```suggestion
		    && msg->content_length > 0)
	   {
```

krizhanovsky (approved):
LGTM

etc/tempesta_fw.conf
const-t: DONE: https://github.com/tempesta-tech/tempesta/wiki/HTTP-security#relaxed-content-type-and-content-length-rules

symstu-tempesta (approved):

PR #2342, opened 2025-02-11
Get rid of per-cpu buffer `g_te_buf`

const-t:
Per-cpu buffer `g_te_buf` replaced with buffer allocated from TfwPool. Using per-cpu buffer as paged data for skb isn't good idea.


krizhanovsky (approved):
LGTM

PR #2344, opened 2025-02-26
Remove extra `module_put`.

EvgeniiMekhanik:
There was an extra `module_put` in case when
`ss_inet_create` fails, remove it.


const-t (approved):
LGTM

PR #2346, opened 2025-03-03
Fix potencial BUG when Tempesta shutdowned

EvgeniiMekhanik:
There was a little optimization (we use `list_empty_careful` before `spin_lock`) in `tfw_http_conn_release`, when Tempesta shutdowned (ss_active is false). This optiomization can lead to double free:
When Tempesta FW shutdowned tfw_mods_stop->ss_stop is called, immediately after it (before clients connections will be dropped) connection with backend is closed by backend and `tfw_http_conn_release` is called. `list_empty_careful` return true and we don't remove this request from seq_list, but on other CPU request can be added! This leads to double free. In fact this optimization doesn't make sense at all, since it only works when tempesta is turned off! Let's use a secure function with a lock here!

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -2895,16 +2895,7 @@ tfw_http_conn_release(TfwConn *conn)
 	 */
 	list_for_each_entry_safe(req, tmp, &zap_queue, fwd_list) {
 		list_del_init(&req->fwd_list);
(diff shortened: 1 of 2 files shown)


krizhanovsky (approved):
LGTM

const-t (approved):
LGTM

PR #2351, opened 2025-03-07
Fix deleteting previosly closed streams.

EvgeniiMekhanik:
Previously we delete previously closed streams in
case when count of closed streams is greater then
TFW_MAX_CLOSED_STREAMS. But there is a case when
count of streams is equal to max concurrent streams when we should close all closed streams to be able to create new streams.

### DIFF
--- a/fw/http2.c
+++ b/fw/http2.c
@@ -1,7 +1,7 @@
 /**
  *		Tempesta FW
  *
- * Copyright (C) 2024 Tempesta Technologies, Inc.
+ * Copyright (C) 2024-2025 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


fw/http2.c
```diff
@@ -368,7 +369,9 @@ tfw_h2_closed_streams_shrink(TfwH2Ctx *ctx)
 	while (1) {
 		spin_lock(&ctx->lock);
 
-		if (closed_streams->num <= TFW_MAX_CLOSED_STREAMS) {
+		if (closed_streams->num <= TFW_MAX_CLOSED_STREAMS
+		    && (ctx->streams_num < max_streams
+			|| !closed_streams->num)) {
```
const-t, line 374: I'm ok with current solution, however I would like to ask, maybe we need account `ctx->streams_num` excluding `closed_streams->num`?

const-t (approved):

fw/http2.c
EvgeniiMekhanik: No! Streams moved to closed_streams queue after sending all data to the client or when RST stream is received. If we don't take them into account we never delete this streams from memory

Issue #2278, opened 2024-11-08
Check that Tempesta FW config exist during Tempesta loading


EvgeniiMekhanik:
Currently if Tempesta FW config doesn't exist we load Tempesta with strange error messages.

It seems that we should check config presence in temepsta.sh

PR #2355, opened 2025-03-12
Fix: check config file existence

MorganaFuture:
**What**
Add a check for the existence of the configuration file before starting Tempesta FW.

**Why**
Currently when Tempesta FW is started without a configuration file, it produces confusing error messages instead of clearly indicating the root cause of the issue.

**Links**
[2278](https://github.com/tempesta-tech/tempesta/issues/2278)

### DIFF
--- a/scripts/tempesta.sh
+++ b/scripts/tempesta.sh
@@ -344,6 +344,13 @@ stop_tfw_logger()
 	fi
 }
 
+check_configuration_file_presense()
+{
+	if [ ! -f "$tfw_cfg_path" ]; then
+		error "Configuration file $tfw_cfg_path does not exist. Please create a configuration file before starting Tempesta FW."
+	fi
+}
+
 start()
 {
 	echo "Starting Tempesta..."
@@ -351,6 +358,8 @@ start()
 	TFW_STATE=$(sysctl net.tempesta.state 2> /dev/null)
 	TFW_STATE=${TFW_STATE##* }
 	TFW_LOGGER_EXEC=$(expr "$TFW_STATE" != "start")
+	
+	check_configuration_file_presense
 
 	if [[ -z ${TFW_STATE} ]]; then
 		setup
@@ -397,6 +406,7 @@ stop()
 
 reload()
 {
+	check_configuration_file_presense
 	update_js_challenge_templates
 	echo "Running live reconfiguration of Tempesta..."
 


krizhanovsky:
@MorganaFuture thank you for the pull request! @EvgeniiMekhanik could you please review it

scripts/tempesta.sh
```diff
@@ -351,6 +351,10 @@ start()
 	TFW_STATE=$(sysctl net.tempesta.state 2> /dev/null)
 	TFW_STATE=${TFW_STATE##* }
 	TFW_LOGGER_EXEC=$(expr "$TFW_STATE" != "start")
+	
+	if [ ! -f "$tfw_cfg_path" ]; then
+		error "Configuration file $tfw_cfg_path does not exist. Please create a configuration file before starting Tempesta FW."
```
EvgeniiMekhanik, line 356: It seems that we should check configuration file presense for reload/restart also. For example in current implementation if we start Tempesta FW then remove configuration file and try to reload we have the same problem, which is described in issue. So I suggest to move this check in separate function with the name `check_configuration_file_presense()` and call it for reload/restart also

EvgeniiMekhanik (changes requested):
It seems that we should check configuration file presense for reload/restart also. For example in current implementation if we start Tempesta FW then remove configuration file and try to reload we have the same problem, which is described in issue. So I suggest to move this check in separate function with the name check_configuration_file_presense() and call it for reload/restart also

scripts/tempesta.sh
MorganaFuture: Added the check_configuration_file_presense function and included it in reload as well.

scripts/tempesta.sh
```diff
@@ -344,13 +344,22 @@ stop_tfw_logger()
...
+
 start()
 {
 	echo "Starting Tempesta..."
 
 	TFW_STATE=$(sysctl net.tempesta.state 2> /dev/null)
 	TFW_STATE=${TFW_STATE##* }
 	TFW_LOGGER_EXEC=$(expr "$TFW_STATE" != "start")
+	
+	check_configuration_file_presense()
```
EvgeniiMekhanik, line 362: in bash we should not use `()` when we call function it is a syntax error (all other is good). And please merge this two commits in one.

MorganaFuture: Fixed

EvgeniiMekhanik (approved):

PR #2357, opened 2025-03-12
Several `skb_fragment` and `ss_skb_data_ptr_by_offset`

EvgeniiMekhanik:
- We should check that skb has linear part (d_size != 0) in `skb_fragment` because if it is not true, skb->data can points to incorrect place (for example address of skb fragment!) if this skb was previosly passed to `__split_linear_data`.
- Fix calculation of place to insert frame header in `ss_skb_data_ptr_by_offset`. If data points to the end of (linear data/fragment) we should use next (zero fragment/next fragment) not current one.


const-t (approved):
LGTM

Issue #2258, opened 2024-10-09
Fix `irqbalance` in scripts


krizhanovsky:
Ubuntu 24 doesn't support `irqbalance` anymore by default, so on Tempesta stop we have a warning
```
# ./scripts/tempesta.sh --stop
Stopping Tempesta...
net.tempesta.state = stop
...unload Tempesta modules
Un-loading Tempesta kernel modules...
Unit irqbalance.service could not be found.
done
```

We should remove it from our our control scripts. Probably just remove since it seems the IRQ binding we do through sysfs.

Or if it's still installed, we just need to correctly handle this case and the case, when it is not installed

PR #2360, opened 2025-03-14
Fix: irqbalance warning

MorganaFuture:
**What**
Added systemd service availability check before attempting to modify and restart the irqbalance service during Tempesta FW shutdown.

Current behavior:

- If irqbalance is installed and running, the function works as before
- If irqbalance isn't installed or running, the function silently does nothing
- The script continues execution without errors either way

**Why**
Ubuntu 24.04 doesn't include irqbalance by default, causing error messages during Tempesta FW shutdown:
`Unit irqbalance.service could not be found.`

**Links**
[2258](https://github.com/tempesta-tech/tempesta/issues/2258)

### DIFF
--- a/scripts/tfw_lib.sh
+++ b/scripts/tfw_lib.sh
@@ -292,13 +292,12 @@ tfw_set_net_queues()
 
 tfw_irqbalance_revert()
 {
-	systemctl status irqbalance.service >/dev/null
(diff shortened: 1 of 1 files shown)


EvgeniiMekhanik (approved):

EvgeniiMekhanik (approved):
LGTM. Please move description to commit message

krizhanovsky:
Merged, @MorganaFuture thank you!

Issue #2313, opened 2025-01-06
Configuration parsing error: mmap_log_buffer_size


symstu-tempesta:
**Describe the issue**
I cannot set the configuration value for mmap_log_buffer_size.

**Expected Behavior**
Tempesta FW starts correctly but currently crashes when the mmap_log_buffer_size configuration parameter is set.

**To Reproduce**
Add the following line to your Tempesta configuration and start the application:
```
mmap_log_buffer_size 4096;
```

**Configuration file**
```config
            listen 80;
            server localhost:8000;

            access_log dmesg mmap mmap_host=localhost mmap_log=/tmp/access.log;
            mmap_log_buffer_size 4096;
```

**Version or commit hash**
80598c97099132d383426ae45dd45f8bf1bd675d

**Stacktrace or debug log**
```
[23215.774224] Start test: t_clickhouse.test_clickhouse_logs.TestClickhouseLogsBaseTest2.test_missing_log
[23215.933958] [tdb] Start Tempesta DB
[23215.952773] [tempesta fw] Initializing Tempesta FW kernel module...
[23216.038606] [tempesta fw] ERROR: the first character is not a letter: '4'
[23216.038609] [tempesta fw] ERROR: syntax error
[23216.038611] [tempesta fw] ERROR: configuration parsing error:
                  5:             4096;
                     ^^^^^^^^^^^^^^^^^
[23216.038613] [tempesta fw] Warning: Configuration parsing has failed. Clean up...
[23216.086674] [tempesta fw] exiting...
[23216.180309] [tdb] Shutdown Tempesta DB
```

krizhanovsky:
Please create a new wiki page for `tfw_logger` and document `mmap_log_buffer_size` there (we don't have anything about this option in our wiki). The mmap part must be moved from https://tempesta-tech.com/knowledge-base/Handling-clients/  to the new wiki  page.

Now `access_log` must just accept two options, which can be used together: `tfw_logger` and `dmesg`. E.g.
```
access_log dmesg tfw_logger=/path/to/config.json
```

See https://github.com/tempesta-tech/tempesta/issues/2421 for the configuration changes: make the transition to the separate JSON configuration and leave the ML configuration part for #2421.

PR #2363, opened 2025-03-16
fix: configuration parsing error mmap_log_buffer_size

MorganaFuture:
**What**
This PR introduces a JSON-based configuration system for the Tempesta FW logger daemon with:

- A dedicated TfwLoggerConfig class for logger settings
- JSON configuration file support via Boost's property_tree
- Two operation modes: daemon (background) and handle (foreground for debugging)
- Support for a logger_config attribute in the access_log directive
- Updates to Tempesta scripts for compatibility
- Backward compatibility with existing configuration

**Why**
The Tempesta logger configuration is currently scattered across multiple parameters, making it hard to manage. A dedicated JSON file provides a cleaner approach that separates logger settings from the main configuration. This structured format improves error handling and simplifies troubleshooting with the new foreground mode.

**Links**
[2313](https://github.com/tempesta-tech/tempesta/issues/2313)

### DIFF
--- a/utils/tfw_logger.cc
+++ b/utils/tfw_logger.cc
@@ -39,21 +39,28 @@
 #include "clickhouse.hh"
 #include "mmap_buffer.hh"
 #include "error.hh"
+#include "tfw_logger_config.hh"
 
 namespace po = boost::program_options;
 
-constexpr char dev_path[]	= "/dev/tempesta_mmap_log";
-constexpr char table_name[]	= "access_log";
-constexpr char pid_file_path[]	= "/var/run/tfw_logger.pid";
-constexpr std::chrono::seconds		wait_for_dev{1};
-constexpr std::chrono::milliseconds	wait_for_stop(10);
-constexpr std::chrono::seconds		reconnect_min_timeout{1};
-constexpr std::chrono::seconds		reconnect_max_timeout{16};
+constexpr char dev_path[] = "/dev/tempesta_mmap_log";
+constexpr std::chrono::seconds wait_for_dev{1};
+constexpr std::chrono::milliseconds wait_for_stop(10);
+constexpr std::chrono::seconds reconnect_min_timeout{1};
+constexpr std::chrono::seconds reconnect_max_timeout{16};
 
 typedef struct {
 	const char		*name;
 	clickhouse::Type::Code	code;
 } TfwField;
+
+enum class CommandResult {
+    CONTINUE,
+    HELP,
+    GENERATE,
+    STOP,
+    ERROR
+};
 
 static const TfwField tfw_fields[] = {
 	[TFW_MMAP_LOG_ADDR]		= {"address", clickhouse::Type::IPv6},
@@ -73,6 +80,7 @@ static const TfwField tfw_fields[] = {
 
 std::atomic<bool> stop_flag{false};
 static thread_local bool uncritical_error;
+static TfwLoggerConfig config;
 
 #ifdef DEBUG
 static void
@@ -232,9 +240,8 @@ callback(const char *data, int size, void *private_data)
 		uncritical_error = false;
 }
 
-void
-run_thread(const int ncpu, const int fd, const std::string &host,
-	   const std::string &user, const std::string &password)
+void 
+run_thread(const int ncpu, const int fd, const TfwLoggerConfig &config)
 {
 	static thread_local std::chrono::seconds timeout(reconnect_min_timeout);
 	cpu_set_t cpuset;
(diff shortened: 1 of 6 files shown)


MorganaFuture:
I have been starting work on [2313](https://github.com/tempesta-tech/tempesta/issues/2313).

MorganaFuture:
@krizhanovsky I slightly modified tfw_logger to make it work with JSON files. I also did some refactoring on it.  Could you assign someone to review these changes?

const-t (changes requested):

utils/tfw_logger.cc
```diff
@@ -333,136 +357,269 @@ void stop_daemon()
...
+    auto loaded_config = TfwLoggerConfig::load_from_file(config_path);
+    if (loaded_config)
+    {
+         config = *loaded_config;
+         std::cout << "Configuration loaded from: " << config_path << std::endl;
+    }
+    else
+    {
+     	// If config file not found or invalid, use command line args
+        std::cout << "Could not load configuration from file, using command line arguments" << std::endl;
```
Kutumov, line 455: In my opinion if expected configuration fails(or its part) it is better to exit in order not to execute the service with undesirable settings. @const-t FYI

utils/tfw_logger_config.cc
```diff
@@ -0,0 +1,257 @@
...
+std::optional<TfwLoggerConfig>
+TfwLoggerConfig::load_from_file(const fs::path &path)
+{
+    if (!fs::exists(path))
+    {
+        std::cerr << "Config file not found: " << path << std::endl;
+        return std::nullopt;
+    }
+
+    try
```
Kutumov, line 42: Please reduce the try area to the line that may throw

utils/tfw_logger_config.cc
```diff
@@ -0,0 +1,257 @@
...
+    {
+        pt::ptree tree;
+        pt::read_json(path.string(), tree);
+
+        TfwLoggerConfig config;
+        config.parse_from_ptree(tree);
+        return config;
```
Kutumov, line 49: Why not `std::move(config)`?

utils/tfw_logger.cc
```diff
@@ -333,136 +357,269 @@ void stop_daemon()
...
+{
+	size_t cpu_cnt;
+ 
+    if (config.get_cpu_count() > 0)
+    {
+        cpu_cnt = config.get_cpu_count();
+    }
+    else
+    {
+        cpu_cnt = std::thread::hardware_concurrency();
```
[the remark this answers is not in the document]
Kutumov: @const-t What's wrong with `std::thread::hardware_concurrency`?

utils/tfw_logger.cc
```diff
@@ -333,136 +357,269 @@ void stop_daemon()
...
+
+void 
+run_worker_threads(int fd, size_t cpu_count)
+{
+	std::vector<std::thread> threads;
+	std::vector<std::future<void>> futures;
+
+	for (size_t i = 0; i < cpu_count; ++i)
+	{
+		std::packaged_task<void(int, int, TfwLoggerConfig)> task(run_thread);
```
Kutumov, line 575: `packaged_task` is redundant here. We can just wait when threads finish using `std::thread::join`

utils/tfw_logger.cc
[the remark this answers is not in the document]
const-t: @Kutumov I'm not sure that it works well with NUMA and with machines that has 100+ cpus. It may return invalid cpu count, but not zero and in this case we even can't fallback to `sysconf`. Even cppref describes it as:
`The value should be considered only a hint`.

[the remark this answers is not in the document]
krizhanovsky: This must be commented in the source code - I was also thinking that `std::thread::hardware_concurrency` is good

utils/tfw_logger.cc
```diff
@@ -333,136 +357,269 @@ void stop_daemon()
...
 			throw Except("Daemonization failed");
+ 
+		// Re-initialize logger after daemonizing
+		setup_logger(config);
+	}
+	else
+	{
+		spdlog::info("Starting in handle (foreground) mode");
+	}
+}
```
krizhanovsky, line 529: I think we don't need daemonization. The modern Linux environments provide this out of the box and in the modern our projects we don't daemonize servers

utils/tfw_logger.cc
```diff
@@ -333,136 +357,269 @@ void stop_daemon()
...
-			std::this_thread::sleep_for(wait_for_dev);
+void 
+create_pid_file(const fs::path &pid_file_path)
+{
+	std::ofstream pid_file(pid_file_path);
+	if (!pid_file)
+		throw Except("Failed to open PID file");
+	pid_file << getpid();
+	pid_file.close();
+}
```
krizhanovsky, line 539: what's the purpose of the pid file? Typically in Linux we also should lock it and set the right permissions. @ttaym you recently worked with a nice small library for PID files, could you please reference it  that we can borrow it's code here as well?

utils/tfw_logger.cc
```diff
@@ -333,136 +357,269 @@ void stop_daemon()
...
-			future.get();
+		size_t cpu_cnt = determine_cpu_count(config);
+
+		daemonize_if_needed(config);
+		set_sig_handlers();
+		create_pid_file(config.get_pid_file());
 
-		spdlog::info("Daemon stopped");
+		// Open the device
+		fd = open_device();
```
krizhanovsky, line 616: Probably the comment is redundant - basically the function name is just the same as the comment

utils/tfw_logger_config.hh
```diff
@@ -0,0 +1,96 @@
...
+  const ClickHouseConfig &get_clickhouse() const { return clickhouse; }
+  bool get_debug() const { return debug; }
+
+private:
+  Mode mode{Mode::DAEMON};                               // Default to daemon mode
+  fs::path log_path{"/var/log/tempesta/tfw_logger.log"}; // Log file path
+  fs::path pid_file{"/var/run/tfw_logger.pid"};          // PID file path
+  size_t buffer_size{4 * 1024 * 1024};                   // 4MB default buffer size
+  size_t cpu_count{0};                                   // 0 means auto-detect
+  ClickHouseConfig clickhouse;                           // ClickHouse configuration
```
krizhanovsky, line 92: Issues with identation and usually we name members as `clickhouse_`, with `_` suffix

MorganaFuture:
The pull request is being moved here https://github.com/tempesta-tech/tempesta/pull/2428

PR #2364, opened 2025-03-17
update `tempesta_installer.sh` (release-0.8)

RomanBelozerov:


pkg/scripts/tempesta_installer.sh
```diff
@@ -68,43 +68,16 @@ log() {
     echo "${TIMESTAMP} [${LEVEL}] - ${MESSAGE}" | tee -a $LOGFILE
```
const-t, line 68: we can delete: `declare -r GITHUB_USER="tempesta-tech"` and `declare -r GITHUB_REPO_TEMPESTA="tempesta"`.

const-t (approved):

pkg/scripts/tempesta_installer.sh
RomanBelozerov: Done

PR #2365, opened 2025-03-18
Mekhanik evgenii/1411 3

EvgeniiMekhanik:


fw/cache.c
```diff
@@ -3308,6 +3271,19 @@ cache_try_service_request(TfwHttpReq *req, tfw_http_cache_cb_t action)
 static void
 cache_req_process_node(TfwHttpReq *req, tfw_http_cache_cb_t action)
 {
+	/*
+	 * If the stream for HTTP/2-request is already closed (due to some
+	 * error or just reset from the client side), there is no sense to
+	 * forward response to the client or request to the server.
+	 * Stream can be closed here for example if cache mode is
+	 * TFW_CACHE_SHARD, this function is called on other numa node
```
const-t, line 3279: In case `TFW_CACHE_SHARD` we may do service request from the cache on remote node, it means that connection may reside on cpu0, but we service request on cpu1. Therefore, while we do service request on cpu1, on cpu0 connection may be already closed. Thus we may have a race  inside `cache_try_service_request() -> tfw_h2_req_unlink_stream_nolock()` isn't?

fw/http2.c
```diff
@@ -412,45 +406,29 @@ tfw_h2_find_not_closed_stream(TfwH2Ctx *ctx, unsigned int id, bool recv)
...
 
 	stream = req->stream;
-	if (!stream) {
-		spin_unlock(&ctx->lock);
+	if (!stream)
 		return;
-	}
 
 	req->stream = NULL;
 	stream->msg = NULL;
```
const-t, line 422: Another subtle place that I would like to discuss with you, it's not related to current PR. 

Here we unlink request from the stream, however this function may run on another cpu due to `CACHE_SHARD` mode, but in [this place](https://github.com/tempesta-tech/tempesta/blob/52d544b53c0f396278f35d622f060eb753d76e27/fw/http.c#L6742) we update `stream->msg` is pipelined request is received. 
The case:
**cpu0**: tfw_cache_process() -> schedule task to another cpu -> exit from tfw_cache_process() -> update `stream->msg`.
**cpu1**: tfw_cache_do_action() -> access already updated stream, but work scheduled with another stream. Therefore, race due to shared `stream`.
Could you check this as well please.

fw/cache.c
EvgeniiMekhanik: We should discuss about it. It seems that we can have a lot of problems with TFW_CACHE_SHARD, because we don't use socket lock

fw/http2.c
EvgeniiMekhanik: Yes agree. We have a problem with a cache shard, so I convert this PR to draft

const-t (approved):

krizhanovsky (commented):
It seems there is a misconception about cache modes:
* `shard` - means that the whole cache is split among NUMA node, i.e. all nodes have different data sets
* `replicated` means that all the NUMA nodes have the identical data sets, but data sets are just copies of each other

I.e. this is about processor packages, not per-CPU things. So I didn't get why the commit messages say "concurrent access' - one or many NUMA nodes do have concurrency and all the cache modes always have concurrency.

I'm not sure if this impacts the patch correctness since checking `req->stream` was needed to not to try to send to closed stream. Can't we close a stream on Tempesta side from another CPU?

EvgeniiMekhanik:
> It seems there is a misconception about cache modes:
> 
>     * `shard` - means that the whole cache is split among NUMA node, i.e. all nodes have different data sets
> 
>     * `replicated` means that all the NUMA nodes have the identical data sets, but data sets are just copies of each other
> 
> 
> I.e. this is about processor packages, not per-CPU things. So I didn't get why the commit messages say "concurrent access' - one or many NUMA nodes do have concurrency and all the cache modes always have concurrency.
> 
> I'm not sure if this impacts the patch correctness since checking `req->stream` was needed to not to try to send to closed stream. Can't we close a stream on Tempesta side from another CPU?

When mode is TFW_CACHE_SHARD we calculate `req->node` and use `tfw_wq_push` to start work in the future on the other cpu (if req->node != numa_node_id()). If mode is not TFW_CACHE_SHARD we call `tfw_cache_do_action` immediately. 
-  If we call `tfw_cache_do_action` immediately (mode is NOT TFW_CACHE_SHARD) there is no chance that stream/connection will be closed on other cpu , because we are under the socket lock on the receive path.
- If we call  `tfw_cache_do_action` stream/connection can be closed on other cpu, but it can be done in any time!

krizhanovsky (approved):
Thank you for the explanation, LGTM!

PR #2368, opened 2025-03-19
Remove unnecessary `tfw_http_hdr_split`

EvgeniiMekhanik:
It seems that we already call `tfw_http_msg_srvhdr_val` two lines below and don't need to call `tfw_http_hdr_split`.


const-t (approved):

krizhanovsky (approved):
LGTM

Issue #2366, opened 2025-03-18
Per vhost cache


const-t:
Cache key consist of `uri + host header`, however may be the case when `http_chains` is configured to route request to other vhost regardless host header. Therefore, after routing Tempesta will try to find/build cache record using the key that belongs to another vhost(specified in host header).

Example: `curl -H"Connection: close" http://app1`

```
vhost app1 {
...
}

vhost app2 {
...
}

http_chains {
hdr "Connection" == "keep-alive" > app1;
-> app2;
}
```
In this case, although we requesting `app1` request will be forwarded to `app2`, but cache record will be stored using key `/ app1`, however expected key is `/ app2`. 

To fix this issue we must take `uri + vhost name` to build the cache key instead of `uri + host header`.

# Testing
Functional test must be implemented.

krizhanovsky:
done in #2435

PR #2376, opened 2025-03-25
fix: Use vhost name instead of Host header in cache key calculation

MorganaFuture:
**What**
Use vhost name instead of Host header in cache key calculation

**Why**
This change ensures proper cache operation when requests are redirected to different vhosts through HTTP chains.
Previously, the cache key was built using uri_path + host (from Host header), which caused incorrect cache behavior when a request with a Host header for one vhost was redirected to a different vhost via HTTP chains. This would lead to storing the cached response under a key that combined the URI with the original Host header, rather than the actual vhost that served the content.
With this change, the cache key is now correctly built using uri_path + vhost_name, ensuring that the cache operates properly even when HTTP chains redirect requests between different vhosts. The code falls back to using the Host header when a vhost is not yet assigned.

**Links**
[2366](https://github.com/tempesta-tech/tempesta/issues/2366)

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -7575,7 +7575,8 @@ cleanup:
 }
 
 /**
- * Calculate the key of an HTTP request by hashing URI and Host header values.
+ * Calculate the key of an HTTP request by hashing URI and vhost name.
+ * If vhost is not yet assigned, fall back to using the Host header.
  */
 unsigned long
 tfw_http_req_key_calc(TfwHttpReq *req)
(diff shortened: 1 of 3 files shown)


fw/t/unit/test.c
```diff
@@ -90,6 +90,7 @@ TEST_SUITE(cfg);
 TEST_SUITE(tfw_str);
 TEST_SUITE(mem_fast);
 TEST_SUITE(http2_parser_hpack);
+TEST_SUITE(http_cache_key);
```
const-t, line 93: Let's rename this to have common suite for all cache functions. 
```suggestion
TEST_SUITE(http_cache);
```

fw/http.c
```diff
@@ -7575,7 +7575,8 @@ tfw_http_hm_srv_send(TfwServer *srv, char *data, unsigned long len)
 }
 
 /**
- * Calculate the key of an HTTP request by hashing URI and Host header values.
+ * Calculate the key of an HTTP request by hashing URI and vhost name.
+ * If vhost is not yet assigned, fall back to using the Host header.
  */
 unsigned long
 tfw_http_req_key_calc(TfwHttpReq *req)
```
const-t, line 7582: You changed only single function, however host header used in other places where key is used. `tfw_http_req_key_calc()` returns key that will be used as key in TDB, but also we do store/compare string key to avoid collisions. See `TFW_CACHE_REQ_KEYITER()`.

fw/t/unit/test_http_cache_key.c
```diff
@@ -0,0 +1,242 @@
...
+	
+	/* Compute key without vhost */
+	req->vhost = NULL;
+	req->hash = 0;
+	key2 = tfw_http_req_key_calc(req);
+	
+	/* Keys should be the same since only uri_path is used for HM requests */
+	EXPECT_EQ(key1, key2);
+	
+	//test_req_free(req); // TODO: kernel stuck here, why?
```
const-t, line 233: Please fix it or remove TODO if it already fixed.

const-t (changes requested):

PR #2378, opened 2025-03-26
Close all sockets when Tempesta FW shutdowned

EvgeniiMekhanik:
When we close socket we check if there is some pending data in socket write queue and if it is true we don't drop connection immediately, but wait until all pending data will be sent (usually socket in TCP_FIN_WAIT1 state in this case).
When Tempesta FW is shutdowned we should not only drop connection, but also close such socket:
- We can't send any more data, because connection is dropped and `sk_write_xmit` callback is equal to zero, so we can't encrypt data.
- Socket can leak if any error occurs, because we don't call `tcp_done` during handling error when socket in TCP_FIN_WAIT1 state.

### DIFF
--- a/fw/sock.c
+++ b/fw/sock.c
@@ -669,7 +669,14 @@ ss_do_close(struct sock *sk, int flags)
 		tcp_set_state(sk, TCP_CLOSE);
 		tcp_send_active_reset(sk, sk->sk_allocation);
 	} else if (tcp_close_state(sk)) {
+		/*
+		 * Set this flag to prevent calling `tcp_done` from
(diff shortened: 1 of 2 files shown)


EvgeniiMekhanik:
It seems that we should discuss about this PR, because I see some problems in current implementation of `tcp_tfw_handle_error` in kernel code.

krizhanovsky (approved):
We discussed the solution and have tried to make the fix more elegant. I have no better proposal, so @EvgeniiMekhanik @const-t I just approve the change and it's your call

const-t (approved):

Issue #2381, opened 2025-03-27
JA5 Filter: LRU list in get_fingerprint_rates is not updated for existing entries, leading to inaccurate eviction


MorganaFuture:
**Describe the issue**
The LRU (Least Recently Used) list update logic within `get_fingerprint_rates` in `ja5_filter.c` is incomplete. When an existing Rates entry is retrieved via `tdb_rec_get_alloc`, its position in the `storage->lru_list` is not updated (i.e., it's not moved to the tail/MRU position). Only newly created entries are correctly added to the tail. This can lead to frequently accessed but older entries being prematurely evicted when TDB is full, as they incorrectly appear as "least recently used.

**Impact**
Reduced effectiveness of the JA5 filter cache, potentially leading to increased TDB churn and minor performance degradation due to unnecessary re-creation of frequently used rate-tracking entries.

**Suggestion**
Add `list_move_tail(&rates->list_node, &storage->lru_list)`

**File**
ja5_filter.c.

krizhanovsky:
Done in https://github.com/tempesta-tech/tempesta/pull/2382

PR #2382, opened 2025-03-27
fix: LRU list in get_fingerprint_rates is not updated for existing en…

MorganaFuture:
**What**
Updated the LRU implementation in get_fingerprint_rates() to properly move accessed records to the end of the LRU list.

**Why**
- The previous implementation only added new entries to the LRU list but didn't update the position of existing entries when accessed
- This resulted in imperfect LRU behavior where frequently accessed entries added earlier could be evicted before rarely accessed entries added later
- The fix ensures that any accessed entry is properly moved to the Most Recently Used position (tail of the LRU list)
- This change improves cache efficiency by properly maintaining the LRU order, keeping frequently used fingerprints in memory and allowing truly least recently used entries to be evicted first

[2381](https://github.com/tempesta-tech/tempesta/issues/2381)


MorganaFuture:
@krizhanovsky Could you please assign someone to review this PR?

fw/ja5_filter.c
```diff
@@ -150,9 +150,13 @@ get_fingerprint_rates(Storage *storage, u64 fingerprint)
...
 
 	spin_lock(&storage->lru_list_lock);
-	/* The record still was not added to the LRU list */
-	if (list_empty(&rates->list_node))
+	if (list_empty(&rates->list_node)) {
+		/* Entry is new or was somehow removed from list, add it to tail (MRU) */
 		list_add_tail(&rates->list_node, &storage->lru_list);
+	} else {
+		/* Entry already exists and is in the list, move it to tail (MRU) */
+		list_move_tail(&rates->list_node, &storage->lru_list);
```
EvgeniiMekhanik, line 158: It seems that on string 130 we iterate in reverse order and delete items. So it seems that first of all we delete items from the tail. So it seems that we should use `list_add_head`  at line 155

MorganaFuture: Done

EvgeniiMekhanik (approved):

krizhanovsky:
@MorganaFuture merged, thank you so much for the contribution!

Issue #2136, opened 2024-06-10
Headers processing optimization


const-t:
# Motivation

`tfw_http_msg_expand_from_pool()` calls a lot of times around the code(especially after merging #1955), during headers processing for both HTTP1 and HTTP2. For every call it perform a lot skb checks, that can be done only once at the end of message construction. We should choose another strategy similar to following: Allocate a buffer, fill the buffer with headers, add the buffer to skb. Looks like we can do this for cached responses as well. In cases when we can calculate target size of headers, we can allocate enough memory and copy headers even without checks.

krizhanovsky:
Need to address comment https://github.com/tempesta-tech/tempesta/pull/1955#discussion_r1426048704 for `__tfw_http_msg_expand_from_pool()`:

> Now the function is called a lot of times, so it should be well optimized. I think it makes sense to make `if (unlikely(skb_headlen(it->skb)))` since even if we have a lot of such skbs, then only the first call hits the condition and all following calls go further. The inner `if` can eliminate `unliekly` - it's no sense to have 2 nested unlikely statements.
>
> Also probably it makes sense to add `if (likely(TFW_STR_PLAIN(s)))` to `TFW_STR_FOR_EACH_CHUNK_INIT()` and `unlikely(room == 0)` in the `while` loop at the below.

const-t:
@krizhanovsky My proposal described in this task almost the same with yours, pre-allocate block around 0.5-1KB(for http1 we can estimate the size relying on the size of forwarded message) maybe from `TfwPool` or `pg_skb_alloc()`, pool may have some advantages, cause it have contignious memory and even if the first allocated block was too small we can expand it by another block of this size by the end of the page. It may allow us to keep small number of fragments. Then copy headers to the new buffer and only after that add the buffer to skb as fragment.

krizhanovsky:
We need to rework the `__tfw_http_msg_expand_from_pool()` allocator and address https://github.com/tempesta-tech/tempesta/pull/2393/files#r2054412506

PR #2393, opened 2025-04-02
TMP

EvgeniiMekhanik:


fw/http_msg.c
```diff
@@ -1679,11 +1645,6 @@ __tfw_http_msg_expand_from_pool(TfwHttpResp *resp, const TfwStr *str,
 
 	BUG_ON(it->skb->len > SS_SKB_MAX_DATA_LEN);
 
-	if (skb_headlen(it->skb)) {
```
krizhanovsky, line 1682: What about to add `WARN_ON_ONCE(skb_headlen(it->skb))` here in case if the assumption that we don't need the transformation is wrong?

EvgeniiMekhanik: Yes ad it.

fw/http_msg.c
```diff
@@ -1666,32 +1668,36 @@ __tfw_http_msg_expand_from_pool(TfwHttpResp *resp, const TfwStr *str,
...
-					r = __tfw_http_msg_move_body(resp,
-								     nskb);
-					if (unlikely(r)) {
+					r = __tfw_http_msg_move_body(resp, nskb,
+								     &frag);
+					if (unlikely(r < 0)) {
 						T_WARN("Error during moving body");
 						return r;
 					}
+					body_was_moved = !!r;
```
const-t, line 1687: Why do we need double negotiation if possible values only 1 and 0?

const-t (approved):

fw/http2.c
```diff
@@ -486,7 +486,9 @@ tfw_h2_hpack_encode_trailer_headers(TfwHttpResp *resp)
 	TfwHttpHdrMap *map = resp->mit.map;
 	TfwHttpHdrTbl *ht = resp->h_tbl;
 	unsigned int i;
-	int r;
+	int r = 0;
+
+	__set_bit(TFW_HTTP_B_RESP_ENCODE_TRAILERS, resp->flags);
```
const-t, line 491: can we rely on `stream->xmit.state` here?

krizhanovsky (changes requested):
Probably we should just drop trailers for HTTP/2 clients, but stop doing the quick dirty fixes.

fw/http2.c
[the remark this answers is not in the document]
EvgeniiMekhanik: I add TODO and link to 2136.

krizhanovsky (approved):
OK, let's fix the flags in #2136

PR #2400, opened 2025-04-10
fix: header table corruption

const-t:
We had a bug during the suffix match in function
`tfw_http_search_cookie()`, before comparing
the name we store current chunk `t = *chunk`, then modify and compare the chunk, after we rostore the chunk `*chunk = t;` that leads to header corruption, because macros `TFW_STR_EQ_CSTR` may assign new
address to `chunk`, therefore restoration happens to a newly assigned adress instead of original adress in `chunk`. Fixed by storing address to temporary
variable and then restore `data` and `len` by this adress.

`__tfw_str_eq_cstr` rewritten to inline function.
Inline functions is more clear, less error prone and suitable in this case, because we don't need to define tmp vars and explicitly save values, only move `chunk` pointer forward. Generated assembly will be the same.

### DIFF
--- a/fw/http_match.c
+++ b/fw/http_match.c
@@ -876,6 +876,31 @@ tfw_http_verify_hdr_field(tfw_http_match_fld_t field, const char **hdr_name,
 	return 0;
 }
 
+/* Simple version of tfw_str_eq_cstr. */
+static bool __always_inline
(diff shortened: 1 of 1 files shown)


const-t:
@krizhanovsky commit extracted from PR #1955.

EvgeniiMekhanik (approved):

PR #2403, opened 2025-04-13
Mekhanik evgenii/1411 7

EvgeniiMekhanik:


fw/hpack.c
```diff
@@ -456,41 +508,61 @@ static inline int
...
-		++hdr->len;
-		if (!TFW_STR_PLAIN(hdr))
-			++last->len;
+	/*
+	 * If the new page was allocated or it->pos points to
+	 * the buffer, which was allocated to store previously
+	 * decoded header (it->to_alloc != 0) we should expand
+	 * header.
+	 */
+	if (!np && !it->to_alloc) {
```
const-t, line 553: The only one thing is not clear for me. Why do we do `tfw_hpack_exp_hdr()` when `it->pos points to the buffer, which was allocated to store previously decoded header`? From my point of view, in this case we can assume that previously allocated buffer(`rspace` indicates how many space in this buffer) being allocated at the same page as new buffer(if not `np` of course). What I mean:
1. We allocate buffer.
2. Process header.
3. Go to to the next header.
3. Use remaining `rspace`, it's not zero, but not enough to process current header completely.
4. `rspace` exhausted. We do `tfw_pool_alloc_not_align_np()` if `!np` it allocates data on the same page where previous `rspace` being allocated before.
5. Adjust the length instead of creating new chunk.

Or there is the case when new buffer will be allocated on the different page? But if so in this case we can't safely use `rspace` as well.

EvgeniiMekhanik: If we allocate a whole page we loose a lot of memory for a small headers i think

const-t: > If we allocate a whole page we loose a lot of memory for a small headers i think
We always doing it allocating pool

EvgeniiMekhanik: According to our discussion we decide to use one strategy for all places where `it->pool` is used. and use rspace if it is available

fw/hpack.c
```diff
@@ -456,41 +518,44 @@ static inline int
...
 		return 0;
 	}
 
+	to_alloc = it->to_alloc ? it->to_alloc : tfw_hpack_buffer_sz(1);
+	r = tfw_hpack_buffer_get(it, to_alloc, false);
+	if (unlikely(r))
+		return r;
+
+	T_DBG3("%s: it->rspace=%lu, sym=%c\n",
+	       __func__, it->rspace, sym);
```
const-t:
```suggestion
	T_DBG3("%s: it->rspace=%lu, sym=%c\n", __func__, it->rspace, sym);
```

const-t (approved):

fw/hpack.c
```diff
@@ -1275,10 +1348,17 @@ tfw_hpack_hdr_set(TfwHPack *__restrict hp, TfwHttpReq *__restrict req,
 
 	d = d_hdr->chunks;
 	TFW_STR_FOR_EACH_CHUNK(s, s_hdr, end) {
+		if (unlikely(it->rspace < s->len)) {
+			r = tfw_hpack_buffer_get(it, it->to_alloc, true);
```
EvgeniiMekhanik, line 1350: Here we have a discussion with @const-t and can't find a solution. Previously we allocate memory only once  `data = tfw_pool_alloc_not_align(it->pool, s_hdr->len)` but this could lead to memory loss if there is no enough memory in pool and we allocate new pages. Now I add extra if in the loop and don't loose memory. Whats better? We can call `tfw_hpack_buffer_get` only once with true flag at the beginning of the function but loose memory as it was before. @krizhanovsky what do you think?

fw/hpack.c
```diff
@@ -562,11 +626,12 @@ huffman_decode_tail(TfwHPack *__restrict hp, TfwHttpReq *__restrict req,
...
 	}
 	if (likely(offset == 0)) {
-		if ((i ^ (HT_EOS_HIGH >> 1)) < (1U << -hp->curr)) {
+		if ((i ^ (HT_EOS_HIGH >> 1)) < (1U << -hp->curr))
 			return T_OK;
-		}
 	}
 	return T_COMPRESSION;
+
+#undef ADJUST_EXTRA_RSPACE
```
krizhanovsky: The macro was removed

fw/hpack.c
```diff
@@ -30,6 +30,54 @@
...
+	return (PAGE_SIZE << order) - sizeof(TfwPoolChunk) - 1;
+}
+
+/**
+ * There are two cases when we allocate space for header:
+ * - it->rspace is equal to zero. In this case we allocate
+ *   `len` bytes;
+ * - it->rspace is not equal to zero. This means that we
+ *   already have buffer with rspace bytes available.
+ *   If this buffer is rather big to contains the whole
```
krizhanovsky:
```suggestion
 *   If this buffer is rather big to contain the whole
```

fw/hpack.c
```diff
@@ -30,6 +30,54 @@
...
+ *   header or it is greater then sizeof(TfwStr) we use
```
krizhanovsky:
```suggestion
 *   header or it is greater than sizeof(TfwStr) we use
```

fw/msg.h
```diff
@@ -84,12 +86,13 @@ typedef struct {
...
 	unsigned long	hdrs_len;
+	unsigned long	rspace;
+	char		*pos;
 	unsigned int	hdrs_cnt;
 	char		__off[0];
 	TfwStr		hdr;
-	char		*pos;
-	unsigned long	rspace;
 	unsigned long	nm_len;
+	unsigned long	to_alloc;
```
krizhanovsky: Why do we need 8 bytes for `hdrs_len`, `nm_len`, `to_alloc`, `rspace` - is it OK if the values ever exceed 4G?  If we extend the structure, let's make some work in optimizing it a little bit - ideally if it becomes smaller than it was

fw/hpack.c
```diff
@@ -30,6 +30,54 @@
 
 #include "hpack_tbl.h"
 
+static inline int
+tfw_hpack_buffer_sz(unsigned long len)
+{
+	unsigned int order = get_order(sizeof(TfwPoolChunk) + len);
+
+	return (PAGE_SIZE << order) - sizeof(TfwPoolChunk) - 1;
```
krizhanovsky: what -1 stands for?

fw/hpack.c
krizhanovsky: There shouldn't be a lot of chunks, so the loop should be short. I think we're fine, at least none of us is able to say what is better with confedence and it doesn't make sense to microbenchmark this place

fw/msg.h
EvgeniiMekhanik: Yes fixed

fw/hpack.c
EvgeniiMekhanik: Yes fixed

fw/hpack.c
EvgeniiMekhanik: Yes fixed

fw/hpack.c
EvgeniiMekhanik: Yes fixed

fw/hpack.c
EvgeniiMekhanik: Yes fixed

PR #2415, opened 2025-05-01
Mekhanik evgenii/1411 9

EvgeniiMekhanik:


const-t (approved):
LGTM, but it seems we still clear the flag in `tfw_hpack_hdr_set()` when if-none-match in dyamic table

PR #2417, opened 2025-05-06
Remove `__restrict` from tfw_hpack_rbuf_calc

const-t:
Removed unnecessary `__restrict`, that hasn't effect on the function. Also `tbl` and `it` might points to the same tructure and this is legal and used in `tfw_hpack_set_rbuf_size()`


const-t:
Port from 0.9 milestone.

PR #2426, opened 2025-05-13
fix: call frang for non-tempesta socket

const-t:
In `tfw_classify_conn_estab()` force Tempesta to call `frang_conn_new()` only for Tempesta's socket.


krizhanovsky (approved):
LGTM. But probably we should check the flag for all `tempesta_ops` callbacks, at least for code consistency and further extensions. We can use likely/unlikely to mitigate the overhead.

const-t:
@krizhanovsky I've added the check to `tfw_classify_conn_close()`, but actually it does nothing, only a hint for reader. In `tfw_classify_tcp()` added as well, condition must be outoptimized there.

PR #2427, opened 2025-05-13
fix: call frang for non-tempesta socket [port 0.8]

const-t:
In `tfw_classify_conn_estab()` force Tempesta to call `frang_conn_new()` only for Tempesta's socket.


krizhanovsky (approved):
LGTM

Issue #2314, opened 2025-01-06
tfw_logger: missing logs during the first second of application startup


symstu-tempesta:
**Describe the issue**
No logs are present in the ClickHouse database during the first second of application startup, although logs do appear in dmesg.

**Expected Behavior**
Logs are present in both dmesg and the ClickHouse access_log table.

**To Reproduce**
Start Tempesta FW, wait for it to begin responding, and then send a request as soon as possible. I reproduced this with a simple test.
```python
import time
from datetime import datetime

from test_suite import tester
from helpers.error import ProcessBadExitStatusException
import clickhouse_connect

__author__ = "Tempesta Technologies, Inc."
__copyright__ = "Copyright (C) 2018-2025 Tempesta Technologies, Inc."
__license__ = "GPL2"


class TestClickhouseLogsBaseTest(tester.TempestaTest):
    backends = [
        {
            "id": "deproxy",
            "type": "deproxy",
            "port": "8000",
            "response": "static",
            "response_content": "HTTP/1.1 200 OK\r\n" "Content-Length: 0\r\n\r\n",
        },
    ]

    tempesta = dict(
        config="""
            listen 80;
            server ${server_ip}:8000;

            access_log dmesg mmap mmap_host=localhost mmap_log=/tmp/access.log;
        """
    )
    clients = [
        {
            "id": "deproxy",
            "type": "deproxy",
            "addr": "${tempesta_ip}",
            "port": "80",
        }
    ]

    def setUp(self):
        super(TestClickhouseLogsBaseTest, self).setUp()

        self.clickhouse_client = clickhouse_connect.get_client()

        res = self.clickhouse_client.command('exists table access_log')

        if res:
            self.clickhouse_client.command('delete from access_log where true')

        try:
            print('removing access log')
            tempesta = self.get_tempesta()
            res = tempesta.node.run_cmd('rm /tmp/access.log')
            assert res == (b'', b'')

        except ProcessBadExitStatusException as e:
            print(e)

    def test_missing_log(self):
        self.start_all_servers()
        self.start_tempesta()
        self.deproxy_manager.start()
        self.start_all_clients()
        self.assertTrue(self.wait_all_connections())

        client = self.get_client('deproxy')
        client.start()

        client.send_request(
            request=client.create_request(method="GET", headers=[]),
            expected_status_code='200',
            timeout=10,
        )

        request_in_dmesg = self.oops.find("GET / HTTP/1.1")
        assert request_in_dmesg is True

[...]

PR #2428, opened 2025-05-21
Morgana future/fix/configuration parsing error mmap log buffer size

MorganaFuture:
What
This PR introduces a JSON-based configuration system for the Tempesta FW logger daemon with:

- A dedicated TfwLoggerConfig class for logger settings
- JSON configuration file support via Boost's property_tree
- Two operation modes: daemon (background) and handle (foreground for debugging)
- Support for a logger_config attribute in the access_log directive
- Updates to Tempesta scripts for compatibility
- Backward compatibility with existing configuration
- Add tests 

Why
The Tempesta logger configuration is currently scattered across multiple parameters, making it hard to manage. A dedicated JSON file provides a cleaner approach that separates logger settings from the main configuration. This structured format improves error handling and simplifies troubleshooting with the new foreground mode.

Links
https://github.com/tempesta-tech/tempesta/issues/2313

### DIFF
--- a/scripts/tempesta.sh
+++ b/scripts/tempesta.sh
@@ -3,7 +3,7 @@
 # Tempesta FW service script.
 #
 # Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
-# Copyright (C) 2015-2024 Tempesta Technologies, Inc.
+# Copyright (C) 2015-2025 Tempesta Technologies, Inc.
 #
 # This program is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License as published by
@@ -38,17 +38,15 @@ tdb_path=${TDB_PATH:="$TFW_ROOT/db/core"}
 tfw_path=${TFW_PATH:="$TFW_ROOT/fw"}
 tls_path=${TLS_PATH:="$TFW_ROOT/tls"}
 lib_path=${LIB_PATH:="$TFW_ROOT/lib"}
-utils_path=${UTILS_PATH:="$TFW_ROOT/utils"}
+logger_path=${LOGGER_PATH:="$TFW_ROOT/utils"}
 tfw_cfg_path=${TFW_CFG_PATH:="$TFW_ROOT/etc/tempesta_fw.conf"}
 tfw_cfg_temp=${TFW_CFG_TMPL:="$TFW_ROOT/etc/tempesta_tmp.conf"}
+tfw_logger_config="$TFW_ROOT/etc/tfw_logger.json"
 
 tfw_logger_should_start=0
 tfw_logger_pid_path="/var/run/tfw_logger.pid"
 tfw_logger_timeout=3
-mmap_host=""
-mmap_log=""
-mmap_user=""
-mmap_password=""
+tfw_logger_config_path=""
 
 lib_mod=tempesta_lib
 tls_mod=tempesta_tls
@@ -126,22 +124,16 @@ templater()
 		fi
 	done < "$tfw_cfg_path"
 
+	tfw_logger_config_path=$(echo "$cfg_content" | grep -oP 'access_log\s+.*logger_config=\K[^;]*' | sed 's/"//g' | sed "s/'//g")
+
 	opts=$(get_opts "$cfg_content" "access_log")
 	while read -r line; do
-		if [ $(opt_exists "$line" "mmap"; echo $?) -ne 0 ]; then
+		if [ $(opt_exists "$line" "mmap"; echo $?) -ne 0 ] || [ $(opt_exists "$line" "logger_config"; echo $?) -ne 0 ]; then
 			tfw_logger_should_start=1
-			mmap_log=$(get_opt_value "$line" "mmap_log")
-			mmap_host=$(get_opt_value "$line" "mmap_host")
-			mmap_user=$(get_opt_value "$line" "mmap_user")
-			mmap_password=$(get_opt_value "$line" "mmap_password")
-
-			[[ -n "$mmap_log" && -n "$mmap_host" ]] ||
-				error "if mmaps enabled in access log, there have to be mmap_host and mmap_log options"
 		fi
 	done <<< "$opts"
 
-	cfg_content=$(remove_opts_by_mask "$cfg_content" "mmap_")
-
+	cfg_content=$(remove_opts_by_mask "$cfg_content" "logger_config=")
 	echo "$cfg_content" > $tfw_cfg_temp
 }
 
@@ -311,36 +303,64 @@ start_tfw_logger()
 		return
 	fi
 
-	if [ -z "$mmap_host" ] || [ -z "$mmap_log" ]; then
-		error "You need to specify 'mmap_host' and 'mmap_log' "`
-		      `"if access_log mmap was specified"
-		return
+	local config_path
+	if [ -n "$tfw_logger_config_path" ]; then
+		config_path="$tfw_logger_config_path"
+		echo "...starting tfw_logger with custom config: $config_path"
+	else
+		config_path="$tfw_logger_config"
+		echo "...starting tfw_logger with default config: $config_path"
 	fi
 
-	"$utils_path/tfw_logger" -H "$mmap_host" -l "$mmap_log" -u "$mmap_user" -p "$mmap_password" ||
-		error "cannot start tfw_logger daemon"
+	# Check if config file exists - CRITICAL ERROR if missing when logger is required
+	if [ ! -f "$config_path" ]; then
+		sysctl -e -w net.tempesta.state=stop 2>/dev/null || true
+		unload_modules
+		tfw_irqbalance_revert
+		error "TFW Logger configuration file not found: $config_path. Cannot start Tempesta with access_log mmap enabled."
+	fi
 
+	# Start daemon
+	"$logger_path/tfw_logger" --config="$config_path" || {
+		sysctl -e -w net.tempesta.state=stop 2>/dev/null || true
+		unload_modules
+		tfw_irqbalance_revert
+		error "cannot start tfw_logger daemon"
+	}
+
+	# Wait for daemon to start and create PID file
 	start_time=$(date +%s)
 	while [[ ! -f "$tfw_logger_pid_path" ]]; do
 		current_time=$(date +%s)
 		elapsed_time=$((current_time - start_time))
 
 		if (( elapsed_time >= tfw_logger_timeout )); then
-			sysctl -e -w net.tempesta.state=stop
+			# Try to cleanup any failed start
+			"$logger_path/tfw_logger" --stop 2>/dev/null || true
+			sysctl -e -w net.tempesta.state=stop 2>/dev/null || true
 			unload_modules
 			tfw_irqbalance_revert
-			error "tfw_logger failed to start, see $mmap_log for details"
+			error "tfw_logger failed to start within $tfw_logger_timeout seconds, see logs for details"
 		fi
 
 		sleep 0.1
 	done
 
+	echo "...tfw_logger started successfully"
 }
 
 stop_tfw_logger()
 {
 	if [ -e $tfw_logger_pid_path ]; then
-		"$utils_path/tfw_logger" -s
+		echo "...stopping tfw_logger"
+		"$logger_path/tfw_logger" --stop || \
+			echo "Warning: Failed to stop tfw_logger daemon gracefully"
+
+		# Remove stale PID file if it still exists
+		if [ -e "$tfw_logger_pid_path" ]; then
+			echo "Warning: PID file still exists, removing it"
+			rm -f "$tfw_logger_pid_path" 2>/dev/null || true
+		fi
 	fi
 }
 
@@ -358,7 +378,7 @@ start()
 	TFW_STATE=$(sysctl net.tempesta.state 2> /dev/null)
 	TFW_STATE=${TFW_STATE##* }
 	TFW_LOGGER_EXEC=$(expr "$TFW_STATE" != "start")
-	
+
 	check_configuration_file_presense
 
 	if [[ -z ${TFW_STATE} ]]; then
--- a/utils/pidfile.cc
+++ b/utils/pidfile.cc
@@ -1,0 +1,216 @@
+/**
+ *		Tempesta FW
+ *
+ * Copyright (C) 2025 Tempesta Technologies, Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License,
+ * or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 59
+ * Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ */
+
+#include "pidfile.hh"
+
+#include <fcntl.h>
+#include <signal.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include <chrono>
+#include <cstring>
+#include <fstream>
+#include <iostream>
+#include <thread>
+
+#include "error.hh"
+
+constexpr std::chrono::milliseconds STOP_WAIT_INTERVAL{10};
+
+int
+pidfile_check(const std::string &fname)
+{
+	struct flock fl;
+	int fd = -1;
+
+	fd = open(fname.c_str(), O_RDWR);
+	if (fd == -1) {
+		if (errno == ENOENT) {
+			errno = 0;
+			return 0;
+		}
+		return -1;
+	}
+
+	// Set locking parameters
+	memset(&fl, 0, sizeof(fl));
+	fl.l_type = F_WRLCK;
+	fl.l_whence = SEEK_SET;
+	fl.l_start = 0;
+	fl.l_len = 0;
+	fl.l_pid = getpid();
+
+	if (fcntl(fd, F_SETLK, &fl) == -1) {
+		close(fd);
+
+		// File is locked by another process
+		if (errno == EAGAIN || errno == EACCES) {
+			errno = 0;
+			throw Except("Daemon is already running");
+		}
+		return -1;
+	}
+
+	// Unlock
+	fl.l_type = F_UNLCK;
+	if (fcntl(fd, F_SETLK, &fl) == -1) {
+		close(fd);
+		return -1;
+	}
+
+	close(fd);
+	return 0;
+}
+
+int
+pidfile_create(const std::string &fname)
+{
+	int fd = -1;
+	struct flock fl;
+	char pid_str[64] = {0};
+	int pid_str_len;
+	mode_t mask;
+
+	// Get PID as string
+	pid_str_len = snprintf(pid_str, sizeof(pid_str), "%ld",
+			       static_cast<long>(getpid()));
+
+	// Remove old PID file
+	if (access(fname.c_str(), F_OK) != -1) {
+		unlink(fname.c_str());
+		errno = 0;
+	}
+
+	// Create new PID file with appropriate permissions
+	mask = S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH; // -rw-r--r--
+	fd = open(fname.c_str(), O_RDWR | O_CREAT, mask);
+	if (fd == -1) {
+		std::cerr << "Cannot create pidfile: " << fname << std::endl;
+		return -1;
+	}
+
+	// Write PID
+	if (write(fd, &pid_str[0], pid_str_len) == -1) {
+		std::cerr << "Cannot write pid into pidfile: " << fname
+			  << std::endl;
+		close(fd);
+		return -1;
+	}
+
+	// Set locking parameters
+	memset(&fl, 0, sizeof(fl));
+	fl.l_type = F_WRLCK;
+	fl.l_whence = SEEK_SET;
+	fl.l_start = 0;
+	fl.l_len = 0;
+	fl.l_pid = getpid();
+
+	// Lock file
+	if (fcntl(fd, F_SETLK, &fl) == -1) {
+		close(fd);
+		std::cerr << "Cannot lock pidfile: " << fname << std::endl;
+		return -1;
+	}
+
+	return fd;
+}
+
+void
+pidfile_remove(const std::string &fname, int fd)
+{
+	struct flock fl;
+
+	// Set unlocking parameters
+	memset(&fl, 0, sizeof(fl));
+	fl.l_type = F_UNLCK;
+	fl.l_whence = SEEK_SET;
+	fl.l_start = 0;
+	fl.l_len = 0;
+
+	// Release lock
+	if (fcntl(fd, F_SETLK, &fl) == -1)
+		std::cerr << "Cannot unlock pidfile: " << fname << std::endl;
+
+	// Close file descriptor
+	close(fd);
+
+	// Remove PID file
+	if (unlink(fname.c_str()) == -1)
+		std::cerr << "Cannot remove pidfile: " << fname << std::endl;
+}
+
+void
+pidfile_stop_daemon(const std::string &fname)
+{
+	pid_t pid;
+	std::ifstream pid_file(fname);
+
+	if (!pid_file)
+		throw Except("No PID file found at '{}'. "
+			     "Is the daemon running?", fname);
+
+	pid_file >> pid;
+	pid_file.close();
+
+	if (pid <= 0)
+		throw Except("Invalid PID in PID file: {}", pid);
+
+	// Send SIGTERM to daemon
+	if (kill(pid, SIGTERM) < 0) {
+		if (errno == ESRCH) {
+			// Process not running - remove stale PID file
+			unlink(fname.c_str());
+			return;
+		}
+		throw Except("Failed to stop daemon (PID {}): {}",
+			     pid, strerror(errno));
+	}
+
+	// Wait for graceful shutdown
+	constexpr int GRACEFUL_WAIT_ITERATIONS = 50; // 50 * 10ms = 500ms
+	for (int i = 0; i < GRACEFUL_WAIT_ITERATIONS; ++i) {
+		if (kill(pid, 0) == -1 && errno == ESRCH)
+			return; // Process died gracefully
+
+		std::this_thread::sleep_for(STOP_WAIT_INTERVAL);
+	}
+
+	// Graceful shutdown failed, try SIGKILL
+	if (kill(pid, SIGKILL) < 0) {
+		if (errno == ESRCH)
+			return; // Process already stopped
+
+		throw Except("Failed to kill daemon (PID {}): {}",
+			     pid, strerror(errno));
+	}
+
+	// Wait for force kill to take effect
+	constexpr int FORCE_WAIT_ITERATIONS = 50; // 50 * 10ms = 500ms
+	for (int i = 0; i < FORCE_WAIT_ITERATIONS; ++i) {
+		if (kill(pid, 0) == -1 && errno == ESRCH)
+			return; // Process force-killed
+
+		std::this_thread::sleep_for(STOP_WAIT_INTERVAL);
+	}
+
+	// If we get here, something is seriously wrong
+	throw Except("Failed to stop daemon (PID {}) even with SIGKILL", pid);
+}
--- a/utils/tfw_logger.cc
+++ b/utils/tfw_logger.cc
@@ -20,65 +20,96 @@
 
 #include <fcntl.h>
 #include <signal.h>
-#include <spdlog/sinks/basic_file_sink.h>
-#include <spdlog/spdlog.h>
 #include <stdio.h>
 #include <unistd.h>
 
+#include <atomic>
 #include <fstream>
-#include <future>
 #include <iomanip>
 #include <iostream>
 #include <sstream>
+#include <thread>
+#include <vector>
 
 #include <boost/program_options.hpp>
-#include <clickhouse/base/socket.h>
-#include <clickhouse/client.h>
 
 #include "../fw/access_log.h"
 #include "clickhouse.hh"
-#include "mmap_buffer.hh"
 #include "error.hh"
+#include "mmap_buffer.hh"
+#include "pidfile.hh"
+#include "tfw_logger_config.hh"
+
+#include <clickhouse/base/socket.h>
+#include <clickhouse/client.h>
+#include <spdlog/sinks/basic_file_sink.h>
+#include <spdlog/spdlog.h>
 
 namespace po = boost::program_options;
 
-constexpr char dev_path[]	= "/dev/tempesta_mmap_log";
-constexpr char table_name[]	= "access_log";
-constexpr char pid_file_path[]	= "/var/run/tfw_logger.pid";
-constexpr std::chrono::seconds		wait_for_dev{1};
-constexpr std::chrono::milliseconds	wait_for_stop(10);
-constexpr std::chrono::seconds		reconnect_min_timeout{1};
-constexpr std::chrono::seconds		reconnect_max_timeout{16};
+constexpr char dev_path[] = "/dev/tempesta_mmap_log";
+constexpr char pid_file_path[] = "/var/run/tfw_logger.pid";
+constexpr char default_config_path[] = "/etc/tempesta/tfw_logger.json";
+constexpr char default_log_path[] = "/var/log/tempesta/tfw_logger.log";
+constexpr std::chrono::seconds wait_for_dev{1};
+constexpr std::chrono::seconds reconnect_min_timeout{1};
+constexpr std::chrono::seconds reconnect_max_timeout{16};
+
+// Global state
+std::atomic<bool> stop_flag{false};
+static thread_local bool uncritical_error;
+static TfwLoggerConfig config;
+
+namespace
+{
+
+/**
+ * Command line options structure
+ */
+struct ParsedOptions {
+	bool				help = false;
+	bool				stop_daemon = false;
+	bool				foreground = false;
+	bool				test_config = false;
+
+	std::optional<fs::path>		config_path;
+	std::optional<std::string>	clickhouse_host;
+	std::optional<uint16_t>		clickhouse_port;
+	std::optional<std::string>	clickhouse_table;
+	std::optional<std::string>	clickhouse_user;
+	std::optional<std::string>	clickhouse_password;
+	std::optional<size_t>		clickhouse_max_events;
+	std::optional<int>		clickhouse_max_wait_ms;
+	std::optional<fs::path>		log_path;
+};
 
 typedef struct {
-	const char		*name;
-	clickhouse::Type::Code	code;
+	const char *name;
+	clickhouse::Type::Code code;
 } TfwField;
 
 static const TfwField tfw_fields[] = {
-	[TFW_MMAP_LOG_ADDR]		= {"address", clickhouse::Type::IPv6},
-	[TFW_MMAP_LOG_METHOD]		= {"method", clickhouse::Type::UInt8},
-	[TFW_MMAP_LOG_VERSION]		= {"version", clickhouse::Type::UInt8},
-	[TFW_MMAP_LOG_STATUS]		= {"status", clickhouse::Type::UInt16},
-	[TFW_MMAP_LOG_RESP_CONT_LEN]	= {"response_content_length", clickhouse::Type::UInt32},
-	[TFW_MMAP_LOG_RESP_TIME]	= {"response_time", clickhouse::Type::UInt32},
-	[TFW_MMAP_LOG_VHOST]		= {"vhost", clickhouse::Type::String},
-	[TFW_MMAP_LOG_URI]		= {"uri", clickhouse::Type::String},
-	[TFW_MMAP_LOG_REFERER]		= {"referer", clickhouse::Type::String},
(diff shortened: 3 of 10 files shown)


scripts/tempesta.sh
```diff
@@ -329,18 +323,19 @@ start_tfw_logger()
...
 	done
-
 }
 
 stop_tfw_logger()
 {
 	if [ -e $tfw_logger_pid_path ]; then
-		"$utils_path/tfw_logger" -s
+		echo "...stopping tfw_logger"
+		"$logger_path/tfw_logger" --stop || \
```
krizhanovsky, line 337: For some reason you removed `stop_daemon()` as well as parsing `-s` or `--stop` options, so this seems not working anymore

logger/tfw_logger.cc
```diff
@@ -0,0 +1,687 @@
...
+   	// Setup signal handlers for graceful shutdown
+   	set_sig_handlers();
+   	
+   	// Open mmap device
+   	spdlog::info("Opening device: {}", dev_path);
+   	fd = open_device();
+   	if (fd < 0) {
+   		spdlog::error("Failed to open device");
+   		return 1;
+   	}
```
krizhanovsky, line 657: Bunch of logic is removed, it seems mostly about daemonization, but I'm not sure that only that logic - it's hard to compare the files line by line. It seems due to my comment https://github.com/tempesta-tech/tempesta/pull/2363#discussion_r2023536183  , which is wrong - in some of our projects we use docker, which doesn't need `daemon()`, but once we tried to manage with systemctl, we needed the daemonization mode. @althusky can provide more details on this

krizhanovsky: pid file also removed. By the way, it seems https://github.com/tempesta-tech/escudo/blob/main/manager/main.cc#L73 implements them in a more accurate way, so it could be better to borrow the code from there

logger/tfw_logger_config.hh
```diff
@@ -0,0 +1,95 @@
...
+	// Override methods for command line arguments
+	void override_log_path(const fs::path& path) { log_path_ = path; }
+	void override_buffer_size(size_t size) { buffer_size_ = size; }
+	void override_cpu_count(size_t count) { cpu_count_ = count; }
+	void override_clickhouse_host(const std::string& host) { clickhouse_.host = host; }
+	void override_clickhouse_port(uint16_t port) { clickhouse_.port = port; }
+	void override_clickhouse_user(const std::string& user) { clickhouse_.user = user; }
+	void override_clickhouse_password(const std::string& password) { clickhouse_.password = password; }
+	void override_clickhouse_max_events(size_t events) { clickhouse_.max_events = events; }
+	void override_clickhouse_max_wait(int ms) { clickhouse_.max_wait = std::chrono::milliseconds(ms); }
```
krizhanovsky, line 80: Too wide lines and other coding style violations, e.g. everything in single line

logger/tfw_logger.cc
```diff
@@ -0,0 +1,687 @@
...
+
+	// Create directory if needed
+	fs::create_directories(config_path.parent_path());
+
+	// Create and save default configuration
+	TfwLoggerConfig default_config;
+	if (!default_config.save_to_file(config_path)) {
+		throw Except("Failed to create default configuration file");
+	}
+}
```
krizhanovsky, line 395: Why do we need this and `TfwLoggerConfig::save_to_file`? We can jsut create a default file and keep in the repo - it will be easier for users to learn from the default file, probably with comments, and we don't need extra C++ programming. At least we do this for Tempesta FW itself

logger/tfw_logger_config.cc
```diff
@@ -0,0 +1,148 @@
...
+	if (clickhouse_.host.empty()) {
+		throw std::runtime_error("ClickHouse host cannot be empty");
+	}
+	if (clickhouse_.port == 0) {
+		throw std::runtime_error("Invalid ClickHouse port");
+	}
+	if (clickhouse_.max_events == 0) {
+		throw std::runtime_error("max_events must be greater than 0");
+	}
+}
```
krizhanovsky, line 148: even for C++ we usually organize the code to read it from bottom to top, i.e. the calee `parse_from_ptree` must be above the caller `TfwLoggerConfig::load_from_file`

logger/test/data/invalid_config.json
```diff
@@ -0,0 +1,4 @@
+{
```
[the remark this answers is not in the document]
MorganaFuture: Tests have been moved to the tempesta-test repository.

logger/test/scripts/setup_test_env.sh
```diff
@@ -0,0 +1,184 @@
+#!/bin/bash
+#
+# Tempesta FW service script.
+#
+# Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
+# Copyright (C) 2015-2024 Tempesta Technologies, Inc.
```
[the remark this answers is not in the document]
MorganaFuture: Tests have been removed from this pr. Not valid anymore

logger/test/README.md
```diff
@@ -0,0 +1,112 @@
+# TFW Logger Tests
```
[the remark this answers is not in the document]
MorganaFuture: Also not valid anymore

logger/tfw_logger_config.cc
MorganaFuture: Fixed

logger/tfw_logger.cc
MorganaFuture: Accepted.

logger/test/Makefile
```diff
@@ -0,0 +1,71 @@
+#		Tempesta FW
+#
+# Copyright (C) 2024 Tempesta Technologies, Inc.
```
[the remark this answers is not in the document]
MorganaFuture: accepted

logger/Makefile
```diff
@@ -33,7 +33,7 @@ CXXFLAGS := -O3 -Wall -Wextra -std=c++23 -DPAGE_SIZE=$(PAGE_SIZE)
 CXXFLAGS += $(shell pkgconf -cflags spdlog)
 CXXFLAGS += -I $(CLICKHOUSE_DIR)
 CXXFLAGS += -I $(CLICKHOUSE_DIR)/contrib/absl
-SRCS := tfw_logger.cc mmap_buffer.cc clickhouse.cc
+SRCS := tfw_logger.cc mmap_buffer.cc clickhouse.cc tfw_logger_config.cc
 OBJS := $(SRCS:.cc=.o)
```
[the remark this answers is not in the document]
MorganaFuture: Accepted

logger/tfw_logger.cc
```diff
@@ -0,0 +1,687 @@
...
+#include <stdio.h>
+#include <unistd.h>
+
+#include <fstream>
+#include <thread>
+#include <iomanip>
+#include <iostream>
+#include <sstream>
+#include <atomic>
+#include <vector>
```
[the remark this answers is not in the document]
MorganaFuture: Done

logger/tfw_logger.cc
```diff
@@ -0,0 +1,687 @@
...
+	// Configuration overrides
+	std::optional<std::string> clickhouse_host;
+	std::optional<uint16_t> clickhouse_port;
+	std::optional<std::string> clickhouse_user;
+	std::optional<std::string> clickhouse_password;
+	std::optional<size_t> clickhouse_max_events;
+	std::optional<int> clickhouse_max_wait_ms;
+	std::optional<size_t> cpu_count;
+	std::optional<fs::path> log_path;
+};
```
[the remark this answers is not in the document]
MorganaFuture: Done

logger/tfw_logger.cc
MorganaFuture: Accepted

logger/tfw_logger_config.hh
MorganaFuture: Now should be fixed

scripts/tempesta.sh
```diff
@@ -126,22 +124,15 @@ templater()
...
+		if [ $(opt_exists "$line" "mmap"; echo $?) -ne 0 ] || [ $(opt_exists "$line" "logger_config"; echo $?) -ne 0 ]; then
 			tfw_logger_should_start=1
-			mmap_log=$(get_opt_value "$line" "mmap_log")
-			mmap_host=$(get_opt_value "$line" "mmap_host")
-			mmap_user=$(get_opt_value "$line" "mmap_user")
-			mmap_password=$(get_opt_value "$line" "mmap_password")
-
-			[[ -n "$mmap_log" && -n "$mmap_host" ]] ||
-				error "if mmaps enabled in access log, there have to be mmap_host and mmap_log options"
 		fi
```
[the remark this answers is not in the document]
MorganaFuture: Done

logger/tfw_logger.cc
```diff
@@ -0,0 +1,687 @@
...
+   }
+   
+   // Cleanup
+   if (fd >= 0) {
+   	close(fd);
+   	spdlog::info("Device closed");
+   }
+   
+   return res;
+}
```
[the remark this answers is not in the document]
const-t: @krizhanovsky No, we don't. clang-tidy only.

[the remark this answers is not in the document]
MorganaFuture: @const-t  @krizhanovsky  I have created [PR](https://github.com/tempesta-tech/tempesta/pull/2434) with clang format

scripts/tempesta.sh
```diff
@@ -126,22 +124,15 @@ templater()
...
 	done <<< "$opts"
 
-	cfg_content=$(remove_opts_by_mask "$cfg_content" "mmap_")
```
const-t, line 143: Why `remove_opts_by_mask()` isn't used? You need to remove `logger_config=/...` from `access_log` directive to not pass it to Tempesta's cfg parser.

scripts/tempesta.sh
```diff
@@ -311,36 +302,57 @@ start_tfw_logger()
...
+		echo "...starting tfw_logger with custom config: $config_path"
+	else
+		config_path="$tfw_logger_config"
+		echo "...starting tfw_logger with default config: $config_path"
 	fi
 
-	"$utils_path/tfw_logger" -H "$mmap_host" -l "$mmap_log" -u "$mmap_user" -p "$mmap_password" ||
+	# Check if config file exists
+	if [ ! -f "$config_path" ]; then
+		error "Configuration file not found: $config_path"
```
const-t: We printing error here, however continuing to load Tempesta. This is confusing. Instead we should print warning if this case is not mandatory for loading or break loading of Tempesta otherwise. Check it in other places as well. For instance if json file has error and TfwLogger failed to load - should we run Tempesta? I think - not.

utils/tfw_logger.cc
```diff
@@ -213,7 +248,7 @@ callback(const char *data, int size, void *private_data)
 
 	dbg_hexdump(data, size);
 
-	 while (size > (int)sizeof(TfwBinLogEvent)) {
+	while (size > (int)sizeof(TfwBinLogEvent)) {
```
const-t: I see that this changes from another commit, but let's fix it. Please replace C style casts with c++ style casts in the `callback()`.

utils/tfw_logger.cc
```diff
@@ -20,65 +20,101 @@
...
+	std::optional<std::string> clickhouse_user;
+	std::optional<std::string> clickhouse_password;
+	std::optional<size_t> clickhouse_max_events;
+	std::optional<int> clickhouse_max_wait_ms;
+	std::optional<size_t> cpu_count;
+	std::optional<fs::path> log_path;
+};
+
+// Forward declarations
+static void setup_signal_handlers();
```
const-t: I would prefer to not mix C and C++ approaches. Using `static` helpers isn't c++ approach, let's use advantages of C++ here. I suggesting to move all those helper functions to anonymous namespace in the beginning of the file and make them non-static.

utils/tfw_logger.cc
```diff
@@ -244,33 +277,47 @@ run_thread(const int ncpu, const int fd, const std::string &host,
...
 				CPU_SET(mbr.get_cpu_id(), &cpuset);
 
-				r = pthread_setaffinity_np(current_thread,
-							sizeof(cpu_set_t), &cpuset);
+				r = pthread_setaffinity_np(
+				    current_thread, sizeof(cpu_set_t), &cpuset);
 				if (r != 0)
-					throw Except("Affinity setting failed");
+					throw Except("Failed to set CPU "
+						     "affinity");
```
const-t: We can not break the string even if it wider than 80 symbols, just to make it available for grepping and improve readability. In this case let's not break the line.

utils/tfw_logger_config.cc
```diff
@@ -0,0 +1,120 @@
...
+	// Parse paths and settings
+	if (auto log_path_opt = tree.get_optional<std::string>("log_path")) {
+		log_path_ = *log_path_opt;
+	}
+
+	buffer_size_ = tree.get<size_t>("buffer_size", buffer_size_);
+	cpu_count_ = tree.get<size_t>("cpu_count", cpu_count_);
+
+	// Validate buffer size
+	if (buffer_size_ < 4096) {
```
const-t: Magic number, please make it constant with comment

utils/tfw_logger.cc
```diff
@@ -310,177 +358,338 @@ set_sig_handlers() noexcept
...
-						 vm["log"].as<std::string>());
-		spdlog::set_default_logger(logger);
-		spdlog::set_level(spdlog::level::info);
-		logger->flush_on(spdlog::level::info);
+static void run_main_loop(int fd)
+{
+	size_t cpu_count;
+
+	// Determine CPU count
+	if (config.get_cpu_count() > 0) {
```
const-t: Giving to user ability to set `cpu_count` looks bad. I see that this is not your code and it have been implemented in the first version of the logger, but we must fix it or at least test it, maybe in another PR.

The problem: Tempesta always allocates buffer for **each** cpu. After, during mmapping it chooses `cpu` where for which buffer is allocated relying on `offset` of mmap and does `mmap` only when `if (area_size == FULL_SIZE) { /* entire buffer is mapping */` - that means we can't mmap two or more buffers for one single thread. On other hand if we specify `cpu_count` less than machine really has cpus, some of buffers will not read at all, thus we lose data.

In current implementation we should always specify `cpu_count` equal to count of cpus.

utils/tfw_logger.cc
```diff
@@ -310,177 +358,338 @@ set_sig_handlers() noexcept
...
+	 * process continues to hold this lock, preventing subsequent
+	 * executions of tempesta.sh.
+	 *
+	 * Close all descriptors before daemonizing.
+	 */
+	closefrom(3);
+
+	// Daemonize if not in foreground mode
+	if (!opts.foreground) {
+		spdlog::info("Daemonizing...");
```
const-t:
```suggestion
		spdlog::debug("Daemonizing...");
```
Just to not see the log during starting tempesta.

const-t (changes requested):

scripts/tempesta.sh
MorganaFuture: Fixed. Added `remove_opts_by_mask()` call to remove `logger_config=` parameter from access_log directive before passing config to Tempesta parser.

scripts/tempesta.sh
MorganaFuture: Fixed. Now when tfw_logger is required but config file is missing/invalid, Tempesta startup fails with critical error instead of continuing.

utils/tfw_logger.cc
MorganaFuture: Fixed

utils/tfw_logger.cc
MorganaFuture: Fixed

utils/tfw_logger.cc
MorganaFuture: Fixed

utils/tfw_logger_config.cc
MorganaFuture: Fixed

utils/tfw_logger.cc
MorganaFuture: Added validation warnings for cpu_count mismatches to educate users about potential data loss. Doesn't break existing functionality - let's fix this in another pull request
What do you think?

utils/tfw_logger.cc
MorganaFuture: Fixed

utils/tfw_logger.cc
const-t: Actually we can drop `cpu_count` from config, it's few lines of code to add it in future, but it will not confusing us and users now.

scripts/tempesta.sh
```diff
@@ -311,36 +303,72 @@ start_tfw_logger()
...
+	if [ ! -f "$config_path" ]; then
+		sysctl -e -w net.tempesta.state=stop 2>/dev/null || true
+		unload_modules
+		tfw_irqbalance_revert
+		error "TFW Logger configuration file not found: $config_path. Cannot start Tempesta with access_log mmap enabled."
 	fi
 
-	"$utils_path/tfw_logger" -H "$mmap_host" -l "$mmap_log" -u "$mmap_user" -p "$mmap_password" ||
+	# Test config file validity before proceeding
+	if ! "$logger_path/tfw_logger" --config="$config_path" --test-config 2>/dev/null; then
```
const-t: Why do we run tfw_logger twice for testing config? We do parse and validate config on each run, if config is invalid tfw_logger exits with error. Seems the first run with `--test-config` is unnecessary.

utils/tfw_logger.cc
MorganaFuture: Removed cpu_count from config

scripts/tempesta.sh
MorganaFuture: Fixed. Removed --test-config

const-t (approved):
LGTM

scripts/tempesta.sh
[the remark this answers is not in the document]
krizhanovsky: Also please document `mmap_log_buffer_size` in the wiki page.

By the way in #2363 I see the changes related to `mmap_log_buffer_size`, the subject for #2313, but this PR seems a pure refactoring and I don't see how does it fix the bug.

utils/tfw_logger.cc
```diff
@@ -20,65 +20,96 @@
...
+
+	std::optional<fs::path> config_path;
+	std::optional<std::string> clickhouse_host;
+	std::optional<uint16_t> clickhouse_port;
+	std::optional<std::string> clickhouse_table;
+	std::optional<std::string> clickhouse_user;
+	std::optional<std::string> clickhouse_password;
+	std::optional<size_t> clickhouse_max_events;
+	std::optional<int> clickhouse_max_wait_ms;
+	std::optional<fs::path> log_path;
```
krizhanovsky: Align structure members for better readability

```suggestion
        bool                            help = false;
        bool                            stop_daemon = false;
        bool                            foreground = false;
        bool                            test_config = false;

        std::optional<fs::path>         config_path;
        std::optional<std::string>      clickhouse_host;
        std::optional<uint16_t>         clickhouse_port;
        std::optional<std::string>      clickhouse_table;
        std::optional<std::string>      clickhouse_user;
        std::optional<std::string>      clickhouse_password;
        std::optional<size_t>           clickhouse_max_events;
        std::optional<int>              clickhouse_max_wait_ms;
        std::optional<fs::path>         log_path;

```

utils/tfw_logger.cc
```diff
@@ -310,177 +359,346 @@ set_sig_handlers() noexcept
...
-		throw Except("No PID file found. Is the daemon running?");
+	// Extract option values
+	if (vm.count("config"))
+		result.config_path = vm["config"].as<fs::path>();
+	if (vm.count("host"))
+		result.clickhouse_host = vm["host"].as<std::string>();
+	if (vm.count("port"))
+		result.clickhouse_port = vm["port"].as<uint16_t>();
+	if (vm.count("table"))
+		result.clickhouse_table = vm["table"].as<std::string>();
```
krizhanovsky, line 420: Please add `table` description to https://github.com/tempesta-tech/tempesta/wiki/TFW-Logger#configuration-options

utils/tfw_logger.cc
```diff
@@ -310,177 +359,346 @@ set_sig_handlers() noexcept
...
-				goto end;
-			if (errno != ENOENT)
-				throw Except("Can't open device");
-			std::this_thread::sleep_for(wait_for_dev);
-		}
-
-		for (i = 0; i < cpu_cnt; ++i) {
-			std::packaged_task<void(int, int, std::string, std::string,
-						std::string)> task(run_thread);
-			futures.push_back(task.get_future());
```
krizhanovsky, line 454: `run_thread()` isn't `noexcept` and it doesn't `catch (...)`. Can `spdlog::debug()` throw? In other words, if you get rid of futures, you need to make sure that `run_thread()` does not neither nor nor in future throw.

utils/tfw_logger.cc
```diff
@@ -310,177 +359,346 @@ set_sig_handlers() noexcept
...
+	}
+
+	spdlog::info("Successfully opened device: {}", dev_path);
+	return fd;
+}
+
+void
+run_main_loop(int fd)
 {
+	auto cpu_count = static_cast<size_t>(sysconf(_SC_NPROCESSORS_ONLN));
```
krizhanovsky, line 556: Please comment why don't we use C++ routine - it's a subtle and interesting thing

utils/tfw_logger.cc
```diff
@@ -310,177 +359,346 @@ set_sig_handlers() noexcept
...
-	std::vector<std::future<void>> futures;
-	unsigned int i;
-	long int cpu_cnt;
-	int fd = -1, res = 0;
-	bool stop = false;
-	std::shared_ptr<spdlog::logger> logger = nullptr;
+	for (size_t i = 0; i < cpu_count; ++i) {
+		threads.emplace_back(
+		    run_thread, static_cast<int>(i), fd, std::ref(config));
+	}
```
krizhanovsky: usually we don't use parentheses around single line statement

utils/tfw_logger.cc
```diff
@@ -310,177 +343,317 @@ set_sig_handlers() noexcept
...
-	catch (const std::exception &e) {
-		log_error(e.what(), logger != nullptr, true);
-		res = 2;
-	}
 
-	for (auto &t : threads)
-		t.join();
+	spdlog::info("All {} worker threads started", cpu_count);
+
+	// Wait for all threads to complete
```
krizhanovsky: somewhat abvious comment

utils/tfw_logger_config.cc
```diff
@@ -0,0 +1,149 @@
...
+	// Validate ClickHouse settings
+	if (clickhouse_.host.empty()) {
+		throw std::runtime_error("ClickHouse host cannot be empty");
+	}
+	if (clickhouse_.port == 0) {
+		throw std::runtime_error("Invalid ClickHouse port");
+	}
+	if (clickhouse_.max_events == 0) {
+		throw std::runtime_error("max_events must be greater than 0");
+	}
```
krizhanovsky: Parentheses around single line statements

utils/tfw_logger.cc
```diff
@@ -502,131 +405,281 @@ parse_all_options(int argc, char *argv[])
...
+static void
+cleanup_resources(int fd, int pidfile_fd)
+{
+	// Close device
+	if (fd >= 0) {
+		close(fd);
+		spdlog::info("Device closed");
+	}
+
+	// Remove PID file
```
krizhanovsky: The comments are obvious, basically just repeating the functions names

krizhanovsky (changes requested):
The PR looks good except several minor and quick to fix things.

However, I didn't see the fixes for #2313 and #2314. I believe there is no fix for the first one. I'm not sure about the second one. If it is fixed, then please describe somewhere what was the reason for the absent logging and how it is fixed. There are a lot of changes and I don't see how it is fixed.

Please make separate new commits for further fixes. This way it will be easier to review the changes, but you'll also be able to squash commits before merge

utils/tfw_logger.cc
MorganaFuture: Removed

utils/tfw_logger.cc
MorganaFuture: Fixed

utils/tfw_logger.cc
MorganaFuture: Removed

utils/tfw_logger_config.cc
MorganaFuture: Fixed

utils/tfw_logger.cc
MorganaFuture: Added

utils/tfw_logger.cc
MorganaFuture: Added `noexcept` to `run_thread()` and wrapped everything in try`-catch(...)` to prevent std::terminate

utils/tfw_logger.cc
MorganaFuture: Acceptde

utils/tfw_logger.cc
MorganaFuture: Added

scripts/tempesta.sh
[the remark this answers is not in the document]
MorganaFuture: I'll add the fix for [#2313](https://github.com/tempesta-tech/tempesta/issues/2313) in the next PR together with the renaming of utils to logger. This is to make the review of these parts easier. The PR will be small, and I’ll open it right after this one is merged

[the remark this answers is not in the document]
MorganaFuture: Do you mean creating a new wiki page for mmap_log_buffer_size?

utils/tfw_logger.cc
krizhanovsky: In code we have `table`, but in wiki `table_name` - typo?

utils/tfw_logger.cc
```diff
@@ -233,67 +268,76 @@ callback(const char *data, int size, void *private_data)
...
-			log_error(e.what(), true, false);
-			/* All the manually thrown exceptions are critical */
-			break;
-		}
-		catch (const std::exception &e) {
-			if (!uncritical_error) {
-				log_error(e.what(), true, true);
-				timeout = reconnect_min_timeout;
-				uncritical_error = true;
+	try {
```
krizhanovsky, line 281: Minus one level of nesting
```suggestion
run_thread(const int ncpu, const int fd, const TfwLoggerConfig &config) noexcept
try {
} catch (...) {
}
```

utils/tfw_logger.cc
```diff
@@ -233,67 +268,76 @@ callback(const char *data, int size, void *private_data)
...
-
-			std::this_thread::sleep_for(timeout);
-
-			if (timeout < reconnect_max_timeout)
-				timeout *= 2;
 		}
+		spdlog::debug("Worker {} stopped", ncpu);
+	} catch (...) {
+		// Never let exceptions escape from thread function
+		// This prevents std::terminate
```
krizhanovsky: It's always better to log anything than just silently ignore an error (exception). Also probably with some logged message you'll need less commenting

scripts/tempesta.sh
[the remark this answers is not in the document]
krizhanovsky: I mean add the option description to https://github.com/tempesta-tech/tempesta/wiki/TFW-Logger or https://github.com/tempesta-tech/tempesta/wiki/Handling-clients

utils/pidfile.cc
```diff
@@ -0,0 +1,221 @@
...
+{
+	int fd = -1;
+	struct flock fl;
+	char pid_str[64] = {0};
+	int pid_str_len;
+	mode_t mask;
+
+	// Get PID as string
+	pid_str_len = snprintf(
+	    pid_str, sizeof(pid_str), "%ld", static_cast<long>(getpid()));
```
krizhanovsky:
```suggestion
        pid_str_len = snprintf(pid_str, sizeof(pid_str), "%ld",
                               static_cast<long>(getpid()));
```

utils/tfw_logger.cc
```diff
@@ -233,67 +268,76 @@ callback(const char *data, int size, void *private_data)
...
+		
+		while (!stop_flag) {
+			try {
+				const auto &ch_cfg = config.get_clickhouse();
+				spdlog::debug("Worker {} connecting to ClickHouse "
+					      "at {}:{}, table: {}",
+					      ncpu,
+					      ch_cfg.host,
+					      ch_cfg.port,
+					      ch_cfg.table_name);
```
krizhanovsky: There is no need to place every parameter on a single line. 
```suggestion
                                              ncpu, ch_cfg.host, ch_cfg.port,
                                              ch_cfg.table_name);
```

Also please use `try {} catch () {}` instead of `{}` for the function, reduce the nesting and adjust the code. The point is in reducing nesting, now writing 10 characters lines of code.

Please adjust the rest of the function appropriately

utils/tfw_logger.cc
```diff
@@ -284,42 +282,42 @@ run_thread(const int ncpu, const int fd, const TfwLoggerConfig &config)
...
+			    ch_cfg.host,
+			    ch_cfg.table_name,
+			    ch_cfg.user ? *ch_cfg.user : "",
+			    ch_cfg.password ? *ch_cfg.password : "",
+			    make_block());
 
-			TfwMmapBufferReader mbr(ncpu, fd, &clickhouse,
-						callback);
+			TfwMmapBufferReader mbr(
+			    ncpu, fd, &clickhouse, callback);
```
krizhanovsky: senseless change breaking the usual way for lines breaking

utils/tfw_logger.cc
```diff
@@ -284,42 +282,42 @@ run_thread(const int ncpu, const int fd, const TfwLoggerConfig &config)
...
 
 			if (!affinity_is_set) {
 				CPU_ZERO(&cpuset);
 				CPU_SET(mbr.get_cpu_id(), &cpuset);
 
-				r = pthread_setaffinity_np(current_thread,
-							   sizeof(cpu_set_t),
-							   &cpuset);
+				r = pthread_setaffinity_np(
+				    current_thread, sizeof(cpu_set_t), &cpuset);
```
krizhanovsky: the same as above

utils/tfw_logger.cc
```diff
@@ -284,42 +282,42 @@ run_thread(const int ncpu, const int fd, const TfwLoggerConfig &config)
...
 				if (r != 0)
-					throw Except("Failed to set CPU "
-						     "affinity");
+					throw Except("Failed to set CPU affinity");
 
 				affinity_is_set = true;
 				spdlog::debug("Worker {} bound to CPU {}",
-					      ncpu, mbr.get_cpu_id());
+					      ncpu,
+					      mbr.get_cpu_id());
```
krizhanovsky: the same as above

utils/tfw_logger.cc
```diff
@@ -373,42 +370,47 @@ parse_command_line(int argc, char *argv[])
...
+	    "ClickHouse password (overrides config)")(
+	    "max-events",
+	    po::value<size_t>(),
+	    "Maximum events before commit (overrides config)")(
+	    "max-wait",
+	    po::value<int>(),
+	    "Maximum wait time in ms before commit "
+	    "(overrides config)")("log-path,l",
+				  po::value<fs::path>(),
+				  "Path to log file (overrides config)");
```
krizhanovsky: broken identation

krizhanovsky (approved):
I approve the changes, but there are still a lot of, new, coding style violations.

@const-t  may I ask you to finalize the reviewing process to get the code aligned with our style? Also I doubt that we should port these changes and #2313 and #2314 to 0.8 - there are huge changes w/or fixing the original issues - too much for a backport.

utils/tfw_logger.cc
MorganaFuture: Accepted

utils/tfw_logger.cc
MorganaFuture: Accepted

utils/pidfile.cc
MorganaFuture: Accepted

utils/tfw_logger.cc
MorganaFuture: Fixed

utils/tfw_logger.cc
MorganaFuture: Fixed

utils/tfw_logger.cc
MorganaFuture: Fixed

utils/tfw_logger.cc
MorganaFuture: Fixed

utils/tfw_logger.cc
MorganaFuture: Fixed try {} catch

MorganaFuture:
@krizhanovsky  This PR fixes [#2313](https://github.com/tempesta-tech/tempesta/issues/2313)

utils/tfw_logger.cc
MorganaFuture: Changed table_name to table in wiki

scripts/tempesta.sh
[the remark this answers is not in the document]
krizhanovsky: @MorganaFuture this comment is still not addressed

utils/tfw_logger.cc
```diff
@@ -20,65 +20,96 @@
...
+
+	std::optional<fs::path>     config_path;
+	std::optional<std::string>  clickhouse_host;
+	std::optional<uint16_t>     clickhouse_port;
+	std::optional<std::string>  clickhouse_table;
+	std::optional<std::string>  clickhouse_user;
+	std::optional<std::string>  clickhouse_password;
+	std::optional<size_t>       clickhouse_max_events;
+	std::optional<int>          clickhouse_max_wait_ms;
+	std::optional<fs::path>     log_path;
```
krizhanovsky: We use tabs and spaces for alignment for less than 8 characters.

I pushed a small commit to fix coding style, but I did not compile it - please make sure that there is no problems.

utils/tfw_logger.cc
```diff
@@ -310,177 +347,324 @@ set_sig_handlers() noexcept
...
 
-	std::cout << "Daemon stopped." << std::endl;
+void
+initialize_logging()
+try {
+	// Create log directory if needed
+	fs::create_directories(fs::path(config.get_log_path()).parent_path());
+
+	auto logger = spdlog::basic_logger_mt(
+	    "access_logger", config.get_log_path().string());
```
const-t:
```suggestion
		auto logger = spdlog::basic_logger_mt("access_logger",
					               config.get_log_path().string());
```

utils/tfw_logger.cc
```diff
@@ -310,177 +347,324 @@ set_sig_handlers() noexcept
...
+		throw Except("Cannot determine CPU count");
+
+	spdlog::info("Using {} CPU(s)", cpu_count);
+	spdlog::info("Starting {} worker threads", cpu_count);
+
+	// Start worker threads
+	std::vector<std::thread> threads;
+	for (size_t i = 0; i < cpu_count; ++i)
+		threads.emplace_back(
+		    run_thread, static_cast<int>(i), fd, std::ref(config));
```
const-t:
```suggestion
		threads.emplace_back(run_thread, static_cast<int>(i), fd,
				     std::ref(config));
```

utils/tfw_logger.cc
MorganaFuture: Accepted

utils/tfw_logger.cc
MorganaFuture: Accepted

utils/tfw_logger_config.cc
```diff
@@ -0,0 +1,116 @@
...
+	// Validate ClickHouse settings
+	if (clickhouse_.host.empty())
+		throw std::runtime_error("ClickHouse host cannot be empty");
+
+	if (clickhouse_.port == 0)
+		throw std::runtime_error("Invalid ClickHouse port");
+
+	if (clickhouse_.table_name.empty())
+		throw std::runtime_error(
+		    "ClickHouse table name cannot be empty");
```
const-t:
```suggestion
		throw std::runtime_error("ClickHouse table name cannot be empty");
```

utils/tfw_logger.cc
```diff
@@ -145,47 +177,51 @@ read_access_log_event(const char *data, int size, TfwClickhouse *clickhouse)
...
+	} else                                                                 \
+		(*block)[ind]->As<col_type>()->Append(0);
 
 	READ_INT(TFW_MMAP_LOG_ADDR, clickhouse::ColumnIPv6, struct in6_addr);
 	READ_INT(TFW_MMAP_LOG_METHOD, clickhouse::ColumnUInt8, unsigned char);
 	READ_INT(TFW_MMAP_LOG_VERSION, clickhouse::ColumnUInt8, unsigned char);
 	READ_INT(TFW_MMAP_LOG_STATUS, clickhouse::ColumnUInt16, uint16_t);
-	READ_INT(TFW_MMAP_LOG_RESP_CONT_LEN, clickhouse::ColumnUInt32, uint32_t);
+	READ_INT(
+	    TFW_MMAP_LOG_RESP_CONT_LEN, clickhouse::ColumnUInt32, uint32_t);
```
const-t:
```suggestion
	READ_INT(TFW_MMAP_LOG_RESP_CONT_LEN,
		 clickhouse::ColumnUInt32, uint32_t);
```

MorganaFuture: Fixed

utils/tfw_logger_config.cc
MorganaFuture: Fixed

scripts/tempesta.sh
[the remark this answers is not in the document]
MorganaFuture: Updated wiki https://github.com/tempesta-tech/tempesta/wiki/Handling-clients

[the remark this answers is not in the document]
MorganaFuture: Updated wiki

PR #2430, opened 2025-05-27
Fix #2361: avoid kernel warning on empty URI in tfw_cache_copy_resp()

### LINKED ISSUE #2361: Kernel warning on the empty uri
Run test from https://github.com/tempesta-tech/tempesta-test/pull/785/commits/da9d7a1cb3d8830c596956364518e70204512f30 causes kernel warning in dmesg, we should fix this problem

WitalyAnisimov:
This patch fixes a kernel warning triggered by malformed or empty URI values 
when Tempesta tries to cache HTTP responses.

Problem:
- When the URI was empty or malformed, `tfw_cache_copy_resp()` and `tfw_cache_entry_key_eq()` 
  could dereference invalid memory or overrun the TDB chunk boundaries.
- This resulted in WARN_ON() or even BUG() being hit inside the kernel, crashing Tempesta FW.

Solution:
- Improved chunk iteration logic and added proper bounds checking in both functions.
- Now Tempesta gracefully handles empty URI or Host headers without triggering kernel warnings.

Validation:
- Kernel no longer logs WARN or BUG when processing such requests.
- Verified via `dmesg` after running Tempesta with empty or malformed authority requests.

Closes: #2361

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -2322,7 +2322,6 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 		p = TDB_PTR(db->hdr, ce->hdrs_304[i]);
 		while (trec && (p + TFW_CSTR_HDRLEN > trec->data + trec->len))
 			trec = tdb_next_rec_chunk(db, trec);
-
 		BUG_ON(!trec);
 
 		ce->hdrs_304[i] = TDB_OFF(db->hdr, p);
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -2271,8 +2271,10 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 		return -ENOMEM;
 	}
 
-	if (WARN_ON_ONCE(tot_len != 0))
+	if (WARN_ON_ONCE(tot_len != 0)) {
+		T_ERR("Cache copy: leftover body data not copied, tot_len = %lu\n", tot_len);
 		return -EINVAL;
+	}
 
 	ce->version = resp->version;
 	tfw_http_copy_flags(ce->hmflags, resp->flags);
@@ -2320,9 +2322,14 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 			continue;
 
 		p = TDB_PTR(db->hdr, ce->hdrs_304[i]);
+
 		while (trec && (p + TFW_CSTR_HDRLEN > trec->data + trec->len))
 			trec = tdb_next_rec_chunk(db, trec);
-		BUG_ON(!trec);
+
+		if (unlikely(!trec)) {
+			T_ERR("Cache copy: failed to update HTTP header\n");
+			return -EFAULT;
+		}
 
 		ce->hdrs_304[i] = TDB_OFF(db->hdr, p);
 	}
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -2320,7 +2320,6 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 			continue;
 
 		p = TDB_PTR(db->hdr, ce->hdrs_304[i]);
-
 		while (trec && (p + TFW_CSTR_HDRLEN > trec->data + trec->len))
 			trec = tdb_next_rec_chunk(db, trec);
 
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -2266,9 +2266,13 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 					   resp, &tot_len);
 	}
 
-	if (unlikely(tot_len != 0))
-		pr_warn_once("leftover body data not copied, tot_len = %lu\n", tot_len);
+	if (unlikely(r)) {
+		T_ERR("Cache: cannot copy HTTP body\n");
+		return -ENOMEM;
+	}
 
+	if (WARN_ON_ONCE(tot_len != 0))
+		return -EINVAL;
 
 	ce->version = resp->version;
 	tfw_http_copy_flags(ce->hmflags, resp->flags);
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -2266,15 +2266,9 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 					   resp, &tot_len);
 	}
 
-	if (unlikely(r)) {
-		T_ERR("Cache: cannot copy HTTP body\n");
-		return -ENOMEM;
-	}
+	if (unlikely(tot_len != 0))
+		pr_warn_once("leftover body data not copied, tot_len = %lu\n", tot_len);
 
-	if (WARN_ON_ONCE(tot_len != 0)) {
-		T_ERR("Cache copy: leftover body data not copied, tot_len = %lu\n", tot_len);
-		return -EINVAL;
-	}
 
 	ce->version = resp->version;
 	tfw_http_copy_flags(ce->hmflags, resp->flags);
@@ -2326,10 +2320,7 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 		while (trec && (p + TFW_CSTR_HDRLEN > trec->data + trec->len))
 			trec = tdb_next_rec_chunk(db, trec);
 
-		if (unlikely(!trec)) {
-			T_ERR("Cache copy: failed to update HTTP header\n");
-			return -EFAULT;
-		}
+		BUG_ON(!trec);
 
 		ce->hdrs_304[i] = TDB_OFF(db->hdr, p);
 	}
--- a/fw/http.c
+++ b/fw/http.c
@@ -3680,7 +3680,6 @@ tfw_h1_req_copy_first_line(TfwHttpReq *req)
 	int r;
 	const TfwStr *meth;
 	static const DEFINE_TFW_STR(meth_get, "GET");
-	static const DEFINE_TFW_STR(slash, "/");
 	static const DEFINE_TFW_STR(sp, " ");
 	static const DEFINE_TFW_STR(ver, " " S_VERSION11 S_CRLF);
 
@@ -3699,13 +3698,7 @@ tfw_h1_req_copy_first_line(TfwHttpReq *req)
 	if (unlikely(r))
 		return r;
 
-	/* uri_path is empty when uri is absolute and doesn't have slash */
-	if (TFW_STR_EMPTY(&req->uri_path))
-		r = tfw_http_msg_expand_from_pool(hm, &slash);
-	else
-		r = tfw_http_msg_expand_from_pool(hm, &req->uri_path);
-	if (unlikely(r))
-		return r;
+	r = tfw_http_msg_expand_from_pool(hm, &req->uri_path);
 
 	return tfw_http_msg_expand_from_pool(hm, &ver);
 }
--- a/fw/http.c
+++ b/fw/http.c
@@ -3699,6 +3699,8 @@ tfw_h1_req_copy_first_line(TfwHttpReq *req)
 		return r;
 
 	r = tfw_http_msg_expand_from_pool(hm, &req->uri_path);
+	if (unlikely(r))
+		return r;
 
 	return tfw_http_msg_expand_from_pool(hm, &ver);
 }
--- a/fw/http_parser.c
+++ b/fw/http_parser.c
@@ -5938,6 +5938,8 @@ Req_Method_1CharStep: __attribute__((cold))
 			__FSM_MOVE_f(Req_UriAbsPath, &req->uri_path);
 		}
 		else if (c == ' ') {
+			/* Absolute URI without path -> set uri_path = "/" */
+			req->uri_path = TFW_STR_F_STRING("/", TFW_STR_COMPLETE);
 			__FSM_MOVE_nofixup(Req_HttpVer);
 		}
 		TFW_PARSER_DROP(Req_UriMarkEnd);
@@ -6057,6 +6059,8 @@ Req_Method_1CharStep: __attribute__((cold))
 			__FSM_MOVE_f(Req_UriAbsPath, &req->uri_path);
 		}
 		else if (c == ' ') {
+			/* Absolute URI without path -> set uri_path = "/" */
+			req->uri_path = TFW_STR_F_STRING("/", TFW_STR_COMPLETE);
 			__FSM_MOVE_nofixup(Req_HttpVer);
 		}
 		TFW_PARSER_DROP(Req_UriAuthorityEnd);
--- a/fw/http_parser.c
+++ b/fw/http_parser.c
@@ -5924,19 +5924,20 @@ Req_Method_1CharStep: __attribute__((cold))
 		__FSM_JMP(Req_UriAbsoluteForm);
 	}
 
-	__FSM_STATE(Req_UriRareFormsEnd, hot) {
+	__FSM_STATE(Req_UriRareFormsEnd, hot)
+	{
 		if (likely(c == '/')) {
 			__msg_field_open(&req->uri_path, p);
 			__FSM_MOVE_f(Req_UriAbsPath, &req->uri_path);
 		}
 		else if (c == ' ') {
-			if (unlikely(req->method == TFW_HTTP_METH_OPTIONS)) {
+			if (req->method == TFW_HTTP_METH_OPTIONS)
 				req->uri_path = TFW_STR_F_STRING("*", TFW_STR_COMPLETE);
-			} else {
+			else
 				req->uri_path = slash;
-			}
 			__FSM_MOVE_nofixup(Req_HttpVer);
 		}
+		TFW_PARSER_DROP(req_UriMarkEnd);
 	}
 
 	__FSM_STATE(Req_UriAbsoluteForm, cold) {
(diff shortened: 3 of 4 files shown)


fw/cache.c
```diff
@@ -1267,16 +1267,26 @@ tfw_cache_entry_key_eq(TDB *db, TfwHttpReq *req, TfwCacheEntry *ce)
 
 	t_off = CE_BODY_SIZE;
 	TFW_CACHE_REQ_KEYITER(c, &req->uri_path, &req->host, u_end, h_start,
-			      h_end, u_fin, h_fin)
+			h_end, u_fin, h_fin)
```
krizhanovsky: broken identation

fw/cache.c
```diff
@@ -2271,8 +2281,15 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 		return -ENOMEM;
 	}
 
-	if (WARN_ON_ONCE(tot_len != 0))
+	/* FIX: safer tot_len check */
```
krizhanovsky: Usually we mark such comments as `FIXME` and we either need to fix the things in this PR or create a new issue and point it out in the comment

fw/cache.c
```diff
@@ -2271,8 +2281,15 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
...
+	if (WARN_ON_ONCE(tot_len < 0)) {
+		T_ERR("Cache copy: negative leftover length = %ld\n", (long)tot_len);
+		return -EINVAL;
+	}
+	if (WARN_ON_ONCE(tot_len != 0)) {
+		T_ERR("Cache copy: unexpected leftover length = %lu\n", tot_len);
 		return -EINVAL;
+	}
```
krizhanovsky: These checks are executed on the hot data path, so do we need both of them just to print different messages? Can't we just print the value of `tot_len`?

fw/cache.c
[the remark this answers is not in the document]
krizhanovsky: There is a comment for `TFW_CACHE_REQ_KEYITER()`:

> uri_path and host are not expected to be empty, because we check
> it previosly but we should not crash.

and the macro pints warnings on empty URI and Host. Meantime, there are parts of code, e.g. [referenced by Konstantin](https://github.com/tempesta-tech/tempesta/blob/master/fw/http.c#L3703), which do allow empty URI. @const-t why do you think that it's bad to adjust the macro to process empty strings? At the moment the macro is used in two place: one of them isn't addressed by this patch and another introduces a bunch of checks making the loop harder to read and slower. It seems there is some code inconsistency - whether we allow or not empty URI and Host. 

By the way, don't we see the warnings from `TFW_CACHE_REQ_KEYITER()` for the test case https://github.com/tempesta-tech/tempesta-test/commit/da9d7a1cb3d8830c596956364518e70204512f30 ?

I remember that Host can be empty, but why do we allow empty URL? We have a security option to prohibit empty Host, but nothing about URL. @EvgeniiMekhanik why don't we block empty URL? Are there any legit use cases?

fw/cache.c
```diff
@@ -1267,16 +1267,26 @@ tfw_cache_entry_key_eq(TDB *db, TfwHttpReq *req, TfwCacheEntry *ce)
 
 	t_off = CE_BODY_SIZE;
 	TFW_CACHE_REQ_KEYITER(c, &req->uri_path, &req->host, u_end, h_start,
-			      h_end, u_fin, h_fin)
+			h_end, u_fin, h_fin)
 	{
-		if (!trec)
+		if (!trec || t_off >= trec->len)
```
krizhanovsky: We already have `if (t_off >= trec->len)`  at the below

fw/cache.c
```diff
@@ -2320,9 +2337,15 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 			continue;
 
 		p = TDB_PTR(db->hdr, ce->hdrs_304[i]);
-		while (trec && (p + TFW_CSTR_HDRLEN > trec->data + trec->len))
+
+		while (trec && (p + TFW_CSTR_HDRLEN > trec->data + trec->len)) {
 			trec = tdb_next_rec_chunk(db, trec);
-		BUG_ON(!trec);
+		}
```
krizhanovsky: According to https://github.com/tempesta-tech/tempesta/blob/master/CodingStyle we don't use braces around single-line statement

fw/cache.c
```diff
@@ -2320,9 +2337,15 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
...
+
+		if (unlikely(!trec)) {
+			T_ERR("Cache copy: ran out of chunks while updating hdrs_304[%d]\n", i);
```
krizhanovsky: `T_ERR()` prints messages for the end-user - such messages must not contain variable names, which are cryptic for users, use only human friendly text. Bebug messages are file to have developer-only messaging

fw/cache.c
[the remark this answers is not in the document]
EvgeniiMekhanik: This test still fails because of WARNING. And I don't see any legal cases when we need empty URL, moreover according chatGPT this is not valid a case and such request is malformed.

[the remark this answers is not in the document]
EvgeniiMekhanik: So I think we should block such request at the same place where we check that host should not be empty

EvgeniiMekhanik (changes requested):
I think we should block request with empty URI

fw/http.c
```diff
@@ -3699,13 +3698,7 @@ tfw_h1_req_copy_first_line(TfwHttpReq *req)
...
 		return r;
 
-	/* uri_path is empty when uri is absolute and doesn't have slash */
-	if (TFW_STR_EMPTY(&req->uri_path))
-		r = tfw_http_msg_expand_from_pool(hm, &slash);
-	else
-		r = tfw_http_msg_expand_from_pool(hm, &req->uri_path);
-	if (unlikely(r))
-		return r;
+	r = tfw_http_msg_expand_from_pool(hm, &req->uri_path);
```
EvgeniiMekhanik, line 3701: need to check 'r'
if (unlikely(r))
    return r;

fw/cache.c
EvgeniiMekhanik: Why we decide this place? We decide to remove WARN_ON_ONCE and BUG_ON below?

fw/http_parser.c
```diff
@@ -6081,6 +6091,8 @@ Req_Method_1CharStep: __attribute__((cold))
 			__FSM_MOVE_f(Req_UriAbsPath, &req->uri_path);
 		}
 		else if (c == ' ') {
+			/* Absolute URI without path → set uri_path = "/" */
+			req->uri_path = tfw_str_slash;
```
EvgeniiMekhanik: We can write req->uri_path = TFW_STR_F_STRING("/", TFW_STR_COMPLETE); and remove `tfw_str_slash`

EvgeniiMekhanik:
```
#define TEST_FULL_REQ(req_host, req_uri_path)				\
	FOR_REQ("GET http://" req_host req_uri_path " HTTP/1.1\r\n\r\n")\
	{								\
		EXPECT_TFWSTR_EQ(&req->host, req_host);			\
		if (SLEN(req_uri_path)) {				\
			EXPECT_TFWSTR_EQ(&req->uri_path, req_uri_path);	\
		} else {						\
			/*						\
			 * If request URI is empty Tempesta FW set	\
			 * default req->uri_path "/".			\
			 */						\
			EXPECT_TFWSTR_EQ(&req->uri_path, "/");		\
		}							\
	}
```
to fix TEST(http1_parser, parses_req_uri)

EvgeniiMekhanik:
Look at the lobel `__FSM_STATE(Req_UriAuthorityEnd, cold) {` near the place which was fixed.
We should also set URI to "/" here
```
else if (c == ' ') {
			/* Absolute URI without path → set uri_path = "/" */
			req->uri_path = TFW_STR_F_STRING("/", TFW_STR_COMPLETE);
			__FSM_MOVE_nofixup(Req_HttpVer);
		}
```

fw/cache.c
```diff
@@ -2320,9 +2322,14 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 			continue;
 
 		p = TDB_PTR(db->hdr, ce->hdrs_304[i]);
+
 		while (trec && (p + TFW_CSTR_HDRLEN > trec->data + trec->len))
 			trec = tdb_next_rec_chunk(db, trec);
-		BUG_ON(!trec);
+
+		if (unlikely(!trec)) {
```
const-t: Why `BUG_ON` replaced by `if`. Do you have tests that can reproduce this `BUG_ON`, which approves that we need to change this logic?

fw/http_parser.c
const-t: From my point of view if we will use `tfw_str_slash` in two or more places having single static `tfw_str_slash` looks good.

fw/cache.c
```diff
@@ -2266,13 +2266,9 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
...
 	}
 
-	if (unlikely(r)) {
-		T_ERR("Cache: cannot copy HTTP body\n");
-		return -ENOMEM;
-	}
+	if (unlikely(tot_len != 0))
+		pr_warn_once("leftover body data not copied, tot_len = %lu\n", tot_len);
 
-	if (WARN_ON_ONCE(tot_len != 0))
```
EvgeniiMekhanik: Why this part of code is changed? It is unrelated changes to the task 2361. It was discussed with @krizhanovsky or @const-t ?

fw/http_parser.c
EvgeniiMekhanik: we can name it `slash` not `tfw_str_slash` and we should initialize it using marco `TFW_STR_F_STRING("/", TFW_STR_COMPLETE);`

fw/http_parser.c
```diff
@@ -5928,6 +5938,8 @@ Req_Method_1CharStep: __attribute__((cold))
 			__FSM_MOVE_f(Req_UriAbsPath, &req->uri_path);
 		}
 		else if (c == ' ') {
+			/* Absolute URI without path -> set uri_path = "/" */
+			req->uri_path = TFW_STR_F_STRING("/", TFW_STR_COMPLETE);
```
EvgeniiMekhanik: It is mistake. It seems that we should set req->uri_path == '*' not '/' here. See RFC7230 5.3.4. @const-t please aloso pay attention on this place

EvgeniiMekhanik: We should also write unit test for this case and new python test to cover it.

EvgeniiMekhanik: In new RFC https://www.rfc-editor.org/rfc/rfc9112.html see please [3.2.4. ](https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2.4)[asterisk-form](https://www.rfc-editor.org/rfc/rfc9112.html#name-asterisk-form)

fw/t/unit/test_http1_parser.c
```diff
@@ -206,7 +206,28 @@ TEST(http1_parser, parses_req_uri)
...
+	FOR_REQ("OPTIONS * HTTP/1.1\r\n\r\n") {			\
+		EXPECT_EQ(req->method, TFW_HTTP_METH_OPTIONS);		\
+		EXPECT_TFWSTR_EQ(&req->uri_path, "*");			\
+	}
+
+#define TEST_OPTIONS_WITHOUT_PATH(req_host)				\
+	FOR_REQ("OPTIONS http://" req_host " HTTP/1.1\r\n\r\n") {	\
+		EXPECT_EQ(req->method, TFW_HTTP_METH_OPTIONS);		\
+		EXPECT_TFWSTR_EQ(&req->host, req_host);			\
+		EXPECT_TFWSTR_EQ(&req->uri_path, "/");			\
```
EvgeniiMekhanik: It seems that there is a mistake here isn't it? According RFC 9112:
```
For example, the request

OPTIONS http://www.example.org:8001 HTTP/1.1

would be forwarded by the final proxy as

OPTIONS * HTTP/1.1
Host: www.example.org:8001
```

So in this case we should also expect uri_path as '*' isn't it?

const-t: Agree with @EvgeniiMekhanik

fw/http_parser.c
```diff
@@ -6042,11 +6050,17 @@ Req_Method_1CharStep: __attribute__((cold))
...
 		__msg_field_finish(&req->host, p);
 		T_DBG3("host len = %i\n", (int)req->host.len);
+
 		if (likely(c == '/')) {
 			__msg_field_open(&req->uri_path, p);
 			__FSM_MOVE_f(Req_UriAbsPath, &req->uri_path);
 		}
 		else if (c == ' ') {
+			if (req->method == TFW_HTTP_METH_OPTIONS)
+				req->uri_path = TFW_STR_F_STRING("*", TFW_STR_COMPLETE);
```
const-t: Why `slash` is `static` but `TFW_STR_F_STRING("*", TFW_STR_COMPLETE);` not? Make them the same.

fw/cache.c
```diff
@@ -2320,8 +2320,10 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 			continue;
 
 		p = TDB_PTR(db->hdr, ce->hdrs_304[i]);
+
```
const-t: Please, check your changes before pushing/committing and revert changes that doesn't relates to current fix. For instance these empty lines.

fw/http_parser.c
```diff
@@ -5922,15 +5924,20 @@ Req_Method_1CharStep: __attribute__((cold))
 		__FSM_JMP(Req_UriAbsoluteForm);
 	}
 
-	__FSM_STATE(Req_UriRareFormsEnd, hot) {
+	__FSM_STATE(Req_UriRareFormsEnd, hot)
+	{
```
const-t: Why you changed formatting here?

fw/http_parser.c
```diff
@@ -5922,15 +5924,20 @@ Req_Method_1CharStep: __attribute__((cold))
...
 		if (likely(c == '/')) {
 			__msg_field_open(&req->uri_path, p);
 			__FSM_MOVE_f(Req_UriAbsPath, &req->uri_path);
 		}
 		else if (c == ' ') {
+			if (req->method == TFW_HTTP_METH_OPTIONS)
```
const-t: I can't see any states that leads to `Req_UriRareFormsEnd` except `Req_UriRareForms`. If so, we can change the name of `Req_UriRareForms` to `Req_UriAsteriskForm` and in this state we can do parse [asterisk form](https://www.rfc-editor.org/rfc/rfc9112.html#name-asterisk-form) :
```
__FSM_STATE(Req_UriRareFormsEnd, hot) {
	if (c == ' ') {
		req->uri_path = TFW_STR_F_STRING("*", TFW_STR_COMPLETE);
		__FSM_MOVE_nofixup(Req_HttpVer);
	}
	TFW_PARSER_DROP(Req_UriRareFormsEnd);
}
```
What do you think, can we?

Also it would be good to add these tests:
```
EXPECT_BLOCK_REQ("OPTIONS *  HTTP/1.1\r\n\r\n");
EXPECT_BLOCK_REQ("OPTIONS */asd HTTP/1.1\r\n\r\n");
EXPECT_BLOCK_REQ("OPTIONS *HTTP/1.1\r\n\r\n");
EXPECT_BLOCK_REQ("OPTIONS ** HTTP/1.1\r\n\r\n");
``

Comment (EvgeniiMekhanik) [reply fw/http_parser.c]: We also need test like this
```
FOR_REQ("OPTIONS http://" "tempesta-tech.com/home?name=value" 
		" HTTP/1.1\r\n\r\n")
	{
		EXPECT_EQ(req->method, TFW_HTTP_METH_OPTIONS);
		EXPECT_TFWSTR_EQ(&req->host, "tempesta-tech.com");
		EXPECT_FALSE(tfw_str_eq_cstr(&req->uri_path, "*", 1, 0));
	}
```

File: fw/http_parser.c [hunk 17]
```diff
@@ -5912,25 +5915,22 @@ Req_Method_1CharStep: __attribute__((cold))
 		__FSM_MOVE_nofixup_n(Req_MUSpace, 0);
 	}
 
-	__FSM_STATE(Req_UriRareForms, cold) {
+	__FSM_STATE(Req_UriAsteriskForm, cold) {
```
const-t: Why did you rename it to `Req_UriAsteriskForm`? This state may lead to Asterisk and to Absolute forms. `Req_UriRareForms` is suitable name.

fw/t/unit/test_http1_parser.c
```diff
@@ -220,6 +241,19 @@ TEST(http1_parser, parses_req_uri)
...
 	TEST_FULL_REQ("natsys-lab.com:8080", "/cgi-bin/show.pl?entry=tempesta");
 
+	TEST_OPTIONS_ASTERISK();
+
+	TEST_OPTIONS_WITHOUT_PATH("example.com");
+	TEST_OPTIONS_WITHOUT_PATH("example.com:8080");
+	TEST_OPTIONS_WITHOUT_PATH("tempesta-tech.com");
+
+	FOR_REQ("OPTIONS http://tempesta-tech.com/home?name=value HTTP/1.1\r\n\r\n")
+		{
```
const-t: broken identation

fw/cache.c
```diff
@@ -2322,6 +2322,7 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
 		p = TDB_PTR(db->hdr, ce->hdrs_304[i]);
 		while (trec && (p + TFW_CSTR_HDRLEN > trec->data + trec->len))
 			trec = tdb_next_rec_chunk(db, trec);
+
```
const-t: delete this please

const-t (changes requested):
Looks good, but need couple of minor changes

EvgeniiMekhanik (approved):

const-t (approved):

PR #2433, opened 2025-05-29
Mekhanik evgenii/1411 rework allocation streams 1

EvgeniiMekhanik:


fw/http_stream_sched.h
```diff
@@ -39,10 +41,20 @@ typedef struct tfw_stream_sched_entry_t {
 	long int			active_cnt;
 	TfwStream			*owner;
 	struct tfw_stream_sched_entry_t	*parent;
+	struct tfw_stream_sched_entry_t *next_free;
```
const-t: It seems we can get rid of `next_free`, each free `TfwStreamSchedEntry` can use first 8 byte to store pointer to the next free `TfwStreamSchedEntry`. As head pointer use `TfwStreamSched::free_list`.

krizhanovsky: Yeah, agree

fw/http2.c
```diff
@@ -248,14 +332,44 @@ tfw_h2_context_clear(TfwH2Ctx *ctx)
...
+
+TfwStreamSchedEntry *
+tfw_h2_alloc_stream_sched_entry(TfwH2Ctx *ctx)
+{
+	TfwStreamSchedEntry *entry;
+
+	if (!ctx->sched.free_list) {
+		if (tfw_h2_context_alloc_extra_sched_entry_list(ctx))
+			return NULL;
+	}
```
krizhanovsky:
```suggestion
	if (!ctx->sched.free_list)
		if (tfw_h2_context_alloc_extra_sched_entry_list(ctx))
			return NULL;
```

fw/http_stream_sched.h
```diff
@@ -27,32 +27,49 @@
...
 	struct eb_root			blocked;
 } TfwStreamSchedEntry;
 
+/**
+ * @next 		- pointer to the next scheduler list;
+ * @entries 		- flexible array of schedulers;
+ */
+typedef struct tfw_stream_sched_list {
+	struct tfw_stream_sched_list *next;
+	TfwStreamSchedEntry entries[];
```
krizhanovsky: Align the members

```suggestion
	struct tfw_stream_sched_list    *next;
	TfwStreamSchedEntry		entries[];
```

fw/http_stream_sched.h
```diff
@@ -51,11 +63,13 @@ typedef struct tfw_stream_sched_entry_t {
  *			  streams storage;
  * @root		- root scheduler of per-connection priority tree;
  * @blocked_streams	- count of blocked streams;
+ * @free_list		- list of preallocated schedulers;
  */
 typedef struct tfw_stream_sched_t {
 	struct rb_root		streams;
 	TfwStreamSchedEntry	root;
 	long int		blocked_streams;
+	TfwStreamSchedEntry	*free_list;
```
krizhanovsky, line 74: Probably we should rearrange the structure members to avoid memory holes, e.g. `TfwStreamSchedEntry` members should be together and small members are coupled

fw/http2.c
```diff
@@ -204,15 +204,97 @@ tfw_h2_cleanup(void)
...
+#undef LIST_SZ_DFLT
+}
+
+static inline int
+tfw_h2_context_alloc_extra_sched_entry_list(TfwH2Ctx *ctx)
+{
+	TfwStreamSchedList *list;
+	struct page *page;
+
+	page = alloc_page(GFP_ATOMIC);
```
krizhanovsky: How many schedulers we have provisioned in the first page and how many will be in this page? Does it make sense to allocate a whole page if we don't fit the first one? Is it possible that we won't utilize it in the most cases? Probably it makes sense to fallback to a slab allocator (BTW does it provide per-cpu allocations?)

In any case we need a solid comment about the number of provisioned scheduler entries, how many of them we allocate each time and which number of streams we expect. We need a good reasoning for an allocation strategy,

fw/tls.c
```diff
@@ -630,8 +630,10 @@ tfw_tls_conn_dtor(void *c)
 	TlsCtx *tls = tfw_tls_context(c);
 
 	if (TFW_CONN_PROTO((TfwConn *)c) == TFW_FSM_H2
-	    && ttls_hs_done(tls))
+	    && ttls_hs_done(tls)) {
```
krizhanovsky:
```suggestion
	    && ttls_hs_done(tls))
	{
```

fw/http2.c
```diff
@@ -204,14 +204,27 @@ tfw_h2_cleanup(void)
 	tfw_h2_stream_cache_destroy();
 }
 
+TfwH2Ctx *
+tfw_h2_context_alloc(void)
+{
+	return (TfwH2Ctx *)kzalloc(sizeof(TfwH2Ctx), GFP_ATOMIC);
```
krizhanovsky: Why not using caching allocator? I think we might have a lot of new HTTP/2 context allocations and the context is of fixed size

fw/http_stream.c
```diff
@@ -127,7 +130,7 @@ tfw_h2_add_stream(TfwStreamSched *sched, unsigned int id, unsigned short weight,
 	if (unlikely(!new_stream))
```
krizhanovsky, line 128: Why do we allocate a stream separately from the H2 context and the schedulers? We know that a TCP connection is almost always per-cpu, so does `kmem_cache_alloc()` provide per-cpu locality? It's safe to assume that a TCP connection, established for HTTP/2, allocates at least 5-10 streams - so why don't we preallocate at lease several streams together with `TfwH2Ctx`, especially having that now we fully decoupled it from the TLS connection?

**UPDATE** OK, I see that we need to keep schedulers together to optimize reprioritization dealing only with schedulers and it could be a case that we touch only one, scheduled stream, per a softirq shot. However, having that we still have at least several streams in the most cases and at least 1 - should we preallocate some of them with the HTTP/2 context instead of calling caching allocator several times?

Also with the next PRs #2445 and #2447 - do we still have a problem with scheduling streams? I.e are the schedulers as big as they are still in this PR and are they hot? Do we still benefit from memory-local schedulers?

krizhanovsky (approved):
The PR is good, but it seems there is still room for improvement - let's discuss on the call.

I have a couple of questions in the comment and I think it's better to address them in comments in the code - it'll be good to have decisions made in this PR well documented (I had hard time to remember our discussions about schedulers locality and that streams locality isn't important).

fw/http_stream.c
EvgeniiMekhanik: Schedulers have the same size in all this PRs. https://github.com/tempesta-tech/tempesta/pull/2447 is not about scheduling, https://github.com/tempesta-tech/tempesta/pull/2445 change `eb_root` to `list_head` which has the same size. In 2445 we minimize count of remove/insert schedulers (now we remove scheduler only when HTTP2 stream window size is exceeded - stream is blocked). and insert it later during processing window update frame. Previously we decide to don't preallocate streams  streams because size of streamis very big (about 900 bytes).

fw/http2.c
EvgeniiMekhanik: Yes good point

fw/http2.c
EvgeniiMekhanik: We have about 50 schedulers on the first page, and about 60 schedulers on each new allocated page.

krizhanovsky: So if we don't fit 50 schedulers, we allocate 60 more, which is close to classic x2. But probably it's rare to see 100 streams and more, right? Do we need to allocate 60 more streams? @const-t @EvgeniiMekhanik we need a good discussion and comment on this

fw/http2.c
const-t: @EvgeniiMekhanik @krizhanovsky  This is outdated code, I can't see this part in final version. I missed something?

fw/http2.c
```diff
@@ -251,11 +301,63 @@ tfw_h2_context_clear(TfwH2Ctx *ctx)
...
+
+TfwStreamSchedEntry *
+tfw_h2_alloc_stream_sched_entry(TfwH2Ctx *ctx)
+{
+	TfwStreamSchedEntry *entry;
+
+	if (unlikely(!ctx->sched.free_list)) {
+		/*
+		 * If count of preallocate streams schedulers is exceeded
+		 * use stadart kernel allocator. There is no sence to
```
krizhanovsky:
```suggestion
		 * use standard kernel allocator. There is no sense to
```

fw/http2.c
```diff
@@ -251,11 +301,63 @@ tfw_h2_context_clear(TfwH2Ctx *ctx)
...
 
+
+TfwStreamSchedEntry *
+tfw_h2_alloc_stream_sched_entry(TfwH2Ctx *ctx)
+{
+	TfwStreamSchedEntry *entry;
+
+	if (unlikely(!ctx->sched.free_list)) {
+		/*
+		 * If count of preallocate streams schedulers is exceeded
```
krizhanovsky: Could you please add that for now it is 50 (50, right?). This is crucial to realize whether the default number of precreated schedulers is adequate

fw/http2.c
```diff
@@ -251,11 +301,63 @@ tfw_h2_context_clear(TfwH2Ctx *ctx)
...
+		 * use stadart kernel allocator. There is no sence to
+		 * allocate the whole page for the new schedulers or
+		 * use special cache for this purpose, because it is a
+		 * very rare case (browsers usually open not more then
+		 * 15 - 20 streams in parallel even if there are much
+		 * more resourses to request).
+		 */ 
+		entry = kzalloc(sizeof(TfwStreamSchedEntry), GFP_ATOMIC);
```
krizhanovsky: Could you please comment what is `sizeof(TfwStreamSchedEntry)` to evaluate how much memory we overruse in this case. If it is small and close the a power of 2, then it's just like caching allocator

PR #2435, opened 2025-06-03
fix: Use vhost name instead of Host header in cache key calculation

MorganaFuture:
**What**
Use vhost name instead of Host header in cache key calculation

**Why**
This change ensures proper cache operation when requests are redirected to different vhosts through HTTP chains.
Previously, the cache key was built using uri_path + host (from Host header), which caused incorrect cache behavior when a request with a Host header for one vhost was redirected to a different vhost via HTTP chains. This would lead to storing the cached response under a key that combined the URI with the original Host header, rather than the actual vhost that served the content.
With this change, the cache key is now correctly built using uri_path + vhost_name, ensuring that the cache operates properly even when HTTP chains redirect requests between different vhosts. The code falls back to using the Host header when a vhost is not yet assigned.

Links
[2366](https://github.com/tempesta-tech/tempesta/issues/2366)


fw/cache.c
```diff
@@ -302,14 +302,13 @@ static DEFINE_PER_CPU(char[RESP_BUF_LEN], g_c_buf);
...
+ * Iterate over request URI and vhost name to process request key.
+ * uri_path is chunked, vhost name is a simple string.
  */
-#define TFW_CACHE_REQ_KEYITER(c, uri_path, host, u_end, h_start,	\
-			      h_end, u_fin, h_fin)			\
+#define TFW_CACHE_REQ_KEYITER(c, uri_path, vhost_name, u_end, v_start,	\
+			      v_end, u_fin, v_fin)			\
 	c = NULL;							\
-	if (!(u_fin = WARN_ON_ONCE(TFW_STR_EMPTY(uri_path)))) {		\
+	if (!(u_fin = WARN_ON_ONCE(TFW_STR_EMPTY(uri_path)))) {	\
```
EvgeniiMekhanik: Broken `\` aliment at the end of the string

fw/t/unit/test_http_cache.c
```diff
@@ -0,0 +1,217 @@
...
+TEST(http_cache, empty_vhost_name)
+{
+	unsigned long key_empty_vhost, key_named_vhost, key_empty_vhost_again;
+	TfwVhost vhost_empty;
+	BasicStr empty_name = { .data = "", .len = 0 };
+
+	memset(&vhost_empty, 0, sizeof(vhost_empty));
+	vhost_empty.name = empty_name;
+
+	TfwStr host_header = { .data = (void *)"some.host.com", .len = 13 };
```
EvgeniiMekhanik: All variables should be declared at the beginning of the function. ISO C90 forbids mixed declarations and code

fw/t/unit/test_http_cache.c
```diff
@@ -0,0 +1,217 @@
...
+	test_req->host = host_header;
+
+	TfwStr uri_path = { .data = (void *)"/test", .len = 5 };
```
EvgeniiMekhanik: ISO C90 forbids mixed declarations and code

fw/t/unit/test_http_cache.c
```diff
@@ -0,0 +1,217 @@
...
+TEST(http_cache, uses_vhost_not_host)
+{
+	unsigned long key1, key2;
+
+	/* Set the Host header to a specific value */
+	TfwStr host = {
+		.data = (void *)"same.host.example.com",
+		.len = 21
+	};
+	test_req->host = host;
```
EvgeniiMekhanik, line 67: ISO C90 forbids mixed declarations and code

fw/t/unit/test_http_cache.c
```diff
@@ -0,0 +1,217 @@
...
+TEST(http_cache, stable_with_http_chains)
+{
+	unsigned long key1, key2;
+
+	/* Set a URI path */
+	TfwStr uri_path = {
+		.data = (void *)"/test/path",
+		.len = 10
+	};
+	test_req->uri_path = uri_path;
```
EvgeniiMekhanik: ISO C90 forbids mixed declarations and code

fw/t/unit/test_http_cache.c
```diff
@@ -0,0 +1,217 @@
...
+	test_req->hash = 0; /* Clear cached hash */
+	key1 = tfw_http_req_key_calc(test_req);
+
+	/* Now change to second vhost, same Host header */
+	test_req->vhost = &test_vhost2;
+	test_req->hash = 0; /* Clear cached hash */
+	key2 = tfw_http_req_key_calc(test_req);
+
+	/* Keys should be different because vhost names are different */
+	EXPECT_NE(key1, key2);
```
EvgeniiMekhanik, line 85: This expect fails, when I run this test.

EvgeniiMekhanik:
I also catch kernel warning at the string 2281 (cache.c) `if (WARN_ON_ONCE(tot_len != 0))`

EvgeniiMekhanik:
I also create brunch and add test. We should always write python tests.

EvgeniiMekhanik (changes requested):
Need to rework

fw/cache.c
MorganaFuture: Fixed

fw/t/unit/test_http_cache.c
MorganaFuture: Fixed

fw/t/unit/test_http_cache.c
MorganaFuture: Fixed

fw/t/unit/test_http_cache.c
MorganaFuture: Fixed

fw/t/unit/test_http_cache.c
MorganaFuture: Fixed

fw/t/unit/test_http_cache.c
MorganaFuture: Tests pass on my machine, could you please double check? 
test_screen

EvgeniiMekhanik:
```
[ 1291.938904] Start test: cache.test_cache.TestCacheH2.test_cache_date
[ 1292.566247] [tdb] Start Tempesta DB
[ 1292.642012] [tempesta fw] Initializing Tempesta FW kernel module...
[ 1293.358834] [tempesta fw] Warning: Vhost default doesn't have certificate with matching SAN/CN.
[ 1293.358834]     Maybe that's fine, but it's worth checking the
[ 1293.358834]     config - if there is no relations between the
[ 1293.358834]     names, then host name confusion attack is possible.
[ 1293.533509] [tempesta fw] Configuration processing is completed.
[ 1293.579842] [tdb] Opened table /opt/tempesta/db/filter0.tdb: size=16777216 rec_size=20 base=000000006d0c7901
[ 1293.801164] [tdb] Opened table /opt/tempesta/db/cache0.tdb: size=268435456 rec_size=0 base=000000000019a777
[ 1293.816134] [tdb] Opened table /opt/tempesta/db/sessions0.tdb: size=16777216 rec_size=312 base=00000000cf75e9e4
[ 1293.829874] [tdb] Opened table /opt/tempesta/db/client0.tdb: size=16777216 rec_size=624 base=00000000c10a1b40
[ 1293.897155] [tempesta fw] Open listen socket on: 0.0.0.0:443
[ 1293.899310] [tempesta fw] Open listen socket on: 0.0.0.0
[ 1293.965049] [tempesta fw] Tempesta FW is ready
[ 1294.037212] ==================================================================
[ 1294.038188] BUG: KASAN: wild-memory-access in __hash_calc+0x162/0x190 [tempesta_lib]
[ 1294.039205] Read of size 1 at addr 00746c7561666564 by task swapper/6/0
```
please build debug kernel with kasan and run python tests (cache, http2_general)  with and without tcp segmentation to check that there is no problems.

EvgeniiMekhanik (changes requested):
I catch bug during run of cache tests

fw/http.c
```diff
@@ -7750,8 +7753,15 @@ tfw_http_req_key_calc(TfwHttpReq *req)
...
-		req->hash ^= tfw_hash_str(&req->host);
+	BUG_ON(!req->vhost);
+
+	/* Convert BasicStr to TfwStr for hashing */
+	vhost_name.data = req->vhost->name.data;
+	vhost_name.len = req->vhost->name.len;
+	vhost_name.flags = 0;
+	vhost_name.nchunks = 0;
+
+	req->hash ^= tfw_hash_str(&vhost_name);
```
const-t: Use `basic_hash_str()` instead of `tfw_hash_str`.

fw/cache.c
```diff
@@ -2415,8 +2424,9 @@ __cache_entry_size(TfwHttpResp *resp)
 	TfwCaTokenArray hdr_del_tokens =
 			tfw_vhost_get_capo_hdr_del(req->location, req->vhost);
 	/* Add compound key size */
+	BUG_ON(!req->vhost);
```
const-t: Please don't set `BUG_ON` everywhere, you checking emptiness of vhost, before calling `__cache_entry_size()`.

fw/cache.c
```diff
@@ -1259,15 +1253,24 @@ tfw_cache_entry_key_eq(TDB *db, TfwHttpReq *req, TfwCacheEntry *ce)
...
 
-	if (req->uri_path.len + req->host.len != ce->key_len)
+	if (req->uri_path.len + vhost_name.len != ce->key_len)
 		return false;
 
 	t_off = CE_BODY_SIZE;
-	TFW_CACHE_REQ_KEYITER(c, &req->uri_path, &req->host, u_end, h_start,
-			      h_end, u_fin, h_fin)
+	TFW_CACHE_REQ_KEYITER(c, &req->uri_path, &vhost_name, u_end, v_start,
+				v_end, u_fin, v_fin)
```
const-t: broken identation

fw/cache.c
```diff
@@ -2081,21 +2087,24 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
...
+	BUG_ON(!req->vhost);
+
+	/* Convert BasicStr to TfwStr */
+	vhost_name.data = req->vhost->name.data;
+	vhost_name.len = req->vhost->name.len;
+	vhost_name.flags = 0;
+	vhost_name.nchunks = 0;
+
+	TFW_CACHE_REQ_KEYITER(field, &req->uri_path, &vhost_name, end1, v_start,
+			end2, u_fin, v_fin)
```
const-t: broken identation

fw/cache.c
```diff
@@ -318,20 +317,15 @@ static TfwStr g_crlf = { .data = S_CRLF, .len = SLEN(S_CRLF) };
...
-	if (!(h_fin = WARN_ON_ONCE(TFW_STR_EMPTY(host)))) {		\
-		if (likely(!TFW_STR_PLAIN(host))) {			\
-			h_start = (host)->chunks;			\
-			h_end = (host)->chunks + (host)->nchunks;	\
-		} else {						\
-			h_start = host;					\
-			h_end = (host) + 1;				\
-		}							\
-		c = c ? : h_start;					\
+	if (!(v_fin = WARN_ON_ONCE(!(vhost_name)->len))) {		\
```
const-t, line 321: Maybe something like this?
```
TfwStr v_start;
TFW_CACHE_REQ_KEYITER(...)
...
...
if (!(v_fin = WARN_ON_ONCE(!(vhost_name).len))) {		\
	(v_start)->data = vhost_name.data;			\ ** Here vhost_name is BasicStr** 
	(v_start)->len = vhost_name.len;			\
	v_end = (v_start) + 1;					\
	c = c ? : v_start;					\
}
...
```

fw/http.c
```diff
@@ -7750,8 +7753,15 @@ tfw_http_req_key_calc(TfwHttpReq *req)
 	if (test_bit(TFW_HTTP_B_HMONITOR, req->flags))
 		return req->hash;
 
-	if (!TFW_STR_EMPTY(&req->host))
-		req->hash ^= tfw_hash_str(&req->host);
+	BUG_ON(!req->vhost);
+
+	/* Convert BasicStr to TfwStr for hashing */
+	vhost_name.data = req->vhost->name.data;
```
EvgeniiMekhanik: Can req->vhost be empty? If yes it seems that we should not use it for `tfw_http_req_key_calc` isn't it? @const-t how do you think?

fw/t/unit/test_http_cache.c
```diff
@@ -0,0 +1,223 @@
...
+	/* Clear the flag for next tests */
+	__clear_bit(TFW_HTTP_B_HMONITOR, test_req->flags);
+}
+
+/**
+ * Test cache key calculation with an empty vhost name.
+ * An empty vhost name should still result in a deterministic key,
+ * different from a key with a non-empty vhost name.
+ */
+TEST(http_cache, empty_vhost_name)
```
EvgeniiMekhanik: Why we use empty vhost for key calculation?

fw/t/unit/test_http_cache.c
```diff
@@ -0,0 +1,223 @@
...
+
+	/* Keys should be different because vhost names are different */
+	EXPECT_NE(key1, key2);
+}
+
+/**
+ * Test that the cache key is the same even if the Host header changes
+ * but the vhost remains the same (which is what happens with HTTP chains)
+ */
+TEST(http_cache, stable_with_http_chains)
```
EvgeniiMekhanik: It seems that this test is a duplicate of previous one. It's better to implement python tests for this case

fw/cache.c
```diff
@@ -2081,21 +2087,24 @@ tfw_cache_copy_resp(TDB *db, TfwCacheEntry *ce, TfwHttpResp *resp, TfwStr *rph,
...
 
-	/*
-	 * Get 'host' header value (from HTTP/2 or HTTP/1.1 request) for
-	 * strict comparison.
-	 */
-	TFW_CACHE_REQ_KEYITER(field, &req->uri_path, &req->host, end1, h,
-			      end2, u_fin, h_fin)
+	BUG_ON(!req->vhost);
+
+	/* Convert BasicStr to TfwStr */
```
EvgeniiMekhanik: It seems that we use this conversion two times. We can implement macros like `TFW_STR_STRING`
for example 
`#define __TFW_STR_STRING(val, len) (same as it was in TFW_STR_STRING but no calculate SLEN).
#define TFW_STR_STRING(val) __TFW_STR_STRING(val, SLEN(val))
```

And here use `__TFW_STR_STRING`

File: fw/cache.c (hunk 11 above)
Comment (EvgeniiMekhanik) [inline fw/cache.c:321]: TFW_STR_EMPTY insteam of direct usage (vhost_name)->len?

File: fw/http.c (hunk 12 above)
Comment (const-t) [reply fw/http.c]: @EvgeniiMekhanik  @MorganaFuture One of the reasons to use vhost instead of host is that `req->vhost` must not be empty when we hit the cache. If `vhost` not found, we not pass a request to the cache.

File: fw/cache.c (hunk 15 above)
Comment (const-t) [reply fw/cache.c]: I believe we can do [this](https://github.com/tempesta-tech/tempesta/pull/2435#discussion_r2207789200) and we don't need conversion at all. Just use `v_start` as local var.

File: fw/http.c (hunk 7 above)
Comment (MorganaFuture) [reply fw/http.c]: Accepted

File: fw/cache.c (hunk 8 above)
Comment (MorganaFuture) [reply fw/cache.c]: Removed

File: fw/cache.c (hunk 9 above)
Comment (MorganaFuture) [reply fw/cache.c]: Fixed

File: fw/cache.c (hunk 10 above)
Comment (MorganaFuture) [reply fw/cache.c]: Fixed

File: fw/cache.c (hunk 11 above)
Comment (MorganaFuture) [reply fw/cache.c:321]: Accepted

File: fw/t/unit/test_http_cache.c (hunk 13 above)
Comment (MorganaFuture) [reply fw/t/unit/test_http_cache.c]: The unit test covers an edge case where vhost exists but has an empty name (name.len = 0). This is important because BUG_ON(!req->vhost) only checks if vhost is not NULL, not if its name is empty, so the function must handle such cases correctly without crashing.

File: fw/cache.c (hunk 11 above)
Comment (MorganaFuture) [reply fw/cache.c:321]: vhost_name is now BasicStr

File: fw/t/unit/test_http_cache.c (hunk 14 above)
Comment (MorganaFuture) [reply fw/t/unit/test_http_cache.c]: Removed, added python test

File: fw/t/unit/test_http_cache.c (hunk 13 above)
Comment (EvgeniiMekhanik) [reply fw/t/unit/test_http_cache.c]: It seems that we can add BUG_ON(!req->vhost->len) also? Vhost is set up during configuration and it seems that name can't be equal to zero. If now please can you add python test to reproduce it.

File: fw/http.c [hunk 16]
```diff
@@ -7750,8 +7750,9 @@ tfw_http_req_key_calc(TfwHttpReq *req)
 	if (test_bit(TFW_HTTP_B_HMONITOR, req->flags))
 		return req->hash;
 
-	if (!TFW_STR_EMPTY(&req->host))
-		req->hash ^= tfw_hash_str(&req->host);
+	BUG_ON(!req->vhost);
```
EvgeniiMekhanik, line 7748: It seems that || !req->vhost->len also

EvgeniiMekhanik (approved):

fw/t/unit/test_http_cache.c
MorganaFuture: You're right, Vhost is set up during configuration, and it seems that the name can't be zero. I removed that test.

fw/http.c
MorganaFuture: There's no point in this check with `BUG_ON(!req->vhost);` — length doesn't exist.

PR #2436, opened 2025-06-04
Fix: correct HTTP method parsing in tfw_http_meth_str2id()

### LINKED ISSUE #2362: Function `tfw_http_meth_str2id` is totally wrong
In `tfw_http_meth_str2id` function we check first letter in chunk and the length of header and return method type according to it. It is totally wrong, because we return method 'PUT' if the real method was 'PAT' or 'POT'
and so on. This function should be reworked

WitalyAnisimov:
The function now properly parses and identifies HTTP methods based on their length and string content. Fixed multiple invalid cases in the old implementation. Fix #2362


fw/http_parser.c
```diff
@@ -11130,66 +11130,85 @@ enum {
...
-	unsigned long mv_len;
-	unsigned char *p;
 	const TfwStr *chunk;
+	const unsigned char *p;
+	size_t len;
 
 	BUG_ON(TFW_STR_PLAIN(m_hdr));
 
-	mv_len = m_hdr->len - H2_METH_HDR_VLEN;
-	/* ':method' name should always be in a single chunk */
```
const-t, line 11140: Please do not remove this comment

const-t (changes requested):
This PR doesn't solve the problem. There are few points to fix:
1. `if (p[0] == 'G' && p[1] == 'E' && p[2] == 'T')` `GET` will be matched with `GETASD` and so on. But must be matched only with `GET`. For other headers the same.
2. `if (p[0] == 'G' && p[1] == 'E' && p[2] == 'T')`  comparison of each symbol looks bad, please see how we do comparisons `tfw_h2_parse_req_hdr_name()`.
3. This function doesn't handle all supported headers listed in `tfw_http_meth_t`, please add missed headers.

fw/http_parser.c
```diff
@@ -11130,66 +11130,85 @@ enum {
 unsigned char
 tfw_http_meth_str2id(const TfwStr *m_hdr)
```
const-t, line 11165: This function is good candidate for unit test, please implement it.

fw/http_parser.c
EvgeniiMekhanik: This function is still wrong. I add test in appropriate branch (with the same name) in tempesta-test to check it. If we don't use huffman encoding and there is a TCP segmentation there are a lot of chunks in `m_hdr`.  May be we can check if there are more then 2 chunks in `m_hdr` we should call some static function with full comparison.

fw/http_parser.c
```diff
@@ -11130,66 +11130,94 @@ enum {
...
-		case 'L':
-			return TFW_HTTP_METH_LOCK;
-		case 'M':
-			return TFW_HTTP_METH_MOVE;
-		case 'P':
-			return TFW_HTTP_METH_POST;
-		default:
-			return _TFW_HTTP_METH_UNKNOWN;
+	switch (len) {
+	case 3:
```
EvgeniiMekhanik: please use enum `TFW_HTTP_MLEN_3C` here and in other places

fw/cache.c
```diff
@@ -1267,16 +1267,26 @@ tfw_cache_entry_key_eq(TDB *db, TfwHttpReq *req, TfwCacheEntry *ce)
 
 	t_off = CE_BODY_SIZE;
 	TFW_CACHE_REQ_KEYITER(c, &req->uri_path, &req->host, u_end, h_start,
-			      h_end, u_fin, h_fin)
+			h_end, u_fin, h_fin)
```
EvgeniiMekhanik, line 1270: This commit is unrelated to this patch. (I mean patching of cache.c)

fw/http_parser.c
```diff
@@ -11130,66 +11130,94 @@ enum {
...
+			if (C4_INT(p, 'M', 'O', 'V', 'E'))
+				return TFW_HTTP_METH_MOVE;
+			if (C4_INT(p, 'L', 'O', 'C', 'K'))
+				return TFW_HTTP_METH_LOCK;
+			if (C4_INT(p, 'C', 'O', 'P', 'Y'))
+				return TFW_HTTP_METH_COPY;
+		}
+		break;
+	case 5:
+		if (chunk->len == 5) {
```
EvgeniiMekhanik: If there is no tcp segmentation or haffman encoding is used we don't need to check it, because it  was already checked in `case 5:`. If we deal with tcp segmentation without huffman this function is wrong at all, because there can be a lot of chunks in `m_hdr`.

fw/http_parser.c
```diff
@@ -11130,66 +11130,94 @@ enum {
...
-		case 'M':
-			return TFW_HTTP_METH_MOVE;
-		case 'P':
-			return TFW_HTTP_METH_POST;
-		default:
-			return _TFW_HTTP_METH_UNKNOWN;
+	switch (len) {
+	case 3:
+		if (chunk->len == 3) {
+			if (p[0] == 'G' && p[1] == 'E' && p[2] == 'T')
```
EvgeniiMekhanik: We have macros in `http_msg.c`:
```

#define TFW_CHAR4_INT(data)	*((u32 *)(data))
#define TFW_CHAR3_INT(data)	((data[2] << 16) | *((u16 *)(data)))
#define TFW_CHAR2_INT(data)	*((u16 *)(data))
#define TFW_CHAR_INT(data)	(data[0])
```

We need use the same one (like C4_INT)  it this function to check it, not `if (p[0] == 'G' && p[1] == 'E' && p[2] == 'T')`

fw/http_parser.c
```diff
@@ -11120,8 +11153,9 @@ enum {
 	TFW_HTTP_MLEN_7C,
 	TFW_HTTP_MLEN_8C,
 	TFW_HTTP_MLEN_9C,
+	TFW_HTTP_MLEN_10C,
 };
-#define H2_METH_HDR_VLEN    7
+#define H2_METH_HDR_VLEN    8
```
EvgeniiMekhanik, line 11158: ? H2_METH_HDR_VLEN is equal to 7 ":method"

fw/http_parser.c
```diff
@@ -11130,66 +11164,89 @@ enum {
...
+		return _TFW_HTTP_METH_UNKNOWN;
+
+	copied = 0;
+	for (i = 1; i < m_hdr->nchunks && copied < len; ++i) {
+		const TfwStr *chunk = TFW_STR_CHUNK(m_hdr, i);
+		size_t to_copy = (len - copied < chunk->len) ?
+				 (len - copied) : chunk->len;
 
-	BUG_ON(TFW_STR_PLAIN(m_hdr));
+		memcpy(method_buf + copied, chunk->data, to_copy);
```
EvgeniiMekhanik, line 11183: It is wrong. We should not use extra copy. First of all tcp segmentation without huffman is a rare case, so we can use simple version of this function (when there are only two chunks). And if count of chunks is greater then 2 use iteration over all chunks and comparison, without copy

krizhanovsky (commented):
Well, #2362 isn't labeled as "good to start", but it seems from #2452 it required deep HTTP/2 logic in Tempesta FW. If #2452 is correct and fixes the issue, then we should go with #2452

Issue #2275, opened 2024-11-02
Cache current time stamp


krizhanovsky:
# Motivation

There are dozens of `tfw_current_timestamp()` calls around the code, which executes `ktime_get_real_ts64()`, a relatively heavyweight kernel function. Meantime, Tempesta FW code doesn't require high precise timestamps, so this can be optimized.

# Scope

Call the ``ktime_get_real_ts64()` only once per a softirq shot and cache the value in a per-cpu variable.

krizhanovsky:
Please also make the `x509_get_current_time()` from https://github.com/tempesta-tech/tempesta/pull/2339/commits/8aefc6af1ae2ccd31dc5723453a06965c7def273 to use the cached timestamp instead of `ktime_get_real_seconds()`

PR #2438, opened 2025-06-09
refactor: optimize tfw current timestamp

MorganaFuture:
**What**
Optimize tfw_current_timestamp() to use kernel-level softirq timestamp caching. When in softirq context, use cached timestamp from softirq_current_timestamp() instead of calling expensive ktime_get_real_ts64() directly.

**Why**
Tempesta makes dozens of tfw_current_timestamp() calls per softirq batch, each executing expensive ktime_get_real_ts64(). This optimization leverages kernel-level caching (updated once per softirq batch) to significantly reduce CPU overhead in high-throughput scenarios while maintaining sufficient timestamp precision for Tempesta's use cases
**Links**
[Kernel patch](https://github.com/tempesta-tech/linux-6.12.12-tfw/pull/3)
[2275](https://github.com/tempesta-tech/tempesta/issues/2275)

### DIFF
--- a/lib/common.h
+++ b/lib/common.h
@@ -20,7 +20,30 @@
 #ifndef __LIB_COMMON_H__
 #define __LIB_COMMON_H__
 
-/* Get current timestamp in secs. */
+#include <linux/interrupt.h>
+
+#ifdef CONFIG_SECURITY_TEMPESTA
+/* Get current timestamp in secs.*/
+static inline long
+tfw_current_timestamp(void)
+{
+	/*
+	 * Use kernel-cached timestamp when in softirq context.
+	 * The timestamp is updated once per softirq batch in handle_softirqs().
+	 * This provides significant performance improvement by avoiding multiple
+	 * expensive ktime_get_real_ts64() calls per softirq batch.
+	 */
+	if (likely(in_serving_softirq()))
+		return softirq_current_timestamp();
+	
+	/* Fallback to direct call outside softirq context */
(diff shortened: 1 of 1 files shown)


MorganaFuture:
This pull request will be opened after these [changes](https://github.com/tempesta-tech/linux-6.12.12-tfw/pull/3) are agreed upon.

lib/common.h
```diff
@@ -20,13 +20,37 @@
 #ifndef __LIB_COMMON_H__
 #define __LIB_COMMON_H__
 
-/* Get current timestamp in secs. */
+#include <linux/interrupt.h>
+
+#ifdef CONFIG_SECURITY_TEMPESTA
```
EvgeniiMekhanik: We don't need it here.

lib/common.h
```diff
@@ -20,13 +20,37 @@
...
+	 * Use kernel-cached timestamp when in softirq context.
+	 * The timestamp is updated once per softirq batch in handle_softirqs().
+	 * This provides significant performance improvement by avoiding multiple
+	 * expensive ktime_get_real_ts64() calls per softirq batch.
+	 */
+	if (likely(in_serving_softirq()))
+		return softirq_current_timestamp();
+	
+	/* Fallback to direct call outside softirq context */
+	struct timespec64 ts;
```
EvgeniiMekhanik: Variable should be declared at the beginning of the function.

PR #2444, opened 2025-06-11
Fix kernel bugs in case of call `tcp_done`.

### LINKED ISSUE #2443: Kernel BUG in case of error in `tcp_push_pending_framers`
Tempesta FW calls  `tcp_push_pending_framers` during handling WINDOW_UPDATE frame. If window became greater that zero we call it to continue send HTTP2 data. If error occurs during this function (in `sk_fill_write_queue` or `tcp_write_xmit` -> `tfw_handle_error`->`tcp_done` will be called. In this case we release socket and drop connection, but is still in use! This leads to kernel bug.
Tests: simulate error in  `tcp_push_pending_framers` which is called during WINDOW_UPDATE processing.

EvgeniiMekhanik:
We call `tcp_push_pending_frames` from Tempesta FW directly in `ss_do_send/ss_do_close/ss_do_shutdown` functions and during WINDOW_UPDATE frame processing if HTTP2 window became greater then zero. In all this cases `tcp_done` should not be called in case of error in `tcp_push_pending_frames`. To prevent it we set special flag `SOCK_TEMPESTA_IN_USE`, when this flag is set `tw_handle_error` hust set socket state to TCP_CLOSE and clear xmit timers not call `tcp_done`. Later Tempesta FW checks socket state and if it is `TCP_CLOSE` close socket and drop connection.

Closes #2443

### DIFF
--- a/fw/http_frame.c
+++ b/fw/http_frame.c
@@ -698,13 +698,13 @@ tfw_h2_wnd_update_process(TfwH2Ctx *ctx)
 		if (*window > 0) {
 			if (ctx->sched.root.active_cnt) {
 				sock_set_flag(sk, SOCK_TEMPESTA_HAS_DATA);
-				sock_set_flag(sk, SOCK_TEMPESTA_IN_USE);
-				tcp_push_pending_frames(sk);
-				sock_reset_flag(sk, SOCK_TEMPESTA_IN_USE);
+				SS_IN_USE_PROTECT({
+					tcp_push_pending_frames(sk);
+				});
 			}
 		}
 
-		return likely(sk->sk_state != TCP_CLOSE) ? T_OK : sk->sk_err;
+		return likely(!ss_sock_is_closed(sk)) ? T_OK : sk->sk_err;
 	}
 
 fail:
--- a/fw/http_frame.c
+++ b/fw/http_frame.c
@@ -683,6 +683,7 @@ tfw_h2_wnd_update_process(TfwH2Ctx *ctx)
 	wnd_incr = ntohl(*(unsigned int *)ctx->rbuf) & ((1U << 31) - 1);
 	if (wnd_incr) {
 		TfwH2Conn *conn = container_of(ctx, TfwH2Conn, h2);
+		struct sock *sk = ((TfwConn *)conn)->sk;
 		long int *window = ctx->cur_stream ?
 			&ctx->cur_stream->rem_wnd : &ctx->rem_wnd;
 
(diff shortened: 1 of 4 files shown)


fw/http_frame.c
```diff
@@ -696,13 +697,14 @@ tfw_h2_wnd_update_process(TfwH2Ctx *ctx)
...
-				tcp_push_pending_frames(((TfwConn *)conn)->sk);
+				sock_set_flag(sk, SOCK_TEMPESTA_HAS_DATA);
+				sock_set_flag(sk, SOCK_TEMPESTA_IN_USE);
+				tcp_push_pending_frames(sk);
+				sock_reset_flag(sk, SOCK_TEMPESTA_IN_USE);
 			}
 		}
 
-		return T_OK;
+		return likely(sk->sk_state != TCP_CLOSE) ? T_OK : sk->sk_err;
```
krizhanovsky: Let's not call Linux socket functions directly from the HTTP layer - this breaks the layering. Please move the functions to sock.[ch]

fw/sock.c
```diff
@@ -493,13 +495,25 @@ ss_do_send(struct sock *sk, struct sk_buff **skb_head, int flags)
 	 * If packets_out is not equal to zero `tcp_push_pending_frames`
 	 * will be called later from `tcp_data_snd_check` when we receive
 	 * ack from the peer.
+	 *
```
krizhanovsky: extra empty line

fw/sock.c
```diff
@@ -493,13 +495,25 @@ ss_do_send(struct sock *sk, struct sk_buff **skb_head, int flags)
...
 	 */
+	sock_set_flag(sk, SOCK_TEMPESTA_IN_USE);
 	if (sock_flag(sk, SOCK_TEMPESTA_HAS_DATA)) {
 		tcp_push_pending_frames(sk);
 	} else {
 		tcp_push(sk, MSG_DONTWAIT, mss, TCP_NAGLE_OFF | TCP_NAGLE_PUSH,
 			 size);
 	}
+	sock_reset_flag(sk, SOCK_TEMPESTA_IN_USE);
```
krizhanovsky: It seems `tcp_push_pending_frames()` surrounded with `SOCK_TEMPESTA_IN_USE` is a designated function. Probably there are 4 functions to protect, so what do you think about a macro like `SS_IN_USE_PROTECT()`, which will call a Linux function with the flag? And please comment the new function(s)/macro that the flag protects against the socket freeing in `tcp_done()`

PR #2445, opened 2025-06-12
Mekhanik evgenii/1411 change scheduler strategy

EvgeniiMekhanik:


fw/http_frame.c
```diff
@@ -2233,50 +2261,56 @@ do {									\
...
-		r = tfw_h2_stream_xmit_process(sk, ctx, stream, !parent,
-					       &snd_wnd);
-
-		/* We don't recalculate deficits of exclusive streams. */
-		if (parent) {
-			deficit = tfw_h2_stream_recalc_deficit(stream);
-			tfw_h2_sched_stream_enqueue(sched, stream, parent,
-						    deficit);
-		} else if (!tfw_h2_stream_is_active(stream)) {
+		stream_is_exclusive =  tfw_h2_stream_is_exclusive(stream);
```
const-t:
```suggestion
		stream_is_exclusive = tfw_h2_stream_is_exclusive(stream);
```

fw/http_stream_sched.c
```diff
@@ -591,54 +526,33 @@ tfw_h2_sched_stream_enqueue(TfwStreamSched *sched, TfwStream *stream,
...
-	u64 deficit;
+	struct list_head *node = &sched->root.active;
 
-	while (node) {
-		TfwStream *stream = eb64_entry(node, TfwStream, sched_node);
+	while (!list_empty(node)) {
+		TfwStream *stream = list_last_entry(node, TfwStream, sched_node);
 
 		if (tfw_h2_stream_is_active(stream)) {
 			T_DBG4("Stream with id (%u) removed from dependency"
```
const-t: Outdated debug meesage, now we don't remove streams from the tree.

fw/http_frame.c
```diff
@@ -1949,55 +1935,53 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
...
 	if (type == HTTP2_CONTINUATION || type == HTTP2_DATA) {
-		it.skb = it.skb_head = stream->xmit.skb_head;
-		if ((r = tfw_http_msg_insert(&it, &data, &frame_hdr_str)))
+		TfwStr dst = {};
+		unsigned int _;
+
+		r = ss_skb_get_room_w_frag(stream->xmit.skb_head,
+					   stream->xmit.skb_head,
+					   data, FRAME_HEADER_SIZE,
+					    &dst, &_);
```
const-t:
```suggestion
		r = ss_skb_get_room_w_frag(stream->xmit.skb_head,
					   stream->xmit.skb_head,
					   data, FRAME_HEADER_SIZE,
					   &dst, &_);
```

fw/http_frame.c
```diff
@@ -2011,22 +1995,77 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
...
+#define ADJUST_BLOCKED_STREAMS_AND_EXIT(len)				\
 do {									\
-	if (*snd_wnd <= FRAME_HEADER_SIZE + TLS_MAX_OVERHEAD)		\
+	/*								\
+	 * If Tempesta FW stop to make frames, because of exceeded	\
+	 * stream->rem_wnd, mark such stream as blocked.		\
+	 */								\
+	ctx->sched.blocked_streams +=					\
+		(stream->rem_wnd <= len && !stream->xmit.is_blocked);	\
+	stream->xmit.is_blocked = stream->rem_wnd <= len; 		\
```
const-t: Maybe something like this? Why do we need check of `!stream->xmit.is_blocked` here, it seems it can't be blocked before calling `ADJUST_BLOCKED_STREAMS_AND_EXIT`? 
```suggestion
	stream->xmit.is_blocked = stream->rem_wnd <= len; 		\
	ctx->sched.blocked_streams += stream->xmit.is_blocked;		\
```
or 
```
	if (stream->rem_wnd <= len) {					\
		stream->xmit.is_blocked = true; 			\
		++ctx->sched.blocked_streams;				\
	}
```

fw/http_frame.c
```diff
@@ -2011,22 +1995,77 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
...
+	max_len -= FRAME_HEADER_SIZE;					\
+	min_len = min(min_to_send, (unsigned long)len);			\
+	frame_length = tfw_h2_calc_frame_length(ctx, stream, type, len,	\
+						max_len); 		\
+	/*								\
+	 * If the lenght of data to send is less then `min_to_send`	\
+	 * use it as a minimum bytes to send.				\
+	 */								\
+	if (frame_length < min_len)					\
+		ADJUST_BLOCKED_STREAMS_AND_EXIT(min_len);		\
```
const-t: We calling `ADJUST_BLOCKED_STREAMS_AND_EXIT()` in the `CALC_FRAME_LENGTH_AND_SET_FRAME_TYPE()`, but `CALC_FRAME_LENGTH_AND_SET_FRAME_TYPE()` called for `HEADERS` frame. It looks that we can block stream during sending `HEADERS` frame if `rem_wnd` is small and `ctx->rsettings.max_frame_sz` is less than 512. It may lead us to crash.

fw/http_stream_sched.c
```diff
@@ -93,86 +84,70 @@ tfw_h2_stream_sched_spin_lock_assert(TfwStreamSched *sched)
 	assert_spin_locked(&((TfwConn *)conn)->sk->sk_lock.slock);
 }
 
+static void
+tfw_h2_stream_sched_insert_per_weight(struct list_head *head,
```
const-t: Maybe `tfw_h2_stream_sched_insert_by_weight` or something similar? Because `per` in this context means `for each`.

fw/http_stream_sched.c
EvgeniiMekhanik: Yes fixed

fw/http_frame.c
EvgeniiMekhanik: Yes fixed

fw/http_stream_sched.c
EvgeniiMekhanik: Yes fixed

fw/http_frame.c
```diff
@@ -2013,22 +1997,78 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
...
+	max_len -= FRAME_HEADER_SIZE;					\
+	min_len = min(min_to_send, (unsigned long)len);			\
+	frame_length = tfw_h2_calc_frame_length(ctx, stream, type, len,	\
+						max_len); 		\
+	/*								\
+	 * If the lenght of data to send is less then `min_to_send`	\
+	 * use it as a minimum bytes to send.				\
+	 */								\
+	if (frame_length < min_len)					\
+		ADJUST_BLOCKED_STREAMS_AND_EXIT(min_len, type);		\
```
const-t, line 2183: I would like to discuss this optimization, I have a doubts. 512 is a pretty large value.

Example:
```
Assume SND_WND=512/min_to_send=512, len=stream->xmit.h_len=2000, TLS_MAX_OVERHEAD=34
max_len = SND_WND - TLS_MAX_OVERHEAD - FRAME_HEADER_SIZE = 512 - 34 - 9 = 469;
min_len = min(min_to_send, len) = 512
frame_length = min3(ctx->rsettings.max_frame_sz, len, max_len) = 469
```

`if (frame_length < min_len)` -> don't send request. Wait for window.
At this place we are waiting for more window, however we could sent 469 bytes of http headers/body and let client to process while we waiting for available window. Maybe I missed something, but I can't see any reason to delay sending. Do we have benchmarks?

fw/http_frame.c
```diff
@@ -2013,22 +1997,78 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
...
+	const unsigned int min_to_send_dflt = 512;
+
+	/*
+	 * Tempesta FW avoid to send frame if the size of frame
+	 * is less then 512 bytes, except when tcp window is less
+	 * then 512 bytes (small mtu) or http2 initial window is
+	 * less then 1024 bytes (client usually sends window update
+	 * frame with a size equal to a half of initial window).
+	 */
+	return min3((unsigned int)snd_wnd, min_to_send_dflt,
```
const-t: We have type casting here, because `tfw_tcp_calc_snd_wnd()` has `unsgined long` return type. Why it have this type if in this function we finding `min` between two `unsigned int`?

EvgeniiMekhanik:
MekhanikEvgenii/1411-change-scheduler-strategy:
http2_general.test_h2_stream_priority.TestStreamPriorityStress.test_stream_priority_stress : 
Ran 1 test in 33.311s
Ran 1 test in 35.152s
Ran 1 test in 37.232s

master:
Ran 1 test in 126.169s
Ran 1 test in 113.844s
Ran 1 test in 121.743s

fw/http_frame.c
EvgeniiMekhanik: I add below the performance results:
small response:
master:
finished in 50.06s, 1537958.50 req/s, 1.15GB/s
finished in 50.07s, 1529491.88 req/s, 1.15GB/s
PR2445:
finished in 50.07s, 1607661.42 req/s, 1.20GB/s
finished in 50.07s, 1590254.26 req/s, 1.19GB/s
big response:
master:
finished in 50.07s, 50731.62 req/s, 4.93GB/s
finished in 50.06s, 51307.80 req/s, 4.98GB/s
PR2445:
finished in 50.06s, 51170.04 req/s, 4.90GB/s
finished in 50.06s, 51582.10 req/s, 4.93GB/s (edited)

EvgeniiMekhanik: We also have a good permanence improvement on  http2_general.test_h2_stream_priority.TestStreamPriorityStress.test_stream_priority_stress

fw/http_frame.c
```diff
@@ -2013,22 +1992,86 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
...
+	*stop = true;							\
+	T_FSM_EXIT();							\
+} while(0)
+
+#define CALC_FRAME_LENGTH_AND_SET_FRAME_TYPE(type, len)			\
+do {									\
+	unsigned int max_len;						\
+	unsigned int min_len;						\
+									\
+	if (*snd_wnd <= FRAME_HEADER_SIZE + TLS_MAX_OVERHEAD) {		\
```
const-t, line 2169: In `tfw_sk_fill_write_queue()` we do `tfw_h2_make_frames()`, however it may exit in this condition `snd_wnd <= FRAME_HEADER_SIZE + TLS_MAX_OVERHEAD` due to lack of the `snd_wnd`. After when we will receive tcp `ack` - `tcp_push_pending_frames()` will be called, but `SOCK_TEMPESTA_HAS_DATA` not set and `tfw_sk_fill_write_queue()` will not be called as expected.

EvgeniiMekhanik: Why it will be false? We check fw_h2_is_ready_to_send in sock_clnt.c and reset flag only if this function returns false.

EvgeniiMekhanik: This flag will be reset only if there is no active streams (no data to send) or http2 window is exceeded. If http2 window is exceeded we receive window update from the client later, set  SOCK_TEMPESTA_HAS_DATA and call tcp_push_pending_frames at the end of the processing received queue

const-t: Yeah, I see. Seems ok

fw/sock.c
```diff
@@ -138,6 +138,21 @@ static struct kmem_cache *ss_cbacklog_cache;
...
 static void ss_linkerror(struct sock *sk, int flags);
 
+/*
+ * In case error occurs when during call `tcp_push_pending_frames` or
+ * some other linux kernel function from Tempesta FW source code, we
+ * just set socket state to TCP_CLOSE in `tcp_tfw_handle_error` and
+ * it is Tempesta FW responsibilty to check socket state and drop
+ * connection and close socket.
+ */
+#define SS_STATE_PROCESS(sk)				\
```
const-t: I would prefer to have a explicit condition instead of macros, because of hiding `return` or at least depict in the name that macros may `return`. But it's up to you and other reviewers.

fw/http_frame.c
```diff
@@ -2013,22 +1992,86 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
...
+		return 0;
+
+	/*
+	 * Tempesta FW avoid to send frame if the size of frame
+	 * is less then 512 bytes, except when mtu is small or http2
+	 * initial window is less then 1024 bytes (client usually sends
+	 * window update frame with a size equal to a half of initial
+	 * window).
+	 */
+	min_to_send = min3(min_to_send_dflt, dst_mtu(dst) >> 4,
```
const-t: You doing `dst_mtu()`, but maybe it would be better to pass `mss_now` to this function to respect tcp headers size, etc?

const-t (approved):
LGTM, but please address the last comments.

fw/http_frame.c
```diff
@@ -2013,22 +1992,86 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
...
+	struct dst_entry *dst = __sk_dst_get(sk);
+	unsigned int min_to_send;
+
+	/* MTU is unknown, we should send any available bytes. */
+	if (unlikely(!dst))
+		return 0;
+
+	/*
+	 * Tempesta FW avoid to send frame if the size of frame
+	 * is less then 512 bytes, except when mtu is small or http2
```
```suggestion
	 * is less than 512 bytes, except when mtu is small or http2
```

fw/http_frame.c
```diff
@@ -2235,50 +2266,54 @@ do {									\
...
+					tfw_h2_stream_clean(ctx, stream);
+				} else {
+					TfwStreamSchedEntry *parent =
+						stream->sched->parent;
+
+					tfw_h2_stream_sched_remove(sched,
+								   stream);
+					tfw_h2_sched_stream_enqueue(sched,
+								    stream,
+								    parent);
```
krizhanovsky: This code is frequently called and I think we can aggregate both the functions to get some advantage, mostly in the propagation functions. This pair of functions is also called from `tfw_h2_stream_sched_move_child()` - we can just rewrite the function and use it here.

Please also make the function names consistent: use `sched` and `stream` in the same order

krizhanovsky (approved):
LGTM with small improvements from the not addressed comments. Just please do the changes in a separate commit.

fw/http_stream_sched.c
```diff
@@ -280,12 +280,36 @@ tfw_h2_stream_sched_has_children(TfwStreamSchedEntry *entry)
...
+
+	tfw_h2_stream_sched_propagate_add_active_cnt(sched, stream);
+}
+
+void
+tfw_h2_stream_sched_reinsert(TfwStreamSched *sched, TfwStream *stream,
+			     TfwStreamSchedEntry *parent)
+{
+	tfw_h2_stream_sched_remove(sched, stream);
+	tfw_h2_stream_sched_enqueue(sched, stream, parent);
```
krizhanovsky: No, the intention behind https://github.com/tempesta-tech/tempesta/pull/2445#discussion_r2328838046 was not just to make the code more beautiful, but actually, to make it _faster_! We should not adjust the weights twice and definitely must not do two propagations, in `tfw_h2_stream_sched_propagate_dec_active_cnt()` and `tfw_h2_stream_sched_propagate_add_active_cnt()`.

fw/http_stream_sched.c
```diff
@@ -483,7 +472,11 @@ tfw_h2_remove_stream_dep(TfwStreamSched *sched, TfwStream *stream)
 	tfw_h2_stream_sched_move_children(sched, stream, parent,
 					  &stream->sched->active,
 					  parent_has_children);
-	BUG_ON(stream->sched->active_cnt);
+
+	if (stream_is_active) {
+		__tfw_h2_stream_sched_propagate_dec_active_cnt(sched, parent,
+							       1);
+	}
```
const-t, line 478:
```suggestion
	if (stream_is_active)
		__tfw_h2_stream_sched_propagate_dec_active_cnt(sched, parent,
							       1);
```

fw/http_stream_sched.c
```diff
@@ -21,35 +21,26 @@
  * H20 uses an O(1) approach described as an Array of Queue. This is the very
```
krizhanovsky, line 21: Please update the copyright year in all the modified files

fw/http_stream_sched.c
```diff
@@ -362,50 +370,40 @@ tfw_h2_add_stream_dep(TfwStreamSched *sched, TfwStream *stream,
...
-		       " new exclusively added strean with id (%u), ctx %px\n",
-		       child->id, deficit, SCHED_PARENT_STREAM(sched, dep),
-		       stream->id, ctx);
-		tfw_h2_stream_sched_move_child(sched, child, stream->sched,
-					       deficit);
+		       " (id %u) of stream with id (%u) to the new exclusively"
+		       " added strean with id (%u), ctx %px\n", child->id,
+		       SCHED_PARENT_STREAM(sched, dep), stream->id, ctx);
+		tfw_h2_stream_sched_remove(sched, child);
+		tfw_h2_stream_sched_enqueue(sched, child, stream->sched);
```
krizhanovsky, line 392: Why in this function we need the 2 double propagation loops and can't use `tfw_h2_stream_sched_reinsert()`?

fw/http_stream_sched.c
```diff
@@ -429,20 +426,13 @@ tfw_h2_stream_sched_move_children(TfwStreamSched *sched, TfwStream *stream,
...
 		T_DBG3("During removing stream with id (%u) from dependency"
-		       " tree, move its child (id %u new deficit %llu)"
-		       " to the new parent stream with id (%u), ctx %px\n",
-		       stream->id, child->id, deficit,
+		       " tree, move its child (id %u) to the new parent stream"
+		       " with id (%u), ctx %px\n", stream->id, child->id,
 		       SCHED_PARENT_STREAM(sched, parent), ctx);
-		tfw_h2_stream_sched_move_child(sched, child, parent, deficit);
+		__tfw_h2_stream_sched_remove(sched, child);
+		__tfw_h2_stream_sched_enqueue(sched, child, parent);
```
krizhanovsky:
```suggestion
		__tfw_h2_stream_sched_reinsert(sched, child, parent);
```

EvgeniiMekhanik: Fixed

fw/http_stream_sched.c
EvgeniiMekhanik: Fixed

fw/http_stream_sched.c
EvgeniiMekhanik: We can use `tfw_h2_stream_sched_reinsert` only if `stream->parent` is the  * same as `parent` or was removed from sheduler. Add comment about it

EvgeniiMekhanik: It's not a problem, because we use only during adding/changing priority of the stream not making frames. It's rather rare case

PR #2447, opened 2025-06-12
Remove unnecessary call of `tfw_hpack_exp_hdr`.

EvgeniiMekhanik:
There is no sence to call `tfw_hpack_exp_hdr` in `BUFFER_VAL_OPEN` macros, we should just reinit `it->hdr`. Old descriptors in `it->hdr` for header's name can be just forgotten as they are not needed any more, since the name is parsed by now, and all necessary information is already in `it->parsed_hdr`. Removing the extra tfw_hpack_exp_hdr` call has two advantages:
- We can reduce size of `TfwMsgParseIter` structure.
- Intersection between of usage `req->pool` in `tfw_hpack_exp_hdr` and during `it->parsed_hdr` allocation can lead to memory reallocation in `tfw_pool_alloc_np`.

### DIFF
--- a/fw/hpack.c
+++ b/fw/hpack.c
@@ -330,7 +330,6 @@ do {								\
 do {								\
 	(it)->hdr.data = (it)->pos;				\
 	(it)->hdr.len = 0;					\
-	(it)->next = 0;						\
 } while (0)
 
 #define	BUFFER_NAME_OPEN(length)				\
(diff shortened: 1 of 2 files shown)


const-t (approved):
LGTM

krizhanovsky (approved):
LGTM

PR #2448, opened 2025-06-13
ISSIE 2408: system verification script and netconsole configuration

symstu-tempesta:


scripts/tempesta.sh
```diff
@@ -461,6 +478,7 @@ while :; do
 		# User CLI.
 		--start)
 			validate_net_devices "$2" "$3"
+			validate_system_configuration_and_run_netconsole
```
RomanBelozerov, line 506: We want to disable this check using the cmd option, or am I wrong?

scripts/tempesta.sh
```diff
@@ -443,6 +449,17 @@ validate_num_of_opt()
 	fi
 }
 
+validate_system_configuration_and_run_netconsole()
```
RomanBelozerov, line 477: I think we should call this method for `start`, `restart`  and `reload` actions. But there is no need to restart the netconsole. You can split the logic in 2 methods.

scripts/troubleshooting/system_verification.py
```diff
@@ -0,0 +1,746 @@
...
+            ),
+            errors=sysctl_params_errors,
+        )
+
+    verification_passed = all({system_check_ok, kernel_config_ok, sysctl_config_ok})
+
+    if verification_passed:
+        tempesta_log('Verification passed successfully!')
+
+    elif verification_passed:
```
RomanBelozerov:
```suggestion
    else:
```

RomanBelozerov (approved):

scripts/tempesta.sh
```diff
@@ -461,6 +483,7 @@ while :; do
 		# User CLI.
 		--start)
 			validate_net_devices "$2" "$3"
+			validate_system_configuration_and_run_netconsole
```
krizhanovsky, line 506: We don't ship the troubleshooting servers with the community version, so tempesta.sh should accept a CLI argument like `--escudo` to enable the functionality. Otherwise, community users will just get errors on the troubleshooting server unavailable

scripts/troubleshooting/system_verification.py
```diff
@@ -0,0 +1,746 @@
+#!/usr/bin/python3
+"""
+Command-line Interface to verify minimal OS and hardware requirements
+"""
```
krizhanovsky, line 4: Please add a copyright header as we do this in other our sources

scripts/troubleshooting/system_verification.py
```diff
@@ -0,0 +1,746 @@
...
+        """
+        Read command line arguments
+
+        :return: key-value arguments
+        """
+        parser = argparse.ArgumentParser(
+            description="Check if OS and Machine is suitable with minimal Tempesta requirements, start Netconsole",
+            epilog='./system_verification.py -nh=192.168.0.100 -nni=eth0 -th=192.168.0.101 -tm=00:00:00:00:00:00',
+            add_help=True
+        )
```
krizhanovsky, line 104: Netconsole isn't required, especially for the community edition

scripts/troubleshooting/tests.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def test_line_in_middle_of_comment(self):
+        test_config = Config(self.config_path)
+        self.assertTrue(test_config.check_params_are_same({
+            'secondline': {'1'},
+        }))
+
+
+if __name__ == '__main__':
+    unittest.main()
```
krizhanovsky: Shouldn't we move the testing script to tempesta-test ?

scripts/tempesta.sh
symstu-tempesta: These parameters are not required because regular users won't be able to configure them properly — the necessary values are unknown to them. Users should only see notifications about certain problems. BUT, it's possible to configure netconsole and it's work

scripts/troubleshooting/system_verification.py
symstu-tempesta: These parameters are not required because regular users won't be able to configure them properly — the necessary values are unknown to them. Users should only see notifications about certain problems

scripts/troubleshooting/tests.py
symstu-tempesta: Hmm, you're probably right — this might actually be better. But it might be a bit difficult to integrate with our testing framework, since these are unit tests. Maybe it's better to move this test to another directory within the project

const-t (approved):

PR #2452, opened 2025-06-19
Rework restoring `req->method` from hpack dynamic table.

### LINKED ISSUE #2362: Function `tfw_http_meth_str2id` is totally wrong

EvgeniiMekhanik:
Previously we use `tfw_http_meth_strid` to restore req->method from hpack dynamic table, but this function was uneffective and wrong. Moreover we should not use this function at all - when we restore some other `req` fields from hpack dynamic table we use `parser->cstate`, and we should use it for restoring `method` field also.

Closes #2362

### DIFF
--- a/fw/hpack.c
+++ b/fw/hpack.c
@@ -1363,23 +1363,14 @@ tfw_hpack_hdr_set(TfwHPack *__restrict hp, TfwHttpReq *__restrict req,
 done:
 	switch (entry->tag) {
 	case TFW_TAG_HDR_H2_METHOD:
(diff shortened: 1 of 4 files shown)


krizhanovsky (approved):
I do like that this PR simplifies the logic to make is smaller and faster, however a review from @const-t is required.

const-t (approved):
LGTM

PR #2454, opened 2025-06-27
fix clang-format according coding style

MorganaFuture:
fix clang-format according coding style

### DIFF
--- a/pkg/debian/tempesta-fw.service
+++ b/pkg/debian/tempesta-fw.service
@@ -11,6 +11,12 @@ Environment="TLS_PATH=/usr/src/tempesta/tls"
 Environment="LIB_PATH=/usr/src/tempesta/lib"
 Environment=TFW_CFG_PATH=/etc/tempesta/tempesta_fw.conf
 Environment="TFW_SYSTEMD=1"
+Environment="TFW_NETCONSOLE_HOST="
+Environment="TFW_NETCONSOLE_PORT="
+Environment="TFW_NETCONSOLE_NI="
+Environment="TFW_TROUBLESHOOTING_HOST="
+Environment="TFW_TROUBLESHOOTING_PORT="
+Environment="TFW_TROUBLESHOOTING_MAC="
 RemainAfterExit=yes
 ExecStart=/lib/tempesta/scripts/tempesta.sh --start
 ExecStopPost=/lib/tempesta/scripts/tempesta.sh --stop
--- a/scripts/tempesta.sh
+++ b/scripts/tempesta.sh
@@ -42,6 +42,12 @@ logger_path=${LOGGER_PATH:="$TFW_ROOT/utils"}
 tfw_cfg_path=${TFW_CFG_PATH:="$TFW_ROOT/etc/tempesta_fw.conf"}
 tfw_cfg_temp=${TFW_CFG_TMPL:="$TFW_ROOT/etc/tempesta_tmp.conf"}
 tfw_logger_config="$TFW_ROOT/etc/tfw_logger.json"
+tfw_netconsole_host="$TFW_NETCONSOLE_HOST"
+tfw_netconsole_port="$TFW_NETCONSOLE_PORT"
+tfw_netconsole_ni="$TFW_NETCONSOLE_NI"
+tfw_troubleshooting_host="$TFW_TROUBLESHOOTING_HOST"
+tfw_troubleshooting_port="$TFW_TROUBLESHOOTING_PORT"
+tfw_troubleshooting_mac="$TFW_TROUBLESHOOTING_MAC"
 
 tfw_logger_should_start=0
 tfw_logger_pid_path="/var/run/tfw_logger.pid"
@@ -463,6 +469,22 @@ validate_num_of_opt()
 	fi
 }
 
+validate_system_configuration()
+{
+  python3 "$script_path/troubleshooting/system_verification.py"
+}
+
+validate_system_configuration_and_run_netconsole()
+{
+  python3 "$script_path/troubleshooting/system_verification.py" \
+   -th "$tfw_troubleshooting_host" \
+   -tp "$tfw_troubleshooting_port" \
+   -tm "$tfw_troubleshooting_mac" \
+   -nh "$tfw_netconsole_host" \
+   -np "$tfw_netconsole_port" \
+   -nni "$tfw_netconsole_ni"
+}
+
 args=$(getopt -o "d:" -a -l "$LONG_OPTS" -- "$@")
 eval set -- "${args}"
 while :; do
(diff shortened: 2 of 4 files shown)


MorganaFuture:
The after and before files will be deleted, this is just a demonstration of how clang works

utils/clang-format-test/before.cc
```diff
@@ -0,0 +1,227 @@
...
+goto error;                                            \
+(*block)[ind]->As<col_type>()->Append(                         \
+    *reinterpret_cast<const val_type *>(p));                   \
+} else                                                                 \
+(*block)[ind]->As<col_type>()->Append(0);
+
+int main() {
+// All these issues should be fixed by clang-format
+return 0;
+}
```
krizhanovsky: I'm not sure if we need to have and keep the broken C++ code in the main tree. For some stuff to look and play around, we have https://github.com/tempesta-tech/blog/ repo

MorganaFuture: I'll remove them, it's just done to make it easier to understand how clang works now for review

utils/clang-format-test/after.cc
```diff
@@ -0,0 +1,255 @@
...
+	badly_formatted_function(int a, int b, int c)
+	{
+		if (a > 0) {
+			std::cout << "Value is positive" << std::endl;
+			return a + b + c;
+		} else if (a < 0)
+			return -1; // This should be on separate lines
+		else
+			return 0;
+	}
```
const-t:
```suggestion
		if (a > 0) {
			std::cout << "Value is positive" << std::endl;
			return a + b + c;
		} 
		else if (a < 0) {
			return -1; // This should be on separate lines
		}
		else {
			return 0;
		}
	}
```

keshonok: Technically, there's no need in the last `else {}`. You can just `return 0`. Unless this is a matter of strict code style.

utils/clang-format-test/after.cc
```diff
@@ -0,0 +1,255 @@
...
+	while (!stop_flag) {
+		try {
+			spdlog::debug("Worker {} connecting to ClickHouse at "
+				      "{}:{}, table: {}",
+				      ncpu,
+				      ch_cfg.host,
+				      ch_cfg.port,
+				      ch_cfg.table_name);
+			TfwClickhouse clickhouse(
+			    ch_cfg.host,
```
const-t: Broken alignment, need to set proper settings for clang.

utils/clang-format-test/after.cc
```diff
@@ -0,0 +1,255 @@
...
+    [TFW_MMAP_LOG_RESP_CONT_LEN] = {"response_content_length",
+				    clickhouse::Type::UInt32},
+};
+
+void
+run_thread(const int ncpu, const int fd) noexcept
+try {
+	static thread_local std::chrono::seconds timeout(reconnect_min_timeout);
+	cpu_set_t				 cpuset;
+	int					 r;
```
const-t: Usually we don't format local variables this way.

utils/clang-format-test/after.cc
```diff
@@ -0,0 +1,255 @@
...
+	int  *ptr2;
+	void *ptr3;
+
+};
+
+int
+func_test(int argc, char *argv[])
+{
+	TestClass obj;
+	int	  result = obj.badly_formatted_function(1, 2, 3);
```
const-t:
```suggestion
	int result = obj.badly_formatted_function(1, 2, 3);
```

utils/clang-format-test/before.cc
const-t: Yes, these files are very helpful for review, we can just drop them before merge.

utils/clang-format-test/after.cc
```diff
@@ -0,0 +1,255 @@
...
+
+	// Function calls with bad parameter alignment
+	obj.another_function(5, 10);
+
+	return 0;
+}
+
+typedef struct {
+	const char	      *name;
+	clickhouse::Type::Code code;
```
const-t:
```suggestion
	const char	       *name;
	clickhouse::Type::Code code;
```

utils/clang-format-test/after.cc
```diff
@@ -0,0 +1,255 @@
...
+				r = pthread_setaffinity_np(
+				    current_thread, sizeof(cpu_set_t), &cpuset);
+				if (r != 0)
+					throw Except(
+					    "Failed to set CPU affinity");
+			}
+		} catch (const std::exception &e) {
+			if (!uncritical_error) {
+				log_error(e.what(), true, true);
+				timeout		 = reconnect_min_timeout;
```
const-t:
```suggestion
				timeout = reconnect_min_timeout;
```

const-t (changes requested):

utils/clang-format-test/after.cc
keshonok: Looking at the Linux kernel source code I can see that the construct with the last `else {}` is used there in similar cases. So, I think this is the preferred way. Perhaps, the explicitness adds clarity.

MorganaFuture: Now it should comply with the code style.

utils/clang-format-test/after.cc
MorganaFuture: Fixed

utils/clang-format-test/after.cc
MorganaFuture: Fixed

utils/clang-format-test/after.cc
MorganaFuture: Fixed

utils/clang-format-test/after.cc
MorganaFuture: Fixed

utils/clang-format-test/after.cc
MorganaFuture: Partially matches, but for some reason instead of
```

function(arg1, arg2,
         arg3, arg4);
```

clang formats it as

```
function(
arg1, arg2,
arg3, arg4);
```

MorganaFuture: So the problem with this behavior is the 80 character limit: https://github.com/tempesta-tech/tempesta/pull/2454/files#diff-f51b0b97b75ed93ef0ecd2de4801f028328659ba1ba0e3d0438db67ab2604f85R84-R88

const-t: clang format it like so when column limit is reached and we can't configure this

const-t (approved):
LGTM

utils/clang-format-test/before.cc
MorganaFuture: Removed files

PR #2455, opened 2025-06-30
Morgana future/refactor/rename utils to logger

MorganaFuture:
rename utils to logger:
    
- Replace hardcoded batching constants with JSON-configurable parameters
- Add max_events and max_wait_ms settings to ClickHouseConfig
- Fix table creation to use dynamic table_name instead of hardcoded "access_log"
- Update tempesta_fw.conf documentation with new logger_config syntax
- Add comprehensive JSON configuration example in etc/tfw_logger.json


etc/tempesta_fw.conf
```diff
@@ -1408,30 +1408,41 @@
...
+# 'mmap' outputs the access log to a memory-mapped buffer at /dev/tempesta_mmap_log.
+#        When 'mmap' is used, a separate tfw_logger daemon must be configured
+#        to read from the buffer and send logs to storage (e.g., ClickHouse).
+# 'logger_config=<CONFIG_FILE>' specifies the path to JSON configuration file
+#                               for tfw_logger daemon (optional, used with mmap).
+#                               If not specified, tfw_logger uses default settings.
+#
+# When 'mmap' mode is enabled, the kernel module creates a memory-mapped buffer
+# that the tfw_logger daemon reads from. The tfw_logger configuration (either
+# via logger_config or /etc/tfw_logger.json) defines where logs are sent:
```
const-t: `either via logger_config or /etc/tfw_logger.json` Now we can't configure tfw_logger via `logger_config` directive. We only can specify config file. Am I right?

const-t (changes requested):
Rejected to discuss of moving `mmap_log_buffer_size` to json config. Current implementation requires some improvements.

etc/tempesta_fw.conf
MorganaFuture: Now we use logger_config only for set path for json config

logger/clickhouse.cc
```diff
@@ -57,7 +50,14 @@ TfwClickhouse::TfwClickhouse(const std::string &host, const std::string &table_n
 
 	client_ = std::make_unique<clickhouse::Client>(opts);
 
-	client_->Execute(table_creation_querry);
+	std::string table_creation_query = "CREATE TABLE IF NOT EXISTS " + table_name_ +
```
const-t: Using raw `table_name_` introduces SQL injection here. I believe we should sanitize `table_name_` before the first usage and for now we can do this during parsing configuration. 
1. We can allow only `_`,`A-Za-z`,`0-9`.
2. We should limit the name by the length. Clickhouse relies on filesystem's file name length. I think 128 will be enough.

const-t: About style: I suggesting to make `table_creation_query` constexpr string_view stored in format of `fmt::format` and then just do `fmt::format(table_creation_query, table_name_);`

const-t (changes requested):

logger/clickhouse.cc
MorganaFuture: Done

MorganaFuture: Done

const-t (approved):
LGTM

krizhanovsky (approved):
LGTM

Issue #1715, opened 2022-10-11
Slow DoS prevention


b3b:
# Motivation

Because of high memory usage per each connection, possible denial of service exists when large resources are consumed by slow clients.

Related issue: https://github.com/tempesta-tech/tempesta/issues/498

# Testing

#### Host1 (slow clients):
Start a lot of slow (1 byte per second) downloads:
  `curl --output /dev/null -H 'Connection: close' --parallel-max 999 --parallel --parallel-immediate 'http://tempesta-host/[1-10000]' --limit-rate 1B`

#### Host2 (legitimate client):
Try to access resource: `curl -v --output /dev/null http://tempesta-host`

Depending on the load, result could be:
- Empty response, client hangs for some time, and receive [FIN,ACK] from Tempesta:
  ```curl: (52) Empty reply from server```
- Partial response is downloaded and connection is closed by the Tempesta side [RST, ACK]:
```
  { [1132 bytes data]
  * Recv failure: Connection reset by peer
  * Closing connection 0
  curl: (56) Recv failure: Connection reset by peer
```

#### Tempesta

- `cat /proc/net/sockstat` show high value of TCP memory usage
- `ss -l` show high `Send-Q` values
- DMESG contain messages: `TCP: out of memory -- consider tuning tcp_mem`

###### tempesta.cfg
```
listen 80 proto=http;
server 127.0.0.1:8000;
cache 0;
```

#### Backend

Backend on port 8000 should return a large response.
Tested on 200MB, larger responses could trigger #1714

### Scenario for quick reproduction

- Reduce TCP memory bound: `sysctl -w net.ipv4.tcp_mem='100 100 100'`
- Start first (slow) client: `curl -v --output /dev/null  http://127.0.0.1 --limit-rate 1B`
- While first client is running, start second client: `curl -v --output /dev/null  http://127.0.0.1`
- Second client hangs until the first client has finishing downloading.

b3b:
> @b3b I'm wondering why `client_header_timeout` and `client_body_timeout` (https://github.com/tempesta-tech/tempesta/wiki/HTTP-security) aren't used in the Tempesta config to prevent the attack?

`client_header_timeout` and `client_body_timeout` was not used to demonstrate the problem of high memory usage per connection.
And these headers alone will not completely remove the problem, but may disrupt large downloads of legitimate users.

b3b:
On a testing host, memory consumption for sockets when only single slow client is connected:
```shell
[...]

Issue #1346, opened 2019-08-29
HTTTP/2 flood prevention


krizhanovsky:
## Scope

[Low-Rate Denial-of-Service Attacks against HTTP/2 Services](https://www.researchgate.net/publication/280591671_Low-Rate_Denial-of-Service_Attacks_against_HTTP2_Services) paper studies HTTP/2 floods using `PING` and `WINDOW_UPDATE` frames. It's not good to fight against the attacks using packet rate limit since a client can send large POST requests for example. From the other hand requests rate in such attacks could be zero. So a new limits *with burst*, just the same as for current request limiting, must be developed.

## Testing

The test must be developed as part of #673 .

## Documentation

Please update https://github.com/tempesta-tech/tempesta/wiki/HTTP-security#frang-security-limits-enforcing-module with the reference to the paper.

RomanBelozerov:
> The HTTP limits must be extended by stream_rate, stream_burst and concurrent_streams just similar to equal connection limits.

Tempesta should send SETTINGS_MAX_CONCURRENT_STREAMS in SETTINGS frame when initializing connection with client.

EvgeniiMekhanik:
During developing this task we should don't forget to change
```C
if (closed_streams->num <= TFW_MAX_CLOSED_STREAMS
    || (max_streams == ctx->streams_num
	&& closed_streams->num))
```
in `tfw_h2_closed_streams_shrink()`

By @EvgeniiMekhanik  : It was already fixed long time ago. https://github.com/tempesta-tech/tempesta/commit/14cd37b67932b9fe248bea98666be9bc6f4e5b81. Moreover this comment is incorrect. During fixing this problem it was found that we should use another condition.

krizhanovsky:
Related CVE and security advisory https://github.com/advisories/GHSA-93p3-5r25-4p75

const-t:
1. First of all we must not implement `stream_rate`, `stream_burst`, we have `request_rate/burst` that works the same.
2. `response_timeout` it's very doubtful. It can help us only if limit whole connection lifetime, without any updates during connection lifetime. For example: Slow read client opens connection with 100 streams, stream 1 sends request, Tempesta scheduled it to upstream. Start `response_timeout`, if response timeout exceeded, reset connection.
Cons of this method:   
    1. We always must know how many time will take transferring of largest file with slowest client.
    2. It's per-connection not per-stream.

[...]

PR #2456, opened 2025-07-01
Mekhanik evgenii/fix 1346 1

EvgeniiMekhanik:


fw/ss_skb.c
```diff
@@ -1335,10 +1339,16 @@ ss_skb_init_for_xmit(struct sk_buff *skb)
 	struct skb_shared_info *shinfo = skb_shinfo(skb);
 	__u8 pfmemalloc = skb->pfmemalloc;
 
-	WARN_ON_ONCE(skb->sk);
+	skb_orphan(skb);
```
EvgeniiMekhanik: Please pay attention on this place. Here we release skb owner and decrease client->mem. This function `ss_skb_init_for_xmit` is called before push skb to the socket write queue. So all skbs in socket write queue are not taken into account for client memory calculation. We release skb owner here, because if don't do it we need a rather big kernel patch to adjust skb memory before it will be passed to socket write queue. @krizhanovsky @const-t  what do you think about it?

fw/http_limits.h
```diff
@@ -189,5 +189,6 @@ int frang_tls_handler(TlsCtx *tls, int state);
 int frang_sticky_cookie_handler(TfwHttpReq *req);
 bool frang_req_is_whitelisted(TfwHttpReq *req);
 int frang_http_hdr_limit(TfwHttpReq *req, unsigned int new_hdr_len);
+void frang_adjust_client_mem(struct sock *sk, int delta);
```
const-t: Unused function

fw/sock_clnt.c
```diff
@@ -692,18 +694,60 @@ tfw_cfgop_max_concurrent_streams(TfwCfgSpec *cs, TfwCfgEntry *ce)
...
+		default:
+			/* Should be checked early. */
+			BUG();
+		}
+	}
+
+	if (!tfw_cli_hard_limit) {
+		tfw_cli_hard_limit = (tfw_cli_soft_limit < U64_MAX / 2) ?
+			tfw_cli_soft_limit * 2 : U64_MAX;
+	}
```
const-t, line 746: Case where soft limit is greater than hard limit is not validated `client_mem 100m 20m;`

fw/ss_skb.h
```diff
@@ -72,11 +72,15 @@ struct tfw_skb_cb {
 	on_send_cb_t	on_send;
 	on_tcp_entail_t on_tcp_entail;
 	unsigned int 	stream_id;
+	unsigned char	tls_type;
```
const-t: Please document this field.

fw/client.h
```diff
@@ -48,4 +52,30 @@ void tfw_cli_abort_all(void);
...
+}
+
+static inline void
+tfw_client_get_light(TfwClient *cli)
+{
+	int rc;
+
+	rc = atomic_inc_return(&cli->refcnt);
+	if (rc == 1)
+		tfw_client_obtain(cli->addr, NULL, NULL, NULL);
```
const-t: When calling tfw_client_get_light() we always have a valid pointer to tdb record, I mean when client already obtained. It implies we can do something like this. To avoid record "obtaining" part.

light_get:
```
TfwClientEntry *ent = (TfwClientEntry *)cli;
TdbFRec *rec = ((TdbFRec *)cli) - 1;
tdb_rec_keep(client_db, rec);
spin_lock(&ent->lock);
++ent->users;
spin_unlock(&ent->lock);
```

fw/http.c
```diff
@@ -6437,6 +6446,21 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb,
 		actor = tfw_http_parse_req;
 		req->ja5h.version = TFW_HTTP_JA5H_HTTP_REQ;
 	}
+	/*
+	 * For tls connections we already set `skb->owner` before
+	 * tls decryption.
+	 */
+	if (!skb->sk)
+		ss_skb_set_owner(skb, conn->peer);
```
const-t: It will be called for splitted skbs during processing pipelined requests, therefore we double count memory that already counted during tls decryption.

fw/sock.c
```diff
@@ -575,6 +579,7 @@ ss_send(struct sock *sk, struct sk_buff **skb_head, int flags)
 				r = -ENOMEM;
 				goto err;
 			}
+			ss_skb_set_owner(twin_skb, skb->sk);
```
const-t: I don't know how accurate this counting should be, but I must highlight this part. Here we double count skb's fragments because `pskb_copy_for_clone()` copies only linear data to `twin_skb`.

const-t (changes requested):

fw/http.c
EvgeniiMekhanik: No! we checked that skb->sk is NULL here. If owner was set on previous during tls decryption it will be not NULL. If skb was splitted we decrement count of used memory  in `ss_skb_split`

fw/client.h
EvgeniiMekhanik: Yes fixed

fw/sock.c
```diff
@@ -421,16 +428,21 @@ ss_skb_tcp_entail(struct sock *sk, struct sk_buff *skb, unsigned int mark,
...
-	struct sk_buff *skb, *tail, *next, *to_destroy;
+	struct sk_buff *tail, *next, *to_destroy;
 	unsigned char tls_type = 0;
 	unsigned int mark = 0;
 	void *opaque_data = NULL;
 	void (*destructor)(void *) = NULL;
 	int r;
 
-	while ((skb = ss_skb_dequeue(skb_head))) {
+	while ((*snd_wnd = tfw_tcp_calc_snd_wnd(sk, mss_now))) {
```
const-t, line 498: We calculate window here, but we don't check it in the loop. I mean we can put into a `write_queue` skb with len 200 when `snd_wnd` is 1.

fw/http_frame.c
```diff
@@ -2395,10 +2417,10 @@ do {									\
 }
 
 int
-tfw_h2_make_frames(struct sock *sk, TfwH2Ctx *ctx, unsigned int mss_now)
+tfw_h2_make_frames(struct sock *sk, TfwH2Ctx *ctx, unsigned int mss_now,
+		   unsigned long snd_wnd)
 {
 	TfwStreamSched *sched = &ctx->sched;
-	unsigned long snd_wnd = tfw_tcp_calc_snd_wnd(sk, mss_now);
```
const-t, line 2400: It doesn't related to this commit, but it seems we don't account `FRAME_HEADER_SIZE` and tls overhead in `snd_wnd` in `tfw_h2_stream_xmit_process()`. As I see we do `*snd_wnd -= frame_length;` but `frame_length` doesn't include  `FRAME_HEADER_SIZE` and size of tls records.

fw/sock.c
```diff
@@ -506,51 +522,33 @@ ss_do_send(struct sock *sk, struct sk_buff **skb_head, int flags)
...
+	 *   (`tcp_push_pending_frames`) will be called again.
 	 */
 	SS_IN_USE_PROTECT({
-		if (sock_flag(sk, SOCK_TEMPESTA_HAS_DATA)) {
-			tcp_push_pending_frames(sk);
-		} else {
-			tcp_push(sk, MSG_DONTWAIT, mss,
-				 TCP_NAGLE_OFF | TCP_NAGLE_PUSH, size);
-		}
+		tcp_push_pending_frames(sk);
```
const-t, line 604: Does it make sense to call `tcp_push_pending_frames()` `if !(*skb_head)`?

fw/ss_skb.c
```diff
@@ -1749,11 +1765,12 @@ ss_skb_set_owner(struct sk_buff *skb, void *owner)
 		 * this SKBs will be passed to the socket write queue.
 		 */
 		BUG_ON(skb->sk);
+		BUG_ON(TFW_SKB_CB(skb)->mem != 0);
```
const-t: Should we panic if `mem` is miscalculated? Even if `mem` is greater than it really used client will be disconnected, but not crashed. Maybe convert it to warning?

const-t (changes requested):

fw/http_frame.c
EvgeniiMekhanik: About `FRAME_HEADER_SIZE` i agree, but we can't adjust size of tls records. To adjust it we should make encryption here.

fw/sock.c
EvgeniiMekhanik: Yes! We call `sk_fill_write_queue` from `tcp_push_pending_frames` and move skbs from connection write queue to socket write queue

fw/sock.c
EvgeniiMekhanik: Fixed using `skb_shift`, but it seems that we should not split skb here if there is no snd_wnd and more over we should not do it during making frames. Its better to push more data then snd_wnd then split skb

fw/sock.c
```diff
@@ -397,40 +404,100 @@ ss_forced_mem_schedule(struct sock *sk, int size)
 	sk_memory_allocated_add(sk, amt);
 }
 
+static bool
+ss_skb_can_collapce(struct sk_buff *to, struct sk_buff *from,
```
const-t:
```suggestion
ss_skb_can_collapse(struct sk_buff *to, struct sk_buff *from,
```

fw/sock.c
```diff
@@ -397,40 +404,100 @@ ss_forced_mem_schedule(struct sock *sk, int size)
...
+
+	if ((skb_shinfo(to)->flags & SKBFL_SHARED_FRAG) !=
+	    (skb_shinfo(from)->flags & SKBFL_SHARED_FRAG))
+		return false;
+
+	return true;
+}
+
+static bool
+ss_skb_try_collapce(struct sock *sk, struct sk_buff *skb,
```
const-t:
```suggestion
ss_skb_try_collapse(struct sock *sk, struct sk_buff *skb,
```

fw/t/unit/helpers.c
```diff
@@ -33,6 +33,7 @@
  */
 #include "helpers.h"
 #include "http_msg.h"
+
```
const-t, line 36: Why need blank line here?

const-t (approved):
LGTM

fw/ss_skb.h
```diff
@@ -63,20 +63,30 @@ typedef void (*on_tcp_entail_t)(void *conn, struct sk_buff *skb_head);
...
  * @is_head		- flag indicates that this is a head of skb list;
  */
 struct tfw_skb_cb {
 	void 		*opaque_data;
 	void 		(*destructor)(void *opaque_data);
 	on_send_cb_t	on_send;
 	on_tcp_entail_t on_tcp_entail;
+	long int	mem;
 	unsigned int 	stream_id;
+	unsigned char	tls_type;
```
krizhanovsky: I didn't get the whole https://github.com/tempesta-tech/tempesta/pull/2456/changes/661629903334a75a897ff1ea065af64410476b7c change at all. This PR does not use `skb->dev` for a connection point as it is said in the commit message. Even if we really going to use `skb->dev` for something else, then we must not store TLS type both in `skb->dev` and `skb->cb` - this makes spaghetti code

fw/http2.c
```diff
@@ -647,7 +647,7 @@ tfw_h2_stream_xmit_prepare_resp(TfwStream *stream)
 	BUG_ON(!resp || resp->msg.skb_head || !resp->req
 	       || !resp->req->conn || !stream->xmit.skb_head);
```
krizhanovsky, line 649: It seems comment https://github.com/tempesta-tech/tempesta/issues/1346#issuecomment-1611029794 isn't addressed. Unfortunately the comment doesn't say how exactly `tfw_h2_closed_streams_shrink()` should be changed and why, so I'm not sure if anyone can remember what the comment is about.

fw/connection.h
```diff
@@ -606,6 +609,8 @@ tfw_peer_for_each_conn(TfwPeer *p, int (*cb)(TfwConn *))
 }
 
 extern unsigned int tfw_cli_max_concurrent_streams;
+extern u64 tfw_cli_soft_limit;
+extern u64 tfw_cli_hard_limit;
```
krizhanovsky: Let's add `mem` to the limits since we will have more per-client limits for system resources, e.g. for CPU

fw/sock.c
krizhanovsky: This comment isn't addressed, the code still has `ss_skb_can_collapce` instead of `ss_skb_can_collapse`

fw/sock.c
```diff
@@ -397,40 +404,109 @@ ss_forced_mem_schedule(struct sock *sk, int size)
...
+		skb->mark = mark;
+		if (tls_type)
+			skb_set_tfw_tls_type(skb, tls_type);
+		ss_forced_mem_schedule(sk, skb->truesize);
+		tcp_skb_entail(sk, skb);
+		tp->write_seq += skb->len;
+		TCP_SKB_CB(skb)->end_seq += skb->len;
+
+		return 0;
+	}
```
krizhanovsky, line 486: The change https://github.com/tempesta-tech/tempesta/pull/2456/changes/227b9c1f8366d71aac3b5f34821e59bc87ff7bcc seems a good performance optimization, but it is irrelevant to #1346 and should have been done in a separate PR

fw/ss_skb.c
```diff
@@ -1336,10 +1343,16 @@ ss_skb_init_for_xmit(struct sk_buff *skb)
...
 
-	WARN_ON_ONCE(skb->sk);
+	skb_orphan(skb);
 
 	skb_dst_drop(skb);
 	INIT_LIST_HEAD(&skb->tcp_tsorted_anchor);
+
+	/*
+	 * dev is used to save connection for memory accounting
+	 * clear it before pass skb to the kernel.
```
krizhanovsky, line 1352: I didn't find where we actually store connection in `skb->dev`

fw/connection.c
```diff
@@ -231,3 +232,102 @@ tfw_connection_unlink_to_sk(TfwConn *conn)
...
+		tfw_h2_context_safe(conn) : NULL;
+	if (!h2) {
+		r = ss_skb_tcp_entail_list(sk, &conn->write_queue,
+					   mss_now, &snd_wnd);
+		if (unlikely(r))
+			return r;
+
+		/*
+		 * `snd_wnd` is not 0 if all skbs was entailed to socket
+		 * write queue.
```
krizhanovsky: Apparently we need to fix the comment as well

fw/connection.c
```diff
@@ -231,3 +232,102 @@ tfw_connection_unlink_to_sk(TfwConn *conn)
...
+		 */
+		if (!conn->write_queue) {
```
krizhanovsky: `unlikely`?

fw/connection.h
```diff
@@ -573,6 +575,7 @@ tfw_connection_validate_cleanup(TfwConn *conn)
 	BUG_ON(!conn);
 	BUG_ON(!list_empty(&conn->list));
 	BUG_ON(conn->stream.msg);
+	BUG_ON(conn->write_queue);
```
krizhanovsky, line 584: Is kernel crash inevitable and it's better to crash now? Otherwise it'd be better to print a warning.

fw/ss_skb.c
```diff
@@ -1706,10 +1719,84 @@ int
 ss_skb_realloc_headroom(struct sk_buff *skb)
 {
 	int delta = MAX_TCP_HEADER - skb_headroom(skb);
+	bool skb_has_owner = (skb->sk != NULL);
+	void *owner;
```
krizhanovsky: It'd be enough just to use and check `owner` and we don't need `skb_has_owner` - if `owner` is not `NULL`, then skb does have an owner

fw/ss_skb.c
```diff
@@ -1706,10 +1719,84 @@ int
...
+	 * kernel code). It is not safe for us to not break `skb->truesize`
+	 * calculation here, so we should orphan skb and then restore it's
+	 * owner later.
+	 */
+	if (skb_has_owner) {
+		owner = skb->sk;
+		mem = TFW_SKB_CB(skb)->mem;
+		old_truesize = skb->truesize;
+		skb_orphan(skb);
+	}
```
krizhanovsky: Do we really need to play with the skb owner? This looks risky and basically `client_mem` is our own accounting. Probably we can account expanded head in parallel with the socket/skb accounting or just do not care about `pskb_expand_head()`.

This might open a DoS attack vector, but this is about _mitigation_, so in general we can afford some accounting inaccuracy to get faster and more reliable code.

fw/ss_skb.c
krizhanovsky: Why don't we make a pointer to a client accounting in `skb->cb` instead of to play with `skb_orphan()`? I'd prefer to avoid this since we can get plenty of crashes in this patch or in later kernel version migrations due to breaking kernel logic about orphaned skbs.

fw/http_frame.c
```diff
@@ -2186,7 +2209,7 @@ do {									\
 
 #define FRAME_XMIT_FSM_NEXT(frame_length, state)			\
 do {									\
-	*snd_wnd -= frame_length;					\
+	*snd_wnd -= frame_length - FRAME_HEADER_SIZE;			\
```
krizhanovsky: Why do we now exclude frame header from `snd_wnd` accounting? It'd be good if commit message for https://github.com/tempesta-tech/tempesta/pull/2456/changes/d000665a40b7fe760d560b9b45c6de177871397e would explain this

fw/ss_skb.c
```diff
@@ -1706,10 +1719,84 @@ int
...
+ss_skb_adjust_sk_mem(struct sk_buff *skb, int delta)
+{
+	TfwClient *cli = (TfwClient *)skb->sk;
+
+	/*
+	 * `cli` can be zero here when this function is called
+	 * from `ss_skb_split` for SKBs which are already orphaned
+	 */
+	if (cli) {
+		TFW_SKB_CB(skb)->mem += delta;
```
krizhanovsky, line 1786: Does it ever reach 0? What about introducing a perfstat counter, which is changed right in the function, so at least we can track whether our accounting works at the global scale?

fw/ss_skb.c
```diff
@@ -1706,10 +1719,84 @@ int
...
+		 * We can safely use `skb->sk` for our purposes until
+		 * this SKBs will be passed to the socket write queue.
+		 */
+		BUG_ON(skb->sk);
+		WARN_ON(TFW_SKB_CB(skb)->mem != 0);
+
+		tfw_client_get((TfwClient *)owner);
+		skb->sk = owner;
+		skb->destructor = ss_skb_destructor;
+		ss_skb_adjust_sk_mem(skb, mem);
```
krizhanovsky: `ss_skb_set_owner()`, as the name implies, is about setting an owner, so it must not adjust memory accounting. In particular, it seems there is a bug: `ss_skb_realloc_headroom()` gets `mem = TFW_SKB_CB(skb)->mem` and passes it as an argument, with the delta, to `ss_skb_set_owner()`, which calls `ss_skb_adjust_sk_mem()` and now we add `mem` to `TFW_SKB_CB(skb)->mem`, not only the delata as `skb->truesize - old_truesize`

fw/ss_skb.c
```diff
@@ -1706,10 +1719,84 @@ int
...
+	}
+}
+
+void
+ss_skb_adjust_sk_mem(struct sk_buff *skb, int delta)
```
krizhanovsky: Now the function doesn't adjust socket memory, but client memory instead, so let's rename it

fw/ss_skb.c
```diff
@@ -1706,10 +1719,84 @@ int
...
+		/*
+		 * All SKBs were orphaned when Tempesta FW received them.
+		 * We can safely use `skb->sk` for our purposes until
+		 * this SKBs will be passed to the socket write queue.
+		 */
+		BUG_ON(skb->sk);
+		WARN_ON(TFW_SKB_CB(skb)->mem != 0);
+
+		tfw_client_get((TfwClient *)owner);
+		skb->sk = owner;
```
krizhanovsky: It was clear why do we need to play with skb->sk when we were adjusting `sk->sk_rmem_alloc`, but why do we need this now, when we only adjust `TfwClient->mem`?

fw/connection.c
```diff
@@ -231,3 +232,102 @@ tfw_connection_unlink_to_sk(TfwConn *conn)
...
+	});
+	if (unlikely(sk->sk_state == TCP_CLOSE))
+		return -ENOMEM;
+	return 0;
+}
+
+int
+tfw_connection_push(TfwConn *conn, unsigned int mss_now)
+{
+	struct sock *sk = conn->sk;
```
krizhanovsky, line 264: `conn` dereferencing must be done after `BUG_ON` at the below

fw/connection.c
```diff
@@ -231,3 +232,102 @@ tfw_connection_unlink_to_sk(TfwConn *conn)
...
+	/*
+	 * This function is called under the socket lock, same as dropping a
+	 * connection. Moreover this function is never called when socket
+	 * state is TCP_CLOSE. When client closes the connection, we drop it
+	 * from tcp_done() -> ss_conn_drop_guard_exit(), and socket state is
+	 * set to TCP_CLOSE, so this function will never be called after it.
+         */
+	BUG_ON(!conn);
+
+	BUG_ON(SS_CONN_TYPE(sk) & Conn_Closing);
```
krizhanovsky: Probably this can be just a warning. Also probably we don't need an empty line before this assertion

fw/connection.c
```diff
@@ -231,3 +232,102 @@ tfw_connection_unlink_to_sk(TfwConn *conn)
...
+	}
+
+	/*
+	 * First of all Tempesta FW entails skb from connection write
+	 * queue (control frames, tls alerts and so on), then if
+	 * `snd_wnd` is not exceeded make frames.
+	 */
+	r = ss_skb_tcp_entail_list(sk, &conn->write_queue, mss_now, &snd_wnd);
+	if (unlikely(r))
+		return r;
```
krizhanovsky: This is just the same as for HTTP/1, so let's use the same code for both the protocols

fw/http2.h
```diff
@@ -204,4 +204,10 @@ tfw_h2_is_ready_to_send(TfwH2Ctx *ctx)
 	return ctx->sched.root.active_cnt && ctx->rem_wnd;
 }
 
+static inline bool
+tfw_h2_or_stream_wnd_is_exceeded(TfwH2Ctx *ctx, TfwStream *stream)
```
krizhanovsky: Is there a typo in name? It's unclear where "or" relates to, definitely not to "h2"

fw/http_frame.c
```diff
@@ -2304,13 +2327,13 @@ do {									\
 		if (stream->xmit.h_len) {
 			T_FSM_JMP(HTTP2_MAKE_CONTINUATION_FRAMES);
 		} else {
-			if (stream->xmit.postponed
+			if (unlikely(stream->xmit.postponed)
 			    && !stream->xmit.frame_length
 			    && !ctx->cur_send_headers) {
```
```suggestion
			    && !ctx->cur_send_headers)
			    {
```

fw/sock.c
```diff
@@ -506,51 +588,33 @@ ss_do_send(struct sock *sk, struct sk_buff **skb_head, int flags)
 
 	if (ss_skb_on_send(conn, skb_head))
 		goto cleanup;
+	else if (*skb_head)
```
krizhanovsky:
```suggestion
	if (*skb_head)
```

fw/ss_skb.c
```diff
@@ -1706,10 +1719,84 @@ int
...
+	if (owner) {
+		/*
+		 * All SKBs were orphaned when Tempesta FW received them.
+		 * We can safely use `skb->sk` for our purposes until
+		 * this SKBs will be passed to the socket write queue.
+		 */
+		BUG_ON(skb->sk);
+		WARN_ON(TFW_SKB_CB(skb)->mem != 0);
+
+		tfw_client_get((TfwClient *)owner);
```
krizhanovsky: This is an additional reference count for the TDB entry - were is corresponding decerment of the reference count?

fw/ss_skb.c
```diff
@@ -1706,10 +1719,84 @@ int
...
+	/*
+	 * `cli` can be zero here when this function is called
+	 * from `ss_skb_split` for SKBs which are already orphaned
+	 */
+	if (cli) {
+		TFW_SKB_CB(skb)->mem += delta;
+		BUG_ON(TFW_SKB_CB(skb)->mem < 0);
+		tfw_client_adjust_mem(cli, delta);
+	}
+}
```
krizhanovsky, line 1790: At the moment only this function accounts `TfwClient->mem` and the function accounts `skb` data only. I believe we also need to account HTTP level data as well. For example, a single skb may produce many HTTP messages and each of the message may unpack a lot of data from HPACK.

Not necessary to account everything supper accurately, but the main data consumers must be accounted.

krizhanovsky (changes requested):
Quite a big and sophisticated change, definitely requiring some discussion. Let's discuss the PR on the next meeting. Besides absent logic and bugs, I see following topics to consider/discuss:
1. how the soft limit of `client_mem` is supposed to be implemented? Shouldn't we revert the `sk->sk_rmem_alloc` patch? I did like the approach since it naturally implements the soft limit, but it's unclear how to implement it now
2. It seems for me that there is no sense to play with `skb_orphan()` if we account everything on application level (no `sk->sk_rmem_alloc` and only `TfwClient` accounting)It seems  `skb_orphan()` is required only for more accurate accounting for `pskb_expand_head()`, but does it really impacts much into overall memory consumption?
3. I believe we also should account HTTP level memory
4. There are test pull requests - https://github.com/tempesta-tech/tempesta-test/pull/615 and https://github.com/tempesta-tech/tempesta-test/pull/618 - but they both are in draft. They must be finished along with this PR and they do need to test accounting accuracy and #1715 scenario - it must be fixed with this PR.

Also it seems there are 2 irrelevant commits for the #1346 and #1715 logic and it'd be better to publish such commits separately to make PRs review and merge faster.

Please make all upcoming changes in separate commits and please write good commit messages even for commits supposed to be squashed.

fw/http2.c
EvgeniiMekhanik: It was already fixed long time ago. https://github.com/tempesta-tech/tempesta/commit/14cd37b67932b9fe248bea98666be9bc6f4e5b81. Moreover this comment is incorrect. During fixing this problem it was found that we should use another condition.

fw/ss_skb.h
EvgeniiMekhanik: Yes fixed, was planning to use it previously and forget to remove commit

fw/connection.h
EvgeniiMekhanik: yes fixed

fw/sock.c
EvgeniiMekhanik: yes fixed, also I make some changes in this PR, because now we pass calculate and `snd_wnd` during pushing skb to socket write queue, so we can use `skb_shift` for more effective performance optimization.

fw/connection.c
EvgeniiMekhanik: yes, add `unlikely`

fw/connection.h
EvgeniiMekhanik: If `write_queue` is not empty, memory leak occurs, but I use BUG_ON here because we use BUG_ON in this function for all other cases

fw/ss_skb.c
EvgeniiMekhanik: Lets talk about it, I don't understand comment

fw/http2.h
EvgeniiMekhanik: Yes fixed

fw/http_frame.c
EvgeniiMekhanik: fixed

fw/sock.c
EvgeniiMekhanik: fixed

fw/ss_skb.c
EvgeniiMekhanik: in `ss_skb_destructor`

fw/ss_skb.c
EvgeniiMekhanik: Yes, I will implement it in separate PR, because current PR is too big

fw/ss_skb.c
EvgeniiMekhanik: yes, save all necessary data in skb->cb

fw/ss_skb.c
EvgeniiMekhanik: yes fixed

fw/connection.c
EvgeniiMekhanik: Remove it, because currently it is not necessary at all.

fw/connection.c
EvgeniiMekhanik: yes fixed

fw/connection.c
EvgeniiMekhanik: Yes remove it

fw/ss_skb.c
EvgeniiMekhanik: remove it during reworking

fw/ss_skb.c
EvgeniiMekhanik: `[skb_set_owner_r](https://elixir.bootlin.com/linux/v6.12.12/C/ident/skb_set_owner_r)` look this function in linux kernel code has the same name, and also adjust memory accounting. we call `skb_orphan` so TFW_SKB_CB(skb)->mem will be 0

fw/connection.c
EvgeniiMekhanik: yes fixed

fw/http_frame.c
EvgeniiMekhanik: yes fixed

fw/sock_clnt.c
```diff
@@ -922,6 +939,15 @@ static TfwCfgSpec tfw_sock_clnt_specs[] = {
 		.allow_repeat = false,
 		.allow_reconfig = true,
 	},
+	{
+		.name = "client_mem",
```
const-t, line 994: @EvgeniiMekhanik @krizhanovsky I should notice that currently `client_mem` overrides frang's `http_body_len`. By default Tempesta can receive 1GB request/response, but with enabled `client_mem` it is limited to `client_mem`. It may be surprising during configuration. It seems we can temporary disable `http_body_len` and set `client_mem` to 1GB until #498 is implemented.

fw/pool.h
```diff
@@ -53,18 +53,20 @@ typedef struct tfw_pool_chunk_t {
  * Memory pool descriptor.
  *
  * @curr	- current chunk to allocate memory from;
+ * @owner	- owner for memory accounting;
  * @order,@off	- cached members of @curr;
  */
 typedef struct {
 	TfwPoolChunk	*curr;
+	void		*owner;
```
const-t: Why not forward-declare `TfwClient` struct?

EvgeniiMekhanik:
Big responses
master:
finished in 40.06s, 64391.32 req/s, 6.12GB/s
2e94835a50ff3f0f0c4b7fc08887e32e49c7f0fa:
finished in 40.10s, 63755.38 req/s, 6.11GB/s - no client mem option, just accounting

fw/client.c
```diff
@@ -182,16 +252,39 @@ tfw_client_addr_eq(TdbRec *rec, void *data)
 }
 
 static void
+tfw_client_free(TdbRec *rec)
+{
+        /* Stats should be updated(dec/inc) only for complete records. */
+        if (!tdb_entry_is_complete(rec))
```
const-t: I believe we don't need this condition for clients, we allocate per-cpu memory even for not complete records. At this moment we are safe, however if in future we fail before set record as complete we will have memleak.

fw/client.c
```diff
@@ -182,16 +252,39 @@ tfw_client_addr_eq(TdbRec *rec, void *data)
...
 tfw_client_ent_init(TdbRec *rec, void *data)
 {
 	TfwClientEntry *ent = (TfwClientEntry *)rec->data;
 	TfwClient *cli = &ent->cli;
 	TfwClientEqCtx *ctx = (TfwClientEqCtx *)data;
+	int cpu;
+
+	assert_spin_locked(&client_db->ga_lock);
+
+	cli->mem = tfw_alloc_percpu(long);
```
const-t: `alloc_percpu()` has `GFP_KERNEL` flag by default, in this place we must set `GFP_ATOMIC | __GFP_ZERO` and remove 
```
for_each_online_cpu(cpu)
		*(per_cpu_ptr(cli->mem, cpu)) = 0;
```

fw/http_frame.c
```diff
@@ -1874,6 +1874,11 @@ tfw_h2_frame_process(TfwConn *c, struct sk_buff *skb, struct sk_buff **next)
 		       parsed, skb->len);
 	}
 
+	r = frang_client_mem_limit((TfwCliConn *)c, true);
```
const-t: I'm worrying about this place, without benchmark I can't say how it is crucial, but seems we introduce a lot of inter-cache traffic, that might be significant on NUMA. Even for small frames, service frames, we call `frang_client_mem_limit()`. Maybe we can have a local per-cpu threshold(for instance 4 * default mtu) to check during request processing and to check the remote cpus only when threshold has been reached and then once per request or client processing stage - when all skbs are processed? However the last makes sense only if we close all connections from the client.

fw/tls.c
```diff
@@ -584,6 +588,8 @@ tfw_tls_send(TlsCtx *tls, struct sg_table *sgt)
 					r = -ENOMEM;
 					goto out;
 				}
+				ss_skb_set_owner(skb, ss_skb_dflt_destructor,
```
const-t, line 591: We have a few places like that, we do `ss_skb_set_owner()` where we assign new value to per-cpu. And then `ss_skb_adjust_data_len()` where we also assign new value to per-cpu and  we do this in the loop. By itself assign new value to per-cpu in the loop is very cheap, it is regular assign. However, we have remote access to this `per-cpu` variable that leads to inter-cache traffic even for local writes. Maybe we can assign values to "local" variable and after some threshold assign it to per-cpu? Not only in this place, do like so everywhere.

fw/cache.c
const-t, line 2822: Here we add client mem for plain http1 as well as other protocols, however for plain http1 memory will not be allocated and `ss_skb_to_sgvec_with_new_pages()`will not be called. I'm not ask to change this behavior it is up to you, but I ask to write a comment to describe this.

fw/http.c
```diff
@@ -6540,6 +6584,23 @@ tfw_http_req_process(TfwConn *conn, TfwStream *stream, struct sk_buff *skb,
...
+	/*
+	 * For tls connections we already set `skb->owner` before
+	 * tls decryption.
+	 */
+	if (!TFW_SKB_CB(skb)->opaque_data) {
+		ss_skb_set_owner(skb, ss_skb_dflt_destructor,
+				 conn->peer, skb->truesize);
+	}
+
+	r = frang_client_mem_limit((TfwCliConn *)conn, false);
```
const-t: FYI: We checking limit of memory, before parsing, cache response build, etc. Those may consume a lot of memory, that can lead to interesting side effect: the current response will live during transfer to the client consuming a lot of memory and only new requests will not received from the same client, even if they are not so big. However checking mem limit in the end of req processing even worse.

fw/sock.c
```diff
@@ -681,6 +668,10 @@ ss_send(struct sock *sk, struct sk_buff **skb_head, int flags)
 				r = -ENOMEM;
 				goto err;
 			}
+			memset(twin_skb->cb, 0, sizeof(twin_skb->cb));
+			ss_skb_set_owner(twin_skb, ss_skb_dflt_destructor,
+					 TFW_SKB_CB(skb)->opaque_data,
+					 skb_headlen(skb));
```
const-t:
```suggestion
					 skb_headlen(twin_skb));
```

const-t: Also we don't account `struct sk_buff` and `struct skb_shared_info` in this place, but in other places we do, why?

fw/client.c
EvgeniiMekhanik: Yes fixed

fw/client.c
EvgeniiMekhanik: Yes fixed

fw/tls.c
EvgeniiMekhanik: As I see on SUT, we have no problem with assigning the only one real place which decrease performance (in case when it is enabled) is frang_client_mem_limit, where we sum client memory for all cpus

fw/cache.c
EvgeniiMekhanik: yes fixed

fw/sock.c
EvgeniiMekhanik: Yes fixed

fw/http_frame.c
EvgeniiMekhanik: Yes this only one place, when we decrease performance for 10 percent (if appropriate option is enabled, and we check client memory consumption). @krizhanovsky is it crusial? Should we accuracy check memory consumption and block client or not?

fw/cache.c
```diff
@@ -2816,7 +2819,8 @@ tfw_cache_add_body_page(TfwMsgIter *it, char *p, int sz, bool h2)
 	skb_fill_page_desc(it->skb, it->frag, page, off, sz);
 	if (!h2)
 		skb_frag_ref(it->skb, it->frag);
-	ss_skb_adjust_data_len(it->skb, sz);
+	else
```
const-t: Why `else` if we add fragment for both h2 and h1?

fw/client.c
```diff
@@ -193,16 +188,69 @@ tfw_client_addr_eq(TdbRec *rec, void *data)
...
+
+	percpu_ref_kill_and_confirm(&cli_mem->refcnt, cli_mem_release);
+}
+
+static inline TfwClientMem *
+tfw_client_mem_alloc(void)
+{
+	TfwClientMem *cli_mem;
+
+	cli_mem = tfw_kmalloc(sizeof(TfwClientMem), GFP_ATOMIC);
```
const-t: It would be good to re-use this memory, good for memcache, but as I see, we can't use it, because during shutdown we may destroy the cache before freeing all objects.
 
Also this is untested place. We didn't do benchmarks with many clients so we can't say what overhead it introduces

fw/client.c
```diff
@@ -193,16 +188,69 @@ tfw_client_addr_eq(TdbRec *rec, void *data)
...
+	free_percpu(cli_mem->mem);
+	kfree(cli_mem);
+}
+
+static void
+cli_mem_release(struct percpu_ref *ref)
+{
+	TfwClientMem *cli_mem = container_of(ref, TfwClientMem, refcnt);
+
+	call_rcu(&cli_mem->rcu_head, __cli_mem_release);
```
const-t: Warning on shutdown: Start Tempesta do single request, then stop Tempesta. 
```
[  183.855417] [tempesta fw] Tempesta FW is ready
[  200.349624] [tdb] Close table 'client0.tdb'
[  200.353864] [tdb] Close table 'sessions0.tdb'
[  200.391901] ------------[ cut here ]------------
[  200.392530] WARNING: CPU: 0 PID: 0 at kernel/rcu/tree.c:2631 rcu_core+0x3d9/0x7c0
[  200.393491] Modules linked in: tempesta_fw(OE) tempesta_db(OE) tempesta_tls(OE) tempesta_lib(OE) veth intel_rapl_msr intel_rapl_common xt_conntrack xt_MASQUERADE nft_masq xfrm_user xfrm_algo nft_chain_nat nf_nat snd_hda_codec_generic nf_conntrack snd_hda_intel xt_addrtype nf_defrag_ipv6 nft_compat nf_defrag_ipv4 snd_intel_dspcfg bridge snd_intel_sdw_acpi snd_hda_codec stp llc snd_hda_core nf_tables overlay qxl snd_hwdep kvm_amd snd_pcm i2c_i801 drm_ttm_helper snd_timer ccp cfg80211 ttm i2c_mux snd binfmt_misc kvm joydev input_leds i2c_smbus lpc_ich drm_kms_helper soundcore virtiofs mac_hid serio_raw sch_fq_codel dm_multipath efi_pstore drm nfnetlink dmi_sysfs qemu_fw_cfg ip_tables x_tables autofs4 btrfs blake2b_generic raid10 raid456 async_raid6_recov async_memcpy async_pq async_xor async_tx xor raid6_pq libcrc32c raid1 raid0 crct10dif_pclmul hid_generic crc32_pclmul ghash_clmulni_intel virtio_net usbhid ahci sha512_ssse3 net_failover psmouse virtio_scsi sha256_ssse3 virtio_rng libahci failover hid virtio_blk
[  200.393538]  aesni_intel
[  200.397304] [tdb] Close table 'cache0.tdb'
[  200.413234]  crypto_simd cryptd [last unloaded: tempesta_lib(OE)]
[  200.413249] CPU: 0 UID: 0 PID: 0 Comm: swapper/0 Tainted: G           OE      6.12.12+ #225
[  200.413257] Tainted: [O]=OOT_MODULE, [E]=UNSIGNED_MODULE
[  200.413257] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS Arch Linux 1.17.0-2-2 04/01/2014
[  200.413259] RIP: 0010:rcu_core+0x3d9/0x7c0
[  200.413264] Code: 02 00 00 80 7c 24 2f 00 0f 85 51 03 00 00 48 85 c0 0f 84 c4 02 00 00 84 d2 74 11 48 8b 7c 24 08 e8 ec 33 00 00 48 85 c0 75 02 <0f> 0b 41 f7 c6 00 02 00 00 74 05 e8 97 99 ff ff 48 8b 7c 24 08 e8
[  200.413270] RSP: 0018:ffffc1b080005f10 EFLAGS: 00010046
[  200.413277] RAX: 0000000000000000 RBX: ffff9c2a182373b8 RCX: 00000000802a0028
[  200.413278] RDX: 0000000000000000 RSI: 0000000000000001 RDI: ffff9c2a182373b8
[  200.413279] RBP: ffff9c2a18237340 R08: ffff9c26c44f3de0 R09: 00000000802a0028
[  200.413279] R10: 00000000802a0028 R11: ffffffffbd74a628 R12: ffffffffbd60a940
[  200.413280] R13: 00000000000355c8 R14: 0000000000000246 R15: ffffffffffffffff
[  200.413283] FS:  0000000000000000(0000) GS:ffff9c2a18200000(0000) knlGS:0000000000000000
[  200.413285] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[  200.413286] CR2: 000055aea04216a8 CR3: 0000000100a96000 CR4: 0000000000750ef0
[  200.413289] PKRU: 55555554
[  200.413291] Call Trace:
[  200.413305]  <IRQ>
[  200.413309]  ? __warn+0x89/0x140
[  200.413320]  ? rcu_core+0x3d9/0x7c0
[  200.413322]  ? report_bug+0x164/0x1a0
[  200.413344]  ? handle_bug+0x58/0xa0
[  200.413361]  ? exc_invalid_op+0x17/0x80
[  200.413363]  ? asm_exc_invalid_op+0x1a/0x20
[  200.413378]  ? rcu_core+0x3d9/0x7c0
[  200.413380]  ? rcu_core+0x269/0x7c0
[  200.413381]  handle_softirqs+0xd9/0x2e0
[  200.413400]  __irq_exit_rcu+0x63/0x80
[  200.413404]  sysvec_apic_timer_interrupt+0x71/0xa0
[  200.413421]  </IRQ>
[  200.413423]  <TASK>
[  200.413423]  asm_sysvec_apic_timer_interrupt+0x1a/0x20
[  200.413436] RIP: 0010:pv_native_safe_halt+0xf/0x20
[  200.413440] Code: 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 f3 0f 1e fa eb 07 0f 00 2d 55 f5 39 00 fb f4 <c3> cc cc cc cc 66 66 2e 0f 1f 84 00 00 00 00 00 90 90 90 90 90 90
[  200.413440] RSP: 0018:ffffffffbd603e88 EFLAGS: 00000206
[  200.413442] RAX: ffff9c2a18200000 RBX: ffffffffbd60a940 RCX: 0000000000000000
[  200.413443] RDX: 0000000000000000 RSI: 0000000000000000 RDI: 000000000012dd64
[  200.413443] RBP: 0000000000000000 R08: 0000000000000001 R09: 0000000000000002
[  200.413443] R10: 0000000000000001 R11: 0000000000000000 R12: 0000000000000000
[  200.413444] R13: 0000000000000000 R14: ffffffffbd60a040 R15: 000000000008a000
[  200.413446]  ? ct_kernel_exit.constprop.0+0x5d/0x80
[  200.413448]  default_idle+0x9/0x20
[  200.413449]  default_idle_call+0x30/0x100
[  200.413451]  do_idle+0x1fb/0x240
[  200.413473]  cpu_startup_entry+0x29/0x40
[  200.413476]  rest_init+0xcc/0xe0
[  200.413477]  start_kernel+0x61b/0x8a0
[  200.413533]  x86_64_start_reservations+0x18/0x40
[  200.413560]  x86_64_start_kernel+0x7a/0x80
[  200.413563]  common_startup_64+0x13e/0x141
[  200.413584]  </TASK>
[  200.413586] ---[ end trace 0000000000000000 ]---
[  200.492377] ------------[ cut here ]------------
[  200.493101] WARNING: CPU: 0 PID: 1419 at kernel/rcu/tree.c:2628 rcu_core+0x70e/0x7c0
[  200.494063] Modules linked in: tempesta_fw(OE) tempesta_db(OE) tempesta_tls(OE) tempesta_lib(OE) veth intel_rapl_msr intel_rapl_common xt_conntrack xt_MASQUERADE nft_masq xfrm_user xfrm_algo nft_chain_nat nf_nat snd_hda_codec_generic nf_conntrack snd_hda_intel xt_addrtype nf_defrag_ipv6 nft_compat nf_defrag_ipv4 snd_intel_dspcfg bridge snd_intel_sdw_acpi snd_hda_codec stp llc snd_hda_core nf_tables overlay qxl snd_hwdep kvm_amd snd_pcm i2c_i801 drm_ttm_helper snd_timer ccp cfg80211 ttm i2c_mux snd binfmt_misc kvm joydev input_leds i2c_smbus lpc_ich drm_kms_helper soundcore virtiofs mac_hid serio_raw sch_fq_codel dm_multipath efi_pstore drm nfnetlink dmi_sysfs qemu_fw_cfg ip_tables x_tables autofs4 btrfs blake2b_generic raid10 raid456 async_raid6_recov async_memcpy async_pq async_xor async_tx xor raid6_pq libcrc32c raid1 raid0 crct10dif_pclmul hid_generic crc32_pclmul ghash_clmulni_intel virtio_net usbhid ahci sha512_ssse3 net_failover psmouse virtio_scsi sha256_ssse3 virtio_rng libahci failover hid virtio_blk
[  200.494102]  aesni_intel crypto_simd cryptd [last unloaded: tempesta_lib(OE)]
[  200.506042] CPU: 0 UID: 0 PID: 1419 Comm: kworker/0:9 Tainted: G        W  OE      6.12.12+ #225
[  200.507170] Tainted: [W]=WARN, [O]=OOT_MODULE, [E]=UNSIGNED_MODULE
[  200.508032] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS Arch Linux 1.17.0-2-2 04/01/2014
[  200.509228] Workqueue: dio/vda2 iomap_dio_complete_work
[  200.509976] RIP: 0010:rcu_core+0x70e/0x7c0
[  200.510571] Code: 8b 15 f6 98 b8 01 48 f7 da 48 85 d2 7e 33 48 8b 55 78 48 85 d2 0f 84 aa fc ff ff eb 31 48 8b 45 78 48 85 c0 0f 85 bc fc ff ff <0f> 0b e9 c6 fc ff ff 48 89 ce 4c 89 ff e8 a0 a4 eb 00 e9 d3 f9 ff
[  200.513008] RSP: 0018:ffffc1b080005f10 EFLAGS: 00010046
[  200.513732] RAX: 0000000000000000 RBX: ffff9c2a182373b8 RCX: ffff9c2a18237460
[  200.514667] RDX: ffffffffffffd8f0 RSI: 0000000000000001 RDI: ffff9c2a182373b8
[  200.515746] RBP: ffff9c2a18237340 R08: 0000000000000001 R09: 7fffffffffffffff
[  200.516730] R10: ffffffffbd6060c0 R11: 00000000000ecef5 R12: ffff9c26c4264000
[  200.517710] R13: 00000000000355c8 R14: 0000000000000246 R15: ffffffffffffffff
[  200.518768] FS:  0000000000000000(0000) GS:ffff9c2a18200000(0000) knlGS:0000000000000000
[  200.519894] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[  200.520685] CR2: 000055aea04216a8 CR3: 0000000101ff2000 CR4: 0000000000750ef0
[  200.521624] PKRU: 55555554
[  200.522073] Call Trace:
[  200.522551]  <IRQ>
[  200.522977]  ? __warn+0x89/0x140
[  200.523538]  ? rcu_core+0x70e/0x7c0
[  200.524075]  ? report_bug+0x164/0x1a0
[  200.524634]  ? handle_bug+0x58/0xa0
[  200.525168]  ? exc_invalid_op+0x17/0x80
[  200.525742]  ? asm_exc_invalid_op+0x1a/0x20
[  200.526347]  ? rcu_core+0x70e/0x7c0
[  200.526883]  ? rcu_core+0x269/0x7c0
[  200.527509]  ? __hrtimer_run_queues+0x141/0x2a0
[  200.528166]  handle_softirqs+0xd9/0x2e0
[  200.528779]  __irq_exit_rcu+0x63/0x80
[  200.529383]  sysvec_apic_timer_interrupt+0x71/0xa0
[  200.530061]  </IRQ>
[  200.530440]  <TASK>
[  200.530827]  asm_sysvec_apic_timer_interrupt+0x1a/0x20
[  200.531580] RIP: 0010:_raw_spin_unlock_irqrestore+0x1d/0x40
[  200.532338] Code: 90 90 90 90 90 90 90 90 90 90 90 90 90 f3 0f 1e fa 0f 1f 44 00 00 e8 b2 0a 00 00 90 f7 c6 00 02 00 00 74 06 fb 0f 1f 44 00 00 <65> ff 0d a4 81 5b 43 74 05 c3 cc cc cc cc 0f 1f 44 00 00 c3 cc cc
[  200.534759] RSP: 0018:ffffc1b082567e58 EFLAGS: 00000206
[  200.535534] RAX: 0000000000000001 RBX: ffff9c26c0ce9f00 RCX: ffff9c26c867d368
[  200.536465] RDX: 0000000000000000 RSI: 0000000000000287 RDI: ffff9c26c867d360
[  200.537508] RBP: ffff9c26c3888800 R08: 0000000000000001 R09: 0000000000000002
[  200.538542] R10: 0000000000000000 R11: 0000000000000002 R12: 0000000000004000
[  200.539603] R13: ffff9c26c4264000 R14: ffff9c270dd39160 R15: 0000000000000000
[  200.540532]  ? _raw_spin_unlock_irqrestore+0xe/0x40
[  200.541209]  aio_complete_rw+0xdb/0x1c0
[  200.541787]  process_one_work+0x16d/0x380
[  200.542380]  worker_thread+0x2cb/0x3e0
[  200.543032]  ? __pfx_worker_thread+0x20/0x20
[  200.543689]  kthread+0xcf/0x100
[  200.544178]  ? __pfx_kthread+0x20/0x20
[  200.544720]  ret_from_fork+0x31/0x60
[  200.545246]  ? __pfx_kthread+0x20/0x20
[  200.545780]  ret_from_fork_asm+0x22/0x60
[  200.546345]  </TASK>
[  200.546707] ---[ end trace 0000000000000000 ]---
[  200.565583] [tdb] Close table 'filter0.tdb'
[  200.566166] [tempesta fw] modules are stopped
[  200.693391] [tempesta fw] exiting...
[  200.805140] [tdb] Shutdown Tempesta DB
```

fw/client.c
```diff
@@ -193,16 +188,69 @@ tfw_client_addr_eq(TdbRec *rec, void *data)
...
+__cli_mem_release(struct rcu_head *rcu)
+{
+	TfwClientMem *cli_mem = container_of(rcu, TfwClientMem, rcu_head);
+
+	free_percpu(cli_mem->mem);
+	kfree(cli_mem);
+}
+
+static void
+cli_mem_release(struct percpu_ref *ref)
```
const-t: It seems this function will be called twice, the first time from `percpu_ref_call_confirm_rcu()` when `per-cpu` will be switched to atomic calling `percpu_ref_kill_and_confirm()` and the second time when refcounter reached zero in `percpu_ref_put_many()`. Please check it.

EvgeniiMekhanik: Yes fixed this and trace

fw/client.c
EvgeniiMekhanik: I thnik that new client initialization is much more rare than if we use atomic_inc?dec on each skb allocation for client reference counting

fw/client.c
```diff
@@ -65,6 +65,9 @@ static struct {
 
 static TDB *client_db;
 
+static atomic_t shutdown_pending = ATOMIC_INIT(0);
```
const-t, line 71: static variable doesn't require zero initialization.

fw/client.c
```diff
@@ -193,16 +194,73 @@ tfw_client_addr_eq(TdbRec *rec, void *data)
...
+
+	cli_mem = tfw_kmalloc(sizeof(TfwClientMem), GFP_ATOMIC);
+	if (unlikely(!cli_mem))
+		return NULL;
+
+	cli_mem->mem = tfw_alloc_percpu_gfp(long, GFP_ATOMIC | __GFP_ZERO);
+	if (!cli_mem->mem)
+		goto free_cli_mem;
+
+	if (percpu_ref_init(&cli_mem->refcnt, cli_mem_release, 0, GFP_ATOMIC))
```
const-t: I can't see call `percpu_ref_exit()` we should do this, to free `data`. I believe we can call it from `release` callback.

fw/client.c
const-t: > I thnik that new client initialization is much more rare than if we use atomic_inc?dec on each skb allocation for client reference counting

Indeed, however it depends on the size of lru, if we have a lot of unique clients we may have performance spikes. It is just warning from my side, I'm not sure that we need to change this especially without benchmark.

@krizhanovsky @EvgeniiMekhanik Anyway in this function `tfw_client_mem_alloc()` we have two allocations, one allocates `TfwClientMem` and the second inside `percpu_ref_init()`.

fw/client.c
```diff
@@ -193,16 +194,73 @@ tfw_client_addr_eq(TdbRec *rec, void *data)
...
+	free_percpu(cli_mem->mem);
+	kfree(cli_mem);
+}
+
+static void
+cli_mem_release(struct percpu_ref *ref)
+{
+	TfwClientMem *cli_mem = container_of(ref, TfwClientMem, refcnt);
+
+	call_rcu(&cli_mem->rcu_head, __cli_mem_release);
```
const-t: It seems we are safe to call `__cli_mem_release()` right here, instead of `call_rcu()`

fw/client.c
```diff
@@ -65,6 +65,9 @@ static struct {
 
 static TDB *client_db;
```
const-t, line 70: @krizhanovsky @EvgeniiMekhanik All this memory counting introduce some overhead, maybe it worth not to allocate all per-cpu staff and so on if `client_mem` is zero? It anyway can't be reconfigured now.

const-t (changes requested):

fw/client.c
EvgeniiMekhanik: yes, but we do it in other places. see sock.c for example

fw/client.c
EvgeniiMekhanik: Yes fixed. This is left over from my other idea)

fw/client.c
EvgeniiMekhanik: yes fixed

fw/client.c
const-t: In some places we do, in some don't. In access_log.c we don't do.

fw/pool.h
const-t: Using forward declaration for `TfwClient` we can use type name here instead of `void *`, like we do for other structures, see `fw/http_types.h"

fw/client.c
keshonok: Technically, you don't know if a macro results in a pure zero value. Moreover, you don't need to know that. You just need to use the macro where appropriate. I believe the use of the macro here is justified and correct.

fw/client.c
const-t: @EvgeniiMekhanik I believe with wait queue, that was added in the last patch we can use memcache instead of kmalloc here.

fw/pool.h
EvgeniiMekhanik: Yes fixed

fw/client.c
EvgeniiMekhanik: I agree with @keshonok. I think that it is good to use this macros everywhere. There is an api and we should use it

fw/client.c
EvgeniiMekhanik: yes fixed

fw/client.c
EvgeniiMekhanik: Yes fixed, moreover preallocated pool was implemeneted

lib/log.h
```diff
@@ -22,48 +22,100 @@
...
-	T_COMPRESSION	= -MAX_ERRNO + 7,
+	/* The message looks good and we can safely pass it. */
+	T_OK		 = 0,
+
 	/*
-	 * Generic error. Connection should be shutdown gracefully
-	 * with TCP_FIN.
+	 * Common error codes high boundary. All common error codes
+	 * should be less then this boundary. Error codes should be listed
+	 * from least crusial to most crusial.
```
const-t, line 43:
```suggestion
	 * from least crucial to most crucial.
```

lib/log.h
```diff
@@ -22,48 +22,100 @@
...
 };
 
+static inline bool
+is_tfw_error_code(int err_code)
+{
+	return err_code < __T_ERROR_CODES_BOUNDARY;
+}
+
+static inline bool
+tfw_error_code_more_crusial(int err_code1, int err_code2)
```
const-t, line 101:
```suggestion
tfw_error_code_more_crucial(int err_code1, int err_code2)
```

const-t, line 101: nit: From my point of view calling this function is more sophisticated than simple `err1 < err2`.

lib/log.h
```diff
@@ -22,48 +22,100 @@
...
+	 * Tempesta send TCP FIN in this case.
+	 */
+	T_BLOCK_WITH_FIN = __T_COMMON_ERROR_CODE_MAX - 5,
+
+	/*
+	 * Internal error codes high boundary. All internal error codes
+	 * should be less then this boundary. Error codes should be listed
+	 * from least crusial to most crusial.
+	 */
+	__T_INTERNAL_ERROR_CODE_MAX = -MAX_ERRNO + T_INTERNAL_ERROR_CODE_MAX,
```
const-t, line 79: Do we need `__T_INTERNAL_ERROR_CODE_MAX`? It is unclear for me what difference between common and internal error codes and why `T_BLOCK` is internal but `T_BLOCK_WITH_RST` is common.

lib/log.h
```diff
@@ -22,48 +22,100 @@
...
+	 * is in range.
+	 */
+	BUILD_BUG_ON(T_BLOCK_WITH_FIN >= MAX_ERRNO);
+	BUILD_BUG_ON(T_BLOCK <= __T_COMMON_ERROR_CODE_MAX);
+
+	return err_code1 < err_code2;
+}
+
+static inline bool
+tfw_error_code_is_crusial(int err_code)
```
const-t, line 114:
```suggestion
tfw_error_code_is_crucial(int err_code)
```

fw/client.c
```diff
@@ -58,12 +59,175 @@ typedef struct {
...
+	percpu_ref_kill(&cli_mem->refcnt);
+	percpu_ref_put(&cli_mem->refcnt);
+}
+
+static inline int
+tfw_cli_mem_init(TfwClientMem *cli_mem)
+{
+	int r;
+
+	cli_mem->mem = tfw_alloc_percpu_gfp(long, GFP_KERNEL | __GFP_ZERO);
```
const-t, line 166: You calling `tfw_cli_mem_init()` from softIRQ as well, but you use only `GFP_KERNEL` flags. For `tfw_percpu_ref_init()` the same.

fw/client.c
```diff
@@ -58,12 +59,175 @@ typedef struct {
...
+tfw_cli_mem_pool_init(void)
+{
+	TfwClientMem *block;
+	int i, r;
+
+	if (WARN_ON_ONCE(!client_cfg.lru_size))
+		return -EINVAL; 
+
+	cli_mem_pool.order =
+		get_order(sizeof(TfwClientMem) * client_cfg.lru_size);
```
const-t, line 212: For large LRUs we potentially can fail and not allocate memory, I believe we must allocate not greater than `MAX_PAGE_ORDER`.

fw/client.c
EvgeniiMekhanik: yes fixed

fw/client.c
EvgeniiMekhanik: yes fixed

lib/log.h
EvgeniiMekhanik: According our private discussion we decide to use this function

fw/client.c
```diff
@@ -159,16 +159,18 @@ tfw_cli_mem_kill_work_fn(struct work_struct *work)
 }
 
 static inline int
-tfw_cli_mem_init(TfwClientMem *cli_mem)
+tfw_cli_mem_init(TfwClientMem *cli_mem, bool from_pool)
```
const-t: Maybe just pass flags would be better? But it is not important I'm ok with it

const-t (approved):
LGTM, however review from @krizhanovsky is required, especially to approve new approach with memory limit checking.

fw/sock_clnt.c
krizhanovsky: @EvgeniiMekhanik @const-t yes, this definitely makes sense. The 1GB limit was just a workaround for #498 and it makes sense to completely remove it and replace with `client_mem`. Just make a temporary limit for the value as 1GB

fw/http.c
krizhanovsky: @const-t is this comment still relevant? I see for example, that `ss_skb_adjust_data_len()` is called from`tfw_hpack_cache_decode_expand()` ->  `tfw_hpack_cache_decode_expand()` -> `tfw_http_msg_expand_data()`, i.e the current code does account HTTP responses

fw/http_frame.c
krizhanovsky: " adjust FRAME_HEADER_SIZE during calculation send window during making frames." does not explain why the header frame size is subtracted. We do not have capacity to fix every small thing, but please **do put attention to commit messages and code commenting**. This is super important especially if such a sophisticated work as this PR and whole 1346

fw/connection.h
krizhanovsky: The point was to comment this **in code, not in PR**

fw/client.c
```diff
@@ -341,7 +342,6 @@ tfw_client_stop(void)
 	if (tfw_runstate_is_reconfig())
 		return;
 
-	tfw_client_free_lru();
```
krizhanovsky: It seems the content of commit https://github.com/tempesta-tech/tempesta/pull/2456/changes/a3ad1762d9fa13ca9d9ef9f2571b371c2bdd74bd and its message are about different things

krizhanovsky (approved):
LGTM for now. I listed all required to dos in https://github.com/tempesta-tech/tempesta/issues/1715#issuecomment-4313947521 and made a small commit to adjust coding style to not to pollute the comment. Let's address the comments in a new PR - it becomes to hard to review this PR - let's just merge it for now.

PR #2339, opened 2025-02-07
Migrate to linux 6.12.12

const-t:

PR #2457, opened 2025-07-02
Optimize tfw_current_timestamp() with per-CPU caching

MorganaFuture:
Replace heavyweight ktime_get_real_ts64() calls with per-CPU cached
  timestamps during softirq processing. Cache is refreshed every second
  and uses atomic jiffies snapshot for consistency.

### DIFF
--- a/lib/common.h
+++ b/lib/common.h
@@ -23,34 +23,63 @@
 #include <linux/percpu.h>
 #include <linux/time.h>
 
-static DEFINE_PER_CPU(struct timespec64, tfw_cached_ts);
-static DEFINE_PER_CPU(unsigned long, tfw_ts_last_update);
-
 #define TFW_TS_REFRESH_INTERVAL (HZ)
 
 /**
+ * Get current timestamp with ktime_get_real_ts64 interface.
+ * Uses per-CPU cached timestamps in softirq context.
+ * WARNING: Must be called from softirq context only.
+ */
+static inline void
+tfw_current_timestamp_ts64(struct timespec64 *ts)
+{
+	static DEFINE_PER_CPU(struct timespec64, tfw_cached_ts);
+	static DEFINE_PER_CPU(unsigned long, tfw_ts_last_update);
+
+	WARN_ON_ONCE(!in_softirq());
+
+	struct timespec64 *cached_ts = this_cpu_ptr(&tfw_cached_ts);
+	unsigned long *last_update = this_cpu_ptr(&tfw_ts_last_update);
+	unsigned long now = jiffies;
+
+	if (unlikely(time_after(now,
+				*last_update + TFW_TS_REFRESH_INTERVAL)))
+	{
+		ktime_get_real_ts64(cached_ts);
+		*last_update = now;
+	}
+
+	*ts = *cached_ts;
+}
+
+/**
  * Get current timestamp in seconds.
- * Uses cached value if available and not stale when in softirq context.
+ * Uses per-CPU cached timestamps in softirq context.
+ * WARNING: Must be called from softirq context only.
  */
 static inline long
 tfw_current_timestamp(void)
 {
-	if (likely(in_serving_softirq())) {
-		struct timespec64 *cached_ts = this_cpu_ptr(&tfw_cached_ts);
-		unsigned long *last_update = this_cpu_ptr(&tfw_ts_last_update);
-		unsigned long now = jiffies;
-		
-		if (unlikely(time_after(now, *last_update + TFW_TS_REFRESH_INTERVAL))) {
-			ktime_get_real_ts64(cached_ts);
-			*last_update = now;
-		}
-		
-		return cached_ts->tv_sec;
-	} else {
-		struct timespec64 ts;
-		ktime_get_real_ts64(&ts);
-		return ts.tv_sec;
-	}
+	struct timespec64 ts;
+
+	tfw_current_timestamp_ts64(&ts);
+	return ts.tv_sec;
+}
+
+/**
+ * Get current timestamp - real-time version.
+ * For use outside of softirq context or when precise real-time is needed.
+ * WARNING: Must be called from process context only.
+ */
+static inline struct timespec64
+tfw_current_timestamp_real(void)
+{
+	struct timespec64 ts;
+
+	WARN_ON_ONCE(in_softirq());
+
+	ktime_get_real_ts64(&ts);
+	return ts;
 }
 
 #endif /* __LIB_COMMON_H__ */
--- a/lib/common.h
+++ b/lib/common.h
@@ -20,13 +20,37 @@
 #ifndef __LIB_COMMON_H__
 #define __LIB_COMMON_H__
 
-/* Get current timestamp in secs. */
+#include <linux/percpu.h>
+#include <linux/time.h>
+
+static DEFINE_PER_CPU(struct timespec64, tfw_cached_ts);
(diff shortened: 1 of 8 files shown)


lib/common.h
```diff
@@ -20,13 +20,37 @@
...
+		
+		if (unlikely(time_after(now, *last_update + TFW_TS_REFRESH_INTERVAL))) {
+			ktime_get_real_ts64(cached_ts);
+			*last_update = now;
+		}
+		
+		return cached_ts->tv_sec;
+	} else {
+		struct timespec64 ts;
+		ktime_get_real_ts64(&ts);
```
EvgeniiMekhanik: According codestyle you should add empty line after variables declaration

lib/common.h
```diff
@@ -20,13 +20,37 @@
...
 {
-	struct timespec64 ts;
-	ktime_get_real_ts64(&ts);
-	return ts.tv_sec;
+	if (likely(in_serving_softirq())) {
+		struct timespec64 *cached_ts = this_cpu_ptr(&tfw_cached_ts);
+		unsigned long *last_update = this_cpu_ptr(&tfw_ts_last_update);
+		unsigned long now = jiffies;
+		
+		if (unlikely(time_after(now, *last_update + TFW_TS_REFRESH_INTERVAL))) {
```
EvgeniiMekhanik: Too long line. Maximum 80 symbols are allowed

lib/common.h
```diff
@@ -20,13 +20,37 @@
 #ifndef __LIB_COMMON_H__
 #define __LIB_COMMON_H__
 
-/* Get current timestamp in secs. */
+#include <linux/percpu.h>
+#include <linux/time.h>
+
+static DEFINE_PER_CPU(struct timespec64, tfw_cached_ts);
```
EvgeniiMekhanik: It seems that it's better to declare this variables inside the function, because they used only in `tfw_current_timestamp`

lib/common.h
```diff
@@ -20,13 +20,37 @@
...
+static DEFINE_PER_CPU(unsigned long, tfw_ts_last_update);
+
+#define TFW_TS_REFRESH_INTERVAL (HZ)
+
+/**
+ * Get current timestamp in seconds.
+ * Uses cached value if available and not stale when in softirq context.
+ */
 static inline long
 tfw_current_timestamp(void)
```
EvgeniiMekhanik, line 64: Is there some places where this function can be called both in softirq and not in softirq context? If no may be split it into two functions and call that one that expected? @const-t how do you think?

EvgeniiMekhanik:
It seems that we apply this patch to current master because there is no changes in kernel.

lib/common.h
const-t: Yes, we should have two functions

const-t: @MorganaFuture Please replace call of `ktime_get_real_ts64` in access log by this cacheable version.

lib/common.h
MorganaFuture: Fixed

lib/common.h
MorganaFuture: Fixed

lib/common.h
MorganaFuture: Done

lib/common.h
MorganaFuture: Done

lib/common.h
```diff
@@ -20,13 +20,58 @@
...
+{
+static DEFINE_PER_CPU(struct timespec64, tfw_cached_ts);
+static DEFINE_PER_CPU(unsigned long, tfw_ts_last_update);
+
+	struct timespec64 *cached_ts = this_cpu_ptr(&tfw_cached_ts);
+	unsigned long *last_update = this_cpu_ptr(&tfw_ts_last_update);
+	unsigned long now = jiffies;
+
+	if (unlikely(time_after(now,
+				*last_update + TFW_TS_REFRESH_INTERVAL))) {
```
const-t:
```suggestion
	if (unlikely(time_after(now,
				*last_update + TFW_TS_REFRESH_INTERVAL))) 
	{
```

const-t:
> Can someone please remind me why this is better than:
> 
>     1. Get and cache per-CPU timer value in timer softirq which runs at HZ rate.
> 
>     2. Get the current cached timer value unconditionally in Tempesta code.
> 
> 
> Thanks!

We are trying to minimize the use of timers because we’ve had issues with them. I realize that using a single, simple timer in this case may not cause problems, but we chose this approach because solution with timers also introduce overhead. For example, it requires inter-cache communication and additional effort to maintain the timer itself. Anyway, the performance difference between these solutions is negligible.

EvgeniiMekhanik (approved):

lib/common.h
```diff
@@ -20,13 +20,63 @@
...
 }
 
+/**
+ * Get current timestamp - real-time version.
+ * For use when precise real-time is needed without caching.
+ */
+static inline void
+tfw_current_timestamp_real(struct timespec64 *ts)
+{
+	ktime_get_real_ts64(ts);
```
EvgeniiMekhanik, line 80: Please add `WARN_ON_ONCE(!in_task())`

lib/common.h
```diff
@@ -20,13 +20,63 @@
 #ifndef __LIB_COMMON_H__
 #define __LIB_COMMON_H__
 
-/* Get current timestamp in secs. */
+#include <linux/percpu.h>
```
krizhanovsky, line 23: Please update the copyright year

tls/tls_ticket.c
```diff
@@ -351,7 +357,8 @@ ttls_tickets_configure(TlsPeerCfg *cfg, unsigned long lifetime,
 	}
 
 	timer_setup(&tcfg->timer, ttls_ticket_rotate_keys, 0);
-	secs = tcfg->lifetime - (tfw_current_timestamp() % tcfg->lifetime);
+	tfw_current_timestamp_real(&ts);
```
krizhanovsky, line 369: These are definitely not all the places where we get timestamps, just today I commented a similar place in #2339. @const-t it seems #2339 will be merged later, so could you please fix this in your PR, but @MorganaFuture please check the whole FW, DB and TLS code for timestamps

tls/ttls.c
```diff
@@ -2850,6 +2850,15 @@ ttls_alpn_ext_eq(const ttls_alpn_proto *proto, const unsigned char *buf,
...
 #if DBG_TLS >= 3
+/**
+ * Debug version of time function that returns incremental counter
+ * instead of real time. Useful for deterministic testing and debugging.
+ * 
+ * NOTE: Currently not used as all ttls_time() calls were replaced with
+ * tfw_current_timestamp(). Kept for potential future debugging needs.
+ * To use it, replace tfw_current_timestamp() calls in TLS code with
+ * ttls_time_debug() when DBG_TLS >= 3.
+ */
```
krizhanovsky: Let's remove the function - if we ever need it, we can fetch it from git

tls/tls_ticket.c
MorganaFuture: Replaced. In the FW and DB modules, I didn’t find any usage of timestamps or anything similar. Only jiffies are used, which is already optimal as a kernel global variable.

tls/ttls.c
MorganaFuture: Done

lib/common.h
MorganaFuture: Added

Issue #2450, opened 2025-06-16
L7 DDoS mitigation utilities


krizhanovsky:
**Description is TODO**

Probably we need to split out some generic code to be able to write more such scripts.

- [x] Block by
 - ja5t and ja5h
 - IP addresses (issue IPset rules instead of Tempesta rules. Check if we can use IPset also with nftables)
- [x] Number of fetched JA5 hashes (10 for now) or IP addresses must be configurable, e.g. CLI argument
- [x] Time slice (60 seconds for now) also must be configurable, e.g. CLI argument
- [x] Optionally (also CLI), validate the model: make a the same query in `N` minutes back (presumably long time ago) and check that the selected JA5 hashes or IP addresses in the main query produced not more than `1/M` of the values from the main query. In other words, we verify that the the IPs and JA5 hashes became aggressive recently. We should not block known top talkers (typically business valuable clients).
- [x] Fetch the clients producing the most error codes. The error codes must also be configurable.
- [x] Fetch the clients producing the most RPS
- [x] Fetch the clients by unusual Geo regions (see Escudo sources and ask @ttaym )
- [x] Fetch the clients having the largest accumulated response time
- [x] `EXCLUDED_USER_AGENTS` should be in a config file
- [x] There should be a manager daemon, which monitors the number of error responses and average response time from Tempesta FW perfstat. It must run it training mode for a specified number of seconds and move to defense mode after that. Training mode determines average per second values and standard deviation. The number of standard deviations must be configured as a defence threshold. On the threshold the daemon must run the queries and validate the models. If a model validated, then issue the filtration rule. Repeat the action until response time and number of error message back to average + specified number of standard deviations (can be less than 1). The training mode should not be mandatory - it should be possible to call the daemon in defense mode right from a monitoring.

The daemon and all the scripts must log all the actions and what they're blocking.

This must be tested against https://github.com/tempesta-tech/tempesta-test/issues/38 - let's discuss how efficient it is and whether we need to fix anything in the design.

PR #2460, opened 2025-07-10
ISSUE 2450: DDOS mitigation script

symstu-tempesta:


scripts/ddos_mitigation/app.py
```diff
@@ -0,0 +1,68 @@
...
+    args = CommandLineArgs.parse_args()
+    app_config = AppConfig(_env_file=args.config)
+    logger.setLevel(getattr(logging, args.log_level or app_config.log_level, "INFO"))
+
+    clickhouse_client = ClickhouseAccessLog(
+        host=app_config.clickhouse_host,
+        port=app_config.clickhouse_port,
+        user=app_config.clickhouse_user,
+        password=app_config.clickhouse_password,
+        database=app_config.clickhouse_database,
```
RomanBelozerov, line 44: We can change a table for [TFWLogger](https://github.com/tempesta-tech/tempesta/wiki/TFW-Logger#configuration-options). It always use `default` database.

scripts/ddos_mitigation/logger.py
```diff
@@ -0,0 +1,16 @@
+import logging
+
+__author__ = "Tempesta Technologies, Inc."
+__copyright__ = "Copyright (C) 2023-2025 Tempesta Technologies, Inc."
+__license__ = "GPL2"
+
+
+logger = logging.getLogger()
```
RomanBelozerov: You shouldn't use root logger. It's bad idea because you will receive unexpected logs and other problems.
For example:
```python3
import logging
import requests

logger = logging.getLogger()
logger.propagate = True
logger.setLevel(logging.DEBUG)
handler = logging.StreamHandler()
handler.setLevel(logging.DEBUG)
handler.setFormatter(
    logging.Formatter("[%(asctime)s][%(name)s][%(levelname)s]: %(message)s")
)
logger.addHandler(handler)

response = requests.get('https://google.com')
```

I received logs:
```text
[2025-07-21 14:21:12,286][urllib3.connectionpool][DEBUG]: Starting new HTTPS connection (1): google.com:443
[2025-07-21 14:21:12,584][urllib3.connectionpool][DEBUG]: https://google.com:443 "GET / HTTP/11" 301 220
[2025-07-21 14:21:12,585][urllib3.connectionpool][DEBUG]: Starting new HTTPS connection (1): www.google.com:443
[2025-07-21 14:21:12,759][urllib3.connectionpool][DEBUG]: https://www.google.com:443 "GET / HTTP/11" 200 None
```

scripts/ddos_mitigation/logger.py
```diff
@@ -0,0 +1,16 @@
...
+logger.propagate = True
+logger.setLevel(logging.DEBUG)
+handler = logging.StreamHandler()
+handler.setLevel(logging.DEBUG)
```
RomanBelozerov: You should remove `handler.setLevel(logging.DEBUG)` because you don't manage the logging levels for handlers and loggers separately. It just doesn't make sense in this case.

Also I recommend use logging.CRITICAL like the default level. I received a loss of performance in our tests.

scripts/ddos_mitigation/config.py
```diff
@@ -0,0 +1,66 @@
...
+    training_mode_duration_min: int = 10
+
+    path_to_ja5t_config: str = "/etc/tempesta/ja5t/blocked.conf"
+    path_to_ja5h_config: str = "/etc/tempesta/ja5h/blocked.conf"
+
+    clickhouse_host: str = "192.168.0.104"
+    clickhouse_port: int = 8123
+    clickhouse_user: str = "default"
+    clickhouse_password: str = ""
+    clickhouse_database: str = "default"
```
RomanBelozerov, line 20: You should add clickhouse table here.

scripts/ddos_mitigation/config.py
```diff
@@ -0,0 +1,66 @@
...
+    default_errors_threshold: Decimal = 5
+
+    stats_window_offset_min: int = 60
+    stats_window_duration_min: int = 60
+    stats_rps_multiplier: int = 100
+    stats_time_multiplier: int = 100
+    stats_errors_multiplier: int = 5
+
+    detectors: set[Literal["threshold", "geoip"]] = {"threshold"}
+    blocking_type: set[Literal["ja5t", "ja5h", "ipset", "nftables", "geoip"]] = {"ja5t"}
```
RomanBelozerov:
```suggestion
    blocking_types: set[Literal["ja5t", "ja5h", "ipset", "nftables", "geoip"]] = {blockers.Ja5tBlocker.name()}
```

scripts/ddos_mitigation/blockers/base.py
```diff
@@ -0,0 +1,63 @@
...
+class BaseBlocker(metaclass=abc.ABCMeta):
+
+    @staticmethod
+    @abc.abstractmethod
+    def name() -> str:
+        """
+        Name of the blocker. Should be using in config blocking_type variable
+        """
+
+    def prepare(self):
```
RomanBelozerov, line 25: Why don't you use `abc.abstractmethod` decorator for all methods? You should remove them from base class. It's strange.

scripts/ddos_mitigation/blockers/nft.py
```diff
@@ -0,0 +1,136 @@
...
+class NFTBlocker(BaseBlocker):
+    def __init__(self, blocking_table_name: str):
+        self.blocking_table_name = blocking_table_name
+
+    @staticmethod
+    def name() -> str:
+        return "nftables"
+
+    def prepare(self):
+        result = run_in_shell("which nft")
```
RomanBelozerov: This method is difficult. Probably, it should be split in several methods?

scripts/ddos_mitigation/blockers/nft.py
```diff
@@ -0,0 +1,136 @@
...
+
+        result = run_in_shell(f"nft list table inet {self.blocking_table_name}_table")
+
+        if result.returncode != 0 and "No such file or directory" in result.stderr:
+            result = run_in_shell(
+                f"nft add table inet {self.blocking_table_name}_table"
+            )
+
+            if result.returncode != 0:
+                raise ValueError(f"Cannot add new table to nft: {result.stderr}")
```
RomanBelozerov: You use this block 26 times:
```python3
result = run_in_shell(cmd)
if result.returncode != 0:
    raise ValueError(msg)
```
I think we can add `check` argument to `run_in_shell` method or create a new method for these blocks. This makes the code more clean.

scripts/ddos_mitigation/access_log.py
```diff
@@ -0,0 +1,216 @@
...
+            f"""
+            SELECT 
+                min(ja5t), 
+                min(ja5h),
+                address,
+                min(user_agent) user_agent,
+                count(1) as total_requests,
+                sum(response_time) as total_time,
+                countIf(status not in (200, 201)) as total_errors
+            FROM access_log
```
RomanBelozerov: The table name is a changeable value

scripts/ddos_mitigation/detectors/base.py
```diff
@@ -0,0 +1,25 @@
...
+        """
+        Name of the detector. Should be used in config
+        """
+
+    async def prepare(self):
+        """
+        Made some prepatation, training, etc
+        """
+
+    async def find_users(self, current_time: int = None) -> list[User]:
```
RomanBelozerov: And you can add `prepare`  like abstractmethod
```suggestion
    @abc.abstractmethod
    async def prepare(self):
        """
        Made some prepatation, training, etc
        """

    @abc.abstractmethod
    async def find_users(self, current_time: int = None) -> list[User]:
```

RomanBelozerov (changes requested):
I think you should clean up the commit history, or you'll make a squish merge?

scripts/ddos_mitigation/config.py
symstu-tempesta: I prefer to avoid heavy imports inside config files, as they can recursively import others and, as a result, loading a simple data type could end up loading the entire application

scripts/ddos_mitigation/blockers/base.py
symstu-tempesta: Not all blockers implement this method, but it's included as part of the lifecycle

scripts/ddos_mitigation/detectors/base.py
symstu-tempesta: The `find_users` method might be abstract; the `prepare` method is also optional, similar to how it's handled in blockers

scripts/ddos_mitigation/config.py
RomanBelozerov: OK, but you should use `blocking_types` instead of `blocking_type`

RomanBelozerov (approved):

scripts/ddos_mitigation/README.md
```diff
@@ -0,0 +1,51 @@
+# DDoS Defender
+
+Block users by JA5T, JA5H, or IP based on Tempesta FW access 
+logs stored in the ClickHouse database.
+
+# How to run
+
+### Requirements:
+- Python 3.10 <=
+- Tempesta FW 0.0.8 <=
```
krizhanovsky:
```suggestion
- Tempesta FW 0.8.0 <=
```

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
+# DDoS Mitigation Script
+
+## Brief
+### How it works
+Tempesta provides extended information about user requests. In this case, we're interested in the user's IP address,
```
krizhanovsky:
```suggestion
Tempesta provides [extended information about user requests](/Handling-clients/#access-log). In this case, we're interested in the user's IP address, 
```

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
+# DDoS Mitigation Script
+
+## Brief
+### How it works
+Tempesta provides extended information about user requests. In this case, we're interested in the user's IP address, 
+as well as their JA5T and JA5H hashes. These hashes allow us to distinguish users based on similar characteristics, 
+such as TLS connection or HTTP request fingerprints.
+
+Additionally, access logs can be stored in ClickHouse, which offers extremely powerful capabilities for analyzing traffic.
```
krizhanovsky:
```suggestion
Additionally, access logs can be stored in [ClickHouse](Access-Log-Analytics/), which offers extremely powerful capabilities for analyzing traffic.
```

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+
+The mitigation script connects to the ClickHouse database and, at regular intervals, analyzes user traffic. 
+It compares aggregated values (such as the total number of requests, accumulated response time, and total number of
+error responses) against predefined thresholds. All of these thresholds can be customized in the application configuration.
+
+To block a user, the mitigation script adds the user's JA5T hash to the Tempesta FW configuration and reloads the server.
```
krizhanovsky:
```suggestion
To block a user, the mitigation script adds the user's JA5 hashes to the Tempesta FW configuration and reloads the server.
```

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+
+### Historical Mode
+The Mitigation Script can be configured to start in historical mode.  To enable it, set the following in your
```
krizhanovsky:
```suggestion
The Mitigation Script can be configured to start in historical mode.  In this mode the script learns form the historical data stored and retrieved from ClickHouse. To enable it, set the following in your 
```

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+It compares aggregated values (such as the total number of requests, accumulated response time, and total number of
+error responses) against predefined thresholds. All of these thresholds can be customized in the application configuration.
+
+To block a user, the mitigation script adds the user's JA5T hash to the Tempesta FW configuration and reloads the server.
+
+### Historical Mode
+The Mitigation Script can be configured to start in historical mode.  To enable it, set the following in your 
+app configuration:
+```.env
+training_mode="historical"
```
krizhanovsky: It seems both cases are allowed, but it's a bit confusing to see CAPSLOCK in example.env and lower case in the Wiki

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+The mitigation script connects to the ClickHouse database and, at regular intervals, analyzes user traffic. 
+It compares aggregated values (such as the total number of requests, accumulated response time, and total number of
+error responses) against predefined thresholds. All of these thresholds can be customized in the application configuration.
+
+To block a user, the mitigation script adds the user's JA5T hash to the Tempesta FW configuration and reloads the server.
+
+### Historical Mode
+The Mitigation Script can be configured to start in historical mode.  To enable it, set the following in your 
+app configuration:
+```.env
```
krizhanovsky: I'm not sure if our wiki translator on the website is able to process `.env` syntax, usually we use `bash` or just empty syntax for config snippets.

scripts/ddos_mitigation/WIKI.md
krizhanovsky: It seems the main script is `defender.py`, so maybe replace "The Mitigation Script" with "Defender" or "L7 DDoS Defender"?

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+persistent_users_window_offset_min=60
+persistent_users_window_duration_min=60
+persistent_users_total_requests: Decimal=1
+persistent_users_total_time=1
+```
+
+A user is marked as persistent if they exceed either of the following thresholds during the specified time window: 
+
+- persistent_users_total_requests: minimum number of requests (RPS)
+- persistent_users_total_time: minimum total response time
```
krizhanovsky:
```suggestion
- `persistent_users_total_requests`: minimum number of requests (RPS)
- `persistent_users_total_time`: minimum total response time
```

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+To train the script using the last 10 minutes of live traffic, you can use:
+
+```.env
+training_mode_duration_min=10
+```
+During this period, the script will gather user activity, calculate average metrics,
+apply multipliers (as in historical mode), and set the thresholds accordingly.
+
+### Persistent Users
+This feature is available only in `historical` and `real` modes, as it requires existing traffic data for analysis.
```
krizhanovsky, line 49: I'd add "Defender reacts on system metrics getting worse and kills the most aggressive clients impacting to the system overload. However, it might false positively kill benign clients who have been working with the system before the degradation event. The set of such _persistent_ clients can also be learnt by Defender." to provide some context for the feature

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+Each User-Agent should be on a separate line.
+
+If you want to use a custom location for this file, you can set the following variable in your config:
+
+```.env
+allowed_user_agents_file_path=/your/custom/path.txt
+```
+
+Requests with matching User-Agents will be ignored by the blocking logic and treated as safe.
+
```
krizhanovsky, line 98: I'd add

> This is quite unreliable way to white list client since many, even relatively simple, DDoS attacks use a pool of real life user agents. In practice this is the most useful for Web API if your clients use some specific User-Agent values. This still won't help if an attacker prepares an attack specifically for your service, but it'd safe to say that 90% or more DDoS attacks aren't prepared for a specific target.

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+By default, the ja5t blocking method is used. However, multiple methods can be specified, including combinations:
+
+```.env
+blocking_types=["ja5t", "ipset"]
+```
+
+### Unblocking Users
+After a DDoS attack, blocked users can be automatically unblocked.
+The default blocking duration is controlled by the blocking_time_min variable (in minutes).
```
krizhanovsky:
```suggestion
The default blocking duration is controlled by the `blocking_time_min` variable (in minutes).
```

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+```.env
+blocking_release_time_min=5
+```
+
+This ensures that users are not blocked longer than necessary, while still maintaining protection during an active attack.
+
+## Prepare Tempesta FW
+The script requires a specific Tempesta FW configuration.
+Let's create two directories inside the Tempesta configuration directory /etc/tempesta/:
+ja5t and ja5h.
```
krizhanovsky:
```suggestion
Let's create two directories inside the Tempesta configuration directory `/etc/tempesta/`:
`ja5t` and `ja5h`.
```

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+```
+
+## How to Defend Your App
+The mitigation script currently provides basic protection suitable for small to medium-sized applications,
+where traffic spikes are not extremely frequent or unpredictable.
+
+### Blog or Online Shop
+These types of applications typically don’t have a large number of concurrent users and often operate within a traffic range of 0 to 50 active users.
+However, it's important to account for the fact that static files (like CSS, JS, and images) are also requested.
+On initial page load, a single user might generate up to 20 or more HTTP requests.
```
krizhanovsky: For me the numbers are too small. E.g. https://tempesta-tech.com/ index produces 86 requests. https://www.kiwoko.com - 172 requests. I'd move to "up to 200" instead of 20

Our website barely reaches 2-3k views monthly, but with if a new blog post hits a hacker news discussion we can easily go significantly above 50 active users for a couple of hours. I think SMB  ecommerce also can get 50-500 online on promos, but 0-50 is a good average.

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+
+Let's estimate:
+- If 10 users are browsing your site concurrently, total requests might reach 200.
```
krizhanovsky: For average it probably make sense to talk abut 10 users with 50 requests, and 500 rps

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+- If there are 50 concurrent users, it might go up to 1000 requests.
```
krizhanovsky: Probably for the second calculation it makes sense to estimate a peak, like 500 and 10,000RPS. However, it still unrealistic since 500 online isn't equal to 500 concurrent visits. Maybe divide this by 60 seconds?..

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+
+Requests alone are not the only important metric.
+
+#### Total Accumulated Response Time
+Static file requests are usually handled directly by Tempesta FW without reaching the backend.
+However, dynamic page generation or API calls (e.g., fetch() requests) hit the backend and consume time.
```
krizhanovsky:
```suggestion
However, dynamic page generation or API calls (e.g., `fetch()` requests) hit the backend and consume time.
```

Dynamic pages are also cached if a CMS allows this.

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+#### Total Errors
+A spike in errors (like 5xx responses) is a strong signal of a problem.
+If you're seeing dozens of such responses, it likely means something is going wrong and needs attention.
+
+#### Example Mitigation Script Configuration
+Based on a typical blog or online shop scenario, the following configuration is a reasonable starting point:
+
+```.env
+default_requests_threshold=300
+default_time_threshold=40
```
krizhanovsky: Need to describe in the wiki how this is computed: accumulated sum of all response times for a client  per a sliding window (how the window is configured?)

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+```
+
+This setup allows the script to observe traffic for 30 minutes, calculate real averages,
+and apply scaled thresholds based on live behavior — which is ideal for dynamic, traffic-intensive apps.
+
+### Testing your App
+There are many tools available to simulate DDoS attacks on your application.
+Some of them — like Apache JMeter — even allow you to write request scenarios and define different RPS (requests per second) loads over time slices.
+
+However, for a more focused and powerful DDoS simulation, we recommend using MHDDoS.
```
krizhanovsky: Please insert links to HDHHoS and JMeter

scripts/ddos_mitigation/access_log.py
```diff
@@ -0,0 +1,234 @@
...
+                    WHEN total_errors >= {errors_threshold} THEN total_errors
+                END as value,
+                CASE 
+                    WHEN total_requests >= {rps_threshold} THEN 0
+                    WHEN total_time >= {time_threshold} THEN 1
+                    WHEN total_errors >= {errors_threshold} THEN 2
+                END as type
+            FROM scored_clients
+            WHERE persistent_user_agent = ''
+            LIMIT {ja5_hashes_limit}
```
krizhanovsky: It's good to have persistent clients, but in my understanding this isn't equal to the model validation from #2450.

Suppose there is a spike of error codes. We select the top clients producing error codes. All of them might be persistent clients - the most producing clients, while the spike is caused by thousands of tiny clients producing only a small fraction of errors. The clients could be caught with computing the most frequent JA5 now (i.e. the new clients have unique JA5).

The point of this example is that:
1. we have an assumption how to classify bad clients - this is our classification model (detector)
2. next we verify whether the model doesn't detect anything in the past, before the triggering event happend
3. if it's OK, we apply the model and block the clients. If not, we need another model - just dominating ja5t from the example.

In other words, #2450 requires a dynamically adjusting system, which tries one detector after another and validates them against the history. If none of the detector is validate, it just gives up.

The problem with the implemented approach is that we have to configure a detector and blocker upfront and we have no idea which attach we'll face and which contermeasures will be effective.

Is it possible to run the logic from a CLI? e.g. for a particular configuration get a list or peristent clients and the list of clients to be blocked by RPS or error codes? This way we can use Defender as a tool for manual incedent response.

scripts/ddos_mitigation/example.env
```diff
@@ -0,0 +1,175 @@
...
+# The precision of average time values statistic. In cases when values are to low, it should
+# made rounding, like 0.5 should be increased to 1, or 0.4 to zero
+STATS_TIME_PRECISION=1
+
+# The precision of average errors values statistic. In cases when values are to low, it should
+# made rounding, like 0.5 should be increased to 1, or 0.4 to zero
+STATS_ERRORS_PRECISION=0.1
+
+# The list of backend response status which are legal
+RESPONSE_STATUSES_WHITE_LIST=[100,101,200,201,204,400,401,403]
```
krizhanovsky: there also should be 3xx redirects

scripts/ddos_mitigation/detectors/threshold.py
```diff
@@ -0,0 +1,276 @@
...
+
+            average_stats = await self.average_stats_load(
+                start_at=int(time.time()) - self.app_config.stats_window_offset_min,
+                period_in_minutes=self.app_config.stats_window_duration_min,
+            )
+            self.set_thresholds(
+                requests_threshold=average_stats.requests,
+                time_threshold=average_stats.time,
+                errors_threshold=average_stats.errors,
+            )
```
krizhanovsky: It seems we define thresholds with average `M` values and will have plenty of false postives. We should compute standard deviations `D` and define the thresholds as the value for [z-score](https://en.wikipedia.org/wiki/Standard_score), e.g. 10 by default

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,347 @@
...
+### Blocking Methods
+The mitigation script supports several methods for blocking users:
+
+| NAME | Description                                |
+|-|--------------------------------------------|
+| ja5t | Client TLS connection fingerprint          |
+| ja5h | Client HTTP fingerprint                    |
+| ipset | Block by  ip using ipset + iptables |
+| nftables | Block by ip using nftables |
+
```
krizhanovsky, line 108: The table is very useful. Could you please also make a chapter for detectors and describe each of them with their configuration options? Also please describe in the geoip whether it's required something extra, like the geoip database, to install

scripts/ddos_mitigation/example.env
```diff
@@ -0,0 +1,175 @@
...
+
+# Time period (in seconds) that defines how often the user blocking task
+# should run to find and block new users.
+BLOCKING_WINDOW_DURATION_SEC=10
+
+# Default number of JA5 hashes to block during a single checking period
+BLOCKING_JA5_LIMIT=10
+
+# Default limit of IP addresses to block during a single period.
+BLOCKING_IP_LIMITS=10
```
krizhanovsky: I was wrong with 10 (it was just for my games with clickhouse) a reasonable default should not be less than 100, maybe 1,000 is more adequate

krizhanovsky (changes requested):
Quite a big work!

My review wasn't super deep, but it seems the most problem is with the computing thresholds as rounded averages. The next, not so serious, issue is models verification (please check unselected check boxes in #2450).

Wiki is great. Just several items to add/improve.

scripts/ddos_mitigation/WIKI.md
symstu-tempesta: The index could be 50 * 200 = 10k, but subsequent pages will be partially cached by the browser. A new page might require ~10 XHRs and up to 20 static files (previews of goods or posts). In this case, the peak RPS over the last 10 seconds should be 2350. We can assume a user needs at least 3 seconds to evaluate a page and another 1–3 seconds to act, so roughly 2350 / 5 = 470 RPS

scripts/ddos_mitigation/access_log.py
symstu-tempesta: **Answer 1 – Persistent users**
---
Persistent users are those who generate the most errors. Filtering persistent users is performed as the final step for efficiency reasons, since comparing long strings for millions of requests is computationally expensive. We still calculate their impact for the global score, but skip them in the final return.
This way, we avoid comparing long strings for all requests and only perform the comparison for items that exceed the limits. In other words, persistent user statistics are not used directly in the output.
The only situation where persistent users could affect results is if their ja5t hashes match those of attackers. In that case, blocking the attacker might also block the persistent user. Ideally, the Defender should identify real persistent users’ ja5t hashes and skip them explicitly. I believe ja5t hashes should always be unique, at least for small- and mid-size applications.

**Answer 2 – Dynamically adjusting system**
---
In my view, the pretraining mode is essentially a basic form of dynamic adjustment. We start the Defender in historical or real-time mode, collect data for the last X minutes or hours, and then set the threshold based on this collected data. Users exceeding the limits are blocked. Alternatively, you can set predefined limits if you have a good understanding of your app’s/system’s capabilities. This approach is usually sufficient for small- and mid-size applications.

However, if we update thresholds periodically during application operation, we may encounter non-trivial situations — for example, day/night traffic fluctuations, promotional campaigns, holidays, Black Friday, etc. These scenarios are common in large applications with hundreds or thousands of online users.
The challenge is determining whether a morning traffic spike is “normal” when comparing it only to the previous hour’s data. In my opinion, such cases should be handled with more complex daily/weekly/monthly patterns. The system could avoid automatic blocking in these situations and instead generate notifications about abnormal traffic, allowing the administrator to choose an action manually.

**Answer 3 – CLI**
---
The application can be started manually but run as a process. It is also possible to add a CLI mode in which the application performs only a single check (even in pretraining mode), blocks the relevant users, and then exits.

** Appending №1 **
---
I think I confused myself earlier.
The “persistent users” in the query you marked are simply whitelisted users, defined by the User-Agent strings listed in a text file.

The “persistent users” in your sense are calculated in training mode. We can configure their retrieval using the following parameters:

```bash
    persistent_users_max_amount: int = 100
    persistent_users_window_offset_min: int = 60
    persistent_users_window_duration_min: int = 60
    persistent_users_total_requests: Decimal = 1
    persistent_users_total_time: Decimal = 1
```  

and this users can have a problem if they really generates more traffic rather than other. To fix it we need to have a special table in database with all this users and filter them out as i do with white listed user agents

** Appendix №2 **
---
After today’s call, you suggested adding a CLI that both displays useful information and verifies the model. That’s easy to add. Let’s align on what it should show and do:

```bash
# start defender monitoring
./defender.py start  --config=defender.env

# show current traffic for period
./defender.py info limits
rps=30, time=5, errors=5

# show current traffic stats
./defender.py info traffic
RPS: 100, TIME: 0.1, ERRORS: 0.1

# show already blocked clients
./defender.py info blocked
Total: 3

ja5t=2309ywfe02, value=34, type=rps
ja5t=1kjdhf82323, value=10, type=time
ja5t=1kjdhf82323, value=7, type=errors

# show clients that potentially could be blocked
./defender.py risky show
Total: 3

ja5t=2309ywfe02, value=34, type=rps
ja5t=1kjdhf82323, value=10, type=time
ja5t=1kjdhf82323, value=7, type=errors

# block clients and show them
./defender.py risky block
Total: 3

ja5t=2309ywfe02, value=34, type=rps
ja5t=1kjdhf82323, value=10, type=time
ja5t=1kjdhf82323, value=7, type=errors
```

scripts/ddos_mitigation/WIKI.md
```diff
@@ -0,0 +1,399 @@
...
+```
+
+## How to Defend Your App
+The Defender currently provides basic protection suitable for small to medium-sized applications,
+where traffic spikes are not extremely frequent or unpredictable.
+
+### Blog or Online Shop
+These types of applications typically don’t have a large number of concurrent users and often operate within a traffic range of 0 to 50 active users.
+However, it's important to account for the fact that static files (like CSS, JS, and images) are also requested.
+On initial page load, a single user might generate up to 200 or more HTTP requests. All farther requests should receive some
```
krizhanovsky, line 275:
```suggestion
On initial page load, a single user might generate up to 200 or more HTTP requests. All further requests should receive some
```

scripts/ddos_mitigation/detectors/threshold.py
krizhanovsky: Why don't we use standard deviation? It could be hard to configure the multipliers, basically, one need to do some data analysis to reveal what the multipliers are.

krizhanovsky (commented):
I still have questions, let's discuss them on a call

scripts/ddos_mitigation/access_log.py
symstu-tempesta: Based on our today call:

1. Different models 
2. Z-score

Issue #2439, opened 2025-06-10
Rapid-reset attack mitigation


const-t:
# Motivation

At this moment to mitigate rapid-reset attack Temesta FW [recommends](https://github.com/tempesta-tech/tempesta/wiki/DDoS-mitigation#http2-rapid-reset) to use frang directives `request_rate/burst`, however there is more effective mitigation approaches.

# Scope
1. Tempesta FW struggles of non-cancelling variant of this attack. To mitigate it effectively we should drop connection with the client after a small number of attempts to exceed max concurrent streams limit(or even drop immediately if it not breaks legal clients). This approach eliminates the need for strict `request_rate/burst` limits in this scenario or at least allows them to be set significantly higher.
2. For regular variant of the attack to not spent time on processing garbage traffic, we should count amount of failed streams in the connction and when it exceeds some value e.g 50%(might be configurable) - disconnect or block the client.

# Test
In our repository there is script to reproduce this attack.

krizhanovsky:
Probably we can generalize the point (1) to drop connection on any detected attack, not only rate limit exceeding, i.e. from functions `tfw_http_req_parse_block()` and `tfw_http_req_block()`.

EvgeniiMekhanik:
I suggest to use control frame limits for both types of this attack. For cancelling variant we block connection if count of rst streams exceeded the limit of control frames. For non-cancelling variant we also use the same limit, because in fact we limit count of generated rst stream.

krizhanovsky:
Fixed in #2462 for 0.9 (`master`) and #2488 for 0.8

PR #2462, opened 2025-07-14
Implement frames flood rate limit

EvgeniiMekhanik:


const-t (commented):
We discussed few changes in the chat, once it will be implemented I will re-review it again

fw/http_limits.c
```diff
@@ -1673,6 +1673,58 @@ frang_http_hdr_limit(TfwHttpReq *req, unsigned int new_hdr_len)
...
 
+static int
+__frang_ctrl_frame_limit(FrangAcc *ra, TfwCliConn *conn,
+			 unsigned int ctrl_frame_rate)
+{
+	TfwH2Ctx *ctx = tfw_h2_context_unsafe((TfwConn *)conn);
+	unsigned long ts = jiffies * FRANG_FREQ / HZ;
+	int i = ts % FRANG_FREQ;
+
+
```
const-t: Empty line

fw/http_limits.c
```diff
@@ -1673,6 +1673,58 @@ frang_http_hdr_limit(TfwHttpReq *req, unsigned int new_hdr_len)
...
+	r = __frang_ctrl_frame_limit(ra, (TfwCliConn *)conn,
+				     dflt_vh->frang_gconf->ctrl_frame_rate);
+
+	if (unlikely(r == T_BLOCK_WITH_RST) && dflt_vh->frang_gconf->ip_block)
+		tfw_filter_block_ip(FRANG_ACC2CLI(ra));
+
+	tfw_vhost_put(dflt_vh);
+
+	return r;
+
```
const-t: Empty line

fw/http_frame.c
```diff
@@ -1030,6 +1037,18 @@ do {									\
 		      ctx->streams_num);				\
 		TFW_INC_STAT_BH(clnt.streams_num_exceeded);		\
 		SET_TO_READ_VERIFY(ctx, HTTP2_IGNORE_FRAME_DATA);	\
+		/*							\
+		 * There are two types of rapid reset attack:		\
+		 * - client opens a lot of streams and then closes them \
+		 *   using RST STREAM. This type of attack is fixed by	\
+		 *   control frame limit;
+		 * - client opens a lot of streams and exceeded max	\
+		 *   concurrent streams limit to make the server generates \
```
```suggestion
		 *   concurrent streams limit to make the server generate \
```

fw/http2.h
```diff
@@ -83,6 +95,7 @@ typedef struct tfw_conn_t TfwConn;
  *			  from _HTTP2_SETTINGS_MAX are used to save what
  *			  settings we sent to the client;
  * @conn		- pointer to h2 connection of this context;
+ * @ctrl_frame_stat	- control frame receiption history;
```
krizhanovsky:
```suggestion
 * @ctrl_frame_stat	- control frame reception history;
```

etc/tempesta_fw.conf
```diff
@@ -1053,6 +1053,7 @@
 #       http_body_len NUM;
 #       http_hdr_len NUM;
 #       http_header_cnt NUM;
+#	ctrl_frame_rate NUM;
```
krizhanovsky: Please define a reasonable default value in the config (note necessary in the C code) to provide secure by default configuration.

And here is an important practical question: what is a reasonable value for the rate limit, having that it covers all the streams and TCP connections for a client and it accounts frequent frames like SETTINGS, WINDOW_UPDATE and PRIORITY.

Shouldn't we actually use a ratio instead of the absolute number? Probably it'd be easier to estimate ration like not more than 20% percent of all frames.

krizhanovsky (approved):
LGTM. Just please fix the minor comments and add the description of the configuration option to 

https://tempesta-tech.com/knowledge-base/DDoS-mitigation/ (see the upcoming blog post and let's discuss on the call)

https://tempesta-tech.com/knowledge-base/On-the-fly-Reconfiguration/#on-the-fly-configuration-possibilities

https://tempesta-tech.com/knowledge-base/HTTP-security/

fw/http_frame.c
```diff
@@ -1030,6 +1037,18 @@ do {									\
 		      ctx->streams_num);				\
```
krizhanovsky: Only `tfw_h2_frame_type_process()` is affected by the rate limiter, so how the 2nd point from #2439 

> For regular variant of the attack to not spent time on processing garbage traffic, we should count amount of failed streams in the connction and when it exceeds some value e.g 50%(might be configurable) - disconnect or block the client.

is implemented? Even for the proposal

> For non-cancelling variant we also use the same limit, because in fact we limit count of generated rst stream.

I don't see rate limiting for the generated RST frames.

krizhanovsky (changes requested):
At the second glance it seems we can do better and there is also 2nd point from #2439 for TODO

fw/http_frame.c
EvgeniiMekhanik: We discuss with @const-t about this pint and decide that current implementation is no worse. look in this patch we ratelimit count of failed streams and disconnect and block the client. But we do it in the same way and the same place  as we do for all other limits. That's why I chose this option.

EvgeniiMekhanik: Look we call `frang_ctrl_frame_limit` in the `VERIFY_MAX_CONCURRENT_STREAMS` macros each time when count of max concurrent streams is exceeded. If ratelimit is exceeded also we disconnect and block client.

krizhanovsky: @EvgeniiMekhanik I still don't understand: if we have `SETTINGS_MAX_CONCURRENT_STREAMS=100` and client constantly establishes say not more than 50 streams, but ALL the streams are reset on our side due to invalid frames from the client - how this attack is blocked?

EvgeniiMekhanik: According https://cloud.google.com/blog/products/identity-security/how-it-works-the-novel-http2-rapid-reset-ddos-attack there are two types of rapid reset attack
1. The client may also assume that the cancellation will take effect immediately when the server receives the RST_STREAM frame, before any other data from that TCP connection is processed.
This attack is called Rapid Reset because it relies on the ability for an endpoint to send a RST_STREAM frame immediately after sending a request frame, which makes the other endpoint start working and then rapidly resets the request. The request is canceled, but leaves the HTTP/2 connection open. (This case is covered by
```
	if (tfw_h2_frame_is_ctrl(hdr_type) &&
	    unlikely(frang_ctrl_frame_limit((TfwConn *)ctx->conn)))
		return T_BLOCK_WITH_RST;
```
Same as all other control frames!
2. The second variant does away with canceling streams entirely, and instead optimistically tries to open more concurrent streams than the server advertised. This case is covered in `VERIFY_MAX_CONCURRENT_STREAMS`

EvgeniiMekhanik: The only one case is not covered in a good way
- The first variant does not immediately cancel the streams, but instead opens a batch of streams at once, waits for some time, and then cancels those streams and then immediately opens another large batch of new streams. This attack may bypass mitigations that are based on just the rate of inbound RST_STREAM frames (such as allow at most 100 RST_STREAMs per second on a connection before closing it). But according the article
""'
These attacks lose the main advantage of the canceling attacks by not maximizing connection utilization, but still have some implementation efficiencies over standard HTTP/2 DDoS attacks. But this variant does mean that any mitigation based on rate-limiting stream cancellations should set fairly strict limits to be effective.
"""
So we can set a strict limit for contol frames

etc/tempesta_fw.conf
krizhanovsky: Agreed on separate hardcoded ratelimits for each control frame type. WINDOW_UPDATE should be relative to DATA frames. Add a configurable multiplier. Add performance counters for connection resets per control frame rate limit - propose to increment multiplier in case of growing the stats. Probably it makes sense to implement this right in HTTP/2 handling to avoid frang callback overheads

fw/http_frame.c
```diff
@@ -999,6 +1003,133 @@ tfw_h2_frame_pad_process(TfwH2Ctx *ctx)
...
+{
+	unsigned long ts = jiffies * FRANG_FREQ / HZ;
+	int i = ts % FRANG_FREQ;
+
+	if (ctrl_frame_stat[i].ts != ts) {
+		ctrl_frame_stat[i].ts = ts;
+		ctrl_frame_stat[i].cnt = 0;
+	}
+	ctrl_frame_stat[i].cnt++;
+
```
krizhanovsky: extra empty line

fw/http_frame.c
```diff
@@ -999,6 +1003,133 @@ tfw_h2_frame_pad_process(TfwH2Ctx *ctx)
...
+	if (unlikely(++ctx->prio_frame_cnt < ctx->lsettings.max_streams))
+		return true;
+
+	if (unlikely(ctx->prio_frame_cnt > PRIO_FRAME_RATE_LIMIT))
+		return false;
+
+	return true;
+
+	#undef PRIO_FRAME_RATE_LIMIT
```
krizhanovsky: I'd not use a macro for only one use right a couple of lines below

fw/http_frame.c
```diff
@@ -999,6 +1003,133 @@ tfw_h2_frame_pad_process(TfwH2Ctx *ctx)
...
+
+static inline bool
+tfw_h2_wnd_update_rate_limit(TfwH2Ctx *ctx)
+{
+	/*
+	 * We allow to send not more then six WINDOW_UPDATE
+	 * frames for one DATA frame by default. We also allow
+	 * to send at least `max_streams` WINDOW_UPDATE frames
+	 * to cover the case when client initiate connection
+	 * with zero window.
```
krizhanovsky: Can we also add a limit to accept no more than 6 WINDOW_UPDATEs say for each kilobyte of received payload data to mitigate Data Dribble on this level as well? We're going to have memory accounting in #1346 to fight with the attack, but I think it still makes sense to have 1 more level of protection - the statistical model also isn't 100% accurate.

fw/http_frame.c
```diff
@@ -999,6 +1003,133 @@ tfw_h2_frame_pad_process(TfwH2Ctx *ctx)
...
+	 */
+	#define WND_UPDATE_FRAME_RATE_LIMIT		\
+		ctx->data_frames_sent * 6 * wnd_update_frame_rate_multiplier
+
+	if (unlikely(++ctx->wnd_update_cnt < ctx->lsettings.max_streams))
+		return true;
```
krizhanovsky: This counter and `prio_frame_cnt` are never reset/decremented, so a long living enough connection (e.g. for a large transfer) will hit the limit and we'll terminate large downloads.

It seems the single integer counter isn't accurate: we can increment it say every 1/10 of second and we don't know when to decrease it and how much. If we clear it, we loose the 9/10 second of history.

etc/tempesta_fw.conf
```diff
@@ -1201,6 +1201,30 @@
...
+# This directive controls the number of WINDOW UPDATE frames that the
+# Tempesta FW can accept per count of DATA frames. Maximum value 1024.
+#
+# Syntax:
+#   window_update_frame_rate_multiplier NUM;
+#
+# Example:
+#   window_update_frame_rate_multiplier 16;
+#
+
```
krizhanovsky: With #2478 we don't have the config comments any more. Please describe the option in https://tempesta-tech.com/knowledge-base/HTTP-security/ instead

fw/procfs.h
```diff
@@ -70,13 +70,23 @@ typedef struct {
...
 typedef struct {
 	TFW_STAT_COMMON;
 	u64	streams_num_exceeded;
+	u64	prio_frame_exceeded;
+	u64	rst_stream_frame_exceeded;
+	u64	settings_frame_exceeded;
+	u64	ping_frame_exceeded;
+	u64	wnd_update_frame_exceeded;
 	u64	msgs_fromcache;
 	u64	online;
```
krizhanovsky, line 91: please document this in https://tempesta-tech.com/knowledge-base/Performance-statistics/

I also saw that for example `streams_num_exceeded` isn't described as well, so please review and update the whole wiki page

fw/http.c
```diff
@@ -152,6 +152,8 @@ static const DEFINE_TFW_STR(STR_CLEN_ZERO, "Content-Length: 0");
  */
 unsigned int max_header_list_size = 0;
 bool allow_empty_body_content_type;
+unsigned int ctrl_frame_rate_multiplier = 0;
+unsigned int wnd_update_frame_rate_multiplier = 0;
```
krizhanovsky: Don't you like shorter name, e.g. `wnd_upd_frame_rate_mul`?..

fw/http_frame.c
```diff
@@ -999,6 +1003,133 @@ tfw_h2_frame_pad_process(TfwH2Ctx *ctx)
...
+
+	if (ctrl_frame_stat[i].ts != ts) {
+		ctrl_frame_stat[i].ts = ts;
+		ctrl_frame_stat[i].cnt = 0;
+	}
+	ctrl_frame_stat[i].cnt++;
+
+
+	if (unlikely(ctrl_frame_stat[i].cnt >
+		     ctrl_frame_rate_limit * ctrl_frame_rate_multiplier))
```
krizhanovsky: We should count all the entries with `ts` not older than 1 second in the array: the array is a history of 8 measurements with 1/8 second granularity. See `frang_conn_limit()`.

fw/http2.h
```diff
@@ -128,6 +148,12 @@ typedef struct tfw_h2_ctx_t {
 	unsigned int    new_settings[_HTTP2_SETTINGS_MAX - 1];
 	DECLARE_BITMAP  (settings_to_apply, 2 * _HTTP2_SETTINGS_MAX - 1);
 	TfwH2Conn	*conn;
+	CtrlFrameStat	rst_stat[FRANG_FREQ];
+	CtrlFrameStat	ping_stat[FRANG_FREQ];
+	CtrlFrameStat	settings_stat[FRANG_FREQ];
```
krizhanovsky: This is 192 bytes for the counters. It'd be better to organize them like `FrangRates`: 3 counters with 1 timestamp

fw/http_frame.c
```diff
@@ -1019,19 +1150,31 @@ tfw_h2_frame_type_process(TfwH2Ctx *ctx)
...
+		SET_TO_READ_VERIFY(ctx, HTTP2_IGNORE_FRAME_DATA);		\
+		/*								\
+		 * There are two types of rapid reset attack:			\
+		 * - client opens a lot of streams and then closes them using	\
+		 *   RST STREAM. This type of attack is fixed by control frame	\
+		 *   limit;							\
+		 * - client opens a lot of streams and exceeded max concurrent	\
+		 *   streams limit to make the server generate a lot of		\
+		 *   RST STREAM responses. We can also use control frame	\
+		 *   limit to fix it;						\
```
krizhanovsky: > We can also use control frame limit to fix it

It's still unclear why. Can you elaborate on this fix in the comment?

krizhanovsky (changes requested):
There are definitely issues with accounting and limiting.

Also it seems we still need to limit the ratio of error-terminates streams caused by https://github.com/tempesta-tech/tempesta/issues/2439#issuecomment-2994206194 , say as not more 10%, and terminate the TCP connection with a client

fw/http_frame.c
EvgeniiMekhanik: yes fixed

fw/http_frame.c
EvgeniiMekhanik: yes fixed

fw/http.c
EvgeniiMekhanik: yes fixed

fw/http2.h
EvgeniiMekhanik: yes fixed

fw/http_frame.c
EvgeniiMekhanik: yes fixed

fw/procfs.h
EvgeniiMekhanik: Will be done after merging PR

etc/tempesta_fw.conf
krizhanovsky: Just leave unresolved to remind about the wiki update

fw/http_frame.c
```diff
@@ -999,6 +1037,146 @@ tfw_h2_frame_pad_process(TfwH2Ctx *ctx)
...
+	}
+
+	return true;
+}
+
+static inline bool
+tfw_h2_wnd_update_rate_limit(TfwH2Ctx *ctx)
+{
+	/*
+	 * We allow to send not more then six WINDOW_UPDATE
```
```suggestion
	 * We allow to send not more than six WINDOW_UPDATE
```

fw/http_frame.c
```diff
@@ -999,6 +1037,146 @@ tfw_h2_frame_pad_process(TfwH2Ctx *ctx)
...
+	 * frames for one DATA frame by default. We also allow
+	 * to send not more then one WINDOW_UPDATE frame for each
```
```suggestion
	 * to send not more than one WINDOW_UPDATE frame for each
```

fw/http2.h
```diff
@@ -21,6 +21,7 @@
 #define __HTTP2__
```
krizhanovsky, line 21: please update the copyright year for the file header

fw/http_frame.c
```diff
@@ -999,6 +1037,146 @@ tfw_h2_frame_pad_process(TfwH2Ctx *ctx)
...
+	 * 256 bytes of sent DATA frames to prevent data dribble
```
krizhanovsky: Could you please describe how did you choose 256?

We discussed on the call that we can and should  keep 256 for 0.8 and use a larger value for 0.9

fw/http_limits.h
```diff
@@ -192,4 +196,10 @@ int frang_sticky_cookie_handler(TfwHttpReq *req);
 bool frang_req_is_whitelisted(TfwHttpReq *req);
 int frang_http_hdr_limit(TfwHttpReq *req, unsigned int new_hdr_len);
 
+static inline int
+frang_time_in_frame(const unsigned long tcur, const unsigned long tprev)
+{
+	return tprev + FRANG_FREQ > tcur;
+}
```
krizhanovsky, line 203: Can you also use the function in `frang_conn_limit()` and maybe other code in http_limit.c

fw/http2.h
EvgeniiMekhanik: yes fixed

fw/http_limits.h
EvgeniiMekhanik: Yes fixed

fw/http_frame.c
```diff
@@ -46,6 +46,41 @@
 #define STREAM_ID_SIZE			4
 #define ERR_CODE_SIZE			4
 
+#define PING_FRAME_RATE_LIMIT		100
+#define SETTINGS_FRAME_RATE_LIMIT	5
+#define RST_FRAME_RATE_LIMIT		5
+#define PRIORITY_FRAME_RATE_LIMIT	30
+
+#define TFw_H2_CTRL_FRAME_RATE_LIMIT(name, limit)			\
```
const-t:
```suggestion
#define TFW_H2_CTRL_FRAME_RATE_LIMIT(name, limit)			\
```

fw/http_frame.c
```diff
@@ -999,6 +1040,75 @@ tfw_h2_frame_pad_process(TfwH2Ctx *ctx)
...
+		return false;
+	}
+
+	if (unlikely(ctx->wnd_update_cnt >
+		     (ctx->data_bytes_sent / 128) * wnd_update_frame_rate_mul)) {
+		TFW_INC_STAT_BH(clnt.wnd_update_frame_exceeded);
+		return false;
+	}
+
+
```
krizhanovsky: extra empty line

fw/http_frame.c
```diff
@@ -999,6 +1040,75 @@ tfw_h2_frame_pad_process(TfwH2Ctx *ctx)
...
+		return true;
+
+	if (unlikely(ctx->wnd_update_cnt >
+		     ctx->data_frames_sent * 6 * wnd_update_frame_rate_mul)) {
+		TFW_INC_STAT_BH(clnt.wnd_update_frame_exceeded);
+		return false;
+	}
+
+	if (unlikely(ctx->wnd_update_cnt >
+		     (ctx->data_bytes_sent / 128) * wnd_update_frame_rate_mul)) {
```
krizhanovsky, line 1065: Probably 128 should go to 0.8 and in master we can have 1024 or something like this

krizhanovsky: agreed to do this in #2445

fw/http.c
```diff
@@ -8349,6 +8352,13 @@ tfw_cfgop_cleanup_allow_empty_body_content_type(TfwCfgSpec *cs)
 	allow_empty_body_content_type = false;
 }
 
+static void
+tfw_cfgop_cleanup_frame_limit(TfwCfgSpec *cs)
+{
+	ctrl_frame_rate_mul = 0;
```
const-t: It seems we set it to zero if live reconfiguration is failed, but Tempesta continues to work. Isn't?

const-t (commented):
LGTM, but have a question

fw/http.c
EvgeniiMekhanik: yes fixed

fw/http_frame.c
EvgeniiMekhanik: yes fixed

PR #2470, opened 2025-07-25
2466: Add ability to specify clickhouse database name for tfw_logger

DenisShkiria:
The database name can be specified in the JSON configuration using the "db_name" key, or as a command-line argument using "--database/-d <name>". The "port" parameter is now taken into account. It existed earlier, but was ignored.

### DIFF
--- a/etc/tfw_logger.json
+++ b/etc/tfw_logger.json
@@ -6,6 +6,7 @@
         "port": 9000,
         "user": "default",
         "password": "",
+        "db_name": "default",
         "table_name": "access_log",
         "max_events": 1000,
         "max_wait_ms": 100
--- a/logger/clickhouse_config.hh
+++ b/logger/clickhouse_config.hh
@@ -24,6 +24,8 @@
 #include <optional>
 #include <string>
 
+#include <boost/property_tree/ptree_fwd.hpp>
+
 #include <fmt/format.h>
 
 struct ClickHouseConfig {
@@ -37,6 +39,12 @@ struct ClickHouseConfig {
 	size_t max_events{1000};
 	// Max time before commit
 	std::chrono::milliseconds max_wait{100};
+
+	void
+	parse_from_ptree(const boost::property_tree::ptree &tree);
+
+	void
+	validate() const;
 };
 
 template <> struct fmt::formatter<ClickHouseConfig> {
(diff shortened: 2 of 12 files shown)


logger/clickhouse_config.hh
```diff
@@ -0,0 +1,72 @@
...
+	parse(fmt::format_parse_context &ctx)
+	{
+		return ctx.begin();
+	}
+
+	template <typename FormatContext>
+	constexpr decltype(auto)
+	format(const ClickHouseConfig &config, FormatContext &ctx)
+	{
+		static constexpr auto msg_template =
```
EvgeniiMekhanik: I catch error during build `declared static in constexpr function`.

logger/tfw_logger_config.hh
```diff
@@ -122,6 +107,12 @@ public:
 		clickhouse_.port = port;
 	}
 
+	void
+	override_clickhouse_db_name(const std::string &db_name)
+	{
+		clickhouse_.db_name = db_name;
```
EvgeniiMekhanik: It seems that we should add checks here and for all other options. Same checks as we do it during parsing configuration file.

EvgeniiMekhanik: I mean if (clickhouse_.db_name.empty())

logger/clickhouse_config.hh
DenisShkiria: I reproduced and fixed this compilation error for GCC 11. However, as far as I know, the default compiler for Ubuntu 24.04 (which as far as I know is the Tempesta target system) is GCC 13.3. I checked the compilation with gcc13 while I was working on it.

logger/tfw_logger_config.hh
DenisShkiria: I moved all of these checks to the `validate()` methods and call them after reading the configuration value from the file and applying the command line arguments. Since there are no guarantees regarding the config values in the TfwLoggerConfig class after this change, I made it a struct and removed all `override_...` functions.

These changes were made in the `2466: Add validation of cmd arguments for tfw_logger` commit.

logger/tfw_logger_config.cc
```diff
@@ -21,99 +21,32 @@
...
 void
 TfwLoggerConfig::parse_from_ptree(const pt::ptree &tree)
 {
-	// Parse paths and settings
-	if (auto log_path_opt = tree.get_optional<std::string>("log_path"))
-		log_path_ = *log_path_opt;
+	if (const auto val = tree.get_optional<std::string>("log_path"))
+		log_path = *val;
+
+	buffer_size = tree.get<size_t>("buffer_size", buffer_size);
```
const-t: Please remove `buffer_size` from config, this parameter can be configured only in Tempesta using `mmap_log_buffer_size`.

DenisShkiria: done

DenisShkiria:
> Looks good. Please don't specify issue number or link to issue in commit message, if project will be moved from github we'll have "outdated" info in history which have nothing useful. Better to have short description of the issue in the message.

I removed the issue id (2466) from the commit messages.

etc/tfw_logger.json
```diff
@@ -6,6 +6,7 @@
         "port": 9000,
         "user": "default",
         "password": "",
+        "db_name": "default",
```
krizhanovsky, line 9: Pleasse update https://github.com/tempesta-tech/tempesta/wiki/Access-Log-Analytics one the PR is merged

PR #2473, opened 2025-07-30
Improve building response from cache.

EvgeniiMekhanik:
Previously during building response from cache
we did two things that greatly reduced perfomance
for plain http and https:
- for plain http we allocate extra skb for body, but we should not do it at all.
- for https we should allocate extra skb not every time. We allocate extra skb, because we should copy skb, which contains headers during tls encryption, so if we add body to such skb we will copy body
too. But if the body size is less then PAGE_SIZE >> 2, it's more effective to do not allocate skb and copy body.
https:
release-0.8			optimization
small:				small:
Transfer/sec:      1.27GB	Transfer/sec:      1.70GB
Transfer/sec:      1.24GB	Transfer/sec:      1.70GB
Transfer/sec:      1.17GB	Transfer/sec:      1.70GB
1K:
Transfer/sec:      1.71GB	Transfer/sec:      2.68GB
Transfer/sec:      1.66GB	Transfer/sec:      2.65GB
http:
release-0.8			optimization
Transfer/sec:      1.19GB	Transfer/sec:      1.86GB
Transfer/sec:      1.22GB	Transfer/sec:      1.98GB

### DIFF
--- a/fw/cache.c
+++ b/fw/cache.c
@@ -2800,25 +2800,42 @@ tfw_cache_add_body_page(TfwMsgIter *it, char *p, int sz, bool h2,
  * to actually reserve any space for h2 frame header.
  */
 static int
-tfw_cache_build_resp_body(TDB *db, TdbVRec *trec, TfwMsgIter *it, char *p,
+tfw_cache_build_resp_body(TDB *db, TdbVRec *trec, TfwHttpResp *resp, char *p,
 			  unsigned long body_sz, bool h2)
 {
+	TfwMsgIter *it = &resp->mit.iter;
+	bool sh_frag = !h2 && TFW_CONN_TLS(resp->req->conn);
 	int r;
-	bool sh_frag = h2 ? false : true;
 
 	if (WARN_ON_ONCE(!it->skb_head))
 		return -EINVAL;
+
 	/*
 	 * If all skbs/frags are used up (see @tfw_http_msg_expand_data()),
 	 * create new skb with empty frags to reference the cached body;
 	 * otherwise, use next empty frag in current skb.
+	 * If sh_frag is true we should copy skb with headers during
+	 * encryption. In this case we calculate count of skbs to build
+	 * response if we force add/not add extra skb for body here.
+	 * - If count of skbs are equal add skb (we don't increase count of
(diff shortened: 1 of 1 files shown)


fw/cache.c
```diff
@@ -2800,25 +2800,42 @@ tfw_cache_add_body_page(TfwMsgIter *it, char *p, int sz, bool h2,
...
 	 * create new skb with empty frags to reference the cached body;
 	 * otherwise, use next empty frag in current skb.
+	 * If sh_frag is true we should copy skb with headers during
+	 * encryption. In this case we calculate count of skbs to build
+	 * response if we force add/not add extra skb for body here.
+	 * - If count of skbs are equal add skb (we don't increase count of
+	 * skbs in response, but avoid extra body copying).
+	 * - If count of skbs are not equal we calculte count of body bytes
+	 * should be copied during encryption and if this count is greater
+	 * then TLS_MAX_PAYLOAD_SIZE create extra skb.
```
const-t: > count is greater then TLS_MAX_PAYLOAD_SIZE create extra skb

But in the condition I can't see where we use TLS_MAX_PAYLOAD_SIZE, looks like outdated comment or it must be rewritten to be more informative.

fw/cache.c
```diff
@@ -2800,25 +2800,42 @@ tfw_cache_add_body_page(TfwMsgIter *it, char *p, int sz, bool h2,
...
 	 */
-	if (!it->skb || it->frag + 1 >= MAX_SKB_FRAGS || sh_frag) {
+	if (!it->skb || it->frag + 1 >= MAX_SKB_FRAGS
+	    || (sh_frag
+		&& ((DIV_ROUND_UP(body_sz + it->skb->len,
+				  SS_SKB_MAX_DATA_LEN) ==
+		     1 + DIV_ROUND_UP(body_sz,
+				      SS_SKB_MAX_DATA_LEN))
+		    || (min(SS_SKB_MAX_DATA_LEN - it->skb->len,
+			    body_sz) > (PAGE_SIZE >> 2))))) {
```
const-t: I propose to extract to static function the condition introduced in this patch and write informative comment, especially describe why `PAGE_SIZE >> 2` has been chosen.

Issue #2451, opened 2025-06-17
Security by default configuration


krizhanovsky:
# Motivation

The prime focus of Tempesta FW is security, so the default configuration must switch on maximum security features.

See report from https://ddos-test.com for our website:

![Image](https://github.com/user-attachments/assets/03ecffc9-a395-4cb1-b075-acb68e053899)

# Scope

Must also be ported to 0.8.x

- [ ] port the security setting from tempesta-tech.com, including `request_rate` (*)
- [ ] do not keep idle TCP connections longer than 30 seconds (probably, just `sysctl` in `tempesta.sh`)

(*) Please check the commits history and comment all of the default security configuration settings, e.g. it's unclear for me why do we have `http_method_override_allowed true` and `http_strict_host_checking false`.

# Testing

Deploy on our website and check that https://ddos-test.com do not report the TCP idle time issues.

PR #2478, opened 2025-08-04
remove old version `tempesta_fw.conf` and add recommended basic config

RomanBelozerov:


etc/tempesta_fw.conf
```diff
@@ -1,1528 +1,70 @@
...
+# Allow only following characters in URI: $%()+,/a-zA-Z0-9&?:-.[]_=
+# https://tempesta-tech.com/knowledge-base/HTTP-security/#custom-character-sets
+http_uri_brange 0x24-0x26 0x28-0x29 0x2b 0x2c 0x2f 0x41-0x5a 0x61-0x7a 0x30-0x39 0x3f 0x3a 0x2d 0x2e 0x5b 0x5d 0x5f 0x3d;
+
+# https://ddos-test.com: The system keeps idle TCP client connections open for more than 45.0 seconds.
+# https://tempesta-tech.com/knowledge-base/Handling-clients/#keep-alive-timeout
+keepalive_timeout 30;
+
+# https://tempesta-tech.com/knowledge-base/Backend-servers/
+srv_group srv_grp_1 { server <server_IP>:<server_port> conns_n=64; }
```
krizhanovsky: This line won't be loaded by Tempesta FW, should be commented out. Probably add some mark to make it visible as the most crucial setting to define

krizhanovsky (approved):
Looks good, but we should do something with `srv_group` it'd should be visible that a user MUST set something here to get working installation

etc/tempesta_fw.conf
```diff
@@ -1,1528 +1,70 @@
...
+block_action attack drop;
+
+# Security settings: https://tempesta-tech.com/knowledge-base/HTTP-security/
+frang_limits {
+  request_rate 200;
+  request_burst 25;
+  tcp_connection_rate 100;
+  tcp_connection_burst 10;
+  concurrent_tcp_connections 500;
+  http_methods head post put get delete;
```
const-t, line 41: In my opinion is doesn't make sense to have these "default" values in config. Default config should contain minimal working configuration, but not override default values by another default values.

The same related to `http_uri_brange` and `keepalive_timeout`. It looks weird, that we have `keepalive_timeout 75` default value in Tempesta and then override the value by another "good" value in the default config.

krizhanovsky: @const-t do you think we should adjust the default config values in the C code for #2451 ?

RomanBelozerov: I added these values from our website. We tested them using our CI, Nexus, Grafana and website. I agree that it looks strange, maybe I'm still misunderstanding  issue #2451.

`keepalive_timeout 75` - https://ddos-test.com/ does not recommend such value.
`http_uri_brange` - like an example. This directive allows almost all characters.
```
  request_rate 200;
  request_burst 25;
  tcp_connection_rate 100;
  tcp_connection_burst 10;
  concurrent_tcp_connections 500;
```
 I do not know if these values should be used. I think we can consider this as a minimal set of security features because they are disabled by default.

const-t: @const-t @RomanBelozerov For `keepalive_timeout` I think we should adjust default value. But for frang we should have either separated config which has more complicated security oriented(something like can be used in prod) configuration or simple example in the current config.

PR #2479, opened 2025-08-04
Mekhanik evgenii/fix collapce and body moving

EvgeniiMekhanik:


fw/http_msg.c
```diff
@@ -1199,6 +1200,16 @@ __tfw_http_msg_move_body(TfwHttpResp *resp, struct sk_buff *nskb, int *frag)
...
+	/*
+	 * After moving body, we should also update `resp->cut`
+	 * to correct removing body flag data later.
+	 */
+	TFW_STR_FOR_EACH_CHUNK(c, &resp->cut, end) {
+		if (c->skb == *body)
+			c->skb = nskb;
+		else
+			break;
+	}
```
krizhanovsky, line 1211: This loop can be large for large responses - can we combine this work with some other to avoid one more loop?

PR #2481, opened 2025-08-08
Denis shkiria 2314 tfw logger missing logs

DenisShkiria:
I made several commits. One is for refactoring, and the other addresses the issue. Please review the changes for each commit rather than the total diff.


logger/mmap_buffer.cc
```diff
@@ -112,12 +112,9 @@ TfwMmapBufferReader::read()
 
 	const char *data = buf_->data + (tail & buf_->mask);
 	const uint64_t size = head - tail;
-	if (size == 0) [[unlikely]]
```
const-t: I can't understand how if fixes the issue? How I see it: if `size == 0` in the callback we don't reach `while` loop and we only call `clickhouse.commit()`. Therefore it looks like we didn't call `commit` previously and fix the problem calling `commit()` few times.

const-t (changes requested):
I would like to have a discussion about fix, to be sure that I got it correctly.

logger/mmap_buffer.cc
DenisShkiria: The commit function buffers records before sending them to ClickHouse:
```
bool
TfwClickhouse::commit()
{
	auto now = now_ms();

	block_.RefreshRowCount();
	if ((now - last_time_ > max_wait_ && block_.GetRowCount() > 0)  <------ bufferization condition
	    || block_.GetRowCount() > max_events_) {
	    ...
```
After these changes, the buffered records will be sent to ClickHouse once the `max_wait_` maximum wait time has passed.

logger/parser.cc
```diff
@@ -0,0 +1,149 @@
...
+	}
+}
+
+[[nodiscard]] std::optional<AccessLog>
+Parser::parse_access_log_event(const TfwBinLogEvent &event)
+{
+	AccessLog access_log;
+	access_log.timestamp = std::chrono::milliseconds(event.timestamp);
+
+	static constexpr std::array kSteps = {
```
krizhanovsky, line 112: we use CamelCase only for types, see https://github.com/tempesta-tech/tempesta/blob/master/CodingStyle

fw/access_log.h
```diff
@@ -79,26 +79,6 @@ typedef enum {
...
-		[TFW_MMAP_LOG_URI] = 0,
-		[TFW_MMAP_LOG_REFERER] = 0,
-		[TFW_MMAP_LOG_USER_AGENT] = 0,
-		[TFW_MMAP_LOG_JA5T] = 8,
-		[TFW_MMAP_LOG_JA5H] = 8,
-		[TFW_MMAP_LOG_DROPPED] = 8
-	};
-	return TfwBinLogFieldsLens[field];
-}
-
```
krizhanovsky, line 101: In general, this make the code quite fragile - now we can easily change the C code w/o updating the C++ part. The C and C++ parts do depend on each other and this must be expressed via headers.

logger/access_log.hh
```diff
@@ -0,0 +1,108 @@
...
+	uint32_t response_content_length;
+	uint32_t response_time;
+	std::string vhost;
+	std::string uri;
+	std::string referer;
+	std::string user_agent;
+	uint64_t ja5t;
+	uint64_t ja5h;
+	uint64_t dropped_events;
+};
```
krizhanovsky, line 45: we align structure members with spaces. We align structure members to minimize memory holes

logger/tfw_logger.cc
```diff
@@ -209,20 +215,23 @@ read_access_log_event(const char *data, int size, TfwClickhouse *clickhouse)
...
-		size -= len;                                                   \
-	} else                                                                 \
-		(*block)[ind]->As<clickhouse::ColumnString>()->Append(         \
-		    std::string(""));
+#define READ_STR(method)							\
+	do {									\
+		const int ind = method + 1; /* column 0 is timestamp */		\
+		if (TFW_MMAP_LOG_FIELD_IS_SET(event, method)) {			\
+			size_t len = *reinterpret_cast<const uint16_t *>(p);	\
+			if (len + 2 > size) [[unlikely]]			\
```
consuelo2210: Can you explain this change? It looks like a bug was here, but what a magic "+2"? It was before, but I am not in context of this project

logger/mmap_buffer.cc
```diff
@@ -87,38 +99,25 @@ TfwMmapBufferReader::run(std::atomic<bool> *stop_flag)
...
-	u64 head, tail;
-
-	head = __atomic_load_n(&buf_->head, __ATOMIC_ACQUIRE);
-	tail = buf_->tail;
+	const auto head = __atomic_load_n(&buf_->head, __ATOMIC_ACQUIRE);
+	const auto tail = buf_->tail;
 
-	if (head - tail == 0) [[unlikely]]
-		return -EAGAIN;
+	const char *data = buf_->data + (tail & buf_->mask);
```
consuelo2210: I think this instruction is better to be after size check

logger/mmap_buffer.hh
```diff
@@ -48,29 +46,27 @@ typedef void (*TfwMmapBufferReadCallback)(const char *data, int size,
...
  *    @read - checks if there is a new data block and executes the callback when
  *        new data is detected.
  */
 class TfwMmapBufferReader {
 public:
-	TfwMmapBufferReader(const unsigned int ncpu, const int fd, void *private_data,
-			    TfwMmapBufferReadCallback cb);
+	using Callback = std::function<void(std::span<const char> data)>;
+
+	TfwMmapBufferReader(unsigned int ncpu, int fd, Callback cb);
```
consuelo2210: Is it a regular way for callbacks? Not by && and with std::forward inside?

logger/mmap_buffer.cc
consuelo2210: Ah, I see you removed this check later

logger/mmap_buffer.cc
consuelo2210: Are you sure that we will reach max_wait_ time in the second time? What if not?

I also have a question about force flag or work in destructor. Why this function doesn't have a force flag to flush something on shutdown, for example?

consuelo2210 (changes requested):
I love the changes in the first refactoring part: I really hate to have pointer in the places where reference has to be, also I appreciate using std::span for ranges instead of c-style pointer+size and love std::function instead c-style callbacks. I feel that these changes were good. But I wasn't able to review the second part of refactoring: for the person outside the project (as I am) it would be much more easier to review this concrete commit in separated moving steps, every in separate commit to make review process more reviewer-friendly.

What about 2 main commits:
The fix of problem doesn't look right for me. But I glad to see changes in spdlog file.

logger/mmap_buffer.cc
DenisShkiria: > Are you sure that we will reach max_wait_ time in the second time? What if not?

If we do not reach the maximum wait time the first time, we will reach it on one of the subsequent attempts. The current polling period is 10 ms, so in the worst case, we will have a maximum wait time divided by 10 ms attempts because each attempt now calls commit().

> I also have a question about force flag or work in destructor. Why this function doesn't have a force flag to flush something on shutdown, for example?

That's a good point. Thanks! Indeed, during shutdown, we may encounter a situation in which some records have been read from shared memory but not sent to ClickHouse because they were buffered in the block buffer. To implement a graceful shutdown, we should try to flush these buffered records before exiting. I will add this logic, though I'm not sure in which PR.

fw/access_log.h
DenisShkiria: This function was not used in the C part, so this change does not affect the connection between the C and C++ parts. However, I agree that it would be beneficial to bind the C++ and C logic at the compilation level.

logger/tfw_logger.cc
DenisShkiria: Sorry, it seems that I did not implement this fix properly. Please take a look at this [commit](https://github.com/tempesta-tech/tempesta/pull/2493/commits/2b20688b16f8d9fec567d084f7c5151c1b2e07f9), I believe it is correct. The main idea is that we don't check if we have enough data in the buffer when parsing a string data structure.

PR #2484, opened 2025-08-11
Fix accounting in frang

EvgeniiMekhanik:
Previously Tempesta FW accounts blocked connections/requests in frang history. This can lead to a problem - if limit of connections/requests is equal to N * (FRANG_REQ - 1) and client opens N connection / sends N request in a one slot. In this case after (FRANG_REQ - 1) period all new connections /requests will be blocked.
This patch fixes this behaviour now Tempesta FW doesn't decrement appropriate counter if frang block connection/request.


fw/http_limits.c
```diff
@@ -181,6 +181,12 @@ frang_req_is_whitelisted(TfwHttpReq *req)
 	return frang_sk_is_whitelisted(req->conn->sk);
 }
 
+static int
+frang_time_in_frame(const unsigned long tcur, const unsigned long tprev)
```
keshonok, line 185: Why not make it inline? Perhaps, it's inlined by the compiler's optimizer but I am not so sure of that.

fw/http_limits.c
```diff
@@ -1501,17 +1510,19 @@ frang_tls_conn_limit(FrangAcc *ra, FrangGlobCfg *conf, int hs_state)
 		break;
 	}
 
-	for (i = 0; i < FRANG_FREQ; i++)
-		if (ra->history[i].ts + FRANG_FREQ >= ts) {
-			sum_new += ra->history[i].tls_sess_new;
-			sum_incomplete += ra->history[i].tls_sess_incomplete;
+	for (j = 0; j < FRANG_FREQ; j++) {
+		if (frang_time_in_frame(ts, ra->history[j].ts)) {
```
const-t, line 1514: @krizhanovsky @EvgeniiMekhanik  The patch looks good and it fix the issue. However should we consider this as an issue? It seems current implementation has some advantages, it slows too fast clients, however introduces some inaccuracy. Is it worth to fix?

Issue #2477, opened 2025-08-04
doc: move description from `tempesta_fw.conf` to  wiki


RomanBelozerov:
# Scope

`tempesta_fw.conf` contain a lot of recommendations\info about directives to setting TempestaFW. It looks like duplicates for our wiki. 
This file should contain the recommended basic configuration, and not be a duplicate of our wiki.

Please move these recommendations to wiki if it is not a duplicate. 

And you should check https://github.com/tempesta-tech/tempesta/wiki/Development-guidelines#documentation before.

The old version was removed - https://github.com/tempesta-tech/tempesta/pull/2478

PR #2487, opened 2025-08-12
remove old version tempesta_fw.conf and add recommended basic config[port 0.8]

const-t:
- remove old version (see #2477);
- add recommended basic config (see #2451);


krizhanovsky (approved):
LGTM

PR #2490, opened 2025-08-13
Fix bug if stale response build fails

EvgeniiMekhanik:
We should zeroed `hmresp->pair` before trying to
build stale response, otherwise we access `hmresp->pair` later during response freeing.


fw/http.c
```diff
@@ -1560,6 +1560,7 @@ __tfw_http_resp_fwd_stale(TfwHttpMsg *hmresp)
 	TfwHttpResp *stale_resp;
 
 	req->resp = NULL;
+	hmresp->pair = NULL;
```
```suggestion
	tfw_http_msg_unpair(hmresp);
```

PR #2492, opened 2025-08-14
Fix bug if stale response build fails

EvgeniiMekhanik:
We should zeroed `hmresp->pair` before trying to
build stale response, otherwise we access `hmresp->pair` later during response freeing. Look we build new response in `tfw_h1_send_resp` so `req->pair` will point to the new response, later both this request and response will be freed in `tfw_http_resp_fwd`. Old request (hmresp) will be freed from `tfw_sock_srv_connect_drop` but `hmresp->pair` point to the already freed response! To prevent it we should zeroed `hmresp->pair` as was described below.

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -1556,7 +1556,7 @@ __tfw_http_resp_fwd_stale(TfwHttpMsg *hmresp)
 	TfwHttpReq *req = hmresp->req;
 	TfwHttpResp *stale_resp;
 
-	req->resp = NULL;
+	tfw_http_msg_unpair(hmresp);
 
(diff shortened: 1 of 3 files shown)


fw/http.c
```diff
@@ -1557,6 +1557,7 @@ __tfw_http_resp_fwd_stale(TfwHttpMsg *hmresp)
 	TfwHttpResp *stale_resp;
 
 	req->resp = NULL;
+	hmresp->pair = NULL;
```
```suggestion
	tfw_http_msg_unpair(hmresp);
```

const-t (approved):

Issue #2437, opened 2025-06-08
Thrash results in Clickhouse access log


krizhanovsky:
**Describe the issue**
On the new website setup installation I get thrash results from clickhouse:
```SQL
SELECT
    timestamp,
    method,
    version,
    status,
    response_content_length,
    response_time,
    vhost,
    uri,
    referer,
    user_agent,
    ja5t,
    ja5h
FROM default.access_log
WHERE user_agent NOT IN ('baremetal', 'tempesta-1', 'tempesta-2', 'vm.web')
ORDER BY response_time DESC
LIMIT 5
FORMAT Pretty

Query id: 37fa2dc3-8ed0-483f-80b2-d14d3e889d05

   ┏━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━┳━━━━━━┓
   ┃               timestamp ┃ method ┃ version ┃ status ┃ response_content_length ┃ response_time ┃ vhost ┃ uri ┃ referer ┃ user_agent ┃ ja5t ┃ ja5h ┃
   ┡━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━━╇━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━╇━━━━━━┩
1. │ 2025-06-06 10:49:01.597 │      0 │       0 │    400 │                       0 │    4294928032 │       │     │ -       │ -          │    0 │    0 │
   ├─────────────────────────┼────────┼─────────┼────────┼─────────────────────────┼───────────────┼───────┼─────┼─────────┼────────────┼──────┼──────┤
2. │ 2025-06-06 10:48:53.182 │      0 │       0 │    400 │                       0 │    4294919616 │       │     │ -       │ -          │    0 │    0 │
   ├─────────────────────────┼────────┼─────────┼────────┼─────────────────────────┼───────────────┼───────┼─────┼─────────┼────────────┼──────┼──────┤
[...]

Issue #2442, opened 2025-06-10
Get rid of busy loop in Tfw logger


const-t:
# Motivation

Now tfw logger reads data from mapped memory every 10ms using busy loop, that seems ineffective and we may have log drops.
 
# Scope
I would suggest to rework this part and get rid of busy-loop. Instead of it we can using polling for each file descriptor. We can achieve that allocating per-cpu buffers in multiple virtual files, instead of single file and then Tfw logger's thread will poll_wait this file descriptor. Once mapped region is changed Tempesta must notify waiting thread. We can use procfs or securityfs for these files, instead of `/dev`. In the Tempesta FW part we need to implement polling waiting/notification (see `struct file_operations`).

consuelo2210:
Do you have any ideas, how to do it with plugins? Plugin don't have their own threads and work only on the thread of tfw_logger. Should they share fd-s for waiting on poll? Is it possible per cpu?

Issue #2399, opened 2025-04-09
Security events observability


krizhanovsky:
# Motivation

We have dozens of conditions to block a request due to security events. Typically we print a security warning to dmesg for this. However, it may flood the log and make the system slower since there are a lot of malicious bots in the Internet. Meantime we need an observability tool to be able to debug application issues. This issue is the "security" part of #537

# Scope

## Clickhouse logging

We need a security log like [access log](https://tempesta-tech.com/knowledge-base/Handling-clients/#clickhouse), but to a different Clickhouse table. Instead of response data we need a enum for the reason why a request or client connection was blocked.

Also need to make the current table for access logging configurable in the `tfw_logger` config.

The enum values must have good and verbose descriptions as C strings as well as short string descriptors. The values and strings should be inserted into a clickhouse table by the `tfw_logger` daemon or a script - TBD.

See JSON configuration example in #2421 

### Events format

#2509 and testing task https://github.com/tempesta-tech/tempesta-test/issues/866 require client port in the events, so the security events, and accordingly the Clickhouse table, must contain fields:

1. client IPv6
2. client port
3. event type, at least 2 bytes, probably 4, - we'll have a lot of events, like "requests burst exceeded"
4. event body, a binary array. E.g. the number of requests, bytes from an HTTTP fields, blocked by parsers etc. This filed must be interpreted depending on the event type. Do not print it in text format to not to spend cycles in softirq and minimize the traffic sent to Clickhouse. We can easily format this with helping tools.
5. Tempesta FW IP/port that triggered the event.
6. flag that `ip_block` rule was applied for the IP

## Counters

Need to implement a /proc counters with the names as short string descriptors. The counters must be collected per-cpu. Ideally per-vhost and/or per server - TBD

## Security event visibility

Tempesta FW handles at least Web-cache deception, HTTP Response Splitting, HTTP Request Smuggling, some SQL injection and XSS and L7 DDoS attacks. Some of them, e.g. HTTP Response Splitting, a logged, but others, e.g. seems Webcache deception according to https://github.com/tempesta-tech/tempesta-test/blob/master/cache/test_cache_deception_attack.py , aren't logged at all. Some of the clients wants to get clear visibility of all blocked attacks as the product KPI, so please make sure that we properly log at least these attacks. Appropriate tests adjustments would be great.

# Testing

https://github.com/tempesta-tech/tempesta-test/issues/796

# Documentation

[...]

PR #2493, opened 2025-08-14
Denis shkiria 2314 tfw logger missing logs fix

### LINKED ISSUE #2314: tfw_logger: missing logs during the first second of application startup

DenisShkiria:
Fixes #2314 and #2437, massive architecture reworking.

### DIFF
--- a/fw/http.h
+++ b/fw/http.h
@@ -276,6 +276,7 @@ typedef struct {
  * @keep_alive		- the value of timeout specified in Keep-Alive header;
  * @iter		- skb expansion iterator;
  * @content_length	- the value of Content-Length header field;
+ * @jrxtstamp		- time the message has been received, in jiffies;
  * @flags		- message related flags. The flags are tested
  *			  concurrently, but concurrent updates aren't
  *			  allowed. Use atomic operations if concurrent
@@ -305,6 +306,7 @@ typedef struct {
 	unsigned int	keep_alive;					\
 	TfwMsgIter	iter;						\
 	unsigned long	content_length;					\
+	unsigned long	jrxtstamp;					\
 	DECLARE_BITMAP	(flags, _TFW_HTTP_FLAGS_NUM);			\
 	TfwConn		*conn;						\
 	void (*destructor)(void *msg);					\
@@ -414,7 +416,6 @@ struct tfw_http_req_t {
 	struct list_head	fwd_list;
 	struct list_head	nip_list;
 	unsigned long		jtxtstamp;
-	unsigned long		jrxtstamp;
 	union {
 		unsigned long		tm_header;
 		long			stale_ce_age;
@@ -487,7 +488,6 @@ typedef struct {
  * HTTP Response.
  * TfwStr members must be the first for efficient scanning.
  *
- * @jrxtstamp	    - time the message has been received, in jiffies;
  * @mit		    - iterator for controlling HTTP/1.1 => HTTP/2 message
  *		      transformation process (applicable for HTTP/2 mode only).
  * @no_cache_tokens - tokens for cache-control directive e.g.
@@ -507,7 +507,6 @@ struct tfw_http_resp_t {
 	unsigned short		status;
 	long			date;
 	long			last_modified;
-	unsigned long		jrxtstamp;
 	TfwHttpTransIter	mit;
 	TfwStr			no_cache_tokens;
 	TfwStr			private_tokens;
--- a/fw/http.h
+++ b/fw/http.h
@@ -379,7 +379,6 @@ typedef struct {
  * @fwd_list	- member in the queue of forwarded/backlogged requests;
  * @nip_list	- member in the queue of non-idempotent requests;
  * @jtxtstamp	- time the request is forwarded to a server, in jiffies;
- * @jrxtstamp	- time the request is received from a client, in jiffies;
  * @tm_header	- time HTTP header started coming. Only rx path;
  * @stale_ce_age - calculated age of stale response. Must be assigned only when
  *		  "cache_use_stale" is configured on tx path with cache;
--- a/logger/clickhouse.cc
+++ b/logger/clickhouse.cc
@@ -41,8 +41,8 @@ constexpr std::string_view table_creation_query_template =
 	"ENGINE = MergeTree() ORDER BY timestamp";
 
 TfwClickhouse::TfwClickhouse(const ClickHouseConfig &config,
-			     clickhouse::Block block)
-	: block_(std::move(block)),
+			     clickhouse::Block *block)
+	: block_(block),
 	  last_time_(now_ms()),
 	  table_name_(config.table_name),
 	  max_events_(config.max_events),
@@ -69,7 +69,7 @@ TfwClickhouse::TfwClickhouse(const ClickHouseConfig &config,
 clickhouse::Block *
 TfwClickhouse::get_block() noexcept
 {
-	return &block_;
+	return block_;
 }
 
 bool
@@ -77,14 +77,13 @@ TfwClickhouse::commit()
 {
 	auto now = now_ms();
 
-	block_.RefreshRowCount();
-	if ((now - last_time_ > max_wait_ && block_.GetRowCount() > 0)
-	    || block_.GetRowCount() > max_events_) {
+	block_->RefreshRowCount();
+	if ((now - last_time_ > max_wait_ && block_->GetRowCount() > 0)
+	    || block_->GetRowCount() > max_events_) {
 
-		client_->Insert(table_name_, block_);
+		client_->Insert(table_name_, *block_);
 
-		for (size_t i = 0; i < block_.GetColumnCount(); ++i)
-			block_[i]->Clear();
+		block_->Clear();
 
 		last_time_ = now;
 
--- a/logger/clickhouse.cc
+++ b/logger/clickhouse.cc
@@ -1,6 +1,11 @@
 /**
  *		Tempesta FW
  *
+ * Clickhouse interfaces using the C++ client library.
+ * For code sameples and the source code reference:
+ *
+ *   https://github.com/ClickHouse/clickhouse-cpp.git
+ *
  * Copyright (C) 2024-2025 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
@@ -17,35 +22,94 @@
  * this program; if not, write to the Free Software Foundation, Inc., 59
  * Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  */
-
 #include <iostream>
 #include <string_view>
 
 #include <fmt/format.h>
 
+#include <spdlog/spdlog.h>
+
+#include "../fw/access_log.h"
+#include "error.hh"
 #include "clickhouse.hh"
 
-static auto
-now_ms()
-{
-	return std::chrono::duration_cast<std::chrono::milliseconds>(
-		std::chrono::system_clock::now().time_since_epoch());
-}
+namespace {
 
 constexpr std::string_view table_creation_query_template = 
 	"CREATE TABLE IF NOT EXISTS {} "
-	"(timestamp DateTime64(3, 'UTC'), address IPv6, method UInt8, "
-	"version UInt8, status UInt16, response_content_length UInt32, "
-	"response_time UInt32, vhost String, uri String, referer String, "
-	"user_agent String, ja5t UInt64, ja5h UInt64, dropped_events UInt64) "
-	"ENGINE = MergeTree() ORDER BY timestamp";
+	"(timestamp DateTime64(3, 'UTC'),"
+	" address IPv6,"
+	" method UInt8,"
+	" version UInt8,"
+	" status UInt16,"
+	" response_content_length UInt64,"
+	" response_time UInt32,"
+	" vhost String,"
+	" uri String,"
+	" referer String,"
+	" user_agent String,"
+	" ja5t UInt64,"
+	" ja5h UInt64,"
+	" dropped_events UInt64"
+	") ENGINE = MergeTree() ORDER BY timestamp";
 
-TfwClickhouse::TfwClickhouse(const ClickHouseConfig &config, ch::Block block)
-	: block_(std::move(block)),
-	  last_time_(now_ms()),
-	  table_name_(config.table_name),
-	  max_events_(config.max_events),
-	  max_wait_(config.max_wait)
+typedef struct {
+	const char			*name;
+	ch::Type::Code			code;
+} TfwField;
+
+static const TfwField tfw_fields[] = {
+	[TFW_MMAP_LOG_ADDR]		= {"address", ch::Type::IPv6},
+	[TFW_MMAP_LOG_METHOD]		= {"method", ch::Type::UInt8},
+	[TFW_MMAP_LOG_VERSION]		= {"version", ch::Type::UInt8},
+	[TFW_MMAP_LOG_STATUS]		= {"status", ch::Type::UInt16},
+	[TFW_MMAP_LOG_RESP_CONT_LEN]	= {"response_content_length", ch::Type::UInt32},
+	[TFW_MMAP_LOG_RESP_TIME]	= {"response_time", ch::Type::UInt32},
+	[TFW_MMAP_LOG_VHOST]		= {"vhost", ch::Type::String},
+	[TFW_MMAP_LOG_URI]		= {"uri", ch::Type::String},
+	[TFW_MMAP_LOG_REFERER]		= {"referer", ch::Type::String},
+	[TFW_MMAP_LOG_USER_AGENT]	= {"user_agent", ch::Type::String},
+	[TFW_MMAP_LOG_JA5T]		= {"ja5t", ch::Type::UInt64},
+	[TFW_MMAP_LOG_JA5H]		= {"ja5h", ch::Type::UInt64},
+	[TFW_MMAP_LOG_DROPPED]		= {"dropped_events", ch::Type::UInt64}
+};
+
+} // anonymous namespace
+
+void
+TfwClickhouse::make_block()
+{
+	block_ = ch::Block();
+
+	auto col = std::make_shared<ch::ColumnDateTime64>(3);
+	block_.AppendColumn("timestamp", col);
+
+	for (int i = TFW_MMAP_LOG_ADDR; i < TFW_MMAP_LOG_MAX; ++i) {
+		const TfwField *field = &tfw_fields[i];
+		auto col = tfw_column_factory(field->code);
+		block_.AppendColumn(field->name, col);
+	}
+
+	// We may read more data in one shot, so reserve more memory.
+	block_.Reserve(max_events_ * 2);
+}
+
+bool
+TfwClickhouse::handle_block_error() noexcept
+{
+	try {
+		block_.Clear();
+		return true;
+	}
+	catch (const std::exception &e) {
+		spdlog::error("Cannot clear a Clickhouse block: {}", e.what());
+	}
+	return false;
+}
+
+TfwClickhouse::TfwClickhouse(const ClickHouseConfig &config)
+	: table_name_(config.table_name),
+	  max_events_(config.max_events)
 {
 	auto opts = clickhouse::ClientOptions();
 
@@ -63,6 +127,13 @@ TfwClickhouse::TfwClickhouse(const ClickHouseConfig &config, ch::Block block)
 	std::string table_creation_query =
 		fmt::format(table_creation_query_template, table_name_);
 	client_->Execute(table_creation_query);
+
+	make_block();
+}
+
+TfwClickhouse::~TfwClickhouse()
+{
+	handle_block_error();
 }
 
 clickhouse::Block &
@@ -71,24 +142,24 @@ TfwClickhouse::get_block() noexcept
 	return block_;
 }
 
-bool
-TfwClickhouse::commit(bool force)
+[[nodiscard]] Error<bool>
+TfwClickhouse::commit(bool force) noexcept
 {
-	auto now = now_ms();
+	try {
+		block_.RefreshRowCount();
 
-	block_.RefreshRowCount();
-	if ((now - last_time_ > max_wait_ && block_.GetRowCount() > 0)
-	    || block_.GetRowCount() > max_events_ || force) {
+		if (block_.GetRowCount() < max_events_ && !force)
+			return false;
 
 		client_->Insert(table_name_, block_);
-
 		block_.Clear();
-
-		last_time_ = now;
-
-		return true;
 	}
-	return false;
+	catch (const std::exception &e) {
+		spdlog::error("Clickhouse insert error: {}", e.what());
+		return error(Err::DB_SRV_FATAL);
+	}
+
+	return true;
 }
 
 template <typename T> std::shared_ptr<clickhouse::Column>
--- a/logger/clickhouse.cc
+++ b/logger/clickhouse.cc
@@ -73,13 +73,13 @@ TfwClickhouse::get_block() noexcept
 }
 
 bool
-TfwClickhouse::commit()
+TfwClickhouse::commit(bool force)
 {
 	auto now = now_ms();
 
 	block_->RefreshRowCount();
 	if ((now - last_time_ > max_wait_ && block_->GetRowCount() > 0)
-	    || block_->GetRowCount() > max_events_) {
+	    || block_->GetRowCount() > max_events_ || force) {
 
 		client_->Insert(table_name_, *block_);
 
--- a/logger/clickhouse.cc
+++ b/logger/clickhouse.cc
@@ -2,7 +2,7 @@
  *		Tempesta FW
  *
  * Clickhouse interfaces using the C++ client library.
- * For code sameples and the source code reference:
+ * For code samples and the source code reference:
  *
  *   https://github.com/ClickHouse/clickhouse-cpp.git
  *
@@ -111,7 +111,7 @@ TfwClickhouse::TfwClickhouse(const ClickHouseConfig &config)
 	: table_name_(config.table_name),
 	  max_events_(config.max_events)
 {
-	auto opts = clickhouse::ClientOptions();
+	auto opts = ch::ClientOptions();
 
 	opts.SetHost(config.host);
 	opts.SetPort(config.port);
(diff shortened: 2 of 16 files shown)


logger/clickhouse.cc
```diff
@@ -41,8 +41,8 @@ constexpr std::string_view table_creation_query_template =
 	"ENGINE = MergeTree() ORDER BY timestamp";
 
 TfwClickhouse::TfwClickhouse(const ClickHouseConfig &config,
-			     clickhouse::Block block)
-	: block_(std::move(block)),
+			     clickhouse::Block *block)
```
consuelo2210: Why not but ref? :)

logger/tfw_logger.cc
```diff
@@ -305,6 +305,7 @@ try {
 	}
 	catch (const Exception &e) {
 		log_error(e.what(), true, false);
+		block.Clear();
```
consuelo2210: don't we need to clear it in any exception? It would be good at least to leave a comment: why we are doing clear and when, and when we are trying to flush after stop

logger/tfw_logger.cc
```diff
@@ -317,6 +318,21 @@ try {
...
+	block.RefreshRowCount();
+	while (block.GetRowCount() > 0)
+	try {
+		spdlog::debug("Worker {} flushing remaining events...", ncpu);
+		TfwClickhouse clickhouse(config.clickhouse, &block);
+		clickhouse.commit(/*force=*/true);
+	}
+	catch (const std::exception &e) {
+		log_error(e.what(), true, false);
+		std::this_thread::sleep_for(timeout);
```
consuelo2210: but we are stopping, why do we need to sleep and change timeout?

logger/clickhouse.cc
DenisShkiria: The both options looks questionable here, so I choose the pointer to because for some reason the following function returns pointer:
```
	clickhouse::Block *get_block() noexcept;
```
and because the form with `&` looks a bit clearer in the following line:
```
TfwClickhouse clickhouse(ch_cfg, &block);
```
But basically the both options are about equal here.

logger/tfw_logger.cc
DenisShkiria: The current logic considers exceptions of the type Exception unrecoverable. The while(1) loop stops when an exception arises, and the thread finishes. In this case, we do not try to flush the contents of the buffer because it can contain corrupted data. For example, the internal rows of the block may have different sizes. When the exception means something else, the block should already be cleared.

> don't we need to clear it in any exception?

No. The current implementation uses exceptions to reestablish the connection to ClickHouse when the ClickHouse-cpp library cannot insert the block into the database. If, for some reason, the block cannot be sent to the DB, an exception is raised, and the connection to the DB is reestablished. However, the block should not be cleared in this case, otherwise part of its data could be lost.

I agree that adding a comment with these details here is worth it. In my opinion, rewriting code to simplify it is better than writing long comments. However, comments are better than nothing.

logger/tfw_logger.cc
DenisShkiria: No, if we're here, it means ClickHouse-cpp lib encountered an issue and threw an exception. We should reestablish the connection and try again.

DenisShkiria:
> the destructor is more propriate way for block cleaning.

Unfortunately, as far as I understand, incorporating flush logic into the destructor is more labor-intensive. When the ClickHouse C++ library sends data to the database, it throws an exception in case of an error. The [recommended](https://github.com/ClickHouse/clickhouse-cpp/issues/184) way to handle the exception is to reset the connection to the DB. To send data to the DB, you should write the code similar to the following:
```
while (true) {
  if (stop_requested())
    break;
    
  try {
    log(inserting...)
    connection->insert(block);
    log (success!)
    break;
  }
  catch (e) {
    log(failed to insert)
  }
  reconnect(connection)
}
```
It is possible to write something similar in the `~TfwClickhouse()`, but we would need to modify `TfwClickhouse()`. I decided that moving the flush logic into a separate cycle would be a less intrusive change.

logger/tfw_logger.cc
consuelo2210: But in this case we just increasing timeout after each failed try. Let's imagine that there is some issue with CH and we can't reconnect to it. So does the program just hang on stop? It doesn't sound right. And I can't see another condition of breaking this circle

logger/clickhouse.cc
consuelo2210: I think it is better to move in the direction when in case of object presence we have a ref. Especially that we create that block with make_block. But this is really minor for this particular review. I am going to close this discussion

logger/tfw_logger.cc
DenisShkiria: Sorry, when I wrote the first comment, I thought we were discussing the previous while(1) {...} cycle. Yes, the suggested logic ignores the stop flag when the block is not empty and tries to send data to ClickHouse until the end. If the connection to ClickHouse has been lost and cannot be reestablished and the block is not empty, then this cycle will continue indefinitely. This also means that the program cannot stop gracefully. My logic is that if the user does not want to wait for a graceful shutdown, which can take an unpredictable amount of time, he or she should send a SIGKILL.

logger/mmap_buffer.cc
```diff
@@ -113,12 +113,14 @@ TfwMmapBufferReader::read()
 	head = __atomic_load_n(&buf_->head, __ATOMIC_ACQUIRE);
 	tail = buf_->tail;
 
-	if (head - tail == 0) [[unlikely]]
-		return -EAGAIN;
+	assert(head >= tail);
+	const int size = static_cast<int>(head - tail);
 
-	callback_(buf_->data + (tail & buf_->mask), head - tail, private_data_);
+	callback_(buf_->data + (tail & buf_->mask), size, private_data_);
```
krizhanovsky: I think it'd be better to not to call `callback()` each time the thread wakes up. Also this fix will break with #2442, when we move from polling to event-driven operation.

Essentially, for now we should emulate call Clickhouse only if a time elapses. I'm thinking about State pattern for `TfwClickhouse`, which will keep track whether we have bach buffer clear or do we need to flush something. This State-like object can replace `callback()`.

The problem isn't so crucial, but there is some overhead and we also do some senseless printing in `dbg_hexdump()` - just might be annoying.

logger/clickhouse.cc
krizhanovsky: Yeah, I think it's also better to prefer reference to pointer as in general it provides more type safety. I'd prefer to adjust the current code to return references.

BTW we have Memory safety guidelines https://tempesta-tech.com/knowledge-base/Memory-safety-guideline/ , which are just a some additions to the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines) - our code must follow the principles. This particular case means that a reviewer overlooked the issue

logger/tfw_logger.cc
```diff
@@ -212,12 +212,16 @@ read_access_log_event(const char *data, int size, TfwClickhouse *clickhouse)
...
+		size -= len_size;                                              \
 		if (len > size) [[unlikely]]                                   \
 			goto error;                                            \
 		(*block)[ind]->As<clickhouse::ColumnString>()->Append(         \
-		    std::string(p + 2, len));                                  \
-		len += 2;                                                      \
+		    std::string(p, len));                                      \
 		p += len;                                                      \
 		size -= len;                                                   \
 	} else                                                                 \
```
krizhanovsky: The change is good, but I'd also replace the macro with a template function

fw/http.h
```diff
@@ -414,7 +416,6 @@ struct tfw_http_req_t {
 	struct list_head	fwd_list;
 	struct list_head	nip_list;
 	unsigned long		jtxtstamp;
-	unsigned long		jrxtstamp;
```
const-t, line 417: Please remove description from the doc of `tfw_http_req_t`.

const-t (approved):
LGTM

logger/tfw_logger.cc
```diff
@@ -173,102 +170,116 @@ make_block()
...
+	read_int<TFW_MMAP_LOG_ADDR, ch::ColumnIPv6, in6_addr>(block, event, p, size);
+	read_int<TFW_MMAP_LOG_METHOD, ch::ColumnUInt8, unsigned char>(
+							block, event, p, size);
+	read_int<TFW_MMAP_LOG_VERSION, ch::ColumnUInt8, unsigned char>(
+							block, event, p, size);
+	read_int<TFW_MMAP_LOG_STATUS, ch::ColumnUInt16, uint16_t>(
+							block, event, p, size);
+	read_int<TFW_MMAP_LOG_RESP_CONT_LEN, ch::ColumnUInt32, uint32_t>(
+							block, event, p, size);
+	read_int<TFW_MMAP_LOG_RESP_TIME, clickhouse::ColumnUInt32, uint32_t>(
```
consuelo2210: clickhouse ->cn

logger/tfw_logger.cc
```diff
@@ -173,102 +170,116 @@ make_block()
...
+							block, event, p, size);
```
consuelo2210: clickhouse -> cn

logger/tfw_logger.cc
```diff
@@ -173,102 +170,116 @@ make_block()
 	return block;
 }
 
-int
-read_access_log_event(const char *data, int size, TfwClickhouse *clickhouse)
+template <auto method, typename ColType, typename ValType>
+void
+read_int(ch::Block &block, const auto *event, const char *&p, size_t &size)
```
consuelo2210: const char *& -> is barely used, I would suggest to use string_view&

consuelo2210: I'd also prefer to have wrapper around this string and around Block. I asked chatgpt to split for this 2 criterias, and I haven't verify the code inside. But the proposal looks similar to me idea:
```cpp
   struct BlockWriter {
    ch::Block& block;

    BlockWriter(ch::Block& blk) : block(blk) {}

    template <typename ColType, typename ValType>
    void write_int(size_t index, ValType val) {
        block[index]->As<ColType>()->Append(val);
    }

    void write_string(size_t index, std::string_view sv) {
        block[index]->As<ch::ColumnString>()->Append(std::string(sv));
    }

    void write_empty_string(size_t index) {
        block[index]->As<ch::ColumnString>()->Append(std::string(""));
    }
};

struct LineReader {
    std::string_view buf;  // buffer to read

    LineReader(const char* data, size_t size) : buf(data, size) {}

    template <typename ValType>
    ValType read_int() {
        if (sizeof(ValType) > buf.size())
            throw std::runtime_error("Buffer too small for integer");

        const ValType* val_ptr = reinterpret_cast<const ValType*>(buf.data());
        buf.remove_prefix(sizeof(ValType));
        return *val_ptr;
    }

    std::string_view read_u16_string() {
        constexpr size_t len_size = sizeof(uint16_t);

        if (len_size > buf.size())
            throw std::runtime_error("Buffer too small for string length");

        const uint16_t len = *reinterpret_cast<const uint16_t*>(buf.data());
        buf.remove_prefix(len_size);

        if (len > buf.size())
            throw std::runtime_error("Buffer too small for string content");

        std::string_view sv(buf.data(), len);
        buf.remove_prefix(len);
        return sv;
    }

    bool empty() const { return buf.empty(); }
};

//TODO: rename to consume_int
template <auto method, typename ColType, typename ValType>
void read_int(const auto* event, LineReader& reader, BlockWriter& writer) {
    constexpr int ind = method + 1;

    if (TFW_MMAP_LOG_FIELD_IS_SET(event, method)) {
        ValType val = reader.read_int<ValType>();
        writer.write_int<ColType>(ind, val);
    } else {
        writer.write_int<ColType>(ind, 0);
    }
}

//TODO: rename to consume_str
template <auto method>
void read_str(const auto* event, LineReader& reader, BlockWriter& writer) {
    constexpr int ind = method + 1;

    if (TFW_MMAP_LOG_FIELD_IS_SET(event, method)) {
        std::string_view sv = reader.read_u16_string();
        writer.write_string(ind, sv);
    } else {
        writer.write_empty_string(ind);
    }
}
```

consuelo2210: The main difference will be in error handling. goto in this case is not a good idea

logger/clickhouse.cc
```diff
@@ -1,6 +1,11 @@
 /**
  *		Tempesta FW
  *
+ * Clickhouse interfaces using the C++ client library.
+ * For code sameples and the source code reference:
```
consuelo2210:
```suggestion
 * For code samples and the source code reference:
```

logger/mmap_buffer.hh
```diff
@@ -53,22 +56,23 @@
  */
 class TfwMmapBufferReader {
 public:
-	using Callback = std::function<void(std::span<const char> data)>;
+	using ProcessEvents = std::function<Error<size_t> (std::span<const char>)>;
```
consuelo2210: ProcessEvents -> EventBufferProcessor ? or EventProcessor ?  ProcessEventsFn ?ProcessEvents looks like we have a vector inside

logger/tfw_logger.cc
```diff
@@ -531,7 +531,16 @@ try {
...
 	logger->flush_on(spdlog::level::info);
-} catch (const spdlog::spdlog_ex &ex) {
+
+	// Set custom log pattern to include thread ID
+	// %Y-%m-%d %H:%M:%S.%e - Date and time with milliseconds
+	// %n - Logger name
+	// %l - Log level
+	// %t - Thread ID
+	// %v - Log message
+	logger->set_pattern("[%Y-%m-%d %H:%M:%S.%e] [%n] [%l] [%t] %v");
```
consuelo2210, line 581: with this change I asked Denis to fix line 516:

if (!opts.foreground) {
		spdlog::debug("Daemonizing...");
		
We haven't initialized spdlog yet here, so it will be better so write in std::cout here

logger/clickhouse.cc
```diff
@@ -17,35 +22,94 @@
...
+{
+	block_ = ch::Block();
+
+	auto col = std::make_shared<ch::ColumnDateTime64>(3);
+	block_.AppendColumn("timestamp", col);
+
+	for (int i = TFW_MMAP_LOG_ADDR; i < TFW_MMAP_LOG_MAX; ++i) {
+		const TfwField *field = &tfw_fields[i];
+		auto col = tfw_column_factory(field->code);
+		block_.AppendColumn(field->name, col);
```
consuelo2210, line 90: Why do we AppendColumn with timestamp first? Could we do it after specific enum values? It could save us from idx+1 code

logger/mmap_buffer.hh
```diff
@@ -53,22 +56,23 @@
...
+	[[nodiscard]] bool run(std::atomic<bool> &stop_flag) noexcept;
 	unsigned int get_cpu_id() const noexcept;
 
 private:
 	TfwMmapBuffer	*buf_;
 	size_t		size_;
-	bool		is_running_;
-	Callback	callback_;
+	ProcessEvents	proc_ev_;
+	TfwClickhouse	&db_;
```
consuelo2210, line 74: The TfwClichHouse with a callback looks a bit strange here. Since we already have db_ as a member, we lose the abstraction of why the callback was created. Previously, we were explicitly doing something with the data; now, there are hints that the data is being inserted into the database, but the actual insertion is hidden inside the callback’s black box. It seems we could safely remove proc_ev_ at this point.

consuelo2210 (changes requested):
I love the way you use idle time to flush the database.

I left some comments during the review, but the main issue for me is that we still have a lot of code that needs to be separated. All reading functions (read_int, read_access_event and so on) should be moved outside of tfw_logger.cc.

I believe that this function:
```cpp
bool TfwMmapBufferReader::run(std::atomic<bool> &stop_flag) noexcept
```
should become a free function:
```cpp
bool run(std::atomic<bool> &stop_flag) noexcept
```


and call consume_event and flush from all EventProcessors (if we have several). I propose the following interface:

```cpp
class IEventProcessor {
public:
    void consume_event();
    void make_background_work(); // Will be called during idle, e.g., commit to ClickHouse
};
```

Every EventProcessor should implement the reading functions (read_int, read_str, consume_event) and, I think, hold a direct reference to TfwClickhouse and write to it during consumption. If we later introduce another writer, we can consider unifying them under a common interface, but for now, we don’t need that.

We can add the following methods to TfwClickhouse and get rid of get_block method (because the class can work only around this one block).

```cpp
template <typename ColType, typename ValType>
void write_int(size_t index, ValType val) {
    block_[index]->As<ColType>()->Append(val);
}

void write_string(size_t index, std::string_view sv) {
    block_[index]->As<ch::ColumnString>()->Append(std::string(sv));
}

void write_empty_string(size_t index) {
    block_[index]->As<ch::ColumnString>()->Append(std::string(""));
}
```
I used template, because it is more close to your code, but we can define IndexType for every index.
template <typename IndexType>
void write_int(IndexType index, IndexType::val_type val){
    block_[IndexType::index_constant]->As<IndexType::col_type>()->Append(val);
}

I think we should get rid of the callback. Previously, it created an abstraction and hid all details of processing. But now, since we have db as a member, that abstraction is broken and only adds unnecessary complexity.

I also rewrite read_int, read_str during another comment to separate reading and writing inside consume_* operation.

So, for me, for now, it is more about separating.

logger/error.hh
```diff
@@ -22,15 +27,78 @@
...
 
 #include <boost/system/system_error.hpp>
 #include <boost/system/error_code.hpp>
 
+/*
+ * ------------------------------------------------------------------------
+ *	std::expected API for the data path
+ * ------------------------------------------------------------------------
+ */
+enum class Err : int {
```
consuelo2210, line 47: I proposed another solution here
https://github.com/tempesta-tech/escudo/issues/176. Have a look, please

logger/clickhouse.cc
krizhanovsky: Yeah, probably it does make sense from technical point of view, but I'd be confused to see timestamp somwhere else than in the first column from `select * from access_log`. And probably there is no hurt with `+ 1`, so I close the comment as resolved, but feel free to reopen if you believe this is important.

logger/tfw_logger.cc
krizhanovsky: @consuelo2210 yes, `std::string_view` is definitely better than `const char *&`. But I didn't get about `LineReader` and `BlockWriter` - what's the point? They are new logical entities and int he proposed approach there are more code, which doesn't look easier to read than the functions? What we get from the 2 new structures?

@ttaym at the moment the logic under discussion is in the generic logging daemon file `tfw_logger.cc` and with the xFW event logging it must be moved to a Tempesta FW specific plugin, so yes, it should be moved to some new class and file. I'm not sure about proposed two separate reading and writing classes, but at least one new class is required.

So far I'm adding a TODO comment in the code.

logger/mmap_buffer.hh
krizhanovsky: Yes, agree. Probably, there should be
1. a class aggregating the existing `process_events()`, `read_access_log_event()`, `read_str()` and `read_int()`. The motivation is https://github.com/tempesta-tech/tempesta/issues/2399 - the same ring buffer will be used for other events with another logic.
2. The new class must proxy data from the ring buffers to Clickhouse. Interfaces in TBD, but it should have only one entry point, not 2, like we have now for `db_` and the callback

logger/error.hh
krizhanovsky: OK, I commented https://github.com/tempesta-tech/escudo/issues/176 : let's leave this for that task. The task in the next milestone and there is no reason to do it in this PR.

consuelo2210 (approved):

logger/tfw_logger.cc
consuelo2210: I'm not sure I was clear earlier. I realized that while we want to separate reading and writing responsibilities, they're currently mixed together in these functions. For example, if we add more events that require integer reading, we’ll end up duplicating code.

In my final review comment, I suggested moving the reader-related functions into the EventProducer class and the writer-related functions into the CH class. I believe this will improve code readability and maintainability.

logger/tfw_logger.cc
```diff
@@ -114,6 +114,24 @@ dbg_hexdump([[maybe_unused]] std::span<const char> data)
...
+ * specific plugin, most likely to a new class(es)
+ * There will be similar functions (class API) for the security events logging.
+ *
+ * The classes should inherit the same interface, e.g.
+ *
+ *	class IEventProcessor {
+ *	public:
+ *		Error<bool> consume_event();
+ *		void make_background_work();
+ *		[[nodiscard]] bool flush(bool force = false) noexcept;
```
consuelo2210, line 130: void make_background_work();
 or 
[[nodiscard]] bool flush(bool force = false) noexcept;
 
 not both, right?

logger/mmap_buffer.cc
```diff
@@ -86,6 +86,12 @@ TfwMmapBufferReader::read() noexcept
 	if (!res)
 		return error(Err::DB_SRV_FATAL);
 
+	// TODO #2399, #182 (escudo xFW): this should be replaced with
+	// IEventProcessor->flush(), but ideally if we can move it to the
+	// free run() function in tfw_logger. Basically, we flush() events
+	// after a ring buffer read, so probably it's just fine to firstly
+	// call all event processors to read events from the associated RBs
```
consuelo2210, line 93: I love this. They really can collect ch::Block inside during reading and have a method

::flush(Clickhouse &db)
{
//std::unique_ptr<Block> block_;
db->commit(block_.release());
}

logger/tfw_logger.cc
krizhanovsky: I think we need the both: one when the main loop is idle and another one for when we read something and mush flush the data

PR #2496, opened 2025-08-19
Fix deadlock in `tfw_http_conn_cli_drop`.

EvgeniiMekhanik:
In `tfw_http_conn_cli_drop` we release all requests in the `seq_queue` under the `cli_conn->seq_qlock`. In this case we can't free `req->resp` under the `cli_conn->seq_lock` if `resp->conn` is not zero and response keeps the last reference to the connection. If response will be freed here, server connection will be released here and all requests from `fwd_list` will be freed. Such requests are freed under the `cli_conn->seq_qlock`, where `cli_conn` is an appropriate client connection, which can be the same as current `cli_conn`. (If it is so deadlock occured). To fix this problem we add all such responses to another list and free it after `cli_conn->seq_lock` will be released.

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -951,6 +951,25 @@ tfw_http_prep_304(TfwHttpReq *req, struct sk_buff **skb_head, TfwMsgIter *it)
 	return 0;
 }
 
+static inline void
+tfw_http_conn_msg_unlink_conn(TfwHttpMsg *hm)
+{
+	/*
+	 * Unlink the connection while there is at least one
+	 * reference. Use atomic exchange to avoid races with
+	 * new messages arriving on the connection.
+	 *
+	 * NOTE: currently this unlink operation is not needed
+	 * for HTTP/2 mode; it is left here as is since it does
+	 * nothing in HTTP/2 mode, because default general stream
+	 * @conn->stream is not used and @conn->stream.msg must
+	 * be always NULL during HTTP/2 processing.
+	 */
+	__cmpxchg((unsigned long *)&hm->conn->stream.msg,
+		  (unsigned long)hm, 0UL, sizeof(long));
+	tfw_connection_put(hm->conn);
+}
+
 /*
  * Free an HTTP message.
  * Also, free the connection instance if there's no more references.
@@ -980,21 +999,7 @@ tfw_http_conn_msg_free(TfwHttpMsg *hm)
 		 * the request.
 		 */
 		WARN_ON_ONCE((TFW_CONN_TYPE(hm->conn) & Conn_Clnt) && hm->pair);
-
-		/*
-		 * Unlink the connection while there is at least one
-		 * reference. Use atomic exchange to avoid races with
-		 * new messages arriving on the connection.
-		 *
-		 * NOTE: currently this unlink operation is not needed
-		 * for HTTP/2 mode; it is left here as is since it does
-		 * nothing in HTTP/2 mode, because default general stream
-		 * @conn->stream is not used and @conn->stream.msg must
-		 * be always NULL during HTTP/2 processing.
-		 */
-		__cmpxchg((unsigned long *)&hm->conn->stream.msg,
-			  (unsigned long)hm, 0UL, sizeof(long));
-		tfw_connection_put(hm->conn);
+		tfw_http_conn_msg_unlink_conn(hm);
 	}
 
 	tfw_http_msg_free(hm);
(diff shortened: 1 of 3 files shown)


fw/http.c
```diff
@@ -3014,11 +3021,40 @@ tfw_http_conn_cli_drop(TfwCliConn *cli_conn)
...
+			 * this response under the `cli_conn->seq_qlock`.
+			 * If response will be freed here, server connection
+			 * will be released here and all requests from
+			 * `fwd_list` will be freed. Such requests are freed
+			 * under the `cli_conn->seq_qlock`, where `cli_conn`
+			 * is a appropriate client connection, which can be
+			 * the same as current `cli_conn`.
+			 */
+			if (!resp->conn
+			    || !tfw_connection_last_ref(resp->conn)) {
```
const-t: Seems this check is not safe. `conn->refcnt` might be changed after `tfw_connection_last_ref()`. For instance:
1. `tfw_connection_last_ref()` returns `false` and right after it `conn->refcnt` decrements on another cpu.
2. We do `tfw_http_resp_pair_free()` and again getting deadlock.

const-t (changes requested):

fw/http.c
```diff
@@ -3014,11 +3021,40 @@ tfw_http_conn_cli_drop(TfwCliConn *cli_conn)
 		smp_mb__before_atomic();
 		set_bit(TFW_HTTP_B_REQ_DROP, req->flags);
 		if (unused) {
-			tfw_http_resp_pair_free(req);
+			resp = req->resp;
```
const-t, line 3025: Maybe we should move logic related to queue cleaning on shutdown from `tfw_http_conn_release()` to earlier stage where we still have one additional reference counter? It will give us ability to free responses/requests from `fwd_list` during closing connection while connection itself still alive.

What I mean:
Calling disconnect for connection -> do `fwd_list` cleaning -> put the last reference to connection -> calling `tfw_http_conn_release()` if it needs. However, looks `tfw_http_conn_release()` can be dropped because it will become unnecessary.

fw/http.c
EvgeniiMekhanik: Yes fixed

fw/http.c
EvgeniiMekhanik: I am not sure about this suggestion. In this case we drop requests with responses on one cpu, where we clean fwd_list and sending response to the client connection on other cpu. I can't reproduce it. (All works,  but I am not sure about race). In current solution we free responses only if  READY flag is set and free connection only after freeing all responses. I also should tell that for requests we also free connection after freeing all requests

EvgeniiMekhanik: And free requests only if no response or READY flag is set

const-t (approved):

fw/http.c
```diff
@@ -3013,11 +3020,41 @@ tfw_http_conn_cli_drop(TfwCliConn *cli_conn)
...
+			 * this response under the `cli_conn->seq_qlock`.
+			 * If response will be freed here, server connection
+			 * will be released here and all requests from
+			 * `fwd_list` will be freed. Such requests are freed
+			 * under the `cli_conn->seq_qlock`, where `cli_conn`
+			 * is a appropriate client connection, which can be
+			 * the same as current `cli_conn`.
+			 */
+			if (!resp->conn
+			    || !__tfw_connection_get_if_last_ref(resp->conn)) {
```
```suggestion
			    || !__tfw_connection_get_if_last_ref(resp->conn))
		{
```

Where do we put `resp->conn`? Here we increment the reference counter, but there is no place where we decrement it (`tfw_http_resp_pair_free()` isn't counted because we had it before, but now we have one more refcount). Probably we have a leak now.

fw/http.c
```diff
@@ -3013,11 +3020,41 @@ tfw_http_conn_cli_drop(TfwCliConn *cli_conn)
...
+				tfw_http_msg_free((TfwHttpMsg *)req);
+			 }
+
 			TFW_INC_STAT_BH(serv.msgs_otherr);
 		}
 	}
 	spin_unlock(&cli_conn->seq_qlock);
+
+	list_for_each_entry_safe(resp, tmp_resp, &resp_del_queue,
+				 msg.seq_list) {
```
```suggestion
				 msg.seq_list)
		{
```

fw/http.c
EvgeniiMekhanik: `__tfw_connection_get_if_last_ref` returns `false` if it doesn't increment reference counter. If it returns true we decrement reference counter later. in the loop which was added in this path.

fw/http.c
```diff
@@ -3015,11 +3022,50 @@ tfw_http_conn_cli_drop(TfwCliConn *cli_conn)
...
 	spin_unlock(&cli_conn->seq_qlock);
+
+	/*
+	 * TODO #687 Should be removed during reworking current architecture of
+	 * the locking of `seq_queue` in client connections and `fwd_queue`
+	 * in server connection.
+	 */
+	list_for_each_entry_safe(resp, tmp_resp, &resp_del_queue,
+				 msg.seq_list)
+		{
```
krizhanovsky, line 3064: broken identation

krizhanovsky (approved):
LGTM

PR #2499, opened 2025-08-21
Fix using uninitialized spinlock

EvgeniiMekhanik:
Each `TfwApmHMStats` structure contains spin_lock for `history` field protection. We should initialize this `spin_lock` for each `TfwApmHMStats` structure.


const-t (approved):

krizhanovsky (approved):
LGTM

PR #2501, opened 2025-08-22
Fix spin_lock bug.

EvgeniiMekhanik:
Previosly in linux kernel 5.10.35 `icsk_accept_queue` lock was initialized in `reqsk_queue_alloc` inside `inet_csk_listen_start` call. In current kernel version locks initialization was moved to the `inet_create` inside `inet_init_csk_locks` function call. Since Tempesta FW uses it's own version of `inet_create` (ss_inet_create) we should also move socket locks initialization to this function.


krizhanovsky (approved):
LGTM

PR #2503, opened 2025-08-25
Allow to use page pool on TX path

const-t:
Some drivers allocate fragments using page pool, those SKBs doesn't free as regular SKBs they are must be returned to the page pool. To mark SKBs that contain a fragment allocated from page pool used `pp_recycle` flag, when this fragment moves to another skb, `pp_recycle` flag also must be set to this skb.

### DIFF
--- a/fw/access_log.c
+++ b/fw/access_log.c
@@ -259,7 +259,7 @@ static void
 do_access_log_req_mmap(TfwHttpReq *req, u16 resp_status,
 		       u64 resp_content_length)
 {
-	u64 *dropped = this_cpu_ptr(&mmap_log_dropped);
+	u64 *dropped = this_cpu_ptr(&mmap_log_dropped), dropped_tmp;
 	TfwBinLogEvent *event;
 	unsigned int room_size;
 	TfwStr referer, ua;
@@ -351,6 +351,8 @@ do_access_log_req_mmap(TfwHttpReq *req, u16 resp_status,
 
 	if (*dropped) {
 		WRITE_FIELD(*dropped);
+		/* Save dropped to restore if commit failed. */
+		dropped_tmp = *dropped;
 		*dropped = 0;
 	} else {
 		TFW_MMAP_LOG_FIELD_RESET(event, TFW_MMAP_LOG_DROPPED);
@@ -358,6 +360,7 @@ do_access_log_req_mmap(TfwHttpReq *req, u16 resp_status,
 
 	if (tfw_mmap_buffer_commit(mmap_buffer, p - data) != 0) {
 		T_DBG("Incorrect data size at commit: %ld", p - data);
+		*dropped = dropped_tmp;
 		goto drop;
 	}
 


fw/http.c
```diff
@@ -1798,10 +1799,16 @@ __tfw_http_free_cleanup(TfwHttpMsgCleanup *cleanup)
 	struct sk_buff *skb;
 
 	while ((skb = ss_skb_dequeue(&cleanup->skb_head)))
-		__kfree_skb(skb);
+		kfree_skb(skb);
```
krizhanovsky: IIRC we use `__kfree_skb()` since we own it and we know that there is no other users (references) - what has changed that we do need to track references now?

fw/http_msg.c
```diff
@@ -1346,12 +1346,10 @@ __tfw_http_msg_move_frags(struct sk_buff *skb, int frag_idx,
 			  TfwHttpMsgCleanup *cleanup)
 {
 	int i, len;
-	struct page *page;
 	struct skb_shared_info *si = skb_shinfo(skb);
 
 	for (i = 0, len = 0; i < frag_idx; i++) {
-		page = skb_frag_page(&si->frags[i]);
-		cleanup->pages[i] = compound_head(page);
+		cleanup->pages[i] = skb_frag_netmem(&si->frags[i]);
```
krizhanovsky, line 1383: What if the skb or frag weren't page pool allocated? Or does netmem handles all the cases now?

const-t: We using `skb_page_unref()` to handle this, it choose the right strategy.

fw/http.c
const-t: Good catch, it was not intended

fw/t/unit/test_http_msg.c
```diff
@@ -84,39 +99,33 @@ TEST(http_msg, hdr_in_array)
...
-		return hmresp;
+		return true;
 
 	page = alloc_page(GFP_ATOMIC);
-	if (!page) {
-		kfree_skb(skb);
-		return NULL;
-	}
+	if (!page)
+		return false;
```
EvgeniiMekhanik, line 128: It seems that you forget to free previously allocated `skb` here

fw/t/unit/test_http_msg.c
```diff
@@ -129,7 +138,7 @@ __test_resp_alloc(TfwStr *head_data, TfwStr *paged_data,
 
 	put_page(page);
 
-	return hmresp;
+	return resp;
```
EvgeniiMekhanik: resp is not bool

fw/t/unit/test_http_msg.c
```diff
@@ -211,8 +212,6 @@ TEST(http_msg, cutoff_linear_headers_and_linear_body)
 
 		EXPECT_ZERO(memcmp(addr, expected_frags[i].data, fragsz));
 	}
-
-	tfw_http_msg_free((TfwHttpMsg *)resp);
```
EvgeniiMekhanik, line 215: Where do you free skb allocated in __test_resp_data_alloc?

const-t: test_resp_free() will free skb that linked with response.

EvgeniiMekhanik (approved):

fw/t/unit/test_http_msg.c
const-t: Will be freed during response freeing in test_resp_free()

fw/t/unit/test_http_msg.c
const-t: Fixed

Issue #2495, opened 2025-08-15
Support --generate option for tfw_logger


DenisShkiria:
# Motivation

Currently, the `--generate` option is documented in the tfw_logger [Wiki](https://github.com/tempesta-tech/tempesta/wiki/Access-Log-Analytics), but it is not supported by the utility. We could remove this option from the wiki, but I think it would be useful to support it because it would help users prepare a configuration file for the utility. Currently, the only way to do so is to rely on the Wiki page documentation and copy and paste the Wiki example into the file before modifying it for one's own needs.

# Scope

Add the new "--generate" or "-g" option to the tfw_logger command line options. This option generates a configuration file with default values. For example:
```
# Generate default configuration
./tfw_logger --generate --config /etc/tempesta/tfw_logger.json
```

# Testing

The functionality should be implemented in a way that allows it to be covered by a unit test.

# Documentation

After making changes, check that the tfw_logger [wiki](https://github.com/tempesta-tech/tempesta/wiki/Access-Log-Analytics) page is valid.

krizhanovsky:
IIRC we got rid of the option because we typically distribute our soft with default configs, like [tempesta_fw.conf](https://github.com/tempesta-tech/tempesta/blob/master/etc/tempesta_fw.conf) . But fixing wiki is definitely worth it.

PR #2504, opened 2025-08-27
Fix BUG if Tempesta FW received skb with small headroom

const-t:
- Tempesta FW reuses received skbs to prevent extra copying. If such skbs have small headroom and later will be pushed to the socket write queue BUG_ON will be occured during pushing TCP and IP headers to such skbs. For all skbs allocated by Tempesta FW we allocate headroom which is equal to MAX_TCP_HEADER. Now for all received skbs we check headroom and if it is less then MAX_TCP_HEADER reallocate skb with MAX_TCP_HEADER headroom.
- Also fix potencial memleak if `ss_skb_chop_head_tail` failed.

Author: @EvgeniiMekhanik
Review: @const-t

### DIFF
--- a/fw/sock.c
+++ b/fw/sock.c
@@ -884,6 +884,7 @@ do {									\
 		if (unlikely(offset > 0 &&
 			     ss_skb_chop_head_tail(NULL, skb, offset, 0) != 0))
 		{
+			__kfree_skb(skb);
 			r = SS_BAD;
 			goto out;
 		}
@@ -902,6 +903,22 @@ do {									\
 		if (SS_CONN_TYPE(sk) & Conn_Stop) {
 			__kfree_skb(skb);
 			continue;
+		}
+
+		if (unlikely(skb_headroom(skb) < MAX_TCP_HEADER)) {
+			struct sk_buff *twin_skb =
+				__pskb_copy_fclone(skb, MAX_TCP_HEADER,
+						   GFP_ATOMIC, true);
+			if (!twin_skb) {
+				T_WARN("Unable to copy SKB with small "
+				       "headroom.\n");
+				__kfree_skb(skb);
+				r = SS_BAD;
+				goto out;
+			}
+
+			__kfree_skb(skb);
+			skb = twin_skb;
 		}
 
 		r = SS_CALL(connection_recv, conn, skb);


fw/sock.c
```diff
@@ -904,6 +905,22 @@ do {									\
 			continue;
 		}
 
+		if (unlikely(skb_headroom(skb) < MAX_TCP_HEADER)) {
+			struct sk_buff *twin_skb =
+				__pskb_copy_fclone(skb, MAX_TCP_HEADER,
```
const-t: We can re-allocate only headroom, not reallocating whole skb using `pskb_expand_head()`. 

And maybe we can do it before sending? Some skbs might be freed during request/response processing, re-alloc of such skbs is unnecessary. But need to be careful with `tail_lock` in this case, we must be sure that `skb_free_head()` in `pskb_expand_head()` will not free old head that was shifted to fragments.

fw/sock.c
```diff
@@ -904,6 +905,22 @@ do {									\
...
+						   GFP_ATOMIC, true);
+			if (!twin_skb) {
+				T_WARN("Unable to copy SKB with small "
+				       "headroom.\n");
+				__kfree_skb(skb);
+				r = SS_BAD;
+				goto out;
+			}
```
krizhanovsky: Why don't we fix it in `MAX_HEADER`, like we do this for TLS headers, i.e. just increase the allocation overhead for all allocated `skb`s?

This is `unlikely` specifier, but when this actually happen? `ss_tcp_process_skb()` processes all the ingress packets, so it is super hot and we should no introduce skb reallocations here. Moreover, in most of the cases we do no retransmit skbs and are just going to introduce the overhead.

krizhanovsky (commented):
Which issue does it fix? The branch name refernces #2495, but it is quite a different issue.

fw/sock.c
```diff
@@ -435,8 +441,28 @@ ss_skb_tcp_entail_list(struct sock *sk, struct sk_buff **skb_head)
...
+
+restore_sk_write_queue:
+	to_destroy = tail ? tail->next : tcp_send_head(sk);
+	if (to_destroy) {
+		tcp_for_write_queue_from_safe(to_destroy, next, sk) {
+			tcp_unlink_write_queue(to_destroy, sk);
+			tcp_wmem_free_skb(sk, to_destroy);
+		}
+	}
+	ss_skb_setup_opaque_data(*skb_head, opaque_data, destructor);
```
const-t, line 464: Looks like passing NULL to `ss_skb_setup_opaque_data()` in case when  `ss_skb_realloc_headroom()` failed on the last skb and `*skb_head` becomes NULL.

EvgeniiMekhanik: No we insert skb to the skb_head back if ` ss_skb_realloc_headroom` fails

PR #2512, opened 2025-09-15
Invent tempesta user-space library for at least error and pidfile

ttaym:
That shared library should be using from at least tfw looger and tempesta escudo. So i prepare for it with installing in /opt/tempesta_escudo well-known path.

### DIFF
--- a/libtus/Makefile
+++ b/libtus/Makefile
@@ -1,0 +1,45 @@
+#		TempestaUser-space Library
+#
+# Copyright (C) 2025 Tempesta Technologies.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License,
+# or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 59
+# Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+
+PREFIX ?= /opt/tempesta_escudo
+LIBDIR ?= ${PREFIX}/lib
+
+CXX		= g++
+# Use gnu++ instead of c++ for standard C header compatibility.
+CFLAGS		= -O2 -ggdb -Wall -std=gnu++23
+LDFLAGS		= -shared -fPIC
+
+OBJECTS	= $(patsubst %.cc, %.o, $(wildcard *.cc))
+
+all : libtus.so
+
+libtus.so: $(OBJECTS)
+	$(CXX) $(CFLAGS) $(LDFLAGS) -o $@ $^
+
+%.o: %.cc
+	$(CXX) $(CFLAGS) $(LDFLAGS) -c $< -o $@
+
+install:
+	@install -d -m 755 ${LIBDIR}
+	install -m 755 libtus.so $(LIBDIR)
+
+clean: FORCE
+	rm -f *.o* *~ libtus.so
+
+FORCE:
+
--- a/libtus/error.hh
+++ b/libtus/error.hh
@@ -33,11 +33,13 @@
 #include <sstream>
 #include <string>
 #include <stdexcept>
-#include <system_error>
 #include <utility>
 
 #include <boost/system/system_error.hpp>
 #include <boost/system/error_code.hpp>
+#include <boost/system/linux_error.hpp>
+
+namespace tus {
 
 /*
  * ------------------------------------------------------------------------
@@ -48,12 +50,6 @@ enum class Err : int {
 	// Clickhouse error.
 	DB_SRV_FATAL,
 };
-
-namespace std {
-	template<>
-	struct is_error_code_enum<Err> : true_type
-	{};
-}
 
 class ErrorCategory : public std::error_category {
 public:
@@ -67,7 +63,7 @@ public:
 	message(int e) const override
 	{
 		switch (static_cast<Err>(e)) {
-			case Err::DB_SRV_FATAL:
+		case Err::DB_SRV_FATAL:
 			return "Database unrecoverable server error";
 		default:
 			return "Unknown error";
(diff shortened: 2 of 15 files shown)


libtus/Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+PREFIX ?= /opt/tempesta_escudo
+LIBDIR ?= ${PREFIX}/lib
+
+CXX		= g++
+# Use gnu++ instead of c++ for standard C header compatibility.
+CFLAGS		= -O2 -ggdb -Wall -std=gnu++23
+LDFLAGS		= -shared -fPIC
+
+OBJECTS	= pidfile.o error.o
```
krizhanovsky: We can generate objects with `OBJECTS := $(patsubst %.cc, %.o, $(wildcard *.cc))` and this will avoid the error that now we don't build an object for error.cc

libtus/error.hh
```diff
@@ -1,12 +1,7 @@
 /**
  *		Tempesta FW
  *
- * Error handling - rules of thumb:
- * 1. prefer std::expected and std::optional on the data plane to reduce
- *    overhead and improve reliability;
- * 2. it's OK to use exceptions on control path for easier error management.
- *
- * Copyright (C) 2024-2025 Tempesta Technologies, Inc.
+ * Copyright (C) 2024 Tempesta Technologies, Inc.
```
krizhanovsky: Seems broken merge

libtus/error.hh
```diff
@@ -105,6 +100,12 @@ public:
 		: std::runtime_error(format_syserr(s))
 	{}
 
+	void
+	print(char *buf, size_t len)
+	{
+		snprintf(buf, len, "%s", what());
+	}
```
krizhanovsky: Do we need this method in Escudo? If yes, then this should be either std::string or std::span, but I don't see any reason to use a raw pointer for error messaging

logger/Makefile
```diff
@@ -16,6 +16,9 @@
 # this program; if not, write to the Free Software Foundation, Inc., 59
```
krizhanovsky, line 16: It seems `libtus/t/tfw_libtus_tests` was commited by a mistake

libtus/error.hh
```diff
@@ -142,15 +144,31 @@ private:
...
+ */
+
+template<typename... Args>
+inline void verify(bool cond, const char* msg, Args&&... args)
+{
+	if (!cond) {
+		throw std::runtime_error(fmt::format(fmt::runtime(msg),
+			std::forward<Args>(args)...));
+	}
+}
```
krizhanovsky, line 162: IIUC this function is considered harmful in Escudo and will be deleted, so let's not to propagate it to Tempesta FW

logger/Makefile
```diff
@@ -16,6 +16,9 @@
...
 # Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 
+PREFIX ?= /opt/tempesta_escudo
```
krizhanovsky: There should be no Escudo stuff in Tempesta

logger/Makefile
```diff
@@ -27,13 +30,15 @@ LDFLAGS += -lcityhash -L $(CLICKHOUSE_BUILD_DIR)/contrib/cityhash/cityhash
...
 LDFLAGS += -labsl_int128 -L $(CLICKHOUSE_BUILD_DIR)/contrib/absl/absl
+LDFLAGS += -ltus -L ../libtus -Wl,-rpath,'$$ORIGIN/../lib:$$ORIGIN/../libtus'
 LDFLAGS += $(shell pkgconf -libs spdlog)
 PAGE_SIZE := $(shell getconf PAGE_SIZE)
 CXXFLAGS := -Wall -Wextra -std=c++23 -march=native -mtune=native \
 	    -DPAGE_SIZE=$(PAGE_SIZE)
 CXXFLAGS += $(shell pkgconf -cflags spdlog)
 CXXFLAGS += -I $(CLICKHOUSE_DIR)
 CXXFLAGS += -I $(CLICKHOUSE_DIR)/contrib/absl
+CXXFLAGS += -I ../libtus
```
krizhanovsky: I think we should include all the headers explicitly. For now we still have
```C++
#include "pidfile.hh"
```
i.e. it's included just as other _local_ files, instead we should include it as `libtus/pidfile.hh`

libtus/error.hh
```diff
@@ -33,11 +28,11 @@
...
 #include <stdexcept>
-#include <system_error>
 #include <utility>
 
 #include <boost/system/system_error.hpp>
 #include <boost/system/error_code.hpp>
+#include <boost/system/linux_error.hpp>
 
 /*
  * ------------------------------------------------------------------------
```
krizhanovsky, line 45: Please introduce a namespace for everything in the library, like `tus`

libtus/error.hh
ttaym: Yes, error.hh file diversed between tempesta fw and escudo, so there can be some minor discrepancies. I'll fix it.

libtus/Makefile
ttaym: Sure

libtus/error.hh
ttaym: That method was added not so long ago in escudo. I'll see what can we do in escudo about it.

libtus/error.hh
ttaym: Ok. Then lets leave it there.

logger/Makefile
ttaym: My bad. I'll fix it.

logger/Makefile
ttaym: Oh, i see. We should of course install only libtus there, not logger.

logger/Makefile
ttaym: Ok

libtus/error.hh
```diff
@@ -33,11 +33,13 @@
...
 #include <string>
 #include <stdexcept>
-#include <system_error>
 #include <utility>
 
 #include <boost/system/system_error.hpp>
 #include <boost/system/error_code.hpp>
+#include <boost/system/linux_error.hpp>
+
+namespace tus {
```
krizhanovsky, line 42: Please also place `pidfile` API into the namespace.

logger/tfw_logger.cc
```diff
@@ -55,6 +55,8 @@ constexpr std::chrono::seconds wait_for_dev{1};
 static std::atomic<bool> stop_flag{false};
 static TfwLoggerConfig config;
 
+using namespace tus;
```
krizhanovsky: Usually, we don't do this and use `tus::` as you did in `mmap_buffer.hh`, please follow this in all the source code files

ttaym:
@const-t I have resolved all comments from Alexander. As you have not seen this PR before, i have squashed all commits into one.

const-t (approved):
LGTM

PR #2513, opened 2025-09-16
 incapsulate block access into clickhouse.cc

ttaym:


logger/Makefile
```diff
@@ -77,4 +82,9 @@ tfw_logger_sanitize:
 	$(MAKE) clickhouse
 	$(MAKE) sanitize
 
+
```
krizhanovsky: extra empty line

logger/tfw_logger.cc
```diff
@@ -134,33 +134,29 @@ dbg_hexdump([[maybe_unused]] std::span<const char> data)
  */
 template <typename ColType, typename ValType>
 void
-read_int(TfwBinLogFields ind, ch::Block &block, const auto *event, const char *&p,
-	 size_t &size)
+read_int(TfwBinLogFields ind, TfwClickhouse &db,
+	 const auto *event, const char *&p, size_t &size)
```
krizhanovsky: Please address todo from the above about replacing `*&p` with `std::string_view`

logger/tfw_logger.cc
```diff
@@ -134,33 +134,29 @@ dbg_hexdump([[maybe_unused]] std::span<const char> data)
...
-
 	if (TFW_MMAP_LOG_FIELD_IS_SET(event, ind)) {
 		const size_t len = tfw_mmap_log_field_len(ind);
 
 		if (len > size) [[unlikely]]
 			throw Except("Incorrect integer eventent length");
 
 		const ValType *val = reinterpret_cast<const ValType *>(p);
-		block[bi]->As<ColType>()->Append(*val);
+		db.append_int<ColType, ValType>(ind, *val);
```
krizhanovsky: Now we have problem of ignored exceptions - we need either to throw exceptions or return a value

logger/Makefile
ttaym: this PR had been produced over still unmerged commit about libtus, so need to review only last commit. Sorry about not clearly uttered that in description.

logger/tfw_logger.cc
ttaym: `in read_access_log_event()` we use std::span, let's use it everywhere here (it better corresponds to non-string binary data concept). I'll do it with std::span and you could see if it is acceptable for you.

logger/tfw_logger.cc
ttaym: Gosh. You are right. I think i should ~~re-throw exceptions after logging~~ just propagate exceptions without catching, as it will be caught in `process_event()`.

logger/clickhouse.cc
```diff
@@ -195,3 +217,19 @@ tfw_column_factory(ch::Type::Code code)
 		throw std::runtime_error("Column factory: incorrect code");
 	}
 }
+
+template void
+TfwClickhouse::append_int<ch::ColumnUInt8, unsigned char>(TfwBinLogFields,
```
consuelo2210: I think, if we will use typetraits for every TfwBinLogFields  type, like:

```cpp
// Primary template — only common fields, like index.
template<TfwBinLogFields FieldType>
struct TfwBinLogTypeTraits
{
    // compile time index
    static constexpr size_t index = static_cast<size_t>(FieldType) + 1;

    // there are no types by default, if anyone forget to add - will be an error (TODO: need to check it)
    using ColType = void;
    using ValType = void;
    //constexpr int len = tfw_mmap_log_field_len()?
};

// Specializations
template<>
struct TfwBinLogTypeTraits<TFW_MMAP_LOG_METHOD>
    : TfwBinLogTypeTraits<TFW_MMAP_LOG_METHOD> // we need index
{
    using ColType = ch::ColumnUInt8;
    using ValType = unsigned char;

};

template<>
struct TfwBinLogTypeTraits<TFW_MMAP_LOG_VHOST>
    : TfwBinLogTypeTraits<TFW_MMAP_LOG_VHOST>
{
    using ColType = ch::ColumnString;
    using ValType = std::string;
};

....

then we will be able to write code like:

template <TfwBinLogFields FieldType>
void TfwClickhouse::append(
    const typename TfwBinLogTypeTraits<FieldType>::ValType& value)
{
    using Traits   = TfwBinLogTypeTraits<FieldType>;
    using ColType  = typename Traits::ColType;
    using ValType  = typename Traits::ValType;

    auto* col = get_column<ColType>(Traits::index); 
    col->Append(value);
}
```
I suppose to add changes above, but also write in TODO the idea below:

In the code we will have
```
void process_field<TFW_MMAP_LOG_ADDR>(db, ev, data);

instead of
read_int<ch::ColumnIPv6, in6_addr>(TFW_MMAP_LOG_ADDR, db, ev, data);

we will have

template <TfwBinLogFields FieldType>
void process_field(TfwClickhouse &db, const auto *event, std::span<const char> &data)
{
    auto value = read_value<FieldType>(event, data);
    db.append<FieldType>(value);
}

size_t
read_access_log_event(TfwClickhouse &db, std::span<const char> data)
{
...
	process_field<TFW_MMAP_LOG_ADDR>(db, ev, data);
	process_field<TFW_MMAP_LOG_METHOD>(db, ev, data);
	process_field<TFW_MMAP_LOG_VERSION>(db, ev, data);
...
}


And just to be more concrete, 2 read_value function
// For integer-like values
template <TfwBinLogFields FieldType>
requires std::is_arithmetic_v<typename TfwBinLogTypeTraits<FieldType>::ValType>
typename TfwBinLogTypeTraits<FieldType>::ValType
read_value(const auto *event, std::span<const char> &data)
{
    using Traits  = TfwBinLogTypeTraits<FieldType>;
    using ValType = typename Traits::ValType;

    ValType value{};
    if (TFW_MMAP_LOG_FIELD_IS_SET(event, FieldType)) {
        const size_t len = tfw_mmap_log_field_len(FieldType);
        if (data.size() < len) [[unlikely]]
            throw tus::Except("Incorrect integer event length");

        value = *reinterpret_cast<const ValType*>(data.data());
        data = data.subspan(len);
    }
    return value;
}

// For strings
template <TfwBinLogFields FieldType>
requires std::same_as<typename TfwBinLogTypeTraits<FieldType>::ValType, std::string>
typename TfwBinLogTypeTraits<FieldType>::ValType
read_value(const auto *event, std::span<const char> &data)
{
    using Traits  = TfwBinLogTypeTraits<FieldType>;
    using ValType = typename Traits::ValType;

    ValType value{};
    if (TFW_MMAP_LOG_FIELD_IS_SET(event, FieldType)) {
        constexpr int len_size = sizeof(uint16_t);
        if (data.size() < len_size) [[unlikely]]
            throw tus::Except("Too short string event");

        const size_t len = *reinterpret_cast<const uint16_t*>(data.data());
        if (data.size() < len_size + len) [[unlikely]]
            throw tus::Except("Incorrect string event length");

        value.assign(data.data() + len_size, len);
        data = data.subspan(len_size + len);
    }
    return value;
}

```

I have not verified the idea above, it just an idea and need to be checked with accuracy.  I think that traits should be here. It will definitely increase a code size. But

1) we will calculate index on compile time, so if these functions is on top of xperf, even +1 can be crucial. Len also can be calculated on compile time for concrete index. So we have some speed benefits.
2) we will reduce the quantity of template parameters, because they are actually strongly connected - it will hide CH internals from read_access_log_event function, until actual writing in CH inside append. We don't need to control type of the field too and write it anywhere in code, it is now a part of index description and we can use just ValType and ColType.
3) we don't need to use c-code in c++ files, so we can create separate helpers to work with log, in c++ way. Upon c-helpers(don't use tfw_mmap_log_field_len in c++ code - use compile time calculation and template direct insertion in code).


I wrote a lot, but suggest to start with traits introducing and with a TODO comment  to moving forward (maybe it will be rewriten in other way).

consuelo2210: @const-t @krizhanovsky I love the direction in which we moved in this PR (only second commit to review in this PR in draft). Alexey asked me to review it, but I don't think that I have right to insist on the implementation. Can you observe the idea and decide do you want to move to type traits or not?

consuelo2210: I am also not so sure about second part of proposal. It just a variant, where we can go. But I don't know, do we really need it or no. So that's why I propose just to add the second part as an idea in comments. Maybe it will just help with thoughts about further direction.

logger/clickhouse.hh
```diff
@@ -65,17 +66,30 @@ public:
 
 	~TfwClickhouse();
 
-	ch::Block &get_block() noexcept;
+	void append_timestamp(uint64_t timestamp) noexcept;
+
+	template <typename ColType, typename ValType>
+	void append_int(TfwBinLogFields field, ValType value) noexcept;
```
consuelo2210: noexcept now can throw? I think that now we have to remove noexcept from description to avoid std::terminate

ttaym: Sorry

logger/clickhouse.cc
consuelo2210: If we introduce only traits, than read_int and read_str will be renamed to read_field with 2 specialization read_field: for int and for string (requires std::same_as<typename TfwBinLogTypeTraits<FieldType>::ValType, std::string> and requires std::is_arithmetic_v<typename TfwBinLogTypeTraits<FieldType>::ValType>).

And code will be 
read_field<TFW_MMAP_LOG_ADDR>(db, ev, data); 
...

read_field<TFW_MMAP_LOG_URI>(db, ev, data); 
...

logger/clickhouse.cc
```diff
@@ -94,6 +95,33 @@ TfwClickhouse::make_block()
...
+{
+	const size_t col_index = field_to_column_index(field);
+	block_[col_index]->As<ch::ColumnString>()->Append(std::string(value));
+}
+
+void
+TfwClickhouse::append_empty_string(TfwBinLogFields field)
+{
+	append_string(field, std::string_view{});
+}
```
krizhanovsky: can't we just replace this with `TfwClickhouse::append_string(field, std::string_view{})` or make a default value for `field` in `append_string()`?

logger/clickhouse.hh
```diff
@@ -65,17 +66,30 @@ public:
...
+	void append_int(TfwBinLogFields field, ValType value);
+
+	void append_string(TfwBinLogFields field, std::string_view value);
+	void append_empty_string(TfwBinLogFields field);
+
 	[[nodiscard]] bool commit(bool force = false) noexcept;
 	bool handle_block_error() noexcept;
 
+private:
+	constexpr size_t
```
krizhanovsky: Please leave comment
```C++
// We store timestamp at index 0
```

logger/tfw_logger.cc
```diff
@@ -134,112 +134,75 @@ dbg_hexdump([[maybe_unused]] std::span<const char> data)
  */
```
krizhanovsky: This comment partially fixes the issues, so let's adjust it as well

logger/tfw_logger.cc
```diff
@@ -134,112 +134,75 @@ dbg_hexdump([[maybe_unused]] std::span<const char> data)
...
 		const size_t len = tfw_mmap_log_field_len(ind);
 
-		if (len > size) [[unlikely]]
+		if (data.size() < len) [[unlikely]]
 			throw tus::Except("Incorrect integer eventent length");
 
-		const ValType *val = reinterpret_cast<const ValType *>(p);
-		block[bi]->As<ColType>()->Append(*val);
+		const ValType *val =
+			reinterpret_cast<const ValType *>(data.size());
```
krizhanovsky:
```suggestion
			reinterpret_cast<const ValType *>(data.data());
```

Please run Tempesta FW with the logger and make sure that you see _adequate_ values in the database. Ideally, if you run Tempesta FW in fron of our [website container](https://github.com/tempesta-tech/tempesta-tech.com/tree/master/container#build-or-update-wordpress-site) (you jsut need to run `./container/lxc/create.py --type=stage` to deploy the container), but a simple file serviced by Nginx is also fine. Run HTTP request to produce 200, 404 and 400 (bad request) responses.

logger/tfw_logger.cc
```diff
@@ -134,112 +134,75 @@ dbg_hexdump([[maybe_unused]] std::span<const char> data)
...
-		if (len > size) [[unlikely]]
+		const size_t len =
+			*reinterpret_cast<const uint16_t *>(data.data());
+		if (data.size() < len_size + len) [[unlikely]]
 			throw tus::Except("Incorrect string event length");
 
-		block[bi]->As<ch::ColumnString>()->Append(std::string(p, len));
-		p += len;
-		size -= len;
+		std::string_view str(data.data() + len_size, len);
```
krizhanovsky:
```suggestion
        data = data.subspan(len_size);
		std::string_view str(data.data(), len);
```
We use `std::span` to not to deal with raw pointers, so let's not do this with `std::span` .

With this fix we'll also need to use `data = data.subspan(len); ` below instead of `data = data.subspan(len_size + len);`.

logger/tfw_logger.cc
```diff
@@ -134,112 +134,75 @@ dbg_hexdump([[maybe_unused]] std::span<const char> data)
...
+	read_int<ch::ColumnUInt32, uint32_t>(TFW_MMAP_LOG_RESP_TIME, db, ev, data);
+
+	read_str(TFW_MMAP_LOG_VHOST, db, ev, data);
+	read_str(TFW_MMAP_LOG_URI, db, ev, data);
+	read_str(TFW_MMAP_LOG_REFERER, db, ev, data);
+	read_str(TFW_MMAP_LOG_USER_AGENT, db, ev, data);
+
+	read_int<ch::ColumnUInt64, uint64_t>(TFW_MMAP_LOG_JA5T, db, ev, data);
+	read_int<ch::ColumnUInt64, uint64_t>(TFW_MMAP_LOG_JA5H, db, ev, data);
+	read_int<ch::ColumnUInt64, uint64_t>(TFW_MMAP_LOG_DROPPED, db, ev, data);
```
krizhanovsky: Once we move all Clickhouse stuff into `TfwClickhouse` we do not need `ch::Column*` types and just can instantiate the types by the C types in clickhouse.cc

logger/clickhouse.cc
krizhanovsky: Traits are an elegant way to generate code from C types to Clickhouse types. While it requires more C++ code, I think binary code will be smaller thanks to small trait teplate, probably mostly eliminated at compile time. However, probably the win isn't so high, so I'd leave this up to @ttaym

logger/tfw_logger.cc
krizhanovsky: Actually, event enum, like `TFW_MMAP_LOG_DROPPED`, is already enough and we can replace both the types with only this enum and we get only `read_field<TFW_MMAP_LOG_ADDR>(db, ev, data)` like it is in https://github.com/tempesta-tech/tempesta/pull/2513#discussion_r2382534666

logger/clickhouse.cc
const-t: @consuelo2210 Idea with type traits looks great for me.

logger/mmap_plugin.cc
```diff
@@ -0,0 +1,166 @@
...
+}
+
+int
+mmap_plugin_init(const ClickHouseConfig *config, void *stop_flag)
+{
+	assert(config);
+
+	plugin_log_info("Mmap plugin initialization");
+
+	global_stop_flag = reinterpret_cast<std::atomic<bool> *>(stop_flag);
```
consuelo2210: Do we have ABI compatibility guarantees between plugins? If so, would using std::stop_token be a more reliable and safer option than std::atomic<bool>?

ttaym: That stop flag is not an opaque pointer anyway because we not produce it inside plugin (in which case there should not be a need to pass it). It is global flag so its type should be known in advance and in fact should be equal to real type of global stop flag.

It it not an std::atomic<bool> just because i decide plugin api to be strictly c compatible. If plugin api would be c++-first then it can be just std::atomic<bool>.

Note that AFAIK it can be just as well be casted to c-compatible _Atomic and used as it if one decide to write c-only plugin. And std::stop_token i suppose cannot be used this way in c-only plugin. (That argument really is not very strong because Alexander wants all processor to be c++ classes anyway)

logger/tfw_logger.cc
```diff
@@ -67,312 +70,166 @@ struct ParsedOptions {
...
 
-		if (mbr.run(stop_flag))
-			break;
-		// ...else, reset the Clickhouse connection.
+	if (mmap_plugin) {
+		auto processor = mmap_plugin->create_processor(ncpu);
+		cpu_id = (static_cast<MmapProcessor *>(
+				processor.get()))->get_cpu_id();
+		if (processor)
+			processors.emplace_back(std::move(processor));
```
ttaym: For revievers. MmapProcessor is coupled with tfw logger. May be better it to not be a plugin, because it is in tfw_logger code tree anyway. But coded as is.

Note cast to MmapProcessor, that is because only mmap processor has sensible exported cpu id, that needed for affinity set. It seems that this is a peculiarity of mmap processor, so it really should _not_ be a part of API. Or may be i cannot get some piece of understanding.

logger/t/test_config.cc
```diff
@@ -78,13 +79,16 @@ TEST_F(ConfigTest, LoadValidConfig)
...
 		"access_log": {
-			"host": "test.host.com",
-			"port": 9001,
-			"db_name": "test_db",
-			"table_name": "test_logs",
-			"user": "testuser",
-			"password": "testpass",
-			"max_events": 500
+			"plugin_path": "./mmap_plugin.so",
+			"clickhouse": {
```
krizhanovsky: I think we don't need the extra level of nesting and can treat `plugin_path` just as other options. There is no other choice than `clickhouse`, so the section name doesn't make much sense

logger/t/test_config.cc
```diff
@@ -362,13 +397,16 @@ TEST_F(ConfigTest, TableNameValidation_DefaultNameValidation)
...
+			}
 		}
 	})");
 
 	auto config = TfwLoggerConfig::load_from_file(config_path);
 	ASSERT_TRUE(config.has_value());
 	EXPECT_NO_THROW(config->validate())
 		<< "Default table name should be valid";
-	EXPECT_EQ(config->clickhouse.table_name, "access_log");
+	EXPECT_EQ(config->clickhouse_mmap->table_name, "access_log");
```
krizhanovsky, line 404: Please also add new tests: no path to a plugin or wrong path (we should correctly terminate with an adequate message), maybe something mode

logger/clickhouse.cc
```diff
@@ -94,6 +95,52 @@ TfwClickhouse::make_block()
 	block_.Reserve(max_events_ * 2);
 }
 
+bool
+TfwClickhouse::do_reconnect() {
```
krizhanovsky:
```suggestion
TfwClickhouse::do_reconnect()
try {
   ...
} catch ...
```

logger/event_processor.cc
```diff
@@ -0,0 +1,105 @@
...
+	return true;
+}
+
+tus::Error<bool>
+EventProcessor::consume_event()
+{
+	auto reconnect_result = handle_reconnection();
+	if (!reconnect_result) {
+		return tus::error(tus::Err::DB_CLT_TRANSIENT);
+	}
```
krizhanovsky:
```suggestion
	if (!handle_reconnection())
		return tus::error(tus::Err::DB_CLT_TRANSIENT);
```

logger/event_processor.hh
```diff
@@ -0,0 +1,56 @@
...
+	EventProcessor& operator=(const EventProcessor&) = delete;
+
+	bool make_background_work() noexcept;
+	[[nodiscard]] bool flush(bool force = false) noexcept;
+	virtual tus::Error<bool> consume_event();
+
+	virtual void request_stop() noexcept = 0;
+	virtual bool stop_requested() noexcept = 0;
+
+
```
krizhanovsky:
```suggestion

```

logger/event_processor.cc
```diff
@@ -0,0 +1,105 @@
...
+}
+
+bool
+EventProcessor::handle_reconnection()
+{
+	if (!db_->needs_reconnect.load(std::memory_order_acquire))
+		return true;
+
+	if (!db_->should_attempt_reconnect())
+		return false;
```
krizhanovsky: Why do we check `needs_reconnect` if we also check at the beginning of  `should_attempt_reconnect()`?

logger/event_processor.cc
```diff
@@ -0,0 +1,105 @@
...
+		std::chrono::steady_clock::now(),
+		std::memory_order_release
+	);
+
+	bool success = false;
+	try {
+		success = db_->do_reconnect();
+	} catch (...) {
+		success = false;
+	}
```
krizhanovsky: `do_reconnect()` does not throw

logger/clickhouse.cc
```diff
@@ -107,22 +154,51 @@ TfwClickhouse::handle_block_error() noexcept
...
-		opts.SetPassword(pswd);
+	return (now - last_attempt) >= timeout;
+}
 
-	client_ = std::make_unique<ch::Client>(opts);
+void
+TfwClickhouse::update_reconnect_timeout(bool success) noexcept
+{
+	if (success) {
+		reconnect_timeout.store(std::chrono::seconds(0));
```
krizhanovsky: We use early exit functions structuring to minimize nesting

```suggestion
		reconnect_timeout.store(std::chrono::seconds(0));
		return;
	}
	....
```

logger/clickhouse.hh
```diff
@@ -65,17 +66,53 @@ public:
...
 	[[nodiscard]] bool commit(bool force = false) noexcept;
 	bool handle_block_error() noexcept;
 
+	bool should_attempt_reconnect() const noexcept;
+	void update_reconnect_timeout(bool success) noexcept;
+	bool do_reconnect();
+
+public:
+	std::atomic<bool> needs_reconnect{false};
+	std::atomic<std::chrono::steady_clock::time_point>last_reconnect_attempt{
```
krizhanovsky:
```suggestion
	std::atomic<std::chrono::steady_clock::time_point> last_reconnect_attempt{
```

krizhanovsky: Why do we need these two as atomics? We have per-EventProcessor `db_` instances, `ClickHouse` have no separate context and called only from `EventProcessor` - where is concurrency?

It seems `needs_reconnect` should be private and probably it'd be good to name with for 1st face, `need_reconnect`

logger/event_processor.cc
```diff
@@ -0,0 +1,105 @@
...
+
+	db_->update_reconnect_timeout(success);
```
krizhanovsky: Why don't `TfwClickhouse::do_reconnect()` update the timeout? It seems now too much stuff is opened in the ClickHouse interface. Probably the whole `EventProcessor::handle_reconnection()` should be inside ClickHose

logger/event_processor.cc
```diff
@@ -0,0 +1,105 @@
...
+		return tus::error(tus::Err::DB_CLT_TRANSIENT);
+	}
+
+	return do_consume_event();
+}
+
+bool
+EventProcessor::handle_reconnection()
+{
+	if (!db_->needs_reconnect.load(std::memory_order_acquire))
```
krizhanovsky: The event loop is a super hot piece of code

```suggestion
	if (!db_->needs_reconnect.load(std::memory_order_acquire)) [[likely]]
```

fw/access_log.h
```diff
@@ -108,4 +108,121 @@ void do_access_log(TfwHttpResp *resp);
...
+template<>
+struct TfwBinLogTypeTraits<TFW_MMAP_LOG_DROPPED>
+	: TfwBinLogTypeCommonTraits<TFW_MMAP_LOG_DROPPED>
+{
+	using ColType = ch::ColumnUInt64;
+	using ValType = uint64_t;
+};
+
+#endif /* __cplusplus */
+
```
krizhanovsky: Why do we need to have this in Tempesta FW code?

logger/clickhouse.cc
```diff
@@ -195,3 +265,19 @@ tfw_column_factory(ch::Type::Code code)
...
+						      uint16_t) noexcept;
+template void
+TfwClickhouse::append_int<ch::ColumnUInt32, uint32_t>(TfwBinLogFields,
+						      uint32_t) noexcept;
+template void
+TfwClickhouse::append_int<ch::ColumnUInt64, uint64_t>(TfwBinLogFields,
+						      uint64_t) noexcept;
+template void
+TfwClickhouse::append_int<ch::ColumnIPv6, in6_addr>(TfwBinLogFields,
+						    in6_addr) noexcept;
```
krizhanovsky: It's still better to get rid of douple specification of the type in the template since `read_int()` also propagates the double typing to `append_int()`

logger/clickhouse.cc
```diff
@@ -94,6 +95,52 @@ TfwClickhouse::make_block()
...
+TfwClickhouse::append_timestamp(uint64_t timestamp)
+{
+	block_[0]->As<ch::ColumnDateTime64>()->Append(timestamp);
+}
+
+template <typename ColType, typename ValType>
+void
+TfwClickhouse::append_int(TfwBinLogFields field, ValType value)
+{
+	const size_t col_index = field_to_column_index(field);
```
krizhanovsky: I think this can be constexpr since we pass `field` from the very upper level as a constant

logger/access_log_plugin/access_log_processor.cc
```diff
@@ -0,0 +1,321 @@
+/**
```
krizhanovsky, line 1: "mmap" is just a way to write events and other logs may use the same technique, so it'd be better to rename the couple of files to something like `access_log_processor.[ch][ch]`

logger/mmap_processor.hh
```diff
@@ -0,0 +1,49 @@
+/**
+ *		Tempesta FW
+ *
+ * Copyright (C) 2024-2025 Tempesta Technologies, Inc.
+
```
krizhanovsky: extra empty line

logger/mmap_processor.hh
```diff
@@ -0,0 +1,49 @@
...
+	bool stop_requested() noexcept override;
+	unsigned int get_cpu_id() const noexcept;
+
+protected:
+	tus::Error<bool> do_consume_event() override;
+
+private:
+	int device_fd_;
+	TfwMmapBuffer *buffer_;
+	size_t size_;
```
krizhanovsky: please align the members as in other places

logger/access_log_plugin.hh
```diff
@@ -0,0 +1,49 @@
...
+ */
+#pragma once
+
+#include <string>
+#include <memory>
+
+#include "../fw/mmap_buffer.h"
+#include "../libtus/error.hh"
+#include "plugin_interface.hh"
+#include "event_processor.hh"
```
krizhanovsky: please order the headers by type and in alphabetical order as we do it usually

logger/event_processor.cc
```diff
@@ -0,0 +1,105 @@
+
```
krizhanovsky: extra empty line

logger/event_processor.cc
```diff
@@ -0,0 +1,105 @@
...
+bool
+EventProcessor::flush(bool force) noexcept
+{
+	if (!handle_reconnection()) {
+		spdlog::debug("DB flushing skipped due to reconnection issues");
+		return false;
+	}
+
+	if (!db_->commit(force ? TfwClickhouse::FORCE : false)) {
+		db_->needs_reconnect.store(true, std::memory_order_release);
```
krizhanovsky: Shouldn't we try to reconnect immediately instead of doing this on the next loop iteration?

logger/plugin.cc
```diff
@@ -0,0 +1,141 @@
...
+#include <dlfcn.h>
+#include <spdlog/spdlog.h>
+#include <fmt/format.h>
+
+#include "../libtus/error.hh"
+#include "event_processor.hh"
+#include "plugin.hh"
+
+Plugin::Plugin(const std::string &plugin_path,
+	       const ClickHouseConfig& config,
```
krizhanovsky:
```suggestion
Plugin::Plugin(const std::string &plugin_path, const ClickHouseConfig& config,
```

logger/plugin.cc
```diff
@@ -0,0 +1,141 @@
...
+			 RTLD_LAZY | RTLD_LOCAL | RTLD_DEEPBIND);
+	if (!handle_)
+		throw tus::Except("Failed to load plugin {}: {}",
+				  plugin_path, dlerror());
+
+	auto get_api = reinterpret_cast<TfwLoggerPluginGetApiFunc>(
+		dlsym(handle_, "get_plugin_api"));
+
+	if (!get_api)
+	{
```
krizhanovsky:
```suggestion
	if (!get_api) {
```

logger/plugin.cc
```diff
@@ -0,0 +1,141 @@
...
+		dlclose(handle_);
+		handle_ = nullptr;
+		throw tus::Except("Plugin {} missing get_plugin_api function",
+				  plugin_path);
+	}
+
+	api_ = get_api();
+	if (!api_ || api_->version != TFW_PLUGIN_VERSION)
+	{
```
krizhanovsky:
```suggestion
	if (!api_ || api_->version != TFW_PLUGIN_VERSION) {
```

Please fix all other similar places

logger/mmap_plugin.cc
```diff
@@ -0,0 +1,166 @@
+/**
```
krizhanovsky, line 1: Do we need to separate this from mmap_processor.cc? Having that we're going to have quite a few plugins and all of them are pretty small, I'd prefer to keep the source tree smaller

logger/mmap_plugin.cc
```diff
@@ -0,0 +1,166 @@
...
+					     e.what()).c_str());
+	}
+
+	return nullptr;
+}
+
+void
+mmap_destroy_processor(void *processor)
+{
+	if (!processor) return;
```
krizhanovsky: Please, do self review and adjust the coding style!

logger/tfw_logger.cc
```diff
@@ -67,312 +70,166 @@ struct ParsedOptions {
...
+	//
+	// TfwClickhouse keeps Block and Client instances, so it should be unique
+	// per an event type (access, security, and xFW).
+	//
+	// All the clickhouses must be flushed on a loop.
+	for (size_t tries = 0; ; ) {
+		bool stop_requested = false;
+		for (const auto& processor : processors) {
+			if (processor->stop_requested())
+				stop_requested = true;
```
krizhanovsky: Now if the kernel side for only one of the plugins (e.g. FW) stops, the whole daemon stops, even if e.g. xFW continues to work

logger/tfw_logger.cc
```diff
@@ -67,312 +70,166 @@ struct ParsedOptions {
...
- * @return the amount of data read, can be less than all available data,
- * e.g. if ClickHouse throws and exception or some event record is broken.
- */
-[[nodiscard]] tus::Error<size_t>
-process_events(TfwClickhouse &db, std::span<const char> data) noexcept
-{
-	size_t read = 0;
+			if (result.value()) {
+				consumed_something = true;
+			}
```
krizhanovsky:
```suggestion
			if (result.value())
				consumed_something = true;
```

logger/tfw_logger.cc
```diff
@@ -67,312 +70,166 @@ struct ParsedOptions {
...
-					     event, p, size);
-
-	return static_cast<size_t>(p - data.data());
-}
+		bool consumed_something = false;
+		for (const auto& processor : processors) {
+			auto result = processor->consume_event();
+			if (!result) [[unlikely]] {
+				spdlog::error("Processor {} error: {}",
+					      processor->processor_id,
```
krizhanovsky: It'd be better to print a processor name, otherwise the error messages will be too cryptic

logger/tfw_logger.cc
```diff
@@ -67,312 +70,166 @@ struct ParsedOptions {
...
-{
-	size_t read = 0;
+			if (result.value()) {
+				consumed_something = true;
+			}
+		}
 
-	dbg_hexdump(data);
+		if (consumed_something) [[likely]]
+		{
```
krizhanovsky:
```suggestion
		if (consumed_something) [[likely]] {
```

logger/tfw_logger.cc
```diff
@@ -67,312 +70,166 @@ struct ParsedOptions {
...
+	std::optional<size_t>		xfw_clickhouse_max_events;
+
 	std::optional<fs::path>		log_path;
 };
 
-#ifdef DEBUG
+// All processors must be non-null
 void
-dbg_hexdump(std::span<const char> data)
+inner_loop(const std::vector<EventProcessorPtr> &processors) noexcept
```
krizhanovsky:
```suggestion
event_loop(const std::vector<EventProcessorPtr> &processors) noexcept
```

logger/clickhouse.cc
```diff
@@ -107,22 +154,51 @@ TfwClickhouse::handle_block_error() noexcept
...
+		} else if (current.count() < 300) {
+			reconnect_timeout.store(current * 2);
+		}
+	}
+}
+
+TfwClickhouse::TfwClickhouse(const ClickHouseConfig &config)
+	: table_name_(config.table_name)
+	, max_events_(config.max_events)
+	, client_options_([config]() {
```
krizhanovsky:
```suggestion
	, client_options_([config]()
	{
```

logger/mmap_plugin.cc
```diff
@@ -0,0 +1,166 @@
...
+	.create_processor	= nullptr,
+	.destroy_processor	= nullptr
+};
+
+constexpr char dev_path[] = "/dev/tempesta_mmap_log";
+constexpr std::chrono::seconds wait_for_dev{1};
+
+std::atomic<bool> *global_stop_flag;
+std::shared_ptr<TfwClickhouse> db;
+int dev_fd = -1;
```
krizhanovsky, line 51: I think we don't have reasons to make these 2 variables global. Why they aren't members of `MmapProcessor`?

Also please rename `MmapProcessor` to `AccessLogProcessor`

ttaym: MmapProcessor is a per worker (per cpu in fact) instance, so if some variables are  shared between all processors we need it to be global variables

logger/clickhouse.cc
ttaym: I cannot see how it can be constexpr if field is argument to function.

But I am doing refactoring in style of type traits of these lines. So it will look more similar to https://github.com/tempesta-tech/tempesta/pull/2513#discussion_r2382534666

logger/clickhouse.hh
ttaym: each eventProcessor has shared pointer to db_ not, so all reconnection logic shared between eventprocessor therefor a need for atomic here.

logger/event_processor.cc
ttaym: `should_attempt_reconnect()` gives cumulative answer about should we try reconnecting or not, while needs_reconnect is false only when reconnecting is done sucessfully. They are not the same.

logger/event_processor.cc
ttaym: Before the PR we had been doing it on next iteration. I do not exactly understand what is a best strategy.

logger/event_processor.cc
ttaym: I cannot see how whole `handle_reconnection()` can be inside clickhouse, but you are right that there is more public stuff then needed. I hide `update_reconnect_timeout()` under private.

logger/mmap_plugin.cc
ttaym: I suppose that if you want we can implement event processors and plugins inside single processor unit, yes.

logger/tfw_logger.cc
ttaym: Yes, that is deficiency of the PR, and i think i will not be on time implementing private stop flags. So it seems that better to do it on further work outside this PR.

I will leave note about it on PR description.

consuelo2210:
Please, also add possible improvement:

In the future, it would be good to eliminate the `sleep` in the event loop.
Ideally, each plugin could signal when new data is available (e.g., via interface and a condition variable),
so the main loop would wake up either on data availability or on a periodic timer and continue processing.
This way we could avoid unnecessary sleeping and make the event loop more responsive.

logger/clickhouse.cc
ttaym: Implemented first variant of traits architecture from @consuelo2210

logger/tfw_logger.cc
consuelo2210: @krizhanovsky Maybe I still need this to fix in Anton's version

logger/clickhouse.hh
consuelo2210: I think that now I can remove atomics from here.

logger/event_processor.cc
consuelo2210: Now interfaces don't have a lot of public methods

logger/clickhouse.cc
consuelo2210: Done

logger/tfw_logger.cc
consuelo2210: Fixed

logger/event_processor.cc
consuelo2210: extra check removed from should_attempt_reconnect, should_attempt_reconnect renamed

logger/clickhouse.hh
consuelo2210: Removed

logger/clickhouse/clickhouse_decorator.cc
```diff
@@ -0,0 +1,99 @@
...
+	}
+}
+
+bool
+ClickHouseDecorator::flush(bool force) noexcept
+{
+	try {
+		block_.RefreshRowCount();
+
+		if (force) {
```
artem98, line 85: I'm not sure it is worth optimization, but if make bool force as template param, we could get rid of one branching on runtime

logger/tfw_logger.cc
```diff
@@ -67,313 +62,186 @@ struct ParsedOptions {
...
+
+		if (++tries < POLL_N) {
+			// Several tries with small sleeping to let the kernel
+			// fill the buffer and not to consume CPU in vain.
+			std::this_thread::sleep_for(delay);
+		}
+		else {
+			// There were nothing to do for POLL_Nms and probably
+			// the system is just idle - good time to flush all
+			// clollected data:
```
artem98: typo: collected

logger/plugin.cc
```diff
@@ -0,0 +1,220 @@
...
+public:
+	ProcessorHandle(TfwLoggerPluginApi *api, void *processor)
+		: api_(api), processor_(processor)
+	{
+		if (!api || !processor || !api_->is_active || !api_->request_stop
+		         || !api_->consume || !api_->send)
+			throw tus::Except("Plugin api is not fully presented");
+	}
+
+	~ProcessorHandle()
```
artem98: NIT: override

logger/plugin.cc
```diff
@@ -0,0 +1,220 @@
...
+	virtual int send(bool force) noexcept override
+	{
+		assert(api_->send);
+		return api_->send(processor_, force);
+	}
+
+	virtual std::string_view name() const noexcept override { return api_->name; };
+
+public:
+	ProcessorHandle& operator=(ProcessorHandle &&other) noexcept
```
artem98: NIT: move above to move constructor

etc/tfw_logger.json
```diff
@@ -1,6 +1,7 @@
 {
     "log_path": "/var/log/tempesta/tfw_logger.log",
     "access_log": {
+        "plugin_path": "/home/kestrel/escudo/fw/logger/access_log_plugin/access_log.so",
```
artem98, line 4: Just for my understanding: what does `kestrel` mean here?

libtus/error.hh
```diff
@@ -79,6 +82,12 @@ make_error_code(Err e) noexcept
 	return {static_cast<int>(e), tfw_error_category()};
```
artem98: NIT: It is possible to use make_error_code_from_int

logger/clickhouse/clickhouse_with_reconnect.cc
```diff
@@ -0,0 +1,115 @@
...
+ClickhouseWithReconnection::update_reconnect_timeout(bool success) noexcept
+{
+	if (success) {
+		reconnect_timeout_ = std::chrono::seconds(0);
+		return;
+	}
+
+	if (reconnect_timeout_.count() == 0)
+		reconnect_timeout_ = std::chrono::seconds(1);
+	else if (reconnect_timeout_.count() < 300)
```
artem98: Maybe we should move it into class as static const field?

logger/tfw_logger.cc
```diff
@@ -67,313 +62,186 @@ struct ParsedOptions {
...
+			// now no new events will be pushed to the buffer
+			// and we can process the rest of the events.
+			for (auto& processor : processors)
+				processor->request_stop();
+		}
 
-	dbg_hexdump(data);
+		bool consumed_something = false;
+		for (auto it = processors.begin(); it != processors.end(); ) {
+			auto& processor = *it;
```
artem98, line 123: Should we check whether the processor is active here?

logger/tfw_logger.cc
```diff
@@ -67,313 +62,186 @@ struct ParsedOptions {
...
-								data.data());
-
-			switch (ev->type) {
-			case TFW_MMAP_LOG_TYPE_ACCESS: {
-				const auto off = read_access_log_event(db, data);
-				data = data.subspan(off);
-				read += off;
-				break;
+			if (consumed) {
+				consumed_something = true;
```
artem98, line 135: Just for my understanding: why don't we proceed to next processor int this case? Is it safe when there is permanent event flow for the first processor?

logger/tfw_logger.cc
artem98: If i'm not wrong it is unsafe cast
Sysconf may return negative value as error code and we try to check unsigned type size_t for <= 0.
I suggest to get cpu_count as it is and later cast it directly to `unsigned`

logger/clickhouse/clickhouse.cc
```diff
@@ -0,0 +1,112 @@
...
+ * Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ */
+#include <spdlog/spdlog.h>
+
+#include "../../libtus/error.hh"
+#include "clickhouse.hh"
+
+
+bool
+TfwClickhouse::reestablish_connection() noexcept
```
artem98, line 32: I thought that the convention from code style is that all 'do_sth' functions return int, not bool
But I'm not sure

logger/access_log_plugin/access_log_plugin.cc
```diff
@@ -0,0 +1,511 @@
...
+		plugin_log_error(fmt::format(
+			"Access log is corrupted, skip current buffer: {}",
+			e.what()).c_str());
+		if (!db.handle_block_error())
+			return tus::error(tus::Err::DB_SRV_FATAL);
+		return 0;
+	}
+	catch (const std::exception &e) {
+		plugin_log_error(fmt::format(
+			"Cought a Clickhouse exception: {}."
```
artem98: typo: Caught

logger/access_log_plugin/access_log_plugin.cc
```diff
@@ -0,0 +1,511 @@
...
+		oss << " ";
+		for (int j = 0; j < 16; ++j) {
+			if (i + j >= buflen)
+				break;
+			const char c = buf[i + j];
+			oss << static_cast<char>(std::isprint(c) ? c : '.');
+		}
+		oss << std::endl;
+	}
+	plugin_log_info(oss.str().c_str());
```
artem98: NIT: little bit faster would be `oss.view().data()`

logger/clickhouse/clickhouse_decorator.cc
artem98: Also, imho, it was more comprehensible, when we had FORCE static constexpr (and NOT_FORCED also )

logger/clickhouse/clickhouse_with_reconnect.cc
```diff
@@ -0,0 +1,115 @@
...
+
+#include <stdexcept>
+
+#include <spdlog/spdlog.h>
+
+#include "../../libtus/error.hh"
+#include "clickhouse_with_reconnect.hh"
+
+ClickhouseWithReconnection::ClickhouseWithReconnection(ch::ClientOptions &&client_options)
+	: TfwClickhouse(std::forward<ch::ClientOptions>(client_options))
```
const-t: Reconnection logic seems broken. There is two cases for reconnect.
1. Retry to connect while establishing new connection. `TfwClickhouse()` throws exception when `ch::Client` can't establish connection then it handles in `mmap_create_processor() where we just print msg and exit, but don't do any retries and just exit.
2. When already established connection was disconnected. It happens silently and when we trying to insert new record we getting "Clickhouse insert error" and ignore it. However we have to restore the connection using `ResetConnection()`. [Link to doc](https://github.com/ClickHouse/clickhouse-cpp?tab=readme-ov-file#retries).

logger/access_log_plugin/Makefile
```diff
@@ -0,0 +1,62 @@
...
+all: $(TARGET)
+
+# We don't want to worry about compatibility, so just link ClickHouse statically.
+# But we need to pass CXXFLAGS to build clickhouse with our building flags.
+$(CLICKHOUSE_LIB):
+	$(MAKE) -C $(CLICKHOUSE_DIR) TARGET_DIR=$(abspath .) CXXFLAGS="$(CXXFLAGS)" -j$(shell nproc)
+
+$(TARGET): $(CLICKHOUSE_LIB) $(OBJS)
+	$(CXX) -shared -fPIC $(CXXFLAGS) -o $@ $(OBJS) \
+		-Wl,--whole-archive,tfw_clickhouse.a,--no-whole-archive $(LDFLAGS)
```
const-t:
```suggestion
all:
# We don't want to worry about compatibility, so just link ClickHouse statically.
# But we need to pass CXXFLAGS to build clickhouse with our building flags.
	$(MAKE) -C $(CLICKHOUSE_DIR) TARGET_DIR=$(abspath .) CXXFLAGS="$(CXXFLAGS)" -j$(shell nproc)
	$(MAKE) $(TARGET)

$(TARGET): $(OBJS)
	$(CXX) -shared -fPIC $(CXXFLAGS) -o $@ $(OBJS) \
		-Wl,--whole-archive,tfw_clickhouse.a,--no-whole-archive $(LDFLAGS)
```
I would like to suggest something like this, because we need to rebuild `logger/clickhouse` directory if files in this directory have been changed. With current approach we skip build, because `$(TARGET)` already built.

const-t (changes requested):

logger/access_log_plugin/Makefile
consuelo2210: I will also remove build in TargetDir. We don't have Monorepo for now, so we don't need to make build more complicated: both projects are building separately, just allow to overload build flags

logger/clickhouse/clickhouse_with_reconnect.cc
consuelo2210: I agree with the first point - it is a problem and I will fix it, but with the second one it is not just
"Clickhouse insert error"

client_->flush(table_name_, block_);
is noexcept, so the only place where we can catch this is inside TfwClickhouse on it's flush. And in this case we aregoing to setup 
	needs_reconnect_ = true; in ClickhouseWithReconnection
	
	By the way, I am not doing it on execute

logger/clickhouse/clickhouse_decorator.cc
consuelo2210: I'll leave it in the comments — not sure it's worth it.

etc/tfw_logger.json
consuelo2210: It is just my local path :) It doesn't important here, because the program only checks parsing

logger/clickhouse/clickhouse_with_reconnect.cc
consuelo2210: I decided to create ReconnectPolicy class

logger/tfw_logger.cc
consuelo2210: I will rename is_active to has_stopped, so it will be clearer, that we don't need to use this check here (we still can consume a collected info from processor, even it has stopped)

logger/tfw_logger.cc
consuelo2210: But we proceed

logger/tfw_logger.cc
consuelo2210: Wow! thank you!

logger/clickhouse/clickhouse_with_reconnect.cc
consuelo2210: ClickHouse establishes a connection only during `execute` and `flush` operations, so reconnection attempts will occur only on these actions.  
~~There is no need to reconstruct the ClickHouse class, since its constructor no longer performs any connection.~~
Also the constructor can cause connection, so I created LazyInitClickhouse, which organizes logic with reconnection by timer

logger/tfw_logger.cc
```diff
@@ -561,9 +550,7 @@ execute_workers() noexcept(false)
 	 * is more reliable on NUMA systems and machines with 100+ CPUs while
 	 * hardware_concurrency() is just a "hint".
 	 */
-	auto cpu_count = static_cast<size_t>(sysconf(_SC_NPROCESSORS_ONLN));
-	if (cpu_count <= 0)
-		throw tus::Except("Cannot determine CPU count");
+	size_t cpu_count = static_cast<size_t>(std::max(sysconf(_SC_NPROCESSORS_ONLN), 1L));
```
artem98: I think we should not ignore possible error codes here

logger/clickhouse/clickhouse.cc
consuelo2210: If you don't mind, I will left it for the time after release. I need to fix Error class too, because now it returns positive values.

logger/clickhouse/lazy_init_clickhouse.hh
```diff
@@ -0,0 +1,105 @@
...
+			return false;
+		return (*ptr)->flush(table_name, block);
+	}
+
+private:
+	/**
+	 * Try to get initialized object.
+	 * Returns nullptr if connection is unavailable.
+	 */
+	Ptr* get() noexcept
```
artem98, line 75: NIT: I think optional would be better here, but it is matter of taste

logger/access_log_plugin/access_log_processor.cc
```diff
@@ -339,5 +339,5 @@ AccessLogProcessor::request_stop() noexcept
 int
 AccessLogProcessor::has_stopped() noexcept
 {
-	return __atomic_load_n(&buffer_->is_ready, __ATOMIC_ACQUIRE);
+	return __atomic_load_n(&buffer_->is_ready, __ATOMIC_ACQUIRE)? 0: 1;
```
artem98: NIT: for codestyle spaces around ? and : are necessary

logger/access_log_plugin/Makefile
```diff
@@ -0,0 +1,62 @@
...
+endif
+
+all: $(TARGET)
+
+# We don't want to worry about compatibility, so just link ClickHouse statically.
+# But we need to pass CXXFLAGS to build clickhouse with our building flags.
+$(CLICKHOUSE_LIB):
+	$(MAKE) -C $(CLICKHOUSE_DIR) TARGET_DIR=$(abspath .) CXXFLAGS="$(CXXFLAGS)" -j$(shell nproc)
+
+$(TARGET): $(CLICKHOUSE_LIB) $(OBJS)
```
const-t: clickhouse directory still not be built, because `tfw_clickhouse.a` exists during re-building, therefore we don't call make for clickhouse directory.

logger/clickhouse/clickhouse_decorator.cc
```diff
@@ -0,0 +1,119 @@
...
+		handle_block_error();
+}
+
+bool
+ClickHouseDecorator::ensure_table_created() noexcept
+{
+	if (!needs_create_table_) [[likely]]
+		return true;
+
+	needs_create_table_ = client_->execute(table_creation_query_);
```
const-t:
```suggestion
	needs_create_table_ = !client_->execute(table_creation_query_);
```
We need to create table only if `execute` returned `false`, isn't?

logger/access_log_plugin/Makefile
consuelo2210: Yeh, sorry, missed it with rework

PR #2514, opened 2025-09-17
Fix write size of response_content_length in mmap access log

krizhanovsky:
We moved to 64-bit response length field in Clickhouse table, but were still writing only 32 bit values to the ring buffer.

Thanks to tfw_mmap_log_field_len() we were reading 32 bits, so this patch only fixes the maximum allowed values, but there were no data corruptions.


const-t (approved):
LGTM

PR #2524, opened 2025-09-30
Fix panic under stress

const-t:


const-t:
After this patch I've noticed that performance grew up on 6.12.12 vs 5.10.35, however only for non cache mode. The results observed on VM with virtio NIC driver. I'm pretty sure it's because of better skb coalescing on the new kernel thanks to better memory management in virtio. We have less skbs, they are with larger fragments. Also we have less skbs with `frag_list` because coalescing happens after GRO.

#### 5.10.35
```
finished in 15.02s, 7869.00 req/s, 769.21MB/s  
requests: 118035 total, 128035 started, 118035 done, 118035 succeeded, 0 failed, 0 errored, 0 timeout  
status codes: 118040 2xx, 0 3xx, 0 4xx, 0 5xx  
```
#### 6.12.12
```
finished in 15.02s, 18251.87 req/s, 1.74GB/s  
requests: 273778 total, 283778 started, 273778 done, 273778 succeeded, 0 failed, 0 errored, 0 timeout  
status codes: 273825 2xx, 0 3xx, 0 4xx, 0 5xx   
```

EvgeniiMekhanik (approved):

fw/cache.c
```diff
@@ -2812,7 +2812,7 @@ tfw_cache_add_body_page(TfwMsgIter *it, char *p, int sz, bool h2)
 }
 
 static inline bool
-tfw_cache_should_append_body_skb(TfwMsgIter *it, unsigned long body_sz,
+tfw_cache_should_append_body_skb(TfwMsgIter *it, unsigned int body_sz,
```
krizhanovsky: It seems we still have problem in other places, e.g. in `tfw_cache_copy_resp()` where we accept `size_t tot_len` and essentially copy it to the cache entry with only 4 bytes of size. Why web objects can not exceed 4GB?

fw/http_msg.c
```diff
@@ -1295,7 +1305,7 @@ __tfw_http_msg_expand_from_pool(TfwHttpMsg *hm, const TfwStr *str,
 					it->frag = frag - 1;
 				}
 
-				skb_room = SS_SKB_MAX_DATA_LEN - it->skb->len;
+				skb_room = MSG_SKB_MAX_DATA_LEN - it->skb->len;
```
krizhanovsky, line 1311: Do we need `skb_room` at all? Now it looks like "unlimited" in practice. This function is hot, so if we can optimize it, we should do this.

fw/cache.c
const-t: The size increased in the last commit.

PR #2530, opened 2025-10-09
Fix Tempesta FW nf hook for ipv6 connections

EvgeniiMekhanik:
- Rewrite `__ipv6_hdr_check` function for correct iteration over all extention headers until upper-layer protocol header is reached.
- Add supporting of `IPPROTO_ICMPV6` protocol for correct working of NDP (neighbor discovery protocol).
- Add debug messages for the cases when `__ipv4_hdr_check` or `__ipv6_hdr_check` fails.

### DIFF
--- a/fw/filter.c
+++ b/fw/filter.c
@@ -16,7 +16,7 @@
  * application level filters must be implemented.
  *
  * Copyright (C) 2014 NatSys Lab. (info@natsys-lab.com).
- * Copyright (C) 2015-2022 Tempesta Technologies, Inc.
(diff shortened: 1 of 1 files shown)


fw/filter.c
```diff
@@ -229,8 +244,10 @@ tfw_ipv6_nf_hook(void *priv, struct sk_buff *skb,
 	struct ipv6hdr *ih;
 
 	ih = __ipv6_hdr_check(skb);
-	if (!ih)
+	if (!ih) {
+		T_DBG("Invalid ipv6 header in skb\n");
```
const-t: Maybe we need statistics counter for this?

PR #2540, opened 2025-10-21
Mekhanik evgenii/more error injection

EvgeniiMekhanik:


tls/tls_ticket.c
```diff
@@ -90,14 +90,28 @@ __ttls_ticket_gen_key(unsigned long ts, const char *secret, char *digest)
...
+		goto exit;
+
+	r = ttls_md_hmac_update(&md_ctx, ticket_key_sym_iv,
+				SLEN(ticket_key_sym_iv));
+	if (unlikely(r))
+		goto exit;
+
+	r = ttls_md_hmac_update(&md_ctx, (unsigned char *)&ts, sizeof(ts));
+	if (unlikely(r))
+		goto exit;
```
krizhanovsky, line 110: Why the error accumulation doesn't work?

PR #2542, opened 2025-10-27
Mekhanik evgenii/more error injection master

EvgeniiMekhanik:


ktest/crypto/hash.h
```diff
@@ -23,6 +23,7 @@
 #include "linux/kernel.h"
 #include "linux/atomic.h"
 
+#define HASH_MAX_DIGESTSIZE	64
```
const-t, line 26: Is seems `SHA256_DIGEST_SIZE` will be more consistent with other code, because only this hmac function can be used. Also this size used in other places with this hmac context.

const-t (approved):
LGTM

ktest/crypto/hash.h
EvgeniiMekhanik: No! In all kernel code HASH_MAX_DIGESTSIZE is used.

const-t: This is not a big deal which size to use for ticket configuration, however

> In all kernel code HASH_MAX_DIGESTSIZE is used.

isn't true, `HASH_MAX_DIGESTSIZE` used in case when user doesn't know which algorithm will be used, but in case when algorithm is hard-coded we can use `DIGEST_SIZE` of this algo. Also if buffer dynamically allocated used `crypto_shash_digestsize()`. For tickets we have hard-coded algorithm in `ttls_tickets_init())` so we can use `SHA256_DIGEST_SIZE`. Also I should notice that `ttls_tickets_configure()` also use sha256 hash function and initialize buffer for it using `SHA256_DIGEST_SIZE`, it implies if `HASH_MAX_DIGESTSIZE` must be used we also corrupt memory in this place, but we not.

PR #2549, opened 2025-11-07
Mekhanik evgenii/fix rescheduling

EvgeniiMekhanik:


fw/http_sched_hash.c
```diff
@@ -177,8 +173,11 @@ __find_best_conn(TfwMsg *msg, TfwHashConnList *cl)
 	 */
 	idx = __bsearch(best_hash, cl);
 	conn = cl->conns[idx].conn;
-	if (likely(__is_conn_suitable(conn, hmonitor)))
-		return conn;
+	if (likely(tfw_srv_conn_get_if_live(conn))) {
```
const-t, line 176: This patch fixes only race with `TFW_CONN_B_RESEND`, however other checks still have races. The connection has shared data that may be read without synchronization with writer and there is no difference when we get a reference to the connection. Let's consider following: 
1. Call `__find_best_conn()`.
2. Call `tfw_srv_conn_get_if_live(conn)` and increment counter.
3. `__is_conn_suitable()` returns true.
4. We chose a connection and all checks have been passed successfully.
5. At this moment another CPU that also used this connection calls `tfw_srv_set_busy_delay()`, that leads to modification of `jbusytstamp`. At this moment we have a connection with passed checks and with incorrect `jbusytstamp`.

`tfw_srv_set_busy_delay()` always called under lock, but reading of the `jbusytstamp` happens without locking.

Also we have comment for `tfw_srv_conn_queue_full()` that describes such race.

I'm sure that we not need to fix all of these races at this moment. However in the commit message we should depict that fixed only case with `TFW_CONN_B_RESEND`, but other cases are still possible.

PR #2554, opened 2025-11-14
Drop server connection if client connection is dropped

EvgeniiMekhanik:
There was a problem with processing responses for requests with already dropped client connection. If some malicious client requests large response and then immediatly closes connection, Tempesta FW continue to receive large response for a long time. If such client does this action in the loop, this leads to huge memory consumption and overload of the server connections queues.
To solve this problem, the simplest possible solution was proposed - during processing response Tempesta FW checks that appropriated request belongs to already dropped client connection, we immediately drop server connection. During server conenction reestabling, we dropped all requests with already dropped client connections and resend all other requests.
In current architecture (with several different locks) there is no good solution to immediately rescheduling requests during dropping client connection, because it should be done under the `fwd_queue` lock of server connection (to which such request belongs to). Moreover we should access such server conenction also under the lock, because it can be deleted at the same time.

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -2078,7 +2078,14 @@ tfw_http_req_fwd_send(TfwSrvConn *srv_conn, TfwServer *srv, TfwHttpReq *req,
 		return -EBADF;
 	}
 
-	if (!(r = tfw_connection_send((TfwConn *)srv_conn, (TfwMsg *)req)))
+	r = tfw_connection_send((TfwConn *)srv_conn, (TfwMsg *)req);
+	if (!(req->msg.ss_flags & SS_F_KEEP_SKB)) {
+		struct sk_buff *skb_head;
+
+		skb_head = arch_xchg(&req->msg.skb_head, NULL);
+		ss_skb_queue_purge(&skb_head);
+	}
+	if (likely(!r))
 		return 0;
 
 	T_DBG2("%s: Forwarding error: conn=[%p] req=[%p] error=[%d]\n",
@@ -3018,7 +3025,7 @@ tfw_http_conn_release(TfwConn *conn)
  * connection threads.
  */
 static void
-tfw_http_conn_cli_drop(TfwCliConn *cli_conn)
+tfw_http_cli_conn_drop(TfwCliConn *cli_conn)
 {
 	TfwHttpReq *req, *tmp_req;
 	TfwHttpResp *resp, *tmp_resp;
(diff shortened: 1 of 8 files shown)


fw/http.c
```diff
@@ -2040,6 +2040,48 @@ tfw_http_req_evict(TfwSrvConn *srv_conn, TfwServer *srv, TfwHttpReq *req,
...
+})
+
+static inline void
+do_access_log_req_lock(TfwHttpReq *req, int resp_status,
+		       unsigned long resp_content_length)
+{
+	spinlock_t *lock;
+
+	if (unlikely(!req->conn))
+		do_access_log_req(req, resp_status, resp_content_length);
```
const-t: Looks we do double log request when `!conn`

fw/http.c
```diff
@@ -2040,6 +2040,48 @@ tfw_http_req_evict(TfwSrvConn *srv_conn, TfwServer *srv, TfwHttpReq *req,
...
+{
+	TfwStream *stream = req->stream;
+
+	if (stream) {
+		req->stream = NULL;
+		stream->msg = NULL;
+	}
+}
+
+#define __REQ_CLI_CONN_LOCK(req)				\
```
const-t: Maybe convert it to static inline function?

fw/http.c
```diff
@@ -2040,6 +2040,48 @@ tfw_http_req_evict(TfwSrvConn *srv_conn, TfwServer *srv, TfwHttpReq *req,
 	       || tfw_http_req_evict_retries(srv_conn, srv, req, eq);
 }
 
+static inline void
+tfw_http_req_unlink_stream(TfwHttpReq *req)
```
const-t: unused

fw/http.c
```diff
@@ -2079,11 +2121,36 @@ tfw_http_req_fwd_send(TfwSrvConn *srv_conn, TfwServer *srv, TfwHttpReq *req,
...
-		struct sk_buff *skb_head;
+	/*
+	 * We set this flag during client connection drop. If this flag is set
+	 * here it means, that current request belongs to already dropped
+	 * client connection.
+	 */
+	if (unlikely(test_bit(__TFW_HTTP_B_REQ_DROP, req->flags))) {
+		/*
+		 * If error occurs during sending request to already dropped
+		 * client connection, delete such request from server
```
const-t, line 2135: Need to rephrase this `sending request to already dropped client connection`.

fw/http.c
```diff
@@ -564,6 +564,38 @@ tfw_http_prep_date(char *buf)
...
+			unsigned long resp_content_length)
+{
+	struct sk_buff *skb_head;
+
+	if (unlikely(!req->conn)) {
+		do_access_log_req(req, resp_status, resp_content_length);
+		return;
+	}
+
+	skb_head = arch_xchg(&req->msg.skb_head, NULL);
```
const-t, line 578: Loosing of access log record, we must avoid this

PR #2558, opened 2025-11-17
Mekhanik evgenii/fault injection mem

EvgeniiMekhanik:


lib/alloc.c
```diff
@@ -0,0 +1,74 @@
...
+
+void *
+tfw__alloc_percpu(size_t size, size_t align)
+{
+	return __alloc_percpu(size, align);
+}
+ALLOW_ERROR_INJECTION(tfw__alloc_percpu, NULL);
+EXPORT_SYMBOL(tfw__alloc_percpu);
+
+#endif
```
krizhanovsky: What about to rename the .c implementation to something saying that this is just fault injection wrappers, not any real allocations, like `fault_injection_alloc.c`?

fw/http_sched_ratio.c
```diff
@@ -24,6 +24,7 @@
 
 #include "sched.h"
 #include "tempesta_fw.h"
+#include "lib/alloc.h"
```
krizhanovsky: Why not to rename alloc.h as well?

linux-6.12.12.patch
```diff
@@ -733,6 +733,30 @@ index 18148e068..3da17f21e 100644
...
+ 
+ struct proc_dir_entry *proc_create_mount_point(const char *name)
+ {
+@@ -577,6 +580,9 @@ struct proc_dir_entry *proc_create_data(const char *name, umode_t mode,
+ 	return proc_register(parent, p);
+ }
+ EXPORT_SYMBOL(proc_create_data);
++#ifdef CONFIG_SECURITY_TEMPESTA
++ALLOW_ERROR_INJECTION(proc_create_data, NULL);
++#endif
```
krizhanovsky, line 756: It seems there is no corresponding PR in https://github.com/tempesta-tech/linux-6.12.12-tfw

db/core/htrie.c
```diff
@@ -251,15 +251,7 @@ tdb_init_mapping(void *p, size_t db_size, unsigned int rec_len)
...
-	if (db_size > TDB_MAX_DB_SZ) {
-		TDB_ERR("too large database size (%lu)", db_size);
-		return NULL;
-	}
-	/* Use variable-size records for large data to store. */
-	if (rec_len > TDB_BLK_SZ / 2) {
-		TDB_ERR("too large record length (%u)\n", rec_len);
-		return NULL;
-	}
+	BUG_ON(db_size > TDB_MAX_DB_SZ || rec_len > TDB_BLK_SZ / 2);
```
krizhanovsky: I don't think this is a good way to fix the problem: at some point we might see kernel crash instead of a user-friendly message in dmesg. This is not data plane, but it still crucial. It's better to handle `NULL` in `tdb_htrie_init()`. We don't have this in our CodingStyle, but I addressed this in https://tempesta-tech.com/blog/memory-safety-and-network-security/ and some of our clients have similar practice to avoid assertions if possible.

fw/main.c
```diff
@@ -233,6 +234,8 @@ tfw_mods_start(void)
 		if ((ret = mod->start())) {
 			T_ERR_NL("Unable to start module '%s': %d\n",
 				 mod->name, ret);
+			if (mod->stop && !mod->started)
+				mod->stop();
```
krizhanovsky, line 238: Thank you: with this fix we have smaller code and no code duplicates!

PR #2567, opened 2025-11-24
Fix BUG if `sock_clnt` module failed to start

EvgeniiMekhanik:


fw/sock_clnt.c
```diff
@@ -831,49 +873,10 @@ tfw_sock_clnt_start(void)
...
-		 * same connections - SS can handle it and it's expected that
-		 * softirqs close some of them while we wait.
-		 */
-		local_bh_enable();
-		schedule();
-		local_bh_disable();
-	}
-	local_bh_enable();
+	if (unlikely(r))
+		tfw_sock_clnt_stop();
```
krizhanovsky: I didn't get this fix. In `tfw_start()` we already have generic way to process such errors: got to `stop_mods` and call `stop` callbacks, `tfw_sock_clnt_stop()` in this case. This fix doubles the logic and also opens a question whether we have similar bugs in other modules.

Also I think this fix should be backported to 0.8

EvgeniiMekhanik: `tfw_sock_clnt_stop` is called only if this module was already started. mod->started is true. `mod->started` is set after module successful start, so if this module failed to start `tfw_sock_clnt_stop` will never called. The same problem is for `sock_srv` module. It is fixed  in https://github.com/tempesta-tech/tempesta/pull/2558

Issue #1940, opened 2023-07-13
Multiple perforance regressions


krizhanovsky:
# Test configuration

Benchmark results on CPU i9-12900HK (0-11 are performance cores) for the current master d7294b86a563bd18d5af51f701821357ade089de (the same as release-0.7 as of Jul 13) vs Nginx 1.18.0 build with OpenSSL 3.0.2. Tempesta FW and Nginx are running in a VM with 2 vCPUs bound to performance cores 0 and 2. All the benchmark tools were running from the host machine. The VM uses multiqueue virtio-net.

Tempesta config:
```
listen 192.168.100.4:443 proto=https;
listen 192.168.100.4:80;

frang_limits {
	http_methods GET HEAD;
	http_uri_len 1000;
}

block_action attack reply;
block_action error reply;

srv_group default {
	server 192.168.100.4:8000;
}

vhost tempesta-tech.com {
	tls_certificate /root/tempesta/etc/tfw-root.crt;
        tls_certificate_key /root/tempesta/etc/tfw-root.key;
	tls_tickets secret="f00)9eR59*_/22" lifetime=7200;

	resp_hdr_set Strict-Transport-Security "max-age=31536000; includeSubDomains";

	proxy_pass default;
}

cache 1;
cache_fulfill * *;

http_chain redirection_chain {
	uri == "/blog"		-> 301 = /blog/1;
	uri == "/blog/"		-> 301 = /blog/1;
	uri == "/services"	-> 301 = /development-services;
	uri == "/services.html"	-> 301 = /development-services;
	uri == "/c++-services"	-> 301 = /development-services;
	uri == "/index.html"	-> 301 = /index;
	uri == "/company.html"	-> 301 = /company;
	uri == "/blog/fast-programming-languages-c-c++-rust-assembly" -> 301 = /blog/fast-programming-languages-c-cpp-rust-assembly;

	-> tempesta-tech.com;
}

http_chain {
	host == "tempesta-tech.com" -> redirection_chain;
}
```

Nginx config servicing `/var/www/tempesta-tech.com` with 26KB index file or `/var/www/html` with a 10 bytes index:
```
user www-data;
worker_processes auto;
pid /run/nginx.pid;

events {
    worker_connections 65535;
    multi_accept on;
    use epoll;
}
worker_rlimit_nofile 65535;

http {
    sendfile on;
    tcp_nopush on;
    tcp_nodelay on;
    keepalive_timeout 65;
    keepalive_requests 100000;
[...]

PR #2568, opened 2025-11-25
Simple settings against simple L7 DDoS

krizhanovsky:


scripts/tempesta.sh
```diff
@@ -225,6 +225,23 @@ setup()
 	sysctl -w net.core.netdev_max_backlog=10000 >/dev/null
 	sysctl -w net.core.somaxconn=131072 >/dev/null
 	sysctl -w net.ipv4.tcp_max_syn_backlog=131072 >/dev/null
+
+	# Limit the number of orphan sockets in FIN-WAIT-1,
+	# quite common for L7 DDoS.
+	sysctl -w net.ipv4.tcp_orphan_retries=1 >/dev/null # timeout for 1s
+	sysctl -w net.ipv4.tcp_retries2=5 >/dev/null # number of retries
+	sysctl -w net.ipv4.tcp_max_orphans=16384 >/dev/null
```
EvgeniiMekhanik, line 233: I am nit sure but what about
```
net.core.rmem_max = 2147483647 (or increase only 4x)
net.core.wmem_max = 2147483647 
# increase Linux autotuning TCP buffer limits 
# min, default, and max number of bytes to use
# allow auto-tuning up to 128MB buffers
net.ipv4.tcp_rmem=4096 65536 1073741824
net.ipv4.tcp_wmem=4096 65536 1073741824 (or increase only 4x as you do it before)
```
For high speed adapters it is recommend to increase this settings also

EvgeniiMekhanik: When I help to test iperf on 100Gb and Trex I use settings from this (https://fasterdata.es.net/host-tuning/linux/test-measurement-host-tuning/)

krizhanovsky: Yeah, I was thinking about these limits and decided not to go with them to not to let just a fistful of connections to waste all TCP resources leading to the same TCP OOM DoS. I added this to https://github.com/tempesta-tech/tempesta/issues/1940#issuecomment-3589204537

PR #2569, opened 2025-11-26
Fix deadlock during server connection release.

EvgeniiMekhanik:
If server connection releases and `TFW_CONN_B_DEL` is set, Tempesta FW reschedule all requests to other backend servers and send error responses to clients.
Responses are sended under the `req_qlock` of appropriate client connection, and also deleted under this lock. If response held the last referense on the appropriate server connection it can be released under the `ret_qlock` also. If such server connection contains requests from appropriate client connetion (which `ret_qlock` is locked) Tempesta FW try to lock already locked spin_lock. To fix this problem use the same solution as we use during client connection release - add responses which held last reference on server connection to special queue and free them later, when `req_qlock` or `seq_qlock` is unlocked.

Fix 2564

### DIFF
--- a/fw/http.c
+++ b/fw/http.c
@@ -3002,6 +3002,52 @@ tfw_http_conn_release(TfwConn *conn)
 	}
 }
 
+static inline void
+tfw_http_free_req_carefully(TfwHttpReq *req, struct list_head *resp_del_queue)
+{
+	TfwHttpResp *resp = req->resp;
+
+	/*
+	 * If `resp->conn` is not zero and response keeps the last
+	 * reference to the connection, we can't free this response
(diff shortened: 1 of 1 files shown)


fw/http.c
```diff
@@ -4872,16 +4882,21 @@ tfw_http_resp_fwd(TfwHttpResp *resp)
 			BUG_ON(!req->resp);
 			list_del_init(&req->msg.seq_list);
 			if (!test_bit(TFW_HTTP_B_CONTINUE_RESP,
-				     req->resp->flags))
-				tfw_http_resp_pair_free(req);
-			else
+				     req->resp->flags)) {
```
```suggestion
				     req->resp->flags))
	{
```

krizhanovsky (approved):
LGTM, but it seems we also need to backport this to 0.8

PR #2579, opened 2025-12-04
fix: Large mmap buffer allocation

const-t:
This patch allows to allocate large mmap buffer, up to 128M. Tempesta tries to allocate non-contiguous high order pages, when it fails rollbacks to order-0 pages. All array of pointers to allocated pages are stored in `struct TfwMmapBufferHolder`, they are freed on Tempesta unloading

### DIFF
--- a/fw/mmap_buffer.c
+++ b/fw/mmap_buffer.c
@@ -1,7 +1,7 @@
 /**
  *		Tempesta FW
  *
- * Copyright (C) 2024-2025 Tempesta Technologies, Inc.
+ * Copyright (C) 2024-2026 Tempesta Technologies, Inc.
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


fw/mmap_buffer.c
```diff
@@ -169,12 +168,163 @@ dev_file_vm_close(struct vm_area_struct *vma)
...
+		return NULL;
+
+	if (size >= PMD_SIZE)
+		shift = PMD_SHIFT;
+	order = shift - PAGE_SHIFT;
+
+	/* Allocate metadata page. */
+	head_page = alloc_pages_node(nid, GFP_KERNEL, 0);
+	if (unlikely(!head_page))
+		return NULL;
```
EvgeniiMekhanik, line 298: It seems that you don't free `pages` here

const-t: Fixed

krizhanovsky (approved):
LGTM

PR #2582, opened 2025-12-05
fix: Don't cleanup request during adjusting

const-t:


fw/http.c
```diff
@@ -3869,51 +3870,51 @@ tfw_h1_adjust_req(TfwHttpReq *req)
...
 	return r;
 
-clean:
-	T_DBG("%s: req [%p] adjusting has failed with code %i\n", __func__, req,
-	      r);
-	__tfw_http_req_cleanup(req);
+err:
+	T_DBG3("%s: req [%p] adjusting has failed with code %i\n", __func__,
+	       req,
+	       r);
```
EvgeniiMekhanik: `r` can be on the same string as req

fw/http.c
```diff
@@ -4488,8 +4490,14 @@ tfw_h2_adjust_req(TfwHttpReq *req)
 
 		do {
 			b_skbs = b_skbs->next;
-			if (WARN_ON_ONCE((b_skbs == old_head)))
+			if (WARN_ON_ONCE((b_skbs == old_head))) {
+				/*
+				 * @new_head will be freed in "err" label.
+				 * Prevent use after free.
+				 */
+				req->msg.skb_head = NULL;
```
EvgeniiMekhanik, line 4506: Why you don't restore old_head here?

const-t: From my point of view it is the same, at this moment we need to drop and free a request and req->msg.skb_head is not used anymore. The issue is double free if req->msg.skb_head and req->cleanup->skb_head point to the same memory and we must NULL one of these pointers I do NULL req->msg.skb_head just for simplicity, otherwise I need reassign skb_head and assign NULL to cleanup->skb_head. Anyway msg.skb_head and clean_up->skb_head will be freed from req's destructor, therefore we are safe and don't access freed memory.

PR #2583, opened 2025-12-05
fix: Don't cache responses to HEAD requests [0.8]

const-t:


const-t:
> LGTM, but why don't we port the cache_cfg.cache changes for tfw_cache_msg_cacheable(), tfw_cache_msg_cacheable() and tfw_cache_process() as we do this for master?

This is refactoring, decided to not include to this port.

PR #2584, opened 2025-12-05
fix: Hang when configuring a disabled NIC [0.8]

const-t:
We may wait for `down` interface forever, don't wait for `up` just configure it.


krizhanovsky (approved):
LGTM

PR #2586, opened 2025-12-12
feat: Add timeout to `ip_block` derective

const-t:


fw/filter.c
```diff
@@ -41,6 +41,7 @@
 #include "tdb.h"
 
 #include "lib/str.h"
+#include "lib/common.h"
```
krizhanovsky, line 43: it'd be better to place it before `lib/str.h` to preserve alphabetical order

fw/filter.c
```diff
@@ -106,14 +111,21 @@ tfw_filter_check_ip(struct in6_addr *addr)
 	TdbIter iter;
 
 	iter = tdb_rec_get(ip_filter_db, tfw_ipv6_hash(addr));
-	while (!TDB_ITER_BAD(iter)) {
+	if (TDB_ITER_BAD(iter))
+		return T_OK;
+	const long ts = tfw_current_timestamp();
+
+	do {
```
krizhanovsky, line 118: Why do we need postconditional loop and check `TDB_ITER_BAD(iter)` before? It seems to avoid getting timestamp in vain, but now it's very lightweight...

const-t: Only to avoid getting timestamp in the loop. Yes, now it is lightweight, did so just because there is no reason to do this in the loop.

fw/filter.c
const-t: fixed

Issue #2578, opened 2025-12-02
Set optimal `tcp_mem` value


const-t:
# Motivation

Current [approach](https://github.com/tempesta-tech/tempesta/blob/3da27768ce1c282f1b523fc08c2523c11f9187f5/scripts/tempesta.sh#L240) to set `tcp_mem` looks inaccurate. On start/stop we simply multiply value that leads to overflow.

# Scope

- [x] Need to find better approach to set optimal `tcp_mem`. Now we assume it must be the value that four times more than calculated value by the kernel, we should calculate it relying on the current available memory, instead of multiplying `tcp_mem` value.
- [x] Need to check other TCP configuration in [tests](https://github.com/tempesta-tech/tempesta-test/blob/aea0b137decc2306e9a284462ce45de17b0bd7bc/test_suite/prepare.py#L8)
- [x] Fix or disable tests that are broken after #2568. Need to check the test suite on CI.

krizhanovsky:
Original [patch](https://github.com/tempesta-tech/tempesta/commit/81338710f2288a70f00925fbd7d1f0c2a7229bf6) reduces `max_orphans` that broke CI.

PR #2587, opened 2025-12-14
The second attempt to improve sysctl settings

### LINKED ISSUE #1763: TCP: out of memory -- consider tuning tcp_mem
# Motivation
While running `h2load` test with many streams, e.g. `h2load https://f35tfw.local -t 2 -c 1000 -D 30 -m 100`, several `TCP: out of memory` warnings can be observed in dmesg.

# Scope
```
[  179.636034] TCP: out of memory -- consider tuning tcp_mem                                                                                                   
[  179.666216] TCP: out of memory -- consider tuning tcp_mem                                                                                                   
[  179.676047] TCP: out of memory -- consider tuning tcp_mem                                                                                                   
[  179.676353] TCP: out of memory -- consider tuning tcp_mem                                                                                                   
[  179.676747] TCP: out of memory -- consider tuning tcp_mem                                                                                                   
[  179.694963] TCP: out of memory -- consider tuning tcp_mem                                                                                                   
[  179.695276] TCP: out of memory -- consider tuning tcp_mem                                                                                                   
[  179.697049] TCP: out of memory -- consider tuning tcp_mem                                                                                                   
[  179.697399] TCP: out of memory -- consider tuning tcp_mem
```

The problem is probably caused by inaccurate `sock` memory accounting. 

This warning is produced by the function below:
```
(net/ipv4/tcp.c)
bool tcp_check_oom(struct sock *sk, int shift)
{
	bool too_many_orphans, out_of_socket_memory;

	too_many_orphans = tcp_too_many_orphans(sk, shift);
	out_of_socket_memory = tcp_out_of_memory(sk);

	if (too_many_orphans)
		net_info_ratelimited("too many orphaned sockets\n");
	if (out_of_socket_memory)
		net_info_ratelimited("out of memory -- consider tuning tcp_mem\n");
	return too_many_orphans || out_of_socket_memory;
}
EXPORT_SYMBOL(tcp_check_oom);
```

Memory accounting is performed by the following routines(and some others too):
```
(include/net/sock.h)
...
static inline bool sk_wmem_schedule(struct sock *sk, int size)
{
	if (!sk_has_account(sk))
		return true;
	return size <= sk->sk_forward_alloc ||
		__sk_mem_schedule(sk, size, SK_MEM_SEND);
}

static inline bool
sk_rmem_schedule(struct sock *sk, struct sk_buff *skb, int size)
{
	if (!sk_has_account(sk))
		return true;
	return size <= sk->sk_forward_alloc ||
		__sk_mem_schedule(sk, size, SK_MEM_RECV) ||
		skb_pfmemalloc(skb);
}
...
```

Setting `net.ipv4.tcp_mem` values ~4 times bigger than the ones calculated on boot, make the warning go away, but might not be the right solution. 

# Testing

Run `h2load` test with the aforementioned options.

krizhanovsky:
Previous patch 81338710f2288a70f00925fbd7d1f0c2a7229bf6 tried to fight with L7 DDoS sacrificing TCP correctness.

In the new version we do only 2 things:
1. Make FIN-WAIT-1 and FIN-WAIT-2 sockets to recycle faster, but still according to the RFCs
2. Increase tcp_mem, which makes sense just for a busy networking box.

We need to use `ip_block` to block an attacker and not let them create TCP connections: it's too late to limit orphaned sockets.

Fix #1763 and #2578

### DIFF
--- a/scripts/tempesta.sh
+++ b/scripts/tempesta.sh
@@ -215,7 +215,7 @@ setup()
 	echo 1 > /proc/sys/kernel/sysrq
 
 	# Automatically immediately reboot on kernel crashes and ignore kernel warnings.
-	echo '1' > /proc/sys/kernel/panic
+	echo 1 > /proc/sys/kernel/panic
 	echo 1 > /proc/sys/kernel/panic_on_oops
 	echo 0 > /proc/sys/kernel/panic_on_warn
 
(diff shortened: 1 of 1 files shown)


scripts/tempesta.sh
```diff
@@ -225,6 +225,35 @@ setup()
...
+	# The minimum number of retries, recommended by RFC 1122.
+	sysctl -w net.ipv4.tcp_retries2=8 >/dev/null
+	sysctl -w net.ipv4.tcp_fin_timeout=10 >/dev/null
+
+	# Increase the total TCP memory to mitigate
+	# "TCP: out of memory -- consider tuning tcp_mem" problem.
+	# This increases the total TCP memory, but leave per-socket limits as
+	# defaults to not allow too memory hungry sockets.
+	#
+	# Linux sets the a sysctl in tcp_init_mem() as ~5%, ~6% and ~9% of
```
const-t:
```suggestion
	# Linux sets the sysctl in tcp_init_mem() as ~5%, ~6% and ~9% of
```

RomanBelozerov (approved):

PR #2589, opened 2025-12-19
proposal: type-safe Error class with compile-time categories and enums.

consuelo2210:
The current implementation provides a type-safe framework for representing and propagating errors in a modular C++ system. Each error is associated with a specific enum and a category, and each category is defined via ErrorCategoryBase with a compile-time stable ID and name. The Error class combines the category ID and the enum value into a single integer code, which can be returned from a module or used for logging.

The error class provides convenient methods to retrieve the category name, the message for the specific enum, and a hexadecimal representation of the code. The operator<< is overloaded to print the full error as Category: Message (0xCODE). A helper function code_to_hex allows formatting the code in hex while preserving the sign if negative.

Modules can define their own categories and enums without changing the core error class. The framework also integrates well with std::expected<T, Error> and can be used to propagate errors or throw exceptions while preserving full traceability and diagnostic information.

NOTE:
Still need a mechanism to recover an error in another module from the integer code. 
It would also be useful to integrate this with std::expected to throw detailed 
exceptions carrying the typed error for propagation outside the module.

Update: I was thinking about adding constructor to Except(Error) and member std::shared_ptr < Error>. In that case we would be able to write detailed info on what(). But maybe we don't need to mix this 2 approaches. Or we can just construct Except with what from Error and forget about Error after it

### DIFF
--- a/libtus/error.hh
+++ b/libtus/error.hh
@@ -43,60 +43,119 @@ namespace tus {
 
 /*
  * ------------------------------------------------------------------------
- *	std::expected API for the data path
+ *	std::expected API for the data path:
+ * 		return std::expected<result, Error<Enum, Category>>
  * ------------------------------------------------------------------------
  */
-enum class Err : int {
-	// Clickhouse error.
-	DB_SRV_FATAL,
-	DB_CLT_TRANSIENT,
+/**
+ * ErrorCategoryBase defines a compile-time contract for error categories.
+ */
+template <typename Derived>
+struct ErrorCategoryBase
+ {
+	/**
+	 * We could use the CategoryId enum instead of uint16_t, but this would make
+	 * Tempesta aware of Escudo modules. On the other hand, having a single enum
+	 * for all categories is more reliable.
+	 */
+	static constexpr uint16_t id()
+	{
+		static_assert(requires { Derived::Id; },
+			"Category must define: static constexpr uint16_t Id");
+		return Derived::Id;
+	}
+
+	static constexpr std::string_view name()
+	{
+		static_assert(requires { Derived::Name; },
+			"Category must define: static constexpr std::string_view Name");
+		return Derived::Name;
+	}
 };
 
-class ErrorCategory : public std::error_category {
+struct TfwCategory: ErrorCategoryBase<TfwCategory>
+{
+	static constexpr uint16_t Id = 0x1000;
+	static constexpr std::string_view Name = "tfw";
+};
+
+/**
+ * Each Error instance is tied to a specific error enum (Enum) and a category
+ * (Category) defined via ErrorCategoryBase.
+ * Categories provide stable identifiers and names, enforced at compile-time.
+ * Each error has a combined integer code() composed of the category ID and the
+ * enum value.
+ */
+template <typename Enum, typename Category>
+class Error
+{
 public:
-	const char *
-	name() const noexcept override
+	constexpr Error(Enum e): enum_val_(static_cast<uint16_t>(e))
+		, category_id_(Category::id()), category_name_(Category::name())
 	{
-		return "error";
+		static_assert(std::is_same_v<std::underlying_type_t<Enum>, uint16_t>,
+			      "Enum underlying type must be uint16_t");
+		code_ = compose_code(category_id_, enum_val_);
 	}
 
-	std::string
-	message(int e) const override
+public:
+	/* Code to return outside the module*/
+	constexpr int
+	code() const noexcept
 	{
-		switch (static_cast<Err>(e)) {
-		case Err::DB_SRV_FATAL:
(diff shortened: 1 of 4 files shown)


libtus/error.hh
```diff
@@ -43,60 +43,119 @@ namespace tus {
...
 
-inline std::error_code
-make_error_code_from_int(int e) noexcept
-{
-	return {e, tfw_error_category()};
-}
+	constexpr std::string_view
+	message() const
+	{
+		return message(static_cast<Enum>(enum_val_));
```
artem98, line 118: I suppose we can add static assert here for clarity

logger/access_log_plugin/access_log_processor.cc
```diff
@@ -317,7 +318,7 @@ AccessLogProcessor::send(bool force) noexcept
 	if (writer_.flush(force))
 		return 0;
 
-	return static_cast<int>(tus::Err::DB_CLT_TRANSIENT);
+	return -ClickhouseError(DB_CLT_TRANSIENT).code();
```
artem98, line 321: I suppose in next iterations we shoul propagate expected to IPluginProcessor

consuelo2210: We can't do it until we have potentially different compilers. Error is not ABI compatible now. I think that it would be good for IPluginProcessor, when we move to one build system for all projects...

consuelo2210: Do you have any ideas how to make it ABI compatible?

artem98: I'm trying to figure it out locally
However, during this process, I have some thought:
imho it is not convenient to fix Enum type as template parameter, cause it makes this type less flexible, maybe we should leave Enum type template param only for constructor and store only message and code?

Example in which I found it inconvenient:

```
Error<Cat1, Enum1> func_returns_error_with_enum1();
Error<Cat1, Enum2> func_returns_error_with_enum2();

Unknown // what should this func return?
outer_func()
{
auto res1 = func_returns_error_with_enum1();
if(res1)
    return res1;
auto res2 = func_returns_error_with_enum2();
if(res2)
    return res2;
    
 return std::nullopt;
}
```
Also we should discuss policy for such functions without result besides error, or not error
std::expected is not so good for it, it can be optional or some special value Error

consuelo2210: Thanks for raising this — it’s an important point. I arrived at similar thoughts while working on the proposal, but decided to set it aside for now to let the idea settle and to get a clearer picture of the exception-handling approach overall.

If we take the presented class, introduce an interface for it, and wrap it with another class, we can hide the details of the concrete enum behind that interface. This would allow us to return std::expected<bool, Error> from the function, while Error can internally represent different enum + category combinations.

logger/access_log_plugin/access_log_processor.cc
```diff
@@ -218,13 +219,13 @@ process_events(AccessLogClickhouseDecorator &db, std::span<const char> data) noe
...
 		if (!db.handle_block_error())
-			return tus::error(tus::Err::DB_SRV_FATAL);
+			return std::unexpected(ClickhouseError(DB_SRV_FATAL));
 		return 0;
 	}
 	catch (const std::exception &e) {
 		spdlog::error("Caught a Clickhouse exception: {}."
 			      " Many events can be lost", e.what());
-		return tus::error(tus::Err::DB_SRV_FATAL);
+		return std::unexpected(ClickhouseError(DB_SRV_FATAL));
```
krizhanovsky, line 228: One of the reasons for the original `Error` was to reduce typing, so I think it makes sense to move `std::expected` to `using` or a parent class that we can just return say `tus::Err::ClickHouse(DB_SRV_FATAL)`

libtus/error.hh
```diff
@@ -43,60 +43,119 @@ namespace tus {
...
+	compose_code(uint16_t category_id, uint16_t enum_val) noexcept
+	{
+		return (category_id << 16) | enum_val;
+	}
+
+private:
+	int			code_;
+	uint16_t		enum_val_;
+	uint16_t		category_id_;
+	std::string_view	category_name_;
```
krizhanovsky, line 140: We have more than 8 bytes now in `Error`, so compiler needs to allocate some stack space to return a value for
```C++
return std::unexpected(ClickhouseError(DB_SRV_FATAL));
```
, which is less efficient than traditional C return code in RAX. The original main intention for `Error` (and described in the header comment) that we can get return code as efficient as in C. And now we loose the performance.

krizhanovsky (changes requested):
It seems we loose performance with the change

consuelo2210:
I will return to new proposal a little bit later

Issue #736, opened 2017-05-29
Move timers to kernel thread


krizhanovsky:
Currently, we do some calculations in timer basis and the calculations can be relatively expensive. Meantime, timer is serviced in the same softirq processing network traffic, so it can lead to packet loss. A separate kernel thread should be introduced to manage such non crucial logic (e.g. connection eviction and statistics).

krizhanovsky:
`tfw_apm_add_srv()` creates a timer for each server, so for scenario of #76 with thousands of servers it creates thousands of timers, so Tempesta FW being w/o any load eats CPU for 10K servers:
```
    26.64%  [tempesta_fw]  [k] tfw_apm_prcntl_tmfn
    16.04%  [kernel]       [k] __lock_acquire
     6.30%  [kernel]       [k] lock_release
     5.67%  [kernel]       [k] run_timer_softirq
     5.48%  [kernel]       [k] check_chain_key
```

A kernel thread should traverse the counters and sleep until timeout for the first updated APM elapses.

krizhanovsky:
The issue simply [hangs a server on 30K server groups](https://github.com/tempesta-tech/tempesta/issues/76#issuecomment-366849543).

krizhanovsky:
Note that #916 introduces multiple `_bh()` synchronization calls, e.g. https://github.com/tempesta-tech/tempesta/pull/916#discussion_r170275718 . With moving some work from softirq (timer) context to kernel threads, the synchronization can be relaxed to remove `_bh()`'s.

krizhanovsky:
#2599 limits the number of upstream servers by 1000, so this task is crucial to fix since there are many use cases with up to 10K upstream servers and can be more.

PR #2599, opened 2026-01-08
Mekhanik evgenii/fix 2588 1

EvgeniiMekhanik:


fw/sock_srv.c
```diff
@@ -605,6 +745,10 @@ tfw_sock_srv_abort_srv(TfwServer *srv)
 static int
 tfw_sock_srv_disconnect_srv(TfwServer *srv)
 {
+	srv->flags |= TFW_SRV_F_REMOVED;
+	smp_mb__after_atomic();
```
const-t: `srv->flags |= TFW_SRV_F_REMOVED` isn't safe. `or` isn't atomic on x86, you need to use `set_bit()`. `smp_mb__after_atomic()` is also optional here, it does nothing on x86, however Tempesta intended to work only on x86.

fw/sock_srv.c
```diff
@@ -483,7 +623,7 @@ tfw_sock_srv_disconnect(TfwConn *conn)
 		 * procedure, and server had not been put. See for details in
 		 * connection's destructor @tfw_srv_conn_release().
 		 */
-		if (del_timer_sync(&conn->timer)) {
+		if (tfw_srv_conn_del_timer_sync(srv_conn)) {
```
const-t, line 802: You changed the function, but not changed comment above it.

fw/sock_srv.c
```diff
@@ -605,6 +745,10 @@ tfw_sock_srv_abort_srv(TfwServer *srv)
 static int
 tfw_sock_srv_disconnect_srv(TfwServer *srv)
 {
+	srv->flags |= TFW_SRV_F_REMOVED;
+	smp_mb__after_atomic();
+	del_timer_sync(&srv->rc_timer);
```
const-t: It seems `timer_shutdown_sync()` is more suitable here, it guarantee that timer will not be re-armed and callback is finished when `timer_shutdown_sync()` is finished.

const-t (changes requested):
Reject for discuss. Current solution well done and solves the problem. But I have a question about approach. @krizhanovsky @EvgeniiMekhanik In this patch we have single timer per server, however we have array of timeouts and we try to re-establish the connection to **the same** server with the different timeouts. The main question - why we assume that some connection will be connected to the same server and some will not and we should try to retry for certain connection later? For example: We try to reconnect for the first connection, we tried few times, current interval set to 1sec. After that, the second connection being dropped and we try to reconnect with 1ms interval for the second connection and with 1sec interval for the first connection. It seems the chance to reconnect for them is the same, but we do re-connections with different interval? Is not enough simple probing logic: On disconnect trying to re-connect after one 1ms for 10 connections, if not connected increase interval and again try to re-connect for 10 connections and so on. If at least one connection being established reset interval. Also it seems with this approach we could use kernel lock-free list `llist.h` and use only `add/del_all` functions to do it lockless.

fw/sock_srv.c
```diff
@@ -122,6 +121,112 @@ tfw_srv_conn_stop(TfwSrvConn *srv_conn)
 	tfw_server_put((TfwServer *)srv_conn->peer);
 }
 
+static inline unsigned int
+tfw_srv_conn_total_recns_to_idx(unsigned int l1, unsigned int l2,
+				unsigned int l3)
+{
+	long long unsigned int curr_recns_n = atomic64_read(&total_recns_n);
+	int b = 64 - __builtin_clzll(curr_recns_n);
```
const-t: I didn't get the idea, why not `b >= 65536 + b >= 131072 + b >= 262144`? Values just for example.

fw/sock_srv.c
```diff
@@ -515,11 +514,12 @@ tfw_sock_srv_disconnect(TfwConn *conn)
 			if (test_bit(TFW_CFG_B_DEL, &srv->flags)) {
 				tfw_connection_abort(conn);
 			} else {
-				r = tfw_connection_close(conn, true);
+				if (tfw_connection_close(conn, true))
+					tfw_connection_abort(conn);
```
krizhanovsky: I'm not sure about this fix: `tfw_connection_close()` calls `ss_close()`, which is one of probable reasons to get non-zero return code here. And now we call `tfw_connection_abort()`, which also leads to `ss_close()`. If we fail in the first function, then we likely fail in the second one and issue a kernel warning.

fw/sock_srv.c
```diff
@@ -515,11 +514,12 @@ tfw_sock_srv_disconnect(TfwConn *conn)
...
 			}
 
 			tfw_connection_put(conn);
-			return r;
+			return 0;
```
krizhanovsky: Now the function should return `void` and `tfw_sock_srv_disconnect_srv()` as well. And `tfw_sock_srv_grace_shutdown_srv()`, and maybe other functions.

fw/sock_srv.c
```diff
@@ -341,15 +390,33 @@ tfw_sock_srv_connect_try(TfwSrvConn *srv_conn)
...
+	 * for the the entire server instead of one timer per connection,
+	 * the kernel can still hang in scenarios where we have thousands
+	 * of servers and tens of thousands of connections per server,
+	 * especially if the network is unstable and we attempt to
+	 * re-establish all those connections simultaneously.
+	 * To avoid this, we track the total number of reconnects and if
+	 * this number exceeds certain thresholds, we reduce the number
+	 * of attempts performed in a single timer callback invocation.
+	 */
+	return max_recns_per_time[tfw_srv_conn_total_recns_to_idx(11, 15, 17)];
```
krizhanovsky: Please explain the numbers in the comment

fw/sock_srv.c
krizhanovsky: `b` is bits, but it seems we can compare `total_recns_n` with the constants, which should make the code simpler.

fw/sock_srv.c
```diff
@@ -122,6 +121,55 @@ tfw_srv_conn_stop(TfwSrvConn *srv_conn)
...
+	 * depending on total reconnect count:
+	 * < 65,536 - commonly used case (array number 0)
+	 * 65,536 – 262,143
+	 * 262,144 – 1,048,575
+	 * >= 1,048,576
+	 */
+	unsigned int tmo_arr_idx = tfw_srv_conn_total_recns_to_idx(17, 19, 21);
+
+	return msecs_to_jiffies(tfw_srv_tmo_vals[tmo_arr_idx][idx]);
+}
```
krizhanovsky, line 212: Quite a good idea with the dynamic timeout change!

krizhanovsky (commented):
I didn't review the whole PR, but just 2 smallest patches. I love the idea, but some work is required to make the code clean - please do this before the next review.

I also agree with @const-t that
1. it makes sense to reconnect with small batches, like `mon(conns_n, 10)`,  to get some minimal concurrency and also not to overload the stack
2. if we issue 10 reconnects, that we can easily see 1 or several unsuccessful and others successful reconnections (e.g. a server is initializing the listening socket). It's OK to try ti reconnect the rest of connection immediately or after a timeout, but we do not need to treat the connections differently

fw/sock_srv.c
EvgeniiMekhanik: I remove this fix. In fact we have a problem in tfw_connection_abort that it can fails (under memory pressure_ and if it fail we catch BUG. We should think how to fix it.

fw/sock_srv.c
```diff
@@ -108,20 +109,272 @@
...
+} RecnsHistoryBucket;
+
+/*
+ * @buckets	- Circular array of history buckets storing per-interval
+ *		  data;
+ * @total	- total number of released connections over the current
+ *		  sliding time window (sum of all buckets);
+ */
+typedef struct {
+	RecnsHistoryBucket buckets[FRANG_FREQ];
```
const-t: Please redefine the constant `FRANG_FREQ` in this file with another name.

fw/server.h
```diff
@@ -38,38 +38,79 @@
 typedef struct tfw_srv_group_t TfwSrvGroup;
 typedef struct tfw_scheduler_t TfwScheduler;
 
+/*
+ * The number of  reconnection attempts during increasing timeout (quick
+ * reconnect) stage.
+ * This number is not included in the total count of reconnection attempts.
+ */
+#define TFW_SRV_TMO_NR		6
```
const-t, line 48: Comment is outdated, now we don't have `quick reconnect` stage, timeout always calculated using total count of reconnections.

fw/sock_srv.c
```diff
@@ -108,20 +109,272 @@
...
+ * timeout between reconnections.
+ */
+enum {
+	TFW_SRV_RECNT_RATE_L1 = 1000,
+	TFW_SRV_RECNT_RATE_L2 = 5000,
+	TFW_SRV_RECNT_RATE_L3 = 15000,
+};
+
+static inline void
+tfw_srv_conn_adjust_inc_recns(TfwSrvConn *srv_conn)
```
const-t:
```suggestion
tfw_srv_inc_recns_in_progress(TfwSrvConn *srv_conn)
```

fw/sock_srv.c
```diff
@@ -154,33 +407,31 @@ tfw_srv_conn_stop(TfwSrvConn *srv_conn)
...
-
-	if (srv_conn->recns < tfw_srv_tmo_nr) {
+	if (srv_conn->recns < TFW_SRV_TMO_NR) {
 		if (srv_conn->recns)
 			T_DBG_ADDR("Cannot establish connection",
 				   &srv_conn->peer->addr, TFW_WITH_PORT);
-		timeout = tfw_srv_tmo_vals[srv_conn->recns];
 	} else {
-		if (srv_conn->recns == tfw_srv_tmo_nr || !(srv_conn->recns % 60))
+		if (srv_conn->recns == TFW_SRV_TMO_NR || !(srv_conn->recns % 60))
```
const-t, line 389: After the changes `TFW_SRV_TMO_NR` is not reliable here, because we have 3 different levels of timeouts. Maybe we can repair/resched connection relying on timeout?

fw/sock_srv.c
```diff
@@ -108,20 +109,272 @@
...
+		{ 1000,  2000,  5000,  10000, 20000, 30000 },
+		{ 5000,  10000, 20000, 30000, 45000, 60000 }
+	};
+	unsigned int idx, tmo_arr_idx;
+	unsigned int timeout;
+	u64 rate;
+
+	assert_spin_locked(&srv->ctrl.recns_lock);
+
+	rate = tfw_srv_total_reconnection_count() >> 3;
```
const-t: magic number, the same in `tfw_srv_update_total_reconnection_count()`

fw/sock_srv.c
EvgeniiMekhanik: I think about it and remove srv->recns. But if count of reconnect is big, but we have only one reconnect attempt for current connection (timeout is big!, because total count of reconnect is big) should we try to reschedule it? It is new load on the system.

EvgeniiMekhanik: @krizhanovsky what do you think?

fw/sock_srv.c
```diff
@@ -446,21 +786,19 @@ tfw_sock_srv_disconnect(TfwConn *conn)
 	TfwSrvConn *srv_conn = (TfwSrvConn *)conn;
 
 	/*
-	 * Exit if connection is already stopping, or if it has never been
-	 * activated (due to some error in @sock_srv start procedure; so,
-	 * consequently, it will never gets to its destructor and will never
-	 * reach the stopped state).
+	 * Exit if connection has never been activated (due to some error
+	 * in @sock_srv start procedure; so, consequently, it will never
+	 * gets to its destructor and will never reach the stopped state).
```
krizhanovsky:
```suggestion
	 * get to its destructor and will never reach the stopped state).
```

fw/sock_srv.c
```diff
@@ -122,6 +121,55 @@ tfw_srv_conn_stop(TfwSrvConn *srv_conn)
...
+		{ 50, 250, 600, 2000, 3000, 4000}
+	};
+	/*
+	 * Calculate what `tfw_srv_tmo_vals` array should be used,
+	 * depending on total reconnect count:
+	 * < 65,536 - commonly used case (array number 0)
+	 * 65,536 – 262,143
+	 * 262,144 – 1,048,575
+	 * >= 1,048,576
+	 */
```
krizhanovsky: Why did you remove the two comments from https://github.com/tempesta-tech/tempesta/pull/2599/changes/dbf9311818214521b4caf9e312adc05dcceb52f1 explaining the hardcoded timeouts?

fw/sock_srv.c
```diff
@@ -108,20 +109,272 @@
...
+						      TFW_SRV_RECNT_RATE_L2,
+						      TFW_SRV_RECNT_RATE_L3);
+	idx = min(srv->ctrl.recns, TFW_SRV_TMO_NR - 1);
+
+	timeout = msecs_to_jiffies(tfw_srv_tmo_vals[tmo_arr_idx][idx]);
+	/*
+	 * Introduce random jitter to the timeout (0 .. timeout/4) to spread
+	 * reconnect attempts and prevent servers from retrying simultaneously.
+	 */
+	timeout += get_random_u32_below(max(timeout >> 2, 1));
```
krizhanovsky: It seems you're using true random generator, which is slow and has limited entropy. The system has limited entropy, which we need for example for security crucial computations on the TLS handshakes and it's not wise to waste it for such a case. If you need some differentiation, then you can use `jiffies`, current CPU id and so on - faster, doesn't need high quality random and doesn't waste entropy

fw/sock_srv.c
krizhanovsky: @EvgeniiMekhanik I believe we spend too much time on discussion and many review iterations for a problem which we never observed - we never saw thousands of reconnections form thousands of servers. I barely can imagine the case actually.

The original problem is about per-connection timers, but this PR tries to implement a very smart approach for massive reconnections from many servers.

The answer is: it's OK for now if it takes time to make all the reconnections. We still need to make at least several reconnection attempts before rescheduling. With this I want to minimize the changes (sticking to the current logic) and finally make the PR reach a merge.

fw/server.c
```diff
@@ -82,11 +82,15 @@ static DECLARE_RWSEM(sg_sem);
...
 {
+	WARN_ON(atomic_read(&srv->ctrl.recns_in_progress));
 	if (srv->cleanup)
 		srv->cleanup(srv);
 	/* Close all connections before freeing the server! */
+	BUG_ON(!list_empty(&srv->ctrl.recns_list));
+	BUG_ON(!list_empty(&srv->ctrl.failed_recns_list));
 	BUG_ON(!list_empty(&srv->conn_list));
 	BUG_ON(timer_pending(&srv->gs_timer));
+	BUG_ON(timer_pending(&srv->rc_timer));
```
krizhanovsky, line 98: We agreed to at least comment all new BUG statements - why it's necessary to crash.

What if we have some connections not closed for the server? If a memory leak, then it's better to print a warning and leak memory than to crash

fw/sock_srv.c
```diff
@@ -23,6 +23,7 @@
 #include <linux/net.h>
```
krizhanovsky, line 23: Please update the copyright year in all the touched files

fw/sock_srv.c
```diff
@@ -108,20 +109,272 @@
...
+}
+
+static inline void
+tfw_srv_conn_mod_timer(TfwSrvConn *srv_conn)
+{
+	TfwServer *srv = (TfwServer *)srv_conn->peer;
+	bool modify = false;
+	int r;
+
+	BUG_ON(!list_empty(&srv_conn->in_reconn_list));
```
krizhanovsky: This is *reconnection*, not the main data plane function. So why do we crash here? Why not to print warning and just exit from the function?

fw/sock_srv.c
```diff
@@ -108,20 +109,272 @@
...
+	unsigned int rc = 0;
+
+	/*
+	 * We set `__TFW_CONN_B_RECONNECT` during connection reestablishing
+	 * from reconnection timer callback and clear this bit if connection
+	 * successfully reestablished from `__reset_retry_timer` or if
+	 * reestablishing failed from `tfw_srv_conn_mod_timer`.
+	 */
+	WARN_ON(test_and_set_bit(__TFW_CONN_B_RECONNECT, &srv_conn->flags));
+	atomic_inc_return(&srv->ctrl.recns_in_progress);
```
krizhanovsky:
```suggestion
	rc = atomic_inc_return(&srv->ctrl.recns_in_progress);
```

fw/sock_srv.c
```diff
@@ -108,20 +109,272 @@
...
+ * is set when a reconnect attempt is initiated.
+ *
+ * If the flag is set, the counter is decremented and the updated value
+ * is returned. If the connection was not part of a reconnect attempt
+ * (e.g., a regular disconnect), the function returns -1.
+ *
+ * The return value is used to distinguish between these cases.
+ */
+static inline int
+tfw_srv_conn_adjust_dec_recns(TfwSrvConn *srv_conn)
```
krizhanovsky:
```suggestion
tfw_srv_dec_recns_in_progres(TfwSrvConn *srv_conn)
```

fw/sock_srv.c
```diff
@@ -108,20 +109,272 @@
...
 	set_bit(TFW_CONN_B_STOPPED, &srv_conn->flags);
 	tfw_server_put((TfwServer *)srv_conn->peer);
 }
 
+static inline unsigned int
+tfw_srv_conn_total_recns_to_idx(unsigned int sum, unsigned int l1,
+				unsigned int l2, unsigned int l3)
+{
+	return (sum >= l1) + (sum >= l2) +
+		(sum >= l3);
```
krizhanovsky:
```suggestion
	return (sum >= l1) + (sum >= l2) + (sum >= l3);
```

fw/sock_srv.c
```diff
@@ -108,20 +109,272 @@
...
+}
+
+/*
+ * Update per-CPU sliding-window reconnection counters.
+ *
+ * Rotates time buckets based on jiffies, resets stale bucket,
+ * and maintains total reconnection count.
+ */
+static inline void
+tfw_srv_update_total_reconnection_count(void)
```
krizhanovsky: There is naming inconsistency: `recns`, `reconnection`, `reconnections`. Please use only one (personally, I'd prefer shorter `recns`)

fw/sock_srv.c
```diff
@@ -108,20 +109,272 @@
...
+
+static inline u64
+tfw_srv_total_reconnection_count(void)
+{
+	int cpu;
+	u64 sum = 0;
+
+	for_each_possible_cpu(cpu) {
+		RecnsHistory *h = per_cpu_ptr(&recns_history, cpu);
+		sum += h->total;
```
krizhanovsky: Here we can sum outdated `total` values - imagine that we didn't hit `i`'th slot in `tfw_srv_update_total_reconnection_count()` for a while and now it participates in this accounting. Nothing, serious, but probably it makes sense to use a more efficient approach https://github.com/tempesta-tech/tempesta/issues/2615

fw/sock_srv.c
```diff
@@ -108,20 +109,272 @@
...
+
+	tfw_srv_update_total_reconnection_count();
+
+	spin_lock_bh(&srv->ctrl.recns_lock);
+
+	r = tfw_srv_conn_adjust_dec_recns(srv_conn);
+	switch (r) {
+	case 0:
+		/*
+		 * Last connection being reestablished from the timer callback.
```
krizhanovsky:
```suggestion
		 * Last connection is being reestablished from the timer callback.
```

fw/sock_srv.c
```diff
@@ -108,20 +109,272 @@
...
+		 * Reconnection attempt from the timer callback failed.
+		 * Add the connection to the head of the failed reconnection
+		 * list.
+		 * The timer will be adjusted later, after all connections
+		 * processed by the timer callback have either been successfully
+		 * reestablished or failed to reconnect.
+		 */
+		WARN_ON(r < 0);
+		list_add_tail(&srv_conn->in_reconn_list,
+			      &srv->ctrl.failed_recns_list);
```
krizhanovsky, line 304: One more names inconsistency: `reconn` and `recns`

fw/sock_srv.c
```diff
@@ -283,26 +534,115 @@ tfw_sock_srv_connect_try(TfwSrvConn *srv_conn)
...
+	spin_lock(&srv->ctrl.recns_lock);
+
+	if (unlikely(!tfw_srv_conn_adjust_dec_recns(srv_conn))) {
+		/*
+		 * If all connections processed by the reconnection timer
+		 * callback were successfully reestablished (i.e., no
+		 * failures remain), increase the number of connections to
+		 * be reestablished on the next timer callback invocation.
+		 */
+		if (list_empty(&srv->ctrl.failed_recns_list)) {
```
krizhanovsky, line 604: This is the only real usage of the list. The other one in `tfw_sock_srv_connect_retry_timer_cb()` is just concatination with `recns_list`. So it seems this list can be replaced with a boolean or a counter

fw/sock_srv.c
krizhanovsky: I think we should not change the number of reconnection attempts. With this patch, because of rate limiting of the reconnections, this could take longer, but it's OK if we really experience multiple upstream failures

krizhanovsky (changes requested):
I think the current approach with adaptive reconnections flow is good (at least I didn't find any issues with it), but the implementation has several issues to fix. Please make fixes in a separate new patch.

fw/sock_srv.c
EvgeniiMekhanik: Yes fixed.

fw/sock_srv.c
EvgeniiMekhanik: yes fixed.

fw/sock_srv.c
EvgeniiMekhanik: yes fixed

fw/server.c
EvgeniiMekhanik: If timer is pending during shutdown, timer callback can be called after Tempesta FW will be unloaded. We catch BUG in this case.

fw/sock_srv.c
EvgeniiMekhanik: yes fixed

fw/sock_srv.c
EvgeniiMekhanik: yes fixed

fw/sock_srv.c
EvgeniiMekhanik: yes fixed. move all this calculations to separate library

fw/server.h
EvgeniiMekhanik: Yes fixed

fw/sock_srv.c
EvgeniiMekhanik: yes fixed

fw/sock_srv.c
EvgeniiMekhanik: I move all sliding window calculation to separate library (same as described in 2615).

fw/sock_srv.c
EvgeniiMekhanik: We try to reestablish connection from this list at the end, So it seems that it is better to use list instead of boolean

EvgeniiMekhanik: We add all other connections to `recns_list`. When we try to reestablish connections we do
list_splice_tail_init(&srv->ctrl.recns_list, &recns_list);
list_splice_tail_init(&srv->ctrl.failed_recns_list, &recns_list);
so logic - all new disconnected connections we add to recns_list all connections which were failed to re-establish we add to failed_recns_list. Try ti reestablish connections from failed_recns_list only after connections from recns_list. If count of connection is big. We skip trying to re establish connections from failed_recns_list, because we exceeded the limit before!

fw/sock_srv.c
EvgeniiMekhanik: yes fixed

lib/sliding_window.h
```diff
@@ -0,0 +1,80 @@
...
+		prev = w->prev_count;
+	} else if (delta == 1) {
+		curr = 0;
+		prev = w->curr_count;
+	} else {
+		return 0;
+	}
+	age = now % window_secs;
+
+	return curr + prev * (window_secs - age) / window_secs;
```
const-t, line 77: I'm not 100% sure, but it seems if do so we can get rid of one division and modulo. But it is up to you.
```
static inline void
tfw_sliding_window_update(TfwSlidingWindow *w, u64 window_secs,
			  unsigned int count)
{
	u64 now = ktime_get_seconds();
	u64 curr_epoch = now;
	u64 delta;

	if (unlikely(!w->epoch)) {
		w->epoch = curr_epoch;
		w->curr_count = count;
		return;
	}

	delta = curr_epoch - w->epoch;
	if (delta < window_secs) {
		w->curr_count += count;
		return;
	}

	w->prev_count = (delta < (window_secs * 2) ? w->curr_count : 0);
	w->curr_count = count;
	w->epoch = w->epoch + window_secs;
}

static inline u64
tfw_sliding_window_get_total(TfwSlidingWindow *w, u64 window_secs)
{
	u64 now = ktime_get_seconds();
	u64 curr_epoch = now;
	u64 curr, prev, delta;

	if (unlikely(w->epoch == 0))
		return 0;

	delta = curr_epoch - w->epoch;
	if (delta < window_secs) {
		curr = w->curr_count;
		prev = w->prev_count;
	} else if (delta < (window_secs * 2)) {
		curr = 0;
		prev = w->curr_count;
	} else {
		return 0;
	}

	return curr + prev * (window_secs - delta) / window_secs;
}
```

fw/sock_srv.c
const-t: No randomness at this moment, comment outdated.

fw/sock_srv.c
```diff
@@ -552,28 +554,24 @@ tfw_sock_srv_calc_max_recns_per_time(TfwServer *srv)
 	};
 	unsigned int idx = min(srv->ctrl.recns_batch_idx,
 			       ARRAY_SIZE(max_recns_per_time) - 1);
-	unsigned max_recns;
-
-	max_recns = max_recns_per_time[idx];
-	max_recns -= get_random_u32_below(max_recns / 5 + 1);
```
const-t: I though we decided to use something like this `rnd = (jiffies + cpu_id) % 5;`, no?

fw/sock_srv.c
```diff
@@ -23,6 +23,7 @@
 #include <linux/net.h>
 #include <linux/wait.h>
 #include <linux/freezer.h>
+#include <linux/random.h>
```
const-t: random is not used

lib/sliding_window.h
EvgeniiMekhanik: No in my code all time is divided into same time slots.  And we can exactly say in which slot we're currently in. In your scenario, a slot and a half could pass between events, and we'd start the new slot from the wrong place.

const-t: Could we just do it this way? Time slot will be updated to the current time and also will be divided into `widnow_secs` slots.

```
...
	w->prev_count = (delta < (window_secs * 2) ? w->curr_count : 0);
	w->curr_count = count;
	w->epoch = curr_epoch;
```

Or if we need the same size for time slot we could do something like this
```
u64 max_epoch = (window_secs * 2);
	if (delta < max_epoch) {
		w->prev_count = w->curr_count;
		w->curr_count = count;
		//w->epoch = w->epoch + delta - (delta - window_secs); OR
        w->epoch = curr_epoch - (delta - window_secs)
	} else {
		w->prev_count = 0;
		w->epoch = curr_epoch;
	}
```
I'm not sure that it makes sense I didn't watch asm code for both of them and didn't profile. Just a suggestion.

const-t (approved):
LGTM

lib/sliding_window.h
EvgeniiMekhanik: During private discussion we decide to choose current version of code

scripts/tempesta.sh
```diff
@@ -19,6 +19,8 @@
 # this program; if not, write to the Free Software Foundation, Inc., 59
```
krizhanovsky, line 19: please update the copyright year

lib/sliding_window.h
```diff
@@ -0,0 +1,80 @@
+/**
+ *		Tempesta kernel library
```
krizhanovsky, line 2: Please write a short description what the file implements and where it's used, describe the algorithm, discuss alternatives (leaking buckets, Frang rate limits). Also please reference the blog post https://blog.cloudflare.com/counting-things-a-lot-of-different-things/

lib/sliding_window.h
```diff
@@ -0,0 +1,80 @@
...
+ * Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ */
+#ifndef __LIB_SLIDING_WINDOW_H__
+#define __LIB_SLIDING_WINDOW_H__
+
+typedef struct {
+	u64 curr_count;
+	u64 prev_count;
+	u64 epoch;
+} TfwSlidingWindow;
```
krizhanovsky, line 71: Please comment the fields and their relations

lib/sliding_window.h
krizhanovsky: I also have the same question - why do we do division (the heaviest instruction!) twice. Please at least comment in the code.

Please make sure that there is only one `div` generated by the compiler (I assume it can optimize `now` division by `window_secs`) for `now / window_secs` and `now % window_secs`.

fw/sock_srv.c
```diff
@@ -111,32 +112,12 @@
...
-	u64 total;
-} RecnsHistory;
-
 /*
  * Per-CPU storage of connection release history aggregated
  * across all servers.
  */
-static DEFINE_PER_CPU(RecnsHistory, recns_history);
+static DEFINE_PER_CPU(TfwSlidingWindow, recns_history);
+static unsigned int recns_history_window = 5;
```
krizhanovsky, line 138: Please comment the constant 5 - why 5 and is it seconds?

fw/server.c
```diff
@@ -101,12 +116,35 @@ tfw_server_destroy(TfwServer *srv)
...
 }
 
 TfwServer *
 tfw_server_create(const TfwAddr *addr)
 {
-	TfwServer *srv = kmem_cache_alloc(srv_cache, GFP_KERNEL | __GFP_ZERO);
+	TfwServer *srv;
+
+	if (!tfw_sg_check_srv_cnt())
+		return NULL;
```
krizhanovsky: In this case we should not just report ""Error handling the server". Let's handle this with a separate error message referencing the #736 issue.

fw/sock_srv.c
krizhanovsky: Oh, finally got it, but it's not obvious from the code. Please add a comment that we need the separate list `failed_recns_list` to place all its members strictly after `recns_list` for reconnection.

krizhanovsky (approved):
There are still places for cleanups and improvements. Let's make one more iteration and finally merge it.

@EvgeniiMekhanik could you please fix my comments and after that @const-t please do one more review, hopefully the final one.

lib/sliding_window.h
EvgeniiMekhanik: yes fixed

fw/sock_srv.c
EvgeniiMekhanik: yes fixed

scripts/tempesta.sh
EvgeniiMekhanik: yes fixed

lib/sliding_window.h
EvgeniiMekhanik: yes fixed

fw/server.c
EvgeniiMekhanik: yes fixed

const-t (approved):
LGTM

PR #2604, opened 2026-01-09
Add warning in case of `dmesg` log usage in release

EvgeniiMekhanik:
Using `dmesg` log option can lead to kernel hung under heavy load. Add to warn about it when Tempesta starts.

### DIFF
--- a/fw/access_log.c
+++ b/fw/access_log.c
@@ -551,6 +551,10 @@ cfg_access_log_set(TfwCfgSpec *cs, TfwCfgEntry *ce)
 		if (strcasecmp(val, "off") == 0) {
 			off = true;
 		} else if (strcasecmp(val, "dmesg") == 0) {
+#ifndef DEBUG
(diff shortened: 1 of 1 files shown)


fw/access_log.c
```diff
@@ -551,6 +551,10 @@ cfg_access_log_set(TfwCfgSpec *cs, TfwCfgEntry *ce)
 		if (strcasecmp(val, "off") == 0) {
 			off = true;
 		} else if (strcasecmp(val, "dmesg") == 0) {
+#ifndef DEBUG
+			T_WARN("Using access log `dmesg` can lead to kernel"
+			       "hung under load.\n");
```
krizhanovsky:
```suggestion
			T_WARN("Using the access log via `dmesg` has significant performance"
						       " overhead and may lead to kernel hangs under high loa\n");
```

krizhanovsky (approved):
LGTM

PR #2610, opened 2026-01-20
Mekhanik evgenii/fix potential bug master

EvgeniiMekhanik:


fw/ss_skb.c
```diff
@@ -1310,7 +1310,8 @@ ss_skb_split(struct sk_buff *skb, int len)
 	 * Initialize GSO segments counter to let TCP set it according to
 	 * the current MSS on egress path.
 	 */
-	tcp_skb_pcount_set(skb, 0);
+	if (skb_tfw_is_present(skb))
```
krizhanovsky:
```suggestion
	if (!skb_tfw_is_present(skb))
```

The commit message explains: "We should not call `tcp_skb_pcount_set` during skb splitting", but the code sets `skb->cb` (BTW not `sb` as in the commit message) IF `present`

fw/http2.c
```diff
@@ -647,7 +647,7 @@ tfw_h2_stream_xmit_prepare_resp(TfwStream *stream)
 	BUG_ON(!resp || resp->msg.skb_head || !resp->req
 	       || !resp->req->conn || !stream->xmit.skb_head);
 
-	tls_type = skb_tfw_tls_type(stream->xmit.skb_head);
+	tls_type = TFW_SKB_CB(stream->xmit.skb_head)->tls_type;
```
krizhanovsky: I saw this change in another PR, but there is still no users for `skb->dev` - why do we need this change?

const-t: The same question. I didn't get this commit.

const-t: The same question. It was done to be able to check who(tempesta or tcp stack) owns skb with `skb_tfw_is_present()`? In this case we have flag that migrates, very subtle.

fw/ss_skb.c
EvgeniiMekhanik: `We should not call `tcp_skb_pcount_set` during skb splitting, if skb belongs to Tempesta` Now (after previous patch with `tls_type`!) we don't set `tls_type` in skb->dev until skb will be pushed to socket write queue, so `skb_tfw_is_present` will return false, until skb will be pushed to socket write queue. We should not call `tcp_skb_pcount_set` because it has no sence, moreover since we use skb->cb in our code this call lead to memory corruption!. After skn will be pushed to socket write queue `tls_type` will be set, so `skb_tfw_is_present` will return true. In this case we should call `tcp_skb_pcount_set` (there is only one case for it, when skb_split is called from tls_encrypt)

fw/http2.c
EvgeniiMekhanik: Yes to be sure that skb_tfw_is_present return false

krizhanovsky (approved):
LGTM

PR #2613, opened 2026-01-29
Trap `SIGINT` during start/stop Tempesta FW

EvgeniiMekhanik:
There are a lot of cases when system remaning in
incorrect state, when tempesta.sh script interrupted and stopped by SIGINT. So trap signal at the beginning of script to prevent it's termination by Ctrl + C

### DIFF
--- a/scripts/tempesta.sh
+++ b/scripts/tempesta.sh
@@ -19,6 +19,10 @@
 # this program; if not, write to the Free Software Foundation, Inc., 59
 # Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 
+set +e
+trap '' SIGINT
+trap 'trap - SIGINT' EXIT
+
 if [ -z "$TFW_SYSTEMD" ]; then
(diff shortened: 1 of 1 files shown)


scripts/tempesta.sh
```diff
@@ -19,6 +19,10 @@
 # this program; if not, write to the Free Software Foundation, Inc., 59
 # Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 
+set +e
+trap '' SIGINT
+trap 'trap - SIGINT' EXIT
```
krizhanovsky, line 24: I just have a concern wouldn't we have hang-like behavior of the script for users?

EvgeniiMekhanik: If we interrupt the script during Tempesta's startup process due to a hang (as is currently the case), the system will remain in an invalid state and will likely require a reboot.

const-t: The same for me, can we initiate tempesta stop on SIGINT?

PR #2617, opened 2026-02-06
FIx memory corruption during cofiguration

EvgeniiMekhanik:
`tfw_cfg_set_mem` set unsigned long value by pointer, not unsigned int, so we should use unsigned long size as a target value, otherwise the next value (`lru_size`) will be overwritten.


const-t (approved):
LGTM

Issue #2556, opened 2025-11-14
Warning on Tempesta restart


krizhanovsky:
A baremetal Tempesta FW setup with plenty of
```
[tempesta fw] Warning: client 1.2.3.4 requested missing server name, reject connection
```
messages in dmesg. On Tempesta FW restart I got
```
[624926.598351] [tempesta fw] Warning: pending active connections for 5s (connections count 0xd7, queues count 0)
[624926.598779] [tempesta fw] Warning:   cpu 0(3), backlog size 0, active connections mask 0x56700000000, cntwork queue size 0, close backlog is empty
[624926.599660] [tempesta fw] Warning:   cpu 1(3), backlog size 0, active connections mask 0x85300000000, cntwork queue size 0, close backlog is empty
[624926.600650] [tempesta fw] Warning:   cpu 2(3), backlog size 0, active connections mask 0x1f000000000, cntwork queue size 0, close backlog is empty
[624926.601751] [tempesta fw] Warning:   cpu 3(3), backlog size 0, active connections mask 0xfffffb6c00000000, cntwork queue size 0, close backlog is empty
[624926.602957] [tempesta fw] Warning:   cpu 4(3), backlog size 0, active connections mask 0xfffffbdf00000000, cntwork queue size 0, close backlog is empty
[624926.604358] [tempesta fw] Warning:   cpu 5(3), backlog size 0, active connections mask 0xfffffe4f00000000, cntwork queue size 0, close backlog is empty
[624926.605690] [tempesta fw] Warning:   cpu 6(3), backlog size 0, active connections mask 0xffffff5a00000000, cntwork queue size 0, close backlog is empty
[624926.607141] [tempesta fw] Warning:   cpu 7(3), backlog size 0, active connections mask 0xfffffc3900000000, cntwork queue size 0, close backlog is empty
[624931.738348] [tempesta fw] Warning: pending active connections for 5s (connections count 0x69, queues count 0)
[624931.739183] [tempesta fw] Warning:   cpu 0(3), backlog size 0, active connections mask 0x55500000000, cntwork queue size 0, close backlog is empty
[624931.740860] [tempesta fw] Warning:   cpu 1(3), backlog size 0, active connections mask 0x84400000000, cntwork queue size 0, close backlog is empty
[624931.742590] [tempesta fw] Warning:   cpu 2(3), backlog size 0, active connections mask 0x1e100000000, cntwork queue size 0, close backlog is empty
[...]

PR #2619, opened 2026-02-12
Fix removing still used clients.

EvgeniiMekhanik:


fw/client.c
```diff
@@ -57,73 +57,111 @@ typedef struct {
...
  */
 static void
-tfw_client_update_lru(unsigned long key)
+tfw_client_update_lru(TfwClient *cli)
 {
-	if (list_empty(&client_lru.freelist)) {
+	assert_spin_locked(&client_db->ga_lock);
+
+	if (client_lru.lru_size >= client_cfg.lru_size) {
 		/* Free list is empty, get the last from LRU list. */
```
const-t: The comment is outdate, please change it.

fw/client.c
```diff
@@ -57,73 +57,111 @@ typedef struct {
...
-		last->key = key;
-		list_add(&last->list, &client_lru.head);
+		/*
+		 * Since we check it under `db->ga_lock` and getting new client
+		 * in `tdb_rec_get_alloc` is also made under `db->ga_lock`, we
+		 * can safely remove client entry in db.
+		 * If client will be removed here, new record will be created
+		 * during establishing new connection.
+		 */
+		if (tfw_client_has_last_ref(last)) {
```
const-t:
```suggestion
			tdb_entry_remove(client_db, last->key, NULL, NULL,
                false);
```

fw/client.c
```diff
@@ -57,73 +57,111 @@ typedef struct {
...
-						     TfwClientLRU, list);
-
-		list_del_init(&new->list);
-		new->key = key;
-		list_add(&new->list, &client_lru.head);
+		list_add(&cli->list, &client_lru.head);
+		client_lru.lru_size++;
 	}
+
+	tfw_client_remove_delayed(10);
```
const-t: I'd like to have it is as constant.

fw/client.h
```diff
@@ -34,6 +34,8 @@
 typedef struct {
 	TFW_PEER_COMMON;
 	TfwClassifierPrvt	class_prvt;
+	struct list_head	list;
+	unsigned long		key;
```
const-t: We can get rid off the `key`, just cast `TfwClient` to `TdbFRec` as you already do `TdbFRec *rec = ((TdbFRec *)curr) - 1;` and use its `key`.

fw/sock_clnt.c
```diff
@@ -88,8 +96,12 @@ tfw_sock_cli_keepalive_timer_cb(struct timer_list *t)
 	 * a deadlock on del_timer_sync(). In case of error try to close
 	 * it one second later.
 	 */
-	if (tfw_connection_close((TfwConn *)cli_conn, false))
-		mod_timer(&cli_conn->timer, jiffies + msecs_to_jiffies(1000));
+	if (TFW_CONN_TYPE(conn) & Conn_Closing) {
+		tfw_connection_abort(conn);
```
EvgeniiMekhanik, line 95: @krizhanovsky please pay attention on comment before. I don't understand idea about deadlock. What does it mean  asynchronously in current context?

fw/client.c
```diff
@@ -57,73 +57,122 @@ typedef struct {
...
+				      list)
+	{
+		if (tfw_client_has_last_ref(curr)) {
+			TdbFRec *rec = ((TdbFRec *)curr) - 1;
+
+			list_del_init(&curr->list);
+			tdb_entry_remove(client_db, rec->key, NULL,
+					 NULL, false);
+		}
+		if (!(budget--))
```
krizhanovsky:
```suggestion
		if (--budget > 0)
```

And I'd suggest safer `int` for `budget` type

db/core/htrie.c
```diff
@@ -231,6 +231,12 @@ tdb_htrie_put_rec(TdbHdr *dbh, TdbRec *rec)
 		tdb_htrie_free_rec(dbh, rec);
 }
 
+bool
+tdb_htrie_rec_has_last_ref(TdbRec *rec)
+{
+	return atomic_read(&rec->refcnt) == 1;
+}
```
krizhanovsky, line 239: Do we need to have is in the TDB module, isn't it better just to declare the function in htrie.h and inline it wherever needed? I don't see a point in the far call

fw/client.c
```diff
@@ -57,73 +57,122 @@ typedef struct {
...
-						     TfwClientLRU, list);
-
-		list_del_init(&new->list);
-		new->key = key;
-		list_add(&new->list, &client_lru.head);
+		list_add(&cli->list, &client_lru.head);
+		client_lru.lru_size++;
 	}
+
+	tfw_client_remove_delayed(COUNT_OF_CLIENTS_TO_DELETE);
```
krizhanovsky: `COUNT_OF_CLIENTS_TO_DELETE` is actually number of clients to scan - we can scan 10 clients and not delete any of them.

fw/client.c
```diff
@@ -57,73 +57,122 @@ typedef struct {
...
-	struct list_head	list;
-	unsigned long		key;
-} TfwClientLRU;
-
 static struct {
 	struct list_head	head;
-	struct list_head	freelist;
-	TfwClientLRU		*mem;
+	struct list_head	for_delay_delete;
+	struct list_head	*curr;
```
krizhanovsky: It's time to comment the structure members - the algorithm around `curr` is questionable since we have to scan probably millions of records, especially under DDoS.

Why don't we remove entries automatically, when TDB reaches zero reference count? We can return a refcounter value from `tdb_put_blk()` and handle it in `tfw_client_put()`. Or register a callback, called by TDB on reaching zero value.

fw/client.c
```diff
@@ -57,73 +57,122 @@ typedef struct {
...
 
 		list_del_init(&last->list);
-		tdb_entry_remove(client_db, last->key, NULL, NULL, false);
-		last->key = key;
-		list_add(&last->list, &client_lru.head);
+		/*
+		 * Since we check it under `db->ga_lock` and getting new client
+		 * in `tdb_rec_get_alloc` is also made under `db->ga_lock`, we
+		 * can safely remove client entry in db.
+		 * If client will be removed here, new record will be created
```
krizhanovsky:
```suggestion
		 * If client is removed here, new record will be created
```

fw/client.c
```diff
@@ -57,73 +57,122 @@ typedef struct {
...
 
 static void
 tfw_client_free_lru(void)
 {
-	int order = get_order(sizeof(TfwClientLRU) * client_cfg.lru_size);
+	spin_lock_bh(&client_db->ga_lock);
+
+	client_lru.curr = &client_lru.for_delay_delete;
+	tfw_client_remove_delayed(ULONG_MAX);
+	BUG_ON(!list_empty(&client_lru.for_delay_delete));
```
krizhanovsky: If only one entry in `for_delay_delete` list has reference counter larger than 1, `tfw_client_remove_delayed()` just reached the list head in `list_for_each_entry_safe_from()` and we crash on the `BUG_ON()`. It seems we need some synchronization here, probably with putting `ga_lock` and waiting for some time.

Also #2556 is about warning, but this path introduces `BUG_ON()`. @EvgeniiMekhanik @const-t be very careful with `BUG()` statements. The kernel coding style also discusses this. Maybe we should make a rule to always comment all `BUG()` statement with explanation why we have to crash.

fw/sock.c
```diff
@@ -1279,6 +1279,14 @@ ss_tcp_state_change(struct sock *sk)
 		sock_set_flag(sk, SOCK_TEMPESTA);
 		ss_active_guard_exit(SS_V_ACT_NEWCONN);
 	}
+	else if (sk->sk_state == TCP_FIN_WAIT2) {
+		/*
+		 * Set keepalive timeout to 10 seconds for forcely connection
+		 * closing if TCP FIN will not received.
```
krizhanovsky: Why do we need https://github.com/tempesta-tech/tempesta/pull/2619/changes/ab4607582449399d33db3e21ce6d19d6a7ff0422 patch at all - we already set  `tcp_fin_timeout` sysctl to 10 in `tempesta.sh`?

If we really need this, then please write a comment here or somewhere else about this and also with the commit message - we must have the behavior well documented.

fw/sock_clnt.c
```diff
@@ -76,10 +76,18 @@ tfw_cli_cache(int type)
 	}
```
krizhanovsky, line 76: Please update the copyright year in all the changed files

fw/sock_clnt.c
krizhanovsky: Apparently "asynchronously" means `sync = false` argument value for `tfw_connection_close()`. I think, in this function we're under a lock (likely socket lock - I didn't check) and can not call TCP closing under it, so we put a closing job into the queue.

This means that it seems calling `tfw_connection_abort()` is wrong. I didn't get deeper into the locking issue since I'm not sure about the commit idea.

EvgeniiMekhanik: I look inside `tfw_connection_close` and see that we never call  `ss_do_close` inside it. If flag is true we call `ss_turnstile_push` if `ss_wq_push` failed. Inside `ss_turnstile_push` we use `ss_cbacklog_cache` but still do real closing procedure in `ss_tx_action`. So I don't see any problem her

fw/client.c
EvgeniiMekhanik: We discuss about it with @const-t. Make changes in TDB here is better, but I implement current solution and add comment in commit message about it.
'''
Also need to add a comment why we implement such solution,
instead of make changes in TDB - during fixing https://github.com/tempesta-tech/tempesta/issues/1346 issue
(currently in progress) we face with the problem of performance
degradation because of a lot of new calls tfw_client_get/put.
To fix this problem we impelent the same `for_delay_delete`
list and do remove these extra `tfw_client_get/put` calls.
'''
In 1396 we adjust memory of all skbs (which belongs to requests and responses using client structure). We can't delete client untill all such skbs will be freed. So I already implement such list in appropriate PR. May be we can implement here another solution and try (if it is possible to avoid list in 1396) If no we still should implement this list. Need to discuss

EvgeniiMekhanik: Also during work on patch (trainging mode) it is necessary to implement a lot of atomics. May be if we can't avoid it we can remove list

fw/client.c
EvgeniiMekhanik: It can't have  reference counter larger than 1. In this case we have memory leak and breaking logic thats why I add BUG_ON

EvgeniiMekhanik: This function is called from cleanup procedure after all stop procedures were called. So all connections should be closed and reference counter should be equal to 1

fw/sock.c
EvgeniiMekhanik: tcp_fin_timeout works only for orphaned sockets (which were closed not shutdowned!). In case when we call tcp_shutdown we should handle this case manually if wan't to close such sockets

fw/sock_clnt.c
krizhanovsky: Yeah, the thing is that now we always do asynchronous closing in context of this comment

fw/client.c
krizhanovsky: > Also during work on patch (trainging mode) it is necessary to implement a lot of atomics. May be if we can't avoid it we can remove list

That's bad - probably we need per-cpu algorithms for accounting. Please make sure that there are no significant degrabations on the multi-core SUT setup before publishing 1346 implementation for review

db/core/tdb.h
```diff
@@ -53,6 +53,7 @@ typedef struct {
 #define TDB_REC_COMMON			\
 	unsigned long	key;		\
 	unsigned int	flags;		\
+	void		*backet;	\
```
krizhanovsky:
```suggestion
	void		*bucket;	\
```

krizhanovsky: Why do we need this? IIRC buckets are aligned and we can get a bucket header pointer by an offset from the stored in-place record or an offset record for large records

db/core/htrie.c
```diff
@@ -221,21 +222,25 @@ tdb_htrie_get_rec(TdbRec *rec)
...
-	if (!refcnt)
-		tdb_htrie_free_rec(dbh, rec);
-}
-
 /* Call only under lock. */
 static void
 tdb_rec_set_remove(TdbRec *rec)
 {
 	rec->flags |= TDB_HTRIE_REC_REMOVED_BIT;
+	rec->backet = NULL;
```
krizhanovsky: Do we still need `TDB_HTRIE_REC_REMOVED_BIT`? In the new logic, it seems not-null pointer is the indicator of removed record. Also it seems we don't check the pointer when we lookup a record, so technically we can lookup and work with the record

db/core/htrie.c
```diff
@@ -221,21 +222,25 @@ tdb_htrie_get_rec(TdbRec *rec)
...
+}
+
+static inline bool
+tdb_put_req_eq_cb(TdbHdr *dbh, TdbRec *rec, void *data)
+{
+	return rec == data;
+}
+
+static inline bool
+tdb_records_are_not_small(TdbHdr *dbh)
```
krizhanovsky: "small" -> fixed size

db/core/tdb.h
EvgeniiMekhanik: We can't do it after https://github.com/tempesta-tech/tempesta/commit/d5b511021d9c1f5f26e9d3e1b8e48a534c93e9f9#diff-398142a2301a5a6c230c34a38413d166ead6433ee19101893b834f2385330907

db/core/htrie.c
EvgeniiMekhanik: we use it only here

db/core/htrie.c
EvgeniiMekhanik: Remove this bit as unnecessary

db/core/htrie.c
EvgeniiMekhanik: fixed

db/core/htrie.c
```diff
@@ -223,30 +223,95 @@ tdb_htrie_get_rec(TdbRec *rec)
...
 }
 
+static inline bool
+tdb_put_req_eq_cb(TdbHdr *dbh, TdbRec *rec, void *data)
+{
+	return rec == data;
+}
+
+static inline bool
+tdb_records_are_not_fixed_size(TdbHdr *dbh)
```
const-t: `tdb_records_are_not_fixed_size()` this name doesn't show the real sense of the function. This `TDB_HTRIE_RALIGN(dbh->rec_len) >= TDB_HTRIE_MINDREC` condition may be true even for fixed size record. If someone decide to use this function to distinguish types of the record(fixed vs variable), we will have a bug. It is really small records, variable size records also might be small. The previous name was correct.

fw/client.c
```diff
@@ -76,54 +70,64 @@ static TDB *client_db;
...
+	TfwClientEntry *ent = (TfwClientEntry *)rec->data;
+	TfwClient *cli = &ent->cli;
 
-	for (i = 0, block = client_lru.mem; i < client_cfg.lru_size; i++)
-		list_add(&block[i].list, &client_lru.freelist);
+	/*
+	 * Client always has extra reference counter, so
+	 * it can be deleted only when client_lru.lru_size is
+	 * exceeded from `tfw_client_update_lru` or when
+	 * Tempesta FW shutdowned from `tfw_client_free_lru`
```
krizhanovsky:
```suggestion
	 * Tempesta FW shut down from `tfw_client_free_lru`
```

fw/connection.c
```diff
@@ -148,13 +148,37 @@ tfw_connection_release(TfwConn *conn)
  *   -ENOMEM	- out-of-memory error occurred.
  */
 int
-tfw_connection_send(TfwConn *conn, TfwMsg *msg)
+tfw_connection_send(TfwConn *conn, TfwMsg *msg, bool *was_shutdowned)
```
krizhanovsky:
```suggestion
tfw_connection_send(TfwConn *conn, TfwMsg *msg, bool *was_shutdown)
```

"Shut" (and "shut down") is an irregular verb with "shut" in all the 3 forms

fw/connection.c
```diff
@@ -148,13 +148,37 @@ tfw_connection_release(TfwConn *conn)
...
 	 */
-	return TFW_CONN_HOOK_CALL(conn, conn_send, msg);
+	r = TFW_CONN_HOOK_CALL(conn, conn_send, msg);
+	if (unlikely(r)) {
+		*was_shutdowned = false;
+		return r;
+	}
+
+	*was_shutdowned = (ss_flags & SS_F_CONN_CLOSE) &&
+		!(ss_flags & __SS_F_FORCE);
```
krizhanovsky: please align this line

fw/connection.c
```diff
@@ -148,13 +148,37 @@ tfw_connection_release(TfwConn *conn)
...
+	if (*was_shutdowned) {
+		BUG_ON(!(TFW_CONN_TYPE(conn) & Conn_Clnt));
```
krizhanovsky: Why do we have to crash here?

fw/connection.c
```diff
@@ -148,13 +148,37 @@ tfw_connection_release(TfwConn *conn)
...
+		/*
+		 * Socket that was closed by `tcp_shutdown` is not orphaned.
+		 * After receiving ack from remote peer, such socket moved
+		 * to TCP_FIN_WAIT2 state and can stay in this state unlimited
+		 * time (`tcp_fin_timeout` doesn't work for such sockets).
+		 * To prevent such situation set connection keepalive timer
+		 * to ten seconds and abort connection if this time is expired.
+		 */ 
+		tfw_cli_conn_mod_timer((TfwCliConn *)conn, 10);
```
krizhanovsky: At this moment we just pushed a work for a ring buffer and did not actually change the socket state, it is not shut down at this moment, so can we have a race with the timer? Having 10 seconds timeout _looks_ safe, but we still need to be sure. Also it'd be good to make 10 as a named constant and make a good reasoning comment about this exact value, especially having that we can't change it in runtime - this must be very robust.

Also we use the RB to reduce contention, but now we're acquiring lock on many CPUs.

I think this should go to the context of actual socket closing, i.e. be called from `ss_do_close()`. Of course not directly no not to break abstractions, but probably as a call back.

One more design question. We decided to get rid of the timer, since timers have high overhead, but this patch adds one more usage of the timer. Why can't we manage a per-CPU queue for such sockets and add a socket to the queue in `ss_do_send()` (also maybe with a callback). The only one CPU will manage the queue and once it has something on a socket it moves the socket to the end of the queue. Once in a while `ss_tx_action()` can check the sockets from the head of the queue.

db/core/htrie.c
```diff
@@ -221,28 +223,93 @@ tdb_htrie_get_rec(TdbRec *rec)
...
+		bckt->rec = 0;
+}
+
+void
+tdb_htrie_put_rec(TdbHdr *dbh, TdbRec *rec)
+{
+	int refcnt = atomic_dec_return(&rec->refcnt);
+
+	BUG_ON(refcnt < 0);
+	if (!refcnt) {
```
krizhanovsky:
```suggestion
	if (refcnt)
	    return;
```

db/core/htrie.c
```diff
@@ -221,28 +223,93 @@ tdb_htrie_get_rec(TdbRec *rec)
...
+		TdbBucket *bckt = rec->bucket;
+		tdb_eq_cb_t *eq_cb = tdb_put_req_eq_cb;
+
+		/*
+		 * There are two different strategies of record deletion:
+		 * - Record can be removed at any time (bckt is equal to zero),
+		 *   but still alive, because there are any active users. In
+		 *   this case, we just delete record here (this strategey
+		 *   ised in cache).
```
krizhanovsky:
```suggestion
		 *   is used in the web cache).
```

db/core/htrie.c
krizhanovsky: Usually, it's better not to overuse negative predicates - I'd suggest to rename it to `tdb_records_are_fixed_size()`

db/core/htrie.c
```diff
@@ -221,28 +223,93 @@ tdb_htrie_get_rec(TdbRec *rec)
...
+
+	BUG_ON(refcnt < 0);
+	if (!refcnt) {
+		TdbBucket *bckt = rec->bucket;
+		tdb_eq_cb_t *eq_cb = tdb_put_req_eq_cb;
+
+		/*
+		 * There are two different strategies of record deletion:
+		 * - Record can be removed at any time (bckt is equal to zero),
+		 *   but still alive, because there are any active users. In
```
krizhanovsky:
```suggestion
		 *   but still alive, because there are some active users. In
```

db/core/htrie.c
```diff
@@ -221,28 +223,93 @@ tdb_htrie_get_rec(TdbRec *rec)
...
+		 *   this case, we just delete record here (this strategey
+		 *   ised in cache).
+		 * - Record was not removed, but all active users gone. here.
+		 *   Remove record before deletion. (this strategy used for
```
krizhanovsky:
```suggestion
		 *   Mark the record as removed record before actual deletion (this strategy used for
```

db/core/htrie.c
```diff
@@ -221,28 +223,93 @@ tdb_htrie_get_rec(TdbRec *rec)
...
+		TdbBucket *bckt = rec->bucket;
+		tdb_eq_cb_t *eq_cb = tdb_put_req_eq_cb;
+
+		/*
+		 * There are two different strategies of record deletion:
+		 * - Record can be removed at any time (bckt is equal to zero),
+		 *   but still alive, because there are any active users. In
+		 *   this case, we just delete record here (this strategey
+		 *   ised in cache).
+		 * - Record was not removed, but all active users gone. here.
```
krizhanovsky:
```suggestion
		 * - Record was not removed, but all active users gone.
```

db/core/tdb.h
```diff
@@ -53,6 +53,7 @@ typedef struct {
 #define TDB_REC_COMMON			\
 	unsigned long	key;		\
 	unsigned int	flags;		\
+	void		*bucket;	\
```
krizhanovsky, line 55: It's better to move the pointer betweeb `key` and `flags` to avoid memory hole

db/core/htrie.c
krizhanovsky: With the new `tdb_htrie_put_rec()` there is a lot of code executed twice. Please update this place and other places calling `tdb_htrie_put_rec()`.

db/core/main.c
```diff
@@ -220,7 +223,11 @@ tdb_rec_next(TDB *db, TdbIter *iter)
 }
 EXPORT_SYMBOL(tdb_rec_next);
 
-/* Decrements reference counter. */
+/**
+ * Decrements reference counter. If reference counter became
+ * equal to zero and record is still alive (was not removed)
+ * remove it, before deletion.
```
krizhanovsky: "remove before delete" is confusing, please update the comment

fw/connection.c
EvgeniiMekhanik: For server connections this timer serves for other purposes. We never use shut down for server connections.

db/core/tdb.h
EvgeniiMekhanik: Yes fixed

db/core/htrie.c
EvgeniiMekhanik: The name should be `tdb_records_are_small_size` as it was before. @const-t mentioned about it in his comment.

fw/connection.c
EvgeniiMekhanik: yes fixed

db/core/htrie.c
EvgeniiMekhanik: I don't understand this point. Here we zeroed bucket in the record in `tdb_rec_set_remove` and in `tdb_htrie_put_rec` we never execute new code.
In fact there are only two cases in all placos where we call `tdb_htrie_put_rec` when record reference  counter became equal to zero. 
If bucket pointer is not NULL we do extra job, otherwise no.
In all places we call remove bucket before put, except `client` module

db/core/main.c
EvgeniiMekhanik: yes fixed

fw/connection.c
EvgeniiMekhanik: About moving it to `ss_do_shutdown` ok, Although I don't like that the number of callbacks is increasing. We decide to remove timer, because we decide to use TCP timer. In tcp timer we still have our own keepalive callback, so I don't see any problem here

db/core/htrie.c
```diff
@@ -221,28 +223,126 @@ tdb_htrie_get_rec(TdbRec *rec)
...
+}
+
+static void
+tdb_bucket_remove_large_record(TdbHdr *dbh, TdbRec *rec,
+			       tdb_eq_cb_t *eq_cb, void *data,
+			       bool force, bool put_record)
+{
+	TdbBucket *bckt = (TdbBucket *)rec->bucket;
+
+	BUG_ON(tdb_rec_is_removed(rec));
```
krizhanovsky: @EvgeniiMekhanik @const-t let's fix a practice to always comment assertions leading to system crash. This extra step will require to pay more attention to such code, both during development and review time and further incidents forensics.

In particular, such statements much be commented why we have to crash and warning isn't enough. For example, if we check a pointer for NULL, we surely should crash before trying to dereference it. In other words, we should assert only if crash is inevitable.

db/core/htrie.c
```diff
@@ -221,28 +223,126 @@ tdb_htrie_get_rec(TdbRec *rec)
...
+}
+
+void
+tdb_htrie_put_rec(TdbHdr *dbh, TdbRec *rec)
+{
+	TdbBucket *bckt;
+	tdb_eq_cb_t *eq_cb;
+	int refcnt = atomic_dec_return(&rec->refcnt);
+
+	BUG_ON(refcnt < 0);
```
krizhanovsky, line 313: Uncommented assertion

db/core/htrie.c
```diff
@@ -659,12 +759,13 @@ tdb_htrie_smallrec_link(TdbHdr *dbh, size_t len, TdbBucket *bckt)
...
 static TdbRec *
-tdb_htrie_create_rec(TdbHdr *dbh, unsigned long off, unsigned long key,
-		     void *data, size_t len, bool complete)
+tdb_htrie_create_rec(TdbHdr *dbh, TdbBucket *bckt, unsigned long off,
+		     unsigned long key, void *data, size_t len, bool complete)
 {
 	char *ptr = TDB_PTR(dbh, off);
 	TdbRec *r = (TdbRec *)ptr;
 
+	BUG_ON(!bckt);
```
krizhanovsky: Uncommented assertion. Is it possible to warn and exit with an error code?

fw/sock_clnt.c
```diff
@@ -341,17 +361,37 @@ tfw_sock_clnt_drop(struct sock *sk)
...
+	 * to TCP_FIN_WAIT2 state and can stay in this state unlimited
+	 * time (`tcp_fin_timeout` which was set to 10 seconds during
+	 * Tempesta FW start doesn't work for such sockets).
+	 * To prevent such situation set connection keepalive timer
+	 * to 10 seconds and abort connection if this time is expired.
+	 */ 
+	tfw_cli_conn_mod_timer((TfwCliConn *)conn, tcp_fin_timeout);
+}
+
+
```
krizhanovsky: extra empty line

db/core/htrie.c
```diff
@@ -251,8 +251,36 @@ tdb_rec_is_removed(TdbRec *rec)
...
+				tdb_htrie_put_rec(dbh, rec);
+			bckt->rec = 0;
+		}
+	} else {
+		/* Remove incomplete record only if @force is set. */
+		if (force && (!eq_cb || eq_cb(dbh, rec, data))) {
+			tdb_rec_set_remove(rec);
+			if (put_record)
+				tdb_htrie_put_rec(dbh, rec);
+			bckt->rec = 0;
```
krizhanovsky: The code is essentially the same for both the cases, so we can just check the conditions and return from the function early and keep the logic small and not duplicated

db/core/htrie.c
```diff
@@ -280,36 +309,40 @@ tdb_bucket_remove_fixed_size_record(TdbHdr *dbh, TdbBucket *bckt,
...
+	eq_cb = tdb_put_req_eq_cb;
+
+	/*
+	 * There are two different strategies of record deletion:
+	 * - Record can be removed at any time (bckt is equal to zero),
+	 *   but still alive, because there are some active users. In
+	 *   this case, we just delete record here (this strategey
+	 *   is used in the web cache).
+	 * - Record was not removed, but all active users gone.
+	 *   Mark the record as removed record before actual deletion (this strategy used for
```
krizhanovsky: too long lone

db/core/htrie.c
```diff
@@ -280,36 +309,40 @@ tdb_bucket_remove_fixed_size_record(TdbHdr *dbh, TdbBucket *bckt,
...
+	 *   clients).
+	 */
+	if (bckt) {
+		write_lock_bh(&bckt->lock);
+		if (tdb_records_are_large(dbh)) {
+			tdb_bucket_remove_large_record(dbh, rec, eq_cb,
+						       rec, true, false);
+		} else {
+			tdb_bucket_remove_small_record(dbh, bckt, eq_cb, rec,
+						       false);
```
krizhanovsky: Do we really need the recursion (`tdb_htrie_put_rec()` -> `tdb_bucket_remove_{large|small}_record()` -> `tdb_htrie_put_rec()`) and is it the only way to write the code with the recursion controlled with the boolean variables?

db/core/htrie.c
krizhanovsky: The problem was in the previous version, not it's gone. The thing was that there was code, called in `tdb_htrie_put_rec()`, and there were places calling exactly the same code right after `tdb_htrie_put_rec()`.

fw/sock_clnt.c
```diff
@@ -341,17 +361,37 @@ tfw_sock_clnt_drop(struct sock *sk)
...
+	/*
+	 * Socket that was closed by `tcp_shutdown` is not orphaned.
+	 * After receiving ack from remote peer, such socket moved
+	 * to TCP_FIN_WAIT2 state and can stay in this state unlimited
+	 * time (`tcp_fin_timeout` which was set to 10 seconds during
+	 * Tempesta FW start doesn't work for such sockets).
+	 * To prevent such situation set connection keepalive timer
+	 * to 10 seconds and abort connection if this time is expired.
+	 */ 
+	tfw_cli_conn_mod_timer((TfwCliConn *)conn, tcp_fin_timeout);
```
krizhanovsky, line 383: please add TODO comment with particular mentioning of issue or pull request where we get rid of the timer.

Once we get rid of the timer, can we also get rid of the new callback `connection_on_shutdown`?

db/core/htrie.c
EvgeniiMekhanik: There is no recursion here. We pass false as a last argument. In this case _put_record is false and we don't call `tdb_htrie_put_rec` again!

db/core/htrie.c
EvgeniiMekhanik: We always have BUG_ON when reference counter is less then zero. It seems that it lead to double free, so I think that BUG_ON is ok

fw/sock_clnt.c
EvgeniiMekhanik: It seems that after removing timer, we should directly update tcp keep alive timer here

db/core/htrie.c
EvgeniiMekhanik: Fixed

PR #2628, opened 2026-03-11
Fix HPACK dynamic table desync

### LINKED ISSUE #2640: Corrupted content error
**Describe the issue**
Time to time we encounter the error "Corrupted content error" due to this error page can't be loaded. 

**To Reproduce**
Simply browse our site or the container with http2.

Also need to add logging for such errors if browser reports it via `RST_STREAM` or `GOAWAY` frames.

### LINKED ISSUE #2662: Kernel panic in stream scheduler
This panic can be triggered by sending the INITIAL_WINDOW_SIZE setting at the moment when sending DATA frames is postponed and the last part of the headers is prepared for sending.

```
[  106.411233] kernel BUG at /usr/src/tfw/tempesta/fw/http_stream_sched.c:123!
[  106.411911] Oops: invalid opcode: 0000 [#1] PREEMPT SMP NOPTI
[  106.412709] CPU: 3 UID: 0 PID: 3330 Comm: python Tainted: G           OE      6.12.12+ #229
[  106.413671] Tainted: [O]=OOT_MODULE, [E]=UNSIGNED_MODULE
[  106.414218] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS Arch Linux 1.17.0-2-2 04/01/2014
[  106.415291] RIP: 0010:tfw_h2_stream_sched_insert_active+0xd5/0xe0 [tempesta_fw]
[  106.416246] Code: 08 00 0f 85 7a ff ff ff 0f 0b 49 8b 40 20 4c 89 48 08 48 89 47 18 48 89 4f 20 4d 89 48 20 c7 47 28 02 00 00 00 c3 cc cc cc cc <0f> 0b 90 0f 1f 84 00 00 00 00 00 90 90 90 90 90 90 90 90 90 90 90
[  106.418404] RSP: 0018:ffffb3b380174ad8 EFLAGS: 00010246
[  106.418958] RAX: ffff9dd8b76482a8 RBX: ffff9dd8b7648040 RCX: 0000000000000017
[  106.419818] RDX: 0000000000000000 RSI: ffff9dd8bc13a578 RDI: ffff9dd8bc13a578
[  106.420845] RBP: ffff9dd8b7648048 R08: ffff9dd8b7648048 R09: 0000000000000400
[  106.421562] R10: ffff9dd895edc5b0 R11: 0000000000000140 R12: ffff9dd89b5b2fa0
[  106.422272] R13: 0000000000000400 R14: 0000000000000000 R15: ffff9dd8b7648198
[  106.423199] FS:  00007f9e158396c0(0000) GS:ffff9ddbd8380000(0000) knlGS:0000000000000000
[  106.424146] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[  106.424840] CR2: 00007f3a874a8350 CR3: 000000013de42000 CR4: 0000000000750ef0
[  106.425722] PKRU: 55555554
[  106.426256] Call Trace:
[  106.426594]  <IRQ>
[  106.426879]  ? die+0x36/0xa0
[  106.427248]  ? do_trap+0xdd/0x100
[  106.427786]  ? tfw_h2_stream_sched_insert_active+0xd5/0xe0 [tempesta_fw]
[  106.428682]  ? do_error_trap+0x6a/0xa0
[  106.429147]  ? tfw_h2_stream_sched_insert_active+0xd5/0xe0 [tempesta_fw]
[  106.429832]  ? exc_invalid_op+0x50/0x80
[  106.430543]  ? tfw_h2_stream_sched_insert_active+0xd5/0xe0 [tempesta_fw]
[  106.431215]  ? asm_exc_invalid_op+0x1a/0x20
[  106.431868]  ? tfw_h2_stream_sched_insert_active+0xd5/0xe0 [tempesta_fw]
[  106.432590]  tfw_h2_sched_activate_stream+0x69/0x80 [tempesta_fw]
[  106.433423]  tfw_h2_apply_new_settings+0x1f3/0x260 [tempesta_fw]
[  106.434198]  ss_skb_tcp_entail+0x43/0x3c0 [tempesta_fw]
[  106.434812]  ss_skb_tcp_entail_list+0x124/0x560 [tempesta_fw]
[  106.435574]  tfw_h2_stream_send_postponed+0x28/0xc0 [tempesta_fw]
[  106.436246]  tfw_h2_make_frames+0x604/0xac0 [tempesta_fw]
[  106.437113]  tfw_connection_fill_sk_write_queue+0x119/0x260 [tempesta_fw]
[  106.437821]  tcp_push_pending_frames+0x33/0x160
[  106.438329]  tcp_rcv_established+0x20f/0x6e0
[  106.438803]  tcp_v4_do_rcv+0x161/0x2c0
[  106.439234]  tcp_v4_rcv+0x1233/0x1560
[  106.439908]  ? raw_local_deliver+0xd5/0x280
[  106.440442]  ip_protocol_deliver_rcu+0x36/0x300
[  106.441131]  ip_local_deliver_finish+0x76/0xa0
[  106.441612]  __netif_receive_skb_one_core+0x89/0xa0
[  106.442284]  process_backlog+0x87/0x140
[  106.442715]  __napi_poll+0x28/0x1c0
[  106.443346]  net_rx_action+0x3a7/0x420
[  106.443775]  ? try_to_wake_up+0x31b/0x680
[  106.444215]  handle_softirqs+0xd9/0x2e0
[  106.444931]  do_softirq.part.0+0x3b/0x60
[  106.445356]  </IRQ>
[  106.445793]  <TASK>
[  106.446064]  __local_bh_enable_ip+0x64/0x80
[  106.446560]  __dev_queue_xmit+0x265/0xfa0
[  106.446992]  ? nft_do_chain_inet+0x58/0x140 [nf_tables]
[  106.447664]  ? nf_hook_slow+0x42/0x100
[  106.448072]  ip_finish_output2+0x2b8/0x580
[  106.448555]  __ip_queue_xmit+0x185/0x4c0
[  106.449116]  __tcp_transmit_skb+0xacf/0xca0
[  106.449561]  tcp_recvmsg_locked+0x29a/0xa40
[  106.450144]  tcp_recvmsg+0x9b/0x200
[  106.450618]  inet_recvmsg+0x56/0x140
[  106.451017]  ? x2apic_send_IPI+0x49/0x60
[  106.451591]  ? ttwu_queue_wakelist+0xd0/0x100
[  106.452042]  sock_recvmsg+0xa6/0xe0
[  106.452635]  ? _raw_spin_unlock_irqrestore+0xe/0x40
[  106.453157]  sock_read_iter+0x93/0x100
[  106.453728]  vfs_read+0x338/0x380
[  106.454106]  ksys_read+0xc1/0x100
[  106.454468]  do_syscall_64+0x4f/0x120
[  106.455103]  entry_SYSCALL_64_after_hwframe+0x76/0x7e
[  106.455662] RIP: 0033:0x7f9e2011baca
[  106.456045] Code: 55 48 89 e5 48 83 ec 20 48 89 55 e8 48 89 75 f0 89 7d f8 e8 98 ca f7 ff 48 8b 55 e8 48 8b 75 f0 41 89 c0 8b 7d f8 31 c0 0f 05 <48> 3d 00 f0 ff ff 77 2e 44 89 c7 48 89 45 f8 e8 f2 ca f7 ff 48 8b
[  106.458148] RSP: 002b:00007f9e15838580 EFLAGS: 00000246 ORIG_RAX: 0000000000000000
[  106.459014] RAX: ffffffffffffffda RBX: 0000000000000433 RCX: 00007f9e2011baca
[  106.459725] RDX: 0000000000000433 RSI: 00007f9e04003f38 RDI: 000000000000002c
[  106.460661] RBP: 00007f9e158385a0 R08: 0000000000000000 R09: 0000000000000000
[  106.461335] R10: 0000000000000001 R11: 0000000000000246 R12: 00007f9e04003e60
[  106.462175] R13: 00007f9e04003f38 R14: 00007f9e15839638 R15: 0000000000000000
[  106.463063]  </TASK>
[  106.463331] Modules linked in: tls tempesta_fw(OE) tempesta_regex(OE) tempesta_db(OE) tempesta_tls(OE) tempesta_lib(OE) veth vhost_vsock vmw_vsock_virtio_transport_common vhost vhost_iotlb vsock intel_rapl_msr intel_rapl_common xt_conntrack xt_MASQUERADE xfrm_user xfrm_algo xt_set ip_set nft_masq nft_chain_nat nf_nat nf_conntrack xt_addrtype nf_defrag_ipv6 nft_compat nf_defrag_ipv4 bridge stp llc nf_tables snd_hda_codec_generic snd_hda_intel snd_intel_dspcfg snd_intel_sdw_acpi snd_hda_codec overlay snd_hda_core kvm_amd qxl snd_hwdep cfg80211 snd_pcm ccp drm_ttm_helper i2c_i801 snd_timer ttm snd i2c_mux binfmt_misc kvm input_leds soundcore i2c_smbus joydev lpc_ich drm_kms_helper virtiofs serio_raw mac_hid sch_fq_codel dm_multipath efi_pstore drm nfnetlink dmi_sysfs qemu_fw_cfg ip_tables x_tables autofs4 btrfs blake2b_generic raid10 raid456 async_raid6_recov async_memcpy async_pq async_xor async_tx xor raid6_pq libcrc32c raid1 raid0 crct10dif_pclmul hid_generic crc32_pclmul ghash_clmulni_intel usbhid virtio_net ahci
[  106.463366]  sha512_ssse3 net_failover psmouse sha256_ssse3 virtio_scsi libahci hid virtio_rng failover virtio_blk aesni_intel crypto_simd cryptd
[  106.475947] ---[ end trace 0000000000000000 ]---
```

const-t:


fw/http_frame.c
```diff
@@ -2210,6 +2210,18 @@ do {									\
 	}
 
 	T_FSM_STATE(HTTP2_MAKE_HEADERS_FRAMES) {
+		/*
+		 * This call doesn't change the stream state, but sets ctx->cur_send_headers.
+		 * We do this to force the stream scheduler to select this
+		 * stream during next sending if current sending of this stream
+		 * has been postponed due to lack of tcp window.
+		 */
+		r = tfw_h2_stream_fsm_ignore_err(ctx, stream, HTTP2_HEADERS, 0);
```
const-t: @krizhanovsky @EvgeniiMekhanik This is alternative solution that lets to choose another stream on the next scheduling cycle, because not encoding header if they can't be sent due to too small window (`snd_wnd <= FRAME_HEADER_SIZE + TLS_MAX_OVERHEAD` window too small for any stream not only for the current), but has worse latency because of postponing encoding. And may postpone already framed stream, that leads to smaller HEADERS frame than possible and forces to use CONTINUATION frame.

From my point of view, setting of `ctx->cur_send_headers` directly is not consistent, but maybe you guys have another vision.

```
diff --git a/fw/http_frame.c b/fw/http_frame.c
index 8e9b29a0c..968d1c360 100644
--- a/fw/http_frame.c
+++ b/fw/http_frame.c
@@ -2161,6 +2161,13 @@ do {									\
 	T_FSM_EXIT();							\
 } while(0)

+#define CALC_MAX_MIN_FRAME_LENGTH(len, out_max, out_min)		\
+do {									\
+	(out_max) = min(TLS_MAX_PAYLOAD_SIZE, *snd_wnd - TLS_MAX_OVERHEAD); \
+	(out_max) -= FRAME_HEADER_SIZE;					\
+	(out_min) = min(min_to_send, (unsigned int)(len));		\
+} while (0)
+
 #define CALC_FRAME_LENGTH_AND_SET_FRAME_TYPE(type, len)			\
 do {									\
 	unsigned int max_len;						\
@@ -2170,9 +2177,7 @@ do {									\
 		*stop = true;						\
 		T_FSM_EXIT();						\
 	}								\
-	max_len = min(TLS_MAX_PAYLOAD_SIZE, *snd_wnd - TLS_MAX_OVERHEAD); \
-	max_len -= FRAME_HEADER_SIZE;					\
-	min_len = min(min_to_send, (unsigned int)len);			\
+	CALC_MAX_MIN_FRAME_LENGTH(len, max_len, min_len);		\
 	frame_length = tfw_h2_calc_frame_length(ctx, stream, type, len,	\
 						max_len); 		\
 	/*								\
@@ -2190,9 +2195,16 @@ do {									\
 	T_FSM_JMP(state);						\
 } while(0)

+#define FRAME_XMIT_FSM_SET_NEXT_STATE(state)				\
+	__fsm_const_state = state
+
 	T_FSM_START(stream->xmit.state) {

 	T_FSM_STATE(HTTP2_ENCODE_HEADERS) {
+		if (*snd_wnd <= FRAME_HEADER_SIZE + TLS_MAX_OVERHEAD) {
+			*stop = true;
+			T_FSM_EXIT();
+		}
 		r = tfw_h2_stream_xmit_prepare_resp(stream);
 		fallthrough;
 	}
@@ -2210,8 +2222,14 @@ do {									\
 	}

 	T_FSM_STATE(HTTP2_MAKE_HEADERS_FRAMES) {
-		CALC_FRAME_LENGTH_AND_SET_FRAME_TYPE(HTTP2_HEADERS,
-						     stream->xmit.h_len);
+		unsigned int min_len, max_len;
+
+		CALC_MAX_MIN_FRAME_LENGTH(stream->xmit.h_len, max_len, min_len);
+		frame_length = tfw_h2_calc_frame_length(ctx, stream,
+							HTTP2_HEADERS,
+							stream->xmit.h_len,
+							max_len);
+
 		if (unlikely(ctx->hpack.enc_tbl.wnd_changed)) {
 			r = tfw_hpack_enc_tbl_write_sz(&ctx->hpack.enc_tbl,
 						       stream);
@@ -2229,6 +2247,12 @@ do {									\
 			return r;
 		}

+		if (frame_length < min_len) {
+			FRAME_XMIT_FSM_SET_NEXT_STATE(HTTP2_SEND_FRAMES);
+			*stop = true;
+			T_FSM_EXIT();
+		}
+
 		FRAME_XMIT_FSM_NEXT(frame_length, HTTP2_SEND_FRAMES);
 	}
```

fw/http_stream.c
```diff
@@ -346,10 +346,12 @@ do {									\
 	}								\
 } while(0)
 
-#define TFW_H2_FSM_TYPE_CHECK(ctx, stream, op, type)			\
+#define TFW_H2_FSM_TYPE_CHECK(ctx, op, type, is_send)			\
 do {									\
 	if ((ctx->cur_##op##_headers					\
-	     && (type != HTTP2_CONTINUATION && type != HTTP2_RST_STREAM)) \
+	     && ((type == HTTP2_HEADERS && !is_send) ||			\
```
EvgeniiMekhanik, line 352: It seems that it is wrong. A HEADERS frame without the END_HEADERS flag set MUST be followed by a [CONTINUATION](https://www.rfc-editor.org/rfc/rfc9113.html#CONTINUATION) frame for the same stream. A receiver MUST treat the receipt of any other type of frame or a frame on a different stream as a [connection error](https://www.rfc-editor.org/rfc/rfc9113.html#ConnectionErrorHandler) ([Section 5.4.1](https://www.rfc-editor.org/rfc/rfc9113.html#ConnectionErrorHandler)) of type [PROTOCOL_ERROR](https://www.rfc-editor.org/rfc/rfc9113.html#PROTOCOL_ERROR).

It seems that HTTP2_RST_STREAM is also forbidden. So it should be
if ((ctx->cur_##op##_headers	and type != HTTP2_CONTINUATION)

EvgeniiMekhanik: I think it's better to check tcp window and do not prepare response. Or ignore tcp windo limit for headers frame
.

EvgeniiMekhanik (changes requested):
I think we should not prepare response if there is no enough TCP window or ignore tcp window during sending headers frame. Otherwise we need make incorrect change in stream state processing. @krizhanovsky what do you think?

EvgeniiMekhanik:
I also found two problems in
f (frame_length < min_len)
		ADJUST_BLOCKED_STREAMS_AND_EXIT(min_len, type);
optimization:
- we stop making frames if stream http2 window is exceeded. But if we have other streams to make frames we should do it.
- we stop making frames if http2 connection window is exceeded. But if we have other streams with headers frames
we should continue make frames for them.

fw/http_stream.c
const-t: Added fix for this as well.

const-t:
> I think we should not prepare response if there is no enough TCP window or ignore tcp window during sending headers frame. Otherwise we need make incorrect change in stream state processing. @krizhanovsky what do you think?

I've updated PR. Added fixes for #2640.  But I didn't implement `I think we should not prepare response if there is no enough TCP window`. From my point of view, if we reached framing function we can prepare headers even if tcp window is not enough, to have prepared headers on the next shot of calling sending function and don't call stream scheduler one more time.  However I attached the version where we don't prepare response if there is no enough TCP window in the comment to see the difference.

Actually this patch allows to call `tfw_h2_stream_fsm_ignore_err()` multiple times for `HEADERS` frame in the state `HTTP2_STREAM_REM_HALF_CLOSED` only during sending that doesn't looks like issue, because we only assign `ctx->cur_send_headers` in this case and not change state of the stream, frames received from the client will not affect this change only sending side.

fw/hpack.c
```diff
@@ -3744,16 +3739,23 @@ tfw_hpack_set_rbuf_size(TfwHPackETbl *__restrict tbl, unsigned short new_size)
 	 * size that occurs in that interval MUST be signaled in a dynamic
 	 * table size update.
 	 */
-	if (tbl->window != new_size && (likely(!tbl->wnd_changed)
-	    || unlikely(!tbl->window) || new_size < tbl->window))
+	if (tbl->window != requested_size && (likely(!tbl->wnd_changed)
```
EvgeniiMekhanik: I read RFC again and look for nghtt2 realization - in some cases we should apply and send two hpack dynamic changes!
For example
4096 - > 1 -> 100 we should send 1 and 100 (minimum and final)
4096 -> 1 -> 0 -> 2 -> 100 -> 3 we should send 0 and 3 (minimum and final)

EvgeniiMekhanik:
There are some questions about patch:
1. If we can't send any other frames for other streams after sending HEADERS FRAME without END_STREAM flag for current stream may be it have sense to don't calculate snd_wnd during sending headers at all? In this case we don't need some commits from this patch.  (UPDATE. I check h20 implementation they don't send all headers to socket immediately< also we rely on window when we account client mem, so we still should calculate snd_wnd during sending headers, but may be not do it so accuracy).
2. We should check how another implementations works with RST STREAM receiving. It seems that now be make some changes not according RFC. It looks like we are trying to simply hide the current problem rather than solve it globally. (I mean that we continue to send headers frames after receiving RST STREAM and more over we don't process it now at all!)

fw/hpack.c
const-t: Good catch, another one bug. Tempesta must be able to send min and max size. RFC cite:

> Multiple updates to the maximum table size can occur between the
   transmission of two header blocks.  In the case that this size is
   changed more than once in this interval, the smallest maximum table
   size that occurs in that interval MUST be signaled in a dynamic table
   size update.  The final maximum size is always signaled, resulting in
   at most two dynamic table size updates.  This ensures that the
   decoder is able to perform eviction based on reductions in dynamic
   table size
   
   However maybe it makes sense to implement it in other PR, to keep this PR small.

EvgeniiMekhanik: Yes also look `nghttp2_hd_deflate_hd_bufs` in nghttp2
```
if (deflater->ctx.hd_table_bufsize_max > min_hd_table_bufsize_max) {

      rv = emit_table_size(bufs, min_hd_table_bufsize_max);

      if (rv != 0) {
        goto fail;
      }
    }

    rv = emit_table_size(bufs, deflater->ctx.hd_table_bufsize_max);
```

fw/http_frame.c
```diff
@@ -199,7 +199,8 @@ do {									\
 			return T_BAD;					\
 		} else if (res == STREAM_FSM_RES_TERM_STREAM) {		\
 			WARN_ON_ONCE(hdr->stream_id != ctx->cur_stream->id); \
-			return tfw_h2_current_stream_send_rst((ctx), err); \
+			return tfw_h2_send_rst_stream(ctx, ctx->cur_stream->id,\
```
EvgeniiMekhanik: May we can implement `tfw_h2_on_tcp_entail_rst` (like tfw_h2_on_tcp_entail_ack). to move stream to closed queue?

EvgeniiMekhanik: Currently it will be moved to closed queue during making data frames (because of error in tfw_h2_insert_frame_header->tfw_h2_stream_fsm_ignore_err but it is not clear (as for me)

const-t: > Currently it will be moved to closed queue during making data frames (because of error in tfw_h2_insert_frame_header->tfw_h2_stream_fsm_ignore_err but it is not clear (as for me)

This is correct for requests for that we will prepare response, however if response will not be prepared we will not add stream to closed and don't clean before connection closing. I will fix this.

const-t:
> I also found two problems in f (frame_length < min_len) ADJUST_BLOCKED_STREAMS_AND_EXIT(min_len, type); optimization:
> 
>     * we stop making frames if stream http2 window is exceeded. But if we have other streams to make frames we should do it.
> 
>     * we stop making frames if http2 connection window is exceeded. But if we have other streams with headers frames
>       we should continue make frames for them.

In this PR I addressed the first point, but the second point still not fixed and I'm sure that not a subject of the current PR and requires some research(maybe some performance test on real and complex web page) on it. Because there are few approaches to fix this and I'm not sure which one is better. Some implementations just queue the headers when they are ready and and in case of exceeded connection window they are just stop sending data. But if differs from our approach. Maybe we need to have another one queue with stream which headers are ready to send and in case of  exceeded connection window we can take stream with highest priority to send headers. However, I have a doubts about any of these solutions, because most of implementations try to keep connection window very large and update it before it exceeded.

fw/http_frame.c
```diff
@@ -2054,6 +2067,43 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
 	bool trailers = false;
 	char *data;
 	int r = 0;
+	unsigned char flags;
+
+	if (type == HTTP2_DATA) {
```
EvgeniiMekhanik: we can't move it here! Look later we check `if (type == HTTP2_CONTINUATION || type == HTTP2_DATA
	    || !stream->xmit.h_len) {` and this check should be done with previous value of h_len (which is not decremented!!)

fw/http_frame.c
```diff
@@ -2438,6 +2409,66 @@ do {									\
...
+			 * restore actual size here.
+			 */
+			stream->xmit.frame_length += frame_length;
+			/**
+			 * Purge stream send queue, but leave postponed
+			 * skbs and rst stream/goaway/tls alert if exist.
+			 */
+			tfw_h2_stream_purge_send_queue(stream);
+
+			if (unlikely(stream->xmit.postponed) &&
```
EvgeniiMekhanik: __tfw_h2_is_ready_to_send_postponed? ctx->cur_send_headers should also be zeroed in tfw_h2_stream_purge_send_queue. Previously it seems that we zero it during stream processing (in case of STREAM_FSM_RES_TERM_STREAM)

fw/http_frame.c
```diff
@@ -2091,8 +2070,11 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
 	}
 
 	data = ss_skb_data_ptr_by_offset(stream->xmit.skb_head,
-					 stream->xmit.frame_length);
-	BUG_ON(!data);
+					 stream->xmit.bytes_to_send);
+	if (unlikely(!data)) {
+		WARN_ONCE(1, "Can't find offset in skb.");
```
EvgeniiMekhanik: May be `if (unlikely(WARN_ONCE(!data, ..)`

fw/http_frame.c
```diff
@@ -2240,117 +2256,91 @@ do {									\
...
+		stream->xmit.headers_frame_length -= bytes_to_send;
+		stream->xmit.bytes_to_send += bytes_to_send;
+		*snd_wnd -= bytes_to_send;
 
-		FRAME_XMIT_FSM_NEXT(frame_length, HTTP2_SEND_FRAMES);
-	}
+		/*
+		 * When prepared (framed) headers are finished don't force
+		 * skb splitting, go to the next frame and try to place the new
+		 * frame header to the current skb. It saves one ss_skb_split
```
EvgeniiMekhanik, line 2471: Why only one? I think we should not split until the end of this function. It seems that we should not split skb for DATA frames also

fw/http_frame.c
```diff
@@ -2165,61 +2139,103 @@ tfw_h2_stream_send_postponed(struct sock *sk, struct sk_buff **skb_head,
...
+
+	stream->xmit.h_len -= frame_length;
+
+	if (!stream->xmit.h_len)
+		flags |= HTTP2_F_END_HEADERS;
+	if (!stream->xmit.b_len && !stream->xmit.t_len)
+		flags |= HTTP2_F_END_STREAM;
+
+	r = tfw_h2_stream_fsm_ignore_err(ctx, stream, type, flags);
+	ctx->cur_send_headers = stream;
```
EvgeniiMekhanik: Why we set cur_send_headers if r != 0 ?

fw/http_frame.c
```diff
@@ -2165,61 +2139,103 @@ tfw_h2_stream_send_postponed(struct sock *sk, struct sk_buff **skb_head,
...
+	unsigned int max_payload_size = ctx->rsettings.max_frame_sz
+		- FRAME_HEADER_SIZE;
+	unsigned char flags = 0;
+
+	unsigned int frame_length = min(max_payload_size, stream->xmit.h_len);
+
+	stream->xmit.h_len -= frame_length;
+	if (!stream->xmit.h_len)
+		flags |= HTTP2_F_END_HEADERS;
+
```
EvgeniiMekhanik, line 2313: Please describe the idea that we don't call  `tfw_h2_stream_fsm_ignore_err` for continuation frames (here or in commit message) to prevent the situation when we receive RST_STREAM and stop send continuations.

fw/http_frame.c
```diff
@@ -2240,117 +2256,91 @@ do {									\
...
+			T_FSM_EXIT();
 		}
-		r = tfw_h2_insert_frame_header(sk, ctx, stream, frame_type,
-					       frame_length);
-		if (unlikely(r))
-			T_FSM_JMP(HTTP2_FRAMING_FAILED);
 
-		FRAME_XMIT_FSM_NEXT(frame_length, HTTP2_SEND_FRAMES);
-	}
+		unsigned int max_len = __tfw_h2_snd_wnd_limit(*snd_wnd);
```
EvgeniiMekhanik: usually we declare variables at the beginning of the block.

fw/hpack.c
```diff
@@ -3732,57 +3733,74 @@ tfw_hpack_set_rbuf_size(TfwHPackETbl *__restrict tbl,
...
 	WARN_ON_ONCE(!tbl->wnd_changed);
-	write_int(tbl->window, 0x1F, 0x20, &tmp);
+	if (tbl->window > tbl->min_window)
+		write_int(tbl->min_window, 0x1F, 0x20, &min_val);
+	write_int(tbl->window, 0x1F, 0x20, &max_val);
 
 	data = ss_skb_data_ptr_by_offset(skb_head,
 					 offset + FRAME_HEADER_SIZE);
-	BUG_ON(!data);
+	if (!data) {
```
EvgeniiMekhanik: if (WARN_INCE(!data)

fw/hpack.c
```diff
@@ -3718,71 +3719,49 @@ tfw_hpack_encode(TfwHttpResp *__restrict resp, TfwStr *__restrict hdr,
...
-					    (TfwHPackETblIter *)tbl);
-		WARN_ON_ONCE(tbl->rb_len > tbl->size);
+	if (requested_size == tbl->window)
+		return;
 
-		tbl->window = new_size;
-		tbl->wnd_changed = true;
-	}
-}
+	unsigned short new_size = min_t(unsigned int, requested_size,
```
EvgeniiMekhanik, line 3739: we always declare variables at the beginning of the block (function)

fw/http_frame.c
```diff
@@ -199,7 +199,19 @@ do {									\
...
+								      err); \
+			} else {					\
+				unsigned int id = ctx->cur_stream->id;	\
+				/**
+				 * If Tempesta is already sending headers,
+				 * only update the stream state and schedule
+				 * sending an RST_STREAM frame, but do not
+				 * remove the stream from the scheduling queue.
+				 */					\
+				return tfw_h2_send_rst_stream(ctx, id, err); \
```
EvgeniiMekhanik: look here https://github.com/tempesta-tech/tempesta/commit/835f017df29ec3940c15c2a0c90073a787f2d993#diff-6044b301803db90a546df674a8dfd1efa458dd1937a1a4d2dfe2b5a19ede5813
it seems that we should also set ctx->cur_stream = NULL;	here. Also we have one more place where we call  `tfw_h2_current_stream_send_rst`. If we decide to close connection not send RST stream in all other places please check should we send RST stream in tfw_h2_wnd_update_process.
We also need a test for this place to be sure that such stream will be moved to closed queue.

fw/http_frame.c
```diff
@@ -294,6 +297,82 @@ tfw_h2_on_send_dflt(void *conn, struct sk_buff **skb_head)
...
+ * WINDOW_UPDATE, PING and SETTINGS/ack frames, because:
+ * 1. The frame may be generated on local cpu.
+ * 2. Response to one of the listed frames may be sent on the same cpu where
+ *    the frame has been received.
+ *
+ * WARNING: This function relies on data sending in @tfw_h2_conn_recv_finish()
+ * DON'T call this function on TX path.
+ */
+static int
+__tfw_h2_send_frame_on_rx(TfwH2Ctx *ctx, TfwFrameHdr *hdr, TfwStr *data)
```
EvgeniiMekhanik, line 313: Can we do it? In this case we don't pass sending frame through the `ss_send/ss_do_send` and can change sending/closing order. For example Tempesta FW received incorrect request and send error response with goaway. This response was send using `ss_send`. On the next skb processing we call this function and send some data in incorrect order.

fw/http_frame.c
```diff
@@ -1652,6 +1715,10 @@ tfw_h2_frame_recv(void *data, unsigned char *buf, unsigned int len,
 		if (ctx->to_read)
 			FRAME_FSM_MOVE(HTTP2_RECV_FRAME_SETTINGS);
 
+		if (test_bit(HTTP2_SETTINGS_NEED_TO_APPLY,
```
EvgeniiMekhanik, line 1736: Also we should apply settings only when ack will be sent to the peer. That's why such sophisticated logic was implemented.  If this way is possible (`__tfw_h2_send_frame_on_rx` and applying here) we can return  to the simple version that was changed in https://github.com/tempesta-tech/tempesta/commit/cd5fc37f62990522c25c6aebaa65a93750777721#diff-6044b301803db90a546df674a8dfd1efa458dd1937a1a4d2dfe2b5a19ede5813

fw/http_frame.c
```diff
@@ -2088,52 +2155,56 @@ tfw_h2_insert_frame_header(struct sock *sk, TfwH2Ctx *ctx, TfwStream *stream,
...
+
+	WARN_ON_ONCE(!tbl->wnd_changed);
+	if (tbl->window > tbl->min_window)
+		write_int(tbl->min_window, 0x1F, 0x20, &min_val);
+	write_int(tbl->window, 0x1F, 0x20, &max_val);
+
+	data = ss_skb_data_ptr_by_offset(skb_head,
+					 offset + FRAME_HEADER_SIZE);
+	if (unlikely(!data)) {
+		WARN_ONCE(1, "Can't find offset in skb.");
```
EvgeniiMekhanik: if WARN_ONCE(!data

fw/http_frame.c
```diff
@@ -2174,60 +2245,112 @@ tfw_h2_stream_send_postponed(struct sock *sk, struct sk_buff **skb_head,
 	return 0;
 }
 
+static __always_inline int
+__tfw_h2_make_headers_frame(TfwH2Ctx *ctx, TfwStream *stream)
+{
+	int r;
+	TfwFrameType type = HTTP2_HEADERS;
+	unsigned int max_payload_size = ctx->rsettings.max_frame_sz
```
EvgeniiMekhanik: max_frame_sz doesn't include FRAME_HEADER_SIZE. we should not calculate max_payload_size in such way

fw/http_frame.c
```diff
@@ -2174,60 +2245,112 @@ tfw_h2_stream_send_postponed(struct sock *sk, struct sk_buff **skb_head,
...
+	return tfw_h2_insert_frame_header(ctx, stream, type, frame_length,
+					  flags,
+					  &stream->xmit.headers_frame_length);
+}
+
+static __always_inline int
+__tfw_h2_make_continuation_frame(TfwH2Ctx *ctx, TfwStream *stream)
+{
+	TfwFrameType type = HTTP2_CONTINUATION;
+	unsigned int max_payload_size = ctx->rsettings.max_frame_sz
```
EvgeniiMekhanik: Incorrect max_payload_size calculation. we should remove FRAME_HEADER_SIZE from send_wnd size not here

fw/http_frame.c
EvgeniiMekhanik: Please check this condition since you set
```
stream->xmit.h_len = stream->xmit.t_len;
stream->xmit.t_len = 0;
```
previously the code below was called for trailer headers, now not called

fw/http_frame.c
```diff
@@ -2249,134 +2372,213 @@ do {									\
...
+			*stop = true;
+			T_FSM_EXIT();
 		}
 
-		r = tfw_h2_insert_frame_header(sk, ctx, stream, frame_type,
-					       frame_length);
-		if (unlikely(r)) {
-			T_WARN("Failed to make headers frame %d", r);
-			return r;
+		unsigned int snd_wnd_budget = __tfw_h2_snd_wnd_limit(*snd_wnd);
```
EvgeniiMekhanik, line 2446: We always declare variables at the beginning of the block of code

fw/http_frame.c
```diff
@@ -2249,134 +2372,213 @@ do {									\
...
-		r = tfw_h2_insert_frame_header(sk, ctx, stream, frame_type,
-					       frame_length);
+		/*
+		 * When prepared (framed) headers are finished don't force
+		 * skb splitting, go to the next frame and try to place the new
+		 * frame header to the current skb. It saves one ss_skb_split
+		 * call.
+		 */
+		const bool has_prepared_headers =
+			!!stream->xmit.headers_frame_length;
```
EvgeniiMekhanik, line 2475: It seems that we can also remove `headers_frame_length` if we do it for data frames also

fw/http_frame.c
```diff
@@ -2249,134 +2372,213 @@ do {									\
...
-					       frame_length);
-		if (unlikely(r)) {
-			T_WARN("Failed to make trail continuation frame %d", r);
-			return r;
+		unsigned int snd_wnd_budget = __tfw_h2_snd_wnd_limit(*snd_wnd);
+		unsigned int allowed_send_len =
+			min(min_to_send, (unsigned int)stream->xmit.b_len);
+
+		if ((snd_wnd_budget - FRAME_HEADER_SIZE < allowed_send_len &&
+		     __socket_has_unsent_data(sk)) ||
```
EvgeniiMekhanik: It seems that we should check ((snd_wnd_budget - FRAME_HEADER_SIZE < allowed_send_len || ctx->rem_wnd < allowed_send_len) && __socket_has_unsent_data(sk)

fw/http_frame.c
```diff
@@ -2389,59 +2591,97 @@ do {									\
...
 		r = tfw_h2_entail_stream_skb(sk, stream,
-					     &stream->xmit.frame_length,
+					     &stream->xmit.bytes_to_send,
 					     true);
 		if (unlikely(r)) {
 			T_WARN("Failed to send frame %d", r);
 			return r;
 		}
-		WARN_ON(stream->xmit.frame_length);
-		if (unlikely(stream->xmit.postponed)
```
EvgeniiMekhanik, line 2445: Why it was removed here?

EvgeniiMekhanik (dismissed):
Need some fixes

fw/http_frame.c
const-t: Outdated, this code is not exists in the final version.

fw/http_frame.c
const-t: Fixed

fw/http_frame.c
const-t: changed

fw/http_frame.c
const-t: Only one because we use whole send window during sending headers. We split in three cases:
1. Headers are large, sending only part of them - window is not enough.
2. All headers are sent, but we have postponed frame, before sending DATA we trying to send postponed frame.
3. Skb, headers frame and send window greater than `TLS_MAX_PAYLOAD_SIZE`. (very unlikely, not possible with default headers frame size)

If none of these are true, we do not split and sending DATA frames without spliting.

fw/http_frame.c
const-t: Changed together with https://github.com/tempesta-tech/tempesta/pull/2628#discussion_r3571259279

fw/http_frame.c
const-t: Added comment to the code

fw/http_frame.c
const-t: No, we don't have this rule. It was required by C89, however current standard C11 allows this.

fw/hpack.c
const-t: changed

fw/hpack.c
const-t: No, we don't have this rule. It was required by C89, however current standard C11 allows this.

fw/http_frame.c
const-t: >Also we have one more place where we call tfw_h2_current_stream_send_rst. If we decide to close connection not send RST stream in all other places please check should we send RST stream in tfw_h2_wnd_update_process.

H2spec requires to have rst_stream in this case and this is the only one place where it requires RST_STREAM, not connection closing. I believe this may happen even with valid client. That's why I didn't change this.

> we should also set ctx->cur_stream = NULL

As I see, we anyway zero `ctx->cur_stream ` in `tfw_h2_context_reinit()` and we are OK even without it. But I've added `ctx->cur_stream = NULL` just to be sure in the future.

>  We also need a test for this place to be sure that such stream will be moved to closed queue.

We have test `test_initiate_stream_reset_during_sending_resp_headers_and_trigger_stream_cleanup()` that triggers cleanup of `closed_queue`, but we can ensure that it really in the `closed_queue` only adding debug.  Maybe you have idea how to do this in other way?

fw/http_frame.c
const-t: I believe we can, and the case described in the comment is the reason why I implemented it this way. It guarantees that after sending a SETTINGS ACK and applying settings, we always prepare a new response with the updated settings. 

In HTTP/2, WINDOW_UPDATE, PING, and SETTINGS frames can be sent in any order. For example, the same behavior occurs when we queue a request for forwarding upstream. While the response is not ready, Tempesta may send a PING or SETTINGS ack frame even if it was received after the request, and this is the correct behavior.

fw/http_frame.c
const-t: I'd prefer to not revert this. The current implementation looks better, because it prevents following situation:
1. Client sends request
2. Tempesta forwards request and wait for response from the upstream.
3. Client sends SETTINGS frame, but only part of it, one setting of five.
4. Tempesta applied this setting, but it can't send ACK, because it waiting for other 4 settings.
5. Response is ready and Tempesta sends the new response using the new applied setting, however this is incorrect, because SETTINGS ACK is not sent - protocol violation.

fw/http_frame.c
const-t: changed

fw/http_frame.c
const-t: Thank you, I really missed this

fw/http_frame.c
const-t: fixed

fw/http_frame.c
const-t: I believe we are ok, I can't see any issues with it.

fw/http_frame.c
const-t: I didn't get this.
`headers_frame_length` is the length of the current framed headers frame. If you asking about this `should_split` condition - we need this to do split in case of frame is greater than `TLS_MAX_PAYLOAD_SIZE`.

fw/http_frame.c
const-t: I described this in the comment https://github.com/tempesta-tech/tempesta/pull/2628#issuecomment-4925514766

If we do like you suggesting we getting silly window syndrome on slow client. Also HTTP2 is another application layer and we must not rely on unACKED data from transport layer. For http2 we should have another mechanism (track http2 window of add timer or something else), however now it works and I'm not sure that we must change this.

fw/http_frame.c
const-t: This part `if (unlikely(stream->xmit.postponed)` is removed because it doesn't make sense now. postponed frames can be in the queue only during sending headers, however now we handling sending of postponed data in the state `HTTP2_SEND_HEADERS_FRAME` so this will duplicate it without effect.

fw/http_frame.c
```diff
@@ -304,15 +308,85 @@ static TfwSkbHooks tfw_h2_rst_frame_skb_hooks = {
...
+	ss_skb_setup_head_of_list(msg.skb_head, msg.skb_head->mark, tls_type);
+
+	if (ctx->cur_send_headers) {
+		ss_skb_queue_splice(&ctx->cur_send_headers->xmit.postponed,
+				    &msg.skb_head);
+	} else {
+		struct sock *sk = conn->sk;
+
+		ss_skb_queue_splice(&conn->write_queue, &msg.skb_head);
+		sock_set_flag(sk, SOCK_TEMPESTA_HAS_DATA);
```
EvgeniiMekhanik: This is not work if there is no data in socket write queue. We check SOCK_TEMPESTA_HAS_DATA in two cases:
1. In ss_do_send (we don't call it after this function)
2. When some data received from the remote peer.
This simple test doesn't work on this branch but correctly work on master
```
    async def test_long_header(self):
        await self.start_all_services()

        client = self.get_client("deproxy")
        server = self.get_server("deproxy")
        client.update_initial_settings(max_header_list_size=1000000)
        client.send_bytes(client.h2_connection.data_to_send())
        await client.wait_for_ack_settings()
        server.set_response(
            "HTTP/1.1 200 OK\r\n"
            + f"Date: {HttpMessage.date_time_string()}\r\n"
            + "Server: debian\r\n"
            + f"{'a'}: {'a' * 500000}\r\n"
            + "Content-Length: 0\r\n\r\n"
        )
        await client.send_request(self.post_request, "200")
```

fw/http_frame.c
```diff
@@ -304,15 +308,85 @@ static TfwSkbHooks tfw_h2_rst_frame_skb_hooks = {
...
+	if (unlikely(r))
+		goto err;
+
+	r = tfw_msg_iter_write(&it, data);
+	if (unlikely(r))
+		goto err;
+
+	ss_skb_setup_head_of_list(msg.skb_head, msg.skb_head->mark, tls_type);
+
+	if (ctx->cur_send_headers) {
```
EvgeniiMekhanik, line 359: We can't apply settings if ctx->cur_send_headers is not NULL.
I write simple test (in appropriate test branch) to test it

const-t: fixed in https://github.com/tempesta-tech/tempesta/pull/2628/changes/4c55392f464ee28f08b0b441990146072b707461

fw/http_frame.c
```diff
@@ -2409,59 +2631,98 @@ do {									\
...
-		}
 	}
+	WARN_ON(stream->xmit.bytes_to_send);
 
 	return r;
 
 #undef FRAME_XMIT_FSM_NEXT
-#undef CALC_FRAME_LENGTH_AND_SET_FRAME_TYPE
+#undef CALC_FRAME_LENGTH_AND_SET_FRAME_TYPE_OR_EXIT
 #undef ADJUST_BLOCKED_STREAMS_AND_EXIT
```
krizhanovsky, line 2764: There is no `ADJUST_BLOCKED_STREAMS_AND_EXIT` macro any more

fw/http_frame.c
```diff
@@ -835,24 +881,51 @@ tfw_h2_rst_stream_process(TfwH2Ctx *ctx)
...
+	 * should close all streams initiated by our side with identifier
+	 * higher than @last_id, and should not initiate new streams until
+	 * connection will be closed (see RFC 7540 section 5.4.1 and section
+	 * 6.8 for details).
+	 */
+	if (err_code)
+		T_DBG("HTTP/2 connection is closed by client with error code:"
+		      " %u, ID of last processed stream: %u\n", err_code,
+		      last_id);
+	WARN_ON(err_code == HTTP2_ECODE_COMPRESSION);
```
krizhanovsky, line 909: The error code is read directly from a client sent GOAWAY frame, so any client can trigger kernel warnings on our side. Probably this should be just `T_DBG()` that we could debug a real client problem if necessary

fw/http2.h
```diff
@@ -210,4 +227,37 @@ tfw_h2_conn_or_stream_wnd_is_exceeded(TfwH2Ctx *ctx, TfwStream *stream)
 	return ctx->rem_wnd <= 0 || stream->rem_wnd <= 0;
 }
 
+static inline int
+tfw_h2_settings_init_list(TfwSettingsList *settings, unsigned char length)
```
krizhanovsky, line 231:
```suggestion
tfw_h2_settings_init_list(TfwSettingsList *settings, int length)
```
as all callers pass `TfwFrameHdr->length` as the parameter value

fw/http_frame.c
```diff
@@ -739,43 +793,45 @@ static int
...
-
-		if (tfw_h2_increment_wnd_sz(window, wnd_incr)) {
-			err_code = HTTP2_ECODE_FLOW;
-			goto fail;
-		}
-
-		if (ctx->cur_stream)
-			tfw_h2_stream_try_unblock(&ctx->sched, ctx->cur_stream);
+	if (!wnd_incr) {
+		tfw_h2_conn_terminate(ctx, err_code);
```
krizhanovsky, line 800: Could you please add a code comment, like the commit message:
```
A zero WINDOW_UPDATE increment has no legitimate use and can be abused
to trigger repeated stream resets. Treat it as suspicious and terminate
the connection regardless of the target stream.
```
This is an important behavior documentation and it'd be good to also have it in code

fw/http_frame.c
```diff
@@ -2194,72 +2297,149 @@ tfw_h2_stream_send_postponed(struct sock *sk, struct sk_buff **skb_head,
...
+	return r;
+
+undo:
+	stream->xmit.b_len += frame_length;
+	return r;
+}
+
+/* Return true if socket has unknowledged or queued data */
+static bool
+__socket_has_inflight_data(struct sock *sk)
```
krizhanovsky, line 2399: Patch https://github.com/tempesta-tech/tempesta/pull/2628/changes/9164780d241403b40287616fa879c645076eb2c6 has deep technical insight int the commit message - it's worth leave it also as a comment in the code, at least a part of the comment.

fw/http_frame.c
```diff
@@ -1658,20 +1757,24 @@ tfw_h2_frame_recv(void *data, unsigned char *buf, unsigned int len,
...
 
-		if ((ret = tfw_h2_settings_process(ctx)))
+		if ((ret = tfw_h2_settings_process_entry(ctx)))
 			FRAME_FSM_EXIT(ret);
 
 		if (ctx->to_read)
 			FRAME_FSM_MOVE(HTTP2_RECV_FRAME_SETTINGS);
 
+		if (ctx->received_settings.num)
+			tfw_h2_apply_new_settings(ctx);
```
krizhanovsky, line 1776: Commit https://github.com/tempesta-tech/tempesta/pull/2628/changes/ced7430d2dea247b6ddcaabd60f2979b1067853d also describes quite subtle logic, which should be commented in the code as well. For example, it's quite not obvious that a client may clean the table with settings 0 - I'd say (without this explanation) that we're just wasting resources applying two consequent updates.

fw/http_frame.c
```diff
@@ -2409,59 +2669,98 @@ do {									\
...
-		}
 	}
+	WARN_ON(stream->xmit.bytes_to_send);
 
 	return r;
 
 #undef FRAME_XMIT_FSM_NEXT
-#undef CALC_FRAME_LENGTH_AND_SET_FRAME_TYPE
+#undef CALC_FRAME_LENGTH_AND_SET_FRAME_TYPE_OR_EXIT
 #undef ADJUST_BLOCKED_STREAMS_AND_EXIT
```
krizhanovsky, line 2764: There are no defines for the 3 `undef`s above any more

fw/http_frame.c
```diff
@@ -2379,26 +2374,66 @@ do {									\
...
-			 * If we there is no headers, data or trailer
-			 * frames to send, all data should be entailed
-			 * to the socket write queue at the beginning
-			 * of this state.
+			 * If Tempesta FW stop to make frames, because of exceeded
+			 * stream->rem_wnd, mark such stream as blocked.
 			 */
-			WARN_ON(stream->xmit.frame_length);
-			fallthrough;
+			WARN_ON(stream->xmit.is_blocked);
```
krizhanovsky, line 2392: `WARN_ON_ONCE()` - can't we get a flood of the warning in the current code?

krizhanovsky (approved):
LGTM, just several minor comments.

I'm wondering if you ran any performance benchmarks to compare this branch with `master`? If not, could you please do this and address performance regressions in #2686 if any.  Probably at least h2load with many streams should be ran, but a more sophisticated tests stressing the framing and streams changes would be good.

fw/http_frame.c
const-t: Replaced with `WARN_ON_ONCE()`. `T_DBG` is not useful here, we need to know that it happened in production. I'd like to have counter for such things that can be monitored.

fw/http2.h
const-t: We were safe with it, we had validation check for the size of the length. But I believe the new version is better and straightforward: https://github.com/tempesta-tech/tempesta/pull/2628/changes/c9b0ee061c0f10ccac66c0a36d588ce584638a67

fw/http_frame.c
const-t: Added comment

fw/http_frame.c
const-t: Added comment

fw/http_frame.c
const-t: Added comment

fw/http_frame.c
const-t: Even if it can't get a flood, it doesn't make sense to had a lot of the warnings. Changed to `WARN_ON_ONCE` and other `WARN_ON` as well.

const-t:
There is no difference. With bigger response result the same.

### PR 2628 content-length: 615
```
finished in 50.04s, 1665344.00 req/s, 1.26GB/s
requests: 83267200 total, 83277200 started, 83267200 done, 83267200 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 83267200 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 62.97GB (67617685300) total, 13.89GB (14909524609) headers (space savings 23.17%), 47.69GB (51209328000) data
                     min         max         mean         sd        +/- sd
time for request:      335us     67.20ms      7.30ms      3.99ms    66.84%
time for connect:     3.94ms     40.58ms     24.02ms      6.80ms    65.00%
time to 1st byte:     6.04ms     57.09ms     31.31ms      8.62ms    68.00%
req/s           :    6940.12    33145.62    16653.25     7767.86    68.00%
```

### PR 2628 content-length: 102400
```
finished in 50.09s, 87537.14 req/s, 8.37GB/s
requests: 4376857 total, 4386857 started, 4376857 done, 4376857 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 4376939 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 418.65GB (449524829701) total, 768.05MB (805356776) headers (space savings 22.69%), 417.41GB (448194095405) data
                     min         max         mean         sd        +/- sd
time for request:     2.92ms       1.78s    103.30ms     40.53ms    80.40%
time for connect:     8.13ms     49.08ms     24.83ms      7.54ms    69.00%
time to 1st byte:    13.74ms     63.86ms     35.75ms      9.68ms    69.07%
req/s           :       0.00     1183.65      896.43      271.04    90.00%
```

### master content-length: 615
```
finished in 49.04s, 1631510.58 req/s, 1.23GB/s
requests: 81575528 total, 81585529 started, 81575529 done, 81575529 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 81575528 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 60.69GB (66239353701) total, 13.60GB (14602019691) headers (space savings 23.18%), 46.72GB (50168950335) data
                     min         max         mean         sd        +/- sd
time for request:      432us     93.93ms      7.84ms      4.71ms    69.54%
time for connect:     5.50ms     43.38ms     25.21ms      6.89ms    70.00%
time to 0st byte:     8.50ms     66.55ms     33.37ms     10.61ms    68.00%
req/s           :    5362.10    40907.77    16314.91     8549.39    71.00%
```

### master content-length: 102400
```
finished in 49.08s, 87423.18 req/s, 8.36GB/s
requests: 4371158 total, 4381159 started, 4371159 done, 4371159 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 4371240 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 417.10GB (448935283643) total, 775.39MB (813050826) headers (space savings 22.50%), 416.87GB (447610791188) data
                     min         max         mean         sd        +/- sd
time for request:     1.60ms       2.22s     97.95ms     45.06ms    81.23%
time for connect:     3.29ms     51.31ms     25.54ms      8.66ms    74.00%
time to 0st byte:     9.94ms     64.92ms     37.35ms     10.50ms    68.37%
req/s           :       -1.00     1322.92      908.83      322.62    81.00%
```

PR #2636, opened 2026-03-23
Fix bug during restoring cache-control header from hpack.

EvgeniiMekhanik:
Previously there was a BUG_ON if we already restore cache-control header from hpack, but it was a mistake, because we check dublicate headers later in `tfw_http_msg_hdr_close` after restoring it. So all requests which contain dublicate cache-control headers, like cache-control: max-age=1
cache-control: max-age=1
trigger this BUG_ON. Fix this problem, now we immediately return T_DROP and drop request, with such dublicate headers.

### DIFF
--- a/fw/http_parser.c
+++ b/fw/http_parser.c
@@ -7402,6 +7402,14 @@ do {									\
 	parser->cstate.is_set = 1;					\
 } while(0)
 
+/*
+ * RFC 9111 does not explicitly define the behavior of caches when multiple
+ * identical Cache-Control directives are present within a single request
+ * (e.g., "Cache-Control: max-age=1, max-age=5").
+ * Empirical testing of the Apache HTTP Server indicates that, in such cases,
+ * it prioritizes the last occurrence of the directive (in this example,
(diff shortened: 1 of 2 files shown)


fw/http_parser.c
```diff
@@ -7451,19 +7451,28 @@ do {									\
...
-	       (req->cache_ctl.flags & TFW_HTTP_CC_STALE_IF_ERROR));
+	/*
+	 * Usually we check dublicate headers in `tfw_http_msg_hdr_close`
+	 * but there there is no sence to continue processing request
+	 * here, drop it immediately.
+	 */
+	if ((req->cache_ctl.flags & TFW_HTTP_CC_MAX_AGE) ||
+	    (req->cache_ctl.flags & TFW_HTTP_CC_MAX_STALE) ||
+	    (req->cache_ctl.flags & TFW_HTTP_CC_MIN_FRESH) ||
+	    (req->cache_ctl.flags & TFW_HTTP_CC_STALE_IF_ERROR))
```
krizhanovsky: The parser is hot, let's make it `unlikely`

PR #2644, opened 2026-04-21
Fix bug in case of `sock_srv` module failed to start.

EvgeniiMekhanik:
During `tfw_sock_srv_stop` we call both `tfw_sg_for_each_srv_reconfig` and `tfw_sg_for_each_srv` for disconnection. If disconnect fails we also call `tfw_sg_for_each_srv` to abort connections. We should also call `tfw_sg_for_each_srv_reconfig` to abort.


const-t (approved):
LGTM

PR #2650, opened 2026-05-03
Fix according review

EvgeniiMekhanik:
- Rename `client_mem_cache` to `tfw_client_mem_cache`.
- Change `in_free_list` from struct list_head to pointer to save 8 bytes.
- Implement 1GB mem limit by default.


EvgeniiMekhanik:
Create new branch with the same commits https://github.com/tempesta-tech/tempesta/pull/2651

PR #2651, opened 2026-05-04
Last fixes according 1715

EvgeniiMekhanik:


EvgeniiMekhanik:
The name of branch is euqal to the appropriate branch in tempesta-test repository.

const-t (approved):
LGTM

fw/t/unit/test.c
```diff
@@ -91,6 +91,12 @@ extern TEST_SUITE_MPART_ARR(http2_parser, H2_SUITE_PART_CNT);
 extern int tfw_pool_init(void);
 extern void tfw_pool_exit(void);
```
krizhanovsky, line 92: This is a new pull request, so please:
1. name the branches according to our guides
2. provide some meaningful description in the title of the PR and it's description

This PR is about #1715. It does not "fixes" #1346. It is from you, not Jinhua.

Maybe we need the branch name, but the title could must be more meaningful

fw/client.h
```diff
@@ -28,7 +28,7 @@ typedef struct tfw_client_mem_t {
 	struct percpu_ref	refcnt;
 	struct work_struct	kill_work;
    	long __percpu		*mem;
-	struct list_head	in_free_list;
+	struct tfw_client_mem_t	*next_free;
```
krizhanovsky: I didn't get your comment:

> We can't do it! Look we allocate and initialize ALL other members only once during tfw_cli_mem_pool_init! In tfw_cli_mem_pool_alloc we only get TfwCLientMem from free_list! So all this fields should be separate!

While a `TfwClientMem` block is in free list, then only `next_free` pointer is used - we do not need the rest of the data structure. Once a block is allocated, we do not need `next_free` and can initialize it and reuse the memory. Initialization is just 8 byte write and the bytes are already in the L1 cache - we can initialize the part of the structure on each allocation.

Also usually list members go first to prefetch the data structure on the list dereference.

Issue #2284, opened 2024-11-19
drop connection data if a client has closed the connection via RST\FIN


RomanBelozerov:
# Motivation

A large memory consumption was noticed when testing [38](https://github.com/tempesta-tech/tempesta-test/issues/38) issue (DDoS mitigation testing). One of the problems that requires investigation - getting TCP FIN\RST in a connection while processing data of this connection. TempestaFW should drop connection data and free up memory as soon as possible. Nginx has 499 status code in access log for this case.

Case example:

1. a client open http1\h2 connection and send a request;
2. TempestaFW forwards the request to a upstream or finds a response in cache;
3.  the client sends TCP FIN\RST before sending the response from TempestaFW;

Expected behaviour - TempestaFW drops connection data and free up memory as soon as possible. Probably, we cannot drop connection data just like that because we must count the number of requests for frang limits (`request_rate` and `request_burst`).

# Testing

Please create a test issue or add http1\h2 tests to tempesta-test.

# Documentation

We should add info about this behaviour in wiki

krizhanovsky:
Crucial since we can't really fight against L7 DDoS with an inefficient `block action reply`. 0.9 release must provide efficient DDoS protection, so move to 0.9

PR #2654, opened 2026-05-06
Mekhanik evgenii/fix 2284

EvgeniiMekhanik:


fw/connection.h
```diff
@@ -438,16 +438,10 @@ tfw_connection_live(TfwConn *conn)
...
-tfw_connection_get_many(TfwConn *conn, int cnt)
-{
-	atomic_add(cnt, &conn->refcnt);
-}
-
 static inline void
 tfw_connection_get(TfwConn *conn)
 {
-	tfw_connection_get_many(conn, 1);
+	atomic_add(1, &conn->refcnt);
```
const-t:
```suggestion
	atomic_inc(&conn->refcnt);
```

fw/sock.c
```diff
@@ -562,20 +556,17 @@ ss_do_send(struct sock *sk, struct sk_buff **skb_head, int flags)
...
 
-	/* If the socket is inactive, there's no recourse. Drop the data. */
-	if (unlikely(!conn || !ss_sock_active(sk)))
-		goto cleanup;
-
 	ss_skb_setup_head_of_list(*skb_head, (*skb_head)->mark, tls_type);
 
-	if (ss_skb_on_send(conn, skb_head))
+	if (ss_skb_on_send(likely(send_is_allowed) ? conn : NULL, skb_head))
 		goto cleanup;
```
const-t, line 586: It seems this is single jump to `cleanup`, it would be better to remove the label and call `ss_skb_queue_purge()` here.

fw/sock.c
```diff
@@ -508,16 +507,23 @@ ss_skb_tcp_entail_list(struct sock *sk, struct sk_buff **skb_head,
...
+
 		/*
 		 * Zero-sized SKBs may appear when the message headers (or any
 		 * other contents) are modified or deleted by Tempesta. Drop
 		 * these SKBs.
+		 * If `on_tcp_entail` callback fails we also free all skbs from
+		 * the list.
 		 */
-		if (!skb->len) {
+		if (skip_list || !skb->len) {
```
const-t, line 529: Looks like we may partially destroy the queue if snd_wnd is 0 and then call `ss_skb_on_tcp_entail()` again, that looks unnecessary.

IIUC the logic is: When `ss_skb_on_tcp_entail()` returns error code we set `skip_list = true` and trying to destroy all skbs until we reach skb with `is_head` set, that indicates beginning of another message. However if `snd_wnd` is zero we break the loop and set `is_head` for the next skb and set `skip_list = false`, as a result we destroyed only part of the message and we continue to send message and call `ss_skb_on_tcp_entail()`.

I believe if `skip_list` is set, we need to destroy all skbs that belongs to the message regardless snd_wnd. 

By the way I would suggest to not count `snd_wnd` on each iteration, only calc it before calling `ss_skb_tcp_entail_list()` and pass it to the function. In the loop just subtract `skb->len` from `snd_wnd`. It seems calling of `tfw_tcp_calc_snd_wnd()` on each iteration doesn't not introduce additional accuracy due to `tfw_tcp_calc_snd_wnd()` doesn't know how many bytes are in write queue, but introduce overhead for re-calculation on each iteration.

fw/sock.c
```diff
@@ -527,8 +533,9 @@ ss_skb_tcp_entail_list(struct sock *sk, struct sk_buff **skb_head,
 
 		r = ss_skb_realloc_headroom(skb);
 		if (unlikely(r)) {
-			ss_skb_queue_head(skb_head, skb);
-			goto restore_sk_write_queue;
+			printk(KERN_ALERT "BADBAD!!!!\n");
```
const-t: Garbage

fw/ss_skb.c
```diff
@@ -1799,3 +1799,11 @@ ss_skb_adjust_client_mem(struct sk_buff *skb, int delta)
 		tfw_client_adjust_mem(cli_mem, delta);
 	}
 }
+
+void
+ss_skb_copy_cb(struct sk_buff *to, struct sk_buff *from)
+{
+	memcpy(&TFW_SKB_CB(to)->copy, &TFW_SKB_CB(from)->copy,
```
const-t: Why do we need this additional copy? `__pskb_copy_fclone()` copies `cb`. It seems we only need to zero it. However if we zero the source skb, what happens if this skb will be re-forwarded due to re-scheduling?

fw/http.c
```diff
@@ -1373,6 +1352,165 @@ tfw_http_req_nip_delist(TfwSrvConn *srv_conn, TfwHttpReq *req)
...
+
+	if (unlikely(!tfw_http_req_is_valid(req))) {
+		spin_lock_bh(&srv_conn->fwd_qlock);
+		/*
+		 * Should be called before removing request from the
+		 * server connection queue.
+		 */
+		if ((TfwMsg *)req == srv_conn->last_msg_sent) {
+			srv_conn->last_msg_sent =
+				__tfw_http_conn_last_msg_sent_prev(srv_conn);
```
const-t, line 1453: I didn't get the logic. This callback may be called `tfw_http_on_tcp_entail_req()` in the middle of sending request into the connection, but we doing some queue management instead of closing connection. Why? Having this only for checking request validity even before sending the first segment also looks unnecessary.

const-t (changes requested):

fw/http.c
EvgeniiMekhanik: No! Callback is set only for `skb_head` and called only once when we entail it. If we start to send some data for request we will send it. So we can remove request only when we try to entail it's first skb.

fw/http.c
```diff
@@ -4170,38 +4265,99 @@ write_merged_cookie_headers(TfwStr *hdr, TfwMsgIter *it)
...
+
+static int
+tfw_h1_req_cutoff_headers(TfwHttpReq *req, TfwHttpMsgCleanup *cleanup)
+{
+	TfwHttpMsg *hm = (TfwHttpMsg *)req;
+	int r;
+
+	if (TFW_STR_EMPTY(&req->body)) {
+		/* Clean current request skb_head if doesn't body exists. */
+		r = tfw_http_msg_set_empty_skb_head(hm, cleanup);
```
const-t: I would like to suggest to not free head_skb if there is no body, we can clean it and re-use, to avoid allocation  in `tfw_http_msg_set_empty_skb_head()`.
```suggestion
		struct sk_buff *skb_free;
		TfwMsgIter *it = &hm->iter;

		/* Clean the current request's skb_head if no body exists. */
		if (skb_headlen(it->skb)) {
			ss_skb_put(it->skb, -skb_headlen(it->skb));
			it->skb->tail_lock = 1;
		}

		__tfw_http_msg_rm_all_frags(it->skb, cleanup);

		skb_free = it->skb_head->next;
		while (skb_free != it->skb_head) {
			ss_skb_unlink(&it->skb_head, skb_free);
			ss_skb_queue_tail(&cleanup->skb_head, skb_free);
			skb_free = it->skb_head->next;
		}
```

const-t (approved):
LGTM, but I left small suggestion.

fw/sock.c
```diff
@@ -546,19 +544,6 @@ ss_skb_tcp_entail_list(struct sock *sk, struct sk_buff **skb_head,
...
-	to_destroy = tail ? tail->next : tcp_send_head(sk);
-	if (to_destroy) {
-		tcp_for_write_queue_from_safe(to_destroy, next, sk) {
-			tcp_unlink_write_queue(to_destroy, sk);
-			tcp_wmem_free_skb(sk, to_destroy);
-		}
-	}
-	if (*skb_head && !TFW_SKB_CB(*skb_head)->is_head)
-		ss_skb_setup_head_of_list(*skb_head, mark, tls_type);
-
```
krizhanovsky, line 552: @const-t @EvgeniiMekhanik I had a look onto b68d2ce8ce95373686c861515317f9e18af28ce4 , which introduced the code and IIRC this also fixed the case when we push a message into a socket and fail somewhere on a middle skb, so a partial message (broken data) is in the write queue. I'm I wrong and is it safe to remove the code now?

fw/ss_skb.c
```diff
@@ -1763,7 +1748,7 @@ void
 ss_skb_dflt_destructor(struct sk_buff *skb)
```
krizhanovsky: Now it's just a regular function, not a destructor and not 'default', any more, so let's rename it to something more appropriate

krizhanovsky (changes requested):
The pull request is probably fine, I reviewed only a couple of patches from it. The only reason for the reject is that I believe we need to adjust our process for so heavy code modifications. @const-t @EvgeniiMekhanik I want to have a discussion on this, probably on the upcoming meeting.

The main problem is that the whole PR is huge having 9 commits of changes from the skb layer to HTTP, reworked callback architecture and so on. The pull request is an implementation of #2284 feature. However, there is no any single commit, message or comment neither in the code nor in the issue or pull request about how all the 9 commits contribute to the problem solution (only 1d9a1d19042aa1282582c410bf0edb01c72b15bf is an exception with a clear purpose).

For example, https://github.com/tempesta-tech/tempesta/pull/2654/changes/1f9c14d64762df62f8c88974c7711de24ba31149 : 

> Rework HTTTP/2 -> HTTP/1 request transformation

why do we need this to solve #2284?

> Use pool-based approach for HTTP/2 to HTTP/1 request transformation.

Maybe it's fine

> Reuse the same pool-based approach already used for HTTP/1 request forwarding, providing a unified implementation.
> Eliminate extra header copying. Previously, headers were written into the linear skb area and then copied by `_pskb_copy_fclone`. Now headers are written directly into skb fragments, reducing memory usage and copy overhead.

I can review the change and see that the code is correct, doesn't hurt performance and doesn't impact security. However, why the change is required? Which problem it solves? It doesn't clear now and will not clear in a year to review what for this code change was made.

Please, in the pull request write a general design for the #2284 feature and describe each of the commits as the steps reaching the design goal.

*Why this is important*
1. The big changes change the project architecture and the whole team must have clear and aligned understanding of the changes. We must review not only the code itself, but the design and architecture.
2. Mature complex projects always have questions "why this code is written this way?" and the current state of the commits doesn't answer this question.

@ykargin @artem98 FYI - I think you also need to keep this in mind and integrate on your projects.

fw/sock.c
EvgeniiMekhanik: Previously, we considered handling ENOMEM in a special way without tearing down the connection, later we decide to don't do it and immediately drop connection. For example in `tfw_tls_encrypt` we do not clean up the socket queue on error, potentially leaving broken data behind. This is currently acceptable because the connection is immediately terminated and the entire queue is subsequently cleaned up.

fw/sock.c
```diff
@@ -527,30 +536,28 @@ ss_skb_tcp_entail_list(struct sock *sk, struct sk_buff **skb_head,
...
 		r = ss_skb_realloc_headroom(skb);
 		if (unlikely(r)) {
-			ss_skb_queue_head(skb_head, skb);
-			goto restore_sk_write_queue;
+			ss_kfree_skb(skb);
+			return r;
 		}
 
 		ss_skb_tcp_entail(sk, skb, mark, tls_type);
+		*snd_wnd = (*snd_wnd > skb->len ? *snd_wnd - skb->len : 0);
```
krizhanovsky, line 543: GPT-5.6 review: `ss_skb_tcp_entail()` -> `ss_skb_try_collapse()` -> `kfree_skb_partial()` -> `__kfree_skb()` frees the `skb`, so at this line we're resolving dead `skb` with `skb->len`.

fw/ss_skb.h
```diff
@@ -28,56 +28,55 @@
...
+typedef struct {
+	on_send_cb_t 		on_send;
+	on_send_fail_cb_t	on_send_fail;
+	on_tcp_entail_cb_t	on_tcp_entail;
+} TfwSkbHooks;
+
+/*
+ * Tempesta FW sk_buff private data.
+ * @cli_mem 		- pointer to TfwClientMem structure for memory
+ *			  accountion;
```
krizhanovsky:
```suggestion
 *			  accounting;
```

fw/http.c
```diff
@@ -4241,7 +4241,7 @@ tfw_h2_req_set_loc_hdrs(TfwHttpReq *req, TfwHdrMods *h_mods)
  * Fuse multiple cookie headers into one.
  * Works only with TFW_STR_DUP strings. */
```
krizhanovsky: Please adjust the comment style

krizhanovsky (approved):
Quite a sophisticated pull request, excellent work! I appreciate the detailed commit messages and commenting - it's much easier to go over such a hard logic. Performance optimizations are also always good. Well done!

There is just a trivial bug found by GPT-5.6 review.

PR #2655, opened 2026-05-09
Design of #1346

krizhanovsky:
This PR is copy of the [1346 design proposal](https://github.com/tempesta-tech/tempesta/issues/1346#issuecomment-4353537980) that we can all discuss in comments and work out a final design.


1346-design.md
```diff
@@ -0,0 +1,167 @@
...
+ * @pending_cpu	- accumulated raw CPU time (in ns) since the last EMA
+ *		  update. This value is periodically folded into @ema;
+ */
+typedef struct {
+    u64 last_ts;
+    s64 ema;
+    u64 pending_cpu;
+} TfwCpuEma;
+```
+Save time at the beginning of SoftIRQ shot and  check CPU usage at the end of SoftIRQ shot (to prevent perfomance regression in case when we do it on each request) .
```
krizhanovsky: Is this `begin_time`?

In one softirq shot we process many requests - how can we apply `begin_time` to all of them?

I think `begin_time` should be the time of receiving an `skb`. We can save the time somewhere (e.g. in a static per-cpu variable) - when we get an `skb` we do not know the client. But we need to call `tfw_client_update_cpu_ema()` not only on forwarding an HTTP message, but also on error responses. At all these calls we should know the socket and `TfwClient`.

1346-design.md
```diff
@@ -0,0 +1,167 @@
...
+**CPU Tracking**
+In addition to `TfwTrainingStat` implement structure and per-cpu array of this structures.
+```C
+/**
+ * Exponential moving average (EMA) tracker for per-CPU time usage.
+ *
+ * The structure is used to accumulate execution time deltas and maintain
+ * a smoothed estimate (EMA) of CPU consumption.
+ *
+ * @last_ts	- timestamp of the last update (in ns). Used to compute
```
krizhanovsky: Isn't it expensive to get time with ns accuracy?

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+Each request tagged with `training_epoch` to prevent mixing old and new training data (we add new field to `request` structure and save epoch in this field). When request removed from server connection queue we don't update `curr` field in case when request belongs to previous epoch. (When request added to server connection queue it always belongs to new epoch if trainging enabled!).
+
+**Current method and alternatives**
+The same problems and altgernatives as for connections.
+
+**CPU Tracking**
+In addition to `TfwTrainingStat` implement structure and per-cpu array of this structures.
+```C
+/**
+ * Exponential moving average (EMA) tracker for per-CPU time usage.
```
krizhanovsky: Good idea, I think EMA should work well here

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+		decay = 1 << SCALE_SHIFT;
+	cpu_ema->ema = cpu_ema->ema *
+		((1 << SCALE_SHIFT) - decay) >> SCALE_SHIFT;
+	cpu_ema->ema += ((s64)usage - (s64)cpu_ema->ema) >> ema_alpha_shift;
+}
+```
+Pass `delta = new_ema - prev_ema` to  `tfw_client_training_adjust_cpu_num` which do the same as ` `tfw_client_training_adjust_req_num`.
+
+**Defence mode**
+In defence mode use `delta_ema` on each SoftIRQ shot to calculate `z = (delta_ema - mean) / std` and if calculated `z > threshold` reject connection with TCP RST and block client by IP if necessary.
```
krizhanovsky: Say we process requests for 1K clients in one SoftIRQ shot, then all of them will use the same `begin_time` and different `now` timestamp - the last client processed has the larges CPU time and the first one the lowest. This a computation bug.

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * The update is performed only if enough time (@min_time_to_adjust)
+ * has passed since the previous update to avoid excessive noise and
+ * high-frequency recalculations.
+ *
+ * The function:
+ *   - computes elapsed time (@dt);
+ *   - converts accumulated CPU time into normalized usage value;
+ *   - applies time-based decay (older history loses weight);
+ *   - updates EMA using a combination of decay and smoothing factor.
```
krizhanovsky: The 2 bulets above are just the idea of EMA, right?

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+ */
+static void
+tfw_client_update_cpu_ema(TfwCpuEma *cpu_ema, u64 delta_cpu)
+{
+	u64 now = ktime_get_ns();
```
krizhanovsky: OK, this one looks fast enough

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+	static const unsigned int ema_alpha_shift = 4;
+
+	cpu_ema->pending_cpu += delta_cpu;
+	if (unlikely(dt < min_time_to_adjust))
+		return;
+
+	cpu_ema->last_ts = now;
+	swap(cpu_ema->pending_cpu, total_cpu);
+	usage = (total_cpu << SCALE_SHIFT) / dt;
+	decay = (dt << SCALE_SHIFT) / time_to_forget_ns;
```
krizhanovsky: What is `SCALE_SHIFT`?

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+ *		  time delta between consecutive measurements;
+ * @ema		- current exponential moving average of CPU usage;
+ * @pending_cpu	- accumulated raw CPU time (in ns) since the last EMA
+ *		  update. This value is periodically folded into @ema;
+ */
+typedef struct {
+    u64 last_ts;
+    s64 ema;
+    u64 pending_cpu;
+} TfwCpuEma;
```
krizhanovsky: This is per TfwClient, right?

1346-design.md
krizhanovsky: Depending on `SCALE_SHIFT` we likely to get always 0 here, since we pass to this code every 1000ns and divide for 0.1s

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+	cpu_ema->last_ts = now;
+	swap(cpu_ema->pending_cpu, total_cpu);
+	usage = (total_cpu << SCALE_SHIFT) / dt;
+	decay = (dt << SCALE_SHIFT) / time_to_forget_ns;
+
+	if (decay > (1 << SCALE_SHIFT))
+		decay = 1 << SCALE_SHIFT;
+	cpu_ema->ema = cpu_ema->ema *
+		((1 << SCALE_SHIFT) - decay) >> SCALE_SHIFT;
+	cpu_ema->ema += ((s64)usage - (s64)cpu_ema->ema) >> ema_alpha_shift;
```
krizhanovsky: In the original EMA old EMA value and the new observation are summed with coeficients `alpha` and `1 - aplha`, but in your case `decay` is independent from `alpha` - why?

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+```
+Pass `delta = new_ema - prev_ema` to  `tfw_client_training_adjust_cpu_num` which do the same as ` `tfw_client_training_adjust_req_num`.
+
+**Defence mode**
+In defence mode use `delta_ema` on each SoftIRQ shot to calculate `z = (delta_ema - mean) / std` and if calculated `z > threshold` reject connection with TCP RST and block client by IP if necessary.
+
+**Current method and alternatives**
+
+**Alternatives**
+1. Use raw CPU time
```
krizhanovsky: What 'raw CPU' time means exactly?

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+* ✔ simple
+* ✔ accuracy
+* ❌ very noisy
+* ❌ strong peaks
+* ❌ Bad normalization
+2. Sliding window average (store CPU usage for the last N ms)
```
krizhanovsky: And how it's better/worse than EMA?

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+3. Use `ema` directly. Currently we measure change, not level (constant high CPU → delta ≈ 0 → no detection).
```
krizhanovsky: This is unclear

1346-design.md
```diff
@@ -0,0 +1,169 @@
+**Training Mode:**
+Collects per-client metrics on each new event aggregates global statistics: number of samples (num of clients),  sum of values (sum), sum of squares (sumsq). At the end of training computes mean and standard deviation, which will be used for z-score calculation in defence mode.
```
krizhanovsky: The training mode must also collect global mean and standard deviations. These values
are used in protection mode.

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+
+**Defence Mode**
+Each new observation is evaluated using z=std / (x−mean). If z > configured_threshold he event is considered anomalous.
```
krizhanovsky: Z-score computed in reverse ratio: `z = (x - mean) / std`. Otherwise you get zeros
for integer computations. BTW it make sense to additionally intoruduce
multiplications, like 1024, to get getter precision of z-score (also multiply the
configured values): `z = |x - mean| * 1024 / std`. Also note the modulus - you can do
this right in z-score computation or on comparison with configuration. I'd prefer the
first be able to handle larger values and/or keep the values in smaller memory.

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+
+**Training mode**
+`conn_curr` is incremented/decremented.
+Track maximum concurrent connections (`conn_max`). When max increases - compute `delta1 = new_max - old_max` and `delta2 = new_max² - old_max²` and use this values to update `sum` and `sumsq`.
+
+**Defence mode**
+Track `conn_curr` on each new opened connection. Calculate `z = (conn_curr - mean) / std` if `z > threshold` reject connection and block client by IP if necessary.
+
+**Epoch handling**
+Each connection tagged with training_epoch to prevents mixing old and new training data (we add new field to `tempesta_sock` and save epoch in this field). When connection closing we don't update `conn_curr` in case when connection belongs to previous epoch. (When connection is opening it always belongs to new epoch if trainging enabled!).
```
krizhanovsky: For which computations do we need `conn_training_epoch` in `TfwClient` and `epoch` in
`TfwTrainingStat`? I din't get what is old and new training data and why it's bad to mix them

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+
+2. Z-score on max, check only max
+* ✔ Fully consistent model
+* ✔ Low false positives
+* ❌ Slow reaction
+* ❌ Attack can stay just below max
+
+In all this cases we don't take into account time awareness (100 connections for 1 second, 100 connections for 1 hour).
+May be it will be good to calculate z-score for connection rate also (or we rely on frang for this case?).
+
```
krizhanovsky, line 616: We already account memory for TfwClient and we also should implement the z-score
logic for memory as weel. The reason is that it could be hard to set the memory limit
or it can be too permissive. Also this is transition to #488 for adaptive client
classification. This shoul work as the current client_mem must never be reached, but
depending on the cimputed values during training mode, a client can be blocked on
smaller memory consumption.

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+Track maximum concurrent connections (`conn_max`). When max increases - compute `delta1 = new_max - old_max` and `delta2 = new_max² - old_max²` and use this values to update `sum` and `sumsq`.
+
+**Defence mode**
+Track `conn_curr` on each new opened connection. Calculate `z = (conn_curr - mean) / std` if `z > threshold` reject connection and block client by IP if necessary.
+
+**Epoch handling**
+Each connection tagged with training_epoch to prevents mixing old and new training data (we add new field to `tempesta_sock` and save epoch in this field). When connection closing we don't update `conn_curr` in case when connection belongs to previous epoch. (When connection is opening it always belongs to new epoch if trainging enabled!).
+
+**Current method and alternatives**
+In current approach during trainging mode we track maximum concurrent connections per client, in defence mode we compare current connections count (`conn_curr`) against a distribution of per-client maximum.
```
krizhanovsky: The idea was to compare `conn_curr` for each of the clients against z-score for
concurrent connections. Not against per client maximum. We should do comparison on
each new connection established. Once a connection is closed, we decrement the
counter.

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+* ✔ effective for burst detection
+* ❌ max vs current mismatch
+* ❌ sensitivity to outliers (one client opens 10000 connections other clients 1 connection during trainging.)
+* ❌ Not good against syn flood. If client open/close connections very fast, such client will not be blocked, because current connection count will be low.
```
krizhanovsky: The 4 pros and cons/alternatives sections are unclear for me (AI generated?)
- they are too brief and I don't have enough context, e.g. what "distribution" is, "consistent model" - what consistent with what?

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+* ❌ Noisy signal
+* ❌ Dominated by low values (many idle clients, if they close there connections durig training)
+
+2. Z-score on max, check only max
+* ✔ Fully consistent model
+* ✔ Low false positives
+* ❌ Slow reaction
+* ❌ Attack can stay just below max
+
+In all this cases we don't take into account time awareness (100 connections for 1 second, 100 connections for 1 hour).
```
krizhanovsky: This task is about resource stress, so "100 connections for 1 second" should be
caught typically via CPU usage.

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+ * @epoch	- training epoch identifier. Compared against the global
+ *		  @g_training_epoch to detect epoch change and trigger
+ *		  reinitialization of @max and @curr.
+ */
+typedef struct {
+	atomic64_t	max;
+	atomic64_t	curr;
+	spinlock_t	lock;
+	unsigned int	epoch;
+} TfwTrainingStat;
```
krizhanovsky: With atomics and spin lock in `TfwTrainingStat` I assume this data structure is going
to be frequently updated from all CPUs, so we're going to get hard contention and
false sharing. We either need to account statistics per-cpu and later merge it or
process TfeClient per-CPU using the message bus. It's not necessary to implement the
last one (if we decide to go this way), but it must be well designed at this moment
the code must ne developed acoring to the future changes to not to rewrite the code
later.

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+```
+We use new implemented function `tfw_client_training_adjust_req_num` both for training and  defence mode.
```
krizhanovsky: and what the function does?

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+**Disabled Mode**
+Internal state used during transitions. Ensures safe updates of shared data (via RCU synchronization). May be it's better to implement this state also, not only as internal state, to prevent any additional calculations, when it is not necessary.
+
+**Connection Count Tracking**
+In`TfwClient` structure we additionally store `unsigned int conn_max`, `int conn_curr` and `unsigned int conn_training_epoch`. We don't need any lock here, because all this fields updated under private `ra->lock` in frang.
+We use new implemented function `tfw_client_training_adjust_conn_num` both for training and  defence mode.
+
+**Training mode**
+`conn_curr` is incremented/decremented.
+Track maximum concurrent connections (`conn_max`). When max increases - compute `delta1 = new_max - old_max` and `delta2 = new_max² - old_max²` and use this values to update `sum` and `sumsq`.
```
krizhanovsky: What `sum` and `sumsq` are? If it is mean and standard deviation, then the
computation is wrong for Welford:
```
n += 1
delta  = new_max - mean
mean  += delta / n
delta2 = new_max - mean
M2    += delta * delta2
```

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+	unsigned int	epoch;
+} TfwTrainingStat;
+```
+We use new implemented function `tfw_client_training_adjust_req_num` both for training and  defence mode.
+
+**Training mode**
+Track `curr` - current in-flight non-idempotent requests. Increment `curr` in `tfw_http_req_enlist`, decrement in  `tfw_http_req_nip_delist`. Also track `max` maximum count  in-flight non-idempotent requests per client. When max increases update global trainging stats, same as we do it for connections (`delta1 = new_max - old_max` and `delta2 = new_max² - old_max²`).
+
+**Defence mode**
+Change signature for `tfw_http_req_enlist` from `void` to `bool`.  Call `tfw_client_training_adjust_req_num` on each new non-idempotent request, calculate z-score, return false if `z > threshold`. `tfw_http_req_enlist` is called from `tfw_http_req_fwd` and `tfw_http_req_fwd_resched`, this functions now return T_BLOCK if `tfw_http_req_enlist` fails.
```
krizhanovsky: No, we compute z-score only on training mode. In defence (protection) mode we only
compare computed value with the current number of indempotent requests in flight.

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+Callers of `tfw_http_req_fwd` and `tfw_http_req_fwd_resched` send 403 error response, drop client connection with TCP RST and block client by IP if these functions return T_BLOCK.
+
+**Epoch handling**
+Each request tagged with `training_epoch` to prevent mixing old and new training data (we add new field to `request` structure and save epoch in this field). When request removed from server connection queue we don't update `curr` field in case when request belongs to previous epoch. (When request added to server connection queue it always belongs to new epoch if trainging enabled!).
```
krizhanovsky: Why do we need this? IIUC this is for the case when `net.tempesta.training` is
changed many times, i.e. there are many transitions between training and protection
modes (maybe with disabled as well). It seems this is a sophistication just to not to
start training from absolute zero, but use requests in flight. Probably, this is not
so big win to make the sophistication, at least in the first implementation.

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+
+**Current method and alternatives**
+The same problems and altgernatives as for connections.
+
+**CPU Tracking**
```
krizhanovsky: If we have cheap and precise enough nanoseconds time, then the current proposal should work. Meantime, I want to propose an alternate or additional change to block malicious users by CPU consumption.

Rework `http_body_chunk_cnt` and `http_header_chunk_cnt` limits as it's hard to unify the values for many-headers long messages and short-headers short-body messages.

Instead we need to detect artificially lowered chunk sizes for HTTP/1 and DATA and CONTINUATION frames for HTTP/2.

We can do this with learning average DATA and CONTINUATION frame size in HTTP/2 and/or data chunk (skb-carried, not HTTP chunk) for both HTTP/1 and HTTP/2.

We should accounb the average (for training and protection modes)  ONLY for multi-chunk messages. I.e. if a message has zero or 1 CONTINUATION or DATA, then we do not compute the average for it.

We learn and analyze average `chunk` size, where `chunk` is a CONTINUATION or DATA frame size for HTTP/2 or skb data chunk in HTTP/1 (not an HTTP chunk size). It's is essentially `total_size / chinks_number`, where `total_size` is the total body or headers size.

The average chunk size is about kilobyte, maybe several kilobytes (with GRE) and we need to catch extremely small chunk sizes. Not only that (it's probably OK to have several occasional small chunks), but when a client sends a lot of small chunks. I.e. I propose to learn and detect multiplication of `N / average_chunk_size * chunks_number` - this feature should have high deviations for normal and attacking connections.

In comparison with the current  `http_body_chunk_cnt` and `http_header_chunk_cnt` limits:
1. this scheme normally handles small messages consisting of 1 small chunk
2. normally handles large messages consisting of many large chunks
3. trigger on large messages consisting on many small chunks
4. these parameters are learnt from traffic and we don't need to specify the hard-to-define limits

krizhanovsky (commented):
I have a lot of comments and questions

1346-design.md
```diff
@@ -0,0 +1,169 @@
...
+	swap(cpu_ema->pending_cpu, total_cpu);
+	usage = (total_cpu << SCALE_SHIFT) / dt;
+	decay = (dt << SCALE_SHIFT) / time_to_forget_ns;
+
+	if (decay > (1 << SCALE_SHIFT))
+		decay = 1 << SCALE_SHIFT;
+	cpu_ema->ema = cpu_ema->ema *
+		((1 << SCALE_SHIFT) - decay) >> SCALE_SHIFT;
+	cpu_ema->ema += ((s64)usage - (s64)cpu_ema->ema) >> ema_alpha_shift;
+}
```
krizhanovsky, line 737: The math is always easy to make wrong computations, so strongly propose to start from a unit test showing the algorithm behavior on different data, see for example `t/unit/user_space/percentiles.c`

1346-design.md
const-t: The same, didn't get the idea. Also I believe we don't consider to use Tempesta against L4 attacks?

1346-design.md
EvgeniiMekhanik: I think that we can do it same as we do it for `client_mem`. We save begin_time at the beginning of `ss_tcp_process_data` and check at the `connection_recv_finish` callback. For client mem we do it to prevent performance degradation, I think for CPU we can do the same.

EvgeniiMekhanik: Socket is known in ss_tcp_process_data, we can get client from sk_user_data (connection)->client same as we do for client mem.

EvgeniiMekhanik: For responses the different case we can do it in `process_resp` function same as we do for client_mem

1346-design.md
EvgeniiMekhanik: Yes

1346-design.md
EvgeniiMekhanik: spin lock is used only for protection during trainging epoch changing

1346-design.md
const-t: > OK, this one looks fast enough

This is fast enough only for one call per SoftIRQ shot. Calling this function often will cause performance degradation.

1346-design.md
const-t: And this is important note. It implies that in training mode we can only compute local values and merge them when we finished processing of the current client. We can use per-cpu counters. But in defence mode we can collect and sum all per-cpu counters in the beginning of processing the client and cache it to compare with z-score even for each request.

1346-design.md
const-t: Why nano seconds why not cpu cycles?

1346-design.md
EvgeniiMekhanik: All requests belongs to the same client (we process only one socket during `ss_tcp_process_data` ). Yes it is not accuracy, but we do the same for `client_mem` to prevent performance degradation.

1346-design.md
EvgeniiMekhanik: yes

1346-design.md
EvgeniiMekhanik: I will try to change it jiffies and check is it ok.

1346-design.md
EvgeniiMekhanik: It is for more accuracy calculation. There is no float calculation in kernel, so we use SCALE_SHIFT for more accuracy

1346-design.md
EvgeniiMekhanik: do not use EMA, use directly time difference

1346-design.md
EvgeniiMekhanik: I mean that we pass delta ema. So we calculate changing of cpu usage.

1346-design.md
EvgeniiMekhanik:
```
static inline void
__calculate_mean_and_std(struct stats *s)
{
	s64 variance;
	u64 total_sum = 0;
	u64 total_sumsq = 0;
	u32 num_clients = 0;

	num_clients = percpu_counter_sum(&s->num);
	total_sum = percpu_counter_sum(&s->sum);
	total_sumsq = percpu_counter_sum(&s->sumsq);

	if (!unlikely(num_clients))
		return;

	s->mean = (total_sum << SCALE_SHIFT) / num_clients;
	variance = ((total_sumsq << SCALE_SHIFT) / num_clients) -
		((s->mean * s->mean) >> SCALE_SHIFT);
	s->std = int_sqrt64(variance << SCALE_SHIFT);
}
```
we calculate mean and std at the end of the training mode, using previously calculated sum and sumsq

1346-design.md
EvgeniiMekhanik: I don't get why we need modulos? If z-score < 0 I think we should not block such clients!

1346-design.md
EvgeniiMekhanik: delta = curr - old_max;
s->sum += delta (per cpu)
n (not always +1), because we increment it only when the client is new!
total_sum = percpu_counter_sum(&s->sum);
s->mean = (total_sum << SCALE_SHIFT) / num_clients;

1346-design.md
EvgeniiMekhanik: As I think we calculate mean and std at the end of the training mode (during switching to defence mode).
Then on each new value we calculate *z_score = ((s64)(val << SCALE_SHIFT) - s->mean) / s->std;
and compare it with configured threshold

1346-design.md
EvgeniiMekhanik: We can have situation when curr value became < 0 (we zeroed all history on new training). So we need epoch and lock for correct synchronization between epoch

1346-design.md
krizhanovsky: Discussed on today call that this solution would be only for HTTP/2 framing attacks, not a generic CPU attacks in sense of #488 (e.g. think about ReDDoS or parser-specific attacks)

1346-design.md
```diff
@@ -0,0 +1,305 @@
...
+ *					  new training epoch starts.
+ * max				- maximum observed value of the tracked metric within the
+ *					  current training epoch (e.g. peak number of in-flight
+ *					  non-idempotent requests);
+ * @epoch			- training epoch identifier. Compared against the global
+ *					  @g_training_epoch to detect epoch change and trigger
+ *					  reinitialization of @max and @counter.
+ */
+typedef struct tfw_client_req_counter_t {
+	struct tfw_client_req_counter_t	*next_free;
```
const-t: Didn't get this, it will be `union`?

1346-design.md
```diff
@@ -104,40 +104,117 @@ Each connection tagged with training_epoch (we add new field to `tempesta_sock`
...
+	 * curr value, so we need syncronization here.
+	 */
+	do {
+		if (curr <= old_max)
+			return true;
+	} while (!atomic_try_cmpxchg(&cli_req_counter->max, &old_max, curr));
+
+	delta1 = curr - old_max;
+	delta2 = (u64)curr * curr - (u64)old_max * old_max;
+	tfw_training_mode_adjust_req_num(delta1, delta2);
```
const-t: This thing is unclear for me. `tfw_training_mode_adjust_req_num()` adjust per-client `sum` and `sumsq`?

1346-design.md
```diff
@@ -0,0 +1,407 @@
+**Training Mode:**
+During the training phase the system collects per-client metrics for each new event and aggregates global statistics required for subsequent z-score calculation in defence mode. For each client only the current maximum number of connections/requests/memory usage is stored. Global statistics include:
+
+- number of samples ("n", i.e. number of active clients),
+- sum of values ("sum"),
+- sum of squared values ("sumsq").
+
+At the end of the training phase the mean and standard deviation are computed and later used for z-score calculation during anomaly detection.
+    sum
+μ = ─── -> mean
```
const-t: This formula can't be viewed in markdown view.

const-t (commented):
I would suggest little bit adding structure to the document. Split into several chapters, at least one chapter for each tracking resource(memory, cpu, request). Maybe math also can be in another chapter, because it is common. Please separate chapters using headers - "##" and "###" in markdown.

benchmark_training/128bit.h
```diff
@@ -0,0 +1,506 @@
+/**
```
const-t: Do we need one more library? I didn't get why we can't use [GCC's __int128 extension ](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fint128.html), linux kernel has Kconfig constant `CONFIG_ARCH_SUPPORTS_INT128` (see `arch/x86/Kconfig`) that indicates support of `__int128` for an architecture and it is enabled for x86_64(this is the only target arch for Tempesta FW) if GCC supports it. See `CC_HAS_INT128`.

1346-design.md
```diff
@@ -0,0 +1,771 @@
...
+```
+This allows CPU costs associated with response handling to be attributed to the corresponding client. Unlike connection, request, and memory tracking, CPU accounting uses an exponentially weighted moving average rather than a simple counter. This provides a smoothed estimate of recent CPU consumption while reducing sensitivity to short processing spikes. The EMA is updated using the following helper:
+```C
+static inline void
+tfw_adaptive_limits_counter_add_ema(TfwAdaptiveLimitLock *limit, int delta)
+{
+	s64 *ema = this_cpu_ptr(limit->counter);
+	static const unsigned int ema_alpha_shift = 4;
+
+	*ema += ((s64)delta - *ema) >> ema_alpha_shift;
```
const-t, line 736: This doesn't look like classic EMA, should we call this EMA? Maybe this algorithm is suitable for us, but I can't see any proofs.

EvgeniiMekhanik: https://wikimedia.org/api/rest_v1/media/math/render/svg/62d45608a6924df932d47051f44465e2096d431c
ema = a * xt + (1 - a) ema_old       (st = a * xt + (1 - a) st-1 wiki formula)
ema = a * xt + ema_old - a* ema_old = ema_old + a * (xt - ema_old) (my formula)

const-t: yes, you are right, my mistake

const-t:
@krizhanovsky @EvgeniiMekhanik This design seems too detailed to me, and the amount of code and implementation detail makes it unnecessarily complicated. Also it would better to add examples of use cases and how to use the feature.

PR #2660, opened 2026-05-21
Fix incorrect server connection reference counting

### LINKED ISSUE #2658: Sometimes server connection reference counter became -1.
**Describe the issue**
Sometimes server connection reference counter became -1.

**To Reproduce**
run tests.deadtime in the loop 

**Version or commit hash**
19dd826fcc88f045f6917a13b5e1a66d32423215
but I think we have the same problem in 0.8

**Stacktrace or debug log**
We don't have any trace, because  we check BUG_ON(rc < TFW_CONN_DEATHCNT); and do not check if rc == -1.
We also delete server connection only when Tempesta FW shutdown and this bug is very rare, so we don't have a trace.

EvgeniiMekhanik:
Implement special function to set correct server connection reference counter, when new server connection established. When we try to establish new connection we set connection reference counting using `tfw_srv_conn_init_as_dead`. If server is shutdown we increment connection reference counter (tfw_sock_srv_disconnect-> __tfw_connection_get_if_not_death). If connection will be established on other cpu at the same time and we just set connection reference counter to 1 using `tfw_connection_revive`, later we put connection reference counter in tfw_sock_srv_disconnect and release such connection. Later when we drop connection, we decrement connection reference counter again, so it became -1. Releasing connection before dropping can lead to different bugs. So here we adjust all previous connection reference counter incrementation.

Closes #2658

### DIFF
--- a/fw/connection.h
+++ b/fw/connection.h
@@ -482,7 +482,7 @@ tfw_connection_put(TfwConn *conn)
 		return;
 
 	rc = atomic_dec_return(&conn->refcnt);
-	BUG_ON(rc < TFW_CONN_DEATHCNT);
+	BUG_ON(rc == -1 || rc < TFW_CONN_DEATHCNT);
 
 	if (likely(rc && rc != TFW_CONN_DEATHCNT))
 		return;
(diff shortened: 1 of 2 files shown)


const-t (approved):
LGTM

PR #2663, opened 2026-06-09
Trainging mod implementation (WIP)

EvgeniiMekhanik:


fw/training.c
```diff
@@ -0,0 +1,626 @@
...
+	 * pointer - we access this pointer under `rcu`. During switching modes
+	 * we first of all disable trainging and then call `synchronize_rcu`,
+	 * so if `tfw_disable_training_or_defence` is called on other cpu, it
+	 * will wait until we finish to collect statistic. If it was called (and
+	 * finished before this function call), `tfw_mode_is_disabled` returns
+	 * true here.
+	 */
+	if (likely(!tfw_mode_is_disabled())) {
+		s = rcu_dereference(g_stats);
+		percpu_counter_add(&s->sum, delta1);
```
const-t: What a reason to use `percpu_counter` instead of simple per-cpu var? `percpu_counter` pretty large and has overhead, must be a reason to use it.

fw/training.h
```diff
@@ -0,0 +1,181 @@
+/**
```
const-t, line 1: I suggest renaming this to `adaptive_limits.c` or similar and use word "training" only in sense of "training mode" as the state of the adaptive limits.

fw/client.h
```diff
@@ -23,41 +23,100 @@
...
+ * @epoch	- training epoch identifier. Compared against the global
+ *		  @g_training_epoch to detect epoch change and trigger
+ *		  reinitialization of @max and @counter.
+ */
+typedef struct tfw_client_counter_t {
+	spinlock_t		lock;
+	atomic_long_t		max;
+	s64 	__percpu	*counter;
+	u16			epoch;
+} TfwClientCounter;
```
const-t: From my point of view we should move this to training.h. All other related structs as well

fw/client.c
```diff
@@ -552,6 +898,310 @@ tfw_client_obtain(TfwAddr addr, TfwAddr *xff_addr, TfwStr *user_agent,
...
+	} while (!atomic64_try_cmpxchg(&counter->max, &old_max, curr));
+
+	*delta1 = curr - old_max;
+	*delta2 = (u64)curr * curr - (u64)old_max * old_max;
+
+	return true;
+}
+
+static bool
+tfw_client_counter_training_check(TfwClientCounter *counter,
```
const-t: It seems `client.c` not the right place for this function. I would prefer to have it in training.c

fw/client.c
```diff
@@ -552,6 +898,310 @@ tfw_client_obtain(TfwAddr addr, TfwAddr *xff_addr, TfwStr *user_agent,
...
+
+	if (tfw_mode_is_disabled())
+		return true;
+
+	curr = tfw_client_counter_get(counter);
+	if (tfw_mode_is_defence())
+		return defence(curr);
+
+	if (tfw_client_counter_change_max(counter, curr, &delta1, &delta2))
+		adjust_num(delta1, delta2);
```
const-t: I would suggest moving update of the global stats to the `tfw_http_conn_recv_finish()`, we don't need live update of the counter during training

const-t (commented):
I see that PR is WIP, but I have few comments for the future.

fw/training.c
EvgeniiMekhanik: Yes fixed

fw/training.h
EvgeniiMekhanik: Yes fixed

fw/client.h
EvgeniiMekhanik: Yes fixed

fw/client.c
EvgeniiMekhanik: fixed

fw/client.c
EvgeniiMekhanik: fixed

fw/adaptive_limits.c
```diff
@@ -0,0 +1,1006 @@
...
+#include "client.h"
+#include "http_limits.h"
+#include "tempesta_fw.h"
+#include "lib/fault_injection_alloc.h"
+#include "lib/128bit.h"
+
+/* Training period in seconds. Zero means disabled. */
+static unsigned int tfw_training_mode_period = 0;
+unsigned int tfw_adaptive_limits_mode =
+	TFW_ADAPTIVE_LIMITS_MODE_IS_DISABLED;
```
const-t, line 30:
```suggestion
unsigned int tfw_adaptive_limits_mode = TFW_ADAPTIVE_LIMITS_MODE_IS_DISABLED;
```

fw/adaptive_limits.c
```diff
@@ -0,0 +1,1006 @@
...
+
+	if (!unlikely(num_clients))
+		return;
+
+	/*
+	 * Use fixed-point scaling only if the accumulated sum is small enough.
+	 * Once the sum approaches the u32 limit, disable scaling to prevent
+	 * arithmetic overflow in subsequent calculations.
+	 */
+	s->scale_shift = total_sum < U32_MAX ? SCALE_SHIFT : 0;
```
const-t, line 286:
```suggestion
	s->scale_shift = total_sum < U32_MAX ? SCALE_SHIFT : 0;
```

fw/adaptive_limits.c
```diff
@@ -0,0 +1,1006 @@
...
+
+/*
+ * Allocate and initialize stats structure.
+ * Returns NULL on failure.
+ */
+static inline struct stats *
+__alloc_stats(void)
+{
+	struct stats *s;
+	gfp_t flags = GFP_KERNEL | __GFP_ZERO;
```
const-t, line 76:
```suggestion
	gfp_t flags = GFP_KERNEL | __GFP_ZERO;
```

fw/adaptive_limits.c
```diff
@@ -0,0 +1,1006 @@
...
+	 * Since total_sum < U32_MAX implies total_sumsq < U64_MAX,
+	 * the fixed-point left shift cannot overflow.
+	 */
+	tmp1 = total_sumsq << s->scale_shift;
+	tmp1 = u128_div_u32(tmp1, num_clients);
+	s->mean = (total_sum << s->scale_shift) / num_clients;
+	tmp2 = (u128)s->mean * (u128)s->mean;
+	tmp2 = tmp2 >> s->scale_shift;
+	variance = tmp1 - tmp2;
+	s->std = u128_sqrt(variance << s->scale_shift);
```
const-t, line 297: I would suggest getting rid of tmp vars to improve readability
```suggestion
	/*
	 * Since total_sum < U32_MAX implies total_sumsq < U64_MAX,
	 * the fixed-point left shift cannot overflow.
	 */
	total_sumsq <<= s->scale_shift;
	total_sum <<= s->scale_shift;
	
	s->mean = total_sum / num_clients;
	mean_sq = ((u128)s->mean * (u128)s->mean) >> s->scale_shift;
	mean_of_sq = u128_div_u32(total_sumsq, num_clients);
	variance = mean_of_sq - mean_sq;
	s->std = u128_sqrt(variance << s->scale_shift);
```

fw/adaptive_limits.c
```diff
@@ -0,0 +1,1006 @@
...
+	 * We store `mean` and `std` values in scaled format, so
+	 * we should convert `val` to scaled format also.
+	 */
+	*z_score = ((s64)(val << s->scale_shift) - (s64)s->mean) / (s64)s->std;
+
+	return true;
+}
+
+static inline void
+tfw_adaptive_limits_adjust_new_client(struct stats __rcu *g_stats)
```
const-t, line 316: Maybe would be better:
```suggestion
tfw_adaptive_limits_inc_num_smaples(struct stats __rcu *g_stats)
```
or
```suggestion
tfw_adaptive_limits_inc_num_clients(struct stats __rcu *g_stats)
```

fw/adaptive_limits.c
```diff
@@ -0,0 +1,1006 @@
...
+	 * rcu pointer dereference should be done under rcu lock,
+	 * to prevent memory corruption.
+	 */
+	BUG_ON(!rcu_read_lock_held());
+	s = rcu_dereference(g_stats);
+	this_cpu_add(*s->num, 1);
+}
+
+static void
+tfw_adaptive_limits_adjust_conn_new_client(void)
```
const-t, line 330:
```suggestion
tfw_adaptive_limits_inc_num_samples_conn(void)
```

```suggestion
tfw_adaptive_limits_inc_num_samples_req(void)
```

....

fw/adaptive_limits.c
```diff
@@ -0,0 +1,1006 @@
...
+}
+
+static void
+tfw_adaptive_limits_adjust_cpu_new_client(void)
+{
+	return tfw_adaptive_limits_adjust_new_client(g_cpu_num);
+}
+
+static inline void
+tfw_adaptive_limits_adjust_new_el(struct stats __rcu *g_stats, u64 delta1,
```
const-t, line 354:
```suggestion
tfw_adaptive_limits_add_sum_sumsq(struct stats __rcu *g_stats, u64 delta1,
```

fw/adaptive_limits.c
```diff
@@ -0,0 +1,1006 @@
...
+	 */
+	BUG_ON(!rcu_read_lock_held());
+	s = rcu_dereference(g_stats);
+	this_cpu_add(*s->sum, delta1);
+	/* `this_cpu_add` is not implemented for 128-bit value. */
+	*this_cpu_ptr(s->sumsq) += delta2;
+}
+
+static void
+tfw_adaptive_limits_adjust_conn_num(u64 delta1, u128 delta2)
```
const-t, line 371:
```suggestion
tfw_adaptive_limits_add_sum_sumsq_conn(u64 delta1, u128 delta2)
```

```suggestion
tfw_adaptive_limits_add_sum_sumsq_req(u64 delta1, u128 delta2)
```

fw/adaptive_limits.h
```diff
@@ -0,0 +1,249 @@
...
+	u16			epoch;
+} TfwAdaptiveLimitLock;
+
+int tfw_adaptive_limits_init(void);
+void tfw_adaptive_limits_exit(void);
+
+bool tfw_adaptive_limits_check_conn_num(TfwAdaptiveLimit *limit, int delta,
+					u16 *epoch);
+void tfw_adaptive_limits_acc_req_num(TfwAdaptiveLimitLock *limit,
+				     int delta, u16 *epoch);
```
const-t, line 188:
```suggestion
void tfw_adaptive_limits_acc_req_num(TfwAdaptiveLimitLock *limit, int delta, u16 *epoch);
```

fw/adaptive_limits.h
```diff
@@ -0,0 +1,249 @@
...
+void tfw_adaptive_limits_acc_mem(TfwAdaptiveLimitLock *limit,
+				 int delta, u16 *epoch);
```
const-t, line 190:
```suggestion
void tfw_adaptive_limits_acc_mem(TfwAdaptiveLimitLock *limit, int delta, u16 *epoch);
```

fw/http.c
```diff
@@ -1357,13 +1358,27 @@ tfw_http_conn_nip_reset(TfwSrvConn *srv_conn)
 		clear_bit(TFW_CONN_B_HASNIP, &srv_conn->flags);
 }
 
+static inline void
+tfw_http_adjust_nip_req(TfwHttpReq *req, int delta)
```
const-t, line 1362: Could you make meaningful name for this function, I can't see any adjustments.

fw/pool.c
```diff
@@ -111,8 +112,9 @@ tfw_pool_free_pages(TfwClientMem *cli_mem, unsigned long addr,
 	pgn = this_cpu_ptr(&pg_next);
 	refcnt = page_count(virt_to_page(addr));
 
-	if (cli_mem)
-		tfw_client_adjust_mem(cli_mem, -(PAGE_SIZE << order));
+	if (cli_mem) {
+		tfw_client_adjust_mem(cli_mem, -(PAGE_SIZE << order), epoch);
+	}
```
const-t, line 117:
```suggestion
	if (cli_mem) 
		tfw_client_adjust_mem(cli_mem, -(PAGE_SIZE << order), epoch);
	
```

fw/adaptive_limits.c
```diff
@@ -0,0 +1,1006 @@
...
+		return r;
+
+	__upgrade_all_stats(new_conn_num, new_req_num, new_mem_num,
+			    new_cpu_num);
+
+	return 0;
+}
+
+static inline int
+__init_z_score(void)
```
const-t, line 240:
```suggestion
setup_training(void)
```
or another name, but `__init_z_score` refers to z-score math calculations

fw/adaptive_limits.c
const-t, line 297: In some cases `s->std` can be zero and it means that training is not successful, maybe we should add warning for this?

fw/adaptive_limits.c
```diff
@@ -0,0 +1,1006 @@
...
+__calculate_z_score(u64 val, struct stats *s, s64 *z_score)
+{
+	if (unlikely(!s->std))
+		return false;
+
+	/*
+	 * We store `mean` and `std` values in scaled format, so
+	 * we should convert `val` to scaled format also.
+	 */
+	*z_score = ((s64)(val << s->scale_shift) - (s64)s->mean) / (s64)s->std;
```
const-t, line 310: `(s64)s->mean` should we cast mean to signed? It may overflow it, whay is it better than use unsigned value?

fw/adaptive_limits.c
```diff
@@ -0,0 +1,1006 @@
...
+	 * If training mode has already completed and the system has switched
+	 * to defence mode, or training was never started, the timer is already
+	 * inactive and no further action is required.
+	*/
+	if (del_timer_sync(&training_timer))
+		tfw_adaptive_limits_prepare_for_defence();
+}
+
+static inline int
+tfw_training_start(void)
```
const-t, line 861: I believe we must add log(dmesg) message about starting training and defense periods. It is hard to use without feedback from the Tempesta.

fw/main.c
```diff
@@ -390,6 +391,47 @@ tfw_ctlfn_state_io(const struct ctl_table *ctl, int is_write,
...
+	unsigned int mode;
+	int r;
+
+	mutex_lock(&tfw_sysctl_mtx);
+
+	if (is_write) {
+		/*
+		 * There are only two available modes to set:
+		 * 0 - defence mode;
+		 * 1 - training mode;
```
const-t, line 411: We have only two modes training and defence, it seems we need mode in which Tempesta is already trained, but limits are not applied to the client. Otherwise I can't imagine how it can be used in following case: Tempesta in training mode, someone started ddos, and now Tempesta under load and trains on this load, then admin or monitoring system has detected ddos and started defense, but Tempesta has spent some time under load, thus it may allow more load in defense than usual that might be not enough for ddos mitigation.

const-t (commented):
In general PR looks good, but sometimes it was hard to read for me, thus most of my comments are related to refactoring.

PR #2665, opened 2026-06-30
Reuse kernel TCP keepalive timer for Tempesta client connections

EvgeniiMekhanik:


fw/sock_clnt.c
```diff
@@ -157,37 +127,13 @@ tfw_cli_conn_release(TfwCliConn *cli_conn)
...
 	int r;
 
 	tfw_connection_get((TfwConn *)cli_conn);
 	r = tfw_connection_send((TfwConn *)cli_conn, msg);
-
-	/*
-	 * If connection was shut down keepalive timer was set
-	 * in `tfw_connection_send`.
-	 */
-	tfw_cli_conn_mod_timer(cli_conn, tfw_cli_cfg_ka_timeout);
```
const-t, line 189: As I understand, the main idea of the timer replacement is to save current behavior, but implement it using kernel's timer handler. But this patch changes behavior of timer and mixes keep-alive of transport and application layers.

key points:
1. Consider to use `sk->sk_timer` for timer handler.
2. We must rearm timer ONLY in `tfw_cli_conn_send()` to have the same behavior as before this patch. (in this patch we don't rearm timer at all).
3. By default TCP keep-alive is disabled, I believe we can completely override tcp KA logic for Tempesta's socket and rely only on the same logic as Tempesta has now. KA timer of application layer always has lower interval than transport so it is enough. In this patch logic from `tcp_keepalive_timer()` is applied before Tempesta's. For instance:
```
if (tp->packets_out || !tcp_write_queue_empty(sk))
		goto resched;
```

const-t (changes requested):

EvgeniiMekhanik:
Drop according private discussion. TCP keepalive timer is not used by default, so we save only some memory for struct timer using this patch.

PR #2666, opened 2026-07-03
Fix NULL dereference when vhost is not found for health monitor requests

EvgeniiMekhanik:
The `http_tbl` module is initialized after the `sock_srv` module. When the number of `srv_group` entries is large and each group contains a `health` directive, Tempesta FW may start sending health monitor requests before the `active_table` is initialized.

In this case, req->vhost is not set (NULL), which later leads to a NULL dereference in the frang module.

This patch prevents scheduling health monitor requests when vhost is NULL, avoiding the crash instead of sending such requests.


const-t (approved):
LGTM

Issue #2673, opened 2026-07-27
kernel BUG at /root/tempesta/fw/sock_clnt.c:140


EvgeniiMekhanik:
**To Reproduce**
tests.fault_injection.test_fault_injection_base.TestStress.test_stress_pool_alloc_h2load

**Version or commit hash**
tempesta 43637c4

**Stacktrace or debug log**
[ 3302.463816] name fail_function, interval 0, probability 10, space 0, times -1
Jul 26 02:58:56 192.168.122.93 [ 3302.463907] ------------[ cut here ]------------
Jul 26 02:58:56 192.168.122.93 [ 3302.463924] WARNING: CPU: 0 PID: 16 at /root/tempesta/fw/http2.c:330 tfw_h2_context_clear+0xb3/0xc0 [tempesta_fw]
Jul 26 02:58:56 192.168.122.93 [ 3302.463974] Modules linked in: tempesta_fw(OE) tempesta_regex(OE) tempesta_db(OE) tempesta_tls(OE) tempesta_lib(OE) tcp_diag inet_diag nf_conntrack_netlink xt_nat xt_tcpudp veth ib_core tls xt_conntrack xt_MASQUERADE bridge stp llc xfrm_user xfrm_algo xt_set ip_set nft_chain_nat nf_nat nf_conntrack nf_defrag_ipv6 nf_defrag_ipv4 xt_addrtype nft_compat nf_tables overlay cfg80211 binfmt_misc intel_rapl_msr intel_rapl_common skx_edac_common nfit kvm_intel kvm rapl i2c_i801 i2c_mux i2c_smbus lpc_ich input_leds joydev mac_hid serio_raw sch_fq_codel dm_multipath efi_pstore netconsole msr drm nfnetlink dmi_sysfs qemu_fw_cfg ip_tables x_tables autofs4 btrfs blake2b_generic raid10 raid456 async_raid6_recov async_memcpy async_pq async_xor async_tx xor raid6_pq libcrc32c raid1 raid0 crct10dif_pclmul crc32_pclmul ghash_clmulni_intel sha512_ssse3 psmouse sha256_ssse3 ahci virtio_net libahci virtio_blk net_failover virtio_rng failover aesni_intel crypto_simd cryptd [last unloaded: tempesta_lib(OE)
Jul 26 02:58:56 192.168.122.93 ]
[...]

PR #2678, opened 2026-08-02
Mekhanik evgenii/fix 2673

EvgeniiMekhanik:


fw/http_stream.c
```diff
@@ -273,8 +273,17 @@ tfw_h2_stream_unlink_nolock(TfwH2Ctx *ctx, TfwStream *stream)
 	tfw_h2_stream_del_from_queue_nolock(stream);
 
 	if (hmreq) {
+		/*
+ 		 * Once hmreq->stream is cleared, the request may be freed
+		 * on another CPU by `tfw_http_on_tcp_entail_req`. Cache the
+		 * flag value before breaking the request-to-stream association.
```
krizhanovsky: Please add a comment that x86-64 TSO keeps load-before-store guarantee of this flag against `tfw_http_conn_msg_free()`, so we don't need a barrier

fw/http_types.h
```diff
@@ -121,6 +121,9 @@ enum {
 	 */
 	TFW_HTTP_B_REQ_HEAD_TO_GET,
 
+	/* Flag indicates, that request was already sent to server. */
+	TFW_HTTP_B_REQ_SENT,
```
krizhanovsky, line 125: Maybe there are no other opportunities, but introducing a new flag for only one set and test looks like a strut. It'd be good to review the logic.

fw/http.c
```diff
@@ -2932,6 +2934,14 @@ tfw_http_resp_pair(TfwHttpMsg *hmresp)
 		goto end;
 
 	list_for_each_entry(req, &srv_conn->fwd_queue, fwd_list) {
+		/*
+		 * Responses may only be paired with requests that have
+		 * already been queued for transmission. Receiving a
+		 * response before that is a backend protocol violation.
+		 */
+		if (!test_bit(TFW_HTTP_B_HMONITOR, req->flags)
+		    && !test_bit(TFW_HTTP_B_REQ_SENT, req->flags))
```
krizhanovsky, line 2963: I have concern about the whole change c10dbfa052b56bca7c4d1c3ef476a87d14d820c2 : 

> This may happen because of a server implementation bug or malicious behavior.

I'm wondering how did we encounter the problem?

Do we need to do anything for buggy upstreams?  Why don't we log that there is an upstream problem?

Malicious could mean request splitting attack, when two or more requests are hidden in one request, so a proxy sees 1 request, but backend sees many. We can not just silently recover - it's good to reset connections, but we must log the incident. Please have a look - we already have a logic for the attack

fw/sock_clnt.c
```diff
@@ -356,19 +357,21 @@ tfw_cli_conn_on_shutdown(TfwConn *conn)
...
-	.connection_drop	= tfw_sock_clnt_drop,
-	.connection_recv	= tfw_tls_connection_recv,
-	.connection_recv_finish = tfw_connection_recv_finish,
-	.connection_on_shutdown	= tfw_cli_conn_on_shutdown,
+	.connection_new			= tfw_sock_clnt_new,
+	.connection_drop		= tfw_sock_clnt_drop,
+	.connection_recv		= tfw_tls_connection_recv,
+	.connection_recv_finish 	= tfw_connection_recv_finish,
+	.connection_on_shutdown		= tfw_cli_conn_on_shutdown,
+	.connection_get_incoming_cpu	= tfw_connection_get_incoming_cpu,
```
krizhanovsky: Websocket backend connections all fail to CPU 0 as `tfw_ws_srv_ss_hooks` doesn't define the `connection_get_incoming_cpu` callback. Also check all other users of the callbacks.

fw/http.c
```diff
@@ -1464,6 +1464,8 @@ tfw_http_on_tcp_entail_req(void *conn, struct sk_buff *skb)
 		return -EINVAL;
 	}
 
+	set_bit(TFW_HTTP_B_REQ_SENT, req->flags);
```
krizhanovsky, line 1467: Why the flag is never cleared? If a server connection dies and the request is rescheduled, it's still keeps the flag - conditions for the request have changed, but the flag persists - this doesn't seem right

fw/sock.c
```diff
@@ -640,7 +640,11 @@ ss_send(struct sock *sk, struct sk_buff **skb_head, int flags)
 	if (unlikely(!*skb_head))
 		return 0;
 
-	cpu = sk->sk_incoming_cpu;
+	cpu = SS_CALL(connection_get_incoming_cpu, sk->sk_user_data);
+        if (cpu != sk->sk_incoming_cpu) {
```
krizhanovsky: Here and at the below space identation instead of tabs

krizhanovsky (changes requested):
I have several concerns on the change

fw/http.c
const-t, line 1467: It seems we should set this flag only when the last skb of the request has been moved to `write_queue` otherwise we may get a bug in case, when the first skb is sent, few skbs of the request still in queue, sending postponed due to lack of the widnow, but flag is set.

fw/sock_clnt.c
EvgeniiMekhanik: yes fixed

fw/sock.c
EvgeniiMekhanik: yes fixed

fw/http.c
EvgeniiMekhanik: No! The problem occurs because we try to access `req` structure in appropriate callback. And if Tempesta FW received unexpected response before this callback invoked request can be already deleted. We call appropriate callback only for the first request skb in the `write_queue`

EvgeniiMekhanik: yes fixed

fw/http_stream.c
EvgeniiMekhanik: Yes fixed

fw/http.c
EvgeniiMekhanik: The purpose of this commit is to protect a request during the interval between its submission for sending and the execution of `tfw_http_on_tcp_entail_req()`.

Previously, once a request was submitted for sending, we did not access it again until a response was received. Extra or unexpected responses were handled within the same processing path by checking:

```c
if (req == (TfwHttpReq *)srv_conn->curr_msg_sent)
```

This assumption changed with the introduction of the `tfw_http_on_tcp_entail_req()` callback. We now access the request again after `ss_do_send()`. The callback is executed asynchronously through a workqueue when the TCP send window becomes available, so there may be a delay between submitting the request and executing the callback.

During this interval, the request must remain valid.

Consider the following sequence:

1. Request `req` is submitted for sending and `srv_conn->curr_msg_sent == req`.
2. `tfw_http_on_tcp_entail_req()` has not been called yet (it will be called after passing through work queue from `ss_tx_action` and when TCP window will be enough).
3. The backend sends a response that we associate with `req`, even though `req` has not actually reached the point where `tfw_http_on_tcp_entail_req()` is called (unexpected response!).
4. While processing that response, we may remove and eventually free `req`.
5. Later, the pending `tfw_http_on_tcp_entail_req()` callback runs and accesses the already freed request, resulting in a use-after-free/crash.

I actually encountered this scenario when responses became reordered after changes involving `sk->sk_incoming_cpu`. Requests were queued in the order `1, 2`, but request `2` was transmitted before request `1`. The backend therefore returned a response to request `2`, while we interpreted it as a response to request `1` and removed request `1`. Later, `tfw_http_on_tcp_entail_req()` was invoked for request `1` and accessed the already freed object.

The new flag closes this lifetime gap. It is set when `tfw_http_on_tcp_entail_req()` has been executed for the request. Until the flag is set, we must not accept a backend response as belonging to that request.

Therefore, if a response arrives for a request whose flag has not yet been set, we treat it as an invalid/unexpected response, report:

```c
T_WARN("Paired request missing, HTTP Response Splitting attack?\n");
```

and terminate the server connection instead of removing a request that may still be referenced by the pending callback.

const-t (approved):
LGTM

scripts/tfw_lib.sh
```diff
@@ -104,7 +104,7 @@ irqbalance_ban_irqs()
 # This function prepares cpu mask for RSS and RPS.
```
krizhanovsky, line 104: Please update the copyright year

fw/sock.c
krizhanovsky: The initial idea behind using `sk->sk_incoming_cpu` and scheduling jobs to the appropriate CPU was to avoid socket locking by many CPUs along with cache contention on socket data. This change breaks this concept.

Why do we need this change if we already have the reordering fix with TFW_HTTP_B_REQ_SENT? Is this change required only for RPS?

fw/http_stream.c
```diff
@@ -273,8 +273,29 @@ tfw_h2_stream_unlink_nolock(TfwH2Ctx *ctx, TfwStream *stream)
...
+		 * request itself.
+		 * The flag load must happen before the store which clears
+		 * `hmreq->stream`. On x86-64, TSO preserves Load->Store ordering,
+		 * so the CPU cannot make the hmreq->stream store visible before
+		 * the preceding flag load. Therefore, no explicit memory barrier
+		 * is required here.
+		 */		
+		bool req_is_fully_parsed =
+			test_bit(TFW_HTTP_B_FULLY_PARSED, hmreq->flags);
+
```
krizhanovsky, line 295: [AI review] x86-64 is OK, but a compiler barrier is required here

fw/sock.c
```diff
@@ -640,7 +640,11 @@ ss_send(struct sock *sk, struct sk_buff **skb_head, int flags)
 	if (unlikely(!*skb_head))
 		return 0;
 
-	cpu = sk->sk_incoming_cpu;
+	cpu = SS_CALL(connection_get_incoming_cpu, sk->sk_user_data);
```
krizhanovsky: [AI review] `ss_send()` (and probably `ss_close()`) can be called on a dying socket (this is why we have live check with `ss_sock_active()` at the below and hold the socket reference count before pushing it to the work queue), so `SS_CALL()` may return `0` and we treat this as a CPU number and schedule a work on the CPU.

fw/sock_clnt.c
```diff
@@ -263,6 +263,7 @@ tfw_sock_clnt_new(struct sock *sk)
 		sk->sk_write_xmit = tfw_tls_encrypt;
 	}
 	sk->sk_fill_write_queue = tfw_sock_clnt_fill_write_queue;
+	conn->incoming_cpu = sk->sk_incoming_cpu;
```
krizhanovsky: [AI review] `conn->incoming_cpu` is initialize after publishing the connection with `tfw_connection_link_peer()`. The same problem in sock_srv.c.

AK: In my understanding we don't have any activity on the client connection until we initialize it in this function and receive a first request. For the server connections we call `tfw_sock_srv_connection_revive()` after the assignment, so the connection becomes available for the rest of the system only after setting the CPU. Probably AIs (both Codex and Claude) are wrong, but please check. Maybe it's worth commenting.

fw/sock.c
```diff
@@ -640,7 +640,11 @@ ss_send(struct sock *sk, struct sk_buff **skb_head, int flags)
...
+	if (cpu != sk->sk_incoming_cpu) {
+		T_DBG("Socket incoming cpu was changed from %d to %d",
```
krizhanovsky:
```suggestion
		T_DBG("Socket incoming cpu was changed from %d to %d\n",
```

fw/sock.c
```diff
@@ -912,8 +916,11 @@ ss_close(struct sock *sk, int flags)
 	if (unlikely(!sk))
 		return T_OK;
 
-	ss_sk_incoming_cpu_update(sk);
-	cpu = sk->sk_incoming_cpu;
+	cpu = SS_CALL(connection_get_incoming_cpu, sk->sk_user_data);
+	if (cpu != sk->sk_incoming_cpu) {
+		T_DBG("Socket incoming cpu was changed from %d to %d",
```
krizhanovsky:
```suggestion
		T_DBG("Socket incoming cpu was changed from %d to %d\n",
```

fw/http_stream.c
```diff
@@ -273,8 +273,29 @@ tfw_h2_stream_unlink_nolock(TfwH2Ctx *ctx, TfwStream *stream)
...
+		 * Cache TFW_HTTP_B_FULLY_PARSED before clearing hmreq->stream,
+		 * while hmreq is still guaranteed to be valid. The value is used
+		 * below to decide whether this path has to free an incomplete
+		 * request itself.
+		 * The flag load must happen before the store which clears
+		 * `hmreq->stream`. On x86-64, TSO preserves Load->Store ordering,
+		 * so the CPU cannot make the hmreq->stream store visible before
+		 * the preceding flag load. Therefore, no explicit memory barrier
+		 * is required here.
+		 */
```
krizhanovsky: [AI review]: white spaces after `*/`

scripts/tfw_lib.sh
```diff
@@ -199,15 +199,18 @@ distribute_queues()
...
 	irq0=${irqs[0]}
-	for i in ${irqs[@]}; do
-		# Wrap around CPU mask if number of queues is
-		# larger than CPUS_N.
-		if [[ $(calc "$i - $irq0") -gt $CPUS_N ]]; then
-			irq0=$i;
+	cpu=0
+
+	for irq in "${irqs[@]}"; do
+        	mask=$(make_cpu_rss_mask "$cpu")
```
krizhanovsky: [AI review] spaces instead of tabs

fw/websocket.c
```diff
@@ -143,6 +144,12 @@ tfw_ws_srv_new_steal_sk(TfwSrvConn *srv_conn)
 	conn->peer = (TfwPeer *)srv;
 	conn->sk = srv_conn->sk;
 	conn->destructor = tfw_ws_conn_release;
+	/*
+	 * Hook `connection_new()` is never called, because ws connection
+	 * can only be bootstraped from http connection, so initialized
+	 * `conn->incoming_cpu` here.
+	 */
+	conn->incoming_cpu = srv_conn->sk->sk_incoming_cpu;
```
krizhanovsky:
```suggestion
	conn->incoming_cpu = srv_conn->sk_incoming_cpu;
```
[AI review] semantically, we assign TfwConn->sk_incomming_cpu instead, so we should use the same, provided by the parent connection.

fw/websocket.c
```diff
@@ -143,6 +144,12 @@ tfw_ws_srv_new_steal_sk(TfwSrvConn *srv_conn)
 	conn->peer = (TfwPeer *)srv;
 	conn->sk = srv_conn->sk;
 	conn->destructor = tfw_ws_conn_release;
+	/*
+	 * Hook `connection_new()` is never called, because ws connection
+	 * can only be bootstraped from http connection, so initialized
```
krizhanovsky:
```suggestion
	 * can only be bootstrapped from HTTP connection, so initialized
```

scripts/tfw_lib.sh
```diff
@@ -199,15 +199,18 @@ distribute_queues()
...
+		if ! echo "$mask" > "/proc/irq/$irq/smp_affinity"; then
+			log "Error: cannot bind IRQ $irq to CPU $cpu, mask=$mask"
```
krizhanovsky, line 216: [AI review, Claude only] need to return here as callers compare the result with "0" and the function still passes to `echo "0"`

scripts/tfw_lib.sh
```diff
@@ -199,15 +199,18 @@ distribute_queues()
 	# for device (not assigned to any of the queues).
 	irqs=(${irqs[@]:1})
 	irq0=${irqs[0]}
```
krizhanovsky: [AI review, Claude only] `irq0` isn't used anymore

scripts/tfw_lib.sh
```diff
@@ -199,15 +199,18 @@ distribute_queues()
...
+			log "Error: cannot bind IRQ $irq to CPU $cpu, mask=$mask"
 		fi
-		delta=$(( i - irq0 ))
-		mask=`make_cpu_rss_mask $delta`
-		echo "$mask" > "/proc/irq/$i/smp_affinity"
+
+        	cpu=$((cpu + 1))
+        	if [[ $cpu -ge $CPUS_N ]]; then
+        	        cpu=0
+        	fi
```
krizhanovsky: [AI review, Claude only] These lines and line 205 are idented with spaces instead of tab

fw/sock_clnt.c
```diff
@@ -356,19 +357,21 @@ tfw_cli_conn_on_shutdown(TfwConn *conn)
...
 static const SsHooks tfw_sock_tls_clnt_ss_hooks = {
-	.connection_new		= tfw_sock_clnt_new,
-	.connection_drop	= tfw_sock_clnt_drop,
-	.connection_recv	= tfw_tls_connection_recv,
-	.connection_recv_finish = tfw_connection_recv_finish,
-	.connection_on_shutdown	= tfw_cli_conn_on_shutdown,
+	.connection_new			= tfw_sock_clnt_new,
+	.connection_drop		= tfw_sock_clnt_drop,
+	.connection_recv		= tfw_tls_connection_recv,
+	.connection_recv_finish 	= tfw_connection_recv_finish,
```
krizhanovsky: [AI review] Here and at the line 363 there is a space immediately after `connection_recv_finish`

fw/connection.h
```diff
@@ -108,6 +108,7 @@ typedef struct tfw_conn_t TfwConn;
 #define TFW_CONN_COMMON					\
 	SsProto			proto;			\
 	TfwGState		state;			\
+	int			incoming_cpu;		\
```
krizhanovsky: [AI review, Claude only] `incoming_cpu` should be place net to `refcnt` to about 4 byte hole

krizhanovsky (changes requested):
AI found several problems, probably not so crucial, but my main concern is the breaking CPU locality with movement from sk->sk_incomming_cpu to TfwConn->incomming_cpu. The CPU local processing is an important performance property, which can not be violated. Moreover, this change seems doesn't fix the original #2673 and introduces extra possible issues, reported by AIs.

scripts/tfw_lib.sh
EvgeniiMekhanik: Yes fixed

fw/sock.c
EvgeniiMekhanik: Remove this commit, but I it seems that for RPS concept is already broken. If `sk_incoming_cpu` changed we schedule SS_SEND/SS_CLOSE operations on other CPU. For example sk_incoming_cpu was  2, we schedule SS_SEND on CPU2, but sk_incoming_cpu changed to 3, but `ss_do_send` will be invoked on CPU 2!

fw/http_stream.c
EvgeniiMekhanik: Yes fixed

fw/sock.c
EvgeniiMekhanik: I totally remove this commit according discussion

fw/sock_clnt.c
EvgeniiMekhanik: I totally remove this commit according discussion

fw/sock.c
EvgeniiMekhanik: I totally remove this commit according discussion

fw/sock.c
EvgeniiMekhanik: I totally remove this commit according discussion

fw/http_stream.c
EvgeniiMekhanik: Yes fixed

scripts/tfw_lib.sh
EvgeniiMekhanik: Yes fixed

fw/websocket.c
EvgeniiMekhanik: I totally remove this commit according discussion

fw/connection.h
EvgeniiMekhanik: I totally remove this commit according discussion

fw/sock_clnt.c
EvgeniiMekhanik: I totally remove this commit according discussion

scripts/tfw_lib.sh
EvgeniiMekhanik: Yes fixed

scripts/tfw_lib.sh
EvgeniiMekhanik: Yes fixed

scripts/tfw_lib.sh
EvgeniiMekhanik: Yes fixed

fw/websocket.c
EvgeniiMekhanik: I totally remove this commit according discussion

krizhanovsky (approved):
LGTM

PR #2682, opened 2026-08-16
Add CLAUDE.md AGENTS.md for AI usage

krizhanovsky:
AGENTS.md is used by Codex, Copilot, Cursor and some other AIs. Claude Code requires CLAUDE.md, so just include AGENTS.md.

The guide is pretty general and does not include the things, derived by the tools automatically, but does include the points frequently missed (in my experience) by the tools during new code generation and review.


consuelo2210 (approved):
LGTM: every point sounds reasonable to me. Let’s check how it works in real life.

PR #2684, opened 2026-08-28
Restore IRQ affinity after queue distribution failure

EvgeniiMekhanik:
Save the original IRQ affinity masks before assigning queue IRQs. If an assignment fails, restore every mask changed during the attempt so that fallback to combined queues or RPS starts from a consistent state.


const-t (approved):
LGTM

